PR #143, opened 2019-08-09
Update getting-started.md because of problems with Windows OS

maku:
Because of problems when executing vendure installation on Windows, getting started guide should be updated. 
The installation process needs a specific python version (node-gyp). Installation of **windows-build-tools** ensures this.


michaelbromley:
Hi, thanks for the contribution!

Do you recall which modules caused the error? Off the top of my head I can think of the `sharp` package needing to be built from source.

PR #153, opened 2019-09-06
fix(race-condition): Fix race condition on DB among worker and server

### LINKED ISSUE #152: Race condition when modifying DB and running server & worker processes concurrently.
**Describe the bug**
With `dbConnectionOptions.synchronize: true`, and changes to the DB (e.g. new CustomFields defined), running the server and worker processes concurrently results in errors presumably caused by a race condition as both of them attempt to make the necessary changes to the DB:

```
[run:server] info 9/6/19, 09:23 - [TypeORM] new columns added: customFieldsMayalsolike3id
[run:server] error 9/6/19, 09:23 - [TypeORM] Query error: Error: ER_DUP_FIELDNAME: Duplicate column name 'customFieldsMayalsolike3id', "ALTER TABLE `product` ADD `customFieldsMayalsolike3id` i
nt NULL" -- [undefined] 
[run:server] error 9/6/19, 09:23 - [TypeOrmModule] Unable to connect to the database. Retrying (1)... trace: 
[run:server] QueryFailedError: ER_DUP_FIELDNAME: Duplicate column name 'customFieldsMayalsolike3id'
[run:server]     at new QueryFailedError (C:\Development\vendure\vendure\src\error\QueryFailedError.ts:9:9)
[run:server]     at Query.<anonymous> (C:\Development\vendure\vendure\src\driver\mysql\MysqlQueryRunner.ts:166:37)
[run:server]     at Query.<anonymous> (C:\Development\vendure\vendure\node_modules\mysql\lib\Connection.js:525:10)
[run:server]     at Query._callback (C:\Development\vendure\vendure\node_modules\mysql\lib\Connection.js:491:16)
[run:server]     at Query.Sequence.end (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\sequences\Sequence.js:83:24)
[run:server]     at Query.ErrorPacket (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\sequences\Query.js:90:8)
[run:server]     at Protocol._parsePacket (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\Protocol.js:291:23)
[run:server]     at Parser._parsePacket (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\Parser.js:433:10)
[run:server]     at Parser.write (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\Parser.js:43:10)
[run:server]     at Protocol.write (C:\Development\vendure\vendure\node_modules\mysql\lib\protocol\Protocol.js:38:16)
[run:server] info 9/6/19, 09:23 - [TypeORM] columns changed in "customer". updating: deletedAt
[run:server] info 9/6/19, 09:23 - [TypeORM] columns changed in "product". updating: deletedAt
```
**To Reproduce**
1. Define a bunch of CustomFields in the VendureConfig)
2. Set `synchronize: true`
3. Start both processes with `concurrently`

**Expected behavior**
The DB should get updated only by the server process.

 - @vendure/core version: 0.2.0
 - Nodejs version: 12

**Possible fix**
My immediate idea is to simple enforce that `synchronize` gets always set to `false` for the worker. This can be done internally in the `bootstrapWorker` function, so the end user does not need to think about it.

lsimone:
#closes #152

### DIFF
--- a/packages/dev-server/index-worker.ts
+++ b/packages/dev-server/index-worker.ts
@@ -2,6 +2,10 @@ import { bootstrapWorker } from '@vendure/core';
 
 import { devConfig } from './dev-config';
 
+// https://github.com/vendure-ecommerce/vendure/issues/152
+// fix race condition when modifying DB
+devConfig.dbConnectionOptions = { ...devConfig.dbConnectionOptions, synchronize: false };
+
 bootstrapWorker(devConfig).catch(err => {
     // tslint:disable-next-line
     console.log(err);
--- a/packages/core/src/bootstrap.ts
+++ b/packages/core/src/bootstrap.ts
@@ -2,7 +2,7 @@ import { INestApplication, INestMicroservice } from '@nestjs/common';
 import { NestFactory } from '@nestjs/core';
 import { TcpClientOptions, Transport } from '@nestjs/microservices';
 import { Type } from '@vendure/common/lib/shared-types';
-import { EntitySubscriberInterface } from 'typeorm';
+import { ConnectionOptions, EntitySubscriberInterface } from 'typeorm';
 
 import { InternalServerError } from './common/error/errors';
 import { ReadOnlyRequired } from './common/types/common-types';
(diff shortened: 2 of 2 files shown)


packages/dev-server/index-worker.ts
```diff
@@ -2,6 +2,10 @@ import { bootstrapWorker } from '@vendure/core';
 
 import { devConfig } from './dev-config';
 
+// https://github.com/vendure-ecommerce/vendure/issues/152
+// fix race condition when modifying DB
+devConfig.dbConnectionOptions = { ...devConfig.dbConnectionOptions, synchronize: false };
```
michaelbromley, line 7: This is the right idea, but putting it here only solves the issue for those people working on the Vendure repo itself. If you were to install the @vendure/core npm package, you would still run into the issue.

I suggest that the fix be put here: https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/bootstrap.ts#L100. That way it fixes the issue for everyone, no matter if they are developing on the Vendure repo or using the npm package.

PR #168, opened 2019-09-25
feat(email-plugin): Added `from` field to the email config

lsimone:
This introduces the `from` field to the email plugin config.
It is processed by handlebars as it is `subject`


packages/email-plugin/src/event-listener.ts
```diff
@@ -117,6 +117,7 @@ export class EmailEventHandler<T extends string = string, Event extends EventWit
     private filterFns: Array<(event: Event) => boolean> = [];
     private configurations: EmailTemplateConfig[] = [];
     private defaultSubject: string;
+    private from: string;
```
michaelbromley, line 120: If the `setFrom()` method is not called on a handler, then "from" will be `undefined`. This is currently causing all the tests to fail because Handlebars doesn't like `undefined`. Initializing `from` to empty string fixes that.

PR #241, opened 2020-01-08
docs: Fix 'Adding new ActionBar buttons' example code

ironynet:
docs: Add 'requirePermission' property in 'Adding new ActionBar buttons' example code, because invisible button.


michaelbromley:
Thanks for the contribution! It just so happens this issue came up in the past week and the last commit (https://github.com/vendure-ecommerce/vendure/commit/292e6d471877422012265fe25e8cdf943cb54cb0) adds a default permission to prevent the "invisible button" problem. But this addition to the docs is welcome anyway, since it demonstrates the existence of this option.

PR #249, opened 2020-01-23
should check if it is a valid or final state

Szbuli:
State machine should return false if trying to change state when in a final state.


michaelbromley:
Hi, thanks for the PR.

I'm not sure I understand the case which this change handles. Could you clarify? Perhaps adding a unit test to illustrate if appropriate?

Szbuli:
Hi,

I thought  for the default states you did not set the "to" field for final states as Cancelled. 

But I see now that you did. If we set an empty "to" field for final states, then the pull request is not relevant anymore.

michaelbromley:
Ok I understand. Yes, the `to` property is required, otherwise there is a TS error, and a terminal state would indeed just have an empty array. I'll close this then, but thanks anyway for taking the time to contribute 👍

PR #281, opened 2020-03-24
fix(docs): update import package for Transport

JWimsingues:


JWimsingues:
Same as here: https://www.vendure.io/docs/typescript-api/worker/worker-options/#workeroptions

michaelbromley:
Good catch, thanks for the fix! :)

PR #284, opened 2020-03-28
bug(admin-ui): Asset drag/drop support in safari

xxKRASHxx:
**STR:**

1.  custom-built admin-ui, `ts-node compile-admin-ui.ts`:
```
import { compileUiExtensions } from '@vendure/ui-devkit/compiler';
import path from 'path'

compileUiExtensions({
    outputPath: path.join(__dirname, 'dist/admin-ui'),
    extensions: [{
        translations: {
            ru: path.join(__dirname, 'translations/ru.json'),
        }
    }],
}).compile!()
```
2. connect admin-ui plugin in `vendure-config.ts`:
```
AdminUiPlugin.init({
  app: {
    path: path.join(__dirname, 'admin-ui/dist')
  },
  ...
});
```

3. open admin web page from Safari browser 13.0.5

**ART:** 
-  blank page
- https://stackoverflow.com/questions/56137343/angular-not-working-on-safari-browser-but-working-well-on-other-browsers

**ERT:**
- page opens as usual

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/asset-file-input/asset-file-input.component.ts
+++ b/packages/admin-ui/src/lib/core/src/shared/components/asset-file-input/asset-file-input.component.ts
@@ -45,7 +45,7 @@ export class AssetFileInputComponent implements OnInit {
     }
 
     @HostListener('document:dragleave', ['$event'])
-    onDragLeave(event: DragEvent) {
+    onDragLeave(event: any) {
         if (!event.clientX && !event.clientY) {
             this.dragging = false;
         }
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi, thanks for the contribution. I don't understand something: the change in your PR is just a TypeScript type annotation. These are removed during compilation and have no effect on the runtime behaviour of the app.

Could you please post any errors in the console of Safari when you see the blank page?

xxKRASHxx:
Screenshot 2020-03-31 at 19 36 23

xxKRASHxx:
> the change in your PR is just a TypeScript type annotation.
I understand, just followed stackoverflow suggestion...
basically, I not so deep in web & angular so bit confused on how to debug such issues.

xxKRASHxx:
@michaelbromley, I think, we need a polyfill for DragDrop events, not the `any` 😑 ...

edxds:
> Hi, thanks for the contribution. I don't understand something: the change in your PR is just a TypeScript type annotation. These are removed during compilation and have no effect on the runtime behaviour of the app.

Angular actually includes DragEvent as a variable in some type decorators that are actually included in the app bundle. This causes the "can't find variable DragEvent" error. Changing DragEvent to Event is enough to fix the error and doesn't involve using 'any'.

michaelbromley:
@edxds Can you confirm that the change in this PR actually resolves the issue on Safari? I don't have access to Safari locally.

edxds:
This PR seems to fix the issue in Safari.

PR #301, opened 2020-04-08
fix(core): add missing translations for default channel not found

jrybinski:
Hi 👋 
There were missing translations for error about missing default channel from channel service.

Here is related function:

https://github.com/vendure-ecommerce/vendure/blob/987b611fa3a7afc5f70183a773299d9d88f4da0c/packages/core/src/service/services/channel.service.ts#L118

Translations for ES are from google translate but I think it's better than nothing. 

I also think we should have the same amount of messages for both languages, even if not everything is translated because, again, any kind of message is better than nothing ;)
 Now in es.json, there are much fewer translations.

Cheers 🥂

### DIFF
--- a/packages/core/src/i18n/messages/en.json
+++ b/packages/core/src/i18n/messages/en.json
@@ -24,6 +24,7 @@
     "create-fulfillment-items-already-fulfilled": "One or more OrderItems have already been fulfilled",
     "create-fulfillment-orders-must-be-settled": "One or more OrderItems belong to an Order which is in an invalid state",
     "create-fulfillment-nothing-to-fulfill": "Nothing to fulfill",
+    "default-channel-not-found":  "Default channel not found",
     "email-address-not-available": "This email address is not available",
     "email-address-not-verified": "Please verify this email address before logging in",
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hi, thanks for taking the time to PR this. 

> I also think we should have the same amount of messages for both languages, even if not everything is translated because, again, any kind of message is better than nothing ;)
Now in es.json, there are much fewer translations.
The way it is supposed to work is that, if there is no Spanish translation for a particular key, it will fall back to the English version.

I've got a feeling the Spanish one should use "por defecto" for "default", based on the [spanish UI translations](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/static/i18n-messages/es.json) which use "por defacto".

jrybinski:
Ok, so should I change Spanish translations? Or remove it and wait for someone who knows this language better and for now we will have fall back version in this case?

michaelbromley:
I'd say just remove it for now and wait for a spanish speaker to notice and get involved :)

jrybinski:
@michaelbromley done

PR #302, opened 2020-04-13
fix(core) Fix swiss and congolese currency codes

Tyratox:
I'm currently experimenting with vendure and so far like it quite a lot.
I currently think of migrating an existing e-commerce shop and am now checking whether I can implement all required features. While setting up a demo system I noticed that the swiss currency code is set to `CHW` instead of `CHF`. This is probably because ISO 4217 defines three currency codes for switzerland but `CHF` is the one usually used.

Screen Shot 2020-04-13 at 15 42 00

Also I noticed that the name used in the channel settings ("WIR euros") can't be found inside the repository so I'm not sure where to do this change/update.

### DIFF
--- a/packages/core/src/api/schema/common/currency-code.graphql
+++ b/packages/core/src/api/schema/common/currency-code.graphql
@@ -58,7 +58,7 @@ enum CurrencyCode {
     "Canadian dollar"
     CAD
     "Congolese franc"
-    CHE
+    CDF
     "Swiss franc"
     CHF
     "Chilean peso"
--- a/packages/core/src/api/schema/common/currency-code.graphql
+++ b/packages/core/src/api/schema/common/currency-code.graphql
@@ -60,7 +60,7 @@ enum CurrencyCode {
     "Congolese franc"
     CHE
     "Swiss franc"
-    CHW
+    CHF
     "Chilean peso"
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi, thanks for the PR! 

> Also I noticed that the name used in the channel settings ("WIR euros") can't be found inside the repository so I'm not sure where to do this change/update.

The human-readable string for each code is handled by the [CurrencyNamePipe](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/shared/pipes/currency-name.pipe.ts) which internally uses the browser's Intl API. So in theory this should always be correct. 

The "CHE" currently listed as "Congolese franc" is then wrong - Congolese franc is "CDF". Would you be able to change that code in the same file too?

Tyratox:
> The human-readable string for each code is handled by the CurrencyNamePipe which internally uses the browser's Intl API. So in theory this should always be correct.

Ah perfect, no change is needed then.

> The "CHE" currently listed as "Congolese franc" is then wrong - Congolese franc is "CDF". Would you be able to change that code in the same file too?

Of course!

michaelbromley:
Thank you for your help 👍

PR #317, opened 2020-04-17
feat(admin-ui): Add PL language with partial translations

jrybinski:
PL language added with partial translations. I will continue working on, just opened PR to share my current status with other dev that want to help with it.


michaelbromley:
I'm afraid you'll need to rebase onto the latest master in order to get these CI tests to pass. If you are not comfortable rebasing, then merging master into your branch is also ok.

jrybinski:
@michaelbromley rebased, should be fine now ;)

PR #318, opened 2020-04-18
fix(ui-devkit) Fix generation of shared module file

Tyratox:
**Describe the bug**
When generating the shared extension module the different extension modules are joined together correctly but if one extension contains multiple modules the default `join` inserts commas which results in a compilation error.

**To Reproduce**
Export multiple ngModules from one extension.

EDIT: It's the default behavior in JS: https://jsfiddle.net/kyp8L5bx/

**Expected behavior**
The imports should be joined together with a '' instead of a comma

### DIFF
--- a/packages/ui-devkit/src/compiler/scaffold.ts
+++ b/packages/ui-devkit/src/compiler/scaffold.ts
@@ -93,7 +93,8 @@ ${extensions
     .map(e =>
         e.ngModules
             .filter(m => m.type === 'shared')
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi, thanks for the PR!

As you might have noticed, the CI failed for no fault of your own. I just pushed a couple of fixes. Would you be so kind as to rebase onto master push again to this PR?

Tyratox:
Of course!

PR #322, opened 2020-04-20
chore: Throw error when no active tax zone in channel

samundra:
- Throw error when no active tax zone is not found in channel


michaelbromley:
Thanks for the help!

PR #329, opened 2020-05-03
feat(core): Added playground and debug config for graphql apis

kai-chu:


packages/core/src/config/vendure-config.ts
```diff
@@ -427,6 +427,39 @@ export interface VendureConfig {
      * @default 'shop-api'
      */
     shopApiPath?: string;
+    /**
+     * @description
+     * The playground config to the admin GraphQL API
+     * [`ApolloServer playground`](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#constructor-options-lt-ApolloServer-gt)
```
michaelbromley: This link appears to be malformed somehow - the `<` & `>` chars are encoded as lt, gt

PR #336, opened 2020-05-09
feat(core): Added playground and debug config for graphql apis

kai-chu:
BREAKING CHANGE: The graphql-playground for the Shop and Admin APIs are now turned off by default, and the Apollo server debug option is also set to false by default (it was formerly true). You can manually configure these values using the VendureConfig.apiOptions object.


michaelbromley:
Excellent work, thanks for your contribution 🙇 👍

Issue #328, opened 2020-05-03
validateVariantOptionIds is memory inefficient and can easily cause the server to crash


Tyratox:
**Describe the bug**
The function
https://github.com/vendure-ecommerce/vendure/blob/451ae50d2316e8f3221c9df23504aac9661337fc/packages/core/src/service/services/product-variant.service.ts#L364-L367
is memory inefficient. 
The function
https://github.com/vendure-ecommerce/vendure/blob/aacfaf4b5911dc10a59d26ad2a17dc5d86e450dc/packages/core/src/service/helpers/utils/get-entity-or-throw.ts#L28-L45
is called which itself calls `findOneInChannel` on line 37

https://github.com/vendure-ecommerce/vendure/blob/aacfaf4b5911dc10a59d26ad2a17dc5d86e450dc/packages/core/src/service/helpers/utils/channel-aware-orm-utils.ts#L32-L48

All functions do work in theory but the generated sql query looks like the following:

[...]

PR #337, opened 2020-05-10
perf(core): Improved performance of validateVariantOptionIds

Tyratox:
This pull request tries to implement the idea mentioned in https://github.com/vendure-ecommerce/vendure/issues/328#issuecomment-623290422.

Step one and two are done at the same time as before as one is required to do the join anyway.
I wasn't able to think of a case where performing the two steps separately would require less memory but I might be missing something.

In addition to the ideas mention I added the eager option to getEntityOrThrow to further decrease the memory required. Is this okay or should this parameter be implemented differently / i.e. a separate function be added with this functionality?


michaelbromley:
Looks good - the `eager` option is fine for `getEntityOrThrow`. Might be useful in optimizing other areas of perf in future, since that function is used a lot all over the service layer. Thanks for your time 👍

PR #352, opened 2020-05-29
feat(docs): Fix example code for admin ui build as deployment step

jrybinski:
Fix for docs and example code for build ui extensions for admin UI as deployment step.


michaelbromley:
Thanks for the fix!

PR #354, opened 2020-05-29
feat(admin-ui) Add Traditional Chinese for i18n-Message

samxchan1021:
Add one more type of Chinese (Traditional) which are for Hong Kong and Taiwan.


michaelbromley:
Thank you very much for the contribution. For now, can you update the PR to remove the renaming of zh.json and the change to vendure-ui-config.json?

The reason is that some prior work needs to be done to properly support region-specific languages (see #356), and once that is done I can expose your zh-TW translations in the admin ui.

By the way, can you provide translations for 

* "Traditional Chinese"
* "Simplified Chinese"

in both languages (just in a comment reply to this is fine)? Then I can update the zh-CN version later with a correct labels once proper support is enabled for regional variations.

samxchan1021:
> Thank you very much for the contribution. For now, can you update the PR to remove the renaming of zh.json and the change to vendure-ui-config.json?
> 
> The reason is that some prior work needs to be done to properly support region-specific languages (see #356), and once that is done I can expose your zh-TW translations in the admin ui.
> 
> By the way, can you provide translations for
> 
> * "Traditional Chinese"
> * "Simplified Chinese"
> 
> in both languages (just in a comment reply to this is fine)? Then I can update the zh-CN version later with a correct labels once proper support is enabled for regional variations.

Of course , give me a second.

michaelbromley:
Thank you very much :)

Issue #279, opened 2020-03-20
Create superadmin from env variables if they exist 


dantium:
**Is your feature request related to a problem? Please describe.**
Superadmin details such as the password are fixed during initial creation - for security and ease of use during deployment it might be good to have a way to set them to something unique.

**Describe the solution you'd like**
It would be nice to have the option for the superadmin user to automatically be created with details from env variables if they exist and if not just use defaults.

For clarity the env variables could be prepended with VENDURE_ i.e: `VENDURE_SUPER_ADMIN_USER_PASSWORD`

PR #358, opened 2020-06-01
feat(core,testing,create): Enable users to specify superadmin credentials

edxds:
This pull requests modifies the core, testing, and create packages to allow the user to specify superadmin credentials during and after app creation.

The core package was modified to add superadmin credentials fields to the config object and to allow reading of the credentials from config in the administrator service, while in the create package new prompts asking for credentials while providing default values were added.

The testing package was modified to allow the `SimpleGraphQLClient` class to access credentials from the `vendureConfig` argument.

This pull request closes issue #279 .

### DIFF
--- a/packages/create/src/gather-user-responses.ts
+++ b/packages/create/src/gather-user-responses.ts
@@ -93,7 +93,7 @@ export async function gatherUserResponses(root: string): Promise<UserResponses> 
             {
                 type: 'text',
                 name: 'superadminPassword',
-                message: 'WWhat password do you want to use for the superadmin user?',
+                message: 'What password do you want to use for the superadmin user?',
                 initial: SUPER_ADMIN_USER_PASSWORD,
             },
         ],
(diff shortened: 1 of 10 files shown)


packages/create/src/gather-user-responses.ts
```diff
@@ -80,6 +84,18 @@ export async function gatherUserResponses(root: string): Promise<UserResponses>
...
+            {
+                type: 'text',
+                name: 'superadminIdentifier',
+                message: 'What identifier do you want to use for the superadmin user?',
+                initial: SUPER_ADMIN_USER_IDENTIFIER,
+            },
+            {
+                type: 'text',
+                name: 'superadminPassword',
+                message: 'WWhat password do you want to use for the superadmin user?',
```
michaelbromley: Typo "WWhat"

edxds: Thanks, good catch! I'll update the pull request.

michaelbromley:
Very nice work, thanks again!

PR #365, opened 2020-06-07
fix(core): Fix findByIdsInChannel to take ids into account

Tyratox:
Apparently findByIdsInChannel didn't actually filter for the ids passed to it.


michaelbromley:
Whaaaaat??? I musn't have had my coffee that day to miss this!

Issue #402, opened 2020-07-06
Custom fields for shippingMethod entity


hendrikdepauw:
**Is your feature request related to a problem? Please describe.**
Our project requires custom fields to be defined on the ShippingMethod entity.

**Describe the solution you'd like**
Make ShippingMethod extend HasCustomFields.

**Describe alternatives you've considered**
None.

**Additional context**
We have/are creating a PR with the changes.

PR #406, opened 2020-07-07
feat(core): Enable custom fields on ShippingMethod entity

hendrikdepauw:
Related to issue #402.


hendrikdepauw:
Hi Michael, could you shed some light on these failing tests? I feel like this has nothing to do with the proposed changes. On top, nothing failed in the other PR. Thanks in advance.

michaelbromley:
Yeah you're right, there's quite some major refactoring in this branch and it looks like I missed updating some tests. Don't worry about it.

Issue #398, opened 2020-07-04
Accept empty as response in ShippingCalculator


jonyw4:
I have some `ShippingCalculator` that requests another server to get the price of the shipping and not always this ShippingCalculator returns a price, because this carrier doesn't ship everywhere. Currently, we can only throw an error, but in this scenario, it's not an error, its just an empty response. Would be nice if the ShippingCalculator accepts an empty response.

jonyw4:
I can do a PR if you approve the idea @michaelbromley 👌

michaelbromley:
Well, this is exactly why we also have the ShippingEligibilityChecker - we have separated the 2 concerns of 

1. Is this carrier valid for this order?
2. If so, what's the price?

The calculator should only get executed if the eligibility checker returns `true`, so your scenario should indeed be an exception.

jonyw4:
Okay, that makes sense in many cases, but in the scenario that I create a ShippingCalculator that uses an external web service/provider and the response it's unknown, how can I resolve? Because I don't know if the shipping address of the customer it's valid for this carrier if I don't ask for the shipping calculator, you got it?

michaelbromley:
The idea is that in your custom ShippingEligibilityChecker, you make the call to the carrier's API, and if you get an empty response, you resolve it with `false`, because that carrier cannot ship this order. If you get a price, then you return `true`. 

I know this means you'll need 2 basically identical calls to their API for the checker + calculator step, but if this is a problem you could e.g. implement caching locally so you do not need to make the second call for the price if you already just asked the API in the checker step.

jonyw4:
I understand but this doesn't feels right...
In this scenario i need to add a custom ShippingCalculator and a ShippingEligibilityChecker to just to configure one shipping calculation / carrier, that can complicate the implementation.
I think that adding a null or empty response for ShippingCalculator its so much easier solution both for developers and to the admin user that will be using the Admin UI.
In addition, if we can return null from ShippingCalculator i can create a custom ShippingEligibilityChecker that can be used for many ShippingCalculator.

michaelbromley:
OK, I get it. So do you propose to change the return type of the `ShippingCalculator.calculate()` method to `ShippingCalculator | undefined`?

If so, I guess the implementation to support it would be as simple as checking for undefined [here](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/entity/shipping-method/shipping-method.entity.ts#L57).

jonyw4:
[...]

PR #407, opened 2020-07-07
Shipping Calculator Changes

### LINKED ISSUE #397: Await all promises in the same time in method getEligibleShippingMethods in ShippingCalculator
I have some plugins using ShippingCalculator that are promises. Would be nice to all promises were loaded at the same time. Currently, the method `getEligibleShippingMethods` awaits the response from the `shippingMethods`  individually.

https://github.com/vendure-ecommerce/vendure/blob/b2f3f082ade6e95374b997651ac517cb2c1a1890/packages/core/src/service/helpers/shipping-calculator/shipping-calculator.ts#L17-L32

A solution would be using [Promises.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all).

jonyw4:
Closes #397 and #398


packages/core/src/service/helpers/shipping-calculator/shipping-calculator.ts
```diff
@@ -14,21 +23,32 @@ export class ShippingCalculator {
...
-                    eligibleMethods.push({ method, result });
-                }
+
+        const checkEligibilityPromises = shippingMethods.map((method) =>
+            this.checkEligibilityByShippingMethod(order, method),
+        );
+        const eligibleMethods = await Promise.all(checkEligibilityPromises);
+
+        // Workaround to remove undefined from type: https://stackoverflow.com/a/51577579
+        const filteredEligibleMethods = eligibleMethods.filter(
```
michaelbromley: You can use the `notNullOrUndefined` util function here (https://github.com/vendure-ecommerce/vendure/blob/master/packages/common/src/shared-utils.ts#L5)

You can import it from "@vendure/common/lib/shared-utils"

packages/core/src/service/helpers/shipping-calculator/shipping-calculator.ts
```diff
@@ -6,6 +6,15 @@ import { Order } from '../../../entity/order/order.entity';
 import { ShippingMethod } from '../../../entity/shipping-method/shipping-method.entity';
 import { ShippingMethodService } from '../../services/shipping-method.service';
 
+type CheckEligibilityByShippingMethodResponse =
```
michaelbromley: I would advise to just define a single type `CheckEligibilityByShippingMethodResponse ` without the `undefined` union, and then in define like `checkEligibilityByShippingMethod(): Promise<CheckEligibilityByShippingMethodResponse | undefined>`. This requires one less type definition and IMO is simpler to understand.

packages/core/src/service/helpers/shipping-calculator/shipping-calculator.ts
jonyw4: Great! When I wrote this part using filter the type definitions didn't get right. Good to know that have a function to deal with this 😁

packages/core/src/service/helpers/shipping-calculator/shipping-calculator.ts
jonyw4: Nice! Good point. I will update

jonyw4:
> See a couple of comments. Also, could you rebase onto the latest `next` branch because I fixed the causes of the failing tests, so then we know whether these changes cause any tests to fail.

Done! 😬

PR #418, opened 2020-07-23
refactor: Update shipping calculator typing

jonyw4:
I realize that when I fix the issue #398 with the PR #407 I forget to add `undefined` in `calculate` function of the shipping calculator. So to fix that I create a `CalculateShippingFnResponse` type to use in all responses in this file.


packages/core/src/config/shipping-method/shipping-calculator.ts
```diff
@@ -90,6 +87,11 @@ export interface ShippingCalculationResult {
     metadata?: Record<string, any>;
 }
 
+export type CalculateShippingFnResponse =
```
michaelbromley: I think a better name would be `CalculateShippingFnResult` - "response" is more associated with a network call.

PR #429, opened 2020-08-05
feat: Remove all order lines mutation in shop gql

jonyw4:


samundra:
Thank you for this PR. I think you will also need to add test cases. @michaelbromley has many test cases in core APIs. Maybe borrowing some code existing code for test could work. I am not sure if there is any doc on *Test* topics or not.

jonyw4:
Alright, I can do that. I will check some test files in Vendure and try to update the PR. Would be nice If you or @michaelbromley have any tips where and what you want to cover in the test so that I can create more accurate test files

michaelbromley:
Nice work @jonyw4! The changes look good (also kudos for updating the docs!), but as @samundra said it would be good to have this tested too.

At the very least, a test to check that the `removeAllOrderLines` mutation does what is says, i.e.

1. addItemToOrder several times with different ProductIds to create an Order with several lines
2. call `removeAllOrderLines`
3. assert that the `order.lines.length === 0` and also that the totals are 0.

This test would live here: https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/e2e/shop-order.e2e-spec.ts

Be careful because in the e2e tests, later tests often depend on the state arrived at from previous tests. So e.g. if you were to test `removeAllOrderLines` in the middle of the suite, it is likely that later tests (which assume certain items in the order) will fail. So it may be best to add this test right at the end on the suite.

jonyw4:
Thanks for the explanation @michaelbromley! I will send the tests.

jonyw4:
Done! 🥳

michaelbromley:
Nice work, well done! 👏

PR #431, opened 2020-08-06
feat: Extended S3Config to accept aws-sdk configuration properties

thomas-advantitge:
Allows a more generic use of the S3 standard. 

As an example, it enables the usage of [MinIO](https://min.io/), an open source object storage which can be useful in several ways, e.g. for local development using storage buckets or as a part of a Kubernetes cluster.
It can also be used as a gateway to other object storages like Google Cloud Storage or Azure. This way, Vendure supports several new storage engines without the need for custom asset storage strategies.

Example usage:
```typescript
storageStrategyFactory: configureS3AssetStorage({
    bucket: process.env.STORAGE_BUCKET || 'default',
    credentials: {
      accessKeyId: process.env.STORAGE_ACCESSKEY || 'default',
      secretAccessKey: process.env.STORAGE_SECRETKEY || 'default',
    },
    nativeS3Configuration: {
      endpoint: process.env.STORAGE_ENDPOINT || 'http://localhost:9000',
      s3ForcePathStyle: true,
      signatureVersion: 'v4',
    },
  }),
```

### DIFF
--- a/packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
+++ b/packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
@@ -1,4 +1,4 @@
-import { AssetStorageStrategy, Injector, Logger } from '@vendure/core';
+import { AssetStorageStrategy, Logger } from '@vendure/core';
 import { Request } from 'express';
 import { Readable, Stream } from 'stream';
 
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks for this. I'm not experience with S3 so I'm glad someone with a bit more knowledge is able to improve the API of the strategy. I'd eventually like to replace the `any` type with the correct typing, but right now it would not be possible without causing TS errors for those not using `aws-sdk`. In future perhaps we can export the S3 stuff in a sub-package so that we can include correct typings without affecting non-s3 users.

I'm pretty sure the CI failures are just random flakiness, so I'm gonna merge anyway.

PR #434, opened 2020-08-08
feat: Custom field length configuration for localeString

thomas-advantitge:


michaelbromley:
Thanks!

Note to self: Add an e2e test for this & update docs.

Issue #440, opened 2020-08-20
Making Orders ChannelAware


hendrikdepauw:
As we would like to use Vendure in a multi-tenant environment, we want to make the Order entity ChannelAware.
We are currently in the process of implementing this feature.

Any advice/pointers are welcome.

hendrikdepauw:
Currently creating a PR so we can get some feedback.

PR #441, opened 2020-08-20
Making Orders ChannelAware #440

hendrikdepauw:
Related to issue #440 .
Still have an issue where the order is not saved under the default channel. I am however using the ChannelService.AssignToCurrentChannel() method.


michaelbromley:
Thank you for the fantastic work on this feature. Especially the thorough e2e test suite. I found the cause of the failing e2e test, and it was the fault of the test, not your changes.

PR #454, opened 2020-09-03
feat(core): Make Customers ChannelAware

hendrikdepauw:


hendrikdepauw:
@michaelbromley I'm not entirely sure what is up with the tests?

hendrikdepauw:
If a customer has a user account and uses the 'login' mutation on a channel that is not yet assigned to his customer entity, the channel will not yet be assigned to it. This will only happen on the next query/mutation. 
This is a side effect of assigning customers to channels before the actual request is handled: before actually logging in there is no customer yet to assign to a channel.

packages/core/e2e/customer-channel.e2e-spec.ts
```diff
@@ -0,0 +1,487 @@
...
+                {
+                    groupId: customerGroupId,
+                    customerIds: [thirdCustomer.id],
+                },
+            );
+        });
+    });
+});
+
+export const CREATE_ADDRESS = gql`
```
michaelbromley, line 342: These gql operations seem to be identical to the ones in the customer.e2e-spec.ts file.
In that case, my suggestion is to move them into /graphql/shared-definition.ts and import from there into both spec files to avoud duplication.

packages/core/src/config/promotion/promotion-condition.ts
```diff
@@ -18,6 +19,7 @@ import { Order } from '../../entity/order/order.entity';
  * @docsPage promotion-condition
  */
 export type CheckPromotionConditionFn<T extends ConfigArgs> = (
+    ctx: RequestContext,
```
michaelbromley, line 22: This part with adding the RequestContext to the check function seems to be mixing concerns. Is there some way that this is connected with making Customer channel-aware? If not, I'd suggest you leave them out of this PR, and create a separate issue/PR with some reasoning for it (I think I can guess, but it's good to make such decisions explicit and formal).

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -35,9 +40,31 @@ export class AuthGuard implements CanActivate {
...
+                );
+                // To avoid assigning the customer to the active channel on every request,
+                // it is only done on the first request and whenever the channel changes
+                if (customer) {
+                    await this.channelService.assignToChannels(Customer, customer.id, [
+                        requestContext.channelId,
+                    ]);
+                }
+            }
+            requestContext = await this.requestContextService.fromRequest(req, info, permissions, session);
```
michaelbromley: Ah much better than the recursion 👍.
I would suggest extracting the first part of this `if` block (ln 49-63) into a well-named method to make it easier to follow the high-level logic of the outer `canActivate` method. E.g.:

```ts
private setActiveChannel(session: CachedSession, requestContext: RequestContext): CachedSession {
  // ...
  }
```

and also think about extracting the `if` expression into a named constant like `activeChannelShouldBeSet` for better readability.

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -35,9 +40,31 @@ export class AuthGuard implements CanActivate {
         const isPublic = !!permissions && permissions.includes(Permission.Public);
         const hasOwnerPermission = !!permissions && permissions.includes(Permission.Owner);
         const session = await this.getSession(req, res, hasOwnerPermission);
-        const requestContext = await this.requestContextService.fromRequest(req, info, permissions, session);
+        let requestContext = await this.requestContextService.fromRequest(req, info, permissions, session);
         (req as any)[REQUEST_CONTEXT_KEY] = requestContext;
```
michaelbromley, line 50: I think this line can be moved down below the next `if` block so that it does not need to be repeated inside the `if` block.

michaelbromley (changes requested):
Great work! Please see the review comments 👍

packages/core/src/config/promotion/promotion-condition.ts
hendrik-advantitge: This is added because the default `CustomerGroupCondition` references `customerService.getCustomerGroups()`. Previously this method did not require a context, but with ChannelAware customers it now does. Therefor a RequestContext needs to be passed to the check method.

packages/core/e2e/customer-channel.e2e-spec.ts
hendrik-advantitge: That is how it was originally coded, but there seems to be a bug (feature?) happening here where if you import from another file that also contains tests, Jest starts executing those tests as well. This creates interference issues and the tests actually fail. A separate util file could be created that contains these definitions so they can be shared?

hendrik-advantitge: I suggest creating an `admin-definitions.ts` file in `core/e2e/graphql` so that the queries are not spread across the tests. The `E2E_ADMIN_QUERY_FILES` constant in `generate-graphql-types.ts` can then be simplified as well. This deserves its own PR in my opinion.

michaelbromley: The existing `shared-definitions.ts` file already does this, but only with definitions used in more than 1 spec file. I'm not sure I agree with putting absolutely _all_ definitions into a single file because of the resulting size of the file and potential increased difficulty finding what you are looking for. But you can PR it and then we see how it looks, if you like.

packages/core/src/config/promotion/promotion-condition.ts
michaelbromley: OK I get it, thanks for clarifying.

michaelbromley:
Thank you so much for the work you put in on this! 🎉

loveyandex:
@michaelbromley how can we define a vendor for a channel?

PR #498, opened 2020-10-14
fix(docs): update findCustomerUser call

macabeus:
On Vendure 0.16.0, now `findCustomerUser` requires `ctx` parameter


michaelbromley:
Thanks!

PR #502, opened 2020-10-14
Fixes in migration guide

chladog:
The migration to channel aware customers query was not correct. Fixed.
Added mandatory ctx object to new style "this.connection.getEntityOrThrow()".

Additionally after generating migration script I had to update the queries to update gaps in old data. Here with these bold values:
**await queryRunner.query(`UPDATE "order" SET "state"='Delivered' WHERE "state"='Fulfilled'`, undefined);**
await queryRunner.query(`ALTER TABLE "fulfillment" ADD "state" character varying NOT NULL **DEFAULT 'Delivered'**`, undefined);

_So maybe you want to include those updates in guide as well._


michaelbromley:
Thank you! I will also add the fulfillment state migration lines too.

PR #511, opened 2020-10-19
AdminUI - Custom order states filtering + highlight

chladog:
-New feature of filtering of orders based on **custom** order states.
-New admin-api query "orderStates" that returns an array of **all configured states** from OrderStateMachine.
-Change of default order state styles to highlight color. Custom order states are usually part of order processing and therefore of higher importance than AddingItems and ArraningPayment, we do not want to be losing them visually.


packages/core/src/api/schema/admin-api/order-states.api.graphql
```diff
@@ -0,0 +1,3 @@
+type Query {
+    orderStates: [OrderState!]
```
michaelbromley, line 2: We already have this info available in the Admin API via the `globalSettings` query, which can return:

```ts
{
  serverConfig: {
    orderProcess: [ ... states ]
  }
}
```

So you can remove this and use that existing query in your Admin UI changes.

michaelbromley (changes requested):
Thanks for the contribution! See my comment about re-using an existing query for the Order states.

packages/core/src/api/schema/admin-api/order-states.api.graphql
chladog: I found it easier to make new PR: https://github.com/vendure-ecommerce/vendure/pull/514
closing this one

PR #514, opened 2020-10-19
AdminUI - Custom order states filtering + highlight

chladog:
-New feature of filtering of orders based on custom order states.
-Change of default order state styles to highlight color. Custom order states are usually part of order processing and therefore of higher importance than AddingItems and ArraningPayment, we do not want to be losing them visually.


michaelbromley:
Thank you for the contribution! 👍

PR #526, opened 2020-10-22
Fix overflow that made website unusable on sm devices

chladog:
For small devices in conjunction with long username (which is usual case as we use emails as usernames) the header causes big overflow that renders ui almost unusable.
vendure_mobile_overflow

As fix in this PR I hid username altogether as it's not crucial information.
vendure_mobile

_If the display of currently logged-in admin is wanted, I would include it in main menu for sm devices._


michaelbromley:
Thanks!

PR #542, opened 2020-11-02
I18n state tokens

chladog:
This PR adds missing translations of Fulfillment, Refund and Payment states and in "Transition to _xxx_" strings and makes custom states translatable.

- change OrderStateI18nPipe to general StateI18nPipe and update translation files accordingly
- use StateI18nPipe in: 
1. Fulfillment states
2. transition-to-state's arguments
3. Payment states
4. Refund states
5. Fulfillment "Mark as _xxx_" strings


michaelbromley:
Thanks!

PR #549, opened 2020-11-06
Fix payment states

chladog:
Add missing payment states tokens


michaelbromley:
Thanks for correcting my Czech grammar 😆

PR #560, opened 2020-11-18
Refine new order list filters

chladog:
- move "Cancelled" from _Open_ to _Closed_.
- add new "Shipped" - reasoning behind is that "Shipped" orders are not of interest to shop managers-by this time they done their part with fulfilling the order. Keeping "Partially Shipped" and "Partially Fulfilled" in _Open_, as there is still some actions to be done. Even if need of manual marking of "delivered", they are still not as important as orders that are awaiting/in the processing.

Improvement here would also be that these presets would be configurable by devs to match an use-case exactly.


michaelbromley:
Thanks, this is a good workflow improvement. 

Re. configurable presets - yes, good idea too. Feel free to open an issue. But not mu priority for the time being.

Issue #309, opened 2020-04-15
Own Logo


radidas:
I would like to add the possibility to change the logo in the admin panel to your own. Thank you .

michaelbromley:
With the referenced fix above, this is now possible using the existing ui extensions system:

```ts
plugins: [
  AdminUiPlugin.init({
    port: 3002,
    app: compileUiExtensions({
      outputPath: path.join(__dirname, 'admin-ui'),
      extensions: [{
         {
           extensionPath: path.join(__dirname, 'my-images'),
           ngModules: [],
           staticAssets: [
             {
               path: path.join(__dirname, 'my-images/small-logo.png'),
               rename: 'cube-logo-75px.png',
             },
             {
               path: path.join(__dirname, 'my-images/large-logo.png'),
               rename: 'cube-logo-300px.png',
             },
           ],
         },
      }],
      devMode: true,
    }),
  }),
],
```

The above config basically just overwrites the logo image files with your own.

PR #610, opened 2021-01-03
Allow customize Admin UI branding options

### LINKED ISSUE #391: Admin UI title, brand color change, vendure info hiding feature
I want a function in the admin UI dashboard where I can modify Admin UI branding as per my own brand like adding a brand color instead of blue everywhere, hiding/modifying the vendure related texts and branding.

Also if possible, css customization features like of wordpress can also be good if possible.

jeancx:
Closes #391 

- [x] Modify interface AdminUiConfig on common/lib/shared-types to accept new configs.
- [x] Make admin-ui-plugin to accept new configs.
- [x] Modify admin-ui to render customizations on component html
- [x] Set title on app load.
- [x] Update docs

What this PR does change:

- Allow to add brand on title and on side of vendure brand.
- Allow to hide Vendure branding.
- Allow to hide Vendure version.

To test:
```
AdminUiPlugin.init({
    adminUiConfig: {
        brand: 'My Store',
        faviconPath: path.join(__dirname, 'favicon.ico'),
        hideVendureBranding: false,
        hideVersion: false,
    },
}),
```


michaelbromley:
Hi!

Wow thanks for all your work on this PR! I'm just catching up on a bunch of things after the New Year break, but I'll make time to review this PR soon.

docs/content/docs/plugins/extending-the-admin-ui/_index.md
```diff
@@ -89,3 +89,8 @@ plugins: [
   }),
 ],
 ```
+
+## Avaliable custom settings
+
+* [Adding UI Translations]({{< relref "adding-ui-translations" >}})
```
jeancx: Hi, i found out that theres is a translation guide folder without link anywhere in the docs, so i added here, but i don'k know if i should have done.

michaelbromley: We don't actually need a manual link to that because Hugo generates a menu item in the docs nav menu, and also a "contents" link in the parent page. See https://www.vendure.io/docs/plugins/extending-the-admin-ui/

jeancx: So i will remove the changes on this file

packages/admin-ui-plugin/src/plugin.ts
```diff
@@ -278,32 +295,56 @@ export class AdminUiPlugin implements OnVendureBootstrap, OnVendureClose {
...
         const maxRetries = 10;
         const retryDelay = 200;
         let attempts = 0;
 
         const pause = () => new Promise(resolve => setTimeout(resolve, retryDelay));
 
         while (attempts < maxRetries) {
-            try {
-                Logger.verbose(`Checking for config file: ${adminUiConfigPath}`, loggerCtx);
-                const configFileContent = await fs.readFile(adminUiConfigPath, 'utf-8');
```
jeancx: here i found out that was reading the file, but the content was not being used, so i replaced by fs.existsSync, because i think the objective of this function was try to find the file until was avaliable or pass max tries. And since i would need the same logic for waiting the favicon to be avaliable, i refactored the function to use in both operations to replace vendure config json and favicon.

michaelbromley:
Alright, I've spent a bit of time looking around the changes and here are my thoughts:

There are currently 2 ways to customize the Admin UI:

1. via the `AdminUiPlugin.init()` options object
2. via the `@vendure/ui-devkit` function `compileUiExtensions()`.

The first method is mainly intended for environment-dependent configuration (api urls etc), whereas the second method is aimed toward functional extensions of the UI via extension modules.

These changes fall somewhere in between - they don't extend the UI per se, but they are purely visual changes not related to the underlying configuration of the admin ui.

I'm tending to think that (at least some of) this functionality belongs in the `ui-devkit` package rather than `admin-ui-plugin`. The reasons for this are:

1. The ui devkit already allows you to replace the logos (in a rather hacky way, see https://github.com/vendure-ecommerce/vendure/issues/309#issuecomment-617816511). This existing functionality could be wrapped in a convenience function. We could add a similar mechanism for the favicon.
2. The `compileUiExtensions` function is executed pre-build. This unlocks more possibilities regarding custom styling. For example, allowing you to specify sass files directly, or even overriding sass variables to enable app-wide colour branding. I've also been thinking of making use of CSS variables for the colours etc to make custom styling even easier. 

I think your approach (of adding options to AdminUiPlugin) makes sense for the branding options though.

Here's what I propose:

1. Reduce the scope of this PR to just the branding options (`brand`, `hideVendureBranding`, `hideVersion`).
2. Do the logos & favicon parts separately as part of the ui-devkit package.
3. Further research and design work needed on how we best deal with custom styles. Might require some refactoring of the admin ui source, e.g. if we want to incorporate CSS variables.

jeancx:
Make sense move logo, icon and css to ui-devkit, i think that even the user could suply his own scaffold dir, enabling him to make advanced customizations on specifc modules.

jeancx:
I have reduced the scope of this PR to  branding options (brand, hideVendureBranding, hideVersion), as suggested.

michaelbromley:
@jeancx Thank you very much for your contribution! Especially the docs - that's really thorough and appreciated.

I was thinking of working on the other 2 aspects myself (branding images & styling) unless you have a particularly strong desire to tackle either?

jeancx:
I just tested today, all workin fine, was a good work, very clean code.

PR #619, opened 2021-01-05
fix(admin-ui): Shipping method validators fix

chladog:
Existing shipping methods can't be updated sometimes due invalid checker/calulcator validators taking effect. This is fix for that. Also some Angular deprecation fixes for _combineLatest_


michaelbromley:
Thanks!

PR #620, opened 2021-01-06
Feat order list shipping column

chladog:
Display of shipping method in the order list view is very useful information for shop managers to keep track of the orders.

I also wanted to bring feature of "custom view" of the columns to the list view (so user could toggle columns as needed), but the table implementation is lacking dynamic columns feature and there were some problems with ordering. As well as sorting would be so much work that I came to conclusion to that usage of ready-to-go table solution like _[ag-grid](https://www.ag-grid.com/)_ would be much better user and dev experience - what do you think?

_PR also contains commit that features option to keep dropdown opened when clicked on their items. (which was part of my discarded dynamic cols effort)_


michaelbromley:
Thanks! I actually happened to need that "don't close on click" functionality when I was implementing something today :D

michaelbromley:
And yeah, the table component is rather limited. I'm not sure I wanna bring in something like agGrid by default though. 

With v0.18.0 you can override the nav menu routes, so this allows you to implement your own order list view, using whatever fancy grid you like.

PR #628, opened 2021-01-08
fix(docs): Fix path in readme to make installation process more clear

oncode:
I stumbled across this when installing, because I got back to the root folder after the 5. step.


michaelbromley:
Thanks!

PR #646, opened 2021-01-12
Order widget i18n fix

chladog:
Make widget buttons translatable and adjust strings to better represent their data ranges. Additionally add feature of "yeasterday" range.


packages/admin-ui/src/lib/static/i18n-messages/cs.json
```diff
@@ -302,10 +302,14 @@
     "latest-orders": "Poslední objednávky",
     "orders-summary": "Souhrn objednávek",
     "remove-widget": "Odebrat widget",
+    "thisMonth": "Tento měsíc",
+    "thisWeek": "Tento týden",
+    "today": "Dnes",
     "total-order-value": "Celkem",
     "total-orders": "Celkem objednávek",
     "widget-resize": "Změnit velikost",
     "widget-width": "Šířka: {width}"
```
michaelbromley: invalid JSON right here 👀

michaelbromley (changes requested):

packages/admin-ui/src/lib/static/i18n-messages/cs.json
chladog: Whoops...

PR #657, opened 2021-01-15
fix(admin-ui): Channel value display

oneyed:
Hey,
I was using the channel-assignment-control.component when i noticed that the channel values were not correctly displayed.
You can check the bug when editing a role with channels.


michaelbromley:
Thanks for the fix! 👍

PR #663, opened 2021-01-21
Run job queues in the worker process

mcfedr:
These changes are aimed at making it easier to scale out a Vendure server.

Currently the job queues are started in the server process, which check for new jobs and runs them. All the built in jobs, then send messages to workers, and wait for the response

It will be more efficient to job the job queue itself in the worker process, and allow it to run independently of the server process. If I run two worker processes I'll get twice the throughput.

### DIFF
--- a/packages/core/e2e/fixtures/test-plugins/with-job-queue.ts
+++ b/packages/core/e2e/fixtures/test-plugins/with-job-queue.ts
@@ -1,6 +1,7 @@
 import { Controller, Get, OnModuleInit } from '@nestjs/common';
 import { JobQueue, JobQueueService, PluginCommonModule, VendurePlugin } from '@vendure/core';
 import { Subject } from 'rxjs';
+import { take } from 'rxjs/operators';
 
 @Controller('run-job')
 class TestController implements OnModuleInit {
@@ -11,22 +12,20 @@ class TestController implements OnModuleInit {
     onModuleInit(): any {
         this.queue = this.jobQueueService.createQueue({
             name: 'test',
-            concurrency: 1,
             process: job => {
-                PluginWithJobQueue.jobSubject.subscribe({
-                    next: () => {
+                return PluginWithJobQueue.jobSubject
+                    .pipe(take(1))
+                    .toPromise()
+                    .then(() => {
                         PluginWithJobQueue.jobHasDoneWork = true;
-                        job.complete();
-                    },
-                    error: err => job.fail(err),
-                });
+                    });
             },
         });
     }
 
     @Get()
-    runJob() {
-        this.queue.add({});
+    async runJob() {
+        await this.queue.add({});
         return true;
     }
 }
--- a/packages/core/e2e/job-queue.e2e-spec.ts
+++ b/packages/core/e2e/job-queue.e2e-spec.ts
@@ -24,6 +24,9 @@ describe('JobQueue', () => {
     const { server, adminClient } = createTestEnvironment(
         mergeConfig(testConfig, {
             plugins: [DefaultJobQueuePlugin, PluginWithJobQueue],
+            workerOptions: {
+                runInMainProcess: true,
+            },
         }),
     );
 
--- a/packages/core/src/api/config/generate-permissions.ts
+++ b/packages/core/src/api/config/generate-permissions.ts
@@ -2,8 +2,8 @@ import { stitchSchemas } from '@graphql-tools/stitch';
 import { GraphQLEnumType, GraphQLInputObjectType, GraphQLSchema } from 'graphql';
 import { GraphQLEnumValueConfigMap } from 'graphql/type/definition';
 
-import { DEFAULT_PERMISSIONS } from '../../common/constants';
-import { getAllPermissionsMetadata, PermissionDefinition } from '../../common/permission-definition';
+import { getAllPermissionsMetadata } from '../../common/constants';
+import { PermissionDefinition } from '../../common/permission-definition';
 
 const PERMISSION_DESCRIPTION = `@description
 Permissions for administrators and customers. Used to control access to
--- a/packages/core/src/api/resolvers/admin/global-settings.resolver.ts
+++ b/packages/core/src/api/resolvers/admin/global-settings.resolver.ts
@@ -19,10 +19,10 @@ import {
     TypeNode,
 } from 'graphql';
 
+import { getAllPermissionsMetadata } from '../../../common/constants';
 import { ErrorResultUnion } from '../../../common/error/error-result';
 import { UserInputError } from '../../../common/error/errors';
 import { ChannelDefaultLanguageError } from '../../../common/error/generated-graphql-admin-errors';
-import { getAllPermissionsMetadata } from '../../../common/permission-definition';
 import { ConfigService } from '../../../config/config.service';
 import {
     CustomFieldConfig,
--- a/packages/core/src/api/resolvers/admin/job.resolver.ts
+++ b/packages/core/src/api/resolvers/admin/job.resolver.ts
@@ -1,5 +1,6 @@
 import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
 import {
+    JobQueue,
     MutationCancelJobArgs,
     MutationRemoveSettledJobsArgs,
     Permission,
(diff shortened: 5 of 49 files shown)


michaelbromley:
Hi, thanks for taking the time to think about this and contribute!

I'm interested to see how you solve this. It could actually simplify things quite a bit, as we will no longer need to send WorkerMessages in many cases.

By the way, can you make this PR against the `next` branch? Since this will be a breaking change.

michaelbromley:
Hi, just as a bit of a heads-up: The next milestone is v1.0. After that I don't want to make any breaking changes. This proposal is of interest and if it turns out to be beneficial it would be good to get it in the 1.0 release. Timeline for 1.0 is on the order of 2-4 weeks. Let me know of your plans with this 👍

mcfedr:
Hi, yes, I've made some more progess, I'll push it here quite likely later today, I wanted it get the new code running in our app first so that I can be sure I've covered the changes I need.

mcfedr:
Ok, so its grown a little from how I first saw it, but thats why one of the key parts of this is to add a strategy using google pub/sub so that I can see that the changes allow this to happen.

Key changes:
- move the polling logic from the `JobQueue` into the `JobQueueStrategy` - this means that we can control how this happens in different ways - some strategies will be polling style, but some push.
- I created an abstract class `PollingJobQueueStrategy` for this refactor, that allows existing `JobQueueStrategy`s to have minimal changes (basically just add `extends...`)
- I've refactored the `JobQueueStrategy` interface to have just `start` and `stop` methods, and created `InspectableJobQueueStrategy` that has the `find*` methods - this is because some Strategies will be unable to implement these functions. This also touches the admin ui - which now directly calls the methods on the strategy, when it implements `InspectableJobQueueStrategy`. For the moment it just shows empty results for other strategies, I'd like to change this so it can either remove this section of ui, or has a message.
- Then a couple of small changes
- Add at last add a PubSubPlugin, that takes a config of existing topics and subscriptions and runs the queues using this.

For the moment I have not touched the use of nest microservice architecture - but the worker process will now call itself through that system, instead of the server process calling the worker process. This is not ideal, its lots of overhead, but i think this can be a separate PR for the sake of clarity.

michaelbromley:
I've pulled your branch and had a look through the changes so far. I like the idea of moving the polling aspect to be strategy-specific implementation detail. That provides a lot more flexibility.

So as I understand it, this is how it will work, please correct me if I misunderstood:

1. JobQueues (`jobQueueService.createQueue(..)`) will be defined on the worker.
2. To send a job to the queue, the server process would call `configService.jobQueueOptions.jobQueueStrategy.add(data)` directly.
3. The `WorkerService` seems to be almost redundant, there are only 3 calls I can find where it is used independently of the JobQueue.
4. This implies that we could probably remove the WorkerService altogether.
5. This would in turn imply that we would be no longer coupled to the Nestjs microservices architecture - which I guess is what you are referring to in the last paragraph. So eventually a worker Controller will just need to make the call to `createQueue()` in the `onApplicationBootstrap` hook, and handle all the logic from the `process` callback, no need for `@MessagePattern()` -decorated methods.
6. However, I guess we'd still keep the worker as a Nestjs microservice (i.e [here in the bootstrapWorker function](https://github.com/vendure-ecommerce/vendure/blob/dc7b3036e5ade97c733cd6b42b04749d49ada294/packages/core/src/bootstrap.ts#L129)), just because we need it to run as a Nest app of some kind.

All in all I like the look of this. Consolidating the JobQueue and Worker will get rid of a whole lot of boilerplate. Some thoughts/questions:

1. The main drawback I can see right now is the fact that only certain JobQueueStrategies will be able to be inspected. Is there really no way to query the status of e.g. the Google pub/sub topics? Will the user have to go to a separate dashboard to see the status of ongoing jobs in that case?
2. Have you thought about how we can incorporate health checking (as in [this code](https://github.com/vendure-ecommerce/vendure/blob/1b84ea7652e52671bf7b5e3b68b7654d612f85ea/packages/core/src/health-check/health-check.module.ts#L27)) into this new architecture?

mcfedr:
4/5: About the WorkerService - Yes, I would be very much inclined to remove it, I think it adds a lot of complications in deploying the system - I could probably do that as well - as I said, I might want to do it as a new PR, after this, just to keep things clearer, what do you think?

6: For the Nestjs microservice - i think it should be possible to replace it with [`NestFactory.createApplicationContext`](https://docs.nestjs.com/standalone-applications#getting-started), so it will run as a cli command style.

1: Queue status - so with pub/sub, the way its architectured, its not designed to get individual job stats, you can see though the monitoring api overall data, e.g. jobs in queue, jobs in progress, number of failed jobs - but that wouldnt allow one to implement the interface as it currently stands - i'm probably imagining that the pubsub plugin could, with further development, provide its own admin extension, that could show these stats - I've also worked a lot with AWS SQS, which is the equivalent, and similarly it doesnt allow to get the status of individual messages. for AMPQ based systems i think its similar.

2: for the health check - I'm not 100% sure - but i would probably just remove that code. In my experience, if my worker is down for whatever reason, i dont want that to be making my api down - and one useful part of the health check is using it in a load balancer to exclude servers that are not working - if we had a queue stats page, it could show that the number of finished jobs isnt growing, so you could know something is wrong.

michaelbromley:
* 4/5 - yes, do that as a separate PR.
* 6 - nice, I didn't know about NestFactory.createApplicationContext - that looks perfect.
* Queue status: Ok, thanks for the explanation. Yes, supplying a custom UI for monitoring might be the best approach then.
* health check: makes sense.

mcfedr:
ok, so its going on our server today(ish) and will have the chance to debug and see if there are any more changes i think need to be made. then ill come and finish this up, i think some tests need fixing still. then ill work on a new PR, based off this, that will remove the microservice setup.

mcfedr:
Ok - I think at this stage, this PR is finished.

The key breaking changes that are here 

`process` function for jobs (`jobQueueService.createQueue({process})`) - the return type is always a promise, job.complete, job.fail should not be used.

Plugins that use `jobQueueService.createQueue` should make sure the queue is created in `onApplicationBootstrap` not `onVendureBootstrap` as the job queue is run in the worker context.

`JobQueueService` no longer has 'inspection' functions (`getJobs`..) these should be accessed on the strategy directly (checking if it implements them)

`Job.id` is no longer of type `ID` but `string|number` - it will be strategy dependant.

`SqlJobQueueStrategy` now throws when the connection is not available, to prevent jobs silently disappearing, this could be noticed if someone is trying to add jobs during init, and before the strategy has been `init`ed.

These changes should only affect people who have a custom `JobQueueStrategy`

`JobQueueStrategy` no longer has the 'inspection' functions, so implementations might want to move to implementing `InspectableJobQueueStrategy`.

`JobQueueStrategy` now has `start` and `stop` functions - but implementations can extend `PollingJobQueueStrategy` and continue to work with the old `next` and `update` implementations. In this case they should be careful to call `super.init` and `super.destroy`

michaelbromley:
Excellent, I'll properly dive into this later today / tomorrow!

From your experience using this branch in your project, does this fully solve the issues you were having with the current architecture?

mcfedr:
Yes - we are able to run multiple server instances and multiple worker instances and they work independently of each other and jobs are running consistently. 

Also as a bonus, the load on our database server is massively reduced - i think because we had already several servers, and in the old architecture, all of them were constantly polling the server for jobs.

michaelbromley:
This is amazing work. Massive improvement in many areas of the code, and I'm really happy to hear about the perf gains you have got on this so far!

Some issues I ran into:

* I've noticed that there are some failing e2e tests which I have traced back to the facet that the DefaultSearchPlugin is attempting to start its job queue from the `onVendureBootstrap` hook. Same with ElasticSearchPlugin. 
* The EmailPlugin fails to start in dev mode, because the dev mailbox server is created twice, in the server and worker, due to the 
call to `this.devMailbox.serve(options);` appearing in the `onApplicationBootstrap` hook. That part should only run in the server context.
* Two unit tests from job-queue-service.spec.ts are failing. Not dug into why that is.

How would you like to proceed? Do you want to fix these issues and update the PR? I'm also happy to fix those parts if you like, but I'll let you decide, since you are already making further commits on top of this on the rm-microservice branch.

mcfedr:
Ill look into these things, I thought I had all the tests running when i last pushed, but i guess i missed a couple of things here - i had been running smaller parts of the suite individually.

mcfedr:
Re perf - you can see when i deploy this quite clearly:

Screenshot 2021-02-11 at 20 27 13

mcfedr:
I think the latest changes should address these problems - tests and e2e all passing for me.

michaelbromley:
All e2e tests passing for me, but I still get a failure on the JobQueueService unit test:

```
@vendure/core: FAIL src/job-queue/job-queue.service.spec.ts (72.203 s)
@vendure/core:   ● JobQueueService › jobs processed in FIFO queue
@vendure/core:     expect(received).toEqual(expected) // deep equality
@vendure/core:     - Expected  - 1
@vendure/core:     + Received  + 1
@vendure/core:       Array [
@vendure/core:         "COMPLETED",
@vendure/core:         "COMPLETED",
@vendure/core:     -   "PENDING",
@vendure/core:     +   "RUNNING",
@vendure/core:       ]
@vendure/core:       147 |         subject.next();
@vendure/core:       148 |         await tick(1);
@vendure/core:     > 149 |         expect(getStates()).toEqual([JobState.COMPLETED, JobState.COMPLETED, JobState.PENDI
NG]);
@vendure/core:           |                             ^
@vendure/core:       150 |
@vendure/core:       151 |         await tick(queuePollInterval);
@vendure/core:       152 |         expect(getStates()).toEqual([JobState.COMPLETED, JobState.COMPLETED, JobState.RUNNI
NG]);
@vendure/core:       at Object.<anonymous> (src/job-queue/job-queue.service.spec.ts:149:29)
@vendure/core:   ● JobQueueService › retries
@vendure/core:     expect(received).toBe(expected) // Object.is equality
@vendure/core:     Expected: "RETRYING"
@vendure/core:     Received: "RUNNING"
@vendure/core:       262 |         subject.next(false);
@vendure/core:       263 |         await tick(1);
@vendure/core:     > 264 |         expect(testJob.state).toBe(JobState.RETRYING);
@vendure/core:           |                               ^
@vendure/core:       265 |         expect(testJob.isSettled).toBe(false);
@vendure/core:       266 |
@vendure/core:       267 |         await tick(queuePollInterval);
@vendure/core:       at Object.<anonymous> (src/job-queue/job-queue.service.spec.ts:264:31)
```

mcfedr:
Running this again, seems to be some instability in this test, ill try and work out whats causing it.

mcfedr:
everytime at the moment...

```
dev/vendure/vendure/packages/core [ job-queue] [!]   took 22s $ yarn jest --config ./jest.config.js src/job-queue/job-queue.service.spec.ts
yarn run v1.22.10
$ /Users/mcfedr/dev/vendure/vendure/node_modules/.bin/jest --config ./jest.config.js src/job-queue/job-queue.service.spec.ts
 PASS  src/job-queue/job-queue.service.spec.ts
  JobQueueService
    ✓ data is passed into job (25 ms)
    ✓ job marked as complete (17 ms)
    ✓ job marked as failed when exception thrown (17 ms)
    ✓ job marked as failed when async error thrown (17 ms)
    ✓ jobs processed in FIFO queue (44 ms)
    ✓ with concurrency (30 ms)
    ✓ processes existing jobs on start (17 ms)
    ✓ retries (47 ms)
    ✓ sets long-running jobs to pending on destroy (2068 ms)
    ✓ should start a queue if its name is in the active list (16 ms)
    ✓ should not start a queue if its name is in the active list (15 ms)

Test Suites: 1 passed, 1 total
Tests:       11 passed, 11 total
Snapshots:   0 total
Time:        4.35 s, estimated 5 s
Ran all test suites matching /src\/job-queue\/job-queue.service.spec.ts/i.
```

michaelbromley:
Just pulled the latest commit and I can confirm this works once the following change is made to `job-queue.service.spec.ts`

```diff
-import { Injector } from '../common';
+import { Injector } from '../common/injector';
```

This fixes a circular dependency issue which was causing the tests to fail (not sure why it used to pass, must be something to do with the fact you re-ordered some imports).

The other failing test in CI looks due to the fact that the pub-sub plugin is not being built before tests are run. You can remedy this by adding an npm script to that package named "ci":

```
"ci": "yarn build"
```

michaelbromley:
If we can get this fixed, this PR can be merged in today:

> The other failing test in CI looks due to the fact that the pub-sub plugin is not being built before tests are run. You can remedy this by adding an npm script to that package named "ci":

> "ci": "yarn build"

mcfedr:
Merging sounds good :)

s3m3n:
@mcfedr how do you deploy the worker to Google Cloud? Can I see somewhere your implementation?

We run Vendure <1.0 on Cloud Run, which is really great for production backend and scales automatically and wasn't available for single instance Vendure below 1.0. But general concept of Google is to keep the resources free until there is something to do. Cloud Run without http request cuts container's CPU to almost zero, worker sql queries take even 70 seconds to return.

Another way could be use of Cloud Tasks or Cloud Functions but both would have to bootstrap whole Vendure before doing it's job which will work 10 times shorter then bootstrap itself, doesn't make a lot of sense.

It seems we have to keep full VM on Compute Engine to make the worker alive all the time with CPU available and subscribe to Queue, but it's the most expensive solution and not auto scallable.

Am I missing something? How did you manage to deploy your Vendure instance on Google Cloud? Maybe you went wide with K8S and own scaling, not managed by Google?

mcfedr:
@s3m3n We also use Cloud Run for the server process, but having the worker there isn't ideal because if there are no http requests the worker will be killed, even if its in the middle of something. So we also have a kubernetes cluster, with the worker running as a deployment, and autoscaling on its CPU usage. 

I'd like to expand the pub/sub plugin to be able to have it also on cloud run, as a push subscriber, by adding a http controller to the plugin.

martijnvdbrug:
@s3m3n @mcfedr I created a GH discussion topic about it here https://github.com/vendure-ecommerce/vendure/discussions/968

Issue #598, opened 2020-12-21
Custom fields for Administrator and Channel


BeepLoveKarki:
Allowing custom fields for Administrator and Channel can act as a great need for various scenarios including the multi-vendor seller.

rohanrajpal:
> Allowing custom fields for Administrator and Channel can act as a great need for various scenarios including the multi-vendor seller.

+1 to this, I also need this feature.

As of now, is there any way to use a plugin and modify the Administrator and Channel entities? I see that it is possible to add a new entity using plugins https://www.vendure.io/docs/plugins/plugin-examples/defining-db-entity, but can't find anything to modify one.

If there is not, how do I go about actually implementing a custom field in Channel? 

My current research says I simply have to add a custom field column to the [Administrator entity](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/entity/administrator/administrator.entity.ts) like done in the [Collection entity](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/entity/collection/collection.entity.ts#L74). Is this all or am I missing something?

Of course the admin UI might also need some modifications.

rohanrajpal:
> > Allowing custom fields for Administrator and Channel can act as a great need for various scenarios including the multi-vendor seller.
> 
> +1 to this, I also need this feature.
> 
> As of now, is there any way to use a plugin and modify the Administrator and Channel entities? I see that it is possible to add a new entity using plugins https://www.vendure.io/docs/plugins/plugin-examples/defining-db-entity, but can't find anything to modify one.
> 
> If there is not, how do I go about actually implementing a custom field in Channel?
> 
> My current research says I simply have to add a custom field column to the [Administrator entity](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/entity/administrator/administrator.entity.ts) like done in the [Collection entity](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/entity/collection/collection.entity.ts#L74). Is this all or am I missing something?
> 
> Of course the admin UI might also need some modifications.

@michaelbromley do you happen to know how should I go about this?

michaelbromley:
@rohanrajpal There is a _little_ bit more involved in enabling custom fields. Take a look at this commit, which added custom fields to the Fulfillment entity. It shows you all the changes needed: https://github.com/vendure-ecommerce/vendure/commit/380f68eb746aff5c5ceebc8d0dbf49c02b25ab16

[...]

PR #670, opened 2021-01-26
feat(core): Support custom fields on Channel entity

rohanrajpal:
Relates to #598


michaelbromley:
Thank you for your help! Let me know if you also plan to add support for the Administrator custom fields too. Otherwise I will do it myself at some point.

rohanrajpal:
> Thank you for your help! Let me know if you also plan to add support for the Administrator custom fields too. Otherwise I will do it myself at some point.

You're welcome! I don't plan to do it for the Administrator for now, but would for sure like to see it soon.

Issue #677, opened 2021-01-29
Channel aware Assets


martijnvdbrug:
Currently, assets are shared between channels. If an asset is uploaded to channel1and the user switches to channel2, it will see the same set of assets. To fully support multi-channel shops, it's desirable to have assets that are channel aware.

**Describe the solution you'd like**
Assets that are uploaded in the context of the current channel should only be available in the current channel.

* Focalpoints are global: Changing focalpoints changes the focalpoint for this asset on all channels
* Deleting an Asset in a non-default channel will result in the asset still being available in the default channel
*  Optional `deleteFromAllChannels` boolean for `deleteAsset` mutation to delete asset from all channels

 ```graphql
 input AssetDeletionOptions {
   force: Boolean
   deleteFromAllChannels: Boolean
 }
 ```

**Considered out of scope for now**: Assigning assets to multiple channels via the Admin UI.

~~**Question** Should there be something like global assets?~~

I am currently working on this feature.

michaelbromley:
> Question Should there be something like global assets?

What do you mean by this? Assets that are available by default in all Channels without needing to be explicitly assigned?

martijnvdbrug:
> > Question Should there be something like global assets?
> 
> What do you mean by this? Assets that are available by default in all Channels without needing to be explicitly assigned?

I was thinking about assets that are assigned to multiple channels, which seems like a must have now that I think about it.

michaelbromley:
Yes, they should work just like any of the ChannelAware entities:

* They are _always_ assigned to the default Channel.
* They may be optionally assigned to one or more sub-channels.

martijnvdbrug:
* What is the desired behavior when an asset is deleted from a channel? Should it always stay in the default channel?
* What if an asset is deleted from the default channel, but it's also in other channels?
* I assume focalPoints should be channel-aware aswell?

michaelbromley:
Good questions:

>What is the desired behavior when an asset is deleted from a channel? Should it always stay in the default channel?
What if an asset is deleted from the default channel, but it's also in other channels?

I think we need to provide a choice to the user. Here's the current deletion API:

```GraphQL
deleteAsset(id: ID!, force: Boolean): DeletionResponse!

type DeletionResponse {
  message: String
  result: DeletionResult!
}

enum DeletionResult {
  "The entity was successfully deleted"
  DELETED
  "Deletion did not take place, reason given in message"
  NOT_DELETED
}
```
[...]

PR #700, opened 2021-02-11
feat(core): Channel aware assets

martijnvdbrug:
Channel aware Assets.

Fixes  issue: #677

Breaking changes: 
* Many-to-many relation from Asset to Channel. 
* `deleteAssets` and `deleteAsset` mutations changed. More detail in #677 

TODO: 
* Make assets assignable to other channels from the Admin UI
* DB migration


michaelbromley:
🔥🔥🔥🔥
Are you planning to do any more commits or is this ready for review now?

martijnvdbrug:
Ready for review!

michaelbromley:
Excellent! Will get around to this later today.

packages/core/src/entity/asset/asset.entity.ts
```diff
@@ -44,4 +45,8 @@ export class Asset extends VendureEntity implements Taggable {
     @ManyToMany(type => Tag)
     @JoinTable()
     tags: Tag[];
+
+    @ManyToMany(type => Channel)
+    @JoinTable()
+    channels: Channel[];
```
michaelbromley, line 51: You should also explicitly add the `ChannelAware` interface to the class definition

packages/core/src/service/services/asset.service.ts
```diff
@@ -167,7 +184,8 @@ export class AssetService {
         }
         const { assetIds, featuredAssetId } = input;
         if (assetIds && assetIds.length) {
-            const assets = await this.connection.getRepository(ctx, Asset).findByIds(assetIds);
+            // const assets = await this.connection.getRepository(ctx, Asset).findByIds(assetIds);
```
michaelbromley: Needs tidying up?

packages/core/src/service/services/asset.service.ts
```diff
@@ -214,8 +232,18 @@ export class AssetService {
...
+    async delete(
+        ctx: RequestContext,
+        ids: ID[],
+        force: boolean = false,
+        deleteFromAllChannels: boolean = false,
+    ): Promise<DeletionResponse> {
+        const assets = await this.connection.findByIdsInChannel(ctx, Asset, ids, ctx.channelId, {
+            relations: ['channels'],
+        });
+        let channelsOfAssets: ID[] = [];
```
michaelbromley, line 243: I'd use `.forEach()` rather than `.map()` just to be explicit that we are only doing side-effects, and we actually don't map to any new array.

packages/core/src/service/services/asset.service.ts
```diff
@@ -258,25 +358,23 @@ export class AssetService {
...
-            const filename = path.basename(filePath);
-            const mimetype = mime.lookup(filename) || 'application/octet-stream';
-            return this.createAssetInternal(RequestContext.empty(), stream, filename, mimetype);
-        } else {
-            throw new InternalServerError(`error.path-should-be-a-string-got-buffer`);
-        }
+    private async hasDeletePermissionForChannels(ctx: RequestContext, channelIds: ID[]): Promise<boolean> {
+        const permissions = await Promise.all(
+            channelIds.map(async channelId => {
+                return this.roleService.userHasPermissionOnChannel(ctx, channelId, Permission.UpdateCatalog);
```
michaelbromley: I think the permission should be `Permission.DeleteCatalog`

packages/core/src/service/services/asset.service.ts
```diff
@@ -239,6 +267,78 @@ export class AssetService {
                 }),
             };
         }
+        const hasDeleteAllPermission = await this.hasDeletePermissionForChannels(ctx, channelsOfAssets);
+        if (deleteFromAllChannels && !hasDeleteAllPermission) {
+            throw new ForbiddenError();
+        }
+        if (!deleteFromAllChannels) {
+            assets.map(asset =>
+                this.channelService.removeFromChannels(ctx, Asset, asset.id, [ctx.channelId]),
```
michaelbromley: Should be be publishing the `AssetChannelEvent` 'removed' here?

packages/core/src/service/services/product.service.ts
```diff
@@ -228,6 +228,8 @@ export class ProductService {
             channelId: input.channelId,
             priceFactor: input.priceFactor,
         });
+        const assetIds: ID[] = ([] as ID[]).concat(...productsWithVariants.map(p => p.assets.map(a => a.id)));
```
michaelbromley: I think it would make sense to apply the `unique()` helper function here since multiple products may have the exact same assets.

michaelbromley (changes requested):
Great work! There are a few minor points to address, plus one case that needs addressing:

If a Product has assets not in the current channel, they are still returned in the `Product` type. They should be filtered out in the field resolver, in the same way we do with [facetValues](https://github.com/vendure-ecommerce/vendure/blob/e8fcb9959ab0717af087398b075ee357e467f1b4/packages/core/src/api/resolvers/entity/product-entity.resolver.ts#L67-L76). An e2e test for this would be good too.

packages/core/src/entity/asset/asset.entity.ts
martijnvdbrug: > If a Product has assets not in the current channel, they are still returned in the `Product` type. They should be filtered out in the field resolver, in the same way we do with [facetValues](https://github.com/vendure-ecommerce/vendure/blob/e8fcb9959ab0717af087398b075ee357e467f1b4/packages/core/src/api/resolvers/entity/product-entity.resolver.ts#L67-L76). An e2e test for this would be good too.

Ahh yes, agreed. Should this be solved in the `asset.resolver` or in the `asset.service` ?

packages/core/src/service/services/asset.service.ts
martijnvdbrug: yes, done.

packages/core/src/service/services/asset.service.ts
martijnvdbrug: Yes, and I also forgot to `Promise.all()` for both async `removeFromChannels`

packages/core/src/entity/asset/asset.entity.ts
michaelbromley: I just noticed that the resolver _always_ gets the data from the AssetService, so in this case it's probably better to do it directly in the service, as a DB-based solution would be slightly more efficient.

martijnvdbrug: I am struggling a bit with making this testable:
If an Asset belongs to a Product, and we remove the Product from a channel, the assets are also removed from that channel. How could we end up with a Product, that has assets which don't belong to the channel?

michaelbromley: 1. Add a product from default into channel 2.
2. Switch to default channel.
3. Add an Asset "A" (which is only available in default channel) to the product.
4. Switch to channel 2 and query the product with assets.
5. Assert that asset "A" is not in the product.assets array.

martijnvdbrug: I am having trouble getting the `asset.asset.channels` relation on ProductAsset. I created a seperate issue for this case: #717

michaelbromley (approved):

PR #701, opened 2021-02-11
Draft: Rm microservice

mcfedr:
This is based from https://github.com/vendure-ecommerce/vendure/pull/663 so is smaller than it looks at the moment.

Its about removing the indirection of having the worker microservice, as the job queue is now called from the worker process.


mcfedr:
So at this point, it builds and tests pass.

- e2e still need some work.
- the docs will need updating.

Key changes that I'm going for

- two bootstraps - one that calls NestFactory.create and the other NestFactory.createApplicationContext
- the worker process is therefore a 'simple' cli script, that uses the context, and calls `app.get(JobQueueService).start();`
  - This is great for simplicity and make it easy for a user to add there own cli commands that also use the context - in our own site we currently hack together something like this, so that we can run a number of console commands - I think I'd like to combine this with the migration.ts script, so its a single commander cli.
- Uses NestModule to add middlewares for the 'sub systems'  - admin-ui, dev mailbox, assets server - this means we can remove the ProcessContext completely, as the modules need not be aware of the context - the middleware is created on demand by Nest in the server context and these methods are not called for the app context.
- This allows a lot of custom plumbing to be removed - OnVendureBootstrap etc.

michaelbromley:
Reading through the massive diff I must admit I am a bit scared by this change! But after pulling the branch so far and exploring the changes this is pretty mind-blowing how much complexity is erased. The whole idea of server/worker is hidden from the developer for the most part.

2 questions:
1. For those cases where we formerly used `workerOptions.runInMainProcess` (e.g. populating data), how would this be handled now? By manually adding:
    ```ts
     .then(app => {
        app.get(JobQueueService).start();
    })
    ```
    after to `bootstrap` call?
2. You got rid of the `beforeVendureBootstrap` hook. This was used by some dev to e.g. implement a Stripe plugin which required them to define some middleware on the Express app and had to be done _before_ the `app.listen()` call. This scenario seems no longer supported? Would you propose an alternative?

mcfedr:
1. Yes - I think this is the most obvious solution - actually - for populate - It might be worth implementing a `drain` function, that would allow us to `await` the queues to all be empty and automatically exit - I think this would be quite simple to add. then it comes `await app.get(PopulateService).populate(); await app.get(JobQueueServer).drain();`

2. According to the [nest lifecycle](https://docs.nestjs.com/fundamentals/lifecycle-events) 

> onApplicationBootstrap() Called once all modules have been initialized, but before listening for connections.

And the `NestModule. configure` is called before listen as well.

And yes - I was aware this is getting big, and removing a lot of your hard work!

michaelbromley:
1. A `drain` method sounds like a good idea here. 
2. You are right! `configure` looks perfect. I mean, I use it already to handle the user-supplied middleware, but for some reason never thought of it as a solution to that particular problem.

> removing a lot of your hard work

Hey as long as it results in a better product, feel free! After all, I have to build on top of Vendure too and this change is gonna make *my* life much easier too 😄 

I'd also like to say how great it is to have another pair of experienced eyes looking at the code deeply. Vendure has grown slowly over the past 2.5 years, starting simple and gradually getting more complex. Many of the parts you've removed represent incremental solutions to problems that came up over time. This made it difficult for me to take a more big-picture view, as I've gotten used to all the cruft 😆 I'm really excited about this change and look forward to getting it merged!

michaelbromley:
@mcfedr any eta on this being completed? I'm ready to start with the 1.0-beta release once this is in!

mcfedr:
Hi, yes, I'm really hoping to find some time early next week, sorry its taken a while

michaelbromley:
Ok, thanks for the update. If you don't find the time, just let me know and I'll happily finish up fixing the tests and update all the docs.

mcfedr:
To be honest, if you have the time to have a look that would be great, I've gotten a bit stuck with these last few, the e2e is failing all over the place for me, even though I see its working in the ci,

michaelbromley:
OK, I managed to get all the tests passing locally. I'll do the remaining work merging & documenting the changes.

mcfedr:
@michaelbromley Good stuff! Thanks for finishing this! I've hopefully got a couple more things coming this way from the team.

PR #703, opened 2021-02-11
perf: improve order quantity update performance

thomas-advantitge:
We encountered a drastic performance decrease when updating order item quantities by large numbers. 
Example: adding 100, 100 and 900 items:
Schermafbeelding 2021-02-11 om 18 25 49

Also, CPU and memory spike up:
image
(note: chart not related to the previous image with requests)

After investigation, saving a list of entities (order items) seems to be a bottleneck. Two improvements are suggested:
* The loop containing a `.save()` now stores the new order items in an array, saving all at once later on.
* Saving all items at once via `.save(orderItems)` did not improve performance drastically. It appears TypeORM has a performance [issue](https://github.com/typeorm/typeorm/issues/1115) saving bulk items via `.save()`. Using the query builder however improves the performance.

Some execution times (in ms) between `.save()` and `.insert()` on the order items:

Next to the creation loop of order items, the price/promotion calculation afterwards was a second performance bottleneck (having a call of ~12s results in ~7s order item loop and ~5s price calculation). The same approach, saving bulk items at once through the query builder, improves the performance.

### DIFF
--- a/packages/core/src/service/helpers/order-modifier/order-modifier.ts
+++ b/packages/core/src/service/helpers/order-modifier/order-modifier.ts
@@ -150,18 +150,18 @@ export class OrderModifier {
             if (!orderLine.items) {
                 orderLine.items = [];
             }
+            const newOrderItems = [];
             for (let i = currentQuantity; i < quantity; i++) {
-                const orderItem = await this.connection.getRepository(ctx, OrderItem).save(
-                    new OrderItem({
-                        listPrice: orderLine.productVariant.price,
-                        listPriceIncludesTax: orderLine.productVariant.priceIncludesTax,
-                        adjustments: [],
-                        taxLines: [],
-                        line: orderLine,
-                    }),
-                );
-                orderLine.items.push(orderItem);
+                newOrderItems.push(new OrderItem({
+                    listPrice: orderLine.productVariant.price,
+                    listPriceIncludesTax: orderLine.productVariant.priceIncludesTax,
+                    adjustments: [],
+                    taxLines: [],
+                    line: orderLine,
+                }));
             }
+            await this.connection.getRepository(ctx, OrderItem).createQueryBuilder().insert().values(newOrderItems);
+            orderLine.items = orderLine.items.concat(newOrderItems);
         } else if (quantity < currentQuantity) {
             if (order.active) {
                 // When an Order is still active, it is fine to just delete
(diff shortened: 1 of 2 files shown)


michaelbromley:
Excellent find! I prepared a benchmark which:

* Creates an order
* Adds a random ProductVariant to it in a quantity of up to 500
* Repeats for 1 minute

Before:
```
http_req_duration..........: avg=2.26s   min=107.8ms med=1.76s   max=8.45s   p(90)=3.1s    p(95)=7.51s
http_reqs..................: 23      0.383328/s
```

After:
```
http_req_duration..........: avg=101.07ms min=86.99ms  med=97.97ms  max=328.97ms p(90)=107ms    p(95)=129ms
http_reqs..................: 592     9.866615/s
```

Incredible! A ~**25x speedup!**

Makes me wonder about what other easy perf wins are lurking in the code base... 🤔

thomas-advantitge:
We just noticed the order items are not saved. The adjustment call returns the correct values, but executing a subsequent order call doesn't.

Unfortunately, this is due to a stupid mistake in the PR which causes the order items not to be saved due to the lack of `.execute()` 😞. It appears to be required for insert queries, while it isn't for select queries.

Luckily, after adding `.execute()`, the order item loop still performs well (it appears the `save` vs `insert` issue still was the bottleneck).

However, the second bottleneck, the `applyPriceAdjustments` section:
```typescript
await this.connection.getRepository(ctx, OrderItem).createQueryBuilder().insert().values(updatedItems).execute();
await this.connection.getRepository(ctx, ShippingLine).createQueryBuilder().insert().values(order.shippingLines).execute();
```
returns duplicate keys: we need an update query. This is not as easy as a bulk `.insert()`, actually this is exactly the goal of `.save()` , but this causes several seconds delay here too.

Two remarks:
* The shippingLines `.save()` delay seems to be neglectable.
* The best alternative I can come up with right now for the `updatedItems` `.save()` is an insert-update-on-duplicate-key query. It appears TypeORM has an `.orUpdate` function for this behaviour:
```typescript
await this.connection.getRepository(ctx, OrderItem).createQueryBuilder().insert().values(updatedItems).orUpdate({ conflict_target: ['id'], overwrite: ['taxLines', 'adjustments', 'listPriceIncludesTax'] }).execute();
```
I tested this quickly and it seems to be working, reducing ~3s calls to ~80ms, but it might need some extra testing. Also, explicitly listing the database fields which need to be overriden feels a bit ugly, especially as the complex calculation logic is hidden from the order service.

PR #705, opened 2021-02-12
perf: improve order quantity update performance (part 2)

thomas-advantitge:
See https://github.com/vendure-ecommerce/vendure/pull/703#issuecomment-777959555.

* Adds missing `.execute()`.
* Fixes unknown order item ids after insert.
* Insert or override on update for applying price changes.

The tests in the previous PR most likely didn't fail as the request returns the correct results. However subsequent requests make the new order items "disappear" again as they weren't stored correctly due to the missing `.execute()`.

### DIFF
--- a/packages/core/src/entity/order-item/order-item.entity.ts
+++ b/packages/core/src/entity/order-item/order-item.entity.ts
@@ -28,7 +28,7 @@ export class OrderItem extends VendureEntity {
     line: OrderLine;
 
     @EntityId()
-    lineId: ID;
+    lineId: ID; // TypeORM requires this ID field on the entity explicitly in order to save the foreign key via `.insert`
 
     /**
      * @description
--- a/packages/core/src/entity/order-item/order-item.entity.ts
+++ b/packages/core/src/entity/order-item/order-item.entity.ts
@@ -27,6 +27,9 @@ export class OrderItem extends VendureEntity {
     @ManyToOne(type => OrderLine, line => line.items, { onDelete: 'CASCADE' })
     line: OrderLine;
 
+    @EntityId()
+    lineId: ID;
+
     /**
      * @description
      * The price as calculated when the OrderItem was first added to the Order. Usually will be identical to the
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1328,8 +1328,25 @@ export class OrderService {
             promotions,
             updatedOrderLine ? [updatedOrderLine] : [],
         );
+        // await this.connection.getRepository(ctx, OrderItem).save(updatedItems, { reload: false });
+        await this.connection
+            .getRepository(ctx, OrderItem)
+            .createQueryBuilder()
+            .insert()
+            .values(updatedItems)
+            .orUpdate({
+                conflict_target: ['id'],
+                overwrite: [
+                    'initialListPrice',
+                    'listPrice',
+                    'listPriceIncludesTax',
+                    'adjustments',
+                    'taxLines',
+                ],
+            })
+            .updateEntity(false)
+            .execute();
         await this.connection.getRepository(ctx, Order).save(order, { reload: false });
-        await this.connection.getRepository(ctx, OrderItem).save(updatedItems, { reload: false });
         await this.connection.getRepository(ctx, ShippingLine).save(order.shippingLines, { reload: false });
         return order;
     }
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1328,7 +1328,13 @@ export class OrderService {
             promotions,
             updatedOrderLine ? [updatedOrderLine] : [],
         );
-        // await this.connection.getRepository(ctx, OrderItem).save(updatedItems, { reload: false });
+        const updateFields: Array<keyof OrderItem> = [
+            'initialListPrice',
+            'listPrice',
+            'listPriceIncludesTax',
+            'adjustments',
+            'taxLines',
+        ];
         await this.connection
             .getRepository(ctx, OrderItem)
             .createQueryBuilder()
@@ -1336,13 +1342,7 @@ export class OrderService {
             .values(updatedItems)
             .orUpdate({
                 conflict_target: ['id'],
-                overwrite: [
-                    'initialListPrice',
-                    'listPrice',
-                    'listPriceIncludesTax',
-                    'adjustments',
-                    'taxLines',
-                ],
+                overwrite: updateFields,
             })
             .updateEntity(false)
             .execute();
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,7 +46,7 @@
     "@nestjs/platform-express": "7.4.4",
     "@nestjs/terminus": "7.0.1",
     "@nestjs/testing": "7.4.4",
-    "@nestjs/typeorm": "7.1.3",
+    "@nestjs/typeorm": "7.1.5",
     "@types/fs-extra": "^9.0.1",
     "@vendure/common": "^0.18.3",
     "apollo-server-express": "2.18.1",
@@ -74,7 +74,7 @@
     "progress": "^2.0.3",
     "reflect-metadata": "^0.1.13",
     "rxjs": "^6.6.3",
-    "typeorm": "0.2.28"
+    "typeorm": "0.2.31"
   },
   "devDependencies": {
     "@types/bcrypt": "^3.0.0",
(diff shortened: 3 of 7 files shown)


michaelbromley:
Unfortunately the e2e tests are not passing even with this version. Somehow I messed up the GitHub actions config a while back so tests are not being run currently on PRs to the `next` branch, which is why we did not know about the test failures. 

I will just revert the last commit for now, but we should continue to work on a better solution to solve this perf issue.

michaelbromley:
I did some investigation into this on Friday and I came up with:

```ts
const newOrderItems: OrderItem[] = [];
for (let i = currentQuantity; i < quantity; i++) {
    const orderItem = new OrderItem({
        listPrice: orderLine.productVariant.price,
        listPriceIncludesTax: orderLine.productVariant.priceIncludesTax,
        adjustments: [],
        taxLines: [],
        line: orderLine,
    } as any);
    newOrderItems.push(orderItem);
}
const result = await this.connection
    .getRepository(ctx, OrderItem)
    .createQueryBuilder()
    .insert()
    .values(newOrderItems)
    .updateEntity(false)       // suppresses the "select" statement following the insert
    .execute();

const items = await this.connection
    .getRepository(ctx, OrderItem)
    .find({where: {line: orderLine}});  // reload all OrderItems so rest of the logic works
orderLine.items = items;
```

thomas-advantitge:
Added new load test:
* Creates an order
* Adds a random ProductVariant to it in a quantity of 999
* Adjusts the quantity to 1
* Adjusts the quantity to 999 (adjustOrderLine request is heavier than addItemToOrder request)
* Adjusts the quantity to 0 (to prevent order-quantity-limit-exceeded error)
* Repeats for 1 minute

Before:
```
http_req_duration..........: avg=4.24s   min=86.46ms med=3.34s  max=8.32s  p(90)=7.88s  p(95)=7.99s   
http_reqs..................: 17      0.235658/s
```

After:
```
http_req_duration..........: avg=284.96ms min=96.2ms  med=259.67ms max=653.99ms p(90)=359.51ms p(95)=413.64ms
http_reqs..................: 213     3.504635/s
```

Not an increase of **x25** but still about **x12**, reducing order item requests from ~7s to ~400ms.

packages/core/src/entity/order-item/order-item.entity.ts
```diff
@@ -27,6 +27,9 @@ export class OrderItem extends VendureEntity {
     @ManyToOne(type => OrderLine, line => line.items, { onDelete: 'CASCADE' })
     line: OrderLine;
 
+    @EntityId()
+    lineId: ID;
```
michaelbromley: Where is this used?

packages/core/src/service/services/order.service.ts
```diff
@@ -1328,8 +1328,25 @@ export class OrderService {
...
         );
+        // await this.connection.getRepository(ctx, OrderItem).save(updatedItems, { reload: false });
+        await this.connection
+            .getRepository(ctx, OrderItem)
+            .createQueryBuilder()
+            .insert()
+            .values(updatedItems)
+            .orUpdate({
+                conflict_target: ['id'],
+                overwrite: [
```
michaelbromley: For maintainability it would be good to extract this array into a var with type `keyof OrderItem` to make it slightly more resistant to refactoring errors.

packages/core/src/entity/order-item/order-item.entity.ts
thomas-advantitge: This was the main key change needed next to the `.execute()`: without it the `.insert()` call ignored the foreign key `lineId` when bulk inserting the order items (a bit strange, had a look in the TypeORM insert query builder to understand it). Resulting in `null` columns in the database.

I however changed the `OrderModifier` code to use `lineId: orderLine.id` instead of `line: orderLine` to be more clear. Awaiting the e2e tests to see if this still behaves the same. Otherwise, I'll revert it and add a comment to the `@EntityId` line.

thomas-advantitge:
@michaelbromley any idea why the postgres & sqljs are failing? They seem to pass locally (tried same postgres version as in the pipeline through docker).

image

michaelbromley:
I'll look into what's going wrong with the tests. There seems to be existing issues with the DefaultSearchPlugin tests on the next branch, so I'll fix those and then take a look at this branch.

michaelbromley:
I'm getting the exact same failures locally on sql.js & postgres as in CI. MySQL passing. Really strange that they are totally different errors. I guess these `insert()` apis are much less tested and rougher than `save()`. I'll see if I can figure out what's up.

michaelbromley:
Working on the Postgres issue:  The source of the failures is the changes in the `OrderService.applyPriceAdjustments()` method. 

Using the `insert().orUpdate()` is not working - rather than updating existing OrderItems, new OrderItems are being created. This is evidenced by the fact that prices are doubled in the failing tests. 

When I log the generated query we get: 
```sql
INSERT INTO "order_item"("createdAt", "updatedAt", "initialListPrice", "listPrice", "listPriceIncludesTax", "adjus
tments", "taxLines", "cancelled", "lineId", "refundId") 
VALUES 
  ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10), 
  ($11, $12, $13, $14, $15, $16, $17, $18, $19, $20) 
ON CONFLICT  ( "id" )  DO UPDATE SET "initialListPrice" = EXCLUDED."initialList
Price", "listPrice" = EXCLUDED."listPrice", "listPriceIncludesTax" = EXCLUDED."listPriceIncludesTax", "adjustments" =
EXCLUDED."adjustments", "taxLines" = EXCLUDED."taxLines"
```

I'm not certain, but it might have something to do with this issue: https://github.com/typeorm/typeorm/issues/2215, i.e. the "id" column is not part of the "INSERT" statement, and perhaps this is when the update is not taking place?

michaelbromley:
OK, getting somewhere now:

Trying to find out why the "id" column is omitted for Postgres, I traced it down to this method in TypeORM: 
https://github.com/typeorm/typeorm/blob/9d2df28b4c8c22024566a86a231db64ae2c09161/src/query-builder/InsertQueryBuilder.ts#L355-L379

```ts
    /**
     * Gets list of columns where values must be inserted to.
     */
    protected getInsertedColumns(): ColumnMetadata[] {
        if (!this.expressionMap.mainAlias!.hasMetadata)
            return [];

        return this.expressionMap.mainAlias!.metadata.columns.filter(column => {

            // if user specified list of columns he wants to insert to, then we filter only them
            if (this.expressionMap.insertColumns.length)
                return this.expressionMap.insertColumns.indexOf(column.propertyPath) !== -1;

            // skip columns the user doesn't want included by default
            if (!column.isInsert) { return false; }

            // if user did not specified such list then return all columns except auto-increment one
            // for Oracle we return auto-increment column as well because Oracle does not support DEFAULT VALUES expression
            if (column.isGenerated && column.generationStrategy === "increment"
                && !(this.connection.driver instanceof OracleDriver)
                && !(this.connection.driver instanceof AbstractSqliteDriver)
                && !(this.connection.driver instanceof MysqlDriver)
                && !(this.connection.driver instanceof AuroraDataApiDriver))
                return false;

            return true;
        });
    }
```

Which shows that any generated, auto-increment columns are omitted _unless_ the driver is Oracle, SQLite, MySQL, Aurora. But above that you can see that this block is skipped _if the columns are explicitly specified_. This lead to a solution:

```ts
 const updateFields: Array<keyof OrderItem> = [
     'initialListPrice',
     'listPrice',
     'listPriceIncludesTax',
     'adjustments',
     'taxLines',
 ];
 await this.connection
     .getRepository(ctx, OrderItem)
     .createQueryBuilder()
     .insert()
     .into(OrderItem, [...updateFields, 'id', 'lineId']) 
     .values(updatedItems)
     .orUpdate({
         conflict_target: ['id'],
         overwrite: updateFields,
     })
     .updateEntity(false)
     .execute();
```

Regarding SQLite failure, it seems to be due to [this line](https://github.com/advantitge/vendure/blob/42bf55a7437d82fdaf3706b525421904ed0c2042/packages/core/src/service/helpers/order-modifier/order-modifier.ts#L173), where the newly-created OrderItems are added to the OrderLine. These new items, however, do not yet have an id, since they haven't been persisted to the DB. This seems the be the root of the sqlite issues. Replacing that line with:
```ts
orderLine.items = await this.connection
  .getRepository(ctx, OrderItem)
  .find({ where: { line: orderLine } });
```
resolves the issue, without seeming to sacrifice much in terms of performance. I'll add before/after benchmarks using the fixes described in this comment to verify.

thomas-advantitge:
> Regarding SQLite failure, it seems to be due to [this line](https://github.com/advantitge/vendure/blob/42bf55a7437d82fdaf3706b525421904ed0c2042/packages/core/src/service/helpers/order-modifier/order-modifier.ts#L173)

Hm strange, thought ids were populated by the line before: `newOrderItems.forEach((item, i) => (item.id = identifiers[i].id));`.

Anyway, thanks for the deep dive into TypeORM and fixing the tests! 🎉

PR #707, opened 2021-02-12
feat: Added custom sender and generator option to emailPlugin config

WilliamMilne:
@michaelbromley - is this what you had in mind? The default options are still instantiated if nothing is passed to the plugin via `EmailPluginOptions`. For a developer to replace these items, they would simply need to create a class that implements `EmailSender` (which I extracted into an interface and renamed the original to `DefaultEmailSender`) and a class that implements EmailGenerator and pass instances of them into the plugin at the VendureConfig level. 

Please let me know if there is a more preferable way to go about this. Thanks!

### DIFF
--- a/packages/email-plugin/src/email-processor.ts
+++ b/packages/email-plugin/src/email-processor.ts
@@ -4,10 +4,16 @@ import fs from 'fs-extra';
 import { deserializeAttachments } from './attachment-utils';
 import { isDevModeOptions } from './common';
 import { loggerCtx } from './constants';
-import { EmailSender } from './email-sender';
+import { DefaultEmailSender } from './email-sender';
 import { HandlebarsMjmlGenerator } from './handlebars-mjml-generator';
 import { TemplateLoader } from './template-loader';
-import { EmailPluginOptions, EmailTransportOptions, IntermediateEmailDetails } from './types';
+import {
+    EmailGenerator,
+    EmailPluginOptions,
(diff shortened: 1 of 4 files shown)


michaelbromley:
Yes, this is exactly what I had in mind! Good work.

Minor point: in the options object, can you rename `customEmailSender` to just `emailSender`? The reason is that the general convention in Vendure is not to add "custom" to configurable parts like strategies/handlers, otherwise everything would be "customThingHandler".

With this change, are you able to successfully integrate with the sendgrid API?

WilliamMilne:
I modified the names to match your suggestion - that makes a lot of sense! 

I intend to work on the Sendgrid API plugin today so if I have success with that I'll let you know. Thank you for your help and guidance on this!

michaelbromley:
Thanks. So, let's see how you get on with the Sendgrid integration before merging this. Then we know that the theoretically correct APIs hold up against the harsh reality of real-world usage 😂

WilliamMilne:
Hey @michaelbromley - I was able to run a successful test on my local machine with an injected EmailSender that uses the Sendgrid API! We aren't production live with our instance of vendure yet so I can't get full real-world usage. What should the next steps be for this PR? Is there a way to re-run the checks here on github?

michaelbromley:
I re-ran the tests, all good 👍 
Thanks for the contribution, I think this will prove a very useful addition.

PR #714, opened 2021-02-15
feat(core): Add order line limit to Vendure configuration

thomas-advantitge:
Yet another quite small generic feature to limit orders per order line. 
Could be a choice to not have this in the core, just a suggestion.

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -359,10 +359,14 @@ export class OrderService {
         customFields?: { [key: string]: any },
     ): Promise<ErrorResultUnion<UpdateOrderItemsResult, Order>> {
         const order = await this.getOrderOrThrow(ctx, orderId);
+        const orderLineByVariant = order.lines.find(line =>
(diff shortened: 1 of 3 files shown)


packages/core/src/service/services/order.service.ts
```diff
@@ -359,10 +359,14 @@ export class OrderService {
         customFields?: { [key: string]: any },
     ): Promise<ErrorResultUnion<UpdateOrderItemsResult, Order>> {
         const order = await this.getOrderOrThrow(ctx, orderId);
+        const orderLineByVariant = order.lines.find(line =>
```
michaelbromley: I think this line will fail when OrderLines exist with the same ProductVariant but different custom fields. This is what the `getExistingOrderLine` method handles. You could move the `existingOrderLine` call up above the validation.

michaelbromley (changes requested):
Thanks! I think its worth having in core. Just take a look at the point about custom field handling.

michaelbromley:
Thanks!

PR #735, opened 2021-02-27
Added hint on how manage sessions in authentication docs

Draykee:
Added hint on how manage sessions in authentication docs


michaelbromley:
Thank you!

PR #748, opened 2021-03-05
fix(core): RememberMe args not passed correctly for NativeAuthenticationStrategy

hendrik-advantitge:
When using the NativeAuthenticationStrategy, the remember me parameter resides under `input`. The `authenticateAndCreateSession( ... )` method expects `MutationAuthenticateArgs` to be passed, where rememberMe is on the same level as `input`.


michaelbromley:
Thank you!

Issue #732, opened 2021-02-25
Pass data from PromotionCondition check() to PromotionAction execute()


karel-advantitge:
**Is your feature request related to a problem? Please describe.**
Currently there is no continuity between the conditions and actions of Promotions. Because of this, in certain cases, we are required to repeat certain calculations and thus, pass some argument to both condition and action. Because of this, sometimes it's preferable to nog have a condition at all and to all checks and calculation in the action.

Example:
The store has a promotion: if you buy 3 sodas, you get a candy bar for free. The candy bar can be any of a list of candybars, with differing prices. This promotion can be cumulated, i.e. If you buy 6 sodas you get 2 candy bars and so on. If more candy bars are in the order, the most expensive one should be the one they get for free. This promotion should be applied to the individual items, not to the order.

Example current execution:
- ~~A PromotionCondition with arguments:~~
    - ~~a list of all sodas~~
    - ~~the required amount of sodas (3 in the case of tis example but we generalize the promotion)~~
- A PromotionItemAction with arguments:
    - a list of all sodas
    - the required amount of sodas (3 in the case of tis example but we generalize the promotion)
    - a list of all candy bars
    - the amount of free candy bars (1 in the case of tis example but we generalize the promotion)

The PromotionCondition can check if the order has the required amount of sodas by comparing with the list of sodas.
The PromotionItemAction can check if the current item is a candy bar by comparing with the list of candy bars. But it needs a list of all sodas as well, because it has to determine how many candy bars should be discounted and whether the current item is one of the candybars that should be discounted. This makes the PromotionCondition irrelevant. The 'condition check' done by the PromotionCondition has to be repeated by the PromotionItemAction anyway (for every item separately)

**Describe the solution you'd like**
Data could be passed from the conditions to the actions.
We could allow the check function of PromotionCondition to not just return a boolean, but also a state, let's say `PromotionConditionState` which is a json object containing calculated values. The test function on Promotion performs all check function in the way it does now, but it also accumulates all PromotionConditionStates in a key-value map where the key is the condition code, resulting in an object, let' say `PromotionState`, of type `{ [code: string]: PromotionConditionState }`.
This object can then be passed to the PromotionAction execute function in the Promotion apply function.
Looking at the code the `await promotion.test(ctx, order)` and `await promotion.apply(ctx, { order })` are always used close together. The test function could return the boolean it does now, as well as the PromotionState. The PromotionState object can then be passed to the apply function.

Example execution with solution:
[...]

PR #752, opened 2021-03-09
feat(core): add promotion state and promotion action-condition dependency

karel-advantitge:
#732

Had a bit of trouble trying to get the typing right. I started with `PromotionActionConfig['conditions']` being an array of string values. But with the array I couldn't get the typing to interpret the string values as literal, while still handling an empty/missing conditions array correctly.
So I made it more analogous to the ConfigArgs object. The difference being that for now in the ConditionConfig, there's not really any useful data to be passed in the values, it's mostly the keys we use. There's a `required` field now, but perhaps there is not a lot of use to it. My guess is in most cases the value for `required` will be `true`
So perhaps somebody has a better solution?

### DIFF
--- a/packages/core/src/common/types/adjustment-source.ts
+++ b/packages/core/src/common/types/adjustment-source.ts
@@ -3,6 +3,8 @@ import { ID } from '@vendure/common/lib/shared-types';
 
 import { VendureEntity } from '../../entity/base/base.entity';
 
+export type TestResult = boolean | object;
+
 export abstract class AdjustmentSource extends VendureEntity {
     type: AdjustmentType;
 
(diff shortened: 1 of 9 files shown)


michaelbromley:
Thanks! Could you make the PR against the `master` branch please? Then I'll take a closer look at the typings and see what the best solution might be.

michaelbromley:
I've spent the morning diving deep into the TS advanced typing rabbit-hole and I found a really nice solution which:

* Allows you to specify the `conditions` of the PromotionAction as an array of PromotionCondition objects
* Extracts the return value of the `PromotionCondition.check()` function automatically
* Gives type-safety based on that return value.

I will add these changes and manually merge this branch locally. Thank you for your work on this!

https://user-images.githubusercontent.com/6275952/110929416-379e8c00-8328-11eb-8bb0-ea9009d7abac.mp4

PR #757, opened 2021-03-11
fix(core): Filter Promotions on Channel before applying to Order

hendrik-advantitge:
When an Order is placed on a certain Channel, Promotions from other Channels are applied as well. Using the PromotionService, the filtering per Channel is automatically applied.

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1345,9 +1345,9 @@ export class OrderService {
                 item.listPriceIncludesTax = priceResult.priceIncludesTax;
             }
         }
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks for this. Are the e2e tests passing for you locally?

hendrik-advantitge:
No, they fail because the promotions are only created on the default channel but some tests are executed on other channels where the promotions do not exist.

michaelbromley:
Ok got it. Do you plan on updating the tests? Fine either way, if not I will update them.

hendrik-advantitge:
Hi Michael, if you could do it that'd be great. Thank you very much!

Issue #764, opened 2021-03-15
Custom field population for relation type


Draykee:
**Is your feature request related to a problem? Please describe.**
Vendure now supports [relation-custom-field-type](https://www.vendure.io/blog/2021/03/announcing-vendure-v1.0.0-beta/#relation-custom-field-type). And I would like to have the opportunity to define custom fields  values with `type: 'relation'` in the `products.csv` file to add those during population. The problem right now is that every value is parsed as a string and not as an object that would be required for a `relation-custom-field-type`.

**products-with-customfield.csv**

| name                  | product:relationalCustomField |
|---|---|
| Sample Product | { "id": 1 }                                    |

**Describe the solution you'd like**
Check if a value can be parsed as an object to correctly interprete the value for the customFields.

**Simple implementation:**
```JS
function isJSON(str) {
    try {
        return (JSON.parse(str) && !!str);
    } catch (e) {
        return false;
    }
}

// import-parser.ts#270
function parseCustomFields(prefix: 'product' | 'variant', r: RawProductRecord): { [name: string]: string } {
    return Object.entries(r)
        .filter(([key, value]) => {
            return key.indexOf(`${prefix}:`) === 0;
        })
        .reduce((output, [key, value]) => {
            const fieldName = key.replace(`${prefix}:`, '');
            return {
                ...output,
               // use value as object if possible
                [fieldName]: isJSON(value) ? JSON.parse(value) : value,
            };
        }, {});
}
```

PR #765, opened 2021-03-15
populate relational custom field of product

Draykee:
Possible solution for #764

Adds the value (from csv cell) as an object to the custom field (defined by column header) if possible.


Draykee:
Oh. I think I f*t up the PR a littble bit with former commits. Is that a problem?

michaelbromley:
No, I can squash it into a single commit when I merge. Will take a proper look tomorrow 👍

michaelbromley:
Have you testing this solution? It looks like a partial solution to me. I.e. it will parse a custom field relations into `{ id: 1 }` but then there needs to be some logic in the `importer.ts` file to actually set the relation in the DB. Right now that object value will just get passed in here: https://github.com/vendure-ecommerce/vendure/blob/e8fcb9959ab0717af087398b075ee357e467f1b4/packages/core/src/data-import/providers/importer/importer.ts#L163

at which point I think something might break.

Draykee:
Yeah I tested it and it worked magically. It was just an idea an I was suprised it worked that well. I think providing the `id` is already enough for the relation.
In my case I wanted to add a relation to a User (a `owner`). I was even able to access the field via the API (before I made the custom field private)

Draykee:
Okay. But this time something went wrong again. just wanted to push https://github.com/vendure-ecommerce/vendure/pull/765/commits/47470f7ff241bd8de02a4a7a635ee2ad99d9ffe5 - a little e2e test to maybe prove that passing an object with "id" is enough.

... or I'm still misunderstanding you concerns.

michaelbromley:
I'll take a look at this tomorrow. Cool that it just works! Thinking more about it I see how that makes sense now. Regarding the commit history, I'll see it I can cleat it up a bit.

michaelbromley:
OK this looks good. 2 points:

1. Might be easiest to merge if you just create a brand new branch from the current `master` branch and then cherry pick only the relevant commits. You might need to create a brand new PR for this, I'm not sure if GH lets you switch branches like that.
2. The e2e is good - but do you think you can incorporate it into the existing `import.e2e-spec.ts` suite?

Draykee:
Made a new PR #776 to `master` with extended `import.e2e-spec.ts` suite

PR #776, opened 2021-03-20
populate relational custom field of product

Draykee:
Possible solution for #764

Adds the value (from csv cell) as an object to the custom field (defined by column header) if possible.
Extends the import e2e test suite


Draykee:
Okay I see that one unit test failed (`src/data-import/providers/import-parser/import-parser.spec.ts`)

```
"customFields": Object {
@vendure/core:     -           "volumetric": "243",
@vendure/core:     +           "volumetric": 243,
@vendure/core:  },
```

Does `JSON.parse("243")` really parse it to `243`? Or is this not a fault of Commit 
282c8e6

michaelbromley:
Oh yeah I think the issue is that `isJSON()` function. So `JSON.parse("123")` does indeed evaluate to the number `123`. I think the `isJSON()` function is not exactly correctly named - since in this case you want to know whether the value _is a string which can be parsed into an object with an id property_.

So you could have something like:

```ts
isRelationObject(value: string) {
  try {
    const parsed = JSON.parse(value);
    return parsed && parsed.hasOwnProperty('id');
  } catch(e) {
    return false;
  }
}
```

would that make sense to you?

Draykee:
I guess we could improve the performance if we would actually return the already parsed value (or the value itself if it failed). But on the other hand I believe that the current implementation is faster to understand and the impact is not that high.

Thanks @michaelbromley for your review!

michaelbromley:
Yes, I think the perf impact of doing JSON.parse twice rather than once is not significant in this case. Thanks for your contribution!

PR #781, opened 2021-03-22
feat(admin-ui): i18n, German translations

Wanztwurst:
Added many and changed some German translations


michaelbromley:
Dankeschön!

PR #786, opened 2021-03-23
feat(core): Add ability to get variants for a specific product in productVariants query

dkostenko-jsninja:
In Vendure we have 2 ways to receive product variants by using product and productVariants queries.

product query finds all product variants for a specific product and in case the product has a lot of variants, the performance of this query will be very slow.

productVariants query has options for pagination and filtering but by using this query we cannot receive variants only for a specific product.

Changes in this pull request will fix this problem. So, in the result, we will have the ability to fetch either all variants or only for the specific product by using the productVariants query.


michaelbromley:
Hi,
thanks for the contribution! Out of interest, how many product variants are you dealing with?

Could you rebase this PR onto the master branch? - there were some failing tests which I've now fixed, so then we can make sure your changes are passing in CI too.

dkostenko-jsninja:
Hello!

I've just done a rebase onto the master branch.

The problem exists for any product which has hundreds of variants.

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -130,8 +131,21 @@ export class ProductVariantService {
...
-        const qb = this.connection.getRepository(ctx, ProductVariant).createQueryBuilder('productVariant');
+    getVariantsByProductId(
+        ctx: RequestContext,
+        productId: ID,
+        options: ListQueryOptions<ProductVariant> = {},
+    ): Promise<PaginatedList<Translated<ProductVariant>>> {
+        const rawConnection = this.connection.rawConnection;
+        const qb = this.connection.getRepository(ctx, ProductVariant).createQueryBuilder('productvariant');
+
+        const filter = parseFilterParams(rawConnection, ProductVariant, options.filter);
```
michaelbromley: I notice that quite a significant amount of logic here is being duplicated from the `ListQueryBuilder.build()` method. I am wondering whether it would be possible to use the ListQueryBuilder, and then add the required clauses (ie. `.andWhere('productvariant.productId = :productId', { productId })` to the returned SelectQueryBuilder object.

Did you try that approach yet?

michaelbromley (commented):
Generally looks good - just see my comment about a possible way to do this with less code duplication. Also, an e2e test for this would be good - do you feel confident to add one? If so it would go [here](https://github.com/vendure-ecommerce/vendure/blob/e5182b73fa1634d46f9d8b14003b74ce72bb717e/packages/core/e2e/product.e2e-spec.ts#L340)

packages/core/src/service/services/product-variant.service.ts
dkostenko-jsninja: Hello! Thank you! I agree that would be better to use `ListQueryBuilder.build()` here. I will do it.

dkostenko-jsninja:
Hello! I implemented `ListQueryBuilder.build()` in `ProductVariantService.getVariantsByProductId()` and added e2e test for "get product variants by product id" case.

packages/core/e2e/product.e2e-spec.ts
```diff
@@ -511,6 +511,45 @@ describe('Product resolver', () => {
...
+                GetProductVariantList.Query,
+                GetProductVariantList.Variables
+            >(GET_PRODUCT_VARIANT_LIST, {
+                options: {
+                    take: 3,
+                    sort: {
+                        price: SortOrder.ASC,
+                    },
+                },
+                productId: '1',
```
michaelbromley: The id should be `'T_1'`

michaelbromley (changes requested):
Thanks, the changes look good!

Just one (very minor) change to the e2e test, and then rebase to latest master because my own bugs will make the ealsticsearch e2e test fail at the current commit 🙃

dkostenko-jsninja:
Hello! I did a rebase and replaced `productId: '1'` with `productId: 'T_1'` in `product.e2e-spec.ts`

michaelbromley (approved):

michaelbromley:
Thank you for your contribution! 👍

PR #797, opened 2021-03-27
Bug/unordered breadcrumbs

### LINKED ISSUE #779: Collection.breadcrumbs not correctly ordered
**Describe the bug**
From a Slack message:

> Hi! Is this the planned behavior in shop-api that breadcrumbs in collection query are returned in reverted sorting at the middle of the list?
For example, collection hierarchy:
collection1 -> collection2 -> collection3 
When we query collection3 breadcrumbs we receive
 [__root_collection__, collection2, collection1, collection3].
As i can see, that occurs due to order of getting ancestors here 

There _is_ currently a passing e2e test for breadcrumb order, but on investigation it appears that it works only because the breadcrumb order matches the order of creation of each collection.

**Expected behavior**
Breadcrumbs should always follow the path from root -> current collection in the correct order.

 - @vendure/core version: 1.0.0-beta.4
 - Nodejs version: any
 - Database (mysql/postgres etc): testing with MariaDB, probably affects all.

Izayda:
Fix #779

### DIFF
--- a/packages/core/e2e/collection.e2e-spec.ts
+++ b/packages/core/e2e/collection.e2e-spec.ts
@@ -65,6 +65,9 @@ describe('Collection resolver', () => {
     let electronicsCollection: Collection.Fragment;
     let computersCollection: Collection.Fragment;
     let pearCollection: Collection.Fragment;
+    let electronicsBreadcrumbsCollection: Collection.Fragment;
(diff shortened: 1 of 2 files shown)


michaelbromley:
Good work, thank you for the fix!

PR #798, opened 2021-03-28
perf(core): Use faster bcrypt implementation

Brooooooklyn:
Since @node-rs is faster and easier to install, see: https://github.com/napi-rs/node-rs/tree/main/packages/bcrypt

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,10 +46,10 @@
     "@nestjs/terminus": "7.1.0",
     "@nestjs/testing": "7.6.13",
     "@nestjs/typeorm": "7.1.5",
+    "@node-rs/bcrypt": "^1.2.1",
     "@types/fs-extra": "^9.0.1",
     "@vendure/common": "^1.0.0-beta.5",
     "apollo-server-express": "2.21.0",
-    "bcrypt": "^5.0.0",
     "body-parser": "^1.19.0",
     "chalk": "^4.1.0",
     "commander": "^7.1.0",
@@ -77,7 +77,6 @@
     "typeorm": "0.2.31"
   },
   "devDependencies": {
-    "@types/bcrypt": "^3.0.0",
     "@types/cookie-session": "^2.0.41",
     "@types/csv-parse": "^1.2.2",
     "@types/express": "^4.17.8",
--- a/packages/core/src/service/helpers/password-cipher/password-ciper.ts
+++ b/packages/core/src/service/helpers/password-cipher/password-ciper.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@nestjs/common';
-import bcrypt from 'bcrypt';
+import bcrypt from '@node-rs/bcrypt';
 
 const SALT_ROUNDS = 12;
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -2942,20 +2942,10 @@
     npmlog "^4.1.2"
     write-file-atomic "^3.0.3"
 
-"@mapbox/node-pre-gyp@^1.0.0":
-  version "1.0.0"
-  resolved "https://registry.npmjs.org/@mapbox/node-pre-gyp/-/node-pre-gyp-1.0.0.tgz#2b809e701da0f6729b47fe78ad4b9dc187a7d2e5"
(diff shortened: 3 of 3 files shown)


michaelbromley:
Hi,
Thanks for the PR. Can you tell me more about the "easier to install" part? I couldn't see any specifics about that in your repo.

Brooooooklyn:
> Hi,
> Thanks for the PR. Can you tell me more about the "easier to install" part? I couldn't see any specifics about that in your repo.

`bcrypt` depends on `node-pre-gyp` which would download pre-build binary from github releases, and it's have many runtime unrelated dependencies. And if download fails, it will try to build from `c` source files, and it requires more system dependencies.

`@node-rs/bcrypt` is pure npm package without postinsatll scripts and system dependencies.

jeancx:
> > Hi,
> > Thanks for the PR. Can you tell me more about the "easier to install" part? I couldn't see any specifics about that in your repo.
> 
> `bcrypt` depends on `node-pre-gyp` which would download pre-build binary from github releases, and it's have many runtime unrelated dependencies. And if download fails, it will try to build from `c` source files, and it requires more system dependencies.
> 
> `@node-rs/bcrypt` is pure npm package without postinsatll scripts and system dependencies.

Hi the main porpose of the node bcrypt is still open source and not compiled is for security reasons, any programmer can see what is happening when it's is being compiled, mainly because the password cryptography is the security core of all system.
So your repository of a, pre-compiled bcrypt may sound a very good good idea, but after it is compiled, how people with no skill of reverse enginering may know if there is any backdoor, or if there is a cryptocoin sucking some resources of the server each time it generates a password hash?
I don't want to sound suspicious, i like the idea of a easy way of a pre compiled code since i had to compile with node-gyp for aws lambda environment, it take me some time to make all work, but is a time well spent if you know you are doing the safe way.
I just want to give my opnion about security issues, as a humble programmer and not a bytecode  reverse enginering guy.

Brooooooklyn:
Oh it's pre compiled for easier installation, but the source code itself is open source: https://github.com/napi-rs/node-rs/tree/main/packages/bcrypt/src. And the pre compiled progress is open source too: https://github.com/napi-rs/node-rs/blob/main/.github/workflows/ci.yaml https://github.com/napi-rs/node-rs/actions/runs/696372751

jeancx:
> Oh it's pre compiled for easier installation, but the source code itself is open source: https://github.com/napi-rs/node-rs/tree/main/packages/bcrypt/src. And the pre compiled progress is open source too: https://github.com/napi-rs/node-rs/blob/main/.github/workflows/ci.yaml https://github.com/napi-rs/node-rs/actions/runs/696372751

Thanks for the answer, I had seen the source code before, but I hadn't analyzed the CI build. I will test your repository later, and make some benchmarks, and if you do well I will put it into production, in some of my projects, but as we say here in Brazil, "Insurance died of old age". 

But I was just giving my opinion as a Vendure user, the final decision is up to the Vendure maintainers.

michaelbromley:
Thanks for both of your input. I am inclined to be somewhat conservative this a dependency as important as bcrypt. From looking at npm stats, your project has seen some decent adoption over the past 6 months which is promising. But the current bcrypt packages has almost 2 orders of magnitude more downloads and hence can be considered more widely tested.

On the other hand, the simplified install is certainly an advantage.

One other point I wonder about is how important is the speed? I thought one of the points of bcrypt is that it is kinda slow to make brute-force attacks less viable. Can you comment on that?

Brooooooklyn:
> One other point I wonder about is how important is the speed? I thought one of the points of bcrypt is that it is kinda slow to make brute-force attacks less viable. Can you comment on that?

The ***Faster*** I mentioned above I mean `Faster when produced the same hash`. The algorithm under the hood is almost same between `@node-rs/bcrypt` and `bcrypt`.  `@napi-rs/bcypt` is faster because the `Rust compiler` and `llvm` produced more optimized binary.

michaelbromley:
I'm going to close this as I don't plan to change the bcrypt lib at this point for reasons mentioned above. Thank you again for bringing the alternative to my attention - it may be worth re-visiting in future once the lib has more of a track record and widespread adoption.

PR #801, opened 2021-03-31
Bug/deleted slugs

### LINKED ISSUE #800: Shop/Admin API product query with slug input returns nothing for existed product
**Describe the bug**
Shop API product query with slug input returns nothing for existed product.

**To Reproduce**
1. Create product with slug 'delete-me'
2. Delete that product
3. Create another product with slug 'delete-me'
4. Try to fetch this product by slug
5. Get nothing

**Expected behavior**
Returned product by slug

 - @vendure/core version: v1.0.0-beta.4
 - Nodejs version v14.16.0

**Additional context**
How to get such broken slugs:
`SELECT * FROM (
                  SELECT slug,
                         COUNT(*),
                         SUM(CASE WHEN "deletedAt" is not null then 1 else 0 end) as deleted_count,
                         SUM(CASE WHEN "deletedAt" is null then 1 else 0 end)     as not_deleted_count
                  FROM product p
                           JOIN product_translation pt on p.id = pt."baseId"
                  GROUP by slug
              ) t
WHERE not_deleted_count>0 and deleted_count>0
order by 3 desc;`

Bug exist here: https://github.com/vendure-ecommerce/vendure/blob/5c837b85c3e03007acff7e31c2a6a5e061766200/packages/core/src/service/services/product.service.ts#L138

Izayda:
Fix #800

### DIFF
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -136,18 +136,24 @@ export class ProductService {
     }
 
     async findOneBySlug(ctx: RequestContext, slug: string): Promise<Translated<Product> | undefined> {
-        const translations = await this.connection.getRepository(ctx, ProductTranslation).find({
-            relations: ['base'],
-            where: { slug },
-        });
-        if (!translations?.length) {
-            return;
-        }
-        const bestMatch =
-            translations.find(t => t.languageCode === ctx.languageCode) ??
-            translations.find(t => t.languageCode === ctx.channel.defaultLanguageCode) ??
-            translations[0];
-        return this.findOne(ctx, bestMatch.base.id);
+        const qb = this.connection.getRepository(ctx, Product).createQueryBuilder('product');
+        FindOptionsUtils.applyFindManyOptionsOrConditionsToQueryBuilder(qb, { relations: this.relations });
+        // tslint:disable-next-line:no-non-null-assertion
+        FindOptionsUtils.joinEagerRelations(qb, qb.alias, qb.expressionMap.mainAlias!.metadata);
+        return qb
+            .innerJoin(ProductTranslation, 'product_translation', 'product_translation.baseId=product.id')
+            .leftJoin('product.channels', 'channel')
+            .andWhere('product_translation.slug = :slug', { slug })
+            .andWhere('product_translation.languageCode = :languageCode', { languageCode: ctx.languageCode })
+            .andWhere('product.deletedAt IS NULL')
+            .andWhere('channel.id = :channelId', { channelId: ctx.channelId })
+            .getMany()
+            .then(products =>
+                products.map(product =>
+                    translateDeep(product, ctx.languageCode, ['facetValues', ['facetValues', 'facet']]),
+                ),
+            )
+            .then(products => products[0]);
     }
 
     async create(ctx: RequestContext, input: CreateProductInput): Promise<Translated<Product>> {
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -140,20 +140,35 @@ export class ProductService {
         FindOptionsUtils.applyFindManyOptionsOrConditionsToQueryBuilder(qb, { relations: this.relations });
         // tslint:disable-next-line:no-non-null-assertion
         FindOptionsUtils.joinEagerRelations(qb, qb.alias, qb.expressionMap.mainAlias!.metadata);
+        const translationQb = this.connection
+            .getRepository(ctx, ProductTranslation)
+            .createQueryBuilder('product_translation')
+            .select('product_translation.baseId')
+            .andWhere('product_translation.slug = :slug', { slug });
+
         return qb
-            .innerJoin(ProductTranslation, 'product_translation', 'product_translation.baseId=product.id')
             .leftJoin('product.channels', 'channel')
-            .andWhere('product_translation.slug = :slug', { slug })
-            .andWhere('product_translation.languageCode = :languageCode', { languageCode: ctx.languageCode })
+            .andWhere('product.id IN (' + translationQb.getQuery() + ')')
+            .setParameters(translationQb.getParameters())
             .andWhere('product.deletedAt IS NULL')
             .andWhere('channel.id = :channelId', { channelId: ctx.channelId })
-            .getMany()
-            .then(products =>
-                products.map(product =>
-                    translateDeep(product, ctx.languageCode, ['facetValues', ['facetValues', 'facet']]),
-                ),
+            .addSelect(
+                "CASE product_translations.languageCode WHEN '" +
+                    ctx.languageCode +
+                    "' THEN 2 WHEN '" +
+                    ctx.channel.defaultLanguageCode +
+                    "' THEN 1 ELSE 0 END",
+                'sort_order',
             )
-            .then(products => products[0]);
+            .orderBy('sort_order', 'DESC')
+            .limit(1)
+            .getMany()
+            .then(products => products[0])
+            .then(product =>
+                product
+                    ? translateDeep(product, ctx.languageCode, ['facetValues', ['facetValues', 'facet']])
+                    : undefined,
+            );
     }
 
     async create(ctx: RequestContext, input: CreateProductInput): Promise<Translated<Product>> {
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -153,11 +153,11 @@ export class ProductService {
             .andWhere('product.deletedAt IS NULL')
             .andWhere('channel.id = :channelId', { channelId: ctx.channelId })
             .addSelect(
-                "CASE product_translations.languageCode WHEN '" +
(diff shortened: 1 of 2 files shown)


packages/core/src/service/services/product.service.ts
```diff
@@ -136,18 +136,24 @@ export class ProductService {
...
-        return this.findOne(ctx, bestMatch.base.id);
+        const qb = this.connection.getRepository(ctx, Product).createQueryBuilder('product');
+        FindOptionsUtils.applyFindManyOptionsOrConditionsToQueryBuilder(qb, { relations: this.relations });
+        // tslint:disable-next-line:no-non-null-assertion
+        FindOptionsUtils.joinEagerRelations(qb, qb.alias, qb.expressionMap.mainAlias!.metadata);
+        return qb
+            .innerJoin(ProductTranslation, 'product_translation', 'product_translation.baseId=product.id')
+            .leftJoin('product.channels', 'channel')
+            .andWhere('product_translation.slug = :slug', { slug })
+            .andWhere('product_translation.languageCode = :languageCode', { languageCode: ctx.languageCode })
```
michaelbromley: This implementation misses a feature: it only looks in the current language and does not perform a fall-back to other languages, as is the case in the current version. That is why the test 
```
Product resolver › product query › falls back to default language slug
```
is failing.

The proper behaviour is:
1. Attempt to match the slug from a ProductTranslation in the current language (`ctx.languageCode`)
2. Else fall back to a match in the `ctx.channel.defaultLanguageCode`
3. Else fall back to any match from any language.

I'm not sure how this logic can be stated as SQL (though I suspect it might be possible with very advanced SQL skills). If not possible, then the fallback must be performed in application code (as it used to be).

michaelbromley (changes requested):
See comment.

Note: There are other e2e tests failing from the Elasticsearch plugin tests - this is unrelated to your changes and is the result of an error I introduced yesterday. If you rebase onto the latest master branch, you should only see failures if there is a problem with your code.

packages/core/src/service/services/product.service.ts
Izayda: Decided to implement slug tests for different translation cases. What result we except in each variant of test?

```javascript
      describe('product query with translations', () => {
            it('by slug without translation arg', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.en;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: translation.slug },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });

            it('by slug de without translation arg', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.de;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: translation.slug },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });

            it('by slug with translation en', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.en;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: translation.slug },
                    { languageCode: LanguageCode.en },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });

            it('by slug de with translation en', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.de;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: 'de-baked-potato' },
                    { languageCode: LanguageCode.en },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });

            it('by slug with translation de', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.en;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: 'en Baked Potato' },
                    { languageCode: LanguageCode.de },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });

            it('by slug de with translation de', async () => {
                const translation = newTranslatedProduct.translations.filter(t => {
                    return t.languageCode === LanguageCode.de;
                })[0];
                const { product } = await adminClient.query<GetProductSimple.Query, GetProductSimple.Variables>(
                    GET_PRODUCT_SIMPLE,
                    { slug: 'de-baked-potato' },
                    { languageCode: LanguageCode.de },
                );

                if (!product) {
                    fail('Product not found');
                    return;
                }
                expect(product.slug).toBe(translation.slug);
            });
        });
```
        
Test results for stable old logic:
```javascript
Error: expect(received).toBe(expected) // Object.is equality

Expected: "de-baked-potato"
Received: "en-baked-potato"
<Click to see difference>


    at Object.<anonymous> (/Users/artem.danilov/WebstormProjects/vendure/packages/core/e2e/product.e2e-spec.ts:496:38)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

Error: expect(received).toBe(expected) // Object.is equality

Expected: "de-baked-potato"
Received: "en-baked-potato"
<Click to see difference>


    at Object.<anonymous> (/Users/artem.danilov/WebstormProjects/vendure/packages/core/e2e/product.e2e-spec.ts:530:38)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

Failed: "Product not found"
Error: Failed: "Product not found"
    at Env.fail (/Users/artem.danilov/WebstormProjects/vendure/node_modules/jest-jasmine2/build/jasmine/Env.js:722:61)
    at fail (/Users/artem.danilov/WebstormProjects/vendure/node_modules/jest-jasmine2/build/jasmine/jasmineLight.js:156:23)
    at Object.<anonymous> (/Users/artem.danilov/WebstormProjects/vendure/packages/core/e2e/product.e2e-spec.ts:544:21)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

image

michaelbromley: That's a good idea to make more thorough tests. So the basic idea is that the query should always return results in the current active language. In these tests this will be en by default, unless you specify otherwise with the query string `languageCode: LanguageCode.de`. This is why those 3 tests are failing - even though you are querying with a de slug, the results will still be in en.

Izayda: Ok.

When querying "by slug de without translation arg" or 'by slug de with translation en' we receive now "en-baked-potato"? Is that correct or we should get None? Now it works because of fourth row in
```javascript
 const bestMatch =
            translations.find(t => t.languageCode === ctx.languageCode) ??
            translations.find(t => t.languageCode === ctx.channel.defaultLanguageCode) ??
            translations[0];
```
 

When querying "by slug en with translation de" we receive now None. Is that right or we should receive 'de-baked-potato'?

michaelbromley: > When querying "by slug de without translation arg" or 'by slug de with translation en' we receive now "en-baked-potato"?

yes this is correct behaviour. We should always return results in the active language - en in this case as this is the channel's default language.

> When querying "by slug en with translation de" we receive now None

Ideally I think we should be able to return the de version here, but looking at the logic I see why this does not currently work. If you can make it work that way, this would be good. If not, it is acceptable to just keep the existing behaviour (returning none).

Izayda: Ok, will implement this logic
```
en slug | en translation | default translation => en slug
en slug | de translation => de slug
de slug | en translation | default translation => en slug
de slug | de translation | translation exists => de slug
de slug | de translation | translation not exists => en slug
```

michaelbromley (approved):

michaelbromley:
💯  _N I C E_  💯 work on the SQL query! Awesome. This is why I love open source.

Izayda:
> 💯 _N I C E_ 💯 work on the SQL query! Awesome. This is why I love open source.

Thank you! I don't know best practices of string formating in JS, so, maybe, some parts of this code maybe more elegant. 
And this two lines 
```       
.getMany()
.then(products => products[0])
```
maybe can be replaced with one. But i couldn't implement this :(
```
.getOne()
```

michaelbromley:
A nicer way to do string interpolation is to use the backticks: `` `interpolate a ${variable}` ``

I added this as well as `getOne()` here: https://github.com/vendure-ecommerce/vendure/commit/ec59bda889814a80d6215bd5a091a866f1de7f70

Izayda:
Thanks! Will use this in next PRs)

PR #806, opened 2021-04-01
feat(core): Add name and description fields to eligiblePaymentMethods query

### LINKED ISSUE #738: eligiblePaymentMethods should return name and description
As requested https://github.com/vendure-ecommerce/vendure/issues/671#issuecomment-788036992

eligiblePaymentMethods query does not return `name` and `description` at the moment. 
`code` can probably also be removed and `id` could be used for relationships instead.

stefanvanherwijnen:
Closes #738 .

`yarn codegen` failed with an error in auth.e2e-spec.ts (not related to this PR) which I couldn't figure out, so I skipped the generation of admin types and thus something might be missing there. But it seems to build and run fine.


michaelbromley:
Thanks! Yeah the codegen issue was an oversight on my part.

Issue #726, opened 2021-02-23
Multiple groups of filters for search products query


Izayda:
**Is your feature request related to a problem? Please describe.**
Allow user to filter by multiple filters. For example, user want to filter some category by color (red (facet_id=1) or white (facet_id=2)) and by material (leather (facet_id=10)).

Now search query allow to input list of facets and logical operator for list https://www.vendure.io/docs/graphql-api/shop/input-types/#searchinput

**Describe the solution you'd like**
Allowing to pass groups of facets to search query.
For example,: `
{
   facets: [
                    {
                       facets_ids: [1,2],
                       logical_operator: "OR"
                    },
                    {
                      facets_ids:[10]
                    }
                ],
  logical_operator: "AND"
}`

**Describe alternatives you've considered**
So, the only way to filter by multiple filters is querying shop-api two times.
- first time with facet_ids = [1,2] and logical operator= Or
- second time with facet_ids=[10]
- merge results at client side

**Additional context**
Probably it can be implemented by ElasticSearch plugin?

chladog:
We discussed this on Slack. Here is my piece of code that can be used to can extend ListQueryBuilder object (Product or ProductVariant) that does similiar to what you ask for, though without customizable logical operator - it will always look between facetValues of same facet (`OR between facetValues of x facet` AND `OR between facetValues of y facet` AND so on...)

`qb.andWhere(`
            "product"."id" IN (
                WITH targets AS (SELECT * FROM facet_value WHERE id IN (:...argProductFacetValues))
                SELECT "productId"
                FROM product_facet_values_facet_value AS map
                INNER JOIN targets AS tgt
                ON tgt.id = map."facetValueId"
                GROUP BY map."productId"
                HAVING COUNT(DISTINCT tgt."facetId") = (SELECT COUNT(DISTINCT "facetId") FROM targets)
            )
            `, { argProductFacetValues: options.facetValueIds });
`

dccarta:
+1 for this. It does _seem_ to make sense to be able to use the OR operator within what I've been calling 'facet value groups', but the AND operator between said groups (although I _would_ say it makes sense, because it's what I'm trying to do with my site 😜 ).

To try and give a real-world example, using place of origin of coffee beans (number in brackets shows products found by search):

FILTERS - NOWT SELECTED
**Region**
- Africa (1)
- South America (3)

**Country**
- Ethiopia (1)
- Colombia (1)
- Brazil (2)

[...]

PR #808, opened 2021-04-06
feat(elasticsearch-plugin): Add facetFilters input for search query

Izayda:
Related #726

This PR hasn't e2e test (only unit tests), because i don't know how to safely extend generated-e2e-shop-types.ts with new field.


michaelbromley:
The implementation looks good, but I'm not sure about the API - I feel it might be confusing that the SearchInput now has both `facetValueIds` and `facetFilters` fields, which essentially do the same thing (filter by facet value).

I understand why you did it this way, because we need the existing SearchInput API to still be compatible with the DefaultSearchPlugin API. I have been trying to think of ways to use the existing API in a backwards-compatible way, e.g.

```graphql
extend input SearchInput {
  facetValueIds: [ID!] | [[ID]]
}
```
but this is not valid in GraphQL unfortunately.

I guess in the docs we should highlight the improved facet-filtering capabilities of the ES plugin and advise devs to favour the use of the `facetFilters` input field, and explain that the `facetValueIds` and `facetValueOperator` fields are mainly there for backwards-compatibility with the DefaultSearchPlugin.

Izayda:
Yes, totally agree with you. Probably, i can extend DefaultSearchPlugin with new input and we can mark old input as deprecated. Seems, it would be more consistent and moreover, allow to write e2e tests.

michaelbromley:
If you have the time to do that I would gladly accept it! I would also suggest renaming to `facetValueFilters` to make it clear we are filtering by FacetValue and not Facet IDs.

Izayda:
Ok, i will do that in couple of days and update PR. 

One moment. How to mark field as deprecated?

Izayda:
Or is there reason to divide implementation on separate PRs?

michaelbromley:
You can keep it in the same PR. To mark as deprecated use the `@deprecated(reason: " ... ")` decorator in the GraphQL SDL

Izayda:
Ok, probably done. Please review :) 


Probably this might be added to docs somewhere:
* # FacetFilters input
 * Moreover, the [SearchInput](/docs/graphql-api/admin/object-types/#searchinput) type is extended with
 * FacetFilters parameter to put multiple conditions of facet filtering.
 *
 * ```SDL
 * extend input SearchInput {
 *    facetFilters: [FacetFilterInput!]
 * }
 * input FacetFilterInput {
 *     facetId: ID
 *     facetIds: [ID!]
 * }
 * ```
 *
 * # Simple OR filter
 * This example translate as facetID=1 OR facet_id=2
 * ```JSON
 * {facetFilters: [{facetIds:[1,2]}]}
 * ```
 * # Simple AND filter
 * This example translate as facetID=1 AND facet_id=2
 * ```JSON
 * {facetFilters: [{facetId:1]},{facetId:2}]}
 * ```
 * # AND and OR filter combination
 * This example translate as facetID=1 AND (facet_id=2 OR facet_id=3)
 * ```JSON
 * {facetFilters: [{facetId:1]},{facetIds:[2,3]}]}
 * ```

michaelbromley:
This is great. I have 1 last bit of feedback:

When I was manually testing out the filters, I found it difficult to remember whether I should be using `facetValueIds` or `facetValueId` in order to get the desired boolean filter expression.

Consider the current naming:
``` 
This example translate as facetID=1 OR facet_id=2
{ facetValueFilters: [{ facetValueIds:[1,2] }]}

This example translate as facetID=1 AND facet_id=2
{ facetValueFilters: [{ facetValueId:1 }, { facetValueId:2 }]}

This example translate as facetID=1 AND (facet_id=2 OR facet_id=3)
{ facetValueFilters: [{ facetValueId:1 }, { facetValueIds:[2,3] }]}
 ```

What do you think of a more explicit naming, such as replacing `facetValueId` with `and`, and `facetValueIds` with `or`:

``` 
This example translate as facetID=1 OR facet_id=2
{ facetValueFilters: [{ or:[1,2] }]}

This example translate as facetID=1 AND facet_id=2
{ facetValueFilters: [{ and:1 }, { and:2 }]}

This example translate as facetID=1 AND (facet_id=2 OR facet_id=3)
{ facetValueFilters: [{ and:1 }, { or:[2,3] }]}
 ```

Izayda:
Great idea! Need 20 minutes to make this rename

michaelbromley:
Did you generate the types with the ES plugin configured?

Izayda:
Yes, ma fault :(

michaelbromley:
OK. Push again with corrected codegen and it's ready to merge 👍

michaelbromley:
Thank you!

PR #814, opened 2021-04-08
perf(core): reduce memory usage of apply-collection-filters job

hendrik-advantitge:
When recalculating all Collections, this query would consume a lot of memory. Instead of querying all Collections before the loop, the Collections are queried individually inside the loop.


michaelbromley:
Thanks!

Issue #810, opened 2021-04-06
Add translation files with plugin config


Draykee:
**Is your feature request related to a problem? Please describe.**
When I create a new error message in my plugin, I want them to be translated like the other vendure errors.

This might not be necessary for the frontend but it might be easier for the frontend developer to always keep the same error format.

**Describe the solution you'd like**
A possiblity to add message files ( like the ones in [core/src/i18n/message](https://github.com/vendure-ecommerce/vendure/tree/master/packages/core/src/i18n/messages) ) via PluginConfig.
All messages should be merged together on bootstrap.

**Additional context**
Translations can already be merged in the admin ui - so this might be reusable.

Draykee:
@michaelbromley Are there special guidelines I should try to follow?
Like should I add a new config option to `VendurePluginMetadata` to import i18n files? Maybe it's also a thing to add it during runtime in the plugin constructor via the `I18nService`

I think the last idea would be the easiest one. I could add a `concat` function that uses `i18next.addResourceBundle`

michaelbromley:
Yes, I prefer the second solution, because I want to keep the API of PluginMetadata as limited as possible.

So the idea would be that in your plugin you inject `I18nService` and then call some new method, e.g. `addTranslations()` which could take a path to a json file, or an array of such.

Draykee:
Just as a reminder: `i18next-node-fs-backend` seems to be deprecated.

michaelbromley:
oh, I didn't know that! Looks like it can be replaced with https://github.com/i18next/i18next-fs-backend, they say 

> It's based on the deprecated i18next-node-fs-backend and can mostly be used as a drop-in replacement.

I wonder what "mostly" means here. Are you interested in attempting to try?

Draykee:
Yeah I can try that. But right now im struggeling finding the correct namespace where to place the new resource bundle at.

Draykee:
I have to import the `I18nService` from a wierd place:
`import { I18nService } from '../../../dist/i18n/i18n.service';`

Isn't it exported in `'@vendure/core'`?

This is the error I get when I try to import from Vendure core in my e2e test plugin
```
@vendure/core:       111 |         try {
@vendure/core:       112 |             DefaultLogger.hideNestBoostrapLogs();
@vendure/core:     > 113 |             const app = await NestFactory.create(appModule.AppModule, {
@vendure/core:           |                         ^
@vendure/core:       114 |                 cors: config.apiOptions.cors,
[...]

PR #819, opened 2021-04-12
Feature: Add translations via plugin application bootstrap

Draykee:
Pull Request for #810 

**Change List:**
- Extended `I18nService` with `addTranslationFile` and `addTranslation`
- Created `translations.e2e-spec.ts` with TranslationTestPlugin
- Changed deprecated[ i18next-node-fs-backend](https://github.com/i18next/i18next-node-fs-backend) to [i18next-fs-backend](https://github.com/i18next/i18next-fs-backend)
- Exports `I18nService` and `I18nError` in public api
- Fixed language detection of i18next module (changed to `lookupQuerystring: 'languageCode'`)
- Added documentation to developer guide and I18nService typescript documentation (used for linking)


Draykee:
@michaelbromley I think this PR should be done and reviewable for you (or anyone else) 🥳

packages/core/src/i18n/i18n.service.ts
```diff
@@ -1,17 +1,29 @@
...
 import path from 'path';
 
 import { GraphQLErrorResult } from '../common/error/error-result';
+import { Logger } from '../config';
 import { ConfigService } from '../config/config.service';
 
 import { I18nError } from './i18n-error';
 
+/**
+ * I18n resources used for translations
```
michaelbromley, line 19: Needs a `@description` tag to be picked up by docs generator.

packages/core/src/i18n/i18n.service.ts
```diff
@@ -21,11 +33,19 @@ export interface I18nRequest extends Request {
  * The `i18next-express-middleware` middleware detects the client's preferred language based on
```
michaelbromley, line 35: Same on the `@description` tag

packages/core/src/i18n/i18n.service.ts
```diff
@@ -44,12 +64,40 @@ export class I18nService implements OnModuleInit {
...
+            const rawData = fs.readFileSync(filePath);
+            const resources = JSON.parse(rawData.toString('utf-8'));
+            this.addTranslation(langKey, resources);
+        } catch (err) {
+            Logger.error(`Could not load resources file ${filePath}`, `I18nService`);
+        }
+    }
+
+    /**
+     * Add a I18n translation file
```
michaelbromley: not a "file" in this case right?

michaelbromley (commented):
Absolutely excellent PR - docs, e2e tests, implementation, 😗 👌 

Just missing some `@description` tags on the JSDoc blocks so that they show up in the docs.

packages/core/src/i18n/i18n.service.ts
kevmtt: Yep. Good old copy+paste error ^^

michaelbromley:
Thank you!

PR #821, opened 2021-04-13
fix(email-plugin): Added filename filter for json files

Draykee:
Will fix: https://github.com/vendure-ecommerce/real-world-vendure/issues/14

Only `.json` files shall be taken into account.


michaelbromley:
Thanks!

PR #842, opened 2021-04-22
feat(core): Improved translations DE; slightly changed EN

Wanztwurst:


packages/core/src/service/services/product-variant.service.ts
```diff
@@ -723,7 +723,7 @@ export class ProductVariantService {
                 const variantOptionIds = this.sortJoin(variant.options, ',', 'id');
                 if (variantOptionIds === inputOptionIds) {
                     throw new UserInputError('error.product-variant-options-combination-already-exists', {
-                        optionNames: this.sortJoin(variant.options, ', ', 'code'),
+                        variantName: variant.name,
```
Wanztwurst: Does variant always have name?

michaelbromley: Yes, but it is not always available. Since the "name" is translatable, it actually exists in a different table (product_translation), and is "hydrated" into the ProductVariant entity using the `translateDeep()` helper function. In this case, I think the name will be `undefined` since I cannot see anywhere that `translateDeep()` is being applied.

Wanztwurst: I've added but not tested it in [another commit](https://github.com/vendure-ecommerce/vendure/pull/842/commits/2ba28fda29a244f4d5178d01baee655cdc30bd66). Is this the way to use translateDeep and access the value?

michaelbromley: From the failing test, it seems that the translateDeep does indeed work! That just means you need to update this e2e test: https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/e2e/product.e2e-spec.ts#L1304

in order to test for the updated error message.

Wanztwurst:
Okay, done! I've learned new things, thank you!

michaelbromley:
Thank you too! 👍

Issue #844, opened 2021-04-25
Asset custom fields are ignored


Draykee:
**Describe the bug**
`AssetService.create(ctx: RequestContext, input: CreateAssetInput)` ignores `input.customFields`

https://github.com/vendure-ecommerce/vendure/blob/e4a795ff7f2743b3615d0ad72187ee233d30e372/packages/core/src/service/services/asset.service.ts#L238

```Typescript
export type CreateAssetInput = {
  file: Scalars['Upload'];
  tags?: Maybe<Array<Scalars['String']>>;
  customFields?: Maybe<Scalars['JSON']>;  //< is ignored
};
```

**To Reproduce**
1. Execute a `createAssets` mutation with `customFields` input
2. Custom fields are ignored

**Expected behavior**
Custom fields are properly parsed and present in the database asset table

PR #845, opened 2021-04-25
fix(core): Fixed missing customField input for assets (#844)

Draykee:
Shall fix (#844)

- Added optional `customFields?: { [key: string]: any }` to `createAssetInternal` signature and passed it in `Asset` constructor. 
- Added `await this.customFieldRelationService.updateRelations(ctx, Asset, input, result);` in `create` function


michaelbromley:
💯 👍

Issue #855, opened 2021-04-30
core: ConfigArgService.validateRequiredFields() fails checking ShippingEligibilityChecker with required arg


Wanztwurst:
It seems saving a shipping method with an assigned ShippingEligibilityChecker which has at least one arg with `required:true` and containing a type which can't be parsed by JSON.parse() always fails with the `configurable-argument-is-required` error.

I think the problem is the `validateRequiredFields` method, which checks the value with:
```typescript
let val: unknown;
if (inputArg) {
    try {
        val = JSON.parse(inputArg?.value);
    } catch (e) {
        // ignore
    }
}
if (val == null) {
    throw new UserInputError('error.configurable-argument-is-required', {
        name,
    });
}
```
As `inputArg.value` is a string in my case,  `JSON.parse(inputArg?.value)` throws an error (which is ignored), `val == null` evaluates to true and the validation fails.

Possible types seem to be from `ConfigArgType`: `'string' | 'int' | 'float' | 'boolean' | 'datetime' | 'ID'`. So my proposition is to change the error-evaluation to something like `!inputArg || inputArg.value === null || input.value === ""` to cover all types (not exactly sure on this, peer review please: are all type covered correctly?).

Code here: [config-arg.service.ts#L91](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/helpers/config-arg/config-arg.service.ts#L91)

Wanztwurst:
I noticed booleans and null get stringified as well – is this on purpose? The tests show the same behaviour. The solution to string verifying then is to put them in double quotation marks I think.

michaelbromley:
Thanks for the report. The reason that all datatypes get represented as strings is because the GraphQL input type `ConfigArgInput` expects a string value. With GraphQL, allowing for multiple data types in inputs is not really possible in a nice way - the technical reason is that input union types are not allowed - so this is a work around.

I think the e2e test is actually incorrect, and the server should be handling strings which are not JSON stringified. In the Admin UI, when values are submitted from such configurable operations, we apply `(value ?? '').toString()` to get the string value we will send to the GraphQL API.

Wanztwurst:
So either the posted values of type string should be put in two pairs of quotes or the different types should be validated differently.

michaelbromley:
[...]

PR #861, opened 2021-05-04
fix(core): Handle different input types in validateRequiredFields()

Wanztwurst:
Implements non-empty checks for different input types of required config-args. Solves #855 

What's not in this commit: Checks for validity of input for actual configured type.


michaelbromley:
Thank you for your help on this 👍

Issue #874, opened 2021-05-10
Refund metadata not displayed


jdG-:
**Describe the bug**
Metadata of refund payment are not displayed
![image](https://user-images.githubusercontent.com/6771601/117729452-25877f80-b1eb-11eb-90c2-1f8f0af2e372.png)

**To Reproduce**
1. In your paymentHandler => createRefund, return a CreateRefundResult with some metadata
2. On an order, create a refund
3. On the order details, the metadata are not displayed

**Expected behavior**
Display the refund metadata

**Environment (please complete the following information):**
 - @vendure/core version: 1.0.0-beta.8
 - Nodejs version: 14.4.0
 - Database (mysql/postgres etc): postgres

PR #875, opened 2021-05-10
fix(admin-ui): Display refund metadata

jdG-:
Display refund's metadata on admin-ui order details. Solves #874


packages/admin-ui/src/lib/order/src/components/order-payment-card/order-payment-card.component.ts
```diff
@@ -16,7 +16,7 @@ export class OrderPaymentCardComponent {
     @Output() settleRefund = new EventEmitter<OrderDetail.Refunds>();
 
     refundHasMetadata(refund?: OrderDetail.Refunds): boolean {
-        return !!refund && Object.keys(refund).length < 0;
```
michaelbromley, line 19: Oh. mustn't have had my coffee yet when I wrote that masterpiece 😆

PR #878, opened 2021-05-13
fix(admin-ui): Update Chinese translate file

kai-chu:
It's been a while since I created the Chinese translate, and lots of new features have been added. I played around the new UI and updated the translation now.


michaelbromley:
Thank you, this is very helpful!

PR #915, opened 2021-06-02
perf(core): Improve performance of apply-collection-filters job

dkostenko-jsninja:
Changes in this pull request will improve the performance of apply-collection-filters job.

I noticed that we use only variants ids before we reassign `collection.productVariants` in `CollectionService.applyCollectionFiltersInternal`.

Based on experiments the performance of the query in `CollectionService.getCollectionProductVariantIds` method will be 8x faster.


michaelbromley:
Wow, this is great! Please could you PR against the `master` branch though, since this is not a feature so it can go in the next patch release :)

dkostenko-jsninja:
Hello! I've just changed the base branch.

Issue #922, opened 2021-06-09
Feature: Guest orders for more than 2h through orderByCode query


Wanztwurst:
The default for guest orders is a 2h time limit to retrieve it via the orderByCode query. I want my guest customers to be able to view their orders, for example to see the current status, after that. 

So the easiest solution is this: an option to disable the 2h check and make permanent access possible (pull request #924 opened).

Alternatively, if you don't like this solution, the time limit could be a configurable option.

PR #923, opened 2021-06-09
feat(core): Option to enable permanent anonymous access to orderByCode

Wanztwurst:
Implements #922 

This commit adds an option to disable the default 2h time window for guests to retrieve their orders through the `orderByCode` query.

Security considerations: The DefaultOrderCodeStrategy [generates the Code](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/common/generate-public-id.ts#L19) from 33 different characters and is 16 characters long – which makes 33^16 different possible codes (~2*10^24). Collisions are highly unlikely, especially if a request limiter is being used.

This commit is non-breaking. Additionally one db query is saved.


Wanztwurst:
I made a mistake – new pull request coming soo,.

PR #924, opened 2021-06-09
feat(core): Option to enable permanent anonymous access to orderByCode

Wanztwurst:
Implements #922 

This commit adds an option to disable the default 2h time window for guests to retrieve their orders through the `orderByCode` query.

Security considerations: The DefaultOrderCodeStrategy [generates the Code](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/common/generate-public-id.ts#L19) from 33 different characters and is 16 characters long – which makes 33^16 different possible codes (~2*10^24). Collisions are highly unlikely, especially if a request limiter is being used.

This commit is non-breaking. Additionally one db query is saved.

### DIFF
--- a/packages/core/src/config/order/order-by-code-access-strategy.ts
+++ b/packages/core/src/config/order/order-by-code-access-strategy.ts
@@ -6,13 +6,13 @@ import { Order } from '../../entity/order/order.entity';
 
 /**
  * @description
- * The OrderByCodeAccessStrategy determins how access to a placed Order via the
+ * The OrderByCodeAccessStrategy determines how access to a placed Order via the
  * orderByCode query is granted.
  * With a custom strategy anonymous access could be made permanent or tied to specific
- * conditions like IP range or a specific Order status.
+ * conditions like IP range or an Order status.
  *
  * @example
- * This example grants access to the requested Order to everyone – unless it's Monday.
+ * This example grants access to the requested Order to anyone – unless it's Monday.
  * ```TypeScript
  * export class NotMondayOrderByCodeAccessStrategy implements OrderByCodeAccessStrategy {
  *     canAccessOrder(ctx: RequestContext, order: Order): boolean {
@@ -38,25 +38,23 @@ export interface OrderByCodeAccessStrategy extends InjectableStrategy {
 /**
  * @description
  * The default OrderByCodeAccessStrategy used by Vendure. It permitts permanent access to
- * the Customer owning the Order and anyone within 2 hours after placing the Order.
+ * the Customer owning the Order and anyone within a given time period after placing the Order
+ * (defaults to 2h).
+ *
+ * @param anonymousAccessDuration value for [ms](https://github.com/vercel/ms), e.g. `2h` for 2 hours or `5d` for 5 days
  *
  * @docsCategory orders
  * @docsPage OrderByCodeAccessStrategy
  */
 export class DefaultOrderByCodeAccessStrategy implements OrderByCodeAccessStrategy {
-    canAccessOrder(ctx: RequestContext, order: Order): boolean {
+    canAccessOrder(ctx: RequestContext, order: Order, anonymousAccessDuration: string = '2h'): boolean {
         // Order owned by active user
-        const activeUserMatches = !!(
-            order &&
-            order.customer &&
-            order.customer.user &&
-            order.customer.user.id === ctx.activeUserId
-        );
+        const activeUserMatches = order?.customer?.user?.id === ctx.activeUserId;
 
         // For guest Customers, allow access to the Order for the following
         // time period
         const anonymousAccessPermitted = () => {
-            const anonymousAccessLimit = ms('2h');
+            const anonymousAccessLimit = ms(anonymousAccessDuration);
             const orderPlaced = order.orderPlacedAt ? +order.orderPlacedAt : 0;
             const now = Date.now();
             return now - orderPlaced < anonymousAccessLimit;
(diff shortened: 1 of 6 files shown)


michaelbromley:
Hi, thanks for the PR. I like the idea and we should support making this configurable, but I'm not sure about the API as given here.

- it does not allow a different access duration (as you mention in the issue)
- there might conceivably be other logic one might want to execute to check whether anonymous access to order details should be allowed, e.g. based on the details of the order or the order customer.

Therefore I think it might be better to make this a function rather than a primitive. We could have an `OrderAccessStrategy`, much in the same way we allow the other options to be defined. It might look something like this:

```ts
interface OrderAccessStrategy extends InjectableStrategy {
  canAccessOrderByCode(ctx: RequestContext, order: Order): boolean | Promise<boolean>
}
```

I find this approach gives maximum flexibility and allows basically _any_ custom requirements to be met. 

Happy to hear your thoughts on this.

Wanztwurst:
The simple switch for permanent access was sufficient for my use case, I only mentioned the configurable time as an alternative way to offer. Of course more freedom is better for all developers, hence an OrderAccessStrategy analogue to the other strategies would be even better! I am going to implement it like you suggested, but I think on Monday at the earliest.

Wanztwurst:
Servus! I've implemented the strategy to regulate access to orderByCode, but to highlight it's purpose it's called different than you proposed: `OrderByCodeAccessStrategy`. For the docs and everything I used OrderCodeStrategy as a template to match the projects standards.

packages/core/src/config/order/order-by-code-access-strategy.ts
```diff
@@ -0,0 +1,67 @@
+import ms from 'ms';
+
+import { RequestContext } from '../../api/common/request-context';
+import { InjectableStrategy } from '../../common/types/injectable-strategy';
+import { Order } from '../../entity/order/order.entity';
+
+/**
+ * @description
+ * The OrderByCodeAccessStrategy determins how access to a placed Order via the
```
michaelbromley: typo "determins" -> "determines"

packages/core/src/config/order/order-by-code-access-strategy.ts
```diff
@@ -0,0 +1,67 @@
...
+ * orderByCode query is granted.
+ * With a custom strategy anonymous access could be made permanent or tied to specific
+ * conditions like IP range or a specific Order status.
+ *
+ * @example
+ * This example grants access to the requested Order to everyone – unless it's Monday.
+ * ```TypeScript
+ * export class NotMondayOrderByCodeAccessStrategy implements OrderByCodeAccessStrategy {
```
michaelbromley, line 17: hah I like the example 😆

packages/core/src/config/order/order-by-code-access-strategy.ts
```diff
@@ -0,0 +1,67 @@
...
+            order &&
+            order.customer &&
+            order.customer.user &&
+            order.customer.user.id === ctx.activeUserId
+        );
+
+        // For guest Customers, allow access to the Order for the following
+        // time period
+        const anonymousAccessPermitted = () => {
+            const anonymousAccessLimit = ms('2h');
```
michaelbromley: Let's make the 2h limit configurable by accepting a constructor argument which defaults to `'2h'`. That means that if you want to e.g. have a 5 hour window, you just need to instantiate with `'5h'` rather than re-implementing the whole strategy.

packages/core/src/config/order/order-by-code-access-strategy.ts
```diff
@@ -0,0 +1,67 @@
...
+ * The default OrderByCodeAccessStrategy used by Vendure. It permitts permanent access to
+ * the Customer owning the Order and anyone within 2 hours after placing the Order.
+ *
+ * @docsCategory orders
+ * @docsPage OrderByCodeAccessStrategy
+ */
+export class DefaultOrderByCodeAccessStrategy implements OrderByCodeAccessStrategy {
+    canAccessOrder(ctx: RequestContext, order: Order): boolean {
+        // Order owned by active user
+        const activeUserMatches = !!(
```
michaelbromley: The original code was written in the days before optional chaining - if you like you could clean this up a bit with `order?.customer?.user?.id === ctx.activeUserId`

packages/core/src/config/order/order-by-code-access-strategy.ts
Wanztwurst: Very good idea!

packages/core/src/config/order/order-by-code-access-strategy.ts
```diff
@@ -0,0 +1,65 @@
...
+ * the Customer owning the Order and anyone within a given time period after placing the Order
+ * (defaults to 2h).
+ *
+ * @param anonymousAccessDuration value for [ms](https://github.com/vercel/ms), e.g. `2h` for 2 hours or `5d` for 5 days
+ *
+ * @docsCategory orders
+ * @docsPage OrderByCodeAccessStrategy
+ */
+export class DefaultOrderByCodeAccessStrategy implements OrderByCodeAccessStrategy {
+    canAccessOrder(ctx: RequestContext, order: Order, anonymousAccessDuration: string = '2h'): boolean {
```
michaelbromley: The `anonymousAccessDuration` argument should be passed to the strategy constructor, not to the `canAccessOrder` method. In this way, the developer will be able to use it like:

```ts
orderByCodeAccessStrategy: new DefaultOrderByCodeAccessStrategy('5h'),
```

michaelbromley (changes requested):
Just one point to fix about the default strategy

packages/core/src/config/order/order-by-code-access-strategy.ts
Wanztwurst: Oopsie, fixed. That's a bit embarrassing.

michaelbromley:
👍💯

Issue #934, opened 2021-06-15
Api to configure NestApplication options - bodyParser


gkielwasser:
When Vendure bootstrap, it creates a NestApplication instance using private options. 
Specifically one of them is whether we want to use the body parser (bodyParser:boolean).

The use of bodyParser need to be configured before the application listen.

We should be able to configure in the vendure config NestApplicationOptions that will be used to create the instance of the NestApplicaiton during bootstrap.

No alternative found so far to disable the global bodyParser.

Below is a concrete exemple I have tried to let a plugin have access to the body raw.

Vendure plugin
```
export class PaymentPlugin implements NestModule {
    configure(consumer: MiddlewareConsumer): any {
        consumer
        .apply(RawBodyMiddleware)
        .forRoutes({
           path: '/payment/webhook',
           method: RequestMethod.POST,
         })
         .apply(JsonBodyMiddleware)
         .forRoutes('*');
    }
}
```
in the RawBodyMiddleware
```@Injectable()
export class RawBodyMiddleware implements NestMiddleware {
    use(req: Request, res: Response, next: () => any) {
        bodyParser.raw({type: '*/*'})(req, res, next);
    }
}
```

then in a webhook endpoint, the body value is read
```
@Controller('payment')
export class PaymentController {
 @Post('webhook/')
  async index(@Body() raw: Buffer) {
     // read the raw body
   }
}
```

Unless NestApplication is started with bodyParser to false, the body will always be parsed
```
const app = await core.NestFactory.create(appModule.AppModule, {
    cors,
    logger: new vendure_logger.Logger(),
    bodyParser: false
});
```

PR #935, opened 2021-06-16
feat(core): Allow to configure body parser

gkielwasser:
Implements [#934 ](https://github.com/vendure-ecommerce/vendure/issues/934)

We currently have no way to control the body parser applied to the global middleware.

This commit would let us configure it the same way we already configure  copple of the NestApplicationOptions using Vendure config.

### DIFF
--- a/packages/core/src/app.module.ts
+++ b/packages/core/src/app.module.ts
@@ -1,7 +1,7 @@
 import { MiddlewareConsumer, Module, NestModule, OnApplicationShutdown } from '@nestjs/common';
-import { Type } from '@vendure/common/lib/shared-types';
 
 import { ApiModule } from './api/api.module';
+import { Middleware, MiddlewareHandler } from './common';
 import { ConfigModule } from './config/config.module';
 import { ConfigService } from './config/config.service';
 import { Logger } from './config/logger/vendure-logger';
@@ -11,9 +11,6 @@ import { I18nService } from './i18n/i18n.service';
 import { PluginModule } from './plugin/plugin.module';
 import { ProcessContextModule } from './process-context/process-context.module';
 import { ServiceModule } from './service/service.module';
-
-// tslint:disable-next-line:ban-types
-type Middleware = Type<any> | Function;
 
 @Module({
     imports: [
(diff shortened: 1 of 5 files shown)


michaelbromley:
Thanks for the contribution. Before merging this I'd like to explore an alternative API design:

* Use the existing `apiOptions.middleware` array
* Extend the middleware object by adding a new flag, e.g. "global" (or maybe some better name) which indicates it will be applied before the call to `app.listen()`.
* During `bootstrap()`, we look for any such middlewares and apply them directly with `app.use()`.

My reasoning for this:

1. It does not introduce another (very specialized) config option. I like to avoid adding new options to the VendureConfig whenever possible.
2. It composes together better with other configurations & plugins. For example, if we have a Stripe plugin which globally disables the `bodyParser` in NestJS, what effect will that have on all the other parts of the project? I can image it might cause some issues to change this. On the other hand, using the design above I think a Stripe plugin could insert a "global" middleware, get the raw request data, and save it to some specific key on the Request object, which can then be read later by the controller handler.

What are your thoughts about this?

gkielwasser:
1- I would agree not to add one by one specific config. In the other hand, currently, at first sight, there is a strong choice to enforce the instanciation of the nest application with only the `core` option. From what I see from the interface NestApplicationOptions, we are also currently not able to control following options globally: bodyParser, httpsOptions, abortOnError. I have to say, I'm not sure if it is a good thing or not.

2- I agree it is better that a plugin is able to add a non-destructive strategy like saving the raw body on a key instead of being able to globally disable that. So the ability to add from a plugin global middlewares looks nice for me.

So for the specific case of the bodyParser I tried a quick POC that apply your solution just to test it would work with the body parser in my use case. It works well.

what I have quickly tested

1- use a given middleware from the vendure config
```
export const useRaw = json({
    verify(req: RawBodyIncomingMessage, res: http.ServerResponse, buf: Buffer) {
        if (req.headers['stripe-signature'] && Buffer.isBuffer(buf)) {
            req.rawBody = cloneBuffer(buf);
        }
        return true;
    },
})
```
2- use it just before `app.listen` with `app.use()`
3- ensure we are able to read the raw body from the plugin that would push this strategy in Vendure config

michaelbromley:
Thanks for your input and testing out the concept. Would you like to have a go at implementing?

gkielwasser:
Ok I have just commited something, waiting for your feedback

michaelbromley:
Thank you, I'll review a little later today 👍

michaelbromley:
@gkielwasser thanks for your very nice work on this 👍

PR #942, opened 2021-06-21
docs: Correct settlePayment function return type

stefanvanherwijnen:
`success: false` seems to be limited to SettlePaymentErrorResult now.


michaelbromley:
Thank you!

Issue #944, opened 2021-06-23
verifyCustomerAccount should be channel independent


karel-advantitge:
**Describe the bug**
When a customer account is being verified through the `verifyCustomerAccount` mutation and the 'vendure-token' header contains a token of a channel the customer is not yet assigned to (e.g. because the token wasn't included in the register mutation), the verification call fails with error: "Cannot locate a Customer for the user"

**To Reproduce**
1. Perform a `registerCustomerAccount` mutation without a 'vendure-token' header
2. Perform the `verifyCustomerAccount` mutation for the newly registered user, but add a 'vendure-token' header for a specific channel

**Expected behavior**
All registered calls are channel independent through the logic of the AuthGuard. If a call comes in with a 'vendure-token' header for a channel the customer is not yet assigned to, the AuthGuard will automatically assign the customer to the channel. I think this logic should be applied in the verification logic as well

PR #945, opened 2021-06-23
fix(core): Make verifyCustomerAccount channel independent

karel-advantitge:
#944


michaelbromley:
The failing e2e test is flaky. Code looks good, I will review properly later on 👍

Issue #931, opened 2021-06-11
Incorrect Quantity Adjustments


WilliamMilne:
**Describe the bug**
In two cases, we were getting an incorrect quantity update to an order line item: 
1. When low stock was available, adjusting an order line, THEN adding that same item to order was causing us to get past the hard cap on the stock of that item.
2. The result from `constrainQuantityToSaleable` appears to be two distinct values depending on the situation: if no adjustment is needed, it is the correct new quantity of the item, but if the desired quantity is above the stock limit then it will return the differential between the existing stock in cart and the maximum stock allowed. 

**To Reproduce**
1. Create a product with 2 stock. 
2. Add it to order with quantity 2. (addItemToOrder mutation triggered)
3. Adjust the order line by incrementing by 1. (adjustOrderLine mutation triggered)
4. Add the item to order again. (addItemToOrder mutation triggered again)
5. You will see the line item quantity become 3. 

**Expected behavior**
1. We expect the stock of an item in cart to never exceed the available stock that the store has.
2. We expect the user to try to add 1000 stock of an item to cart. If they do, we want their cart to be filled with exactly the remaining inventory.
3. If the user already has an item in cart and tries to add more than the store has, we want to see the maximum available stock value in their cart. 

**Environment (please complete the following information):**
 - @vendure/core version: 1.0.1
 - Nodejs version 12.20.2
 - Database (mysql/postgres etc): postgres

**Additional context**
We seem to have solved the problem but are having trouble with the PR process. The two fixes required are: 
In order.service.ts: 
1. Change line 451 from: 
``` await this.orderModifier.updateOrderLineQuantity(ctx, orderLine, quantity, order);```
to: 
```await this.orderModifier.updateOrderLineQuantity(ctx, orderLine, correctedQuantity, order);```

2. Change line 403 from: 
```await this.orderModifier.updateOrderLineQuantity(ctx, orderLine, correctedQuantity, order);```
to: 
```
if (correctedQuantity < quantity) {
    const newQuantity = (existingOrderLine ? existingOrderLine?.quantity : 0) + correctedQuantity;
    await this.orderModifier.updateOrderLineQuantity(ctx, orderLine, newQuantity, order);
} else {
    await this.orderModifier.updateOrderLineQuantity(ctx, orderLine, correctedQuantity, order);
}
```
        
This second change must occur because we need to treat the correctedQuantity value differently if it was modified. There is probably a longer-term fix here, but this patch will solve the problem in our experience.

[...]

PR #983, opened 2021-07-12
Fix(core) incorrect quantity adjustment

WilliamMilne:
Fixed some edge cases for quantity adjustment. Addresses issue #931.


packages/core/e2e/shop-order.e2e-spec.ts
```diff
@@ -520,17 +520,56 @@ describe('Shop orders', () => {
...
+
+            // But it should still add the item to the order
+            expect(addItemToOrder!.order.lines[1].quantity).toBe(100);
+
+            const { adjustOrderLine } = await shopClient.query<
+                AdjustItemQuantity.Mutation,
+                AdjustItemQuantity.Variables
+            >(ADJUST_ITEM_QUANTITY, {
+                orderLineId: 'T_8',
+                quantity: 99,
```
michaelbromley: I am a bit confused by this part. The test says "adjustOrderLine with quantity > stockOnHand" but here I see that quantity (99) is _less than_ stockOnHand (100). Or am I missing something?

WilliamMilne: You're completely right! I have updated that to now be 101 instead and am pushing that code now.

michaelbromley:
I pulled down your branch and played around with it locally. One thing I noticed is that the new test seems to pass event when I revert your changes to `order.service.ts`. Can you confirm whether you see the same?

If so, could you see if you can create a test which fails before the changes, and passes after them?

quangpdt:
The old service is passed with new test is understandable. The incorrect quantity is only returned when we make another query to get the Active order, which is not included in the test case.

@WilliamMilne: Thanks for the fix, I have been looking for the solution for couple days and tried to fix it myself, but clearly it's not covered enough comparing to yours. 

Personally, I found this issue is a critical one and need to be fix asap. Would be great if @michaelbromley could help us out, the PR owner has not given new update for more than a month.

WilliamMilne:
@quangpdt and @michaelbromley - I was able to get a test to fail on `master` and pass on this branch by adding a `GetActiveOrder` query and checking the quantity. On `master`, the quantity was `101`, but on the branch, it was the correct value of `100`. Unfortunately I've hit a small snag and am encountering some failing tests that appear to be failing on both master and the branch: 
image
image

@michaelbromley do you have any suggestions on how to fix these issues so I can push the latest changes? Thanks!!

WilliamMilne:
@michaelbromley - it appears that the customers[0].email issue was solved by deleting the existing e2e data in the folder. 

I was also using an outdated version of node, which was causing the asset server test to crash. I've pushed the latest change now! Sorry about the multiple comments!

quangpdt:
Thanks for the great fix @WilliamMilne!

michaelbromley:
Thank you @quangpdt for your input and @WilliamMilne for your effort to wrap this up. I will review again and all being well, merge this tomorrow.

michaelbromley:
Looks good, thanks all! Will do a patch release this week including this fix.

PR #998, opened 2021-07-21
feat(admin-ui): add admin-ui Italian translation

artcoded-dev:
Adding Italian translation for Admin UI


michaelbromley:
Grazie mille!

Issue #1018, opened 2021-08-04
ElasticSearch plugin: rework product index to include all variant data 


Izayda:
**Is your feature request related to a problem? Please describe.**
Now Elastic plugin store for product index for many properties only first variant data
```
productVariantId, productVariantName, productVariantAssetId, productVariantPreview, productVariantPreviewFocalPoint, currencyCode
```
This can cause some bugs, like https://github.com/vendure-ecommerce/vendure/issues/1017

**Describe the solution you'd like**
Products should be grouped. It can be reached by two ways:
1. Separate product index with all variants data
2. Merging product and variant index into one

**Describe alternatives you've considered**
Manually maintain logic of grouping like now.

Izayda:
This issue can be closed

PR #1024, opened 2021-08-08
fix(elastic search plugin): search grouped by product works only for first variant sku

### LINKED ISSUE #1017: ElasticSearch plugin: search grouped by product works only for first variant sku
**Describe the bug**
Search grouped by product works only for first sku variant.
This happens cause of only first variant sku value is taken into index while indexing product
https://github.com/vendure-ecommerce/vendure/blob/43ce7224edada7ba53096deb62e2b16e[…]f3112c9/packages/elasticsearch-plugin/src/indexer.controller.ts

**To Reproduce**
1. Go to any product with several variants
2. Copy sku of second variant
3. Try to search by this SKU
4. See nothing

**Expected behavior**
Product can be found by any SKU variant

**Additional context**
DefaultSearchPlugin doesn't have this problem. 
Related issue: https://github.com/vendure-ecommerce/vendure/issues/1018

Izayda:
Close #1017

Fixed for sku and variantName

### DIFF
--- a/packages/elasticsearch-plugin/e2e/e2e-helpers.ts
+++ b/packages/elasticsearch-plugin/e2e/e2e-helpers.ts
@@ -60,6 +60,36 @@ export async function testMatchSearchTerm(client: SimpleGraphQLClient) {
     ]);
 }
 
+export async function testMatchFirstSku(client: SimpleGraphQLClient) {
+    const result = await client.query<SearchProductsShop.Query, SearchProductsShop.Variables>(
+        SEARCH_PRODUCTS_SHOP,
+        {
+            input: {
+                term: 'L2201308',
+                groupByProduct: true,
+            },
+        },
+    );
+    expect(result.search.items.map(i => i.productName)).toEqual([
+        'Laptop',
+    ]);
+}
+
+export async function testMatchSecondSku(client: SimpleGraphQLClient) {
+    const result = await client.query<SearchProductsShop.Query, SearchProductsShop.Variables>(
+        SEARCH_PRODUCTS_SHOP,
+        {
+            input: {
+                term: 'L2201508',
+                groupByProduct: true,
+            },
+        },
+    );
+    expect(result.search.items.map(i => i.productName)).toEqual([
+        'Laptop',
+    ]);
+}
+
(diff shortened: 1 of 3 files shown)


michaelbromley:
Just thought about this a bit more and I'm now not sure that this is a desirable solution:

If we have `groupByProduct: true` and search by SKU and if we want to display the SKU in the results, it will now be a concatenation of _all_ variant SKUs right? I don't think this will be expected or useful for many applications. Same with variant name.

I.e. it solves the problem of getting the search hit, but introduces unwanted noise into the result data. What do you think?

Izayda:
Yeah, i agree with you. What behavior should we expect?

michaelbromley:
Ideal behaviour is this:

1. Product "Cheese" has 2 variants: Cheddar/CH1, Edam/ED1
2. Search with `groupByProduct: true` term "ch1"
3. result: 
    ```
    {
       productName: "Cheese",
       productVariantName: "Cheddar"
       sku: "CH1"
     }
     ```
4. Search with `groupByProduct: true` term "ed1"
5. result: 
    ```
    {
       productName: "Cheese",
       productVariantName: "Edan"
       sku: "ED1"
     }
     ```

Not sure how this would be possible given the current way we handle indexing in this plugin though.

Izayda:
Ok, i see! 
Seems, i have to implement this issue #1018

michaelbromley:
Yes. I will close this in favour of the other issue.

Izayda:
Probably we should use https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

Asked for some help of my friend (he is elastic expert).

michaelbromley:
Yes that looks promising. And 👍  for asking an expert. That will save a lot of time and effort!

Izayda:
Expert suggests me two approaches:
1. Store all data in one index with granularity of variants with denormalization of product data for each row. Then use [collapse](https://www.elastic.co/guide/en/elasticsearch//reference/7.x/collapse-search-results.html) parameter to group data by product
2. Store all data in one index with granularity of products with nested field for variants.

First approach allows easily to update index rows and allow to sort result on only relevant hits. 
Second allows not to denormalize data in index, but is hard to manage in updating variants in index.

I'll try first variant

PR #1025, opened 2021-08-09
fix typo in shopApiPath description

mohsendehghan158:


michaelbromley:
Thank you for the fix!

PR #1029, opened 2021-08-11
Fix stock and allocated are incorrectly updated due to multiple same orderline in one order

### LINKED ISSUE #1028: Stock and allocated are incorrectly updated due to multiple same orderline in one order
**Describe the bug**

current apples in stock: 50
currently being allocated: 20
order
1. apple x 10, delivery date 10/10/2021 (in memory product variant with allocated: 20)
2. apple x 5, delivery date 10/12/2021 (in memory product variant with allocated: 20)

because apple are considered as two order items due to different delivery dates, however, they are the same product variant, current code will use the allocated number that is stored in memory to calculate new stock. 

after the first loop: 
the allocated number is 20 + 10 = 30 allocated,
however, the second loop will cause a bug because it is still using the old allocated number in memory to calculate new allocation: 20 + 5 = 25 instead of  30 + 5 = 35, the correct  allocated number should be 20 + 10 + 5 = 35.

**To Reproduce**

**Expected behavior**

 - @vendure/core version:
 - Nodejs version
 - Database (mysql/postgres etc):

**Additional context**

simpian:
closes #1028 

always read db to get the correct allocation number instead of reading in memory stock.

### DIFF
--- a/packages/core/src/service/services/stock-movement.service.ts
+++ b/packages/core/src/service/services/stock-movement.service.ts
@@ -83,7 +83,10 @@ export class StockMovementService {
         const allocations: Allocation[] = [];
         const globalTrackInventory = (await this.globalSettingsService.getSettings(ctx)).trackInventory;
         for (const line of order.lines) {
-            const { productVariant } = line;
+            const productVariant = await this.connection
+            .getRepository(ctx, ProductVariant)
+            .andWhere(`productvariant.id = ${line.productVariant.id}`)
+            .getOne();
             const allocation = new Allocation({
                 productVariant,
                 quantity: line.quantity,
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/stock-movement.service.ts
```diff
@@ -83,7 +83,11 @@ export class StockMovementService {
         const allocations: Allocation[] = [];
         const globalTrackInventory = (await this.globalSettingsService.getSettings(ctx)).trackInventory;
         for (const line of order.lines) {
-            const { productVariant } = line;
+            const productVariant = await this.connection
```
michaelbromley: This line is giving a TS compiler error because the type of productVariant here is `ProductVariant | undefined`. You could use
```ts
const productVariant = await this.connection.getEntityOrThrow(ctx, ProductVariant, line.productVariant.id);
```
to solve this.

packages/core/src/service/services/stock-movement.service.ts
```diff
@@ -125,7 +129,11 @@ export class StockMovementService {
             value.items.push(orderItem);
         }
         for (const lineRow of orderLinesMap.values()) {
-            const { productVariant } = lineRow.line;
+            const productVariant = await this.connection
```
michaelbromley: Same issue as above re TS compiler error.

michaelbromley (changes requested):
Thank you for the issue and fix! I've left some comments for needed changes.

Ideally we should have an e2e test case for this. If you think you can create one let me know. Otherwise I can handle that part.

PR #1035, opened 2021-08-13
allow custom host id when creating new entity for orderable assets

### LINKED ISSUE #1034: allow custom host id when creating new entity for orderable assets
I want to create a page entity with orderable assets, however, I could not reuse product-assets component because it's just for products, so i created a custom asset picker, but then I realize that the backend asset service only allows these entity types:

```
                case 'Product':
                    return 'productId';
                case 'ProductVariant':
                    return 'productVariantId';
                case 'Collection':
                    return 'collectionId';
```

I change the logic to allow custom entity, the naming is somewhat assuming that the developer follows the same naming convention, it should automatically work.

**Additional context**

simpian:
closes #1034

### DIFF
--- a/packages/core/src/service/services/asset.service.ts
+++ b/packages/core/src/service/services/asset.service.ts
@@ -45,6 +45,9 @@ import { TransactionalConnection } from '../transaction/transactional-connection
 import { ChannelService } from './channel.service';
 import { RoleService } from './role.service';
 import { TagService } from './tag.service';
+
+import { camelCase } from 'typeorm/util/StringUtils';
+
 // tslint:disable-next-line:no-var-requires
 const sizeOf = require('image-size');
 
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/asset.service.ts
```diff
@@ -557,15 +560,19 @@ export class AssetService {
...
+                case 'Product':
+                    return 'productId';
+                case 'ProductVariant':
+                    return 'productVariantId';
+                case 'Collection':
+                    return 'collectionId';
+                default:
+                    return `${camelCase(entityName, true)}Id`;
+            }   
+        } catch (error) {
```
michaelbromley: This try-catch block is redundant, since there is no possibility of an error being thrown from the switch statement. You can omit this and just use the switch statement instead.

michaelbromley (changes requested):
Thanks for the issue & fix. See my comment re the try-catch block.

michaelbromley:
👍

packages/core/src/service/services/asset.service.ts
```diff
@@ -565,8 +568,8 @@ export class AssetService {
             case 'Collection':
                 return 'collectionId';
             default:
-                throw new InternalServerError('error.could-not-find-matching-orderable-asset');
-        }
+                return `${camelCase(entityName, true)}Id`;
```
simpian, line 571: @michaelbromley I accidentally added true flag in this function, can you please remove true flag that will actually turn the first letter capital making this Title case instead of camelCase, 

sorry for the inconvenience, thanks!

michaelbromley: Done in v1.1.5

Issue #1032, opened 2021-08-13
The order has been paid and the status is paymentauthorized. Click Modify order, and the page does not jump to the edit page


skyearthfeng:
**Describe the bug**
The order has been paid and the status is paymentauthorized. Click Modify order, and the page does not jump to the edit page
Browser development tool display error:
![新建位图图像 (3)](https://user-images.githubusercontent.com/2148134/129305002-a51d23b0-3dae-4608-a632-c32e010078c1.png)

**To Reproduce**

**Expected behavior**
the page does not jump to the edit page

**Environment (please complete the following information):**
 - @vendure/core version:10.2.2
 - Nodejs version:v12.22.1
 - Database (mysql/postgres etc):sqlite

**Additional context**

skyearthfeng:
The problem has been found. Just modify it as follows：
![1629190484(1)](https://user-images.githubusercontent.com/2148134/129695697-3582cb80-99ae-480d-bf1a-918f89a1a550.png)

michaelbromley:
Hi, I cannot reproduce this on the current latest version 1.1.3.

In your report you put the Vendure version as `@vendure/core version:10.2.2`. This is not a valid version - can you check again?

skyearthfeng:
Sorry, @ vendure / core version: 10.2.2, wrong. It should be @ vendure / core version: 1.1.3. 

My browser is:：
    Microsoft Edge
    版本 92.0.902.73 (官方内部版本) (64 位)

skyearthfeng:
The same problem is tested in chrome v92.0.4515.159 browser
![新建位图图像 (5)](https://user-images.githubusercontent.com/2148134/130008526-487f6dc5-df5f-488d-9e19-179ff5f4a778.png)

michaelbromley:
Ok. Looks like it could be related to https://github.com/apollographql/apollo-client/issues/7608

Can you tell me the exact version of Apollo Client installed using `npm ls @apollo/client`?

skyearthfeng:
"name": "@apollo/client",
  "version": "3.3.11",
  "description": "A fully-featured caching GraphQL client.",

michaelbromley:
Hi, sorry about the long delay on getting back on this. I have been doing some reading and it seems there can be memory leak issues with `shareReplay`:

* https://ncjamieson.com/avoiding-takeuntil-leaks/
* https://ncjamieson.com/whats-changed-with-sharereplay/
[...]

PR #1038, opened 2021-08-17
Fix issue #1032,Click Modify order, and the page does not jump to the…

skyearthfeng:
Fix issue #1032,Click Modify order, and the page does not jump to the edit page


michaelbromley:
Closing as the original issue went stale.

Issue #1036, opened 2021-08-14
Cannot create channel via admin ui in safari when only a single zone exists


njdancer:
**Describe the bug**
When trying to create a channel you must specify a default tax zone and shipping zone. The form used to create a channel in the admin UI leaves these fields empty when viewing in Chrome but populates them in Safari. If you only have a single zone available then this zone becomes preselected however this is not detected by the form validation. I assume the angular private state only gets this value on change and as the preselected value cannot be unselected the field is never able to change, therefore the form is always considered invalid and can never be submitted.

**To Reproduce**
1. Start Vendure with the admin-ui plugin enabled and a single zone configured.
2. In Safari, login to the admin UI, visit Settings > Channel and select "CREATE NEW CHANNEL"
3. Attempt to complete the form and save the channel. You should observe that this is impossible as the default tax zone and default shipping zones are permanently considered invalid.

<img width="648" alt="image" src="https://user-images.githubusercontent.com/666810/129431644-3a20994c-df13-4692-bafc-6261ed61e144.png">

**Expected behavior**
Safari should behave the same as chrome, i.e. the field is empty until selected by the user. Alternatively, the behaviour could be updated so that this preselected value is picked up by the validation logic and allows the form to complete. This makes a lot of sense in the case where only a single zone exists however I understand this may be uncommon and is less desirable when multiple zones exist.

**Environment (please complete the following information):**
 - @vendure/core version: 1.1.3
 - Nodejs version: v14.17.2
 - Database (mysql/postgres etc): postgres
 - Safari: 14.1.1(Mac OS Big Sur 11.4)

**Additional context**
This can be worked around for now by creating an unused zone or simply avoiding Safari and using Chrome(haven't tested other browsers).

Would also like to just say great job on Vendure. Have only started using it recently but am really enjoying it so far. I'm not terribly familiar with Angular and I imagine this is quite a simple fix for someone who is. With that said I may be able to assist with a resolution if this hangs around for too long.

michaelbromley:
Hi Nick, thanks for the report. I'm developing on Windows so never came across this, and have no easy way to test a fix unfortunately. 

I'll add it to the "planned" list and see if someone in the Slack group running macOS wants to have a go at fixing. If you do decide to fix, just leave a comment here so people know it is being worked on.

CatalinCsnMaster:
I'll take this one 😄

PR #1040, opened 2021-08-18
fix(admin-ui): fix Safari dropdown auto-select (#1036)

CatalinCsnMaster:
fix(admin-ui): fix Safari dropdown auto-select

Relates to #1036 . This commit adds an extra empty option tag to default shipping and taxing zones, in order for Safari to select these and not the first real option.


michaelbromley:
Thanks for the fix! 

As a reference for future, the fix implemented is described in this SO answer: https://stackoverflow.com/a/23638053/772859

PR #1086, opened 2021-09-13
bugfix(elastic-search-plugin): use only one index in elastic-search-plugin

### LINKED ISSUE #1017: ElasticSearch plugin: search grouped by product works only for first variant sku

### LINKED ISSUE #1018: ElasticSearch plugin: rework product index to include all variant data

Izayda:
Close #1018 
Close #1017  

@michaelbromley how can i pack new version of package and push it to npm to test on dev?

### DIFF
--- a/packages/elasticsearch-plugin/src/build-elastic-body.ts
+++ b/packages/elasticsearch-plugin/src/build-elastic-body.ts
@@ -90,15 +90,14 @@ export function buildElasticBody(
         ensureBoolFilterExists(query);
         query.bool.filter.push({ term: { enabled: true } });
     }
+
     if (priceRange) {
         ensureBoolFilterExists(query);
-        query.bool.filter = query.bool.filter.concat(createPriceFilters(priceRange, false, !!groupByProduct));
+        query.bool.filter = query.bool.filter.concat(createPriceFilters(priceRange, false));
     }
     if (priceRangeWithTax) {
         ensureBoolFilterExists(query);
-        query.bool.filter = query.bool.filter.concat(
-            createPriceFilters(priceRangeWithTax, true, !!groupByProduct),
-        );
+        query.bool.filter = query.bool.filter.concat(createPriceFilters(priceRangeWithTax, true));
     }
 
     const sortArray = [];
@@ -109,19 +108,34 @@ export function buildElasticBody(
             });
         }
         if (sort.price) {
-            const priceField = groupByProduct ? 'priceMin' : 'price';
+            // const priceField = groupByProduct ? 'priceMin' : 'price';
+            const priceField = 'price';
             sortArray.push({ [priceField]: { order: sort.price === SortOrder.ASC ? 'asc' : 'desc' } });
         }
     }
-    return {
-        query: searchConfig.mapQuery
-            ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
-            : query,
-        sort: sortArray,
-        from: skip || 0,
-        size: take || 10,
-        track_total_hits: searchConfig.totalItemsMaxSize,
-    };
+
+    if (groupByProduct) {
+        return {
+            query: searchConfig.mapQuery
+                ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
+                : query,
+            sort: sortArray,
+            from: skip || 0,
+            size: take || 10,
+            track_total_hits: searchConfig.totalItemsMaxSize,
+            collapse: { field: `productId` },
+        };
+    } else {
+        return {
+            query: searchConfig.mapQuery
+                ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
+                : query,
+            sort: sortArray,
+            from: skip || 0,
+            size: take || 10,
+            track_total_hits: searchConfig.totalItemsMaxSize,
+        };
+    }
 }
 
 function ensureBoolFilterExists(query: { bool: { filter?: any } }) {
@@ -130,35 +144,16 @@ function ensureBoolFilterExists(query: { bool: { filter?: any } }) {
     }
 }
 
-function createPriceFilters(range: PriceRange, withTax: boolean, groupByProduct: boolean): any[] {
+function createPriceFilters(range: PriceRange, withTax: boolean): any[] {
     const withTaxFix = withTax ? 'WithTax' : '';
-    if (groupByProduct) {
-        return [
-            {
-                range: {
-                    [`price${withTaxFix}Min`]: {
-                        gte: range.min,
-                    },
+    return [
+        {
+            range: {
+                ['price' + withTaxFix]: {
+                    gte: range.min,
+                    lte: range.max,
                 },
             },
-            {
-                range: {
-                    [`price${withTaxFix}Max`]: {
-                        lte: range.max,
-                    },
-                },
-            },
-        ];
-    } else {
-        return [
-            {
-                range: {
-                    ['price' + withTaxFix]: {
-                        gte: range.min,
-                        lte: range.max,
-                    },
-                },
-            },
-        ];
-    }
+        },
+    ];
 }
--- a/packages/elasticsearch-plugin/src/build-elastic-body.ts
+++ b/packages/elasticsearch-plugin/src/build-elastic-body.ts
@@ -108,34 +108,24 @@ export function buildElasticBody(
             });
         }
         if (sort.price) {
-            // const priceField = groupByProduct ? 'priceMin' : 'price';
             const priceField = 'price';
             sortArray.push({ [priceField]: { order: sort.price === SortOrder.ASC ? 'asc' : 'desc' } });
         }
     }
 
+    const body: SearchRequestBody = {
+        query: searchConfig.mapQuery
+            ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
+            : query,
+        sort: sortArray,
+        from: skip || 0,
+        size: take || 10,
+        track_total_hits: searchConfig.totalItemsMaxSize,
+    };
     if (groupByProduct) {
-        return {
-            query: searchConfig.mapQuery
-                ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
-                : query,
-            sort: sortArray,
-            from: skip || 0,
-            size: take || 10,
-            track_total_hits: searchConfig.totalItemsMaxSize,
-            collapse: { field: `productId` },
-        };
-    } else {
-        return {
-            query: searchConfig.mapQuery
-                ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
-                : query,
-            sort: sortArray,
-            from: skip || 0,
-            size: take || 10,
-            track_total_hits: searchConfig.totalItemsMaxSize,
-        };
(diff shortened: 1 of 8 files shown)


michaelbromley:
Just pulled this down locally to review. I'm getting 3 failing e2e tests at the moment - do you see the same?

```
 FAIL  e2e/elasticsearch-plugin.e2e-spec.ts (46.07 s)
  ● Elasticsearch plugin › admin api › updating the index › when grouped, disabled is false if at least one variant is
 enabled

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
        Object {
    -     "enabled": true,
    +     "enabled": false,
          "productId": "T_1",
        },
      ]
  ● Elasticsearch plugin › admin api › channel handling › removing from channel with multiple languages

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "product de",
    + ]

  ● Elasticsearch plugin › admin api › multiple language handling › indexes product variant-level languages

    TypeError: Cannot read property 'productName' of undefined

```

packages/elasticsearch-plugin/src/types.ts
```diff
@@ -8,6 +8,7 @@ import {
     SearchResult,
 } from '@vendure/common/lib/generated-types';
 import { ID, JsonCompatible } from '@vendure/common/lib/shared-types';
+import { unique } from '@vendure/common/lib/unique';
```
michaelbromley: This is unused, can be removed.

Izayda:
Yes, i fixed 2 tests of 3. Now fixing last test. I'll write your question in DM in slack about last test today

packages/elasticsearch-plugin/src/build-elastic-body.ts
```diff
@@ -109,19 +108,34 @@ export function buildElasticBody(
             });
         }
         if (sort.price) {
-            const priceField = groupByProduct ? 'priceMin' : 'price';
+            // const priceField = groupByProduct ? 'priceMin' : 'price';
```
michaelbromley: Comment can be cleaned up, and maybe we don't need the const anymore?

packages/elasticsearch-plugin/src/build-elastic-body.ts
```diff
@@ -109,19 +108,34 @@ export function buildElasticBody(
...
-            ? searchConfig.mapQuery(query, input, searchConfig, channelId, enabledOnly)
-            : query,
-        sort: sortArray,
-        from: skip || 0,
-        size: take || 10,
-        track_total_hits: searchConfig.totalItemsMaxSize,
-    };
+
+    if (groupByProduct) {
+        return {
```
michaelbromley: These 2 `if` branches are almost identical, apart from the `collapse` property right? I think it makes sense to create a single object and then have add the `collapse` part after:
```ts
const body = { 
  // query, sort, from, size, track
};
if (groupByProduct) {
  body.collapse = { field: `productId` };
}
return body;
```

packages/elasticsearch-plugin/src/elasticsearch.service.ts
```diff
@@ -193,13 +192,14 @@ export class ElasticsearchService implements OnModuleInit, OnModuleDestroy {
         );
         if (groupByProduct) {
             try {
-                const { body }: { body: SearchResponseBody<ProductIndexItem> } = await this.client.search({
-                    index: indexPrefix + PRODUCT_INDEX_NAME,
+                const { body }: { body: SearchResponseBody<VariantIndexItem> } = await this.client.search({
+                    index: indexPrefix + VARIANT_INDEX_NAME,
                     body: elasticSearchBody,
                 });
+                const totalItems = await this.totalHits(ctx, input, groupByProduct);
```
michaelbromley, line 199: totalItems now needs a separate search? Does this have any noticeable perf impact?

packages/elasticsearch-plugin/src/elasticsearch.service.ts
```diff
@@ -293,18 +335,37 @@ export class ElasticsearchService implements OnModuleInit, OnModuleDestroy {
...
+                                field: `productId`,
+                            },
+                        },
+                    },
                 },
-            },
-        };
+            };
+        } else {
+            elasticSearchBody.aggs = {
```
michaelbromley: The `elasticSearchBody.aggs.collection.terms` part of these 2 `if` branches is identical right? In that case, it can be extracted outside the `if` statement:
```ts
elasticSearchBody.aggs = {
    collection: {
        terms: {
            field,
            size: this.options.searchConfig.collectionMaxSize,
        },
    },
};
if (groupByProduct) {
    elasticSearchBody.aggs.collection.aggs = { /*.... */ };
}
```

packages/elasticsearch-plugin/src/elasticsearch.service.ts
```diff
@@ -445,29 +497,30 @@ export class ElasticsearchService implements OnModuleInit, OnModuleDestroy {
...
         };
-        this.addCustomMappings(result, source, this.options.customProductMappings);
+        ElasticsearchService.addCustomMappings(result, source, this.options.customProductMappings, true);
         return result;
     }
 
-    private getSearchResultAssets(
-        source: ProductIndexItem | VariantIndexItem,
-    ): { productAsset: SearchResultAsset | undefined; productVariantAsset: SearchResultAsset | undefined } {
+    private static getSearchResultAssets(source: ProductIndexItem | VariantIndexItem): {
```
michaelbromley: Why `static`?

michaelbromley (commented):
Very nice work so far! I left a few comments and questions. Once we have all the tests passing, I'd also like to do a little bit of benchmarking to see if there is any performance change here (good or bad).

packages/elasticsearch-plugin/src/elasticsearch.service.ts
Izayda: Yeah, regarding to [docs](https://www.elastic.co/guide/en/elasticsearch//reference/7.x/collapse-search-results.html#collapse-search-results)

`The total number of hits in the response indicates the number of matching documents without collapsing. The total number of distinct group is unknown.`

I deployed this version of plugin yesterday to dev env. But i have no idea how to test perf.
NPM package: https://www.npmjs.com/package/elasticsearch-plugin-test

packages/elasticsearch-plugin/src/elasticsearch.service.ts
Izayda: It was WebStorm suggestion. In Python static functions are rather quicker. Revert it

Izayda:
Packaged version https://www.npmjs.com/package/elasticsearch-plugin-test

Now i'm testing on my dev env

packages/elasticsearch-plugin/src/elasticsearch.service.ts
michaelbromley: Ok, got it. I'll see if I can test perf on this point. Although if that is the only way to do it, the results are moot.

packages/elasticsearch-plugin/src/indexer.controller.ts
```diff
@@ -549,12 +466,12 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
                     updatedProductVariants.forEach(v => (v.enabled = false));
                 }
                 Logger.verbose(`Updating Product (${productId})`, loggerCtx);
-                if (updatedProductVariants.length) {
-                    operations.push(...(await this.updateVariantsOperations(updatedProductVariants)));
+                const languageVariants: LanguageCode[] = [];
+                languageVariants.push(...product.translations.map(t => t.languageCode));
+                for (const variant of product.variants) {
+                    languageVariants.push(...variant.translations.map(t => t.languageCode));
```
michaelbromley, line 472: I have been testing the latest commit with my real project, and keep running into a "Maximum call stack size exceeded" error which causes the `reindex` job to fail.

I have slowly traced it down to this line, I _think_ which might be causing the issue. The job keeps failing on the same product, which has 139 variants. After this line has executed, the `languageVariants` array consists of `'en'` x 140! (1 for the product + 139 for each variant).

This is then causing 140x the amount of work in line 493 to execute **19,460** times, rather than 139 times. I think this is related to the call stack error.

I believe the missing part is adding the `unique()` helper function to only use unique values in the `languageVariants` array.

michaelbromley (changes requested):

packages/elasticsearch-plugin/src/indexer.controller.ts
Izayda: Yeah, in previous version of this PR there was not any problem. I also caught this error today
```
worker_1  | verbose  [ElasticsearchPlugin] Updating 1 Products
worker_1  | verbose  [ElasticsearchPlugin] Deleting 1 Product (id: 5990)
worker_1  | verbose  [ElasticsearchPlugin] Deleting 312 ProductVariants
worker_1  | verbose  [ElasticsearchPlugin] Updating Product (5990)
worker_1  | error  [Vendure Worker] {}
```

Izayda: Fixed. Successfully rebuild index on my dev. Pushed fresh version of package to npm

michaelbromley (approved):
This is very good now. I ran some rough performance tests against my real dataset of ~10k variants. The search query includes the `facetValues` field - searching only the items would be faster.

```
before:
reindex: 201s
avg grouped search: 77ms
avg ungrouped search: 66ms

after:
reindex: 177s
avg grouped search: 101ms
avg ungrouped search: 77ms
```

So as I expected, indexing has sped up quite a bit since we only have a single index now, whereas querying is marginally slower, probably due to the additional query required to get the total count.

Izayda:
Maybe this two queries can be sent in parallel?

Izayda:
I think, in release notes should be pointed, that after deploying index should be recreated (that's why i will release this package at night today on production)

michaelbromley:
> Maybe this two queries can be sent in parallel?

Yes that could be a worthwhile optimization.

Yes, will make a note of the need to recreate index in the blog post.

Thank you for putting so much work into this plugin. You have really made a big difference and I'm glad I have such capable support!

Izayda:
Ok, will try to optimize, thank you!

No problem, I really like to contribute to Vendure!

PR #1102, opened 2021-09-19
fix(core):  Fixed wrong event type when a variant is created

### LINKED ISSUE #1095: No `created` event is fired for ProductVariantEvent
**Describe the bug**
When a new ProductVariant is created I expect an event of the type `created` to be fired. I only get an `updated` event and all the product variants in it have the same `createdAt` and `updatedAt` values, which means they were just created.

**To Reproduce**
Add this to any init method of a plugin that subscribes to events

```
this.eventBus.ofType(ProductVariantEvent).subscribe(event => {
      console.log('Productvariant Event', event);
});
```

**Expected behavior**
A `created` event to be fired when product variants are created.

 - @vendure/core version: 1.2.0
 - Nodejs version: 14
 - Database: Postgres 13
 -

skid:
Fixes #1095

### DIFF
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -321,7 +321,7 @@ export class ProductVariantService {
             ids.push(id);
         }
         const createdVariants = await this.findByIds(ctx, ids);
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks!

Issue #728, opened 2021-02-24
Search query doesn't work properly with localization and collectionSlug input


Clever-Up:
**Describe the bug**
If a collection is configured for 2 or more languages with different slugs, if you perform a "search" query using shop api localized url (`xx/shop-api?languageCode=en`, for instance), it fails to find the products in that collection (which on turn are localized in both languages), as it seems to expect to always expect to receive the default language slug and not the localized one.
Replacing the collectionSlug with collectionId obviously fixes the issue and is a good workaround.

**To Reproduce**
1. Go to Vendure admin and configure 2 supported languages (e.g. italian main language, english secondary language)
2. Create a new collection and configure **different slugs** for each language
3. Associate at least one product to this collection and localize it for the 2 (or more) languages
4. Go to the graphql playground with the `languageCode=en` parameter (where en is the secondary language)
5. Perform a search query using the secondary language slug and see empty result

**Expected behavior**
Localized search query should "use" the localized collection slug

**Environment (please complete the following information):**
 - @vendure/core version: 0.18.3
 - Database (mysql/postgres etc): postgres

michaelbromley:
Thanks for the report. I can reproduce this 👍

Clever-Up:
I have just posted a comment on #712 describing my latest tests, as I guess both these issue are related to products search indexing

PR #1114, opened 2021-09-24
fix(core): fix collection update event generation

### LINKED ISSUE #1015: ElasticSearch plugin: update Collection
**Describe the bug**
In the ElasticSearch plugin, a product or productVariant contains a list of collection slugs. When a collection is updatet, the indexer service is only triggered when the collection filters are changed. When the collection slug is updatet without changing the filters, the reindexing is not carried out and the products remain in ElasticSearch with the wrong slug.

**To Reproduce**
1. Create collection with slug `slug-1`
2. Add product to collection
3. Update collection so that slug is now `slug-2`
4. Elastic still contains the product with `slug-1`

**Expected behavior**
Collection updates should be propagated correctly to the ElasticSearch Plugin.

 - @vendure/core version: v1.0.3

Izayda:
Close #1015

### DIFF
--- a/packages/core/src/service/services/collection.service.ts
+++ b/packages/core/src/service/services/collection.service.ts
@@ -44,7 +44,7 @@ import { AssetService } from './asset.service';
 import { ChannelService } from './channel.service';
 import { FacetValueService } from './facet-value.service';
 
-type ApplyCollectionFiltersJobData = { ctx: SerializedRequestContext; collectionIds: ID[] };
+type ApplyCollectionFiltersJobData = { ctx: SerializedRequestContext; collectionIds: ID[]; applyToChangedVariantsOnly?: boolean; };
 
 @Injectable()
 export class CollectionService implements OnModuleInit {
@@ -64,7 +64,8 @@ export class CollectionService implements OnModuleInit {
         private slugValidator: SlugValidator,
         private configArgService: ConfigArgService,
         private customFieldRelationService: CustomFieldRelationService,
-    ) {}
+    ) {
+    }
 
     async onModuleInit() {
         const productEvents$ = this.eventBus.ofType(ProductEvent);
@@ -99,7 +100,7 @@ export class CollectionService implements OnModuleInit {
                     }
                     completed++;
                     if (collection) {
-                        const affectedVariantIds = await this.applyCollectionFiltersInternal(collection);
+                        const affectedVariantIds = await this.applyCollectionFiltersInternal(collection, job.data.applyToChangedVariantsOnly);
                         job.setProgress(Math.ceil((completed / job.data.collectionIds.length) * 100));
                         this.eventBus.publish(
                             new CollectionModificationEvent(ctx, collection, affectedVariantIds),
@@ -362,10 +363,12 @@ export class CollectionService implements OnModuleInit {
             await this.applyFiltersQueue.add({
                 ctx: ctx.serialize(),
                 collectionIds: [collection.id],
+                applyToChangedVariantsOnly: false,
             });
+        } else {
+            const affectedVariantIds = await this.getCollectionProductVariantIds(collection);
+            this.eventBus.publish(new CollectionModificationEvent(ctx, collection, affectedVariantIds));
         }
-        const affectedVariantIds = await this.getCollectionProductVariantIds(collection);
-        this.eventBus.publish(new CollectionModificationEvent(ctx, collection, affectedVariantIds));
         return assertFound(this.findOne(ctx, collection.id));
     }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/collection.service.ts
```diff
@@ -364,6 +364,8 @@ export class CollectionService implements OnModuleInit {
                 collectionIds: [collection.id],
             });
         }
```
michaelbromley, line 371: If `input.filters` is populated, then a `CollectionModificationEvent` will be published twice - once here on line 368, and also once on line 105 after the "apply-collection-filters" job has complemented.

Therefore I think lines 367 & 368 should be inside an `else {} ` block.

Izayda:
Don't know why check failed. It's strange

michaelbromley (changes requested):
Ah this was simpler than I had thought! Just check my comment and if you agree, make a minor change.

packages/core/src/service/services/collection.service.ts
Izayda: It was my first implementation :) But this doesn't work. Cause https://github.com/vendure-ecommerce/vendure/blob/6f7cc34acda54693fc3ab33cf9f209ecabfc45cb/packages/core/src/service/services/collection.service.ts#L434  method checks only new/deleted products.

Without else block everything works as expected, because search plugins reduces ids here: https://github.com/vendure-ecommerce/vendure/blob/6f7cc34acda54693fc3ab33cf9f209ecabfc45cb/packages/elasticsearch-plugin/src/plugin.ts#L325

michaelbromley: Ah ok that makes sense. In that case, would it make sense to at least add a call to `unique()` in the plugin, since in this case there will likely be duplicate variant IDs sent with the 2 events? (if so, would also apply to the DefaultSearchPlugin)

I've also got a nagging feeling that there is a potentially problematic race condition here:

1. The "apply-collection-filters" job is enqueued.
2. The CollectionModificationEvent is published with the _current_ variants associated with the Collection
3. The job gets processed, updates the variants based on the filter changes, and then publishes another CollectionModificationEvent, with the new set of variant ids.

I _think_ it's actually ok, because once all these variant IDs arrive at the search indexer, it will grab the fresh, up-to-date data on each variant. Just thinking "out loud" here - maybe you can spot any issues with this flow?

Izayda: I changed logic of `applyCollectionFiltersInternal` and `update`.

If `input.filters` is provided in `update` method, we settle job with `applyToChangedVariantsOnly=false`.
`applyCollectionFiltersInternal` checks `applyToChangedVariantsOnly` variable and if it equals `true` (default) - then only new/deleted variants are processed. if it equals `false` - then all variants are processed.

If `input.filters` is not provided in `update` method hen all variants are processed.

So, no need to think about that in plugins and no chance to race condition.

Izayda:
Probably close #728

packages/core/src/service/services/collection.service.ts
```diff
@@ -460,11 +465,19 @@ export class CollectionService implements OnModuleInit {
...
         const preIdsSet = new Set(preIds);
         const postIdsSet = new Set(postIds);
-        const difference = [
-            ...preIds.filter(id => !postIdsSet.has(id)),
-            ...postIds.filter(id => !preIdsSet.has(id)),
-        ];
-        return difference;
+
+        if (applyToChangedVariantsOnly === undefined ? true : applyToChangedVariantsOnly) {
```
michaelbromley: This could be simplified to 
```ts
if (applyToChangedVariantsOnly ?? true) { }
```
where `??` is the nullish-coalescing operator, where the right-hand-side of the operator gets evaluated only if the left-hand side is "nullish" (null or undefined)

OR 

you could define a default value in the method signature, and then you can do a simplified if check: 
```ts
applyCollectionFiltersInternal(collection: Collection, applyToChangedVariantsOnly = true) {
   // ...
   if (applyToChangedVariantsOnly) {}
}
```

packages/core/src/service/services/collection.service.ts
```diff
@@ -431,7 +436,7 @@ export class CollectionService implements OnModuleInit {
     /**
      * Applies the CollectionFilters
      */
-    private async applyCollectionFiltersInternal(collection: Collection): Promise<ID[]> {
+    private async applyCollectionFiltersInternal(collection: Collection, applyToChangedVariantsOnly?: boolean): Promise<ID[]> {
```
michaelbromley: Can you add a short description of what "applyToChangedVariantsOnly" means and why we have it here? I will surely forget in the future and then need to dig through git history to get the context back!

michaelbromley (commented):
A couple of comments, otherwise looking good!

packages/core/src/service/services/collection.service.ts
Izayda: Added description

packages/core/src/service/services/collection.service.ts
Izayda: Added second variant

michaelbromley:
Thanks!

PR #1122, opened 2021-09-29
Fix Spanish translation of facets

JorgeFPastor:
All e-commerce sites in Spanish use the word "Etiquetas" instead of "Facetas", so it will be understood a lot better with this change.


michaelbromley:
Thank you!

PR #1130, opened 2021-10-01
feat(elasticSearchPlugin): add inStock attribute and filter #870

### LINKED ISSUE #870: Enable filter to only show in-stock products and SKUs
I believe this will need update to both shop-api and admin-api.

Izayda:
Partly close #870 

@michaelbromley i didn't write e2e tests. Would you like to review this code and after that i'll write e2e tests. 
Another question. Is it reasonable to implement this attribute and filter for DefaultSearchPluign? It will change index table structure.

### DIFF
--- a/packages/elasticsearch-plugin/src/build-elastic-body.ts
+++ b/packages/elasticsearch-plugin/src/build-elastic-body.ts
@@ -27,6 +27,7 @@ export function buildElasticBody(
         priceRangeWithTax,
         priceRange,
         facetValueFilters,
+        inStock,
     } = input;
     const query: any = {
         bool: {},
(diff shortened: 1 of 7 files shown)


michaelbromley:
The changes look good to me - go ahead with e2e tests!

> Is it reasonable to implement this attribute and filter for DefaultSearchPluign? It will change index table structure.

Yes, in #870 I proposed:

> For the DefaultSearchPlugin this will cause a breaking schema change. Therefore I propose to put it behind a flag so we can release the update in a non-breaking way in version 1.1.

This could be added as an option passed to a static `init()` method as is the convention with plugins.

Izayda:
Added unit tests and e2e tests to elasticsearch-plugin.

Tried to implement this logic in defaultsearch-plugin behind a flag, but i have no idea how to dynamically expand db entity, graphql schema etc. 

@michaelbromley can you implement this logic in defaultsearch-plugin by yourself? I'll will look how would you do that and will learn some typescript magic)) There is for `@ts-ignore` in e2e test due to my inability to expand defaultsearch-plugin

Issue #1143, opened 2021-10-10
[elasticsearch-plugin] Allow script_fields in search request body


Draykee:
**Is your feature request related to a problem? Please describe.**
There is currently no possibility to set `script_fields` to allow script evaluation for each hit.
https://www.elastic.co/guide/en/elasticsearch/reference/7.8/search-fields.html#script-fields

**Describe the solution you'd like**

- Add a search config entry to return a script_fields configuration.
- The field that contains the evaluated script value needs to be mapped to the search result
- `"_source": true,` needs to be set in the request body to still include the whole index document (When using script fields, the default response will not contain the _source field but only the custom field we have defined)

**Additional context**
Assuming the product should receive a location via custom fields. I would like to add geo location based search and therefore I extend the SearchInput with `latitude` , `longitude` (and als `distance`). Here is a example of a possible configutation:
```JS
indexMappingProperties: {
     location: {
         type: 'geo_point', // contains function arcDistance
     },
},
customProductMappings: {
     location: {
         graphQlType: 'String',
         valueFn: (product: Product) => {
             const custom = product.customFields.location;
             return `${custom.latitude},${location.longitude}`;
         },
     }
},
mapQuery: (query: any, input: SearchInput, searchConfig: DeepRequired<SearchConfig>) => {
  // Dummy implementation of the geo distance search
  const distance = input.radius;
  const lat = input.latitude;
  const lon = input.longitude;
  
  if (lat && lon && distance) {
      query.bool.must = {
          match_all: {},
      };
      query.bool.filter = {
          geo_distance: {
              distance: `${distance}km`,
              location: {
                  lat,
                  lon,
              },
          },
      };
  
  }
  
  return query;
},
// The new config entry !
scriptFields: (input) => {
     // assuming SearchInput was extended with latitude and longitude
    // this script would add a 
     const lat = input.latitude;
     const lon = input.longitude;
     return {
         distance: {
             script: `doc['location'].arcDistance(${lat}, ${lon})`,
         }
     }
}
```

This script will add a field with the distance like this inside the ElasticSearch response:
```JS
"fields": {
  "distance": [
    103861.9371616157
  ]
}
```

Draykee:
[...]

PR #1144, opened 2021-10-11
feat(elasticsearch): Extend search config with fields for script evaluation of every hit (#1143)

Draykee:
- Extend search config with fields for script evaluation of every hit (#1143)
- Added `customScriptFields` to search result


michaelbromley:
I would personally do it locally on the command line, using a rebase command. Let me know if you need any help with that (rebase can get a bit tricky sometimes).

PR #1145, opened 2021-10-11
feat(elasticsearch): Extend search config with fields for script evaluation of every hit (#1143)

Draykee:
- Extend search config with fields for script evaluation of every hit (#1143)
- Added customScriptFields to search result

### DIFF
--- a/packages/elasticsearch-plugin/src/options.ts
+++ b/packages/elasticsearch-plugin/src/options.ts
@@ -351,7 +351,7 @@ export interface SearchConfig {
     /**
      * @description
      * Sets `script_fields` inside the elasticsearch body which allows returning a script evaluation for each hit
-     * @since 1.2.4
+     * @since 1.3.0
      * @example
      * ```TypeScript
      * indexMappingProperties: {
@@ -370,13 +370,14 @@ export interface SearchConfig {
      * },
      * scriptFields: {
      *      distance: {
-     *          graphQlType: 'Number'
-     *          valFn: (input) => {
+     *          graphQlType: 'Int',
+     *          environment: 'product',
+     *          scriptFn: (input) => {
      *              // assuming SearchInput was extended with latitude and longitude
      *              const lat = input.latitude;
      *              const lon = input.longitude;
      *              return {
-     *                  script: `doc['location'].arcDistance(${lat}, ${lon})`,
+     *                  script: `doc['product-location'].arcDistance(${lat}, ${lon})`,
      *              }
      *          }
      *      }


Izayda:
> Is this always the case?

Yes. WIth [this PR](https://github.com/vendure-ecommerce/vendure/pull/1086) product data and variant data are stored together in one index. So, we need to consider cases, when we create customProductMapping and customVariantMapping with similar names. That's why at [storage layer](https://github.com/vendure-ecommerce/vendure/blob/4efe258c6f56bd9ed8232263cd99f256f2c37141/packages/elasticsearch-plugin/src/indexer.controller.ts#L756) there are prefixes.

packages/elasticsearch-plugin/src/build-elastic-body.ts
```diff
@@ -134,6 +144,33 @@ function ensureBoolFilterExists(query: { bool: { filter?: any } }) {
...
+    input: ElasticSearchInput,
+    groupByProduct?: boolean,
+): any | undefined {
+    if (scriptFields) {
+        const fields = Object.keys(scriptFields);
+        if (fields.length) {
+            const result: any = {};
+            for (const name of fields) {
+                const scriptField = scriptFields[name];
+                if (scriptField.environment === 'product' && groupByProduct === true) {
```
Izayda, line 158: Can `scriptField.environment` be a Type with three options - Product, Variant, Both?

kevmtt: Yes

packages/elasticsearch-plugin/src/indexer.controller.ts
```diff
@@ -805,6 +805,11 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
             productCollectionSlugs: [],
             productChannelIds: product.channels.map(c => c.id),
         };
+        const customMappings = Object.entries(this.options.customProductMappings);
```
kevmtt: Shall I apply this code here?
```JS
const productCustomMappings = Object.entries(this.options.customProductMappings);
for (const [name, def] of productCustomMappings) {
    item[`product-${name}`] = def.valueFn(v.product, variants, languageCode);
}
````

Otherwise I will get an error in the e2e test because the script_fields can't evaluate the indicies without the customProductMapping. From my understanding we should only index the product mappings and not the one for variants.. (because there are none)

@michaelbromley

File: packages/elasticsearch-plugin/src/options.ts [hunk 3]
```diff
@@ -348,6 +348,43 @@ export interface SearchConfig {
...
+     *      location: {
+     *          type: 'geo_point', // contains function arcDistance
+     *      },
+     * },
+     * customProductMappings: {
+     *      location: {
+     *          graphQlType: 'String',
+     *          valueFn: (product: Product) => {
+     *              const custom = product.customFields.location;
+     *              return `${custom.latitude},${location.longitude}`;
```
michaelbromley, line 367: Should this second interpolated value be `custom.longitude`?

kevmtt: Yes it should

michaelbromley: ok thanks. So, for your info, I have taken your commits into the `minor` branch locally (I had to do it locally to clean up the git history a bit) and I'm really excited about the kind of flexibility this unlocks! Thank you for your efforts on this!

michaelbromley:
Your commits are now merged into the `minor` branch, thank you!

These changes will be made available in the next version, 1.3, which I hope to release this week! 🥳

PR #1150, opened 2021-10-12
fix(core): Fix shipping price when the promotion is not applicable anymo

huseyinalp:
Fix the following issue;
assume that we have a shipping promotion, and it gives free delivery after 30 USD.

The customer added some items to the cart and reached the threshold of free delivery. Shipping price is free as expected.
Then he removed one item from the cart and if cart total is lower than threshold of free delivery, the shipping price should be calculated again.

So we should removed previously added adjustments from shipping line items if there isn't applicable promotion for shipping.


michaelbromley:
Wow great catch, thank you!

PR #1199, opened 2021-10-29
feat(core): Add CSV import in multiple languages

heiko-r:
It turned out a little more complex than anticipated. I tried to maintain as much compatibility with the monolingual import as possible, but this required adding an optional `mainLanguage` parameter to `parseProducts`, which is used to specify the language to use for the `Product.name`, `Product.slug`, etc. properties. So even though the `ParsedProduct` and `ParsedProductVariant` interfaces have additional properties now, I think it should be compatible with most, if not all, existing usages. Not sure if you consider this a breaking change.

Also a note: I pushed with `--no-verify`, because there's an error during e2e testing:
```
@vendure/core: FAIL e2e/custom-fields.e2e-spec.ts (11.346 s)
@vendure/core:   ● Custom fields › validation › invalid datetime
@vendure/core:     expect(received).toEqual(expected) // deep equality
@vendure/core:     Expected: StringContaining "The custom field 'validateDateTime' value [2019-01-01T05:25:00.000Z] is less than the minimum [2019-01-01T08:30]"
@vendure/core:     Received: "fail is not defined"
@vendure/core:       13 |     };
@vendure/core:       14 | }
@vendure/core:     > 15 |
@vendure/core:          | ^
@vendure/core:       at Object.<anonymous> (utils/assert-throws-with-message.ts:15:33)
```
It wasn't me! 😬 It was already broken before I made any changes!  😬😉 (and I also didn't investigate it any further yet)

### DIFF
--- a/docs/content/developer-guide/importing-product-data.md
+++ b/docs/content/developer-guide/importing-product-data.md
@@ -78,9 +78,9 @@ To import custom fields with `list` set to `true`, the data should be separated 
 
 #### Importing data in multiple languages
 
-If a field is translatable (i.e. of `localeString` type), you can use column names with an append language codes (e.g. `name:en`, `name:de`, `product:keywords:en`, `product:keywords:de`) to specify its value in multiple languages.
+If a field is translatable (i.e. of `localeString` type), you can use column names with an appended language code (e.g. `name:en`, `name:de`, `product:keywords:en`, `product:keywords:de`) to specify its value in multiple languages.
 
-Use of language codes has to be consistent throughout the file, i.e. each translated field has to use the same set of translated columns, or none, in which case the generic column's value will be used for all translations.
+Use of language codes has to be consistent throughout the file. You don't have to translate every translatable field. If there are no translated columns for a field, the generic column's value will be used for all languages. But when you do translate columns, the set of languages for each of them needs to be the same. As an example, you cannot use `name:en` and `name:de`, but only provide `slug:en` (it's okay to use only a `slug` column though, in which case this slug will be used for both the English and the German version).
 
 ## Initial Data
 
--- a/packages/core/src/data-import/providers/import-parser/import-parser.ts
+++ b/packages/core/src/data-import/providers/import-parser/import-parser.ts
@@ -35,8 +35,6 @@ const requiredColumns: string[] = [
 ];
 
 export interface ParsedOptionGroup {
-    name: string;
-    values: string[];
     translations: Array<{
         languageCode: LanguageCode;
         name: string;
@@ -45,8 +43,6 @@ export interface ParsedOptionGroup {
 }
 
 export interface ParsedFacet {
-    facet: string;
-    value: string;
     translations: Array<{
         languageCode: LanguageCode;
         facet: string;
@@ -55,7 +51,6 @@ export interface ParsedFacet {
 }
 
 export interface ParsedProductVariant {
-    optionValues: string[];
     sku: string;
     price: number;
     taxCategory: string;
@@ -70,15 +65,9 @@ export interface ParsedProductVariant {
             [name: string]: string;
         };
     }>;
-    customFields: {
-        [name: string]: string;
-    };
 }
 
 export interface ParsedProduct {
-    name: string;
-    slug: string;
-    description: string;
     assetPaths: string[];
     optionGroups: ParsedOptionGroup[];
     facets: ParsedFacet[];
@@ -91,9 +80,6 @@ export interface ParsedProduct {
             [name: string]: string;
         };
     }>;
-    customFields: {
-        [name: string]: string;
-    };
 }
 
 export interface ParsedProductWithVariants {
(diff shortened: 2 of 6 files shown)


docs/content/developer-guide/importing-product-data.md
```diff
@@ -76,6 +76,12 @@ To import custom fields with `list` set to `true`, the data should be separated
 ... ,tablet|pad|android
 ```
 
+#### Importing data in multiple languages
+
+If a field is translatable (i.e. of `localeString` type), you can use column names with an append language codes (e.g. `name:en`, `name:de`, `product:keywords:en`, `product:keywords:de`) to specify its value in multiple languages.
+
+Use of language codes has to be consistent throughout the file, i.e. each translated field has to use the same set of translated columns, or none, in which case the generic column's value will be used for all translations.
```
michaelbromley: I don't fully understand what is meant by this:

"each translated field has to use the same set of translated columns, or none, in which case the generic column's value will be used for all translations"

Could you elaborate in some way? Maybe an example to illustrate what you mean.

packages/core/src/data-import/providers/importer/importer.ts
```diff
@@ -6,6 +6,7 @@ import ProgressBar from 'progress';
 import { Observable } from 'rxjs';
 import { Stream } from 'stream';
 
+import { InternalServerError } from '../../..';
```
michaelbromley: can you import this from the file itself rather than the top-level barrel file?

packages/core/src/data-import/providers/import-parser/import-parser.ts
```diff
@@ -46,13 +80,16 @@ export interface ParsedProduct {
     slug: string;
```
michaelbromley: Are these "main" translatable properties still needed? As far as I can see, all translations put in the `translations` array, and the mainLanguage ones are duplicated in the root object. Is my understanding correct?

packages/core/src/data-import/providers/import-parser/import-parser.spec.ts
```diff
@@ -51,23 +77,33 @@ describe('ImportParser', () => {
...
         });
 
+        it('works with multilingual input', async () => {
+            const importParser = new ImportParser(mockConfigService);
+
+            const filename = path.join(__dirname, 'test-fixtures', 'multiple-languages.csv');
+            const input = fs.createReadStream(filename);
+            const result = await importParser.parseProducts(input);
+
+            expect(result.results).toMatchSnapshot();
```
michaelbromley, line 87: I wish I never used snapshot tests now. They are just so opaque and I can't just see by glancing at the e2e test what particular outcome is being tested for. I understand why you used it here - just following my lead. You don't need to change it, but you can if you want. Mostly I just felt the need to vent about snapshot tests :D

packages/core/src/data-import/providers/importer/importer.ts
```diff
@@ -158,15 +159,29 @@ export class Importer {
...
-                    {
-                        languageCode,
-                        name: product.name,
-                        description: product.description,
-                        slug: product.slug,
-                        customFields,
-                    },
-                ],
+                translations:
+                    !product.translations || product.translations.length === 0
```
michaelbromley: I commented on the interface about the need for the top-level translatable fields (name, description, slug). If they are indeed not essential, then this part could be simplified to just the `map` operation part. Same would go for the other places using the same pattern below.

michaelbromley (commented):
This is very good! I made a bunch of comments, some of which might then require some changes.

The addition of the mainLanguage arg is not a breaking change since it is optional _and_ I would be very surprised if anyone is using that method directly in their custom code.

Re the "fail is not defined" - I think it may be related to https://github.com/facebook/jest/issues/11698

packages/core/src/data-import/providers/import-parser/import-parser.ts
heiko-r: Correct, I'm adding all translations to the `translations` arrays, because that makes the most sense for subsequent importing. I added the whole `mainLanguage` duplication part only to keep `ParsedProduct` etc. compatible with the current implementation. I personally don't need these 'main' translations at all and would be happy to remove those properties along with the `mainLanguage` parameter. But I'm not sure if that would break things for anyone else...

packages/core/src/data-import/providers/import-parser/import-parser.spec.ts
heiko-r: Haha, so true. I found it tedious to check if the snapshot actually contained the intended output. But they do work now, so I feel a bit too lazy to change them now ;-)

packages/core/src/data-import/providers/importer/importer.ts
heiko-r: Yes, getting rid of the whole `mainLanguage` part would simplify things quite a bit. The only worry is backwards compatibility...

packages/core/src/data-import/providers/import-parser/import-parser.ts
michaelbromley: I think we can reasonably consider this an internal API, in the sense that it is not documented and is not exported from the `@vendure/core` package, so if anyone _did_ rely on it, they would need to do a deep import, which is in itself a bit of a signal that it is not a public API.

So I would say it is ok to remove the top-level props that are duplicated, and clean up the rest of the PR accordingly.

heiko-r: Is it safe to put all custom fields, including those that are not of type `localeString`, into `translations.customFields`, and in return, to put `localeString` fields of a random translation into the "main" `customFields` property? I think this should work, but I'm not sure. Otherwise, we would have to keep at least the `customFields` "main" property.

heiko-r: And on second thought, we need to keep the `mainLanguage` parameter after all, in order to be able to set the import language for untranslated CSV files to `ctx.languageCode`.

docs/content/developer-guide/importing-product-data.md
heiko-r: Tried to use too few words. Is the new version more clear?

packages/core/src/data-import/providers/import-parser/import-parser.ts
heiko-r: Kept the `mainLanguage` parameter but removed those top-level translatable fields.

docs/content/developer-guide/importing-product-data.md
michaelbromley: Much better, makes sense now 👍

michaelbromley:
Love all that deleted code in the second commit 😄 

Thank you for your work on this!

PR #1200, opened 2021-10-29
Feature/mollie payment integration

martijnvdbrug:
Initial draft of the Mollie payment integration, including e2e tests.

1. Mollie paymentMethod is created via the admin UI, with a redirectUrl and an apiKey
2. When the storefront calls createPayment, a lookup is done which paymentMethod is being used, based on apiKey.
3. The id of the method is included in the webhookUrl, so that we can retrieve the method+apiKey when a webhook comes in
4. When a webhook comes in, a call to Mollie is made to retrieve payment details
5. If the status is paid, the payment is updated.
6. If any other status, do nothing, because maybe another webhook will come later.

Refunds can be async: For now refund is set as `settled` when we receive anything but 'failed' from Mollie. (So it could still be pending or queued)


packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -0,0 +1,178 @@
...
+    TestOrderWithPaymentsFragment,
+} from '@vendure/core/e2e/graphql/generated-e2e-shop-types';
+import { GET_CUSTOMER_LIST } from '@vendure/core/e2e/graphql/shared-definitions';
+import {
+    ADD_ITEM_TO_ORDER,
+    ADD_PAYMENT,
+    GET_ORDER_BY_CODE,
+    GET_ORDER_BY_CODE_WITH_PAYMENTS,
+} from '@vendure/core/e2e/graphql/shop-definitions';
+import { proceedToArrangingPayment } from '@vendure/core/e2e/utils/test-order-utils';
```
martijnvdbrug: Are these import allowed in test files?

packages/payment-plugin/e2e/payment-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+`;
+
+export const CREATE_PAYMENT_METHOD = gql`
+    mutation CreatePaymentMethod($input: CreatePaymentMethodInput!) {
+        createPaymentMethod(input: $input) {
+            ...PaymentMethod
+        }
+    }
+    ${PAYMENT_METHOD_FRAGMENT}
+`;
```
martijnvdbrug: If imports to other packages in the monorepo are allowed, then these duplicate queries are also not needed.

packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
michaelbromley: You can do it with relative imports, I think that might be safer. But you can also consider duplicating the code in the plugin dir - that is potentially more robust and resistant to changes in core e2e tests.

packages/payment-plugin/e2e/payment-helpers.ts
martijnvdbrug: Based on previous comment: code will stay duplicated to make the plugin more self contained.

packages/payment-plugin/package.json
```diff
@@ -1,30 +1,36 @@
...
+    },
+    "dependencies": {
+        "@mollie/api-client": "^3.5.1"
+    },
+    "devDependencies": {
+        "@types/braintree": "^2.22.15",
+        "@vendure/common": "^1.3.2",
+        "@vendure/core": "^1.3.2",
+        "@vendure/testing": "^1.3.2",
+        "braintree": "^3.0.0",
```
martijnvdbrug, line 32: Why is braintree a dev dependency? Should mollie be a devdependency as well?

packages/payment-plugin/tsconfig.json
```diff
@@ -1,10 +1,10 @@
...
-    "sourceMap": true
-  }
+    "extends": "../../tsconfig.json",
+    "compilerOptions": {
+        "declaration": true,
+        "removeComments": false,
+        "noLib": false,
+        "skipLibCheck": true,
+        "sourceMap": true
+    }
```
martijnvdbrug, line 9: prettier did this

packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
martijnvdbrug: This is fixed.

packages/payment-plugin/package.json
michaelbromley: Yes, make it a dev dependency please. It is this way because if I want to install the payment plugin but only want to use Mollie, I don't also want to install the Braintree (Stripe, Adyen etc) packages too. The docs need to include the instruction to install the required package manually.

packages/payment-plugin/tsconfig.json
michaelbromley: ok, that's no biggie.

packages/payment-plugin/src/mollie/README.md
```diff
@@ -0,0 +1,33 @@
...
+### 1. Add in plugin list in `vendure-config.ts`:
+
+```js
+MolliePlugin.init('https://yourhost.io/');
+```
+
+Pass your publicly available Vendure host to the plugin. This is used by Mollie to let the plugin know when the status
+of a payment changed.
+
+**Use something like `localtunnel` to test on localhost!**
```
michaelbromley: Maybe add a link to the GH repo.

packages/payment-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+export class MolliePlugin {
+    static context = 'MolliePlugin';
+    static host: string;
+
+    /**
+     * Initialize the mollie payment plugin
+     * @param vendureHost is needed to pass to mollie for callback
+     */
+    static init(vendureHost: string): typeof MolliePlugin {
```
michaelbromley: I would recommend using an object here rather than a string, so that in future if other config options need to be exposed, we can easily do so without breaking changes or needing to define overloads. And in that case, use the pattern for defining a symbol for the options so it can be used with the Nest DI as in the [plugin template](https://github.com/vendure-ecommerce/plugin-template/blob/master/src/plugin.ts#L45)

packages/payment-plugin/src/mollie/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### 2. Admin ui
+
+1. Go to the Vendure admin
+2. Go to `settings > paymentMethods` and create a payment method with the name `mollie-payment-YOUR_CHANNEL_TOKEN`. This
```
michaelbromley: Is that name constraint still applicable?

packages/payment-plugin/src/mollie/mollie.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+    init(injector) {
+        paymentMethodService = injector.get(PaymentMethodService);
+    },
+    createPayment: async (ctx, order, amount, args, _metadata): Promise<CreatePaymentResult> => {
+        try {
+            const { apiKey } = args;
+            let { redirectUrl } = args;
+            redirectUrl = redirectUrl.endsWith('/') ? redirectUrl.slice(0, -1) : redirectUrl; // remove appending slash
+            const paymentMethods = await paymentMethodService.findAll(ctx);
+            const paymentMethod = paymentMethods.items.find(pm =>
```
michaelbromley, line 43: This assumes that there will only be 1 PaymentMethod using that particular mollie api key. Is that a safe assumption? Or could there be a scenario with e.g. 2 PaymentMethods using the same api key but with different redirectUrls? In that case it would be safer to check against both the apiKey and redirectUrl.

packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        },
+    };
+    let shopClient: SimpleGraphQLClient;
+    let adminClient: SimpleGraphQLClient;
+    let server: TestServer;
+    let started = false;
+    let customers: GetCustomerListQuery['customers']['items'];
+    let order: TestOrderFragmentFragment;
+    beforeAll(async () => {
+        registerInitializer('sqljs', new SqljsInitializer('__data__'));
```
michaelbromley: Import the `testConfig` constant not from `@vendure/testing`, but from `'../../../e2e-common/test-config';`. Then this call to `registerInitializer` should not be needed, because all initializers for the supported DBs will be implicitly registered. Actually this _should_ already be the case since you are already importing `TEST_SETUP_TIMEOUT_MS` from that file.

michaelbromley (commented):
Looking good so far! Just a few points to look at.

packages/payment-plugin/package.json
martijnvdbrug: Makes sense. I've moved Mollie to devDependencies, and added peerDependencies for both Braintree and Mollie with their current major version: `3.x`

packages/payment-plugin/src/mollie/README.md
martijnvdbrug: Nope, removed it!

packages/payment-plugin/src/mollie/mollie.plugin.ts
martijnvdbrug: Fixed.

packages/payment-plugin/src/mollie/README.md
martijnvdbrug: Done

packages/payment-plugin/src/mollie/mollie.handler.ts
martijnvdbrug: Fair point. In this case the controller only uses the apiKey from the method, so there is no harm in having the wrong paymentMethod as long as it has the same apiKey.

That said, this is a dangerous assumption for the future, so I fixed it :)

packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
martijnvdbrug: Also fixed :ok_hand:

martijnvdbrug:
> Looking good so far! Just a few points to look at.
Ready for re-review, all points should resolved

packages/payment-plugin/src/mollie/README.md
michaelbromley: There's a space before `github` that needs removing.

packages/payment-plugin/src/mollie/mollie.handler.ts
```diff
@@ -0,0 +1,123 @@
...
+        },
+        redirectUrl: {
+            type: 'string',
+        },
+    },
+    init(injector) {
+        paymentMethodService = injector.get(PaymentMethodService);
+        options = injector.get(PLUGIN_INIT_OPTIONS);
+    },
+    createPayment: async (ctx, order, amount, args, _metadata): Promise<CreatePaymentResult> => {
```
michaelbromley: Return type should also include `CreatePaymentErrorResult`, so that in the `catch` block you can return a result with the state `Error`, which is more correct than `Declined`

packages/payment-plugin/src/mollie/mollie.handler.ts
```diff
@@ -0,0 +1,123 @@
...
+                    public: {
+                        redirectLink: payment.getPaymentUrl(),
+                    },
+                },
+            };
+        } catch (err) {
+            Logger.error(err, loggerCtx);
+            return {
+                amount: order.totalWithTax,
+                state: 'Declined' as const,
```
michaelbromley: Change to `Error` and then you can move the `errorMessage` prop to the top level, and it will be available in the standard ErrorResult in the API.

packages/payment-plugin/src/mollie/README.md
```diff
@@ -0,0 +1,31 @@
...
+1. Add the plugin to your VendureConfig `plugins` array:
+
+```js
+MolliePlugin.init({ vendureHost: 'https://yourhost.io/' });
+```
+
+2. In the admin UI set the `redirectUrl`, this is the url that is used to redirect the end-user. I.E. `https://storefront/order`
+3. Set the your Mollie apiKey in the `apiKey` field.
+
+**Use something like [localtunnel](https:// github.com/localtunnel/localtunnel) to test on localhost!**
```
michaelbromley: Add an example:
```shell
npx localtunnel --port 3000 --subdomain my-shop-local-dev
> your url is: https://my-shop-local-dev.loca.lt     <- use this as the vendureHost for local dev.
```

packages/payment-plugin/src/mollie/mollie.handler.ts
```diff
@@ -0,0 +1,123 @@
...
+                metadata: {
+                    errorMessage: err.message,
+                },
+            };
+        }
+    },
+    settlePayment: async (order, payment, args): Promise<SettlePaymentResult> => {
+        return { success: true };
```
michaelbromley, line 87: Could add a comment here that "Logic for settlement is handled by the webhook in mollie.controller.ts"

packages/payment-plugin/src/mollie/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Setup
+
+1. Add the plugin to your VendureConfig `plugins` array:
+
+```js
+MolliePlugin.init({ vendureHost: 'https://yourhost.io/' });
+```
+
+2. In the admin UI set the `redirectUrl`, this is the url that is used to redirect the end-user. I.E. `https://storefront/order`
+3. Set the your Mollie apiKey in the `apiKey` field.
```
michaelbromley: typo "the your"

packages/payment-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    };
+    let shopClient: SimpleGraphQLClient;
+    let adminClient: SimpleGraphQLClient;
+    let server: TestServer;
+    let started = false;
+    let customers: GetCustomerListQuery['customers']['items'];
+    let order: TestOrderFragmentFragment;
+    beforeAll(async () => {
+        const devConfig = mergeConfig(testConfig, {
+            logger: new DefaultLogger({ level: LogLevel.Debug }),
```
michaelbromley: Remove the logger now as it is too noisy in CI

michaelbromley (changes requested):
Great work! Just a few small points to clean up and then ready to merge!

michaelbromley (approved):
🌟

Issue #1202, opened 2021-10-31
InitialData object on importing-product-data.md is not completely based on typescript definition


zmilan:
**Describe the bug**
InitialData object on importing-product-data.md is not completely based on typescript definition. It misses the definition of payment methods and roles.

**To Reproduce**
1. Try to copy-paste example from the documentation into the editor and you would be notified by editor :)

**Expected behavior**
Official examples should work with copy-paste.

**Environment (please complete the following information):**
 - @vendure/core 1.3.0

**Additional context**

PR #1203, opened 2021-10-31
docs(Developer Guide, Importing Product Data): Added example of payment method and roles definition

zmilan:
Added examples of payment method and roles definition. Based on the [issue](https://github.com/vendure-ecommerce/vendure/issues/1202).


michaelbromley:
Thank you!

Issue #1181, opened 2021-10-20
elastic-search: hide customProductMappings from api


Draykee:
**Is your feature request related to a problem? Please describe.**
I want to map and search for some custom products properties without exposing the exact value to the public api.

Usecases:

- A hidden rank that is applied to promote some products
- A hidden product owner id (I never want to expose user ids)
- Exact product locations

**Describe the solution you'd like**
Extend the custom mapping type with a new field:
```TS
type CustomMappingDefinition<Args extends any[], T extends CustomMappingTypes, R> = {
    graphQlType: T;
    hidden?: boolean; // **NEW** ( should default to false )
    valueFn: (...args: Args) => R;
}
```

And inside the `generateCustomMappingTypes` function a filter could just be applied like this:
```TS
const productMappings = Object.entries(options.customProductMappings || {}).filter(([, value]) => !value.hidden);
```

**Describe alternatives you've considered**
I couldn't come up with an alternative that doesn't produce a huge overhead.

Draykee:
@michaelbromley I guess this issue can be closed :)

michaelbromley:
Yes!

PR #1212, opened 2021-11-08
feat(elasticsearch-plugin): Added option to hide indexed fields in api

Draykee:
Added option to hide indexed fields in api #1181

I hope it's okay to already push it to minor branch.


michaelbromley:
Very nice work, thank you!

PR #1218, opened 2021-11-10
fix(docs): Fix GraphQL operation type for eligibleShippingMethods in Shop API guide

vrosa:
`eligibleShippingMethods` is a [query](https://www.vendure.io/docs/graphql-api/shop/queries/#eligibleshippingmethods) and not a mutation as per docs. Currently the link doesn't take the user to the right page.


michaelbromley:
Good catch, thank you!

PR #1222, opened 2021-11-15
Vendure entity events

### LINKED ISSUE #1219: More Events for each entity
I would like to know when a user get's verified. Currently there is no event for this. Having more events can be extremly helpful for analyzing activities.

A event for each entity and every special activity (like coupon code usage, promotions, etc.)
Maybe we should make a list first, but implementing it should be easy.

Overwriting each resolver. But that's a hughe overhead

kevmtt:
A work in progress PR to review changes and disscuss.

### DIFF
--- a/packages/core/src/event-bus/events/asset-event.ts
+++ b/packages/core/src/event-bus/events/asset-event.ts
@@ -1,21 +1,37 @@
-import { RequestContext } from '../../api/common/request-context';
+import { CreateAssetInput, DeleteAssetInput, UpdateAssetInput } from '@vendure/common/lib/generated-types';
+
+import { RequestContext } from '../../api';
 import { Asset } from '../../entity';
-import { VendureEvent } from '../vendure-event';
+import { VendureEntityEvent } from '../vendure-entity-event';
+
+type AssetInputTypes = CreateAssetInput | UpdateAssetInput | DeleteAssetInput;
 
 /**
  * @description
  * This event is fired whenever aa {@link Asset} is added, updated
- * or deleted.
+ * or deleted. The `input` property is only defined for `'created'` & `'updated'` event types.
  *
  * @docsCategory events
  * @docsPage Event Types
+ * @since 1.4
  */
-export class AssetEvent extends VendureEvent {
+export class AssetEvent extends VendureEntityEvent<Asset, AssetInputTypes> {
     constructor(
         public ctx: RequestContext,
-        public asset: Asset,
+        public entity: Asset,
         public type: 'created' | 'updated' | 'deleted',
+        public input?: AssetInputTypes,
     ) {
-        super();
+        super(entity, type, ctx, input);
+    }
+
+    /**
+     * Return an asset field to become compatible with the
+     * deprecated old version of AssetEvent
+     * @deprecated Use `entity` instead
+     * @since 1.4
+     */
+    get asset(): Asset {
+        return this.entity;
     }
 }
--- a/packages/core/src/event-bus/events/asset-event.ts
+++ b/packages/core/src/event-bus/events/asset-event.ts
@@ -9,8 +9,7 @@ type AssetInputTypes = CreateAssetInput | UpdateAssetInput | DeleteAssetInput | 
 
 /**
  * @description
- * This event is fired whenever a {@link Asset} is added, updated
- * or deleted. The `input` property is only defined for `'created'` & `'updated'` event types.
+ * This event is fired whenever a {@link Asset} is added, updated or deleted.
  *
  * @docsCategory events
  * @docsPage Event Types
--- a/packages/core/src/event-bus/events/asset-event.ts
+++ b/packages/core/src/event-bus/events/asset-event.ts
@@ -1,14 +1,15 @@
 import { CreateAssetInput, DeleteAssetInput, UpdateAssetInput } from '@vendure/common/lib/generated-types';
+import { ID } from '@vendure/common/lib/shared-types';
 
 import { RequestContext } from '../../api';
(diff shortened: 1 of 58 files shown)


packages/core/src/event-bus/events/customer-address-event.ts
```diff
@@ -1,21 +1,37 @@
-import { RequestContext } from '../../api/common/request-context';
-import { Address } from '../../entity/address/address.entity';
-import { VendureEvent } from '../vendure-event';
+import { CreateAddressInput, UpdateAddressInput } from '@vendure/common/lib/generated-types';
```
kevmtt, line 1: @michaelbromley Why am I not able to import this? (I get an error from the check-imports script)

michaelbromley: This should not raise an error - the check-imports script only complains if it has been imported from `@vendure/common/src/...` rather than `lib`.

packages/core/src/event-bus/events/asset-event.ts
```diff
@@ -1,21 +1,38 @@
...
+import { VendureEntityEvent } from '../vendure-entity-event';
+
+type AssetInputTypes = CreateAssetInput | UpdateAssetInput | DeleteAssetInput | ID;
 
 /**
  * @description
- * This event is fired whenever aa {@link Asset} is added, updated
- * or deleted.
+ * This event is fired whenever a {@link Asset} is added, updated
+ * or deleted. The `input` property is only defined for `'created'` & `'updated'` event types.
```
michaelbromley: Why is the input not defined for deletion?

packages/core/src/event-bus/events/coupon-code-event.ts
```diff
@@ -0,0 +1,24 @@
+import { ID } from '@vendure/common/lib/shared-types';
+
+import { RequestContext } from '../../api/common/request-context';
+import { VendureEvent } from '../vendure-event';
+
+/**
+ * @description
+ * This event is fired whenever an {@link Asset} is assigned or removed
```
michaelbromley: Update doc block to match event

packages/core/src/event-bus/events/customer-group-entity-event.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @description
+ * This event is fired whenever a {@link CustomerGroup} is added, updated or deleted.
+ * This Entity shall replace the {@link CustomerGroupEvent} later!
+ *
+ * @docsCategory events
+ * @docsPage Event Types
+ * @since 1.4
+ */
+export class CustomerGroupEntityEvent extends VendureEntityEvent<CustomerGroup, CustomerGroupInputTypes> {
```
michaelbromley, line 19: Unfortunate that I used the `CustomerGroupEvent` name already for the old "change" event. Now we have this annoyingly inconsistent name, but I understand why you did it otherwise it is a breaking change.

Could you are a `// TODO: Rename to CustomerGroupEvent in v2` comment so we can remember to fix this in the next major version?

packages/core/src/event-bus/events/history-entry-event.ts
```diff
@@ -0,0 +1,36 @@
...
+type HistoryInput =
+    | {
+          type: HistoryEntryType;
+          data?: any;
+      }
+    | ID;
+
+/**
+ * @description
+ * This event is fired whenever one {@link HistoryEntry} is  is added, updated or deleted.
```
michaelbromley: duplicate "is"

packages/core/src/event-bus/vendure-entity-event.ts
```diff
@@ -0,0 +1,33 @@
+import { RequestContext } from '../api';
+
+import { VendureEvent } from './vendure-event';
+
+/**
+ * @description
+ * The base class for all entity events used by the EventBus system.
+ * <ul>
+ *     <li>For event type `'updated'` the entity should be the one before applying the patch.</li>
```
michaelbromley: You can use the regular markdown syntax for the list here:
```

* For event ...
* For event ...
```

packages/core/src/service/services/customer.service.ts
```diff
@@ -668,15 +673,15 @@ export class CustomerService {
...
             data: { address: addressToLine(createdAddress) },
         });
-        this.eventBus.publish(new CustomerAddressEvent(ctx, createdAddress, 'created'));
+        this.eventBus.publish(new CustomerAddressEvent(ctx, createdAddress, 'created', input));
         return createdAddress;
     }
 
     async updateAddress(ctx: RequestContext, input: UpdateAddressInput): Promise<Address> {
-        const address = await this.connection.getEntityOrThrow(ctx, Address, input.id, {
+        const address = await this.connection.getEntityOrThrow<Address>(ctx, Address, input.id, {
```
michaelbromley: The generic argument here should not be needed. Ditto on line 684, 717, 720.

michaelbromley (changes requested):
This is an absolutely epic PR, thank you so much for all the work you put in on it! Just a few minor points to address. Regarding the Order-related events, I will take a look at that once these changes are merged in.

packages/core/src/service/services/customer.service.ts
kevmtt: That's true. Maybe my IDE was just wrong ^^

packages/core/src/event-bus/events/asset-event.ts
kevmtt: It is, I just forgot that I was able to implement it aswell.

michaelbromley:
Thank you!

Issue #1221, opened 2021-11-14
Payment Method: Remove an argument of an already saved payment method


nemo-js:
If I remove an argument of a custom payment handler, but there are already payment methods (with this handler) saved on the database, I get the following runtime error every time I call addPaymentToOrder:
`TypeError: Cannot read property 'type' of undefined
at PaymentMethodHandler.argsArrayToHash (C:\Users\jimne\Documents\coding\hestafta\av-hestafta-shop\node_modules\@vendure\core\src\common\configurable-operation.ts:412:10)`

I have fixed the error, should I create a PR on the master branch?

PR #1224, opened 2021-11-16
Fix issue #1221,  Remove an argument of an already saved payment method causes runtime error

nemo-js:
Fix issue #1221,  Remove an argument of an already saved payment method causes runtime error


michaelbromley:
Thanks for the fix!

Issue #1220, opened 2021-11-13
elasticsearch-plugin: Allow custom sort parameter input


Draykee:
**Is your feature request related to a problem? Please describe.**
Currently we can only sort by price and name using the elasticsearch plugin.

```GraphQL
"Current definition of sort parameter"
input SearchResultSortParameter {
    name: SortOrder
    price: SortOrder
}
```
Those fields are then hardcoded and parsed into the `sort` field of the elasticseach request body.
```JS
const sortArray = [];
    if (sort) {
        if (sort.name) {
            sortArray.push({
                'productName.keyword': { order: sort.name === SortOrder.ASC ? 'asc' : 'desc' },
            });
        }
        if (sort.price) {
            const priceField = 'price';
            sortArray.push({ [priceField]: { order: sort.price === SortOrder.ASC ? 'asc' : 'desc' } });
        }
    }
```

**Describe the solution you'd like**
I would like to have an option to push additional sort parameters to the `sortArray`. I guess the only way to do that is another option inside the `searchConfig` configuration.

```TS
mapSort?: (input: SearchInput) => string[]; // default = [ ]
 ```

The biggest problem is that we are not able to extend a graphQL input (defined in `packages/core/src/api/schema/common/common-types.graphql`). Maybe a search plugin needs to define those to generate them at runtime?

**Describe alternatives you've considered**
I'm not sure if we are able to apply sorting within the query, but if not there is no alternative.

michaelbromley:
Hi, could this not be solved with the existing [ExtendSearchInputType](https://www.vendure.io/docs/typescript-api/elasticsearch-plugin/elasticsearch-options/#extendsearchinputtype) option, combined with mapQuery?

Draykee:
No this is not possible. 
1. `ExtendSearchInputType` only allows primitive graphql types and not the SortOrder type.
2. `mapQuery` is only used for the `query` property in the elasticsearch request body. I think sort parameters needs to be applied with the `sort` property. (But I'm not 100% sure)

michaelbromley:
OK got it, thanks for clarifying. I'll review the PR later today 👍

Draykee:
Resolved with: https://github.com/vendure-ecommerce/vendure/pull/1230

PR #1230, opened 2021-11-17
feat(elasticsearch-plugin): Added custom sort parameter mapping (#1220)

kevmtt:
Solves #1220 

@michaelbromley Hopefully this can be deployed soon too :)

### DIFF
--- a/packages/elasticsearch-plugin/src/build-elastic-body.ts
+++ b/packages/elasticsearch-plugin/src/build-elastic-body.ts
@@ -2,7 +2,7 @@ import { LanguageCode, LogicalOperator, PriceRange, SortOrder } from '@vendure/c
 import { DeepRequired, ID, UserInputError } from '@vendure/core';
 
 import { SearchConfig } from './options';
-import { CustomScriptMapping, ElasticSearchInput, SearchRequestBody } from './types';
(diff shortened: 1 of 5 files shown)


packages/elasticsearch-plugin/src/build-elastic-body.ts
```diff
@@ -109,14 +109,13 @@ export function buildElasticBody(
...
-    const sortArray = [];
+    const sortArray: ElasticSearchSortInput = [];
     if (sort) {
         if (sort.name) {
             sortArray.push({
                 'productName.keyword': { order: sort.name === SortOrder.ASC ? 'asc' : 'desc' },
             });
-        }
-        if (sort.price) {
+        } else if (sort.price) {
```
michaelbromley: This is a change in behaviour - now sorting by name/price is an exclusive OR. Was there a particular reason for this change? Does it not make sense to allow sorting on name _and_ price?

kevmtt: Oh, no! This was a left-over from a first tryout!
Awesome that you saw this! :)

Draykee:
@michaelbromley I think the failed test is not related to my changes.

PR #1238, opened 2021-11-23
Fixes stream not being instance of ReadStream

HarunKilic:
The stream will not be recognized as ReadStream if used createReadStream from the regular FS package. 
This will fix it.


HarunKilic:
Not sure why thos one failed @michaelbromley

michaelbromley:
It's a random failure that sometimes happens when publishing to Verdaccio in the CI job. Nothing to do with your code. Will merge now, thanks!

PR #1242, opened 2021-11-26
Fixed batch size error on Postgresql

HarunKilic:
When indexing many items with multiple channels and languages, the postgresql will throw an error because of parameter limit to exceed. 
Setting Chunk size will fix the problem.


michaelbromley:
🙏

PR #1274, opened 2021-12-09
fix(admin-ui): Fix tax rate permissions so product variants do not need access to customer groups

vrosa:
As per discussion on Slack:

> `TaxRateEntityResolver.customerGroup()` kicks in to resolve the `TaxRate.customerGroup` field and requires either `ReadCustomer` or `ReadCustomerGroup` permissions. I found it curious that a field resolver needs additional permissions on top of what its parent query (`TaxRateResolver.taxRates()`) does.
Is my assumption to just rely on `ReadProduct` for a full read-only view of products associated to a channel correct or are these added permissions really necessary?

> I think the real issue here is that there is actually no need to even access the `TaxRate.customerGroup` field when listing ProductVariants. I just re-used an existing graphql fragment which is used in the main TaxRate list view. So the solution would be to alter the query to omit that field for the ProductVariants view.

I've leveraged the same convention found in other places, e.g. `GetProductSimple` and `GetProductVariantListSimple`, and created a new query for tax rates that returns less fields - in this case, `customerGroup` - and uses the `Simple` suffix. I couldn't find other examples in the codebases so please let me know if I should take a different approach.

### DIFF
--- a/packages/admin-ui/src/lib/catalog/src/components/variant-price-detail/variant-price-detail.component.ts
+++ b/packages/admin-ui/src/lib/catalog/src/components/variant-price-detail/variant-price-detail.component.ts
@@ -25,7 +25,7 @@ export class VariantPriceDetailComponent implements OnInit, OnChanges {
 
     ngOnInit() {
         const taxRates$ = this.dataService.settings
-            .getTaxRates(999, 0, 'cache-first')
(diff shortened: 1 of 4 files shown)


michaelbromley:
Yes this is good. Good job for taking the time to get familiar with the code base and discern an existing convention 👍

PR #1275, opened 2021-12-09
fix(core): Add Permission.ReadProduct to Allow decorator of TaxCategoryResolver.taxCategories

vrosa:
This is closely related to #1258 and something I missed when testing read-only permissions.

`TaxCategoryResolver.taxCategories()` is also ultimately called as part of the product view flow; in this case, by `ProductDetailComponent`. It already allows `ReadCatalog` so it makes sense to allow `ReadProduct` as well.


michaelbromley:
Good catch 👍

PR #1276, opened 2021-12-09
fix(docs): Fix typos in README and collection-filter link comment

vrosa:
Just fixing a couple doc issues I came across:

`README.md`:
* The default port number for the dev server is 3000 and not 4200 as defined in [shared-constants.ts](https://github.com/vendure-ecommerce/vendure/blob/5f5f767e145938959a0f17d3abe1a25091c6779b/packages/common/src/shared-constants.ts#L5)
* `fragments` as opposed to `fragements`

`collection-filter.ts`
* Link to `default-collection-filters.ts` currently returns a 404. Updating the path to the correct one

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ If you do not specify the `DB` argument, it will default to "mysql".
 
 1. `cd packages/admin-ui`
 2. `yarn start`
-3. Go to http://localhost:4200 and log in with "superadmin", "superadmin"
+3. Go to http://localhost:3000 and log in with "superadmin", "superadmin"
 
 ### Code generation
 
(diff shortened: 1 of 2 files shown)


michaelbromley:
This is good except that port 4200 _is_ correct if you are running the admin UI in dev mode. That is, if you are actively working on the admin ui, making changes  to it, then you want to be running it on port 4200 cos that is the Angular dev server port.

But if you just want to run the prod-mode admin ui, then you use port 3000.

vrosa:
@michaelbromley Sorry, my bad. That's the Admin UI and not the dev server so it's port 4200 indeed. I've updated the pull request.

michaelbromley:
Thank you 👍

Issue #1290, opened 2021-12-15
Admin-UI: Shortcut cmd+u is not working on macOS


Spielboerg:
**Describe the bug**
The shortcut `cmd+u` for displaying all the places the current screen can be extended (New feature in v1.4: https://vimeo.com/656178155#t=235s) does not work on macOS. However, `ctrl+u` works, but ctrl is not the default modifier key for such actions.

**Expected behavior**
`cmd+u` works on mac.

PR #1291, opened 2021-12-15
Fix(admin-ui): cmd+u shortcut on macOS

Spielboerg:
Related issue: #1290 

This fix will allow users to use `ctrl+u` and `cmd+u` on macs.

Documentation: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey


michaelbromley:
Thank you!

Issue #1120, opened 2021-09-27
Reindexing crashes on big installations


Izayda:
**Describe the bug**
Installation has 120k products * 12 channels. While reindexing it generates 4millions of operations.

It has two problems:
1. RAM consumption of both DefaultSearchPlugin and ElasticSearchPlugin
2. ERRCONNRESET on Elastic after sending operations to bulk api 

**To Reproduce**
Steps to reproduce the behavior (needs test):
1. Generate many variants with many channels
2. Try to reindex

**Expected behavior**
Reindex complete successfully with reasonable RAM consumption.

**Environment (please complete the following information):**
 - @vendure/core version: 1.2.0

**Additional context**
From slack: https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1632655423007600

michaelbromley:
Perhaps I have a totally wrong mental model of what is going on here, but the way I read the code, it currently seems to work like this:

1. The `try-catch-finally` block starting on line 241 creates a new index for the reindexing. (242)
2. The original index gets copied over to this new index (265)
3. Point the variants alias at this new index (286)
4. Remove the alias which was pointing to the original index (294)
5. Delete the original index (313)
6. Delete the temp "reindex" alias (324)
7. After all that is done, we run the actual deletions and additions to the index from line 338 onwards. This is where we start populang all the bulk commands into the `operations` array, which gets really big.
8. On line 369 we actually execute these operations on the variant index.
9. Then the function immediately returns.

So given that, I if at line 369 we just update like 10% of the SKUs, then the variants index will be in a partially-complete state until the rest of the batched operations complete. That's what I don't get.

Izayda:
Yeah, you are right. I'm sorry, I was pretty sure, that we implemented "silent" update of variants. But only "silent" index recreation on start of reindex job was implemented here https://github.com/vendure-ecommerce/vendure/pull/996 

But, as i can see, "silent" update of variants can be easily achieved just moving block of operations processing upper.

ttournie:
We also go the issue because we have a huge amount of product in production and get the issue.

I am working on a fix for this issue.

Has a quick fix I tried to simply chunk the operation but I get some error on some chunks about channelID.

The best approche seems to chunk the request to the database (now we get ALL the product at once) and create all the operation related. Instead we could request X amount of  products, create the operation, then send them to ES, and continue for X more.

And Switch the alias when it's done. What do you think ?

[...]

PR #1292, opened 2021-12-15
Fix issue 1120 - Reindexing crashes on big installations

ttournie:
This is a proposition to fix the following issue #1120 

We are currently using this for our own indexing, the goal is to have a more scalable reindex.

### DIFF
--- a/packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
+++ b/packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
@@ -44,6 +44,7 @@ import {
 
 import { createIndices, getClient, getIndexNameByAlias } from './indexing-utils';
 import { MutableRequestContext } from './mutable-request-context';
+const REINDEX_CHUNK_SIZE = 2500;
 
 export const defaultProductRelations: Array<EntityRelationPaths<Product>> = [
     'variants',
@@ -232,7 +233,6 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
         return asyncObservable(async observer => {
             return this.asyncQueue.push(async () => {
                 const timeStart = Date.now();
-                const operations: BulkVariantOperation[] = [];
                 const ctx = MutableRequestContext.deserialize(rawContext);
 
                 const reindexTempName = new Date().getTime();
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thank you for this contribution! I will set aside some time in the coming days to properly review it.

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
```diff
@@ -335,42 +335,97 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
...
+                Logger.verbose(`Will reindex ${totalDeletedProductIds} deleted products`, loggerCtx);
+
+                let deletedProductIds = [];
+                let skip = 0;
+                let finishedDeletedItem = 0;
+                do {
+                    const operations: BulkVariantOperation[] = [];
+                    deletedProductIds = await this.connection
+                        .getRepository(Product)
+                        .createQueryBuilder('product')
```
michaelbromley: `skip` & `take` are not being used in this query, so this part is not currently being batched.

michaelbromley (changes requested):
See the comment re skip/take.

This PR looks good - the only thing is that the batched updates are being performed on the "live" index (the one that the `variants` alias is pointing to). I think this means that the live `search` query will return partial results if the reindex job is in progress. Have you not run into this scenario in your application?

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
ttournie: My bad I have it on my own code, I missed it when creating the PR. I add it

ttournie:
> See the comment re skip/take.
> 
> This PR looks good - the only thing is that the batched updates are being performed on the "live" index (the one that the `variants` alias is pointing to). I think this means that the live `search` query will return partial results if the reindex job is in progress. Have you not run into this scenario in your application?

Well we do not have any problem no.

Has I understand it the reindex index is created by 'cloning' the one that was used (or direclty on it if it fails creating it). So the result are the same has it was before the reindexing at first and then, it gets updated little by little (here by 2500 products) so the first 2500 product will appears before the last 2500 in the search, yes. But I do not see the issue

michaelbromley:
Whether the partial index is an issue I think depends on your needs and also things like the number of channels and languages, which can increase indexing time and leave your storefront longer in a partially-indexed state.

In any case, after you fixed the skip/take part, this looks good to merge and then I will do a few further tests with getting the full index to update transparently.

michaelbromley:
From a recent Slack conversation:

> Maybe one improvement for the full reindex could be to recreate the indice from scratch (not from the current index).
then index the product that needs to be indexed/variant
then switch from the old index to the new.
Right now there is a lot of 'cleaning' operation when reindexing (soft deleted product for exemple) and has the application lives it can have a lot's of old products that should not need to be an operation (edited) 

> I made some changes in this regard and it's much better, I might update my PR

I'll wait on the review to give you time to add these improvements to the PR.

michaelbromley:
I did some memory profiling on this:

27,255 product variants, running `reindex` mutation.

image


blue lines = existing ElasticSearchPlugin
green lines = ElasticSearchPlugin with the changes in this PR
orange line = DefaultSearchPlugin for comparison

michaelbromley:
Rebased and merged in https://github.com/vendure-ecommerce/vendure/commit/bce86f6259daea74e50a29022eab0cb10c134d43

PR #1297, opened 2021-12-16
Add order event

davidspiessens:


michaelbromley:
Thanks, this looks good. Could you make the PR against the `minor` branch though, because this will be a new feature for the next minor version.

michaelbromley:
Thanks, and are you able to clean up the git history so that we just have the commits related to your change? Maybe you need to rebase? Let me know if you need help with this part.

davidspiessens:
If you do a squash and merge, this should be fixed?

Op ma 20 dec. 2021 10:24 schreef Michael Bromley ***@***.***>:

PR #1299, opened 2021-12-19
fix(server): Changed datatype for data in job_record to longtext

### LINKED ISSUE #1298: Job data truncates after large list product variants or channel need to be indexed
**Describe the bug**

I have about 15 channels and 1200 variants, when we add/update new product variants, vendure job kicks in to update index, however whenever we add/update a new channel, system is sending a request to job queue to update index for that specific channel and job data can be huge, when i use default job queue, job data gets truncated and resulted in a malformed JSON string.

Here's the article for different simple column types for text, right now, job data column is using `simple-json` and translated to `text` in MySQL, i wonder if we could change this to `longtext` which can store more strings hence enabling default job queue to handle larger job data

https://chartio.com/resources/tutorials/understanding-strorage-sizes-for-mysql-text-data-types/

**To Reproduce**

**Expected behavior**

 - @vendure/core version:
 - Nodejs version
 - Database (mysql/postgres etc):

**Additional context**

simpian:
This is to close #1298 

Changed `simple-json` to `longtext` to take more characters in job data

### DIFF
--- a/packages/core/src/plugin/default-job-queue-plugin/job-record.entity.ts
+++ b/packages/core/src/plugin/default-job-queue-plugin/job-record.entity.ts
@@ -13,7 +13,7 @@ export class JobRecord extends VendureEntity {
     @Column()
     queueName: string;
 
-    @Column('simple-json', { nullable: true })
+    @Column('longtext', { nullable: true })
     data: any;
 
     @Column('varchar')


simpian:
just realized that sqlite and postgres don't support longtext since text and longtext are considered the same in both DB.

michaelbromley:
Hi, thanks for the contribution. A couple of issues here:

1. Changing the data type will cause a schema change, which is a breaking change. We can't do this change by default, it would need to be behind an opt-in flag.
2. The 'simple-json' type tells TypeORM to apply automatic JSON stringify/parse on the data. Just changing to `longtext` or any other text type would probably break something.

simpian:
got it

Issue #1303, opened 2021-12-20
Verified user gets `unregistered` using (refreshCustomerVerification) 


HarunKilic:
**Describe the bug**
When an user is verified (registered), they will be unverified (unregistered) using the `refreshCustomerVerification` mutation. 

**To Reproduce**

1. Set requireVerification to true
2. Register an user (can also be done at admin panel).
3. Verify the user and confirm in admin panel.
4. Use `refreshCustomerVerification` with the newly verified user. 
5. The user should now be `unregistered`

**Expected behavior**
An verified user should stay verified. Unless an admin wants to unverify which is uncommon.

**Environment (please complete the following information):**
 - @vendure/core version: 1.4.1

PR #1304, opened 2021-12-20
Fixed verification

HarunKilic:
Fixes the issue: #1303


HarunKilic:
This has a funny error again :)

michaelbromley:
The CI failure can be ignored. Unfortunately the jobs sometimes randomly fail, but it is clear that your change is fine since only 1 out of 9 failed (and they don't even touch the code change you made).

PR #1318, opened 2021-12-30
fix(admin-ui): Changes file name from 'pt_PT' to 'pt' to be compatible with Intl API

filipemoraes:
Changes file name from 'pt_PT' to 'pt' to be compatible with Intl API and fix Portuguese translations.


michaelbromley:
Thanks! I dug a bit further into this and found that we don't need to rename the translation files, but instead just make sure we always pass valid values into the Intl API. Portuguese is a kind of edge-case because the region (PT or BR) is part of the translation file, which is not the case for any of the other supported languages.

PR #1323, opened 2022-01-05
fix(core): Order promotions passing state correctly

### LINKED ISSUE #1322: Promotion condition won't pass a state to the associated promotion action
**Describe the bug**
Promotion condition won't pass a state to the associated action

**To Reproduce**
1. Create a promotion condition
2. Pass some data as a state (object, e.g. `{ foo: 123 }`)
3. Try to use this state in an associated promotion action (4th parameter)
4. Notice `{ '0': undefined }` in place of expected state

**Expected behavior**
`{ foo: 123 }` should be passed to the action's state

**Additional context**
Looking at the code the cause is somewhat obvious, the action's state matcher uses Object.keys for a conditions' array which messes up `pick()` function. Proposing PR soon.

hzub:
fixes #1322 

Looking at the code the cause is somewhat obvious, the action's state matcher uses `Object.keys` for a conditions' array which messes up `pick()` function. This code reverts to using `map()` with code.

### DIFF
--- a/packages/core/src/config/promotion/promotion-action.ts
+++ b/packages/core/src/config/promotion/promotion-action.ts
@@ -309,7 +309,7 @@ export class PromotionOrderAction<
 
     /** @internal */
     execute(ctx: RequestContext, order: Order, args: ConfigArg[], state: PromotionState) {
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thank you!

PR #1334, opened 2022-01-10
POC: Plugin schema generator and api module schema stitching

rem4ik4ever:
I was initially thinking of using TypeGraphQL, but figures out that `@nest/graphql` plugin actually supports similar functionality. This is POC that was started in Vendure Slack and further moved to Vendure discussions [here](https://github.com/vendure-ecommerce/vendure/discussions/1326)

**Why?**
As I was working on a custom plugin I noticed how cumbersome the process of defining entities and resolver and then duplicating them in `api-extension` file in `gql` format. Which kind of puts developer in position to make sure that both are correctly written. 

**What are the changes?**
This POC introduces few things: 
1.`yarn generate` command from plugin root will generate schema based on entities and resolver defined in plugin
2. Instead of extending base schema when building API for plugins in `@vendure/core`, it will use schema stitching from `@graphql-tools/stitch`. It will take care of extending Query and Mutation type as well as resolve duplicate definition of scalar types 
```gql
""" 
will be merged and deduped
""" 
scalar DateTime 

""" 
No need to write extend
"""
extend type Query {} 
extend type Mutation {}
```

Would love to hear some feedback. Thanks


packages/core/src/api/config/configure-graphql-module.ts
```diff
@@ -139,10 +148,29 @@ async function createGraphQLOptions(
...
+                    })
+                    .filter(e => notNullOrUndefined(e.schema))
+                    .map(e =>
+                        makeExecutableSchema({
+                            typeDefs: e.schema as DocumentNode,
+                            resolvers: e.resolvers as any,
+                        }),
+                    ),
+            ],
+        });
```
rem4ik4ever, line 173: Here is where all schema stitching magic happens.

packages/core/src/plugin/type-gql-plugin-test/entitiy/node.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    @Field(() => GID)
+    id: ID;
+
+    @Field()
+    createdAt: Date;
+
+    @Field()
+    updatedAt: Date;
+}
```
rem4ik4ever, line 20: BaseNode is exactly the same as VendureEntity with only difference that it defines GraphQL fields.

packages/core/src/plugin/type-gql-plugin-test/entitiy/pagintedList.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+        // here we use the runtime argument
+        @Field(type => [TItemClass])
+        // and here the generic type
+        items: TItem[];
+
+        @Field(type => Int)
+        itemsCount: number;
+    }
+    return PaginatedResponseClass;
+}
```
rem4ik4ever, line 21: Basic PaginationList implementation

packages/core/src/plugin/type-gql-plugin-test/generate.ts
```diff
@@ -0,0 +1,17 @@
...
+async function generateSchema() {
+    const app = await NestFactory.create(GraphQLSchemaBuilderModule);
+    await app.init();
+
+    const gqlSchemaFactory = app.get(GraphQLSchemaFactory);
+    const schema = await gqlSchemaFactory.create([CarResolver]);
+    fs.writeFileSync(__dirname + '/schema.graphql', printSchema(schema));
+}
+
+generateSchema();
```
rem4ik4ever, line 17: Example script to generate `schema.graphql` file for plugin in the root of the plugin folder

packages/core/src/plugin/type-gql-plugin-test/schema.graphql
```diff
@@ -0,0 +1,21 @@
...
+scalar DateTime
+
+type Query {
+  cars: PaginatedResponseClass!
+}
+
+type PaginatedResponseClass {
+  items: [Car!]!
+  itemsCount: Int!
+}
```
rem4ik4ever, line 21: Example of generated schema

packages/core/src/plugin/type-gql-plugin-test/type-gql-test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { VendurePlugin } from '..';
+
+import { CarResolver } from './api/car.resolver';
+import CarEntity from './entitiy/car.entity';
+
+const path = __dirname + '/schema.graphql';
+const schemaFile = loadSchemaSync(path, { loaders: [new GraphQLFileLoader()] });
+const schema = gql`
+    ${printSchema(schemaFile)}
+`;
```
rem4ik4ever, line 19: This snippet Imports schema file to VendurePlugin

michaelbromley:
I've just spent some time looking through this POC. It is quite exciting! I have some questions:

1. I'm not really clear on the differences of `extendSchema()` which we currently use, and `stitchSchemas()`. Is it just needed because `gqlSchemaFactory.create([...])` returns a `GraphQLSchema` object rather than a `DocumentNode`?
2. If we used a solution like this, I'd want to get rid of the step where you need to run a "generate" script to output the schema file and then read it in again. I wonder if there is a way to do this at bootstrap-time without needing to output a file, but to keep everything in-memory? I'm not sure though whether that would have an adverse effect of cold start time. But my thinking is, if we are going to try to improve developer ergonomics for this part, we should make it _seamless_ and not ask the developer to worry about extra steps.
3. Would it be possible to define a field on `Car` that is only visible in the Admin API? That would certainly be needed.

rem4ik4ever:
> I've just spent some time looking through this POC. It is quite exciting! I have some questions:
> 
> 1. I'm not really clear on the differences of `extendSchema()` which we currently use, and `stitchSchemas()`. Is it just needed because `gqlSchemaFactory.create([...])` returns a `GraphQLSchema` object rather than a `DocumentNode`?
> 2. If we used a solution like this, I'd want to get rid of the step where you need to run a "generate" script to output the schema file and then read it in again. I wonder if there is a way to do this at bootstrap-time without needing to output a file, but to keep everything in-memory? I'm not sure though whether that would have an adverse effect of cold start time. But my thinking is, if we are going to try to improve developer ergonomics for this part, we should make it _seamless_ and not ask the developer to worry about extra steps.
> 3. Would it be possible to define a field on `Car` that is only visible in the Admin API? That would certainly be needed.

Hey Michael! 
Thanks for having a look. 
1. Partially yes, converting from `GraphQLSchema` to `DocumentNode` the other part of the problem is that `generate` commands generates schema as is there is no `Query` or `Mutation` types to extend, resulting in schema error `You cant have same type twice` 
2. For sure, i'm trying to figure out a way to make sure that there are less manual steps to make this work. No clear solution for now, but thanks for suggestion attaching generate process to bootstrap can work.
3. That I didn't take into consideration. I'll play around to see how we can achieve this. 

I'll put some time later  next week to address those.

jbeckton:
Not too long ago I was working with NexusJS and one thing I thought was slick was that you could pipe schema generation right into Apollo Server at run time from typescript files, rather than generating the .graphql text files. 

Technically speaking, if Vendure used NexusJS in place of the NestJS GraphQL mechanism then all the Vendure plugin developers need to do is make their types available to the vendure core at runtime.

https://nexusjs.org/docs/api/make-schema#makeschema

jbeckton:
Something else to consider. NestJS has an 'autoSchema' feature.

`The autoSchemaFile property value is the path where your automatically generated schema will be created. Alternatively, the schema can be generated on-the-fly in memory. To enable this, set the autoSchemaFile property to true:`

````
GraphQLModule.forRoot<ApolloDriverConfig>({
  driver: ApolloDriver,
  autoSchemaFile: true,
}),
````

https://docs.nestjs.com/graphql/quick-start#code-first

using the 'include' property you can tell NestJS which modules to scan for gql, maybe the plugin modules can be passed to this?

````
GraphQLModule.forRoot({
  include: [CatsModule],
}),
````

I have never used this feature of NestJS so I don't know what kind of challenges it may pose for Venture, I'm just throwing some ideas out there. But maybe Vendure plugin developers will not have to write the schema extensions if the api schemas are not generated until run time along with plugin schemas?

Issue #1350, opened 2022-01-17
Add a configurable prefix to job queue names


skid:
In cloud environments, we might use a service to handle our job queues (Amazon SQS / Google Cloud Tasks). If we have multiple deployments on the same account / project, job queues will collide - and multiple environments will end up using the same job queue. This can be easily solved if we had a unique prefix added to the job queue names for each deployment.

PR #1351, opened 2022-01-17
Jobqueue prefix

skid:
Adds a `prefix` option to the jobQueue configuration. Solves: https://github.com/vendure-ecommerce/vendure/issues/1350


skid:
Wrong target.

PR #1373, opened 2022-01-25
refactor(core): Allow to mock self-refreshing cache for channels and zones

ashitikov:
Currently there is no way to mock caches on `ChannelService` and `ZoneService`.
This pull requests adds `createCache()` function to allow developers override and mock it if needed.
Also `initZones()` and `initChannels()` now ensures that cache exists, and do not recreate it on each call.


michaelbromley:
Thank you for the contribution!

PR #1386, opened 2022-01-28
fix(admin-ui): Allow serving admin-ui on custom domains

ashitikov:
Suppose you have domain vendure.local and you resolve this domain to 127.0.0.1 via /etc/hosts.
Using `compileUiExtensions` with enabled dev mode you will be able to compile and serve, but accessing `https://vendure.local:4002/admin` you will get `Invalid host header` error.

Supplying parameter `disableHostCheck` to `angular.json` fix this issue.
Serving happens only on dev mode, so no strict safety is needed.


michaelbromley:
Hi,
Thanks for the PR and for bringing this issue to my attention. I did some more research on this issue and I'm not comfortable with disabling the host check by default. I know the risk is probably fairly minimal, but [this thread](https://github.com/webpack/webpack-dev-server/issues/887) has a good explanation of the attack that this check is designed to mitigate.

I would rather solve this by adding a new option to the AdminUiPlugin init options which allows you to pass arbitrary arguments to the `compileUiExtensions` function which will then get passed to `ng serve` or `ng build`. This would allow you to pass the `--disable-host-check` arg, and also make it easy to handle other issues that people run into with the Angular config.

ashitikov:
Good catch.

I did some improvements as you pointed, but unfortunately unable to test this on dev-server due to
```
[run:server] Error: Cannot find module '@vendure/admin-ui'
[run:server] Require stack:
[run:server] - /Users/me/vendure/packages/ui-devkit/compiler/scaffold.js
```
lerna bootstrap + lerna build is okay.

This happens each time I'm trying to use `compileUiExtensions` on dev config.
@michaelbromley Do you have any ideas why this happens? Can you review these changes please?

packages/ui-devkit/src/compiler/types.ts
```diff
@@ -218,6 +227,17 @@ export interface UiExtensionCompilerOptions {
      * @default 4200 | undefined
      */
     watchPort?: number;
+
+    /**
+     * @description
+     * Allows to pass additional arguments for watcher (in development mode) and compiler (in production mode).
+     *
```
michaelbromley, line 244: I'd make it explicit that these are being passed to the `ng build` or `ng serve` commands and include a link:

---
```
Additional command-line arguments which will get passed to the [ng build](https://angular.io/cli/build) 
command (or [ng serve](https://angular.io/cli/serve) if `devMode = true`).
```

Also, please add the following tag to show when this was introduced:

```
@since 1.5.0
```

michaelbromley (changes requested):
Thanks for this! Just a couple of changes in the documentation is all.

Also, can you please make this PR against the `minor` branch? Since it is a new feature, per semver it will go in the next minor release (1.5.0).

Regarding problems getting `compileUiExtensions()` to work in the dev server - yes I am aware of this and I believe it relates to the way Lerna symlinks the dependencies. For testing that part, I usually use Verdaccio to publish a local version (there's even an npm script for that in the root package.json) and then install it in a different location and test that way. A bit slow I know, but it works.

ashitikov:
Hey, fixed conflicts. Made some changes to pass arguments in more elegant way

michaelbromley:
Thanks!

PR #1362, opened 2022-01-19
Fix/long nav items

Spielboerg:
Related issue: https://github.com/vendure-ecommerce/vendure/issues/1361

Commit 2d260aeaf6b5845f6851ca582f5ff1c42ff5ff50 fixes the overflowing:  
margin

This leads to a left shift of the elements (I can also change that if you want the same look as before):  
navigation 

Commit 11eadc0ef056bb120fb5182e43bed02eb329fd46 improves the alignment of the navigation elements:  
navigation-item

The second change results in minimally higher navigation elements. I can also try to restore the original look here.

PR #1390, opened 2022-01-31
fix(admin-ui): Fix hyphenation of long words

Spielboerg:
@michaelbromley I found another problem related to the navigation elements that I have missed in my last pull request (#1362). Here is an example based on a ridiculously long word (left before the fix, right after the fix):

hyphenation

Mobile:  
Bildschirmfoto 2022-01-31 um 17 41 18@2x

I had to change the `lang` attribute of the `html` element to match the current language code to make `hyphens` work correctly (https://css-tricks.com/almanac/properties/h/hyphenate/):

> Note that hyphens is language-sensitive. Its ability to find break opportunities depends on the language, defined in the lang attribute of a parent element. Not all languages are supported yet, and support depends on the specific browser.

### DIFF
--- a/packages/admin-ui/src/lib/core/src/components/main-nav/main-nav.component.scss
+++ b/packages/admin-ui/src/lib/core/src/components/main-nav/main-nav.component.scss
@@ -18,6 +18,7 @@ nav.sidenav {
     }
     .nav-group-header {
         margin: 0;
+        line-height: 1.2;
     }
     .nav-link {
         display: inline-flex;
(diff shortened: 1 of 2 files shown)


Spielboerg:
The failed builds are probably related to this issue: https://github.com/angular/angular/issues/20351

What do you think, should I add `"strictMetadataEmit": true` to the `tsconfig.json` or `tsconfig.app.json` or add the suppression `/** @dynamic */`?

michaelbromley:
Hi, thanks for this fix! We can always rely on the German language to provide us with the edge-cases for the UI 😆 

I think I'd rather use a `@dynamic` annotation, just because the effect is more localized. I don't fully understand all the implications of changing the `strictMetadataEmit` setting, but I'm more happy to change a setting for a single file than the whole project.

michaelbromley:
Dankeschön!

Issue #1377, opened 2022-01-25
Error with internal customField on Address type


thewebkit:
**Describe the bug**
Hello There ! I'm getting an error when I try to add an internal customField on the Address type 

**To Reproduce**
1. Add a custom field with `public: false, internal: true` to the Address type in `vendure-config.ts` :
```
Address: [ { name: 'some_custom_name', type: 'int', public: false, internal: true } ]
```
2. Run `ts-node migration.ts generate ...` & `ts-node migration.ts run`
4. Run `yarn start` and get the following error :

```bash
[run:*server] [ExceptionHandler] Unknown type "AddressCustomFields". Did you mean "CustomField"?
[run:*server] Error: Unknown type "AddressCustomFields". Did you mean "CustomField"?
[run:*server]     at assertValidSDLExtension (...\graphql\validation\validate.js:124:11)
[run:*server]     at Object.extendSchema (...\graphql\utilities\extendSchema.js:77:43)
[run:*server]     at Object.addGraphQLCustomFields (...\@vendure\core\src\api\config\graphql-custom-fields.ts:216:2)
[run:*server]     at buildSchemaForApi (...\@vendure\core\src\api\config\configure-graphql-module.ts:147:18)
[run:*server]     at async createGraphQLOptions (...\@vendure\core\src\api\config\configure-graphql-module.ts:94:25)
```

**Expected behavior**
Since the field is internal and not public, it should not be picked up by the `addGraphQLCustomFields` function

**Environment (please complete the following information):**
 - @vendure/core version: `1.4.5`
 - Nodejs version : `14.17.0`
 - Database (mysql/postgres etc): `sqlite`
 
 
 I've tested with two customFields : if one of them is public, the error does not occur. So my guess is we have to check if all fields are private (internal or !public). I'll look into and try to make a PR if I can !

michaelbromley:
Hi @thewebkit - just checking whether you plan to PR for this issue? If not, I can handle it and get it into the next patch release this week.

thewebkit:
Hey there @michaelbromley ! 

Sorry for the delay; but I had so much going on this week that it went out my mind. 

I've just created the PR; It's my first time posting a PR in a public repo though, so I hope I didn't made any mistakes !

Cheers!

michaelbromley:
Thank you! Congrats on your first public PR 🥳 

I left a comment - take a look and update if you think it makes sense.

PR #1394, opened 2022-02-03
fix(core): Fix OrderAddress type AddressCustomFields error (#1377)

thewebkit:
Relates to #1377. This commit adds an additional check for public/internal customFields on the `OrderAddress` schema type before adding `AddressCustomFields` to it.

### DIFF
--- a/packages/core/src/api/config/graphql-custom-fields.ts
+++ b/packages/core/src/api/config/graphql-custom-fields.ts
@@ -185,7 +185,10 @@ export function addGraphQLCustomFields(
         }
     }
 
-    if (customFieldConfig.Address?.length) {
+    const publicAddressFields = customFieldConfig.Address?.filter(
+        config => !config.internal && (publicOnly === true ? config.public !== false : true),
+    );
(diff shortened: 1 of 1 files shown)


packages/core/src/api/config/graphql-custom-fields.ts
```diff
@@ -185,7 +185,10 @@ export function addGraphQLCustomFields(
         }
     }
 
-    if (customFieldConfig.Address?.length) {
+    const publicAddressFields = customFieldConfig.Address?.filter(
+        config => !config.internal && (publicOnly === true ? config.public !== false : true),
+    );
+    if (customFieldConfig.Address?.length && publicAddressFields?.length) {
```
michaelbromley, line 191: I think the `customFieldConfig.Address?.length` check can be removed now right? The `publicAddressFields` value would already cover this.

michaelbromley: Actually I'll just refactor this separately, no worries.

PR #1401, opened 2022-02-07
Enhance build.ts script for admin-ui-plugin

deusama:
- Fix `yarn build` if path contains spaces
- Wait until removal of compile dir is done before starting new build


michaelbromley:
Thanks for this! And good catch on the `removeSync`.

What OS are you testing this on, by the way?

deusama:
Currently tried this on macOS 12

michaelbromley:
Ok, thanks. I'll test it locally on Windows and if all is well I think we can assume it is safe to merge 👍

Issue #1087, opened 2021-09-14
Payments plugin


michaelbromley:
All e-commerce apps need to accepts payments of some kind. Right now we don't have any official integrations, so every project has to implement their own. This has the following negative results:

1. Increased time/cost to implement an app on Vendure
2. Duplicated work re-implementing e.g. a Stripe integration that could be shared
3. Reliance on 3rd-party plugins which might not be well-maintained or tested.

**Describe the solution you'd like**
Vendure should ship with support for some of the more popular payment providers. This would be housed in a `@vendure/payments-plugin` npm package, and developed as part of this monorepo.

This package will contain a number of plugins, one for each supported payment provider. 

## Prior Art

* The real-world-vendure repo has a working (but perhaps not full-featured) [Braintree integration](https://github.com/vendure-ecommerce/real-world-vendure/tree/a1e0a5a94f2227f1583a1873a1b7408e1c00f3cc/src/plugins/braintree).

**existing 3rd-party payment plugins**

Provider | Author | Repo | Last updated | notes
----------|--------|---------|--------|----------
Mollie | @martijnvdbrug | [link](https://github.com/martijnvdbrug/pinelab-vendure-plugins/tree/master/packages/vendure-plugin-mollie) | 2 months ago
Mollie | @stefanvanherwijnen | [link](https://github.com/simsustech/vendure-plugin-payment-mollie) | 6 months ago
Braintree | @artcoded-dev  | [link](https://github.com/artcoded-net/vendure-braintree-plugin) | 17 days ago | extracted from real-world-vendure
PayUMoney | @landofcoder | [link](https://github.com/landofcoder/vendure-payumoney-payment-plugin) | 12 months ago
Stripe | @c0ldfront | [link](https://github.com/c0ldfront/stripe-vendure-plugin) | 12 months ago
Stripe | @gaiusmathew | [link](https://github.com/gaiusmathew/stripe-payment-plugin) | 12 days ago | Based on the c0ldfront repo
RazorPay | @Deathwish98 | [link](https://github.com/Deathwish98/vendure-razorpay-payment-plugin) | 2 months ago
Pagar.me | @jonyw4  | [link](https://github.com/jonyw4/vendure-pagarme-plugin) | 10 months ago

## Supported Payment Providers

There are many, many payment providers. We cannot support them all, so we need to carefully choose in order to be able to keep this package maintainable. The criteria for inclusion would be:

[...]

PR #1417, opened 2022-02-10
feat(payments-plugin): Add Stripe integration

vrosa:
Complements #1087.

The goal of this PR is to add Stripe integration to `@vendure/payments-plugin` using the [Custom payment flow ](https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=elements) and [Payment Intents API](https://stripe.com/docs/payments/payment-intents).  A rough, high-level sequence diagram of a successful payment for an order can be seen below, which aids in explaining the key components in more detail:

stripe_plugin

* `stripe.plugin.ts`
Actual plugin class.

* `stripe.resolver.ts`
Contains the resolver for the `createStripePaymentIntent` Shop API mutation, which requires an active order in the current session.

* `stripe.service.ts`
Provides methods to interact with Stripe via the `stripe` Node library, most importantly, the creation of a payment intent for the active order. It also supports creating/linking the Customer to Stripe via the `storeCustomersInStripe` plugin option.

* `stripe.controller.ts`
Contains the webhook endpoint which listens for payment intent status updates from Stripe - success or failure - and ultimately calls `OrderService.addPaymentToOrder`.

* `stripe.handler.ts`
Actual payment method handler for Stripe. `createPayment` always returns a `Settled` payment because the webhook is only called when the payment is already settled in Stripe (in case of success). That is, `OrderService.addPaymentToOrder` should always succeed.

* `raw-body.middleware.ts`
Middleware for the webhook route which adds the raw request body to the incoming message object, which is the [format Stripe requires](https://stripe.com/docs/webhooks/signatures#verify-official-libraries). Credit to [Gab from the Slack channel](https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1625483446188600?thread_ts=1625483114.188400&cid=CKYMF0ZTJ).

I have also updated/fixed a few very minor typos and cosmetic documentation issues.

**Decisions / Caveats**

* As discussed on Slack, API key and Webhook signing secret must be passed in as plugin options as opposed to payment method handler args since both are needed in `stripe.controller.ts` before we have access to any Vendure-specific data (such as channel token).
* I chose to expose the payment intent creation as a Shop API endpoint as opposed to have it as part of the payment method handler's `createPayment` because, as the name implies, it's just an _intent_. Stripe requires a created intent in order to display the payment form, that is, the call to `stripe.createPaymentIntent` must take place as soon as the customer lands on the payment page. I didn't want to "lock" the customer by transitioning the order to `ArrangingPayment` and calling `addPaymentToOrder` at that point because they might change their mind and go back to their shopping cart to add more items. Therefore, this solution updates the order at the very last minute once we get confirmation from Stripe in the webhook. As a consequence, the `ArrangingPayment` state is pretty much ignored and only transitioned to given the order state machine requires so.
* As discussed on Slack, typings for the new Customer custom field `stripeCustomerId` is not working for some reason. I used `any` and added a `TODO` where appropriate.

### DIFF
--- a/packages/payments-plugin/src/stripe/stripe.controller.ts
+++ b/packages/payments-plugin/src/stripe/stripe.controller.ts
@@ -1,4 +1,4 @@
-import { BadRequestException, Controller, Headers, HttpStatus, Post, Req, Res } from '@nestjs/common';
+import { Controller, Headers, HttpStatus, Post, Req, Res } from '@nestjs/common';
 import {
     ChannelService,
     InternalServerError,
--- a/packages/payments-plugin/src/stripe/stripe.controller.ts
+++ b/packages/payments-plugin/src/stripe/stripe.controller.ts
@@ -1,4 +1,4 @@
-import { BadRequestException, Controller, Headers, Post, Req } from '@nestjs/common';
+import { BadRequestException, Controller, Headers, HttpStatus, Post, Req, Res } from '@nestjs/common';
 import {
     ChannelService,
     InternalServerError,
@@ -6,17 +6,22 @@ import {
     Logger,
     Order,
     OrderService,
-    OrderStateTransitionError,
     PaymentMethod,
     RequestContext,
     TransactionalConnection,
 } from '@vendure/core';
+import { OrderStateTransitionError } from '@vendure/core/dist/common/error/generated-graphql-shop-errors';
+import { Response } from 'express';
 import Stripe from 'stripe';
 
 import { loggerCtx } from './constants';
 import { stripePaymentMethodHandler } from './stripe.handler';
 import { StripeService } from './stripe.service';
 import { IncomingMessageWithRawBody } from './types';
+
+const missingHeaderErrorMessage = 'Missing stripe-signature header';
+const signatureErrorMessage = 'Error verifying Stripe webhook signature';
+const noPaymentIntentErrorMessage = 'No payment intent in the event payload';
 
 @Controller('payments')
 export class StripeController {
@@ -31,28 +36,34 @@ export class StripeController {
     async webhook(
         @Headers('stripe-signature') signature: string | undefined,
         @Req() request: IncomingMessageWithRawBody,
+        @Res() response: Response,
     ): Promise<void> {
         if (!signature) {
-            throw new BadRequestException('Missing stripe-signature header');
+            Logger.error(missingHeaderErrorMessage, loggerCtx);
+            response.status(HttpStatus.BAD_REQUEST).send(missingHeaderErrorMessage);
+            return;
         }
 
         let event = null;
         try {
             event = this.stripeService.constructEventFromPayload(request.rawBody, signature);
         } catch (e: any) {
-            Logger.error(`Error verifying Stripe webhook signature ${signature}: ${e.message}`);
-            throw new BadRequestException('Error verifying Stripe webhook signature');
+            Logger.error(`${signatureErrorMessage} ${signature}: ${e.message}`, loggerCtx);
+            response.status(HttpStatus.BAD_REQUEST).send(signatureErrorMessage);
+            return;
         }
 
         const paymentIntent = event.data.object as Stripe.PaymentIntent;
         if (!paymentIntent) {
-            throw new BadRequestException('No payment intent in the webhook payload');
+            Logger.error(noPaymentIntentErrorMessage, loggerCtx);
+            response.status(HttpStatus.BAD_REQUEST).send(noPaymentIntentErrorMessage);
+            return;
         }
 
         const { metadata: { channelToken, orderCode, orderId } = {} } = paymentIntent;
 
         if (event.type === 'payment_intent.payment_failed') {
-            const message = paymentIntent.last_payment_error && paymentIntent.last_payment_error.message;
+            const message = paymentIntent.last_payment_error?.message;
             Logger.warn(`Payment for order ${orderCode} failed: ${message}`, loggerCtx);
             return;
         }
@@ -90,7 +101,10 @@ export class StripeController {
         });
 
         if (!(addPaymentToOrderResult instanceof Order)) {
-            Logger.error(`Error adding payment to order ${orderCode}: ${addPaymentToOrderResult.message}`);
+            Logger.error(
+                `Error adding payment to order ${orderCode}: ${addPaymentToOrderResult.message}`,
+                loggerCtx,
+            );
             return;
         }
 
@@ -110,11 +124,9 @@ export class StripeController {
     }
 
     private async getPaymentMethod(ctx: RequestContext): Promise<PaymentMethod> {
-        const method = await this.connection.getRepository(ctx, PaymentMethod).findOne({
-            where: {
-                code: stripePaymentMethodHandler.code,
-            },
-        });
+        const method = (await this.connection.getRepository(ctx, PaymentMethod).find()).find(
+            m => m.handler.code === stripePaymentMethodHandler.code,
+        );
 
         if (!method) {
             throw new InternalServerError(`[${loggerCtx}] Could not find Stripe PaymentMethod`);
--- a/packages/payments-plugin/src/stripe/stripe.controller.ts
+++ b/packages/payments-plugin/src/stripe/stripe.controller.ts
@@ -1,0 +1,125 @@
+import { BadRequestException, Controller, Headers, Post, Req } from '@nestjs/common';
+import {
+    ChannelService,
+    InternalServerError,
+    LanguageCode,
+    Logger,
+    Order,
+    OrderService,
+    OrderStateTransitionError,
+    PaymentMethod,
+    RequestContext,
+    TransactionalConnection,
+} from '@vendure/core';
+import Stripe from 'stripe';
+
+import { loggerCtx } from './constants';
+import { stripePaymentMethodHandler } from './stripe.handler';
+import { StripeService } from './stripe.service';
+import { IncomingMessageWithRawBody } from './types';
+
+@Controller('payments')
+export class StripeController {
+    constructor(
+        private connection: TransactionalConnection,
+        private channelService: ChannelService,
+        private orderService: OrderService,
+        private stripeService: StripeService,
+    ) {}
+
+    @Post('stripe')
+    async webhook(
+        @Headers('stripe-signature') signature: string | undefined,
+        @Req() request: IncomingMessageWithRawBody,
+    ): Promise<void> {
+        if (!signature) {
+            throw new BadRequestException('Missing stripe-signature header');
+        }
+
+        let event = null;
+        try {
+            event = this.stripeService.constructEventFromPayload(request.rawBody, signature);
+        } catch (e: any) {
+            Logger.error(`Error verifying Stripe webhook signature ${signature}: ${e.message}`);
+            throw new BadRequestException('Error verifying Stripe webhook signature');
+        }
+
+        const paymentIntent = event.data.object as Stripe.PaymentIntent;
+        if (!paymentIntent) {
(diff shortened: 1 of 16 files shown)


vrosa:
Having a look at the Mollie integration PR I just realised I didn't add e2e tests. I haven't written Vendure integration tests yet but please let me know if that's a blocker and I can update the PR (it will take a little while though).

michaelbromley:
@vrosa This is one of the best PRs I've ever had in terms of documentation. Thank you so much!

I will set aside some time either tomorrow or early next week to properly review and test this out. The e2e tests are not a blocker, but would be nice to have eventually.

vrosa:
@michaelbromley My pleasure!

I started working on e2e tests today and I _might_ have found a couple opportunities for improvements as to how plugin-specific GraphQL operations are tested. If you don't mind, I'd like to create a follow-up PR adding the missing tests as well as serving as a proof of concept.

There's also some functionality that I removed from my own plugin which could benefit others that deal with marketplaces: [transfer to Stripe connected accounts](https://stripe.com/docs/connect/charges-transfers). I was thinking of introducing the capability at a future pull request in case this PR gets eventually approved.

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+        private stripeService: StripeService,
+    ) {}
+
+    @Post('stripe')
+    async webhook(
+        @Headers('stripe-signature') signature: string | undefined,
+        @Req() request: IncomingMessageWithRawBody,
+    ): Promise<void> {
+        if (!signature) {
+            throw new BadRequestException('Missing stripe-signature header');
```
michaelbromley: Have you checked how the logs look when we throw an exception in the controller? Should we also use `Logger.error()` and pass the `loggerCtx` so that the logs show "StripePlugin" to make it explicit?

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        let event = null;
+        try {
+            event = this.stripeService.constructEventFromPayload(request.rawBody, signature);
+        } catch (e: any) {
+            Logger.error(`Error verifying Stripe webhook signature ${signature}: ${e.message}`);
```
michaelbromley: Add `loggerCtx` here too.

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        const paymentIntent = event.data.object as Stripe.PaymentIntent;
+        if (!paymentIntent) {
+            throw new BadRequestException('No payment intent in the webhook payload');
+        }
+
+        const { metadata: { channelToken, orderCode, orderId } = {} } = paymentIntent;
+
+        if (event.type === 'payment_intent.payment_failed') {
+            const message = paymentIntent.last_payment_error && paymentIntent.last_payment_error.message;
```
michaelbromley: Opportunity to use optional chaining? `paymentIntent.last_payment_error?.message`

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+            authorizedAsOwnerOnly: false,
+            channel,
+            languageCode: LanguageCode.en,
+        });
+    }
+
+    private async getPaymentMethod(ctx: RequestContext): Promise<PaymentMethod> {
+        const method = await this.connection.getRepository(ctx, PaymentMethod).findOne({
+            where: {
+                code: stripePaymentMethodHandler.code,
```
michaelbromley: This does not look sound to me. It is assuming that the admin will name the PaymentMethod "stripe" but what if they name it "stripe payment" (or whatever) instead?

We need a solid way to get a reference to the actual PaymentMethod. E.g. we could pass it as an argument to the `createStripePaymentIntent` mutation, and then store it in the PaymentIntent metadata along with the orderId etc.

(I also noticed that we make the exact same mistake in the Braintree plugin, so I'll need to fix that too)

packages/payments-plugin/src/stripe/stripe.service.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        if (!order || !order.customer) {
+            // This should never happen
+            return undefined;
+        }
+
+        const { customer } = order;
+
+        // TODO: fix typings
+        if ((customer.customFields as any).stripeCustomerId) {
```
michaelbromley: Spent a bit of time trying to get this to work but also to no avail!

I found this issue: https://github.com/microsoft/TypeScript/issues/46617 which _might_ be related.

michaelbromley (changes requested):
Really great work! See a few comment I left. Main one is that we should pass the payment method code to the payment intent mutation, so that we can be sure we are selecting the correct method later in the flow.

packages/payments-plugin/src/stripe/stripe.controller.ts
vrosa: The logs show the error message and the stack trace - I'm using the `DefaultLogger` below but the same info should be displayed by other custom loggers as well:

image

This is due to `ExceptionLoggerFilter`, [which automatically handles](https://github.com/vendure-ecommerce/vendure/blob/53babf654176d5d88cd7c1c7f88be98f8dd31a57/packages/core/src/api/middleware/exception-logger.filter.ts) these exceptions and log their data.

Compared to using `Logger.error()` and setting the status code in the response instead, we basically have the same output but no stack trace in the logs:

```typescript
@Post('stripe')
async webhook(
    @Headers('stripe-signature') signature: string | undefined,
    @Req() request: IncomingMessageWithRawBody,
    @Res() response: Response
): Promise<void> {
    if (!signature) {
        Logger.error('Missing stripe-signature header');
        response.status(HttpStatus.BAD_REQUEST).send('Missing stripe-signature header');
        return;
    }

    // ...
}
```

image

Should we stick to the former as it's more concise and generates more log data? If that's the case, I'll update the message to include the logger context:

```javascript
throw new BadRequestException(`[${loggerCtx}] Missing stripe-signature header`);
```

packages/payments-plugin/src/stripe/stripe.controller.ts
vrosa: Good catch.

packages/payments-plugin/src/stripe/stripe.service.ts
vrosa: That was a good lead. I managed to get it working using the full path to `CustomerCustomFields` in both Braintree and Stripe plugins.

* `stripe/types.ts`:
```typescript
import '@vendure/core/dist/entity/custom-entity-fields';

declare module '@vendure/core/dist/entity/custom-entity-fields' {
    interface CustomCustomerFields {
        stripeCustomerId?: string;
    }
}
```

* `braintree/types.ts`:
```typescript
import '@vendure/core/dist/entity/custom-entity-fields';

// ...

declare module '@vendure/core/dist/entity/custom-entity-fields' {
    interface CustomCustomerFields {
        braintreeCustomerId?: string;
    }
}
```

Intellisense support in VSCode is a bit flaky but it works and builds fine.

packages/payments-plugin/src/stripe/stripe.controller.ts
vrosa: You're right. The code above worked in my tests because yeah, I named the payment method "Stripe" as well.

To be honest, I pretty much copied and pasted the code from the Braintree plugin but in fact I was thinking of retrieving the payment method whose **_handler code_** equates to "stripe" since that never changes. What if we used the query below instead to pick the right method?

```typescript
const method = (await this.connection.getRepository(ctx, PaymentMethod).find())
  .find(method => method.handler.code === stripePaymentMethodHandler.code);
```

From my understanding, you cannot filter, i.e use `where`, `simple-json` columns such as `method.handler` so we'd have to retrieve all payment methods first and then pull the Stripe one. I tested this approach and it works. I believe it has 2 benefits over taking payment method as a parameter from the Storefront:
1. It offers a simpler API
2. The Storefront doesn't need to know about a piece of information that might change at any time, for instance, if an Admin updates the payment method name and code on the Admin UI. Even if `eligiblePaymentMethods` is called, the Storefront would still need to understand how to select Stripe

I'm fine either way but please let me know what you think.

packages/payments-plugin/src/stripe/stripe.controller.ts
michaelbromley: You can pass a 3rd argument to `Logger.error()` for the stack trace like this:
```ts
Logger.error('Missing stripe-signature header', loggerCtx, e.stack);
```

In this case, we don't have an existing error object, but you could do this I think:
```ts
const message = 'Missing stripe-signature header'
Logger.error(message, loggerCtx, new Error(message).stack);
```

Worth trying out to see if it preserves the same level of information as just throwing. If so, it's preferable because it makes the logs more uniform, thus easier to parse.

packages/payments-plugin/src/stripe/stripe.controller.ts
michaelbromley: Yes, your idea is better for the reasons given. Let's go with that 👍

packages/payments-plugin/src/stripe/stripe.service.ts
michaelbromley: Great that you got it working! In that case, let's add a comment linking to that TS issue explaining why we are using a deep import, otherwise someone might "clean it up" in the future and break things without knowing why.

packages/payments-plugin/src/stripe/stripe.controller.ts
vrosa: > it makes the logs more uniform, thus easier to parse

Gotcha. In fact, the stack trace is not important as all we need is the logger context and the message to see what went wrong. I'm updating it so the output is as follows:

image

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+    InternalServerError,
+    LanguageCode,
+    Logger,
+    Order,
+    OrderService,
+    PaymentMethod,
+    RequestContext,
+    TransactionalConnection,
+} from '@vendure/core';
+import { OrderStateTransitionError } from '@vendure/core/dist/common/error/generated-graphql-shop-errors';
```
vrosa, line 13: I had to use deep import here because the `OrderStateTransitionError` class exported by `@vendure/core` comes from the [Admin API generated types](https://github.com/vendure-ecommerce/vendure/blob/e227463a7f78bdb0fed51964ef2e7c9eb676a057/packages/core/src/common/index.ts#L6) whereas the error thrown by `OrderService.transitionToState` is a [Shop API generated type](https://github.com/vendure-ecommerce/vendure/blob/e227463a7f78bdb0fed51964ef2e7c9eb676a057/packages/core/src/service/services/order.service.ts#L68).

packages/payments-plugin/src/stripe/stripe.controller.ts
vrosa: I haven't updated the Braintree plugin to use the same logic - even though it's a simple change - as it would require testing, which I'm not able to do.

michaelbromley: That's fine, I'll handle that part.

packages/payments-plugin/src/stripe/stripe.controller.ts
michaelbromley: Oh that's such a pain with those issues, I've run into that stuff myself. OK, this should be fine anyway - since it's in the monorepo, any refactors to core will trigger an error here that can be fixed at the same time.

Issue #1127, opened 2021-09-30
Canceling of job doesn't work


Izayda:
**Describe the bug**
After starting reindexing i cancel job. But i can see job is still processed.
Video https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1632919005096600
It was on redis job queue. But probably bug also exists in default job plugin.

**To Reproduce**
1. Run a job
2. Cancel it
3. See logs that job is still processed

**Expected behavior**
Job is successfully canceled.

**Environment (please complete the following information):**
 - @vendure/core version: 1.2.2
 - Database (mysql/postgres etc): postgresql

michaelbromley:
For BullMQ, there is an open issue to allow cancellation of running jobs: https://github.com/taskforcesh/bullmq/issues/632

I will check into the DefaultJobQueuePlugin.

michaelbromley:
Ok, so on further investigation there is no easy way to cancel a running job, since the `process` function is Promise-based. As mentioned in the lined BullMQ issue above, there is the possibility of using the AbortController API but this is only stable in Node.js v16+, and I cannot see that it is natively integrated with Promises.

So one possible way to "cancel" a `process` function is to query the job state, and return from the function is the state is `CANCELLED`. Example:

```ts
process: async job => {
    let progress = 0;
    while (progress < 100) {

        if (job.state === JobState.CANCELLED) {
          // manually abort if job is cancelled
          return;
        }

        await new Promise(resolve => setTimeout(resolve, job.data.intervalMs));
        progress += 10;
        job.setProgress(progress);
    }
    Logger.info(`Completed job ${job.id}`);
    return 'Done!';
},
```

The problem is that to make all jobs cancellable, we would have to add such logic into all existing jobs, which in many cases would require quite a bit of refactoring.

Izayda:
Hmmm, yeah, no easy way. Is there problem in default job plugin?

michaelbromley:
In the default plugin, the job will appear as CANCELLED, but if the `process` function has started, it will continue until completion anyway. This is a fundamental issue with Promises unfortunately.

PR #1420, opened 2022-02-10
Fix for job cancellation issue #1127

carathorys:
I think this was the root cause why sometimes I were not able to cancel some jobs.

I've created an example job which was running in `setInterval()` _(not included here)_. If I subscribe to the job's updates, sometimes it doesn't gave me back the job state after I cancel it. 
I digged in the source, and find this where I were able to fix the issue for myself. As I've seen that someone else also opened an issue for this, I made the PR. 

If you need anything else, please let me know!


michaelbromley:
Thank you! I'm not totally sure whether this is exactly the same issue as #1127, since this is specific to subscribing to jobs. But a good improvement nonetheless!

PR #1430, opened 2022-02-16
fix doc for OrderItemPriceCalculationStrategy

christopheblin:
as seen in slack, I updated the doc to list all mutations that trigger OrderItemPriceCalculationStrategy

In fact, there are only four of them since the strategy is only triggered if the 3rd optional parameter `updatedOrderLines` is given and only 4 methods provide it


michaelbromley:
Thanks!

Issue #1432, opened 2022-02-16
Use a shop-api endpoint to get Mollie payment intent instead of addPaymentToOrder 


martijnvdbrug:
**Is your feature request related to a problem? Please describe.**
When using Mollie as paymentmethod, an order can get stuck in the `PaymentAuthorized` state as described here: https://github.com/vendure-ecommerce/vendure/issues/1330

**Describe the solution you'd like**
Use a `createMolliePaymentIntent` or `getMolliePaymentUrl` endpoint to get the Mollie redirectUrl for an order payment, and add Mollie as paymentMethod to the order on callback from Mollie. Similar to the solution for Stripe described here: https://github.com/vendure-ecommerce/vendure/pull/1417 

**Describe alternatives you've considered**
Allow orders to stay in PaymentCreated state after addOrderToPayment has been called. Also described in #1330 

**Additional context**
How do we deal with this breaking change? Do we keep a hybrid solution where both `addPaymentToOrder` and the new endpoint exist?

Since there is no way to deprecate the `addPaymentToOrder` mutation for Mollie only, I suggest we don't go through the trouble of maintaining both solutions and just do the breaking change.

You can assign this issue to me.

michaelbromley:
I would also vote to just do the break. I'll announce it in the blog post etc, but I think the impact will be small because the package is not yet widely adopted.

martijnvdbrug:
@michaelbromley We have 2 ways of implementing this:

### 1. GraphQL query
A GraphQL query `createMolliePaymentIntent` which returns the redirectUrl. The client would need to redirect to Mollie for the payment.

:heavy_check_mark:  This option is in line with the Stripe plugin.
:heavy_check_mark:  An error can be shown in the storefront when createPaymentIntent failed, because the redirect was retrieved via an AJAX call
:x:  Extra network call, so a bit slower and a slightly higher chance of failure


### 2. Rest + Redirect
A rest endpoint `payWithMollie` which does something like this:
```js
const redirectUrl = await mollie.createPayment()
res.redirect(302, redirectUrl)
```
:heavy_check_mark:  Saves 1 network call from the storefront. Less complexity on the storefront side
:x: Cannot be done via AJAX call, so the user will leave the storefront, even when redirect creation fails.

I slightly favor option 1. What do you think?

michaelbromley:
I'd go with option 1 too. Giving more control to the client at the slight expense of extra steps to implement I think is the better choice in the long-run.

[...]

Issue #1340, opened 2022-01-12
Save chosen Mollie payment method (creditcard, paypal and so on) in payment metadata of vendure


nabbl:
It would be great if the mollie payment plugin would save the selected payment method from mollie into the database.

So that when we have to make some reports we can include that information for the customer.

michaelbromley:
Hi, thanks for the suggestion. 

The way this would be done is to put the additional data in the `metadata` object [here](https://github.com/vendure-ecommerce/vendure/blob/42c0be4b49d9e234677206ea1a7eb4f5f034ae0f/packages/payments-plugin/src/mollie/mollie.controller.ts#L48) I think. It probably also makes sense to save additional metadata which might also be useful in debugging payment issues etc. It would be something like this:

```
dbPayment.metadata = {
    ...dbPayment.metadata,
    status: molliePayment.status,
    paidAt: molliePayment.paidAt,
    amount: molliePayment.amount,
    amountCaptured: molliePayment.amountCaptured,
    description: molliePayment.description,
    method: molliePayment.method,
    profileId: molliePayment.profileId,
    settlementAmount: molliePayment.settlementAmount,
    settlementId: molliePayment.settlementId,
    customerId: molliePayment.customerId,
    orderId: molliePayment.orderId,
};
await this.connection.getRepository(Payment).save(dbPayment);
``` 

Would you be interested in having a go at a PR for this?

martijnvdbrug:
Sure. I'm currently workin on #1432 aswell, so I can also do this one. Hopefully I will have a PR somewhere next week.

PR #1437, opened 2022-02-21
Mollie payment intent + Stripe unauthorized settlement fix

martijnvdbrug:
Fixes for Mollie payments plugin. https://github.com/vendure-ecommerce/vendure/issues/1432 and https://github.com/vendure-ecommerce/vendure/issues/1340

**Breaking** changes, as storefronts need to change their flow.

+ a fix for the Stripe plugin


michaelbromley:
Note: please also test compatibility with @mollie/api-client v3.6.x - I just deleted my lockfile and 3.6 got installed and the e2e tests started failing. On reverting to 3.5.1, tests passed again.

martijnvdbrug:
@michaelbromley `createMolliePaymentIntent()` should only work when a shippingMethod has been set on the order. Otherwise we get in trouble later when trying to settle the order via the webhook. This is too late, because the payment has already been done at that point. 

This results in some duplicate business logic:`MollieService`, now also checks if an order has a shippingMethod set. Is there a nicer way to check if the order is `ready for transition to ArrangingPayment`? I see internally there is something like `onTransitionStart`

michaelbromley:
I don't think there is a nicer way to do that. It's just this bit that you need right? https://github.com/vendure-ecommerce/vendure/blob/a8e68227456fb5483f4ef4bfb9128f408d583221/packages/core/src/service/helpers/order-state-machine/order-state-machine.ts#L132-L134

packages/payments-plugin/README.md
```diff
@@ -0,0 +1,3 @@
+# Payments plugin
+
+For documentation, see https://www.vendure.io/docs/typescript-api/payments-plugin
```
martijnvdbrug, line 3: Moved this here, so there is a link on NPM to the docs.

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -115,33 +143,20 @@ describe('Mollie payments', () => {
...
-        await shopClient.query<AddItemToOrder.Mutation, AddItemToOrder.Variables>(ADD_ITEM_TO_ORDER, {
-            productVariantId: 'T_1',
-            quantity: 2,
-        });
-        await proceedToArrangingPayment(shopClient);
-        const { addPaymentToOrder } = await shopClient.query<
-            AddPaymentToOrder.Mutation,
-            AddPaymentToOrder.Variables
-        >(ADD_PAYMENT, {
+        await setShipping(shopClient);
```
martijnvdbrug, line 146: Without shipping a user cannot request a payment intent

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -168,14 +183,25 @@ describe('Mollie payments', () => {
...
+
     it('Should fail to refund', async () => {
         let mollieRequest;
         nock('https://api.mollie.com/')
             .post(/.*/, body => {
                 mollieRequest = body;
                 return true;
             })
-            .reply(200, { status: 'failed' });
+            .reply(200, { status: 'failed', resource: 'payment' });
```
martijnvdbrug, line 204: The new Mollie node client checks for the field `resource`. This is why the tests were failing after upgrading from Mollie 3.5 to 3.6

packages/payments-plugin/src/mollie/mollie.handler.ts
```diff
@@ -31,69 +31,34 @@ export const molliePaymentHandler = new PaymentMethodHandler({
...
-        } catch (err) {
-            Logger.error(err, loggerCtx);
-            return {
-                amount: order.totalWithTax,
-                state: 'Error',
-                errorMessage: err.message,
-            };
+        // Creating a payment immediately settles the payment in Mollie flow, so only Admins and internal calls should be allowed to do this
+        if (ctx.apiType !== 'admin') {
+            throw Error(`CreatePayment is not allowed for apiType '${ctx.apiType}'`);
```
martijnvdbrug, line 51: Is this safe? It's not possible for a shop-api user to spoof the apiType right?

michaelbromley: It is determined by this logic: https://github.com/vendure-ecommerce/vendure/blob/42aa5fbac5194a3f80c6e40a9ae0c5a608d65dfa/packages/core/src/api/common/get-api-type.ts#L18

So no, not possible to spoof from front-end. Possible vuln could be inadvertently introduced by the developer if she defines a `administrators` query in the Shop API though. So I probably should add some mitigation for that possibility.

packages/payments-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -57,28 +61,16 @@ export interface MolliePluginOptions {
  * PaymentMethod was given the code "mollie-payment-method".
```
michaelbromley, line 61: Docs need updating here to not refer to "addPaymentToOrder"

packages/payments-plugin/src/mollie/mollie-shop-schema.ts
```diff
@@ -0,0 +1,18 @@
+import { gql } from 'graphql-tag';
+
+export const shopSchema = gql`
+    type MolliePaymentIntentError {
```
michaelbromley: Implement the `ErrorResult` interface here for easier generic error handling on the client.

packages/payments-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -57,28 +61,16 @@ export interface MolliePluginOptions {
...
- *      payments {
- *          id
- *          metadata
- *      }
- *    }
- *    ...on ErrorResult {
- *      errorCode
- *      message
- *    }
+ *       url
```
michaelbromley: Since you are returning a union type, use the fragment spread syntax and show both the happy path and also the error case being handled.

michaelbromley (changes requested):
Excellent work Martijn! Just a few small points to look at and then it can be merged.

packages/payments-plugin/src/mollie/mollie-shop-schema.ts
martijnvdbrug: Done, it now uses the errorCode `ORDER_PAYMENT_STATE_ERROR` when an order is not 'ready' for payment intent.

PR #1460, opened 2022-03-05
fix(core): Support usage GQL interface on relational custom field

ashitikov:
Currently, if you use GraphQL interface on relational custom field, admin-ui will not be able to query all fields, that is presented in this interface. This pull request fixes global config generation.


michaelbromley:
Thank you 👍

PR #1469, opened 2022-03-11
fix(core): Add order-testing service to index.ts

teresaWienkBorgert:
Added the order-testing service to the index.ts to be able to use it in the implementation.

Fixes #1469


michaelbromley:
Thank you!

PR #1472, opened 2022-03-14
fix(core): Truthy check for custom fields in importer

rheelly:
Added a truthy check in the processCustomFieldValues function
in the data importer so that no attempts are made to write empty strings
into the databse (e.g. for float or int fields)  

If there is an empty field in the given CSV file (e.g. `,,,`) these are parsed as empty strings which will then lead to a driver error by typeORM for fields with a data-type where this would be non-valid.


michaelbromley:
Hi,
Thanks for the contribution. It looks like this is changing the behaviour with regard to custom field lists - previously an empty string would be imported `[]`, but now it is imported as `null`.

Could you perhaps create a failing e2e test for the specific scenario you are trying to handle here, and then make a change which fixes your test without regressing any existing tests? Let me know if you need any assistance on that part. Thanks!

rheelly:
I added some data into the product.csv and the snapshots for the tests and reflected what I meant in the `imports products` test. Is this alright or do I need to make a separate test-case for this?  

The thing is that with the additional line in the `product-import.csv` where the weight field of the productVariant is now empty all importer tests would fail with the previous behavior with the following error: 
```
@vendure/core: $ jest --config ../../e2e-common/jest-config.js --package=core
@vendure/core: FAIL e2e/import.e2e-spec.ts (20.863 s)
@vendure/core:   ● Import resolver › imports products
@vendure/core:     Int cannot represent non-integer value: ""
@vendure/core:       88 |         } else {
@vendure/core:       89 |             const errorResult = typeof result === 'string' ? { error: result } : result;
@vendure/core:     > 90 |             throw new ClientError(
@vendure/core:          |                   ^
@vendure/core:       91 |                 { ...errorResult, status: response.status },
@vendure/core:       92 |                 { query: print(query), variables },
@vendure/core:       93 |             );
@vendure/core:       at SimpleGraphQLClient.query (../../testing/src/simple-graphql-client.ts:90:19)
@vendure/core:       at Object.<anonymous> (import.e2e-spec.ts:82:31)
@vendure/core:   ● Import resolver › imports products with multiple languages
@vendure/core:     String cannot represent value: ""
@vendure/core:       88 |         } else {
@vendure/core:       89 |             const errorResult = typeof result === 'string' ? { error: result } : result;
@vendure/core:     > 90 |             throw new ClientError(
@vendure/core:          |                   ^
@vendure/core:       91 |                 { ...errorResult, status: response.status },
@vendure/core:       92 |                 { query: print(query), variables },
@vendure/core:       93 |             );
@vendure/core:       at SimpleGraphQLClient.query (../../testing/src/simple-graphql-client.ts:90:19)
@vendure/core:       at Object.<anonymous> (import.e2e-spec.ts:254:31)
@vendure/core: PASS e2e/default-search-plugin.e2e-spec.ts (38.543 s)
```

michaelbromley:
> I added some data into the product.csv and the snapshots for the tests and reflected what I meant in the imports products test. Is this alright or do I need to make a separate test-case for this?

the `product-import.csv` is only used in 2 e2e test suites - `import.e2e-spec.ts` and `populate.e2e-spec.ts`, so I think it should be ok assuming the tests in the "populate" suite don't start breaking.

I do however see that the `custom-fields.e2e-spec.ts` suite is failing [here](https://github.com/vendure-ecommerce/vendure/runs/5585043976?check_suite_focus=true#step:9:29). Is that a regression or is it the case that it was wrong before?

michaelbromley:
ok, I'll take a closer look into this soon when I get a bit of time. Thanks for your work so far!

michaelbromley:
Just took a look into this. What I think is happening is this: that `custom-fields.e2e-spec.ts` file has a test to make sure that if a custom list field has a default value specified, it will be used when a new entity is created. Previously, when importing the string list custom field, the value would end up `undefined` and so TypeORM would use the default value. By changing the importer to use a default of `[]`, this behaviour started failing.

I made a small modification to handle this and now everything is merge into master by a8c44d1e. Thanks again for your help!

PR #1481, opened 2022-03-19
fix(core): Copy context on transaction start. Do not allow to run queries after transaction aborts.

ashitikov:
**The problem**
Current architecture of transaction and context guarantees that if one of operation inside of transaction block fails, whole transaction aborts. To bound queries to transaction we use getRepository(ctx, Entity) function, where ctx is the context which bound to transaction:
```
executeInTransaction(ctx, ctx => {
  await getRepository(ctx, Entity).update(...) // op1
  await getRepository(ctx, Entity2).delete(...) // op2
  ... lots of another ops
})
```
If op2 fails, then op1 and other ops in the transaction block aborts, that is what developer actually expects.

Say you have this code:
```
executeInTransaction(ctx, ctx => {
  op1 = getRepository(ctx, Entity).update(...)
  op2 = getRepository(ctx, Entity2).delete(...)
  await Promise.all([op1, op2, op3, opN...])
})
```
This way developer expects the same behavior as in previous example, but instead, if op2 fails, other operations may execute some queries outside of transaction as the result. This happens because of two things:
1. _Promise.all()_ does not abort promises that are already in process of execution
2. getRepository(ctx, Entity) contains check on _isReleased_ property of query runner: https://github.com/vendure-ecommerce/vendure/blob/b70cee883662281b5653ceaa12617c4d0b08a37a/packages/core/src/connection/transactional-connection.ts#L76.
So imagine that op1, op2 and op3 runs in parallel, if op2 fails, op3 may still execute _getRepository(ctx, Entity)_ at some point and you get repository with fresh QueryRunner instance, that is able to execute sql queries. 

**The solution**
1. Do not check _isReleased_ property, to ensure _getRepository(ctx, Entity)_ will get repository that is bound to released query runner. So released query runner will throw an exception in case of any query execute.
2. Copy RequestContext and pass copied instance to _executeInTransaction()_ function, to make sure that original context will be able to run queries outside of transaction.

**Note**
Since transaction binds to copied context, developer must execute queries using copied context, otherwise queries will run outside of transaction.

### DIFF
--- a/packages/core/e2e/database-transactions.e2e-spec.ts
+++ b/packages/core/e2e/database-transactions.e2e-spec.ts
@@ -13,6 +13,14 @@ import {
     TRIGGER_ATTEMPTED_UPDATE_EMAIL,
 } from './fixtures/test-plugins/transaction-test-plugin';
 
+type DBType = 'mysql' | 'postgres' | 'sqlite' | 'sqljs';
+
+const itIfDb = (dbs: DBType[]) => {
+    return dbs.includes(process.env.DB as DBType || 'sqljs')
+        ? it
+        : it.skip
+} 
+
 describe('Transaction infrastructure', () => {
     const { server, adminClient } = createTestEnvironment(
         mergeConfig(testConfig(), {
@@ -67,6 +75,26 @@ describe('Transaction infrastructure', () => {
         expect(verify.users.length).toBe(2);
         expect(!!verify.admins.find((a: any) => a.emailAddress === 'test2')).toBe(false);
         expect(!!verify.users.find((u: any) => u.identifier === 'test2')).toBe(false);
+    });
+
+    it('failing mutation with promise concurrent execution', async () => {
+        try {
+            await adminClient.query(CREATE_N_ADMINS, {
+                emailAddress: 'testN-',
+                failFactor: 0.4,
+                n: 10
+            })
+            fail('Should have thrown');
+        } catch (e) {
+            expect(e.message).toContain('Failed!');
+        }
+
+        const { verify } = await adminClient.query(VERIFY_TEST);
+
+        expect(verify.admins.length).toBe(2);
+        expect(verify.users.length).toBe(2);
+        expect(!!verify.admins.find((a: any) => a.emailAddress.includes('testN'))).toBe(false);
+        expect(!!verify.users.find((u: any) => u.identifier.includes('testN'))).toBe(false);
     });
 
     it('failing manual mutation', async () => {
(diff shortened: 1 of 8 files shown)


michaelbromley:
ok I see some tests failing relating to the transactions. If you cannot debug this let me know and I'll take a look myself when I have a little time.

ashitikov:
> ok I see some tests failing relating to the transactions. If you cannot debug this let me know and I'll take a look myself when I have a little time.

Hey, It's strange, but I cannot reproduce failing tests in _"e2e/database-transactions.e2e-spec.ts"_:
```
@vendure/core: PASS e2e/database-transactions.e2e-spec.ts
```
This test successfully pass on my envorinment using sqljs. Am I missing something?

ashitikov:
Hey! Seems I finally reproduced and resolved transaction tests issues after a lot of time. Here is a list of changes with some notes:
1. Added e2e test to correctly reproduce and fix an issue. You can see that new test skips sqljs and sqlite databases. Unfortunately, this can't be fixed on these databases because TypeORM do not release actually _QueryRunner_ after calling _release()_ (Generally, query runner must set isReleased = true property after release). https://github.com/typeorm/typeorm/blob/486f8c582170dbadbd875e00d09d26359b2fd8be/src/driver/sqlite-abstract/AbstractSqliteQueryRunner.ts#L63 . After some research, I found that query runner deeply cached, and every time you call createQueryRunner(), connection returns the same instance of query runner. 
2. Function _executeInTransaction()_ now copies request context. This has to be done to protect original context from mutation. So inside transaction block we may do whatever we want with local copy of context. When transaction completes, query runner correctly releases. So after that, any query with this context will throw an exception from TypeORM: Query runner already released. This is the main point what am I trying to achieve with this PR. In case of parallel execution there will be no data race between promises. 
3. Event bus replaces RequestContext after awaiting query runner release. In fact after query runner releases, context become invalid. To make sure event bus subscription will be able to query some data, we replace invalid request context with a copy, that does not have transaction manager.
4. Transaction() decorator now associate transactional request context with current execution handler. Ctx() decorator on its own returns transactional request context if available. This is done because default request context from request object do not hold transaction manager anymore. 

This implementation is not a perfect solution, but clean with some workarounds. The main reason is not to have breaking changes. I tried to invent a solution that will support sqljs and sqlite as well (using Proxy on query runner and its manager), but lost a couple of days on it - it does not work as expected.

michaelbromley:
This is really impressive work. Some of the most complex parts of Vendure to deal with, and you've done it in a clean and well-documented way. Thank you!

PR #1485, opened 2022-03-23
fix(payment-plugin,stripe): fix state transitioning 

casabian:
Fix state transitioning for orders that are already in 'ArrangingPayment' state for stripe payment_intent webhook endpoint.

This happens when using vue-storefront/vendure when adding a billing address ([see updateAddressDetails/index.ts](https://github.com/vuestorefront-community/vendure/blob/15b03f4da3398a968e7bd64976b8bf9958456c64/packages/api-client/src/api/updateAddressDetails/index.ts#L26))

Same implementation as for mollie payment service ([mollie.service.ts](https://github.com/vendure-ecommerce/vendure/blob/c6cf4d4e214f3941bc5f27017b0354dda732f9e0/packages/payments-plugin/src/mollie/mollie.service.ts#L138))

### DIFF
--- a/packages/payments-plugin/src/stripe/stripe.controller.ts
+++ b/packages/payments-plugin/src/stripe/stripe.controller.ts
@@ -8,16 +8,16 @@ import {
     OrderService,
     PaymentMethod,
     RequestContext,
-    TransactionalConnection,
+    TransactionalConnection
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thank you!

PR #1487, opened 2022-03-24
Fix: added a return(202) to stripe webhook in case of successfull payment

julienTeam:
Fix: added a return(202) to stripe webhook in case of successfull payment


michaelbromley:
Thank you for the fix! One question, in the docs here (https://stripe.com/docs/webhooks/quickstart), it talks about returning a `200` response rather than `202`: 
image

Is there some other documentation regarding the `202` response?

julienTeam:
Michael, you are correct, I didn't read this part and it worked with 202 but indeed, better to stick with the documentation and apply a 200, I'll correct this.

michaelbromley:
Thank you!

PR #1504, opened 2022-04-04
feat(core): Make search strategy configurable via plugin options

casabian:


casabian:
Hey @michaelbromley I'm not quite sure why the unit tests fail. On my local machine all tests pass as expected.

michaelbromley:
Hi, I'll take a look locally at those unit tests. Might just be a temp issue with GH actions, because the logs look weird - it just gets cancelled with no error output. Will review later today.

packages/core/src/plugin/default-search-plugin/search-strategy/search-strategy.ts
```diff
@@ -2,12 +2,14 @@ import { SearchInput, SearchResult } from '@vendure/common/lib/generated-types';
...
 import { RequestContext } from '../../../api';
+import { InjectableStrategy, Injector } from '../../../common';
 
 /**
  * This interface defines the contract that any database-specific search implementations
  * should follow.
  */
-export interface SearchStrategy {
+export interface SearchStrategy extends InjectableStrategy {
+    init: (injector: Injector) => void | Promise<void>;
```
michaelbromley: The `init` method should not need to be re-declared here, since it will be inherited from the InjectableStrategy interface.

packages/core/src/plugin/default-search-plugin/types.ts
```diff
@@ -35,6 +37,8 @@ export interface DefaultSearchPluginInitOptions {
      * @default false
      */
     bufferUpdates?: boolean;
+
```
michaelbromley, line 40: Can you add a documentation block here? It should have `@description`, `@since 1.6.0` and `@default undefined` tags. If you are not sure what to write for the description let me know and I can assist.

michaelbromley (changes requested):
Thank you for this - a very nice, clean PR following existing patterns & style. I re-ran the unit tests and they passed this time - so there was no issue with your changes.

I just added a couple of small points to address.

casabian:
I'll fix and add your points soon

casabian:
Hi @michaelbromley your commit https://github.com/vendure-ecommerce/vendure/commit/8d87f058926a27b584561104c229d0ad420bc483 has introduced typescript build errors.

michaelbromley:
I'll push some updates soon which should fix the build issues and then I'll review your recent changes 👍

michaelbromley:
Minor branch tests are passing now. Can you merge these in/rebase?

packages/core/src/plugin/default-search-plugin/default-search-plugin.ts
```diff
@@ -192,6 +195,16 @@ export class DefaultSearchPlugin implements OnApplicationBootstrap {
...
+        await this.initSearchStrategy();
+    }
+
+    private async initSearchStrategy(): Promise<void> {
+        const injector = new Injector(this.moduleRef);
+        const searchService = injector.get(FulltextSearchService);
+        if (typeof searchService.searchStrategy.init === 'function') {
+            await searchService.searchStrategy.init(injector);
+        }
     }
```
michaelbromley, line 214: We should also have a corresponding `destroySearchStrategy()` method here, in case the strategy implements the `destroy()` method. It would be called from the `onApplicationShutdown` hook (see the config module for example)

michaelbromley (changes requested):
Looking good! I just spotted one last thing and then it will be ready to merge :)

packages/core/src/plugin/default-search-plugin/default-search-plugin.ts
casabian: Done 👍

michaelbromley:
Thank you for the excellent work on this!

casabian:
> 

You're welcome always happy to help with your awesome project

PR #1509, opened 2022-04-06
feat(core): Pass shipping method to calculator and eligibility checker

ashitikov:
Currently there is not way to get shipping method for which calculator and eligibility checker called. This PR adds such ability.


michaelbromley:
Thank you!

PR #1550, opened 2022-05-06
fix(core): Job update doesn't emit if progress didn't changed

ashitikov:
**The problem**
Currently, getting polling updates through `job.updates()` is not always possible, because of a mistake in operator `distinctUntilChanged`. Current logic is to ignore updates if `progress` OR `state` is equal to previous value. So if your job do not update progress, you will not receive any update through polling. Also this bug spawns a memory leak, because `takeUntil` operator placed after `distinctUntilChanged`

**The solution**
Change operator to ignore updates if `progress` AND `state` are equal to previous values.


michaelbromley:
Thanks for the fix!

Issue #1553, opened 2022-05-06
Add SearchEvent


Draykee:
**Is your feature request related to a problem? Please describe.**
I would like to track and know about the search query usage.

**Describe the solution you'd like**
Broadcast an event whenever a search is executed.

My only concern is that the EventBus might be spammed a lot. @michaelbromley do you have a forecast if this could have an impact on the performance? Or is it neglectable?

PR #1554, opened 2022-05-06
Feature/search event

kevmtt:
Added a simple implementation to track the search inputs to solve #1553

### DIFF
--- a/packages/core/src/event-bus/events/search-event.ts
+++ b/packages/core/src/event-bus/events/search-event.ts
@@ -1,0 +1,22 @@
+import { SearchInput } from '@vendure/common/lib/generated-types';
+
+import { RequestContext } from '../../api/common/request-context';
+import { VendureEvent } from '../vendure-event';
+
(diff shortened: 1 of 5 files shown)


packages/core/src/event-bus/events/search-event.ts
```diff
@@ -0,0 +1,22 @@
...
+    [extendedInputField: string]: any;
+};
+
+/**
+ * @description
+ * This event is fired whenever a search query is executed.
+ *
+ * @docsCategory events
+ * @docsPage Event Types
+ * @since 1.5.3
```
michaelbromley, line 16: Set this as `1.6.0` since it will be going in the next minor release.

michaelbromley (changes requested):
Looks good - just a minor change on the documentation.

michaelbromley:
Since I want to release v1.6 today and want to include this feature, I went ahead and updated the `@since` decorator on the SearchEvent and merged locally. Merged in 0592f942.

PR #1564, opened 2022-05-16
feat(core): Pass payment method to handler and eligibility checker

ashitikov:
Currently there is not way to get payment method for which handler and eligibility checker called. This PR adds such ability.


michaelbromley:
Thanks! This is a logical addition based on the same feature you added for ShippingMethods. 👍

PR #1579, opened 2022-05-23
feat(core): savepoints support (nested transactions).

ashitikov:
**Motivation**

Nested transactions (savepoints) may be highly helpful in case if the developer wants to perform complex logic blocks with sql-queries:
```
transaction1 {
  1. Create new customer
  2. transaction2 {
      1. Change product quantity
      2. Add new products
    } catch() {}
  3. Assign new customer to group
}
```
Currently, there is no way to perform nested transactions, because typeorm (0.2.41 does not support savepoints) as well as vendure allows only one-level transactions at `startTransaction()` function. In the example above, everything in transaction2 will be performed in the same database transaction as everything in transaction1. If something bad happened in transaction2 we will not be able to perform step 3 (assign new customer to group), cause database transaction is already rolled back or attached query runner is released. 

```
@Mutation
@Transaction
async doSmth() {
  withTransaction(ctx, ctx => {
     await getRepository(ctx, User).create(...);
  })
  await getRepository(ctx, Administrator).create(...);
}
```
Another case is that having mutation with `@Transaction` decorator, vendure wraps function body to withTransaction() block, so create user query will be executed in the same transaction as administrator creation from this example. This is also unexpected by developer.
**Solution**

TypeORM 0.2.45 introduces savepoints support: https://github.com/typeorm/typeorm/blob/master/CHANGELOG.md#0242-2022-02-16.  Every time we call startTransaction TypeORM will go deeper in current transaction by creating a savepoint. 
On vendure's side:
1. We should allow developers to call startTransaction multiple times. 
2. Copying context on transaction start (https://github.com/vendure-ecommerce/vendure/pull/1481) works as expected only on the first (root) level of transaction. TypeORM manages levels inside query runners, so calling in parallel queries mixed with nested transactions (savepoints) will lead to unexpected results, cause some queries may be executed on another level of transaction. For example:
```
transaction1 { ctx1 =>
  transaction2 { ctx2 =>
     await Promise.all(Create 3 users in parallel with ctx2 context, where second query fails)
  }
}
```
Developer expects that all 3 users will be executed in the nested transaction (ctx2), however, from database perspective it may look like:
```
START TRANSACTION;
  SAVEPOINT s1;
    CREATE USER1;
    CREATE USER2; <---- FAILS
  ROLLBACK TO SAVEPOINT s1;
  CREATE USER3;
COMMIT;
```
You can see that third user will still be created in higher level transaction (ctx1), although developer expects that if transaction2 rollbacks, then no users will be created. In the most cases, this should not happen. Potentially, this can be partially resolved by using protected member `transactionDepth` of query runner, but it some kind of hack.
The benefits outweigh the potential risks of this feature, so I decided to implement this.

**Implementation**

Changes includes re-implementation of active transaction guard, and new active transaction check on release. Some tests are added to cover main cases.

**Bonus**

I saw a potential bug on event-bus, where events are published even on transaction rollback, which is wrong, I think. This pull request contains a fix for this problem.

**Note**

Because of TypeORM upgrade, I should note here, that it contains breaking changes. I think these changes are not affects somehow vendure.
BREAKING CHANGE (TypeORM): update listeners and subscriber no longer triggered by soft-remove and recover (https://github.com/typeorm/typeorm/blob/master/CHANGELOG.md#0242-2022-02-16).

Original slack thread: https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1650625231400079

### DIFF
--- a/packages/core/src/event-bus/event-bus.ts
+++ b/packages/core/src/event-bus/event-bus.ts
@@ -3,10 +3,11 @@ import { Type } from '@vendure/common/lib/shared-types';
 import { Observable, Subject } from 'rxjs';
 import { filter, mergeMap, takeUntil } from 'rxjs/operators';
 import { EntityManager } from 'typeorm';
+import { notNullOrUndefined } from '../../../common/lib/shared-utils';
 
 import { RequestContext } from '../api/common/request-context';
 import { TRANSACTION_MANAGER_KEY } from '../common/constants';
-import { TransactionSubscriber } from '../connection/transaction-subscriber';
+import { TransactionSubscriber, TransactionSubscriberError } from '../connection/transaction-subscriber';
 
 import { VendureEvent } from './vendure-event';
 
@@ -82,6 +83,7 @@ export class EventBus implements OnModuleDestroy {
             takeUntil(this.destroy$),
             filter(e => (e as any).constructor === type),
             mergeMap(event => this.awaitActiveTransactions(event)),
+            filter(notNullOrUndefined)
         ) as Observable<T>;
     }
 
@@ -109,7 +111,7 @@ export class EventBus implements OnModuleDestroy {
      * * https://github.com/vendure-ecommerce/vendure/issues/520
      * * https://github.com/vendure-ecommerce/vendure/issues/1107
      */
-    private async awaitActiveTransactions<T extends VendureEvent>(event: T): Promise<T> {
+    private async awaitActiveTransactions<T extends VendureEvent>(event: T): Promise<T | undefined> {
         const entry = Object.entries(event).find(([_, value]) => value instanceof RequestContext);
 
         if (!entry) {
@@ -123,7 +125,9 @@ export class EventBus implements OnModuleDestroy {
             return event;
         }
 
-        return this.transactionSubscriber.awaitCommit(transactionManager.queryRunner).then(() => {
+        try {
+            await this.transactionSubscriber.awaitCommit(transactionManager.queryRunner);
+
             // Copy context and remove transaction manager
             // This will prevent queries to released query runner
             const newContext = ctx.copy();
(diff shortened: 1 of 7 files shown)


packages/core/src/event-bus/event-bus.ts
```diff
@@ -123,7 +123,7 @@ export class EventBus implements OnModuleDestroy {
             return event;
         }
 
-        return this.transactionSubscriber.awaitRelease(transactionManager.queryRunner).then(() => {
+        return this.transactionSubscriber.awaitCommit(transactionManager.queryRunner).then(() => {
```
michaelbromley: I'm wondering whether there is an issue here - if the transaction fails and there is no commit, we will have a promise that never resolves right?

ashitikov: Currently no, because It awaits a release. You can inspect that release happens inside finally block, so anyway transaction completes or fails it releases. Currently the problem exists only if transaction fails: subscribers are still be able to receive an events

ashitikov: Ah, sorry, I though you are about code before changes. I see your point, yes, it seems we will have a promise that will never resolve. I'm not sure whether this can be considered as a memory leak

michaelbromley: Yes, I think this could potentially lead to a memory leak. Perhaps it should work like this: if the transaction fails, the `awaitActiveTransactions()` promise rejects, and we silently handle that rejection without emitting a value on the `.ofType()` observable.

ashitikov:
Good catch @michaelbromley about potential memory leak. Fixed this by re-writing an internal transaction awaiting function in transaction subscriber. Current implementation take care of expected an unexpected events without breaking changes.

PR #1585, opened 2022-05-26
fix(elasticsearch-plugin): Fix permissions for pendingSearchIndexUpdates query

vrosa:
This [commit](https://github.com/vendure-ecommerce/vendure/commit/152e64b) updated the permissions for the `pendingSearchIndexUpdates` query for `SearchResolver` and `AdminFulltextSearchResolver` but missed `AdminElasticSearchResolver`. This PR aims to fix that problem.


michaelbromley:
Thanks for the fix!

PR #1592, opened 2022-05-28
fix(admin-ui): Fix NavMenuItem.onClick callback not triggered

### LINKED ISSUE #1549: The `onClick` callback of a `NavMenuItem` is not triggered
**Describe the bug**
The `onClick` callback of a `NavMenuItem` object is not triggered when the user clicks the corresponding menu item on the Admin UI.

**To Reproduce**
Add a new nav menu item to the Admin UI using `addNavMenuItem` and pass, say, `() => console.log('CLICK')` to the `onClick` property of the `config` object parameter.

**Expected behavior**
The function passed in as the `onClick` property should be called when the user clicks the menu item - in this case, log "CLICK" to the console.

 - @vendure/core version: 1.5.1
 - Nodejs version: 14.18.1
 - Database (mysql/postgres etc): N/A

**Additional context**
I noticed the `onClick` property was introduced as part of [this commit](https://github.com/vendure-ecommerce/vendure/commit/ef47c62cafa3795b3af4d82f173031e950bad179), which also adds the same attribute to the `ActionBarItem` type. The code to handle the click event is there for an action bar item but not for the nav menu item - I wasn't sure whether or not it was supposed to have been added in the first place so I didn't want to raise a pull request.

vrosa:
Fixes #1549.

### DIFF
--- a/packages/admin-ui/src/lib/core/src/components/main-nav/main-nav.component.html
+++ b/packages/admin-ui/src/lib/core/src/components/main-nav/main-nav.component.html
@@ -24,6 +24,7 @@
                                     [attr.data-item-id]="section.id"
                                     [routerLink]="getRouterLink(item)"
                                     routerLinkActive="active"
+                                    (click)="item.onClick && item.onClick()"
                                 >
                                     <ng-container *ngIf="item.statusBadge | async as itemBadge">
                                         <vdr-status-badge


packages/admin-ui/src/lib/core/src/components/main-nav/main-nav.component.html
```diff
@@ -24,6 +24,7 @@
                                     [attr.data-item-id]="section.id"
                                     [routerLink]="getRouterLink(item)"
                                     routerLinkActive="active"
+                                    (click)="item.onClick && item.onClick()"
```
michaelbromley: We'd need to pass the click event here to satisfy the type of the `onClick` function:
```
item.onClick($event)
```

michaelbromley (changes requested):
Thanks, that's a good catch. Just one point to fix and then it's all good 👍

Issue #1598, opened 2022-06-02
Missing CustomMappingsResolver in Admin-api


alexisvigoureux:
**Describe the bug**
I have extended the customProductVariantMappings of the ElasticsearchPlugin

```
customProductVariantMappings: {
  deliveryType: {
    graphQlType: 'String',
    valueFn: (variant: ProductVariant) => {
      return variant.customFields.delivery_type ?? ''
    },
  },
}
```

When I call the endpoint on the shop-api, everything is fine, I get results.

```
query search($take: Int, $skip: Int, $term: String!) {
              search(input: {take: $take, skip: $skip, groupByProduct: false, term: $term }) {
                  items {
                       productName
                       customMappings {
                        ...on CustomProductVariantMappings {
                           delivery_type
                       }
                  }
              }
          }
```

But when I request it through the admin-api, I got this error:

```
"Abstract type \"CustomMappings\" must resolve to an Object type at runtime for field \"SearchResult.customMappings\". Either the \"CustomMappings\" type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function."
```

**Expected behavior**
I expect that a request on the admin-api work like on the shop-api

**Environment (please complete the following information):**
 - @vendure/core version: 1.6.1
 - Nodejs version: 14.17
 - Database (mysql/postgres etc): postgres

**Resolution**
Here is the PR to solve the issue: https://github.com/vendure-ecommerce/vendure/pull/1599

PR #1599, opened 2022-06-02
fix(elasticsearch-plugin): Missing CustomMappingsResolver in Admin-api

alexisvigoureux:
## Issue
https://github.com/vendure-ecommerce/vendure/issues/1598

### DIFF
--- a/packages/elasticsearch-plugin/src/plugin.ts
+++ b/packages/elasticsearch-plugin/src/plugin.ts
@@ -214,7 +214,27 @@ import { ElasticsearchOptions, ElasticsearchRuntimeOptions, mergeWithDefaults } 
         },
     ],
     adminApiExtensions: {
-        resolvers: [AdminElasticSearchResolver, EntityElasticSearchResolver],
+        resolvers: () => {
+            const { options } = ElasticsearchPlugin;
+            const requiresUnionResolver =
+                0 < Object.keys(options.customProductMappings || {}).length &&
+                0 < Object.keys(options.customProductVariantMappings || {}).length;
+            const requiresUnionScriptResolver =
+                0 <
+                    Object.values(options.searchConfig.scriptFields || {}).filter(
+                        field => field.context !== 'product',
+                    ).length &&
+                0 <
+                    Object.values(options.searchConfig.scriptFields || {}).filter(
+                        field => field.context !== 'variant',
+                    ).length;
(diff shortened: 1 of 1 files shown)


packages/elasticsearch-plugin/src/plugin.ts
```diff
@@ -214,7 +214,27 @@ import { ElasticsearchOptions, ElasticsearchRuntimeOptions, mergeWithDefaults }
...
+                0 < Object.keys(options.customProductVariantMappings || {}).length;
+            const requiresUnionScriptResolver =
+                0 <
+                    Object.values(options.searchConfig.scriptFields || {}).filter(
+                        field => field.context !== 'product',
+                    ).length &&
+                0 <
+                    Object.values(options.searchConfig.scriptFields || {}).filter(
+                        field => field.context !== 'variant',
+                    ).length;
```
michaelbromley: Lines 218 - 230 are duplicated exactly for both shopApiExtensions and adminApiExtensions. Could you extract this into a function like

```ts
function getCustomResolvers(options: ElasticsearchRuntimeOptions) {
    const requiresUnionResolver =
        0 < Object.keys(options.customProductMappings || {}).length &&
        0 < Object.keys(options.customProductVariantMappings || {}).length;
    const requiresUnionScriptResolver =
        0 <
            Object.values(options.searchConfig.scriptFields || {}).filter(
                field => field.context !== 'product',
            ).length &&
        0 <
            Object.values(options.searchConfig.scriptFields || {}).filter(
                field => field.context !== 'variant',
            ).length;
    return [
        ...(requiresUnionResolver ? [CustomMappingsResolver] : []),
        ...(requiresUnionScriptResolver ? [CustomScriptFieldsResolver] : []),
    ];
}
```
and then you can use it like
```ts
return [
    ShopElasticSearchResolver,
    EntityElasticSearchResolver,
    ...getCustomResolvers(ElasticsearchPlugin.options),
];
```

michaelbromley (changes requested):
Thanks! Just one suggestion here to de-duplicate that large code block.

packages/elasticsearch-plugin/src/plugin.ts
alexisvigoureux: Sure

michaelbromley (approved):

PR #1603, opened 2022-06-04
feat(core): Deprecation of getRepository without context argument

ashitikov:
**Motivation**
Having getRepository() with optional RequestContext argument is not safe:
```
withTransaction(ctx, ctx => {
  await getRepository(ctx, User).create(...)
  await getRepository(Administrator).create(...) <--- A serious mistake here, this query will be out of transaction.
});
```
Therefore current way to get repository from connection is error prone: it's simple to forget to pass context. 

**Solution**
Deprecate getRepository(Entity) function without RequestContext argument. Introduce `getOuterRepository()` function to enforce explicitly usage.
```
withTransaction(ctx, ctx => {
  await getRepository(ctx, User).create(...)
  await getRepository(Administrator).create(...) <--- A deprecation warning thrown by IDE. Developer should pay an attention
  // If this is intended behavior, developer should use
  // await getOuterRepository(Administrator).create(...) <--- No warning here
});
```

**Bonus**
Allow to construct new EntityNotFoundError with custom entity.

**Further steps**
Completely remove signature of getRepository without RequestContext argument in v2.0.

### DIFF
--- a/packages/core/src/connection/transactional-connection.ts
+++ b/packages/core/src/connection/transactional-connection.ts
@@ -7,7 +7,6 @@ import {
     EntitySchema,
     FindOneOptions,
     FindOptionsUtils,
-    getRepository,
     ObjectType,
     Repository,
 } from 'typeorm';
@@ -55,6 +54,8 @@ export class TransactionalConnection {
      * Returns a TypeORM repository. Note that when no RequestContext is supplied, the repository will not
      * be aware of any existing transaction. Therefore calling this method without supplying a RequestContext
      * is discouraged without a deliberate reason.
+     * 
+     * @deprecated since 1.6.3: Use {@link TransactionalConnection.getOuterRepository getOuterRepository()} function instead.
      */
     getRepository<Entity>(target: ObjectType<Entity> | EntitySchema<Entity> | string): Repository<Entity>;
     /**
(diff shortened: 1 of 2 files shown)


packages/core/src/connection/transactional-connection.ts
```diff
@@ -55,6 +54,8 @@ export class TransactionalConnection {
      * Returns a TypeORM repository. Note that when no RequestContext is supplied, the repository will not
      * be aware of any existing transaction. Therefore calling this method without supplying a RequestContext
      * is discouraged without a deliberate reason.
+     * 
+     * @deprecated since 1.6.3: Use {@link TransactionalConnection.getOuterRepository getOuterRepository()} function instead.
```
michaelbromley: Since it's going to be in the minor branch, change it to `since 1.7.0`

packages/core/src/connection/transactional-connection.ts
```diff
@@ -78,14 +79,28 @@ export class TransactionalConnection {
...
 
+    /**
+     * @description
+     * Returns a TypeORM repository, which is not aware of any existing transaction. 
+     * Therefore calling this method is discouraged without a deliberate reason.
+     * It's recommended to use {@link TransactionalConnection.getRepository getRepository(ctx, Entity)} instead.
+     * 
+     * @since 1.6.3
+     */
+    getOuterRepository<Entity>(
```
michaelbromley: I'm not 100% sure I like this naming. Will people know what is meant by "outer repository"? I wonder if there could be a name which better communicates what this is doing. 

- `getRepositoryOutsideTransaction`?
- `getRepositoryWithoutContext`?

not sure I like those any better... Any other ideas?

michaelbromley (commented):
Good - but I'm not sure about the naming yet.

packages/core/src/connection/transactional-connection.ts
ashitikov: Well, `getNoContextRepository()` is an option, but It's not better than you said. 
Another option is to delete getOuterRepository function completely, and instead force developer to use `get() rawConnection`. For example: `this.connection.rawConnection.getRepository()`. By the way, this class is called `TransactionalConnection` which means that every member of this function, like `getRepository()` or smth else should relate somehow to transaction. If developer does not need transactional repository we may force developer to inject original typeorm Connection class.

ashitikov:
Completely removed `getOuterRepository()`. Developer should use `rawConnection.getRepository()` instead. This is much more familiar solution without additional methods.

michaelbromley:
Yes! I love this solution which just removes the API completely. The most elegant code is no code ;)

PR #1614, opened 2022-06-08
Feat(core): variant price calculator

JMPJNS:
### Description
This PR adds a new helper service that can calculate the price of a variant, including promotions, without having to manually create the order, orderLines, orderItems... every time.

You can either supply a base order where it adds the variant and executes the promotions on, or use the currently active order for the session customer (default)

There are still some open fixme and todos that need to be adressed, i was hoping that @michaelbromley can give some input on those (what errors should be used and if it should just return the regular price in case it can't be calculated)

### DIFF
--- a/packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
+++ b/packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { CreateAddressInput } from '@vendure/common/lib/generated-types';
 
 import { RequestContext } from '../../../api';
+import { InternalServerError } from '../../../common';
 import { ConfigService } from '../../../config';
 import { Order, OrderItem, OrderLine, ProductVariant, Promotion, ShippingLine } from '../../../entity';
 import { CustomerService } from '../../services/customer.service';
@@ -32,21 +33,20 @@ export class VariantCalculator {
         initialOrder?: Order,
     ) {
         const customerId = initialOrder?.customer?.id ?? ctx.activeUserId;
-        if (!customerId)
-            // FIXME use the appropriate Error type, instead of generic Error
-            throw new Error('no active user or order customer');
+        if (!customerId) {
+            throw new InternalServerError('no active user or order customer');
+        }
 
         const { orderItemPriceCalculationStrategy } = this.configService.orderOptions;
-        if (!orderItemPriceCalculationStrategy)
-            // FIXME use correct Error type
-            throw new Error('no item price calculation strategy found');
+        if (!orderItemPriceCalculationStrategy) {
+            throw new InternalServerError('no item price calculation strategy found');
+        }
 
         let order: Order;
         if (initialOrder) {
             order = new Order(initialOrder);
         } else if (ctx.session?.activeOrderId) {
-            const found = await this.orderService.findOne(ctx, ctx.session?.activeOrderId);
-            order = new Order(found);
+            order = (await this.orderService.findOne(ctx, ctx.session?.activeOrderId)) ?? new Order();
         } else {
             order = new Order();
         }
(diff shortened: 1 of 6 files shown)


packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
```diff
@@ -0,0 +1,128 @@
...
+        ctx: RequestContext,
+        variant: ProductVariant,
+        promotions: Promotion[],
+        quantity = 1,
+        initialOrder?: Order,
+    ) {
+        const customerId = initialOrder?.customer?.id ?? ctx.activeUserId;
+        if (!customerId)
+            // FIXME use the appropriate Error type, instead of generic Error
+            throw new Error('no active user or order customer');
```
michaelbromley: You can use InternalServerError here.

packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
```diff
@@ -0,0 +1,128 @@
...
+     */
+    public async applyVariantPromotions(
+        ctx: RequestContext,
+        variant: ProductVariant,
+        promotions: Promotion[],
+        quantity = 1,
+        initialOrder?: Order,
+    ) {
+        const customerId = initialOrder?.customer?.id ?? ctx.activeUserId;
+        if (!customerId)
```
michaelbromley: Use curly braces even on single-line if clauses, just to fit in with the code style of the rest of the project.

packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
```diff
@@ -0,0 +1,128 @@
...
+            throw new Error('no active user or order customer');
+
+        const { orderItemPriceCalculationStrategy } = this.configService.orderOptions;
+        if (!orderItemPriceCalculationStrategy)
+            // FIXME use correct Error type
+            throw new Error('no item price calculation strategy found');
+
+        let order: Order;
+        if (initialOrder) {
+            order = new Order(initialOrder);
```
michaelbromley, line 47: If the `initialOrder` is of type `Order`, you should not need to construct a new one here right? Or is there some reason you need to?

packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
```diff
@@ -0,0 +1,128 @@
...
+        } else if (ctx.session?.activeOrderId) {
+            const found = await this.orderService.findOne(ctx, ctx.session?.activeOrderId);
+            order = new Order(found);
```
michaelbromley: Same - the `orderService.findOne` call will return an `Order` so no need to instantiate a new one, unless there is a specific reason I missed.

michaelbromley (commented):
Hi!
Thanks for the PR! I've left a few comments but my overall comment is: I'm not sure whether or not I want to have this in the core yet, because I'm not sure whether it is the best approach. 

This is a really tricky problem to solve, but the nice thing is that all of this code can live as a plugin without requiring any changes to core, right?

Are you already using this in a project? If so, I'd be interested to hear about:

1. How do you use it? Do you have a new query which is called on the product detail page to show discount prices? Or do you have a field resolver on ProductVariant?
2. What's the performance like? There's quite a bit of computation going on here to get that final price - especially if the activeOrder already has many OrderLines. Do you notice any performance impact from querying this?

packages/core/src/service/helpers/variant-calculator/variant-calculator.ts
JMPJNS: wouldn't adding orderLines to initialOrder mutate the order that is passed to the function, since its members are references and not copied values

JMPJNS:
I will do some benchmarking with really large orders to see how it impacts performance

JMPJNS:
and yes all of this can live inside a plugin, currently we have a custom query to get discounted prices for variants

davidhoeck:
We will close this because of in-activity and missing information. As everything can live within a plugin at this point, it is not really necessary to change it in the core.

Issue #1618, opened 2022-06-09
InitializedEvent to make sure the default channel is loaded


Draykee:
**Is your feature request related to a problem? Please describe.**
Right now the only way to initialize something is to use the `onModuleInit` function for NestJS modules.
Vendure currently uses the `InitializerService` to make sure settings, channels, roles, etc. are initialized and I would like to append init functions there.

**Describe the solution you'd like**
The `InitializerService` shall emit an `InitializedEvent` event once all services got initialized.

PR #1619, opened 2022-06-09
feat(event): Added event for a finished initialization (#1618)

Draykee:
Added event for a finished initialization (#1618)

### DIFF
--- a/packages/core/src/event-bus/events/initializer-event.ts
+++ b/packages/core/src/event-bus/events/initializer-event.ts
@@ -6,7 +6,7 @@ import { VendureEvent } from '../vendure-event';
  *
  * @docsCategory events
  * @docsPage Event Types
- * @since 1.6.3
+ * @since 1.7.0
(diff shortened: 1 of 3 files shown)


packages/core/src/event-bus/events/initializer-event.ts
```diff
@@ -0,0 +1,15 @@
+import { VendureEvent } from '../vendure-event';
+
+/**
+ * @description
+ * This event is fired when vendure finished initializing its services inside the {@code InitializerService}
+ *
+ * @docsCategory events
+ * @docsPage Event Types
+ * @since 1.6.3
```
michaelbromley: This will be in the next minor release, so `@since 1.7.0`

michaelbromley (changes requested):
Thanks! This is good - just a minor change in the docs.

michaelbromley:
Thanks!

PR #1639, opened 2022-06-28
fix(core): Use RequestContext if possible anywhere

ashitikov:
This is an adoption for coming deprecation of getRepository(Entity) without ctx argument https://github.com/vendure-ecommerce/vendure/pull/1603. 

Also fixes potential hanging on simultaneous transactions.
Why does this happen?
```
withTransaction(ctx, ctx => {
  // QueryRunner1 instantiated, which is bound to transaction
  await getRepository(ctx, Entity1).create(); // Using QueryRunner1
  await getRepository(Entity2).update(); // <--- Here is a problem. It acquires new connection from pool. Using QueryRunnerX
  await getRepository(Entity3).delete(); // Using QueryRunner1
});
```
If some code inside transaction executes a query without ctx argument `getRepository(Entity2).update()`, typeorm with pg-node driver acquires a connection from pool (QueryRunnerX), instead using an existing connection session, which is bound to current transaction (QueryRunner1). 
Default connection pool limit is 10. 
Once you reach the limit, pg-node will not resolve new connection promises until new connection will be available, so it just waits (QueryRunnerX is hanging). Because we await `getRepository(Entity2).update();` node js is also waits to be able to continue and hangs, and transaction will never complete, holding locks.

Temporary solution - to increase connection limit in the pool, using connection options in vendure config:
```
extra: {
  max: 50 // default 10
}
```

This PR is still in progress, as well as problem investigation.

### DIFF
--- a/packages/core/src/api/resolvers/admin/channel.resolver.ts
+++ b/packages/core/src/api/resolvers/admin/channel.resolver.ts
@@ -52,8 +52,8 @@ export class ChannelResolver {
         if (isGraphQlErrorResult(result)) {
             return result;
         }
-        const superAdminRole = await this.roleService.getSuperAdminRole();
-        const customerRole = await this.roleService.getCustomerRole();
+        const superAdminRole = await this.roleService.getSuperAdminRole(ctx);
+        const customerRole = await this.roleService.getCustomerRole(ctx);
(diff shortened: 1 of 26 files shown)


michaelbromley:
Thank you for this extremely thorough and careful work! Very, very much appreciated!

PR #1672, opened 2022-07-22
feat(core): Declare setDefaultContext in VendureLogger

vrosa:
### Context:

Our application uses a custom [Winston](https://github.com/winstonjs/winston) logger implementing `VendureLogger`, which behind the scenes sends logs to New Relic but whose output is similar to `DefaultLogger`. At first, we had quite a few log entries with an "[undefined]" prefix because, in different places, messages are logged without a `context` parameter. `DefaultLogger` handles this scenario by internally defining a "VendureServer" default context, which is overridden by the worker process to use "VendureWorker" instead. However, this logic is not flexible and only applies to instances of `DefaultLogger`.

### Proposed solution:

The idea is to define `setDefaultContext` as an optional method of `VendureLogger` - therefore not resulting in breaking changes - and update `bootstrapWorker` to call the function only if defined in the logger object. This way, custom implementations of `VendureLogger` can also benefit of such functionality if they deem so.

Another tiny update was on a commented out line of code from `dev-server/dev-config.ts`: `DefaultJobQueuePlugin.init` expects an object as an argument.

### DIFF
--- a/packages/core/src/bootstrap.ts
+++ b/packages/core/src/bootstrap.ts
@@ -98,9 +98,7 @@ export async function bootstrap(userConfig: Partial<VendureConfig>): Promise<INe
 export async function bootstrapWorker(userConfig: Partial<VendureConfig>): Promise<VendureWorker> {
     const vendureConfig = await preBootstrapConfig(userConfig);
     const config = disableSynchronize(vendureConfig);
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks, nice addition 👍

PR #1678, opened 2022-07-23
feat(core): a way to define custom states in type-safe manner

ashitikov:
** Motivation **
Developer has a great way to extend Order/Payment/Fulfillment processes with a custom states. But it lacks typescript safety, so after defining new state, you can stuck with an execution of standard service methods, like:
```
this.orderService.transitionPaymentToState(ctx, id, 'MyCustomState'); // <-- Here is an typescript error, because MyCustomState is missing from PaymentState
```

** Solution **
It's not possible currently to somehow merge declarations directly for types, but instead, we may use interfaces and `keyof` in this way:

```
export interface CustomStates {}
export type PaymentState = 'Created' | Pending | ... | keyof CustomStates;

// Merge declarations
export interface CustomStates {
  MyCustomState: never;
}

this.orderService.transitionPaymentToState(ctx, id, 'MyCustomState'); // <-- Now okay, because it's a part of PaymentState
```
Using this technique enforces `PaymentState` to have  keys from interface, that could be merged with custom declarations in similar way, as we do with custom fields.

### DIFF
--- a/docs/content/developer-guide/customizing-the-order-process/index.md
+++ b/docs/content/developer-guide/customizing-the-order-process/index.md
@@ -49,6 +49,7 @@ export const customerValidationProcess: CustomOrderProcess<'ValidatingCustomer'>
   },
 };
 ```
+
 This object means:
 
 * the `AddingItems` state may _only_ transition to the `ValidatingCustomer` state (`mergeStrategy: 'replace'` tells Vendure to discard any existing transition targets and replace with this one). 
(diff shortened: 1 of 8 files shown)


michaelbromley:
Oh this is a smart solution! Will take some time to properly review & test it out in the coming week.

michaelbromley:
OK just tested this out locally. Works really nicely! Can you just make the PR against `minor` since it is a feature, then I'll merge.

ashitikov:
Yeah, changed :)

PR #1684, opened 2022-07-26
feat(admin-ui,ui-devkit): Support Clarity Sass variable overrides

vrosa:
### Context

Although the Admin UI supports theming via CSS custom properties, there's a vast amount of styling in Clarity that, despite being customisable via Sass variables, is not accessible to Vendure developers. The design system framework is undergoing a transformation period where, slowly, Sass variables are being converted to CSS custom properties, but there are still [hundreds](https://github.com/vmware-clarity/ng-clarity/blob/689a572344149aea90df1676eae04479795754f3/projects/angular/src/utils/_variables.clarity.scss) of styling rules that are controlled by the CSS preprocessor. At the moment, there's no way to properly configure theming of such rules as intended on the Admin UI; instead, new declarations targeting specific elements must be added to `globalStyles`, often times using the `!important` keyword to override built-in definitions.

### Proposed solution

Provide support for developers to override Sass variables defined in Clarity. This way, the Admin UI styles can be _**fully**_ customised by those who want to embrace Clarity's design system (like my team).

### Description of changes

* Update `@vendure/ui-devkit` to support a new `SassVariableOverridesExtension` extension which, similar to the existing `GlobalStylesExtension`, accepts a single reference to a Sass file that is meant to contain Clarity variable definitions. The file, if provided, is imported _**before**_ the global `styles.scss` file, which ensures that any declared values will override Clarity defaults. The extension is optional and can be supplied as stand-alone or in conjunction with `GlobalStylesExtension`.
* Update the global `_variables.scss` stylesheet to import Clarity's main variables file. This way, the design system's Sass variables are also available for the consumption of Vendure's components, which can already consume CSS custom properties. Clarity's mixins file had to be imported as well since it's required by some of their variables.
* Add a new `_sass-overrides.scss` stylesheet, which should host Clarity variable definitions whose values are overridden on the Admin UI. The file is the first import of both `ui-extension.theme.scss` and `styles.scss` in order to have the new values propagated properly. At the moment, it only redefines `$clr-sidenav-width`, whose [default value is `18%`](https://github.com/vmware-clarity/ng-clarity/blob/689a572344149aea90df1676eae04479795754f3/projects/angular/src/layout/nav/_variables.sidenav.scss#L25) and is now set to `10.8rem` as per current UI. This value was hardcoded in `main-nav.components.scss` and `app-shell.component.scss`, which can now rely on the corresponding Sass variable as intended.
* Update documentation accordingly and fix a couple typos.
* Slightly update the importer function in `compile-styles.js` script to conform to the [official API reference](https://sass-lang.com/documentation/js-api/modules#LegacyImporterResult).

#### Notes

* No build time regression was observed
* No generated CSS file size regression was observed (styles.css and theme.css)
* Development and testing were executed using [this approach](https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1644346353760739?thread_ts=1643753248.067699&cid=CKYMF0ZTJ)

### DIFF
--- a/docs/content/plugins/extending-the-admin-ui/admin-ui-theming-branding/_index.md
+++ b/docs/content/plugins/extending-the-admin-ui/admin-ui-theming-branding/_index.md
@@ -65,7 +65,7 @@ You can replace the Vendure logos and favicon with your own brand logo:
 
 ## Theming
 
-Much of the visual styling of the Admin UI can be customized by providing your own themes in a Sass stylesheet. The Admin UI uses [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to control colors and other styles. Here's a simple example which changes the color of links:
+Much of the visual styling of the Admin UI can be customized by providing your own themes in a Sass stylesheet. For the most part, the Admin UI uses [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to control colors and other styles. Here's a simple example which changes the color of links:
 
 1. Install `@vendure/ui-devkit`
 2. Create a custom stylesheet which overrides one or more of the CSS custom properties used in the Admin UI:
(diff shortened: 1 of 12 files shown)


docs/content/plugins/extending-the-admin-ui/admin-ui-theming-branding/_index.md
```diff
@@ -100,3 +100,35 @@ Much of the visual styling of the Admin UI can be customized by providing your o
...
+    import { compileUiExtensions } from '@vendure/ui-devkit/compiler';
+
+    export const config: VendureConfig = {
+      // ...
+      plugins: [
+        AdminUiPlugin.init({
+          app: compileUiExtensions({
+            outputPath: path.join(__dirname, 'admin-ui'),
+            extensions: [{
+              variableOverrides: path.join(__dirname, 'my-variables.scss')
```
michaelbromley: The actual code check in `isSassVariableOverridesExtension` is checking for a `variableStyles` property. So either this documentation is wrong, or the check is wrong.

michaelbromley (commented):
Great work! Just one point on the documentation to check.

docs/content/plugins/extending-the-admin-ui/admin-ui-theming-branding/_index.md
vrosa: Good catch. To be honest, I struggled a bit with the property name since I didn't like either haha. I decided to rename it to `sassVariableOverrides` to be as precise as possible and not cause confusion with CSS custom properties, which are also "variables".

packages/admin-ui/src/lib/static/styles/global/_sass-overrides.scss
```diff
@@ -0,0 +1,3 @@
+// Note: variables defined in this file should use the default! flag so they can be
+// overridden by custom Admin UI applications
+$clr-sidenav-width: 10.8rem !default;
```
vrosa, line 3: This is a late addition. Using the `default!` flag allows the developer to override this and other default Vendure values on their Admin UI app - which I need, actually.

Issue #1685, opened 2022-07-26
AdjustOrderLine to 0 for last orderLine results in error


martijnvdbrug:
**Describe the bug**
When the activeOrder has only 1 orderline left, and the client changes the amount of the orderline to 0, Vendure returns an error:
`Float cannot represent non numeric value: -Infinity`

**To Reproduce**
1. Add an item to order
2. Execute query (see context for exact query) `adjustOrderLine` with quantity 0
3. See error `Float cannot represent non numeric value: -Infinity`

**Expected behavior**
I would expect the query to return an order without any orderLines

**Environment (please complete the following information):**
 - @vendure/core version: 1.6.4
 - Nodejs version 16
 - Database (mysql/postgres etc): mysql

**Additional context**
Exact error message:
```
{
  "errors": [
    {
      "message": "Float cannot represent non numeric value: -Infinity",
      "locations": [
        {
          "line": 60,
          "column": 13
        }
      ],
      "path": [
        "removeAllOrderLines",
        "taxSummary",
        0,
        "taxRate"
      ]
    }
  ],
  "data": null
}
```
Query used:
```
fragment OrderFields on Order {
  id
  code
  state
  active
  totalWithTax
  subTotalWithTax
  shippingWithTax
  customer {
    id
    firstName
    lastName
    phoneNumber
    emailAddress
  }
  shippingAddress {
    fullName
    company
    streetLine1
    streetLine2
    city
    postalCode
    country
  }
  billingAddress {
    fullName
    company
    streetLine1
    streetLine2
    city
    postalCode
    country
  }
  shippingLines {
    shippingMethod {
      id
      code
      name
    }
  }
  lines {
    id
    quantity
    linePriceWithTax
    featuredAsset {
      id
      preview
      thumbnail
    }
    productVariant {
      id
      sku
      name
      priceWithTax
    }
  }
  taxSummary {
    taxRate
    taxTotal
    taxBase
  }
  payments {
    id
    state
    errorMessage
    metadata
  }
  discounts {
    description
    amountWithTax
  }
  couponCodes
}

mutation adjustOrderLine($orderLineId: ID!, $quantity: Int!) {
  adjustOrderLine(orderLineId: $orderLineId, quantity: $quantity) {
    ... on Order {
      ...OrderFields
    }
    ... on ErrorResult {
      errorCode
      message
    }
  }
}
```

This bug writes the following line to the shippingLines table:
`[{"description":"shipping tax","taxRate":null}]`
[...]

Issue #1331, opened 2022-01-07
Admin version is always latest


zgaduj:
**Describe the bug**
When new version of Vendure was released and my CI/CD creating new image I see every time newest version.
My package.json have versions:
```
    "@vendure/admin-ui-plugin": "1.4.2",
    "@vendure/asset-server-plugin": "1.4.2",
    "@vendure/core": "1.4.2",
    "@vendure/email-plugin": "1.4.2",
    "@vendure/ui-devkit": "1.4.2",
```
And yarn.lock:
```
"@vendure/admin-ui-plugin@1.4.2":
  version "1.4.2"
  resolved "https://registry.yarnpkg.com/@vendure/admin-ui-plugin/-/admin-ui-plugin-1.4.2.tgz#a1633d050c6b95c4a55fde394a45b6bb6156914b"
  integrity sha512-....==
  dependencies:
    fs-extra "^10.0.0"
```
So why in Admin UI I have 1.4.3? Somebody can explain this?

Now in Production is mismatch versions.. Very good when I have e.g. `1.3.4` version and admin `1.4.3` :D 
Everywhere Errors like a fireworks

michaelbromley:
Hi,

This is very strange, and I have no explanation for it. The npm packages are immutable, so there is no way a new release could have altered the contents of v1.4.2.

If you can, try running `yarn why @vendure/admin-ui-plugin` and see if there are somehow multiple versions being installed.

zgaduj:
After deleted node_modules, yarn.lock, admin-ui folder and `yarn install`:
```
root@2c4e8fdffbcd:/home/node/app# yarn why @vendure/admin-ui-plugin
yarn why v1.22.17
[1/4] Why do we have the module "@vendure/admin-ui-plugin"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "@vendure/admin-ui-plugin@1.3.4"
info Has been hoisted to "@vendure/admin-ui-plugin"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "32.11MB"
info Disk size with unique dependencies: "32.41MB"
info Disk size with transitive dependencies: "32.53MB"
info Number of shared dependencies: 4
Done in 0.77s.
```

michaelbromley:
Am I understanding this correctly:

* Your package.json and lockfile show v1.4.2
* but yarn is installing v1.3.4
[...]

PR #1687, opened 2022-07-26
fix(ui-devkit): pin version

martijnvdbrug:
I tried downgrading to 1.6.2 because of [an issue](https://github.com/vendure-ecommerce/vendure/issues/1685), but my admin kept saying I was running 1.6.4, while I definitely downgraded to 1.6.2 for all my dependencies. Turned out to be the transient dependency `admin-ui`  in `ui-devkit`.

By pinning this version, the ui-devkit will always use its corresponding `common` and `admin-ui` versions, and not the latest patch.


martijnvdbrug:
@michaelbromley could you try and rerun the tests? I think it's highly unlikely that pinning 2 dependencies causes the e2e to fail...

michaelbromley:
I think this issue is the same as what I discuss in this comment: https://github.com/vendure-ecommerce/vendure/issues/1331#issuecomment-1013226791

In short, this change will get overwritten by Lerna next time we publish a new version.

martijnvdbrug:
The issue does mention that this should work without range prefix, so no `~` or `^` combined with `--exact`. Lerna will still update the dependencies in consuming projects.

Shall I add `--exact` to this PR?

Just tested:
image
Ran `yarn lerna publish --exact`
Result:
image

martijnvdbrug:
@michaelbromley Could this be the result of the lerna `--exact` flag? 
`The Angular CLI requires a minimum Node.js version of either v14.15, or v16.10.` It only fails for windows with node 12.x.

If so, lets close this. Too much hassle for a very small, debatable improvement

michaelbromley:
Hi Martijn,
Thanks for the extra investigation in this one. The `--exact` flag looks promising in this case. I'm just wondering whether there might be any unforeseen effects of switching to exact versions in all Vendure packages now. E.g. if some plugin specifies a peerDependency with a `^` range, and the core packages specify an exact version, could that lead to issues? I think it's very hard to know since the interactions of all those dependencies get really complex pretty quickly.

Let me sit on this one a bit, I need some time to properly investigate. But I'll prioritize that Admin UI issue for the next patch release though :)

martijnvdbrug:
@michaelbromley thanks for the response. I'm also a bit hesitant about any side effects for this one... Let's leave this PR open for reference. Its not a pressing issue for me, I worked around it by adding this to the package.json: 
```json
  "resolutions": {
    "@vendure/admin-ui": "1.6.2",
    "@vendure/common": "1.6.2"
  },
  ```

martijnvdbrug:
Closing this, we can still lookup closed PR's :)

Issue #1683, opened 2022-07-25
OrderItemPriceCalculationStrategy doesn't receive relation custom fields


jacobfrantz1:
**Describe the bug**
I'm using a relation list custom field on OrderLine, but this field doesn't show up in the orderLineCustomFields parameter of calculateUnitPrice on the OrderItemPriceCalculationStrategy.

Additionally, relation list customFields are not differentiating order lines. They are stacking on the same order line even though the relation customFields are different.

**Expected behavior**
I would expect to see relation custom fields hydrated/passed to the calculateUnitPrice function

**Environment (please complete the following information):**
 - @vendure/core version: 1.6.4
 - Nodejs version 16.14.2
 - Database: Postgres 14

PR #1690, opened 2022-07-27
Fix orderline customfield relations

jacobfrantz1:
Fixes issue #1683 

I eagerly loaded OrderLine customFields in certain places.
Changed OrderLine customField equality function
Added tests

I was finding a lot of edge cases when I was programming this. This might not fix all of them.

### DIFF
--- a/packages/core/e2e/shop-order.e2e-spec.ts
+++ b/packages/core/e2e/shop-order.e2e-spec.ts
@@ -130,6 +130,7 @@ describe('Shop orders', () => {
                     { name: 'notes', type: 'string' },
                     { name: 'privateField', type: 'string', public: false },
                     { name: 'lineImage', type: 'relation', entity: Asset },
+                    { name: 'lineImages', type: 'relation', list: true, entity: Asset },
                     { name: 'dropShip', type: 'boolean', defaultValue: false },
                 ],
             },
@@ -280,6 +281,9 @@ describe('Shop orders', () => {
                                 lineImage {
                                     id
                                 }
+                                lineImages {
+                                    id
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


michaelbromley:
Hi!

Thanks for this PR - I can see a lot of careful work went into it 👍 

A couple of questions:

1. You added the OrderLine custom fields to the default relations of `OrderService.findOne()`. When an order is accessed via the GraphQL API, these relations will be resolved by auto-generated field resolvers for custom field relations. Are you covering the case of direct usage of the `fineOne()` method? If so, is there an argument for allowing the consumer to manually specify the `relations` argument rather than always loading OrderLine relations by default?
2. Was the custom field comparison incorrectly handling list custom field relations before? I can see you added to the e2e tests for this and changed the logic - would your new tests fail prior to your changes to `OrderModifier` line 645-654?

michaelbromley:
I found that if you set `eager: true` in your custom field config, it will indeed be joined in the OrderItemPriceCalculationStrategy.

I'm therefore tending to think that we should not default to always joining the custom field relations in `OrderService.findOne()`. And also possibly we shouldn't automatically join them in [OrderModifier](https://github.com/vendure-ecommerce/vendure/pull/1690/files#diff-94b737961f33940377b7c04051fbe103945a11b609a44d6f977c33bc62fcfac5R143-R145) either. 

The logic fixing the custom field equality check + the e2e tests is most definitely needed, however.

What do you think?

jacobfrantz1:
I agree that we shouldn't eagerly join fields that may not be needed. I wasn't aware of the eager prop on custom field config, and I think that will work well, but we should add documentation to OrderItemPriceCalculationStrategy.

michaelbromley:
Would you like to add a commit which reverts the eager loading part? Then I can merge this PR :)

PR #1696, opened 2022-07-31
fix(default-search-plugin): Language fallback

ashitikov:
**Motivation**
Say you have a product with EN translation. You use default search plugin, and two languages are available: EN, AZ.
Searching product with EN languageCode gives you a product, but when you search AZ, results are empty. 
My expectations around this, to be product available in AZ search, but it should be in default translation, which is determined by channel's `defaultLanguageCode `, so in EN.

**Solution**
Use `LEFT JOIN` clause on the same table, to join identical records but with default language code, if preferred language code and default one are different. 
Example generated query:
```sql
SELECT
  "si"."languageCode" AS "si_languageCode",
  "si"."enabled" AS "si_enabled",
  "si"."productName" AS "si_productName",
  "si"."productVariantName" AS "si_productVariantName",
  "si"."description" AS "si_description",
  "si"."slug" AS "si_slug",
  "si"."sku" AS "si_sku",
  "si"."price" AS "si_price",
  "si"."priceWithTax" AS "si_priceWithTax",
  "si"."facetIds" AS "si_facetIds",
  "si"."facetValueIds" AS "si_facetValueIds",
  "si"."collectionIds" AS "si_collectionIds",
  "si"."collectionSlugs" AS "si_collectionSlugs",
  "si"."channelIds" AS "si_channelIds",
  "si"."productPreview" AS "si_productPreview",
  "si"."productPreviewFocalPoint" AS "si_productPreviewFocalPoint",
  "si"."productVariantPreview" AS "si_productVariantPreview",
  "si"."productVariantPreviewFocalPoint" AS "si_productVariantPreviewFocalPoint",
  "si"."inStock" AS "si_inStock",
  "si"."productInStock" AS "si_productInStock",
  "si"."productVariantId" AS "si_productVariantId",
  "si"."channelId" AS "si_channelId",
  "si"."productId" AS "si_productId",
  "si"."productAssetId" AS "si_productAssetId",
  "si"."productVariantAssetId" AS "si_productVariantAssetId"
FROM
  "search_index_item" "si"
  LEFT JOIN "search_index_item" "sil" ON "si"."productVariantId" = "sil"."productVariantId"
  AND "si"."channelId" = "sil"."channelId"
  AND "si"."productId" = "sil"."productId"
  AND "si"."productAssetId" = "sil"."productAssetId"
  AND "si"."productVariantAssetId" = "sil"."productVariantAssetId"
  AND "si"."languageCode" != "sil"."languageCode"
  AND "sil"."languageCode" = 'az'
WHERE
  1 = 1
  AND "si"."languageCode" IN ('az, 'en')
  AND "sil"."languageCode" IS NULL
  AND "si"."channelId" = 2
ORDER BY
  "si"."productVariantId" ASC
```
**Implementation Notes**
Because of join, we should use prefix `si.` everywhere, this was already has been done, except a few places. 

**Additional Context**
Slack thread: https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1658917239434719

### DIFF
--- a/packages/core/e2e/default-search-plugin.e2e-spec.ts
+++ b/packages/core/e2e/default-search-plugin.e2e-spec.ts
@@ -1541,7 +1541,7 @@ describe('Default search plugin', () => {
                     SEARCH_PRODUCTS,
                     {
                         input: {
-                            take: 1,
+                            take: 100,
                         },
                     },
                     {
@@ -1592,25 +1592,34 @@ describe('Default search plugin', () => {
                 await awaitRunningJobs(adminClient);
             });
 
+            it('fallbacks to default language', async () => {
+                const { search } = await searchInLanguage(LanguageCode.af)
+                // No records for AF language, but we expect > 0
+                // because of fallback to default language (EN)
+                expect(search.totalItems).toBeGreaterThan(0);
+            })
+
             it('indexes product-level languages', async () => {
                 const {search: search1} = await searchInLanguage(LanguageCode.de);
 
-                expect(search1.items[0].productName).toBe('laptop name de');
-                expect(search1.items[0].slug).toBe('laptop-slug-de');
-                expect(search1.items[0].description).toBe('laptop description de');
+                expect(search1.items.map(i => i.productName)).toContain('laptop name de');
+                expect(search1.items.map(i => i.productName)).not.toContain('laptop name zh')
+                expect(search1.items.map(i => i.slug)).toContain('laptop-slug-de');
+                expect(search1.items.map(i => i.description)).toContain('laptop description de');
 
                 const {search: search2} = await searchInLanguage(LanguageCode.zh);
 
-                expect(search2.items[0].productName).toBe('laptop name zh');
-                expect(search2.items[0].slug).toBe('laptop-slug-zh');
-                expect(search2.items[0].description).toBe('laptop description zh');
+                expect(search2.items.map(i => i.productName)).toContain('laptop name zh');
+                expect(search2.items.map(i => i.productName)).not.toContain('laptop name de');
+                expect(search2.items.map(i => i.slug)).toContain('laptop-slug-zh');
+                expect(search2.items.map(i => i.description)).toContain('laptop description zh');
             });
 
             it('indexes product variant-level languages', async () => {
                 const {search: search1} = await searchInLanguage(LanguageCode.fr);
 
-                expect(search1.items[0].productName).toBe('Laptop');
-                expect(search1.items[0].productVariantName).toBe('laptop variant fr');
+                expect(search1.items.map(i => i.productName)).toContain('Laptop');
+                expect(search1.items.map(i => i.productVariantName)).toContain('laptop variant fr');
             });
         });
     });
--- a/packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
+++ b/packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
@@ -41,7 +41,7 @@ export class MysqlSearchStrategy implements SearchStrategy {
         const facetValuesQb = this.connection
             .getRepository(ctx, SearchIndexItem)
             .createQueryBuilder('si')
-            .select(['MIN(si.productId)', 'MIN(productVariantId)'])
+            .select(['MIN(si.productId)', 'MIN(si.productVariantId)'])
             .addSelect('GROUP_CONCAT(si.facetValueIds)', 'facetValues');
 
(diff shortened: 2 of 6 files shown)


packages/core/e2e/default-search-plugin.e2e-spec.ts
```diff
@@ -1592,25 +1592,34 @@ describe('Default search plugin', () => {
...
+                expect(search.totalItems).toBeGreaterThan(0);
+            })
+
             it('indexes product-level languages', async () => {
                 const {search: search1} = await searchInLanguage(LanguageCode.de);
 
-                expect(search1.items[0].productName).toBe('laptop name de');
-                expect(search1.items[0].slug).toBe('laptop-slug-de');
-                expect(search1.items[0].description).toBe('laptop description de');
+                expect(search1.items.map(i => i.productName)).toContain('laptop name de');
```
michaelbromley, line 1605: What's the reason for no longer using the zero'th index here? Does the new query alter the order of results?

packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
```diff
@@ -40,8 +41,8 @@ export class MysqlSearchStrategy implements SearchStrategy {
         const facetValuesQb = this.connection
             .getRepository(ctx, SearchIndexItem)
             .createQueryBuilder('si')
-            .select(['MIN(productId)', 'MIN(productVariantId)'])
-            .addSelect('GROUP_CONCAT(facetValueIds)', 'facetValues');
+            .select(['MIN(si.productId)', 'MIN(productVariantId)'])
```
michaelbromley: It's missing the `si.` in front of the `productVariantId` here I believe (this caused an error in manual testing I did).

michaelbromley (commented):
Great work! See comments.

So overall I think this is more desirable behaviour, but it _is_ quite a change in how the product list works, and I'm wondering whether it makes more sense to introduce it as part of v1.7 as a feature, rather than a fix. In which case the PR should be made against the `minor` branch.

This change will also make the DefaultSearchPlugin behave differently to the ElasticsearchPlugin, so I think we will need to update that one too to also fall back to the default translations. I'm not sure if you have experience with ES but let me know if you'd like to take that on too.

packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
ashitikov: Yeah, good catch, I don't have mysql environment for proper testing. I had a hope that e2e test will cover everything.

packages/core/e2e/default-search-plugin.e2e-spec.ts
ashitikov: The reason is that prior to this change there was only one search record, because there was no fallback on default language. So checking with 0 index may return a row in default language, rather than yours. I don't think this can be counted as altering of result's order, rather than additional search results.

ashitikov: About 1.7: I partially agree with you, it introduces slightly a new behavior, but current behavior is buggy in term of searching on multi-language environment. Imagine you work with any i18n solution: when you request a translation by key, but you don't have a translation for that, i18n will just return you a default translation, this is some kind of fallback. Also I can say that other graphql queries, like `products()`, `collections()` successfully fallbacks to default language if translation is missing. 

If you are intended to merge this anyway in 1.7, just tell me, I'll switch the branch

ashitikov: ```This change will also make the DefaultSearchPlugin behave differently to the ElasticsearchPlugin, so I think we will need to update that one too to also fall back to the default translations. I'm not sure if you have experience with ES but let me know if you'd like to take that on too.```
I'm new to ES, maybe there are somebody other people who could work around this :)

michaelbromley:
> but current behavior is buggy in term of searching on multi-language environment. Imagine you work with any i18n solution: when you request a translation by key, but you don't have a translation for that, i18n will just return you a default translation, this is some kind of fallback. Also I can say that other graphql queries, like products(), collections() successfully fallbacks to default language if translation is missing.

Thanks for your input on this. It might be a moot point because it may be that the next version is 1.7 anyway - it's almost ready. I'm going to leave this PR open for now until I decide what I want to do, but I'm happy with the work so thank you again!

ashitikov:
Hey, any news regarding this PR? I changed branch to minor :)

michaelbromley:
I was actually planning to merge just before releaseing v1.7, but since you now target `minor` I'll merge it right away :) 

Thank you for this excellent work!

PR #1721, opened 2022-08-17
fix(core): Prevent negative order total with fixed discounts

### LINKED ISSUE #1717: Negative order total with fixed amount discount
**Describe the bug**
When a promotion with a fixed discount is greater than the order total, the order total becomes negative.

**To Reproduce**
(this also works on the vendure demo env)
1. Create a promotion that discounts an order by $500
2. Add an item to cart with a price below $500
3. In your cart, or active order, the total price will be -$400

**Expected behavior**
I would expect the order total to be 0 (debatable).

In my case, the Mollie payment fails, because negative. I could fix this in the payment handler, but it might be expected as order total?

 - Nodejs version 16

**Additional context**
image
image

martijnvdbrug:
Prevent negative order total when using fixed discounts. Resolve #1717

### DIFF
--- a/packages/core/src/config/promotion/actions/order-fixed-discount-action.ts
+++ b/packages/core/src/config/promotion/actions/order-fixed-discount-action.ts
@@ -13,7 +13,7 @@ export const orderFixedDiscount = new PromotionOrderAction({
         },
     },
     execute(ctx, order, args) {
-        return -Math.min(args.discount, order.totalWithTax);
+        return -Math.min(args.discount, order.total);
     },
     description: [{ languageCode: LanguageCode.en, value: 'Discount order by fixed amount' }],
 });
--- a/packages/core/src/config/promotion/actions/order-fixed-discount-action.ts
+++ b/packages/core/src/config/promotion/actions/order-fixed-discount-action.ts
@@ -13,7 +13,7 @@ export const orderFixedDiscount = new PromotionOrderAction({
         },
     },
     execute(ctx, order, args) {
-        return -args.discount;
+        return -Math.min(args.discount, order.totalWithTax);
     },
     description: [{ languageCode: LanguageCode.en, value: 'Discount order by fixed amount' }],
 });
--- a/packages/dev-server/dev-config.ts
+++ b/packages/dev-server/dev-config.ts
@@ -66,8 +66,8 @@ export const devConfig: VendureConfig = {
             assetUploadDir: path.join(__dirname, 'assets'),
         }),
         DefaultSearchPlugin.init({ bufferUpdates: true, indexStockStatus: false }),
-        BullMQJobQueuePlugin.init({}),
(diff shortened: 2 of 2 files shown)


packages/dev-server/dev-config.ts
```diff
@@ -66,8 +66,8 @@ export const devConfig: VendureConfig = {
             assetUploadDir: path.join(__dirname, 'assets'),
         }),
         DefaultSearchPlugin.init({ bufferUpdates: true, indexStockStatus: false }),
-        BullMQJobQueuePlugin.init({}),
-        // DefaultJobQueuePlugin.init(),
```
martijnvdbrug, line 70: With this line, I am not able to start the dev-server locally without errors. (Docs don't mention anything about a local redis instance)

packages/dev-server/dev-config.ts
```diff
@@ -110,7 +110,7 @@ function getDbConfig(): ConnectionOptions {
         case 'sqlite':
             console.log('Using sqlite connection');
             return {
-                synchronize: false,
```
martijnvdbrug, line 113: With a clean setup (without local .sql files) the dev-server won't start without synchronize

packages/dev-server/dev-config.ts
michaelbromley: Yeah that's fine. I just use the BullMQ job queue when I develop because if I ever need to debug DB queries, it is was easier without the polling noise caused by the DefaultJobQueuePlugin.

packages/core/src/config/promotion/actions/order-fixed-discount-action.ts
```diff
@@ -13,7 +13,7 @@ export const orderFixedDiscount = new PromotionOrderAction({
         },
     },
     execute(ctx, order, args) {
-        return -args.discount;
+        return -Math.min(args.discount, order.total);
```
michaelbromley, line 16: ah you use the `order.total` rather than `totalWithTax` because the promotions all operate pre-tax, right? Otherwise we could still end up negative. Is that the case?

martijnvdbrug: Correct! Also tested with both `channel.pricesIncludeTax=true` and `channel.pricesIncludeTax=false` to be sure

martijnvdbrug:
Windows publishing fails again because of some network error :( @michaelbromley can you rerun failed jobs?

michaelbromley:
Yeah the publish job can be flaky but we can safely ignore it when just a single one fails. Usual networking issues on GH's side.

Just check my second comment and confirm I'm understanding correctly 👍

martijnvdbrug:
@michaelbromley Done! (Thought I already replied, but maybe I had to 'submit review' first)

michaelbromley:
Thanks Martijn!

PR #1722, opened 2022-08-17
docs: Docs improvement

martijnvdbrug:
Even after having contributed a few times already, I am still struggling with testing changes locally. This doc improvement might help other contributors.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,26 @@ DB=<mysql|postgres|sqlite> yarn dev-server:start
 ```
 If you do not specify the `DB` argument, it will default to "mysql".
 
-### 6. Launch the admin ui
+### 6. Testing your changes locally
+This example tests the changes of the `core` package locally, but it will also work for other packages.
+Open 2 terminal windows: 
+- Terminal 1 for watching and compiling the changes of the package you are developing
+- Terminal 2 for running the dev-server 
+```
+# Terminal 1
+cd packages/core
+yarn watch
+```
+After the changes in your package are compiled you have to stop and restart the dev-server:
+```
+# Terminal 2
+cd packages/dev-server
+DB=sqlite yarn start
+```
+The dev-server will now have your local changes from the `core` package.
 
-1. `cd packages/admin-ui`
-2. `yarn start`
-3. Go to http://localhost:4200 and log in with "superadmin", "superadmin"
+
+Lerna links to the `dist` folder of the packages, so you **don't** need to rerun 'yarn bootstrap'
 
 ### Code generation
 


README.md
```diff
@@ -82,11 +82,26 @@ DB=<mysql|postgres|sqlite> yarn dev-server:start
 ```
 If you do not specify the `DB` argument, it will default to "mysql".
 
-### 6. Launch the admin ui
```
martijnvdbrug, line 85: Launching the admin separately isn't needed any more right? The dev-server handles this?

martijnvdbrug:
@michaelbromley failed pipeline because of network again. This might make your live easier: https://stackoverflow.com/questions/71574593/how-to-automatically-retry-github-action-jobs-on-failure
Introduced it to my own e2e tests last week as well

martijnvdbrug:
Tests failed with a timeout `"Exceeded timeout of 10000 ms for a test.`

This PR only changes some .md files, so this PR shouldn't be the cause of e2e test failure...

README.md
michaelbromley: If you want to actually make & test changes to the Admin UI, then yes, you still need to run in from the /admin-ui package. But if you are just using the admin ui to test server features, then you don't need to run it separately.

So we should still include those instructions, but make it clear that it is only needed when working on the admin ui itself.

README.md
```diff
@@ -82,11 +82,26 @@ DB=<mysql|postgres|sqlite> yarn dev-server:start
...
+### 6. Testing your changes locally
+This example shows how to test changes to the `core` package locally, but it will also work for other packages.
+1. Open 2 terminal windows: 
+- Terminal 1 for watching and compiling the changes of the package you are developing
+- Terminal 2 for running the dev-server 
+```
+# Terminal 1
+cd packages/core
+yarn watch
```
michaelbromley, line 107: Changes to GraphQL APIs & shared code in the /common package will not get picked up by this watch. Therefore there is a `watch:common-core` script in the root package.json which will watch both. That's the one I use for core dev.

michaelbromley (changes requested):
Good idea to improve these docs! I left a couple of comments

README.md
martijnvdbrug: Ahh oke, so for changes to core we use `watch:core-common` and for all other packages we can do `cd packages/<package> && yarn watch` ?

README.md
martijnvdbrug: Good point, I will add it to the docs again

README.md
michaelbromley: Yes 👍

michaelbromley:
Looks good now, thanks! 😄

PR #1729, opened 2022-08-22
Feat(core): Slug unique per channel instead of globally unique

### LINKED ISSUE #1720: Unique slug per channel instead of globally unique
When channel-A defined the slug for a collection or product as `clothing`, channel-B cannot use this slug anymore.

I would like slugs to be unique per channel. I cannot think of a use case where slugs should be globally unique.

Just using the Vendure generated slug `clothing-2`

**Additional context**
This goes for both products and collections.

martijnvdbrug:
Check for existence of a slug in the current channel instead of globally. Resolves #1720

### DIFF
--- a/packages/admin-ui/tsconfig.json
+++ b/packages/admin-ui/tsconfig.json
@@ -21,7 +21,7 @@
       "./typings"
     ],
     "lib": [
-      "es2017",
+      "es2019",
       "dom",
       "esnext.asynciterable"
     ],
--- a/packages/core/e2e/collection.e2e-spec.ts
+++ b/packages/core/e2e/collection.e2e-spec.ts
@@ -5,7 +5,7 @@ import {
     facetValueCollectionFilter,
     variantNameCollectionFilter,
 } from '@vendure/core';
-import { createTestEnvironment } from '@vendure/testing';
+import { createTestEnvironment, E2E_DEFAULT_CHANNEL_TOKEN } from '@vendure/testing';
 import gql from 'graphql-tag';
 import path from 'path';
 
@@ -17,9 +17,11 @@ import { productIdCollectionFilter, variantIdCollectionFilter } from '../src/ind
 import { COLLECTION_FRAGMENT, FACET_VALUE_FRAGMENT } from './graphql/fragments';
 import {
     Collection,
+    CreateChannel,
     CreateCollection,
     CreateCollectionInput,
     CreateCollectionSelectVariants,
+    CurrencyCode,
     DeleteCollection,
     DeleteProduct,
     DeleteProductVariant,
@@ -47,6 +49,7 @@ import {
     UpdateProductVariants,
 } from './graphql/generated-e2e-admin-types';
 import {
+    CREATE_CHANNEL,
     CREATE_COLLECTION,
     DELETE_PRODUCT,
     DELETE_PRODUCT_VARIANT,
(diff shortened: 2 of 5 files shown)


packages/admin-ui/tsconfig.json
```diff
@@ -21,7 +21,7 @@
       "./typings"
     ],
     "lib": [
-      "es2017",
```
martijnvdbrug: Not sure why this is here. This has already been merged into the minor branch

packages/core/e2e/collection.e2e-spec.ts
```diff
@@ -538,9 +587,17 @@ describe('Collection resolver', () => {
...
+                        id: 'T_1',
+                        name: '__root_collection__',
+                    },
+                },
+                {
+                    id: 'T_11',
                     name: 'Pear Breadcrumbs',
                     parent: {
-                        id: 'T_8',
+                        id: 'T_9',
```
martijnvdbrug, line 600: A new collection has been created in the second channel, so all ID's are +1. This test also orders by ID, so `T_9` becomes `T_10`, which is then placed at the beginning of the array.

packages/core/src/service/helpers/slug-validator/slug-validator.ts
```diff
@@ -29,7 +30,7 @@ export type TranslationEntity = VendureEntity & {
     id: ID;
     languageCode: LanguageCode;
     slug: string;
-    base: any;
+    base: Collection | Product;
```
martijnvdbrug: We assume our TranslationEntity type to have `slug`, which means it can only be a Collection or Product I think?

michaelbromley: This would cause an issue if someone wanted to use this class in an external plugin, where they might have defined a custom entity which follows the same pattern. It would in fact be a breaking change.

Perhaps there is a way to express this more generally, but I'd say spending a lot of time getting a type perfectly right here is not high priority, since we only actually reference it here: https://github.com/vendure-ecommerce/vendure/blob/834c05c514ad2abfa94f56afa0560c3e205ea2ca/packages/core/src/service/helpers/slug-validator/slug-validator.ts#L81

packages/core/e2e/collection.e2e-spec.ts
michaelbromley: Yep that's fine

michaelbromley:
I answered a couple of your comments. It looks like you might have to pull the `minor` branch locally and `git rebase origin/minor` to get rid of commit [9e96fb2](https://github.com/vendure-ecommerce/vendure/pull/1729/commits/9e96fb2c0685280ed682adebf20b4fb0c8d9b636) from this PR.

Also don't commit your dev-config.ts file.

packages/core/src/service/helpers/slug-validator/slug-validator.ts
martijnvdbrug: Good point. For some reason, I thought slugvalidator was only used internally. Reverted :ok_hand:

martijnvdbrug:
Fixed, ready for re-review

michaelbromley:
Thanks, this is good!

It looks like one of my earlier changes introduced the failure for the postgres e2e tests, nothing to do with your stuff.

Issue #1735, opened 2022-08-24
Cannot transition an order from AddingItems to custom State if there is no LineItems


tecoad:
**Describe the bug**
Cannot transition an order from AddingItems to custom State if there is no LineItems

**To Reproduce**
1. Create a custom mutation to be able to create an order without having to add orderitem.
2. Create a custom orderState to transition from 'addingItems'
3. Try to change the orderState from addingItems to the custom orderState created
4. Would will get a transition error "Syntax error at or near \")\"","

**Expected behavior**
Expect to be able to transition the order 

**Environment (please complete the following information):**
 - @vendure/core version: latest 1.6.1
 - Nodejs version 16
 - Database (mysql/postgres etc): postgress

@ashitikov has helped to check the bug at Slack:
``
Happened on 119 line. Because you have no order lines at all, variantIds  is an empty array, which get spreaded, so a query becomes like variant.id IN () , but your database is waiting for an elements inside IN expression.  
``
``
Fix is quite simple for this: We just should check whether order lines empty, and if not empty, execute query and check, otherwise continue
``

PR #1736, opened 2022-08-24
fix(core): Transition from AddingItems to custom state with an empty order lines

ashitikov:
Fixes this: https://github.com/vendure-ecommerce/vendure/issues/1735
Slack thread: https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1661264397433399

In case of spreading an empty array into query builder, typeorm leaves an empty brackets in clause: `IN ()`, which is interpreted by database as a syntax error. To accomplish that, an additional check added.


michaelbromley:
🙌

PR #1737, opened 2022-08-24
fix(core): Translation don't use DefaultLanguage of Channel (#1730)

### LINKED ISSUE #1730: Translation don't use DefaultLanguage of Channel
**Describe the bug**
If we ask both shop and admin with language, that don't exist, we always get english translation (not the default translation of the channel).

Bug exists here:
https://github.com/vendure-ecommerce/vendure/blob/dc3a4f035baf4fd68c011387f679bad0b3e0e394/packages/core/src/service/helpers/utils/translate-entity.ts#L47

**To Reproduce**
1. Create channel with default language != `en`
2. Translate some products assigned to channel
3. Fire any api method with languageCode that don't exist and get `en` translation

**Expected behavior**
Get the translation with default language for channel

 - @vendure/core version: v1.6.5

Izayda:
Fix #1730

### DIFF
--- a/packages/core/src/service/helpers/utils/translate-entity.ts
+++ b/packages/core/src/service/helpers/utils/translate-entity.ts
@@ -9,20 +9,20 @@ import { VendureEntity } from '../../../entity/base/base.entity';
 // prettier-ignore
 export type TranslatableRelationsKeys<T> = {
     [K in keyof T]: T[K] extends string ? never :
-    T[K] extends number ? never :
-    T[K] extends boolean ? never :
-    T[K] extends undefined ? never :
-    T[K] extends string[] ? never :
-    T[K] extends number[] ? never :
-    T[K] extends boolean[] ? never :
-    K extends 'translations' ? never :
-    K extends 'customFields' ? never : K
+        T[K] extends number ? never :
+            T[K] extends boolean ? never :
+                T[K] extends undefined ? never :
+                    T[K] extends string[] ? never :
+                        T[K] extends number[] ? never :
+                            T[K] extends boolean[] ? never :
+                                K extends 'translations' ? never :
+                                    K extends 'customFields' ? never : K
 }[keyof T];
 
 // prettier-ignore
 export type NestedTranslatableRelations<T> = {
     [K in TranslatableRelationsKeys<T>]: T[K] extends any[] ?
-        [K, TranslatableRelationsKeys<UnwrappedArray<T[K]>>]:
+        [K, TranslatableRelationsKeys<UnwrappedArray<T[K]>>] :
         [K, TranslatableRelationsKeys<T[K]>]
 };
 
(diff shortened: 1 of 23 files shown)


packages/core/src/service/helpers/utils/translate-entity.ts
```diff
@@ -84,11 +89,12 @@ export function translateEntity<T extends Translatable & VendureEntity>(
 export function translateDeep<T extends Translatable & VendureEntity>(
     translatable: T,
     languageCode: LanguageCode,
+    channelLanguageCode: LanguageCode,
```
michaelbromley: This is going to be a breaking change for anyone using translateDeep externally. The function _is_ exported from `@vendure/core` but its not documented anywhere in the docs. So I think it's ok to do this, but I'll just add a note to the changelog just in case.

michaelbromley: Thinking about it more, maybe we should rename to `fallbackLanguageCodes` and accept an array of LanguageCode.
This allows us to have more fallback mechanisms in the future without changing the API again. E.g. at the call site we would be able to provide `[ctx.languageCode, config.defaultLanguageCode]`. What do you think?

michaelbromley (commented):
See my comment. What do you think about the API with `fallbackLanguageCodes: LanguageCode[]` ?

packages/core/src/service/helpers/utils/translate-entity.ts
Izayda: Yeah, i think providing array will be good idea.

So, i can rework it now and change channelLanguageCode to fallbackLanguageCodes array (that will be empty for default). 

But i have 2 questions:
1) it seems that channelLanguageCode should be required param. In other way somebody can forget about providing channel default language and we will receive new bug. fallbackLanguageCodes will not solve this issue. Maybe this should be some structure of {languageCode: ctx.LanguageCode, fallbackLanguages: [], channelLanguage: ctx.channel.languageCode, defaultLanguage: config.defaultLanguageCode}
2) How can we provide config.defaultLanguageCode to translateDeep function? Every call of this function should pass this param? Or maybe function should read config independently?

michaelbromley: How about this:

1. We give `translateDeep` a very generic API that accepts a singe language tuple argument:
```ts
export function translateDeep<T extends Translatable & VendureEntity>(
    translatable: T,
    languageCodes: [LanguageCode, ...LanguageCode[]],
    translatableRelations: DeepTranslatableRelations<T> = [],
): Translated<T> {
    // ...
}
```
This means it _must_ be supplied with at least 1 languageCode. Then inside `translateEntity()` we just loop over the languageCodes tuple to find the first match.

2. In order to make this not error-prone in the services that call it, we define a new helper service which wraps it in a very simple API

```ts
const translated = this.translator.translate(product, ctx);
```
the `Translator.translate()` method will take care of supplying the correct sequence of fallbacks:

```ts
class Translator {

  translate(
    translatable: T,
    ctx: RequestContext,
    translatableRelations: DeepTranslatableRelations<T> = [],
  ) {
    return translateDeep(translatable, [
      ctx.languageCode, 
      ctx.channel.defaultLanguageCode,
      this.configService.defaultLanguageCode,
     // we can even add more here, e.g. system locale detection?
    ], translatableRelations);
  }
}
```

The downside is that we now need to inject the `Translator` class into all the services that use it (16 I think). The upside is that the exposed API is was simpler whilst also letting us provide a really robust fallback mechanism in 1 place. Also easy to change in future.

What do you think?

michaelbromley: Further thought - in this way we can even make the `translateDeep` API change non-breaking by making the `languageCode` argument of type `LanguageCode | [LanguageCode, ...LanguageCode[]]`.

Izayda: Ok, i will try to implement right now

Izayda:
Fixed! But there are some depreciated methods, that still use translateDeep
image

michaelbromley:
I love this new API! It was a pleasure to collaborate and brainstorm ideas for this design with you. Great work 🤝

Issue #1725, opened 2022-08-19
Create methods for assign/remove collections/facets to/from channels


Izayda:
**Is your feature request related to a problem? Please describe.**
I want to extend my current product/variants/facets and collections with multiple languages and currencies. Current way is to use `Translations` and `Channels`. But there no way to assign facets and collection to multiple channels through admin-api.

**Describe the solution you'd like**
New 4 methods in admin-api:
- assignFacetToChannel
- removeFacetFromChannel
- assignCollectionToChannel
- removeCollectionFromChannel

**Describe alternatives you've considered**
Use TypeScript Api through some plugin to assign collection/facets to channels.

Izayda:
Thank you!

So, i should pass right relations to findByIds method :)

PR #1740, opened 2022-08-24
feat(core): Add facet/collection channel methods (#1725)

Izayda:
Add assignFacetsToChannel, removeFacetsFromChannel, assignCollectionsToChannel, removeCollectionsFromChannel methods (#1725)


michaelbromley:
Adding some comments from Slack:

---

Izayda: What actions should be triggered in search plugin? Full reindex?

Michael: Interesting question. With Facets, I think we need to consider what happens when we remove a facet from Channel A, but it is still assigned to a Product in Channel A. I just briefly looked at the code and I think maybe we're already missing something here. I need to spend more time on this one.

Izayda: I remember, that we can’t “reverse engineer” what products and variants will be affected by changing some collection/facet, cause custom mapping can be implemented through plugin. So, yeah, it is interesting question 🙂 

---

Current status: I intend to get this into v1.8 but need to resolve the issues raised above.

Izayda:
So, what can i do here along with the resolving merge conflict?

michaelbromley:
OK so how about we handle the above case like this:

- On removing a Facet from a Channel
- we check to see if any Products or ProductVariants in that channel have that Facet assigned
- If yes, we can either (in order of complexity):
    1. Fail with an error
    2. Automatically remove that Facet from the affected Products/Variants
    3. Return an ErrorResult which informs the user that the facet is used in channel products, and then also provide a "force" argument (like we do with the `deleteFacet` mutation currently) that would automatically remove the Facets as in 2 above.

Option 3 is the most complete handling, but would require some changes to the current gql schema to allow a union return type and then a bit more logic for the 2 modes. But the general outline of the implementation can be found in the [deleteFacet method source](https://github.com/vendure-ecommerce/vendure/blob/4900e1eab994b30cc953faa524503e2f0264f50a/packages/core/src/service/services/facet.service.ts#L175-L214).

Izayda:
And what do you think about adding facet/collection to channel? Full reindex? :)

michaelbromley:
Just adding a Facet to a channel should not need to change the search index (should it?). 

But adding a Collection will require not a full reindex but a running of the 'apply-collection-filters' queue for the added Collection. This would be done [here](https://github.com/vendure-ecommerce/vendure/blob/4a9ec5c87ad54351c5f48753f170846f9b912dc9/packages/core/src/service/services/collection.service.ts#L89) in the CollectionService - you can listen for the new `CollectionChannelEvent` and add a job to the queue with just that channel ID.

Removing a Collection will require reindex just for those ProductVariants which are part of the Collection.

Izayda:
Hi @michaelbromley!

I have switched branch from master to minor. Also have added some logic for affected variants/products.

Can you check this code? I'm not sure in using right ctx in every part of code and logic of affected items

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
         this.rootCollection = this.translator.translate(newRoot, ctx);
         return this.rootCollection;
     }
+
+
+    /**
+     * @description
+     * Assigns a Collections to the specified Channel
```
michaelbromley: Should be "Assigns Collections" - no "a".

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+     */
+    async assignCollectionsToChannel(
+        ctx: RequestContext,
+        input: AssignCollectionsToChannelInput,
+    ): Promise<Array<Translated<Collection>>> {
+        const hasPermission = await this.roleService.userHasPermissionOnChannel(
```
michaelbromley, line 728: I realized we also have to check the UpdateCollection permission. Background: I started off with just Catalog permission, that covers products, assets, collections, facets. Later I realized it needs to be more granular so created individual CRUD permissions for each entity. So now we need to consider both, and make sure User has permissions on at least 1 of them.

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+     * Remove a Collections from the specified Channel
+     */
+    async removeCollectionsFromChannel(
+        ctx: RequestContext,
+        input: RemoveCollectionsFromChannelInput,
+    ): Promise<Array<Translated<Collection>>> {
+        const hasPermission = await this.roleService.userHasPermissionOnChannel(
+            ctx,
+            input.channelId,
+            Permission.UpdateCatalog,
```
michaelbromley: Same thing with also checking UpdateCollection permission.

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+            throw new ForbiddenError();
+        }
+        const collections = await this.connection
+            .getRepository(ctx, Collection)
+            .findByIds(input.collectionIds);
+
+        await Promise.all(
+            collections.map(async collection => {
+                await this.channelService.assignToChannels(ctx, Collection, collection.id, [input.channelId]);
+                return this.eventBus.publish(new CollectionChannelEvent(ctx, collection, input.channelId, 'assigned'));
```
michaelbromley: I noticed that the `ChannelService.assignToChannels()` method publishes a generic `ChangeChannelEvent` which actually contains all the data of the new `CollectionChannelEvent`. I'm thinking that we can just rely on that event, rather than defining new events especially for Collection and Facet channel changes.

So why do we have a dedicated `AssetChannelEvent` and `ProductChannelEvent`? That's just because they were created _before_ the generic `ChangeChannelEvent`, which basically makes them obsolete but I needed to keep them so as to not create a breaking change.

So I'd suggest removing those 2 new custom events you defined. Less code is always nice!

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+                this.eventBus.publish(new CollectionModificationEvent(ctx, collection, affectedVariantIds));
+                return this.eventBus.publish(new CollectionChannelEvent(ctx, collection, input.channelId, 'removed'));
+            }),
+        );
+
+        return this.connection.findByIdsInChannel(
+            ctx,
+            Collection,
+            collections.map(f => f.id),
+            defaultChannel.id,
```
michaelbromley: I think we should be keeping to the active channel, `ctx.channelId` rather than the default channel.

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+        });
+
+
+        return this.connection.findByIdsInChannel(
+            ctx,
+            Collection,
+            collections.map(f => f.id),
+            ctx.channelId,
+            {},
+        ).then(clcs =>
```
michaelbromley, line 759: I'd personally prefer the var to be named `collections` - this kind of abbreviation is not very clear IMO.

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+     * Assigns a Facets to the specified Channel
+     */
+    async assignFacetsToChannel(
+        ctx: RequestContext,
+        input: AssignFacetsToChannelInput,
+    ): Promise<Array<Translated<Facet>>> {
+        const hasPermission = await this.roleService.userHasPermissionOnChannel(
+            ctx,
+            input.channelId,
+            Permission.UpdateCatalog,
```
michaelbromley: See the comment I left on the CollectionService file - same applies here with the UpdateFacet permission.

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
 
         return candidate;
     }
+
+    /**
+     * @description
+     * Assigns a Facets to the specified Channel
```
michaelbromley: Should be "Assigns Facets" without the "a"

packages/core/src/service/services/collection.service.ts
```diff
@@ -709,4 +716,94 @@ export class CollectionService implements OnModuleInit {
...
+            clcs.map(collection => this.translator.translate(collection, ctx)),
+        );
+    }
+
+    /**
+     * @description
+     * Remove a Collections from the specified Channel
```
michaelbromley: drop the "a" here too

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+            }),
+        );
+
+        return this.connection.findByIdsInChannel(
+            ctx,
+            Facet,
+            facets.map(f => f.id),
+            ctx.channelId,
+            {},
+        ).then(fcts =>
```
michaelbromley: Name this `facets` - for 2 extra chars it's worth the improved readability!

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+            fcts.map(facet => translateDeep(facet, ctx.languageCode)),
+        );
+    }
+
+    /**
+     * @description
+     * Remove a Facets from the specified Channel
```
michaelbromley: drop the "a"

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+     */
+    async removeFacetsFromChannel(
+        ctx: RequestContext,
+        input: RemoveFacetsFromChannelInput,
+    ): Promise<DeletionResponse[]> {
+        const hasPermission = await this.roleService.userHasPermissionOnChannel(
+            ctx,
+            input.channelId,
+            Permission.UpdateCatalog,
```
michaelbromley: also add check for UpdateFacet permission

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+                const both = !!(productCount && variantCount) ? 'both' : 'single';
+                const i18nVars = { products: productCount, variants: variantCount, both };
+                let message = '';
+                let result: DeletionResult;
+
+                if (!isInUse) {
+                    await this.channelService.removeFromChannels(ctx, Facet, facet.id, [input.channelId]);
+                    await this.eventBus.publish(new FacetChannelEvent(ctx, facet, input.channelId, 'removed'));
+                    result = DeletionResult.DELETED;
+                } else if (input.force === undefined ? false : input.force) {
```
michaelbromley: side note - this can be expressed as `if (input?.force)`

packages/core/src/service/services/facet.service.ts
```diff
@@ -240,4 +245,110 @@ export class FacetService {
...
+    }
+
+    /**
+     * @description
+     * Remove a Facets from the specified Channel
+     */
+    async removeFacetsFromChannel(
+        ctx: RequestContext,
+        input: RemoveFacetsFromChannelInput,
+    ): Promise<DeletionResponse[]> {
```
michaelbromley: I see why you've decided to return an array of DeletionResponse, but I don't think we should do it this way. All the other `removeXFromChannel`mutations return an array containing the removed entities. We should attempt to stick as close to that as possible.

In this case I suggest you create a new ErrorResult:

```gql
type FacetInUseError implements ErrorResult {
    errorCode: ErrorCode!
    message: String!
    productCount: Int!
    variantCount: Int!
}
```

and then have the mutation return an array of either Facet or FacetInUseError as a new union type:

```gql
union RemoveFacetFromChannelResult = Facet | FacetInUseError

Mutation {
  removeFacetsFromChannel(input: RemoveFacetsFromChannelInput!): [RemoveFacetFromChannelResult!]!
}
```

This solution is getting rather advanced because now you also need to codegen the FacetInUseError TS class. Please let me know if you'd like my assistance with this part.

michaelbromley (changes requested):
See my comments. Thanks for the work on this so far - reviewing your changes allowed me to wrap my head around the problem fully for the first time. 

I think we're almost there - a few cleanups and the main change is with how to handle the removeFacetsFromChannel return type.

As I mentioned in the comment - I'm happy to collaborate on the last few details so let me know if you would like me to handle any of the remaining parts.

packages/core/src/service/services/collection.service.ts
Izayda: Ok, i also can mark that asset and product event are depreciated

packages/core/src/service/services/collection.service.ts
Izayda: Thanks! I often make mistakes with "a" and "the"

packages/core/src/service/services/collection.service.ts
Izayda: But we already check this permission in resolver. Should we also check this permissions in resolver?

packages/core/src/service/services/collection.service.ts
Izayda: It shadows name of const defined above

packages/core/src/service/services/facet.service.ts
Izayda: I need some help here. I tried to implement, code is in current branch. But have some problems with typification

michaelbromley: No problem. You've got it 90% there - I'll wrap the last bits up and try to get this merged today :)

michaelbromley:
Merged in https://github.com/vendure-ecommerce/vendure/commit/34840c943d5d57d1e652352d89196ceb5dc666f8

Thanks for your help on this! 🙏

PR #1745, opened 2022-08-25
password change checks pw validity

### LINKED ISSUE #1734: Password change does not follow password validation strategy
**Describe the bug**
when requesting a password change, the password isn’t verified against the default validation strategy

**To Reproduce**
1. Set up a default validation strategy
2. request a password change
3. type in new password which is miss-matching the validation strategy
4. submit and log in with weak password

**Expected behavior**
Password change should also use the default validation strategy 

 - Nodejs version 16-lts

**Additional context**

Conor-Burns:
fixes #1734


michaelbromley:
Hey thanks! Just check a couple of build errors: https://github.com/vendure-ecommerce/vendure/runs/8025580665?check_suite_focus=true#step:6:40

I think you just need to also add the `PasswordValidationError` return type to the method that calls it too?

Conor-Burns:
should work now - sorry for the trouble

michaelbromley:
Thanks Conor! 👍

Issue #1750, opened 2022-08-28
Order's deleteOrder method malfunction


saymow:
**Describe the bug**
When deleting an order by passing its id, the order is not loaded with its shippingLines, which causes an exception. 

![image](https://user-images.githubusercontent.com/52419335/187093421-81edc45b-59f6-418a-a394-cd96cbdb125d.png)

**To Reproduce**
1. Inject the OrderService class in your plugin
2. Try to delete an order passing its id

**Expected behavior**
The shipping lines should be loaded and the order deleted normally.

**Environment**
 - @vendure/core version: 1.6.5
 - Nodejs version: 16.16.0
 - Database: postgres

PR #1751, opened 2022-08-28
fix(core): delete order method

saymow:
Fixes issue #1750


michaelbromley:
Thank you @saymow for your first contribution to the Vendure repo! 👍

PR #1793, opened 2022-09-19
fix(core): 'productVariantId' in group statement is ambiguous

william-p:
When you uncheck `Group by product`  in Products view we have this MySQL Error (with red notification in admin interface):

`ERROR 1052 (23000): Column 'productVariantId' in group statement is ambiguous`

This is the generated SQL Query:
```sql
SELECT MIN(`si`.`productId`), MIN(`si`.`productVariantId`), GROUP_CONCAT(`si`.`facetValueIds`) AS `facetValues`, 1 as score, BIT_OR(`si`.`enabled`) AS `productEnabled` FROM `search_index_item` `si` LEFT JOIN `search_index_item` `sil` ON              `si`.`channelId` = `sil`.`channelId` AND `si`.`productVariantId` = `sil`.`productVariantId` AND `si`.`productId` = `sil`.`productId` AND `si`.`productAssetId` = `sil`.`productAssetId` AND `si`.`productVariantAssetId` = `sil`.`productVariantAssetId`             AND `si`.`languageCode` != `sil`.`languageCode`             AND `sil`.`languageCode` = 'en'          WHERE `si`.`languageCode` IN ('en', 'fr') AND `sil`.`languageCode` IS NULL AND `si`.`channelId` = '5c4c2665-118a-483e-85d8-083980424196' GROUP BY productVariantId;
```

### DIFF
--- a/packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
+++ b/packages/core/src/plugin/default-search-plugin/search-strategy/mysql-search-strategy.ts
@@ -46,7 +46,7 @@ export class MysqlSearchStrategy implements SearchStrategy {
 
         this.applyTermAndFilters(ctx, facetValuesQb, { ...input, groupByProduct: true });
         if (!input.groupByProduct) {
-            facetValuesQb.groupBy('productVariantId');
+            facetValuesQb.groupBy('si.productVariantId');
         }
         if (enabledOnly) {
             facetValuesQb.andWhere('si.enabled = :enabled', { enabled: true });


michaelbromley:
Hi, thanks for the PR!

I'm a little confused as, although your change makes sense and looks good, I cannot reproduce the error locally. What version of MySQL are you running?

william-p:
Hi, i have not seen this error before upgrade to 1.7.2 (from 1.7.1) in my dev env.

I have rollback to 1.7.1 with same error, strange. All the time i drop database and re-bootstrap / re-import all my products.

```
MySQL: 10.1.26-MariaDB-1~jessie
Node: v16.16.0

michaelbromley:
OK, very weird. I'm merging anyway because, as I said, the change looks correct and there obviously _is_ some edge-case where it appears to cause an issue, even if we don't understand how to reproduce 😆 

Thanks again!

Issue #1794, opened 2022-09-21
devMode in compileUiExtension throws error on next version


HarunKilic:
**Describe the bug**
When enabling devMode in compileUiExtension, it will throw an error saying: `unknown argument --base-path`

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.0-next.17
 - Nodejs version: 16.17.0
 - Angular version: 13.2.3

**Additional context**
The issue might be because the option is deprecated and should be set in angular.json instead: 
https://stackoverflow.com/questions/71604562/unknown-option-base-href-angular-13-ng-serve

HarunKilic:
I've managed to get the ui part working by: 
* Removing `--base-path=${basePath}` from `\node_modules\@vendure\ui-devkit\compiler\compile.js` as its deprecated now. 
* Changed `"baseHref": "/",` in `\admin-ui\angular.json` to `"baseHref": "/admin/",`

The app will now run successfully on port 4200. 
BUT, it will also try to query on port 4200, which should be on port 3000, where Vendure actually runs.

michaelbromley:
Pasting from Slack conversation to make it public:

---

Thanks for taking the time to try to solve this. In general, running on port 4200 doesn't work due to the way that the Admin UI is trying to infer the api url. It defaults to auto which just uses the browser url to guess. You should be able to explicitly set
```ts
AdminUiPlugin.init({ 
  //..
  adminUiConfig: {
    apiHost: 'http://localhost/'
    apiPort: '3000'
  } 
}),
```

HarunKilic:
Just to sum up for a temporary solution: 

1. Remove `--base-path=${basePath}` from `\node_modules\@vendure\ui-devkit\compiler\compile.js`
2. Change `"baseHref": "/"`, in `\admin-ui\angular.json` to `"baseHref": "/admin/"` (you will have to run the app with `compileUiExtensions` atleast once without `devMode` so that the folder can be generated.)
3. Add  
```
AdminUiPlugin.init({ 
  //..
  adminUiConfig: {
   //..
    apiPort: '3000'
  } 
}),
```

HarunKilic:
To make this a permanent solution, the `--base-path` in the cli should be added to angular.json with dynamic values somehow.

martijnvdbrug:
Same problem. Not sure if still needed, but these are the dependencies installed in my project after removing yarn.lock
```shell
yarn list --pattern @angular
yarn list v1.22.17
[...]

PR #1805, opened 2022-09-27
fix(admin-ui): Temporary fix for deprecated option --base-href

HarunKilic:
This fixes the problem: https://github.com/vendure-ecommerce/vendure/issues/1794

But the --base-href is not dynamic anymore. It's set as `/admin/` in the angular.json.


michaelbromley:
Closing this as it has been fixed in the major branch.

Issue #1821, opened 2022-10-04
Promotion customFields relations not persisted


karel-advantitge:
**Describe the bug**
Relations defined in de customFields config for entity `Promotion` don't work.
Other entities with customFields use the `CustomFieldRelationService` in their respective create and update functions to persist the relations.
The `PromotionService` does not do this in its `createPromotion` and `updatePromotion` functions.

I'll make a PR to fix this issue

PR #1822, opened 2022-10-04
fix(core): persist customField relations in PromotionService

karel-advantitge:
#1821


michaelbromley:
Good catch! Thanks for the fix! 👍

PR #1825, opened 2022-10-05
Feat/mollie paymentmethod selection

### LINKED ISSUE #1809: Make Mollie payment method selectable from storefront
Most users are familiar with selecting the payment method on the stores page and then getting redirected to the payment provider. Right now this is not possible with the Mollie plugin, since we always get redirected to the Mollie method-selector, resulting in a slightly confusing checkout flow (Clicking "Buy now" before selecting payment method, or clicking "Buy now" on mollie.com which _may_ both violate german law).
There is a way to embed the method selection in the storefront described [in the Mollie Docs](https://docs.mollie.com/payments/build-your-own-checkout)

Things that are needed:

1. A new query for getting the available mollie payment methods. [Mollie API Reference](https://docs.mollie.com/reference/v2/methods-api/list-methods)
2. Another CreateMolliePaymentIntent input for setting the mollie method. The method can be set [here](https://github.com/vendure-ecommerce/vendure/blob/master/packages/payments-plugin/src/mollie/mollie.service.ts#L82).
3. Error handling

Only hardcoded, hacky experiments I would rather not share.

**Additional context**
I would be willing to pay for the development of this feature, since I really need this for the optimal checkout flow. I would try it myself but I am not that good with typescript and the Vendure plugin architecture yet.
That being said: I really love Vendure and I am looking forward to the time I finish my new shop and I can swich from WooCommerce to Vendure!

martijnvdbrug:
Allow pre-selecting a payment method before being redirected to Mollie's hosted checkout.
closes #1809


martijnvdbrug:
@michaelbromley Pipeline fails with `Entry point @vendure/admin-ui/catalog has a circular dependency on itself.`. I also have this error locally, but it already occurred before I made any changes. Any ideas what change causes this?

michaelbromley:
@martijnvdbrug there were some bad imports in the admin ui, which were fixed but I hadn't pushed my latest commits to the branch. Just did that now, so you should be able to rebase onto the latest minor and force push an update to this PR 👍

martijnvdbrug:
@michaelbromley fixed

packages/payments-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -64,21 +64,49 @@ export interface MolliePluginOptions {
...
  * platform.
  *
+ * 'molliePaymentMethodCode' is an optional parameter that can be passed to skip Mollie's hosted payment method selection screen
+ * You can get available Mollie payment methods with the following query:
+ *
+ * ```GraphQL
+ * {
+ *  molliePaymentMethods(input: { paymentMethodCode: "mollie-payment-method" }) {
+ *    id
+ *    description
```
michaelbromley, line 92: I think this example is missing the `code` field right?

michaelbromley (commented):
Nice work Martijn! Just a small omission in the documentation I believe.

packages/payments-plugin/src/mollie/mollie.plugin.ts
martijnvdbrug: Not anymore!

michaelbromley:
Thanks! ⚡

PR #1826, opened 2022-10-06
fix(core): Export translator.service from core

alexisvigoureux:
Added the translator service to the index.ts to be able to use it in the implementation.
Will allow it to be imported in `@vendure/core` instead of `@vendure/core/dist/service/helpers/translator/translator.service`


michaelbromley:
Thank you!

PR #1841, opened 2022-10-13
fix(core): Order fixed discount considers channel pricesIncludeTax

Hbowers:
Reference: https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1665539143010169

Check the channel `pricesIncludeTax` flag to determine if the upper bounds on a order fixed discount should be tax inclusive.


michaelbromley:
Thanks Harrison! 👍

PR #1850, opened 2022-10-26
chore: Added nx in packages

cwerner1:
enables nx in the vendure project.
  * Basic dependency setup ist done.
  * nx is still using the legacy build commands.
  * `yarn run build:nx ` can be used


michaelbromley:
Major branch (v2) is now on Lerna v6 (based on Nx). Right now I'm happy to stick with Lerna since it is maintained by the Nx team anyway and shares the core engine. Thanks for your contribution nonetheless 👍

PR #1860, opened 2022-11-07
feat(admin): sort orders on customer details page

### LINKED ISSUE #1827: In the admin, the customer's orders should be reverse sorted by date
We have a lot of orders per customer and it's always a pain when an admin tries to locate the latest order in the admin screen. 

A simple default sort by orderPlacedAt would save us a lot of time.

teresaWienkBorgert:
Sort orders on customer details page by orderPlacedAt: 

# Related Issue

- Resolve #1827

### DIFF
--- a/packages/admin-ui/src/lib/customer/src/components/customer-detail/customer-detail.component.ts
+++ b/packages/admin-ui/src/lib/customer/src/components/customer-detail/customer-detail.component.ts
@@ -512,6 +512,7 @@ export class CustomerDetailComponent
             .getCustomer(this.id, {
                 take: this.ordersPerPage,
                 skip: (this.currentOrdersPage - 1) * this.ordersPerPage,
+                sort: { orderPlacedAt: SortOrder.DESC },
             })
             .single$.pipe(
                 map(data => data.customer),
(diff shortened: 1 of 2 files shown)


michaelbromley:
Thanks for the contribution! One question: you modified the `CustomerDetailComponent.fetchOrdersList()` method, which looks good as that is called when the list pagination is changed. But does this change also correctly sort the orders on initial load of the customer detail page?

From a quick look, it seems like that would be set in the [CustomerResolver](https://github.com/vendure-ecommerce/vendure/blob/15a7869a12613955a0b8dec54aa3ce17fbc86e59/packages/admin-ui/src/lib/customer/src/providers/routing/customer-resolver.ts#L27) class. Can you verify this to see if a change is needed there too?

teresaWienkBorgert:
> Thanks for the contribution! One question: you modified the `CustomerDetailComponent.fetchOrdersList()` method, which looks good as that is called when the list pagination is changed. But does this change also correctly sort the orders on initial load of the customer detail page?
> 
> From a quick look, it seems like that would be set in the [CustomerResolver](https://github.com/vendure-ecommerce/vendure/blob/15a7869a12613955a0b8dec54aa3ce17fbc86e59/packages/admin-ui/src/lib/customer/src/providers/routing/customer-resolver.ts#L27) class. Can you verify this to see if a change is needed there too?

Ohh yes you are right, needed to add the same sort order in the customer resolver as well 👍

michaelbromley:
Thanks Teresa 👍

PR #1862, opened 2022-11-07
feat(admin-ui): Login UI refresh

davidhoeck:
This PR implements a nicer looking user interface of the login screen. There is now an option that will be loaded from Unsplash by default. Furthermore the image is also configurable via the Admin UI config, so users can set their custom images.

### DIFF
--- a/packages/admin-ui/src/lib/login/src/components/login/login.component.scss
+++ b/packages/admin-ui/src/lib/login/src/components/login/login.component.scss
@@ -18,11 +18,21 @@
         align-items: stretch;
         position: relative;
 
+        @media (max-width: $breakpoint-medium) {
+            flex-direction: column;
+            height: auto;
+            width: 100%
+        }
+
         .login-wrapper-image {
             height: 100%;
             flex-grow: 1;
             position: relative;
 
+            @media (max-width: $breakpoint-medium) {
+                height: 300px;
+            }
+    
             img {
                 display: block;
                 width: 100%;
@@ -53,6 +63,10 @@
                     font-weight: bold;
                     color: white;
                     margin-bottom: 20px;
+
+                    @media (max-width: $breakpoint-medium) {
+                        font-size: 1.2rem;
+                    }
                 }
 
                 .login-wrapper-image-copyright {
@@ -76,6 +90,13 @@
             box-shadow: 0px 20px 25px rgba(0,0,0,0.1);
             overflow: hidden;
             border-radius: 5px;
+            flex-shrink: 0;
+
+            @media (max-width: $breakpoint-medium) {
+                height: auto;
+                width: 100%;
+                padding: 20px;
+            }
 
             .login-title {
                 font-weight: bold;
--- a/packages/admin-ui/src/lib/login/src/components/login/login.component.scss
+++ b/packages/admin-ui/src/lib/login/src/components/login/login.component.scss
@@ -137,19 +137,6 @@
     }
 }
 
-@media screen and (max-width: $breakpoint-small) {
-    .login-wrapper {
-        justify-content: center;
-        background-color: transparent;
-
-        .login {
-            margin: 20px;
-            padding: 24px;
-        }
-    }
-}
-
-
 .version {
     flex: 1;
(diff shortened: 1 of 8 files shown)


packages/admin-ui/src/lib/login/src/components/login/login.component.ts
```diff
@@ -35,6 +46,26 @@ export class LoginComponent {
         });
     }
 
+    loadImage() {
+        this.httpClient
+            .get('https://login-image.vendure.io', {
+                params: new HttpParams().append('domain', window.location.hostname),
```
michaelbromley: Do we need to explicitly append the domain here? Is that not something that we'll get for free from the express request object? (If you are using Express for the image server).

packages/admin-ui/src/lib/login/src/components/login/login.component.scss
```diff
@@ -1,60 +1,151 @@
...
+                display: flex;
+                flex-direction: column;
+                align-items: flex-start;
+                justify-content: flex-end;
+                padding: 30px;
+
+                .login-wrapper-image-title {
+                    font-size: 1.6rem;
+                    font-weight: bold;
+                    color: white;
```
michaelbromley, line 66: How about a more muted colour here such as `#afafaf` to make the greeting stand out better?

packages/admin-ui/src/lib/login/src/components/login/login.component.scss
```diff
@@ -1,60 +1,151 @@
...
+
+                    @media (max-width: $breakpoint-medium) {
+                        font-size: 1.2rem;
+                    }
+                }
+
+                .login-wrapper-image-copyright {
+                    p {
+                        font-size: 0.6rem;
+                        color: white;
```
michaelbromley, line 78: Same here with muted text colour

packages/admin-ui/src/lib/login/src/components/login/login.component.scss
```diff
@@ -1,60 +1,151 @@
...
-            padding: 24px;
+    .login-wrapper-inner {
+        background: #fff;
+        width: 1120px;
+        height: 590px;
+        display: flex;
+        justify-content: flex-start;
+        align-items: stretch;
+        position: relative;
+
```
michaelbromley, line 22: I experimented with adding `border-radius: 3px; overflow: hidden;` to this class to make the edges a bit softer. I personally like it, what do you think?

packages/common/src/shared-types.ts
```diff
@@ -303,6 +303,11 @@ export interface AdminUiConfig {
      * @default false
      */
     hideVersion?: boolean;
+    /**
+     * @description
+     * The custom login image
+     */
```
michaelbromley, line 311: There should be `@since 1.9.0` here to show in the docs in what version this option was added.

michaelbromley (commented):
Looks waaaay better, good work!

I left a few comments. I think we should also create an RFC around telemetry too before we launch this feature, I want to be explicit about what we data we plan to collect.

packages/admin-ui/src/lib/login/src/components/login/login.component.scss
dlhck: Yeah, why not ✅

packages/admin-ui/src/lib/login/src/components/login/login.component.ts
dlhck: Well, the question is - is it always the same, when you are using the admin ui behind a CDN or load balancer?

michaelbromley: I think so, because this is a client-side app, the request will always come from the browser rather than the server hosting the admin ui app. So it _should_ be exactly the same as what you would get from `window.locations.hostname`, if my understanding is correct.

dlhck: Yep, just tried it. You're correct. If it is not the same, it would show two or three different images maximum, so also not really a problem tbh.

Issue #1871, opened 2022-11-10
init method not called for TaxLineCalculationStrategy 


casabian:
Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@vendure/core/dist/config/config.module.js b/node_modules/@vendure/core/dist/config/config.module.js
index 9cf10b5..cf6cd68 100644
--- a/node_modules/@vendure/core/dist/config/config.module.js
+++ b/node_modules/@vendure/core/dist/config/config.module.js
@@ -67,7 +67,7 @@ let ConfigModule = class ConfigModule {
         const { assetNamingStrategy, assetPreviewStrategy, assetStorageStrategy } = this.configService.assetOptions;
         const { productVariantPriceCalculationStrategy, stockDisplayStrategy } = this.configService.catalogOptions;
         const { adminAuthenticationStrategy, shopAuthenticationStrategy, sessionCacheStrategy, passwordHashingStrategy, passwordValidationStrategy, } = this.configService.authOptions;
-        const { taxZoneStrategy } = this.configService.taxOptions;
+        const { taxZoneStrategy, taxLineCalculationStrategy } = this.configService.taxOptions;
         const { jobQueueStrategy, jobBufferStorageStrategy } = this.configService.jobQueueOptions;
         const { mergeStrategy, checkoutMergeStrategy, orderItemPriceCalculationStrategy, process, orderCodeStrategy, orderByCodeAccessStrategy, stockAllocationStrategy, } = this.configService.orderOptions;
         const { customFulfillmentProcess } = this.configService.shippingOptions;
@@ -85,6 +85,7 @@ let ConfigModule = class ConfigModule {
             assetPreviewStrategy,
             assetStorageStrategy,
             taxZoneStrategy,
+            taxLineCalculationStrategy, 
             jobQueueStrategy,
             jobBufferStorageStrategy,
             mergeStrategy,
```

PR #1872, opened 2022-11-10
fix: issue #1871 - return taxLineCalculationStrategy

casabian:
TaxLineCalculationStrategy was not returned from ConfigModule#getInjectableStrategies and thus the init method of implemented strategies was not called.


michaelbromley:
Hi @casabian, it looks like this PR got a bit mixed up with other changes on the branch. Could you either clean it up with a force push, or create a new PR (whichever is easiest for you)? Thanks.

casabian:
I'll make a new PR

PR #1877, opened 2022-11-12
Add email plugin transport options for AWS SES

chregu:
Has the advantage of having a sending rate to prevent hitting limits at AWS SES and is officially supported by nodemailer
https://nodemailer.com/transports/ses/

Would be great, if this could be included. Would save us eventual future headaches with send limits of AWS.

### DIFF
--- a/packages/email-plugin/src/types.ts
+++ b/packages/email-plugin/src/types.ts
@@ -7,6 +7,7 @@ import SMTPTransport from 'nodemailer/lib/smtp-transport';
 import { EmailGenerator } from './email-generator';
 import { EmailSender } from './email-sender';
 import { EmailEventHandler } from './event-handler';
+import SESTransport from 'nodemailer/lib/ses-transport'
 
 /**
  * @description
@@ -112,6 +113,7 @@ export type EmailTransportOptions =
     | SendmailTransportOptions
     | FileTransportOptions
     | NoopTransportOptions
+    | SESTransportOptions
     | TestingTransportOptions;
 
 /**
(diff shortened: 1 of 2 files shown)


packages/email-plugin/src/types.ts
```diff
@@ -133,6 +135,55 @@ export interface SMTPTransportOptions extends SMTPTransport.Options {
...
+ *     }),
+ *   ],
+ * };
+ *  ```
+ * @docsCategory EmailPlugin
+ * @docsPage Transport Options
+ */
+export interface SESTransportOptions extends SESTransport.Options {
+    type: 'ses';
+    /**
```
michaelbromley: Is this comment supposed to be here?

packages/email-plugin/src/types.ts
```diff
@@ -133,6 +135,55 @@ export interface SMTPTransportOptions extends SMTPTransport.Options {
...
 
+/**
+ * @description
+ * The SES transport options of [Nodemailer](https://nodemailer.com/transports/ses//)
+ *
+ * See [Nodemailers's SES docs](https://nodemailer.com/transports/ses/) for more details
+ * 
+ * @example
+ * ```TypeScript
+ *  import { SES, SendRawEmailCommand } from '@aws-sdk/client-ses'
```
michaelbromley: due to the way the TS parser works when parsing JSDoc comments, you need to escape the `@` symbol in code blocks otherwise it thinks it is a tag like `@example`. So should read `\@aws-sdk/client-ses` (the escape symbol is removed when generating the docs).

michaelbromley (changes requested):
Thanks, this is great!
Just a couple of points to review. Also, could you make the PR against `minor` since it is adding a new feature?

chregu:
Rebased against the minor branch

chregu:
No idea why it still says "1 review requesting changes". Anyway, all is done. Just saying that it's not blocked by that (but it's also not urgent, in the next minor version would of course be great)
image

michaelbromley:
We have a couple of flakey e2e tests at the moment to do with database non-determinism which I need to sort out. Nothing wrong with your changes. Thanks!

PR #1878, opened 2022-11-13
fix(payments-plugin): Return 200 on Stripe payment failed event

vrosa:
The Stripe webhook handler does not return a 200 status code response for `payment_intent.payment_failed` events as per [official docs](https://stripe.com/docs/webhooks#acknowledge-events-immediately). This PR fixes that behaviour, which was initially [flagged on Slack](https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1668154455407779).


michaelbromley:
Thanks!

PR #1880, opened 2022-11-14
feat(core)!: Improve channel service findAll

alexisvigoureux:
Our need is to retrieve a list of channels by id

```
const { items } = await this.findAll(ctx, {
   filter: {
     id: { in: input.channelIds },
   },
 })
```

This PR standardize the findAll function of ChannelService, by accepting a list query options and relations, and returning a paginated list.


michaelbromley:
I agree with this in principle but this will be a breaking change to anyone currently using the `findAll` method. So it would need to go on the `major` branch.

michaelbromley:
Thanks Alexis 👍

PR #1387, opened 2022-01-28
Fix: add Subscription to parentType

danieldflrss:
Add Subscription to parent type resolver action
fix problem when trying to authenticate a graphql subscription

PR #1881, opened 2022-11-14
Fix: add Subscription to parentType

chregu:
Add Subscription to parent type resolver action
fix problem when trying to authenticate a graphql subscription

replaces https://github.com/vendure-ecommerce/vendure/pull/1387


michaelbromley:
Thanks :)

Issue #1824, opened 2022-10-05
Feat: Use Mollie's 'Order API'


martijnvdbrug:
**Is your feature request related to a problem? Please describe.**
Mollie recommends using the Orders API for most cases (more details below). The new Orders API also comes with some extra features: https://docs.mollie.com/orders/why-use-orders 

The most important improvement for Vendure is that we can use `Pay later` payment methods.

Mollie gives these reasons for **NOT** using the orders API:
* When you do not collect customer details like the address, it is not possible to use the Orders API. Then you can choose for the Payments API to create a simple payment instead.
* When you only want to receive donations and you do not sell any digital or physical products.
* You are already using the Payments API and do not want to use all the nice extra features of the Orders API.

I don't think any of these reasons apply to Vendure, so, we migrate to the Orders API :rocket: 

**Describe the solution you'd like**
* **Pay after delivery payment methods such as Klarna Pay later, Klarna Slice it or in3, Klarna Pay now, as well as eco vouchers, gift vouchers, and meal vouchers require the Orders API** 
* Some payment methods communicate the order lines to the customer in their hosted payment pages or in an invoice. For example, the customer will see the order lines on the checkout page of PayPal.
You can manage the created orders in the [Mollie Dashboard](https://www.mollie.com/en/features/dashboard).

**Describe alternatives you've considered**
Keep using the payments API, but this means we can't use any of the pay later methods ever.

**Additional context**
With the new Order API we can have 2 payment flows, 1 for traditional payment methods and 1 for pay-later methods:
* In both cases a payment intent is created, and the order stays in the `AddingItems` state
* With the traditional flow, the payment is immediately settled via webhook (this is currently the only flow)
* The pay-later flow is a little different: The first webhook will have status `Authorized`, and later, when the order in Mollie is transitioned to `Shipped`, another webhook will follow with status `Paid`.

So the Mollie plugin should be able to handle these flows:

1. Create payment intent > payment settled
2. Create payment intent > payment authorized > payment settled

Quote from the Mollie docs:

>  Pay later payment methods will have an **authorized** status. Shipping is required, and it ensures you will be settled. Mollie expects Vendure to update the status of an order in Mollie to 'Shipped'. When all items are shipped, the order is completed.

![image](https://user-images.githubusercontent.com/6604455/201627261-47756926-0f7a-42ff-9a5f-988251d89cc6.png)

PR #1884, opened 2022-11-15
feat(payments-plugin): Use Mollie's Order API

martijnvdbrug:
See #1824 for more details.

This PR implements the Mollie orders API, so we can use pay later methods.

mollie

Important changes:
* Orders MUST have a shipping or billing address with at least city, postalCode, street and country. Payment intent creation will fail if this is not the case
* You can now use pay-later methods: Orders with a pay-later method will be in PaymentAuthorized and need to be settled/captured via the admin UI or programatically. Mollie wants you to do this when you actually ship the order. The payment MUST be captured within 28 days, otherwise the order in Mollie will expire, and your payment can never be captured/settled. https://docs.mollie.com/orders/status-changes#expired

**AutoCapture**
If you don't want to worry about capture/settlement, you can configure auto-capture per payment method. Auto-capture immediately captures/settles the payment after being authorized. This can be useful for digital products or services, where there is no physical shipping moment.

Inspired by the way Shopify handles pay-later methods with auto capturing: https://docs.klarna.com/platform-solutions/shopify/payments/klarna-payments/order-management/

New e2e tests added:
image


packages/payments-plugin/README.md
```diff
@@ -1,3 +1,19 @@
 # Payments plugin
 
 For documentation, see https://www.vendure.io/docs/typescript-api/payments-plugin
+
+## Development
+
+### Mollie local development
```
martijnvdbrug, line 7: Move dev docs to readme instead of public plugin docs, because its only for contributors.

packages/payments-plugin/e2e/mollie-dev-server.ts
```diff
@@ -0,0 +1,90 @@
...
+import { molliePaymentHandler } from '../package/mollie/mollie.handler';
+
+import { CREATE_PAYMENT_METHOD } from './graphql/admin-queries';
+import { CreatePaymentMethod } from './graphql/generated-admin-types';
+import { AddItemToOrder } from './graphql/generated-shop-types';
+import { ADD_ITEM_TO_ORDER } from './graphql/shop-queries';
+import { CREATE_MOLLIE_PAYMENT_INTENT, setShipping } from './payment-helpers';
+
+/**
+ * This should only be used to locally test the Mollie payment plugin
```
martijnvdbrug, line 28: Do we want this file in GIT? This is meant for contributors (as mentioned in the readme). It's quite the hassle to setup a local env that actually communicates with the live Mollie env, so this file is very useful for any future contributors (and me for any future changes)

This dev-server only requires a Mollie apiKey and creates ready to test orders.

packages/payments-plugin/src/mollie/mollie.helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+            givenName: customer.firstName,
+            familyName: customer.lastName,
+            email: customer.emailAddress,
+        };
+    }
+}
+
+/**
+ * Map all order and shipping lines to a single array of Mollie order lines
+ */
```
martijnvdbrug, line 28: @michaelbromley Will he function below always result in the order total? Mollie will validate if all lines add up to the order total, including taxes and discounts. If not, Mollie says no, and we won't be able to create a payment intent.

I don't like recreating this logic, but I've seen this a few times now: accounting software and external fulfillment platforms that want all lines with prices, taxes, shipping and discounts AND want it to add up to the order total.

martijnvdbrug:
@michaelbromley Build fails with `Error: ../core/src/i18n/i18n.service.ts(7,21): error TS7016: Could not find a declaration file for module 'i18next-fs-backend'. '/home/runner/work/vendure/vendure/node_modules/i18next-fs-backend/cjs/index.js`
But I'm pretty sure this is not related to this PR?

packages/payments-plugin/README.md
michaelbromley: Yes this makes sense.

packages/payments-plugin/e2e/mollie-dev-server.ts
michaelbromley: Yes for sure, keep this in the repo!

packages/payments-plugin/package.json
```diff
@@ -1,40 +1,41 @@
...
-        "@vendure/common": "^1.8.4",
-        "@vendure/core": "^1.8.4",
-        "@vendure/testing": "^1.8.4",
-        "braintree": "^3.0.0",
-        "nock": "^13.1.4",
-        "rimraf": "^3.0.2",
-        "stripe": "^8.197.0",
-        "typescript": "4.3.5"
-    }
+  "name": "@vendure/payments-plugin",
```
michaelbromley: Looks like some unnecessary formatting of indentation happened here.

packages/payments-plugin/src/mollie/mollie.helpers.ts
michaelbromley: The logic looks correct for the order lines themselves. The order total is further composed of the shipping and any surcharges. You can see the entire internal calculation for totals here: https://github.com/vendure-ecommerce/vendure/blob/81426fa01ad1ff82a587444c3768944e99c5e6b6/packages/core/src/service/helpers/order-calculator/order-calculator.ts#L453-L478

packages/payments-plugin/src/mollie/mollie.helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+export function toAmount(value: number, orderCurrency: string): Amount {
+    return {
+        value: (value / 100).toFixed(2),
+        currency: orderCurrency,
+    };
+}
+
+/**
+ * Lookup one of Mollies allowed locales based on an orders countrycode or channel default.
+ * If both lookups fail, resolve to en_US to prevent payment failure
```
michaelbromley, line 72: What does `locale` signify in Mollie? Is it only for localization? If so this is fine.

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -68,16 +75,16 @@ export class MollieService {
...
         const [order, paymentMethod] = await Promise.all([
-            this.activeOrderService.getOrderFromContext(ctx),
+            this.activeOrderService.getActiveOrder(ctx, undefined),
             this.getPaymentMethod(ctx, paymentMethodCode),
         ]);
         if (!order) {
             return new PaymentIntentError('No active order found for session');
         }
-        await this.entityHydrator.hydrate(ctx, order, { relations: ['lines', 'customer', 'shippingLines'] });
+        await this.entityHydrator.hydrate(ctx, order, { relations: ['customer', 'lines', 'lines.productVariant', 'shippingLines', 'shippingLines.shippingMethod'] });
```
michaelbromley: With EntityHydrator, you don't need to list the intermediate joins like you do with TypeORM APIs. I.e you can put `lines.productVariant` and it will automatically know to join `lines` first.

michaelbromley (commented):
Martijn, excellent PR! I just left a few comments, nothing major to change as far as I can see.

packages/payments-plugin/src/mollie/mollie.helpers.ts
martijnvdbrug: I have included shipping in the calculation, but surcharges...

Are surcharges included in the `line.proratedLinePrice` price? 

Mollie only knows orderLines, it doesn't know the concept of surcharges. The docs state this:

> The total amount of the line, including VAT and discounts. Adding all totalAmount values together should result in the same amount as the amount top level property.

So, the sum of `lines.totalAmount` must equal the `order.totalAmount`. 

Is the solution here to add surcharges as separate lines?

michaelbromley: > Are surcharges included in the line.proratedLinePrice price?

No, they are separate.

In that case then yes, map surcharges to lines and Mollie should be happy.

packages/payments-plugin/src/mollie/mollie.helpers.ts
martijnvdbrug: Yep, only for showing the customers language during the hosted checkout. It is a mandatory field for some reason.

> Allows you to preset the language to be used in the hosted payment pages shown to the consumer.

martijnvdbrug:
> > But I'm pretty sure this is not related to this PR. Should I try and fix it in this PR?
> 
> No ignore it for now. If you wanna have a go at fixing, do it in a separate PR.

It does also block the e2e tests right now, is that an issue?

michaelbromley:
> It does also block the e2e tests right now, is that an issue?

Ah ok yeah we should fix that then. I'll take a look and see if I can figure out what's causing it...

packages/payments-plugin/src/mollie/mollie.handler.ts
```diff
@@ -1,17 +1,17 @@
...
     CreatePaymentErrorResult,
     CreatePaymentResult,
     CreateRefundResult,
     Logger,
     PaymentMethodHandler,
-    PaymentMethodService,
     SettlePaymentResult,
 } from '@vendure/core';
-import { Permission } from '@vendure/core';
+import { PaymentState } from '@vendure/core/src';
```
michaelbromley: This is the cause of the build error. Should be imported from @vendure/core.

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -1,9 +1,15 @@
...
+import { CreateParameters } from '@mollie/api-client/dist/types/src/binders/orders/parameters';
 import { Inject, Injectable } from '@nestjs/common';
 import {
     ActiveOrderService,
     ChannelService,
     EntityHydrator,
+    ErrorResult,
     LanguageCode,
     Logger,
     Order,
```
michaelbromley, line 15: Add the `OrderStateTransitionError` import here rather than the current deep import.

michaelbromley (changes requested):
Found the build issue

packages/payments-plugin/src/mollie/mollie.handler.ts
martijnvdbrug: Oops, so it IS caused by this PR :1st_place_medal:

packages/payments-plugin/src/mollie/mollie.helpers.ts
martijnvdbrug: @michaelbromley Not really familiar with surcharges. How can I add a promotion that will add a surcharge to the order? I would like to test if Mollie allows negative numbers on an order line + include a surcharge in the e2e test.

michaelbromley: Surcharges are not related to Promotions. Promotions can only make "adjustments", which is what we already take into account in the `discountedPrice` and `proratedPrice`. However, promotions _can_ make an orderline negative - this is an edge case and almost always an error in the PromotionAction logic. But should be gracefully handled if it comes up.

Surcharges must be manually added using `OrderService.addSurchargeToOrder`. There is currently no built-in mutation for this.

martijnvdbrug: Thanks, I will try to fix this today.

martijnvdbrug: image
Mollie says 'OK'. I will add it to the e2e test and round up this PR

michaelbromley:
Martijn, thanks so much for the huge amount of work you put into this! 🙏 🥳

martijnvdbrug:
Happy to contribute! (And a healthy dose of self-interest, of course)

Issue #1889, opened 2022-11-19
Draft order should filter the product variant list


carathorys:
**Describe the bug**
When creating a new order in the admin-ui it can happen that you select a disabled product variant, which will result in an error message on the admin-ui.

**To Reproduce**
1. Go to 'Products'
2. Click on 'Edit' on a product
3. Go to the Product Variants tab
4. Disable some of them
5. Go to 'Orders'
6. Create a new draft order
7. Search for the disabled product variant
8. Add it to your order by clicking to the `Add item to order` button
9. See the error: ` No ProductVariant with the id 'XYZ' could be found`

**Expected behavior**
If a product can't be added to an order _(eg.: the product variant is disabled)_ it shoudn't be listed in the product selector.

**Environment (please complete the following information):**
 - @vendure/core version: 1.8.x
 - Nodejs version: not relevant
 - Database (mysql/postgres etc): not relevant

**Additional context**

PR #1890, opened 2022-11-20
fix(admin-ui): Draft order product selector should show only enabled products

carathorys:
This PR tries to address the issue #1889 what I've reported: 
1. Enabled the `search` query on the `admin-api` to be filtered to the enabled product variants
2. Adds the option to use this filter to the `product-selector` component
3. Adds the option to the product data service to be used

### DIFF
--- a/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
+++ b/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
@@ -381,8 +381,7 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
                 body: JSON.stringify({ id: mockData.mollieOrderResponse.id }),
                 headers: { 'Content-Type': 'application/json' },
             });
-            // tslint:disable-next-line:no-non-null-assertion
-            const { order: adminOrder } = await adminClient.query(GET_ORDER_PAYMENTS, { id: order!.id });
+            const { order: adminOrder } = await adminClient.query(GET_ORDER_PAYMENTS, { id: order?.id });
             expect(adminOrder.state).toBe('ArrangingPayment');
         });
 
@@ -457,7 +456,7 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
                 adminClient,
                 order.lines[0].id,
                 10,
-                // tslint:disable-next-line:no-non-null-assertion
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 order.payments!.find(p => p.amount === 108990)!.id,
                 SURCHARGE_AMOUNT,
             );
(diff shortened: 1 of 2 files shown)


michaelbromley:
Thanks for the report & PR!

I have 1 hesitation here about this solution: if we extend the search query input, then we'll also have to extend it for the Elastic plugin too (plus any other search-compatible plugins such as the Typesnse-based advanced search plugin which is planned).

I'd like to suggest that we instead consider still listing disabled products, but we use the `SearchResult.enabled` field in the search query to get the status of the result. We can then mark it as disabled in the list and make it so it cannot be added.

I think it might even be more useful for an admin to see that the product exists, but is disabled, rather than not seeing it at all. 

What do you think?

carathorys:
Hi! 
I was thinking on this as well, but I ended up here because: 
This is defined in the Full text search resolver, which will invoke the search index whatever it should. 
I thought that it's implemented in the Elastic plugin as well as that will implement the same interface _(now I see that it's not using the same resolver)_.
Filtering on the client side wouldn't be an optimal solution because of the paging I think _(it can be implemented, but why pass everything from the DB to the Server, then to the client?)_

If it is OK, I'll check the Elastic plugin as well _(I've never worked with Elastic before, so it can take some time)_, but I would like to provide a proper solution for this :) 
Also, if you think this shouldn't be done there, I can live with that, this was just my opinion :)

carathorys:
Yes, I can imagine this as a solution:
If a product is disabled we indicate it in the dropdown, and we doesn't allow the user/admin to select that variant. Or it should be still selectable, and when selected we show that the variant is disabled?

michaelbromley:
We should definitely show it is disabled in the dropdown - don't want to require an extra click to discover this data. As to whether it is selectable... I'd say it's safer to make it selectable (other uses of the same component might want to be able to select disabled items), and then we'd disable the "add to order" button.

carathorys:
Hi Michael

I've checked the Elastic plugin, it's only one line change to enable the filtering there:
`elasticsearch-resolver.ts`: 
```
await this.elasticsearchService.search(ctx, args.input, args.input.enabledOnly);
```

Also earlier I've added an optional parameter to the product selector component (`vdr-product-selector`) where you can still define whether you would like to get all the products, or just the enabled ones, with the default value of the current behavior to show everything, and then just made the `vdr-draft-order-variant-selector` to filter the results. 

_If you wish, I can try to make those additional changes of course, it was just more straightforward to me implement this way (why should I show an option to you, if then I doesn't allow you to use it?)._

michaelbromley:
Hi @carathorys, apologies for the delay - I got busy with recent releases etc.

To return to this issue: what I'd like to do for the time being is to make this an admin-ui only change, and leave the core queries as they are.

Would you be willing to modify your PR / make a new one to implement this?

Issue #1893, opened 2022-11-22
Database error when dealing with big collection.


ttournie:
**Describe the bug**
When trying to create/update a collection with a lots of variants in it we get an error msg from the datase Driver because au the 
Typeorm save call in applyCollectionFiltersInternal.

This is because the chunk option does not work for linked entities.

**To Reproduce**
Create a collection with a bit amount of variant (we have a collection with 34 000 items)

**Expected behavior**
This should not break the job.

**Environment (please complete the following information):**
 - @vendure/core version: 1.8.3
 - Nodejs version 16.16
 - Database (mysql/postgres etc): postgres

**Additional context**
I made a fix and optimised a bit the apply-collection-filter job by also making this job pop another job for each collection that need updating.

I will make a PR

ttournie:
Perfect

PR #1896, opened 2022-11-22
Fix/issue/1893

ttournie:
Fix scaling issue when there is a lot of collections/variant in a collection

- Fix applyCollectionFiltersInternal for big collection
- Make apply-collection-filter pop unitary job to avoid having a Big slow job but tiny quick one

### DIFF
--- a/packages/core/e2e/default-search-plugin.e2e-spec.ts
+++ b/packages/core/e2e/default-search-plugin.e2e-spec.ts
@@ -936,6 +936,8 @@ describe('Default search plugin', () => {
                 await awaitRunningJobs(adminClient);
                 // add an additional check for the collection filters to update
                 await awaitRunningJobs(adminClient);
+                // add an additional check for the collection filters unitary to update
+                await awaitRunningJobs(adminClient);
                 const result1 = await doAdminSearchQuery({ collectionId: 'T_2', groupByProduct: true });
 
                 expect(result1.search.items.map(i => i.productName)).toEqual([
--- a/packages/core/src/service/services/collection.service.ts
+++ b/packages/core/src/service/services/collection.service.ts
@@ -559,6 +559,15 @@ export class CollectionService implements OnModuleInit {
         return filters;
     }
 
+    private chunkArray = <T>(array: T[], chunkSize: number): T[][] => {
+        const results = []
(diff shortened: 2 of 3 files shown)


michaelbromley:
Hi Thierry, thanks for the PR!

I'm wanting to get a deeper understanding of this issue, so I have a few questions:

> we get an error msg from the datase Driver because au the
Typeorm save call in applyCollectionFiltersInternal.

What is the error exactly that you get?

> Make apply-collection-filter pop unitary job to avoid having a Big slow job but tiny quick one

Is there an advantage to doing it this way (adding a job from within a job) rather than just directly adding a single collection id to the existing `applyFiltersQueue`?

Apart from not causing the DB error, do you notice any other performance changes using this code, e.g. speed or memory usage?

ttournie:
> Hi Thierry, thanks for the PR!
> 
> I'm wanting to get a deeper understanding of this issue, so I have a few questions:
> 
> > we get an error msg from the datase Driver because au the
> > Typeorm save call in applyCollectionFiltersInternal.
> 
> What is the error exactly that you get?
> 
> > Make apply-collection-filter pop unitary job to avoid having a Big slow job but tiny quick one
> 
> Is there an advantage to doing it this way (adding a job from within a job) rather than just directly adding a single collection id to the existing `applyFiltersQueue`?
> 
> Apart from not causing the DB error, do you notice any other performance changes using this code, e.g. speed or memory usage?

Hello,

The error msg was : bind message has 24880 parameter formats but 0 parameters

For the job popping other jobs, that is a good question, It might be possible to directly have several unitary job popping yes. Not sur if it's a problem for the Es job buffer thought.
The popping of unitary job might not be necessary but it avoid waiting for huge collection to build before doing the next one and might be an issue if you have a lot's of collection/sub-collection like we do.

For the performance, For us it's quicker now for big collection because of how the query relation builder works.

michaelbromley:
Thanks for the extra info. I pulled this PR locally and played with it a bit. I found that the main improvements you made to the TypeORM calls + the chunking alone give a big benefit in perf. I did not like the splitting of jobs, because it generates too much noise - hundreds of jobs can be created each time a product is changed.

I added another optimization on top and will include some benchmark data in issue #1893

Issue #1543, opened 2022-05-02
Migrate monorepo from Lerna to Nx


michaelbromley:
This monorepo is built on Lerna, a tool for managing monorepos which handles linking sibling packages and also makes the versioning & publishing of multiple packages much easier. Lerna has served us very well for the past ~4 years, but unfortunately has been unmaintained for a while and recently officially announced the fact. 
![image](https://user-images.githubusercontent.com/6275952/166199839-215ab742-ba4f-438b-87e8-8ded5382c320.png)

Therefore we should move to another tool to handle these concerns. Recently I've been using [Nx](nx.dev) in some other projects and it is very, very good. It is not primarily aimed at the exact same use-case that we've been using Lerna for - specifically it does not handle all the publishing stuff automatically. But the monorepo management aspect is much more sophisticated. They even have a [guide for migrating from Lerna to Nx](https://nx.dev/migration/adding-to-monorepo). 

This repo has, over the years, accumulated quite a lot of custom scripts and checks to ensure a smooth release cycle. This will all take some verification and no doubt quite a bit of manual modification to work in Nx.

So the plan is to start with that migration guide, and then go though each of the custom scripts (doc gen, code gen, release checks, version bumps etc) and get each working.

This will be done on the `master` branch and then merge up into `minor` and `major`.

michaelbromley:
In the mean time, Nrwl took over maintenance of Lerna and started releasing new versions!

https://twitter.com/victorsavkin/status/1529493314621145090

So for the short-term I'll probably stick with Lerna and keep Nx migration open as a future project when there's a bit more time for non-critical tasks.

cwerner1:
Hi @michaelbromley 

I added a basic configuration for the project. So the nx features like the dependency graph is available.

I`ve also tested the new lerna version 5 where the nx integrationcan be used and takes advantage of the nx dependency management.

https://github.com/vendure-ecommerce/vendure/pull/1850

michaelbromley:
Hey @cwerner1, thanks! What are your thoughts on moving to Lerna v6 vs Nx? 

The thing I get the most value from in Lerna is the way it automatically manages dependency version bumps on release. I'm not sure whether that's something that is well supported in Nx or whether we'd need a custom plugin for that.

I was under the impression that migrating from Lerna v4 to the new Nx-powered v6 would be a good first step.

asonnleitner:
What about rushjs? @michaelbromley

asonnleitner:
[...]

PR #1902, opened 2022-11-24
chore: Update to Lerna v6

michaelbromley:
Relates to #1543
Running into this issue: https://github.com/nrwl/nx/issues/10822#issuecomment-1326795930 which prevents me from moving over to this on master right now.


michaelbromley:
This is done now in the major branch.

PR #1904, opened 2022-11-25
Added additional sort order to orderService.findOne query

### LINKED ISSUE #1903: Add additional sort order to `orderSerivce.findOne`
We have a use case where we add multiple items at the same time to the order, thus all orderItems and orderLines have the same createdAt time. While retrieving such kind of orders, the included orderLines switch the order.

It would be nice to additionally sort the included orderLines of an order by the productVariantId to have a unique order while fetching an order with orderLines created at the exact same time.

teresaWienkBorgert:
Added additional sort order to orderService.findOne query to avoid changing sort order of orderLines if they are created at the exact same time. 

resolves #1903


teresaWienkBorgert:
I could not find ant tests for the sort order of an order, in case I need to extend the tests just let me know :)

michaelbromley:
No, there are not tests for this, but none of the other tests break so it's all good. Thanks for the fix!

PR #1908, opened 2022-11-28
fix(core): Importing custom boolean field works as intended

### LINKED ISSUE #1907: Boolean custom fields don't get populated from csv files.
**Describe the bug**
When using a csv file to import products, having a boolean customField never gets the right value.

**To Reproduce**
1. Create a boolean customField on products or productVariants
2. Import products using said customField
3. DB always gets the default value for it, never respecting the csv data.

**Expected behavior**
It should reflect what the csv file data.

**Environment:**
 - @vendure/core version: 1.8.4
 - Nodejs version: 16.18.1
 - Database (mysql/postgres etc): postgresql

knoid:
fixes #1907 by explicitly parsing boolean fields like `value.toLowerCase() === 'true'`.


michaelbromley:
Thank you for your contribution! 🙏

PR #1909, opened 2022-11-30
feat(elasticsearch-plugin): Independently access to customMappings

Izayda:
### Description
Access to customProductVariantMappings and customProductMappings independently from groupByProduct by adding two new fields - customProductVariantMappings and customProductMappings.

This PR don't break current GQL scheme, just expand it.

### Use case
We always use `groupByProduct=true`, because our product can have a lot of variants and we don't want to spam our listings of rather similar items. 
Current logic of elasticsearch plugin select the best productVariant attributes after applying filters (through facets values). But this logic is not applied to customMapping - i can't select customProductVariantMappings when i have `groupByProduct=true`. Real case - we have customProductVariantMapping `allAssets` with all image of variant. I want to apply filter `color=blue` and get grouped by product result with blue images.

### DIFF
--- a/packages/elasticsearch-plugin/src/options.ts
+++ b/packages/elasticsearch-plugin/src/options.ts
@@ -165,7 +165,7 @@ export interface ElasticsearchOptions {
      * @description
      * Custom mappings may be defined which will add the defined data to the
      * Elasticsearch index and expose that data via the SearchResult GraphQL type,
-     * adding a new `customMappings` field, customProductMappings, customProductVariantMappings.
+     * adding a new `customMappings`, `customProductMappings` & `customProductVariantMappings` fields.
      *
      * The `graphQlType` property may be one of `String`, `Int`, `Float`, `Boolean`, `ID` or list
      * versions thereof (`[String!]` etc) and can be appended with a `!` to indicate non-nullable fields.
@@ -175,8 +175,8 @@ export interface ElasticsearchOptions {
      * parsed to the elasticsearch index.
      *
      * This config option defines custom mappings which are accessible when the "groupByProduct"
-     * input options is set to `true`. Anyway, customProductVariantMappings can be accessed by using
-     * customProductVariantMappings field.
+     * input options is set to `true`. In addition, custom variant mappings can be accessed by using
+     * the `customProductVariantMappings` field, which is always available.
      *
      * @example
      * ```TypeScript
@@ -227,8 +227,8 @@ export interface ElasticsearchOptions {
     /**
      * @description
      * This config option defines custom mappings which are accessible when the "groupByProduct"
-     * input options is set to `false`. Anyway, customProductMappings can be accessed by using
-     * customProductMappings field.
+     * input options is set to `false`. In addition, custom product mappings can be accessed by using
+     * the `customProductMappings` field, which is always available.
      *
      * @example
      * ```SDL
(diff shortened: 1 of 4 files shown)


packages/elasticsearch-plugin/src/options.ts
```diff
@@ -165,7 +165,7 @@ export interface ElasticsearchOptions {
      * @description
      * Custom mappings may be defined which will add the defined data to the
      * Elasticsearch index and expose that data via the SearchResult GraphQL type,
-     * adding a new `customMappings` field.
+     * adding a new `customMappings` field, customProductMappings, customProductVariantMappings.
```
michaelbromley: should read:

"adding new `customMappings`, `customProductMappings` & `customProductVariantMappings` fields"

packages/elasticsearch-plugin/src/options.ts
```diff
@@ -175,7 +175,8 @@ export interface ElasticsearchOptions {
      * parsed to the elasticsearch index.
      *
      * This config option defines custom mappings which are accessible when the "groupByProduct"
-     * input options is set to `true`.
+     * input options is set to `true`. Anyway, customProductVariantMappings can be accessed by using
+     * customProductVariantMappings field.
```
michaelbromley: >input options is set to `true`. Anyway, customProductVariantMappings can be accessed by using
customProductVariantMappings field.

change to 

"input options is set to `true`. In addition, custom variant mappings can be accessed by using
the `customProductVariantMappings` field, which is always available."

packages/elasticsearch-plugin/src/options.ts
```diff
@@ -222,7 +227,8 @@ export interface ElasticsearchOptions {
     /**
      * @description
      * This config option defines custom mappings which are accessible when the "groupByProduct"
-     * input options is set to `false`.
+     * input options is set to `false`. Anyway, customProductMappings can be accessed by using
+     * customProductMappings field.
```
michaelbromley: "input options is set to `false`. In addition, custom product mappings can be accessed by using
the `customProductMappings` field, which is always available."

michaelbromley (changes requested):
Hi Artem, thanks for the PR!

I was wondering whether it would have been possible to avoid creating new fields, and allowing the user to select _both_ product & variant custom mappings in the existing `customMappings` field. But after a bit of experimentation it appears that this is not possible. Unless we add a new type like `AllCustomMappings` which combines all the fields from the product & variant custom mappings. But even then I'm not sure that would work in a non-breaking way.

Perhaps in this case we should deprecate the `customMappings` field, and suggest using the specific new fields instead? I don't really like the redundancy, unless there is some good reason to keep both that I didn't think of?

Also I just left a couple of comments on improving the documentation added.

Izayda:
Hi @michaelbromley , 

Have fixed your comments. 

I tried to implement this feature in current approach, but it seems, that it will break schema anyway. So, it seems for me we can just use new approach and mark `customMappings` as depreciated (how?), but maybe there is some reason in current implementation.

michaelbromley:
It should be possible to mark as deprecated with a directive like'
```gql
extend type SearchResult {
  customMappings: CustomMappings! @deprecated(reason: "Use customProductMappings or customProductVariantMappings")
}
```

Izayda:
Ok, give me couple of minutes

michaelbromley:
Thanks!

PR #1928, opened 2022-12-08
docs: Separate adding and overriding nav items

elauser:
We couldn't find the documentation part which lets us add custom actions in the admin UI to existing pages.
Having this page split up into adding-navigation-items and overriding-navigation-items would make it better visible at a glance, and we would have probably found it when we scanned through the docs.


michaelbromley:
Should the "overriding nav items" section really be separated from the "adding nav items" page? I was more thinking along the lines of a split like:

- modifying the nav menu (adding + overriding nav items)
- adding navigation to pages (action bar items)

Thoughts?

elauser:
Hm.. yes the split is still a bit unclear.
Maybe have the following two sections

- Modifying navigation items
  - Extending the NavMenu
  - Override existing items
- Add actions to pages
  - Adding ActionBar buttons
  - Adding BulkActions

@michaelbromley what do you think of that?

elauser:
@michaelbromley implemented the suggested changes. Anything to add/remove?

michaelbromley:
Looks good, thanks!

PR #1930, opened 2022-12-09
feat(event-bus): Use instanceof check for of type operator

yackinn:


yackinn:
Let me know what we need tests for when we've agreed on an api for the event bus.

michaelbromley:
Hi,

Thanks for your contribution :) I've spent a bit more time thinking about this. I'd like to _only_ add the `filter()` method. The reasons are:

1. It is the most flexible. 
2. I want to add as few new APIs as possible, as a general rule.
3. It's trivial to use a `e => e instanceof MyBaseEvent` predicate to get the `ofInstance()` functionality.
4. I think `ofInstance()` might cause confusion - it's not obvious by the name alone what the difference is between `ofType()` and `ofInstance()`. Both methods take a type - not an instance - as an argument. The distinction is quite subtle and it forces the user to think too much about the internal implementation details. 
5. With `.filter()`, that implementation detail is made explicit in user code.

As this is a new feature, it should be made against the `minor` branch too.

yackinn:
@michaelbromley Please check and verify

packages/core/src/event-bus/event-bus.ts
```diff
@@ -81,9 +81,27 @@ export class EventBus implements OnModuleDestroy {
     ofType<T extends VendureEvent>(type: Type<T>): Observable<T> {
         return this.eventStream.asObservable().pipe(
             takeUntil(this.destroy$),
-            filter(e => (e as any).constructor === type),
+            filter(e => e.constructor === type),
```
yackinn, line 84: The typecast doesn't seem to be needed.

yackinn:
Why are the jobs failing?

michaelbromley:
The jobs are failing because of a config issue after I added the latest plugin - nothing to do with your code.

Thanks very much for your contribution!

PR #1967, opened 2023-01-04
add fioBank and addPaymentToArder

igoriev36:


michaelbromley:
Hi, is this PR intentional?

I've just taken a look at fiobank, and I see it is a Czech bank. I would not add a regional bank to the core plugins, as we only have resources to maintain a few of the biggest global payment providers.

I would encourage you to create a stand-alone plugin which we can then list on our community plugins page.

PR #1987, opened 2023-01-15
fix(docs): expand list of translatable objects

seminarian:


michaelbromley:
`PaymentMethod` does not currently implement the `Translatable` interface, which is why it is not included in that list.

However, that list is perhaps incomplete, as these are the entities which _do_ implement the interface:

image

So we're not currently mentioning facet values, product option groups and product options.

And maybe it is worth noting that the labels & descriptions are also localizable for:

- custom fields
- CollectionFilter
- PaymentMethodHandler
- PromotionAction
- PromotionCondition
- ShippingCalculator
- ShippingEligibilityChecker

docs/content/user-guide/localization/index.md
```diff
@@ -5,14 +5,26 @@ weight: 10
...
+* FacetValue
 * Products
-* Shipping methods
+* ProductOptions
+* ProductOptionGroups
+* ProductVariants
+* ShippingMethods
+
+Vendure supports **admin-facing** (Admin API and Admin UI) localization by allowing you to define translations for labels and descriptions of the following objects:
+* CustomFields
```
michaelbromley, line 22: I think it needs a newline in between the paragraph in order to correctly render the list.

Issue #1997, opened 2023-01-20
Cache-Control header for AssetServer plugin


Draykee:
**Is your feature request related to a problem? Please describe.**
Set a default Cache-Control header for assets served by the AssetServer plugin and make it configurable

**Describe the solution you'd like**
A a configurable Cache-Control header that is set on the file response

Draykee:
https://imagekit.io/blog/ultimate-guide-to-http-caching-for-static-assets/

![image](https://user-images.githubusercontent.com/19676051/213682650-bec9d5ec-8e3c-421e-8a2b-7bb5d639eb38.png)

Draykee:
https://ashton.codes/set-cache-control-max-age-1-year/

This article recommends 1year cache policy, basically because of CDNs being able to cache them for a long time:
![image](https://user-images.githubusercontent.com/19676051/213683227-202e47ec-237e-4ab6-b1c2-f901e9fd5db4.png)

PR #2005, opened 2023-01-23
Added a configurable Cache-Control header

Draykee:
Added a configurable Cache-Control header to asset response #1997

### DIFF
--- a/packages/asset-server-plugin/src/plugin.ts
+++ b/packages/asset-server-plugin/src/plugin.ts
@@ -16,7 +16,7 @@ import fs from 'fs-extra';
 import path from 'path';
 
 import { getValidFormat } from './common';
-import { loggerCtx } from './constants';
+import { DEFAULT_CACHE_HEADER, loggerCtx } from './constants';
 import { defaultAssetStorageStrategyFactory } from './default-asset-storage-strategy-factory';
 import { HashedAssetNamingStrategy } from './hashed-asset-naming-strategy';
 import { SharpAssetPreviewStrategy } from './sharp-asset-preview-strategy';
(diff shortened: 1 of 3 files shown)


packages/asset-server-plugin/src/types.ts
```diff
@@ -117,4 +135,11 @@ export interface AssetServerOptions {
     storageStrategyFactory?: (
         options: AssetServerOptions,
     ) => AssetStorageStrategy | Promise<AssetStorageStrategy>;
+    /**
+     * @description
```
michaelbromley, line 141: Could you add here what the default value is:

```
Defaults to publicly cached for 6 months.

@default 'public, max-age=15552000'
```

packages/asset-server-plugin/src/types.ts
```diff
@@ -43,6 +43,24 @@ export interface ImageTransformPreset {
...
 
+/**
+ * @description
+ * A configuration option for the Cache-Control header in the AssetServerPlugin asset response.
+ *
+ * @docsCategory AssetServerPlugin
+ */
+export type CacheConfig = {
+    /**
+     * The max-age=N response directive indicates that the response remains fresh until N seconds after the response is generated.
```
michaelbromley, line 55: You need to add `@description` tags to properties also so that the text shows up in the docs.

packages/asset-server-plugin/src/plugin.ts
```diff
@@ -232,6 +232,20 @@ export class AssetServerPlugin implements NestModule, OnApplicationBootstrap {
                 }
                 res.contentType(mimeType);
                 res.setHeader('content-security-policy', `default-src 'self'`);
+                // Configure Cache-Control header
```
michaelbromley: I think we should make all of this logic run once on startup, and save the resulting cache header string to a private class field, otherwise it is being run on every single asset request.

You could argue this is a micro-optimization, but I also think it'll make for more readable code.

michaelbromley (changes requested):
Thanks, this looks very nice! Just a few small changes :)

michaelbromley:
Looks like the tests are failing because the `cacheHeader` option needs to be marked as optional.

Draykee:
> Looks like the tests are failing because the `cacheHeader` option needs to be marked as optional.

Sorry, no clue how this happend! Hopefully it's passing now :)

michaelbromley:
Looks good, thanks!

PR #2009, opened 2023-01-26
feat(ui-devkit): Add "exclude" option to UI extensions

vrosa:
### Context

As custom UI extensions mature, a Vendure developer might want to write unit tests or perhaps [Storybook stories](https://storybook.js.org/docs/react/writing-stories/introduction) and choose to colocate them with the corresponding component. Currently, UI Devkit copies **everything** inside the `extensionPath` directory, which could also bring unintended files to the output.

### Proposed solution

Allow developers to optionally specify files within their UI extension to be skipped during the copy process.

### Description of changes

* Update `@vendure/ui-devkit` to support a new optional `exclude` property in `AdminUiExtension` objects, which takes a string array and works the exact same way as [TypeScript's `tsconfig.json` counterpart](https://www.typescriptlang.org/tsconfig#exclude).
* Update `typescript-doc-parser.ts` to support escaped `/*` tokens in comment descriptions. I decided to introduce this additional change because the best example of `exclude` is using a [globstar pattern](https://metacpan.org/dist/File-Globstar/view/lib/File/Globstar.pod), which causes a [multi-line comment to end](https://github.com/microsoft/tsdoc/issues/166). Escaping it using the same existing approach as the `@` token seemed like an obvious choice, however, I'm more than happy not to include changes to that logic and use a different pattern in the `exclude` comment description.

### DIFF
--- a/packages/ui-devkit/src/compiler/scaffold.ts
+++ b/packages/ui-devkit/src/compiler/scaffold.ts
@@ -1,6 +1,7 @@
 /* tslint:disable:no-console */
 import { spawn } from 'child_process';
 import * as fs from 'fs-extra';
+import glob from 'glob';
 import * as path from 'path';
 
 import {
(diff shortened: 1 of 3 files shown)


michaelbromley:
I'm very happy that you and your team are exercising the admin ui extensions heavily and finding these points to improve! Thanks again!

PR #2012, opened 2023-01-30
fix(admin-ui): Redirect to loginUrl if configured on Forbidden GraphQL error

vrosa:
When [`loginUrl`](https://www.vendure.io/docs/typescript-api/admin-ui-plugin/admin-ui-config/#loginurl) is configured for a custom Admin UI, the app should redirect the user to that URL on GraphQL 403 errors and not to the built-in `/login` route. This PR fixes that behaviour - a couple notes:

* The `redirectTo` query string param is not appended to the URL to keep things simple, as `loginUrl` itself can have query string parameters, be a relative path, etc, and the custom login page would require logic similar to `LoginComponent` to decode the redirect URL.
* Code to display an error notification is skipped since the app is terminated once the user is redirected.


michaelbromley:
Thanks!

PR #2016, opened 2023-01-31
fix(core): Fix auth of admin and customer users with the same email

vrosa:
Slack thread [here](https://vendure-ecommerce.slack.com/archives/CKYMF0ZTJ/p1674795935727559).

Anyone can be both customer and an administrator using the same email address, since `identifier` is not a unique key in the `user` table. However, some of the existing logic - most notably authentication - ignores that fact and retrieves the first user in the database with a given email without taking user type (Customer vs Admin) or context (Admin vs Shop API) into account. An example of a consequence can be reproduced following the steps below:

* Register a customer and set, say, test@test.com as their email address
* Create a new administrator in the Admin UI using the same email
* Attempt to login to the Admin UI using the credentials of the administrator created above. We always get an "invalid credentials" error even with the right password

In this case, the issue occurs because the user associated to the customer record was created first, and when matching the provided admin password the user connected to the administrator account is never retrieved.

The main change this PR introduces is an update to the `UserService.getUserByEmailAddress` method, which is ultimately used for, among others:

* Login
* Password reset
* Some customer-related operations
* Resolving the `user` field for customers and administrators

Instead of retrieving the first user with the provided email, it now takes an optional `userType` parameter for cases where user type _**cannot**_ be inferred by the API type (Admin or Shop):

* Resolving `user` for customers, since querying the entity is available in both APIs
* Creating or updating a customer, whose operations are also available in both APIs

In all other cases, user type is inferred by API type.

### DIFF
--- a/packages/core/e2e/auth.e2e-spec.ts
+++ b/packages/core/e2e/auth.e2e-spec.ts
@@ -1,6 +1,6 @@
 /* tslint:disable:no-non-null-assertion */
 import { SUPER_ADMIN_USER_IDENTIFIER, SUPER_ADMIN_USER_PASSWORD } from '@vendure/common/lib/shared-constants';
-import { createTestEnvironment } from '@vendure/testing';
+import { createErrorResultGuard, createTestEnvironment, ErrorResultGuard } from '@vendure/testing';
 import { DocumentNode } from 'graphql';
 import gql from 'graphql-tag';
 import path from 'path';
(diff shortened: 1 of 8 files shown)


vrosa:
> Excellent PR!! Thank you!
> 
> e2e tests are failing though - do you get any failures locally?

Glad to hear!

To be honest, I was so confident that I was running `core` tests individually and commented out my local `pre-push` hook to speed things up and get the changes to remote. It turned out that a very last minute change broke a few of the tests 😬 I added a missing pair of parentheses and a check for an eventual `_entityPrefix` in `dbConnections`. Everything seems to be ok now!

Issue #2015, opened 2023-01-30
Soft deleted product options should be ignored


giovramirez:
**Describe the bug**
Soft deleted product group options are being displayed on the admin UI even if all options are deleted as well.

![image](https://user-images.githubusercontent.com/3448883/215503518-c14ee926-a4fb-40c7-826c-1f798ed08469.png)

![image](https://user-images.githubusercontent.com/3448883/215503711-d7e016d3-6d1f-4dc6-b822-b9c66e28c0c9.png)

Now, If I try to create a new product variant for a product that has a deleted group option, it will fail after a click the **Add X variants** button displaying that's necessary to create new options for the deleted group.

![image](https://user-images.githubusercontent.com/3448883/215505903-cccbee05-aaeb-4f9a-80c5-8b3add760fa5.png)

**To Reproduce**
1. Go to 'Catalog > Products > Edit (A product with a group and options set) > Product variants > Manage variants'
2. Delete all product variants.
3. Delete all group options.
4. Delete group.
5. See error, deleted group still showing up.
6. Create a new group called 'test'.
7. Add new options '1' and '2' for the group 'test'.
8. Fill 'sku' and 'price' for new two product variants.
9. Click on Add 2 variants.
10. See error

**Expected behavior**
1. Deleted groups should not be displayed on the Admin UI.
2. Products with deleted groups should be ignore these groups and create variants only for the active ones.

**Environment (please complete the following information):**
 - @vendure/core version: v1.9.2
 - Nodejs version: 16.15.0
 - Database (mysql/postgres etc): mariadb

giovramirez:
@michaelbromley I want to add some technical details on how to fix these issues. If you agree with them or have some comments about it please let me know! 

I would like to contribute with my first PR to Vendure!

PR #2017, opened 2023-01-31
Fix/ignore soft deleted product options

giovramirez:
Fixes https://github.com/vendure-ecommerce/vendure/issues/2015

**Cause:** If a product group has one or more deleted options, the remove group from product function fails.

**Fix:** Excludes deleted options to not try to redelete them. So, it allows to set deleteAt in the ProductOptionGroup entity correctly.


michaelbromley:
Thank you and congratulations on your first contribution to Vendure core! 🥳

giovramirez:
Thanks Michael! 💪

PR #2018, opened 2023-02-02
fix(admin-ui): Fix UI for long channel list tab on Administrator detail page

vrosa:
Fix small cosmetic issue on the Administrator detail page when the channel list is too long. As it can be seen below, it overflows the content box and doesn't respect the boundaries of the view. A quick solution is to present a horizontal scrollbar when necessary.

#### Before

before_list

#### After

after_list


michaelbromley:
Nice catch, never tested with such long channel names 😆

PR #2020, opened 2023-02-03
chore(job-queue-plugin)!: Update bullmq & redis dependencies

alexisvigoureux:
## 🎯 Goal

Update bullmq and ioredis dependencies

## 📝 Approach

- Removal of the QueueScheduler class from the job-queue-plugin due to the bullmq [refactoring in v2](https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/changelog.md).
- `@types/ioredis` is deprecated, `ioredis` provides its own type definitions now with the v5. The migration was made by following the [guideline](https://github.com/luin/ioredis/wiki/Upgrading-from-v4-to-v5).

BREAKING CHANGE: The minimum Redis recommended version is 6.2.0 [link](https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/changelog.md#breaking-changes-1).
(that’s why the PR is made against the major branch)

## Note
The `getAllBufferIds` function is called 2 times: [call 1](https://github.com/vendure-ecommerce/vendure/blob/master/packages/job-queue-plugin/src/bullmq/redis-job-buffer-storage-strategy.ts#L29), [call2](https://github.com/vendure-ecommerce/vendure/blob/master/packages/job-queue-plugin/src/bullmq/redis-job-buffer-storage-strategy.ts#L40) but the variable is not used. So I didn't add a workaround to the missing [scanstream](https://github.com/luin/ioredis/blob/main/lib/cluster/index.ts#L613) function on Redis cluster


michaelbromley:
Thanks for the PR!

I'll properly review this once I've finished a couple of other tasks on the major branch.

xrev87:
The only thing you might find here is that TypeORM 0.2.45 has a peer dependency on an older version of IORedis

PR #2038, opened 2023-02-14
feat(testing): Turn productsCsvPath into an optional property for test server initialization

vrosa:
Currently, `productsCsvPath` is a required field when [initialising the test server](https://www.vendure.io/docs/developer-guide/testing/#initialize-the-server), regardless whether or not the developer needs product data to begin with. An alternative is to provide a CSV with headers only and not data rows, which is not a big deal but as a matter of convenience it would be nice to skip the property entirely. This PR updates the attribute to be optional.

### DIFF
--- a/docs/content/developer-guide/testing.md
+++ b/docs/content/developer-guide/testing.md
@@ -95,7 +95,7 @@ afterAll(async () => {
 
 An explanation of the options:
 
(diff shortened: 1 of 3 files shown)


michaelbromley:
Yep nice improvement 👍

PR #2047, opened 2023-02-21
fix(core): Filter out deleted options groups

### LINKED ISSUE #2046: Switch products from multiple variants to a single one fails
**Describe the bug**
I'm trying to switch products from multiple variants to a single one. However, [productVariantService.validateVariantOptionIds](https://github.com/vendure-ecommerce/vendure/blob/7b0fae1bda1238d5022178ea4e7729085ead318a/packages/core/src/service/services/product-variant.service.ts#L724) is not filtering out deleted optionGroups so it's throwing this exception: **UserInputError: error.product-variant-option-ids-not-compatible**

**To Reproduce**

Calling these functions from a custom plugin sequentially:

1. Call productVariantService.getVariantsByProductId
2. Call productVariantService.softDelete
3. Call productOptionGroupService.deleteGroupAndOptionsFromProduct
5. Call productVariantService.create

Step **5** fails 

**Expected behavior**
Step **5** should allow to create the new product variant.

 - Nodejs: v16.15.0
 - Database (mysql/postgres etc): mysql

giovramirez:
Fixes #2046

### DIFF
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -741,7 +741,7 @@ export class ProductVariantService {
         if (
             !samplesEach(
                 optionIds,
-                optionGroups.map(g => g.options.map(o => o.id)),
+                activeOptions.map(g => g.options.map(o => o.id)),
             )
         ) {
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thank you!

PR #2055, opened 2023-02-25
Updated sql.js dependency

melanki:


michaelbromley:
Thank you for the contribution!

PR #2056, opened 2023-02-27
EmailPlugin: Feat/dynamic transport settings

### LINKED ISSUE #2043: EmailPlugin: Allow Email SMTP settings per channel
I would like to be able to have different SMTP settings per channel, preferably stored in the database, so that administrators can configure their own mail server.

My initial thought is to transform the `transport` property in the `EmailPlugin.init()` to a function, so that developers can dictate where SMTP settings are coming from:
```ts
    EmailPlugin.init({
      transport: async (ctx) => {

        // Get your SMTP settings based on the requestContext
        const settings = await myService.getSmtpSettings(ctx);

          return {
            type: 'smtp',
            host: settings.host,
            port: settings.port,
            auth: {
              user: settings.user,
              pass: settings.password,
            }
          }
          ...
    })
```
This way we can, for example, get SMTP config stored on custom fields on a Channel

I am currently using a transactional mail service, that also enables me to send emails from different domains with a single SMTP connection, which works fine, but all channels are forced to use the same transactional email service.

I would like to hear your thoughts on this. Is this something that should be included in Vendure, or is this a use case for a separate plugin?

## Additional context
The process function is already async, so it shouldn't be to hard to resolve transport settings here: https://github.com/vendure-ecommerce/vendure/blob/7b0fae1bda1238d5022178ea4e7729085ead318a/packages/email-plugin/src/email-processor.ts#L77

martijnvdbrug:
Closes #2043

Allow dynamic resolving of email transport settings:
```ts
EmailPlugin.init({
    templatePath: path.join(__dirname, '../test-templates'),
    handlers: [],
    transport: async (injector, ctx) => {
        if (ctx) {
          return injector.get(MyTransportService).getSettings(ctx);
        } else {
          return defaultSettings;
        }
    }
});
```

### DIFF
--- a/packages/email-plugin/src/plugin.ts
+++ b/packages/email-plugin/src/plugin.ts
@@ -276,7 +276,7 @@ export class EmailPlugin implements OnApplicationBootstrap, OnApplicationShutdow
         private jobQueueService: JobQueueService,
         private processContext: ProcessContext,
         @Inject(EMAIL_PLUGIN_OPTIONS) private options: EmailPluginOptions,
-    ) { }
+    ) {}
 
     /**
      * Set the plugin options.


packages/email-plugin/src/handlebars-mjml-generator.ts
```diff
@@ -3,8 +3,8 @@ import fs from 'fs-extra';
 import Handlebars from 'handlebars';
 import mjml2html from 'mjml';
 import path from 'path';
-import { EmailGenerator } from './email-generator';
 
+import { EmailGenerator } from './email-generator';
```
martijnvdbrug, line 7: Linter? Not sure what happened here

martijnvdbrug:
Closing this, as I am combining it in a PR with another email issue. See https://github.com/vendure-ecommerce/vendure/pull/2059 for new PR

PR #2058, opened 2023-02-28
Feat/channel aware stripe

martijnvdbrug:
This PR implements config for Stripe per channel: A user can set ApiKey and webhooks per channel via the Admin UI. 

image

It would be nice to encrypt the ApiKey based on an env var, before storing it in the DB, but I can't find a way to hook into the saving of handler args. 

The PR also includes a Stripe dev-server to test out changes during development:
image

E2e tests are still missing the webhook mocking, but that's not part of this PR. I've manually verified payment flow with the dev-server and local tunnel:
image
(admin UI looks nice!)


packages/admin-ui/src/lib/core/src/common/introspection-result.ts
```diff
@@ -1,234 +1,314 @@
...
-        UpdateOrderItemsResult: [
-            'InsufficientStockError',
-            'NegativeQuantityError',
-            'Order',
-            'OrderLimitError',
-            'OrderModificationError',
-        ],
-        UpdatePromotionResult: ['MissingConditionsError', 'Promotion'],
-    },
+      export interface PossibleTypesResultData {
```
martijnvdbrug: I regenerated types. How to prettify everything?

michaelbromley: There should be a pre-commit hook that applies Prettier to all changed files. Is that not working?

michaelbromley:
Thanks for this @martijnvdbrug! I'll review it soon when I get some time to dedicate to this. 

In the mean time I'll invite @vrosa to see if he has any feedback, as the original author of the Stripe integration.

michaelbromley:
By the way, I just pushed a *lot* of changes to the major branch, including all the TypeORM breaking changes. I'm afraid you'll need to do a bit of extra work to integrate these changes.

I made some notes on the main changes in the TypeORM APIs:

Ref: https://typeorm.io/changelog#030httpsgithubcomtypeormtypeormpull8616-2022-03-17

- TypeORM upgrade requires Node.JS version `14+`
- changing where `null` to use `IsNull()`
- changing return types to `null` rather than `undefined`
- change `findOne(id)` to `findOne({ where: { id } })`
- change `where: { order: myOrder }` to `where: { order: { id: myOrder.id }}`
- change `findInIds(ids)` to `find({ where: { id: In(ids) } })`

martijnvdbrug:
> By the way, I just pushed a _lot_ of changes to the major branch, including all the TypeORM breaking changes. I'm afraid you'll need to do a bit of extra work to integrate these changes.
> 
> I made some notes on the main changes in the TypeORM APIs:
> 
> Ref: https://typeorm.io/changelog#030httpsgithubcomtypeormtypeormpull8616-2022-03-17
> 
> * TypeORM upgrade requires Node.JS version `14+`
> * changing where `null` to use `IsNull()`
> * changing return types to `null` rather than `undefined`
> * change `findOne(id)` to `findOne({ where: { id } })`
> * change `where: { order: myOrder }` to `where: { order: { id: myOrder.id }}`
> * change `findInIds(ids)` to `find({ where: { id: In(ids) } })`

vrosa:
> Thanks for this @martijnvdbrug! I'll review it soon when I get some time to dedicate to this.
> 
> In the mean time I'll invite @vrosa to see if he has any feedback, as the original author of the Stripe integration.

Thanks for the heads up, I'll have a closer look later this week if that's ok.

martijnvdbrug:
@michaelbromley Not sure why the generated types changed in core. Looks like formatting. How can I make sure formatting is up to date?

 If I run `yarn format .` in the root of the project it changes **a lot** of files that are not part of this PR. Also, didn't formatting used to be automatic on commit or push?

martijnvdbrug:
@michaelbromley and @vrosa could you guys take a look at this PR, I keep updating the PR because it goes out of sync with the major branch :upside_down_face: 

(@michaelbromley the [generated type formatting issue](https://github.com/vendure-ecommerce/vendure/pull/2058#issuecomment-1466363273) is still there, not sure what to do about it. I had to regenerate types)

martijnvdbrug:
@michaelbromley Is the `major` development ready? I can't build the major branch locally, which results in not being able to build the payments-plugin, because of type mismatches (`DocumentNode`, probably because of version mismatches)

michaelbromley:
@martijnvdbrug You're right, it was in a broken state after a recent merge from master. Should be working now hopefully!

packages/admin-ui/src/lib/core/src/common/introspection-result.ts
martijnvdbrug: Nope, `yarn husky install` gives me `error Command "husky" not found.`. I see the dependency. But this is something for later. I will try to manually lint or prettier

martijnvdbrug: NVM, it is working now, not sure why. I did get an Out Of Memory so I had to change the command to `NODE_OPTIONS=\"--max-old-space-size=8096\" lint-staged`

package.json
```diff
@@ -87,7 +87,7 @@
     "hooks": {
       "commit-msg": "commitlint -e $HUSKY_GIT_PARAMS",
       "post-commit": "git update-index --again",
-      "pre-commit": "lint-staged",
+      "pre-commit": "NODE_OPTIONS=\"--max-old-space-size=8096\" lint-staged",
```
martijnvdbrug, line 90: Not sure if this is an Ubuntu specific issue, but the OOM also happens in my own repo's with eslint, and this is how I fixed it.  https://stackoverflow.com/questions/53230823/fatal-error-ineffective-mark-compacts-near-heap-limit-allocation-failed-javas

martijnvdbrug:
Strange, e2e tests here are also failing with `Error: Cannot find module '@ardatan/aggregate-error'
Require stack:
- /home/martijn/git/vendure/node_modules/@nestjs/graphql/node_modules/@graphql-tools/utils/index.cjs.js`
Is it caching node_modules somehow? Yarn.lock is not part of this PR

michaelbromley:
Not sure why, maybe caching? Anyway, I found a few more issues with e2e tests and just pushed some changes to major.

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -43,46 +43,39 @@ export class StripeController {
...
-            event = this.stripeService.constructEventFromPayload(request.rawBody, signature);
-        } catch (e: any) {
-            Logger.error(`${signatureErrorMessage} ${signature}: ${JSON.stringify(e.message)}`, loggerCtx);
-            response.status(HttpStatus.BAD_REQUEST).send(signatureErrorMessage);
-            return;
-        }
-
+        const event = request.body as Stripe.Event;
         const paymentIntent = event.data.object as Stripe.PaymentIntent;
+        const { metadata: { channelToken, orderCode, orderId } = {} } = paymentIntent;
```
vrosa: Destructuring here is going to throw an error in case `paymentIntent` is `null` or `undefined`. These new lines should be moved to after the `if (!paymentIntent)` check, since retrieving the order depends on a valid payment intent in the first place:

```ts
const paymentIntent = event.data.object as Stripe.PaymentIntent;
if (!paymentIntent) {
  Logger.error(noPaymentIntentErrorMessage, loggerCtx);
  response.status(HttpStatus.BAD_REQUEST).send(noPaymentIntentErrorMessage);
  return;
}

const { metadata: { channelToken, orderCode, orderId } = {} } = paymentIntent;
// more code
```

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -43,46 +43,39 @@ export class StripeController {
...
             Logger.info(`Received ${event.type} status update for order ${orderCode}`, loggerCtx);
             return;
         }
-
-        const ctx = await this.createContext(channelToken, request);
-
-        const order = await this.orderService.findOneByCode(ctx, orderCode);
-        if (!order) {
-            throw Error(`Unable to find order ${orderCode}, unable to settle payment ${paymentIntent.id}!`);
+        try {
```
vrosa: I reckon it makes the most sense to have this check right after an order is retrieved to ensure we're processing an actual message from Stripe before proceeding any further:

```ts
if (!order) {
  throw Error(`Unable to find order ${orderCode}, unable to settle payment ${paymentIntent.id}!`);
}

try {
  event = this.stripeService.constructEventFromPayload(request.rawBody, signature);
}
// more code
```

packages/payments-plugin/src/stripe/stripe.handler.ts
```diff
@@ -50,39 +72,47 @@ export const stripePaymentMethodHandler = new PaymentMethodHandler({
     },
 
     async createRefund(ctx, input, amount, order, payment, args): Promise<CreateRefundResult> {
-        const result = await stripeService.createRefund(payment.transactionId, amount);
```
vrosa, line 53: Out of interest, why did you remove `createRefund` from `StripeService` and moved the code here? The idea was to encapsulate interactions with the `Stripe` object inside the service so the handler doesn't need to know about the Stripe API, e.g. `stripe.refunds.create`.

packages/payments-plugin/src/stripe/stripe-client.ts
```diff
@@ -0,0 +1,17 @@
+import Stripe from 'stripe';
+
+/**
+ * Wrapper around the Stripe client that exposes ApiKey and WebhookSecret
+ */
+export class VendureStripeClient extends Stripe {
+
+    constructor(
+        public apiKey: string,
```
vrosa: It's a minor thing but from what I could tell `apiKey` doesn't need to be public as it's not referenced anywhere unlike `webhookSecret` [here](https://github.com/vendure-ecommerce/vendure/pull/2058/files#diff-6d83de7289055b3d0fddfd6ca1e7e3343c119d4f650173bfd30194c56c61f935R73).

packages/payments-plugin/src/stripe/stripe.controller.ts
martijnvdbrug: Good point. fixed

packages/payments-plugin/src/stripe/stripe.controller.ts
martijnvdbrug: Also valid feedback. Fixed

packages/payments-plugin/src/stripe/stripe.handler.ts
martijnvdbrug: Why I did it: simplicity, the function was just proxying to the stripe client.

I have changed it back, encapsulation the stripe client inside the service is also a good point.

packages/payments-plugin/src/stripe/stripe-client.ts
martijnvdbrug: You're right, it can be private.

yasserlens:
I'd very much appreciate having an update on where this is at the moment - I eagerly need it to launch a new store on the same vendure instance that already serves another store. Thank you for your work on this! 🙏

martijnvdbrug:
@michaelbromley Breaking change docs:

## Updating Stripe from `v1.x` to `v2`

The Stripe plugin has been made channel aware. This means your api key and webhook secret are now stored in the database, per channel, instead of environment variables.

To migrate to `v2` of the Stripe plugin from `@vendure/payments` you need to:
1. Remove the `apiKey` and `webhookSigningSecret` from the plugin initialization in `vendure-config.ts`:
```diff
-StripePlugin.init({
-    apiKey: process.env.YOUR_STRIPE_SECRET_KEY,
-    webhookSigningSecret: process.env.YOUR_STRIPE_WEBHOOK_SIGNING_SECRET,
-    storeCustomersInStripe: true,
-}),
+StripePlugin.init({
+    storeCustomersInStripe: true,
+ }),
```
2. Start the server and login as administrator.
3. For each channel that you'd like to use Stripe payments, you need to create a payment method with payment handler `Stripe payment` and the `apiKey` and `webhookSigningSecret` belonging to that channel's Stripe account.

image

michaelbromley:
Thanks @martijnvdbrug! 

This will be in 2.0.0-beta.2, expected next week 🥳

PR #2062, opened 2023-03-01
Fix: Merge nested relations correctly

jnugh:
We have a custom field for `Order` that contains a relation list. Something like this:

```ts
{
  name: "xxx",
  type: "relation",
  list: true,
  entity: TargetEntity,
  readonly: true,
  internal: true,
},
```

the relation works great when reading the order directly using the `order(id: "orderId")` query. However when reading it as part of a list via `activeCustomer` the relation is not populated correctly. This is because the current version:

```ts
finalEntity[entry.relation] = entry.entity[entry.relation];
```

doe snot work for nested relations. In our case `entry.relation` would have been `customFields.xxx`.

### DIFF
--- a/packages/core/e2e/fixtures/test-plugins/list-query-plugin.ts
+++ b/packages/core/e2e/fixtures/test-plugins/list-query-plugin.ts
@@ -6,6 +6,7 @@ import {
     Ctx,
     Customer,
     CustomerService,
+    HasCustomFields,
     ListQueryBuilder,
     LocaleString,
     Order,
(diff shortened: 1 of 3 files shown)


michaelbromley:
Hi!
Thanks for the PR!
Is there any chance you'd have time to write an e2e test which fails before your fix and passes after it? ListQueryBuilder is a really key part of the code base, and I like to make sure all behaviour is well tested because it is so complex too.

jnugh:
Hi, sure thing. Very good idea indeed - this made me understand the initial problem a little better.
This is not really about multi level nested relations but about the `customFields` aliasing that also uses `.` as a separator.
When resoling a normal relation the relations array will always contain each item of the path which is why this works most of the time.

Se lets's say we resolve `order.lines.productVariant` the relations array will contain `['lines', 'lines.productVariant']` thus the original code incorrectly copied `order['lines.productVariant']` but still correctly applied `order['lines']` therefore this will work.

When using `customFields` no join needs to be made. Thus the relations array looks like this: `['customFields.relation']` therefore `order['customFields']` will never be applied. It might also be possible to only apply the first segment without recursion, not sure if this makes a real difference.

Anyways - the test case I added will fail with `Cannot return null for non-nullable field TestEntity.customFields.` without the fix.

michaelbromley:
Thanks! 🙏

PR #2072, opened 2023-03-10
fix(docs): make consistent with other docs

seminarian:
I noticed that the value of APP_ENV doens't match what's being used in vendure-config.ts or on https://www.vendure.io/docs/deployment/production-configuration/

### DIFF
--- a/packages/harden-plugin/src/harden.plugin.ts
+++ b/packages/harden-plugin/src/harden.plugin.ts
@@ -35,7 +35,7 @@ import { HardenPluginOptions } from './types';
  *   plugins: [
  *      HardenPlugin.init({
  *        maxQueryComplexity: 650,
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks! I think we should use `APP_ENV` since this is the variable we ship with in the `@vendure/create` standard installation, and the standard `NODE_ENV` would have a value of `'development'` anyway.

seminarian:
> Thanks! I think we should use `APP_ENV` since this is the variable we ship with in the `@vendure/create` standard installation, and the standard `NODE_ENV` would have a value of `'development'` anyway.

O yeah good catch confused those two! Should be fixed now?

michaelbromley:
Perfect, thanks!

PR #2075, opened 2023-03-14
fix(payments-plugin): multiple payments and stock check

### LINKED ISSUE #2026: Mollie Payment Plugin is not deducting payments already present on order (e.g. giftcards)
**Describe the bug**
Payments already present on the order are not deducted from the total amount passed to mollie.

**To Reproduce**
1. Add a payment which doesnt cover the whole amount of the order
2. Add a mollie payment intent
3. See the total amount, not the amount which is left to pay

**Expected behavior**
Deduct payments already present in the order.

**Additional context**
My proposal would be to import the function `totalCoveredByPayments` and change [this line](https://github.com/vendure-ecommerce/vendure/blob/7fdd875c770eb87b6921f9780be5fb8921e61588/packages/payments-plugin/src/mollie/mollie.service.ts#L117) to
`amount: toAmount(order.totalWithTax - totalCoveredByPayments(order), order.currencyCode),`

### LINKED ISSUE #2030: Can not add payment to order when one of the items is out of stock
**Describe the bug**

An order can not be transitioned to ArrangingPayment when one of the items is out of stock. This causes errors for payment methods that use a paymentIntent-like setup (Mollie and Stripe), because the order stays in AddingItems, and is only transitioned to ArrangingPayment **after** a payment has been made.

**To Reproduce**
1. Allocate stock only when transitioned to PaymentSettled using the StockAllocationStrategy. `return toState === 'PaymentSettled';`
2. Add item X to cart
3. Change the stock of X in the admin ui to 0
4. Pay for the order using Mollie
5. Order will stay in AddingItems even though the customer has made a payment.

**Expected behavior**
**Option 1.** Check before creating payment intent. Can we do some kind of transition dry-run to see if the order is "transitionable" to ArrangingPayment? Or should we add the stock check manually to the Mollie and Stripe plugin

**Option 2.** Try to transition to ArrangingPayment before redirecting to Stripe or Mollie. Downside here is that the customer can't edit the cart anymore if he decides not to pay, and thus returning to the storefront that now has an order in the ArrangingPayment state.

**Option 3:** Use stock allocation. This will only sort the problem for my shop, other users can run into the same issue when using stock allocation on settlement.

 - @vendure/core version: 1.9.2
 - Nodejs version: 16

martijnvdbrug:
* Account for payments that have already been made on an order. Closes  #2026 
* Check salable stock before creating a payment intent. Closes #2030

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1070,7 +1070,6 @@ export class OrderService {
         if (payment.state === 'Declined') {
             return new PaymentDeclinedError(payment.errorMessage || '');
         }
-
         return this.transitionOrderIfTotalIsCovered(ctx, order);
     }
 
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1070,6 +1070,7 @@ export class OrderService {
         if (payment.state === 'Declined') {
             return new PaymentDeclinedError(payment.errorMessage || '');
         }
+        
         return this.transitionOrderIfTotalIsCovered(ctx, order);
     }
 
--- a/packages/payments-plugin/e2e/mollie-dev-server.ts
+++ b/packages/payments-plugin/e2e/mollie-dev-server.ts
@@ -2,10 +2,12 @@ import { AdminUiPlugin } from '@vendure/admin-ui-plugin';
 import {
     ChannelService,
     DefaultLogger,
+    DefaultSearchPlugin,
     Logger,
     LogLevel,
     mergeConfig,
     OrderService,
+    PaymentService,
     RequestContext,
 } from '@vendure/core';
 import { createTestEnvironment, registerInitializer, SqljsInitializer, testConfig } from '@vendure/testing';
@@ -35,6 +37,7 @@ import { CREATE_MOLLIE_PAYMENT_INTENT, setShipping } from './payment-helpers';
     const config = mergeConfig(testConfig, {
         plugins: [
             ...testConfig.plugins,
+            DefaultSearchPlugin,
             AdminUiPlugin.init({
                 route: 'admin',
                 port: 5001,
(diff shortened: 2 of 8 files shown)


michaelbromley:
It looks like one of the e2e tests is flaky - it is sometimes passing and sometimes failing:

```
Mollie payments › Should successfully refund the Mollie payment
@vendure/payments-plugin:     Nock: No match for request {
@vendure/payments-plugin:       "method": "GET",
@vendure/payments-plugin:       "url": "https://api.mollie.com/v2/orders/ord_mockId?embed=payments",
@vendure/payments-plugin:       "headers": {
@vendure/payments-plugin:         "accept": "application/hal+json",
@vendure/payments-plugin:         "user-agent": "Node/v14.21.3 Mollie/3.6.0",
@vendure/payments-plugin:         "authorization": "***",
@vendure/payments-plugin:         "accept-encoding": "gzip",
@vendure/payments-plugin:         "content-type": "application/json"
@vendure/payments-plugin:       }
@vendure/payments-plugin:     }
```

martijnvdbrug:
I couldn't exactly reproduce, but after disabled the failing test, some other test failed. That indicates that one of the Nock mock requests was being used by multiple testcases. I've refactored all testcases to use exact path matching (no wildcard mock requets) to prevent this. 
I'm pretty confident this works, but I can't be sure, because can't reproduce.

michaelbromley:
Yeah I just pulled this branch locally and all tests pass for me too. 😠  Gonna dig around a bit...

martijnvdbrug:
There is a dirty hack: just make the mock request persistent with `persist()`, but that doesn't really fix the root cause, and thra frustrates me.... 

Looking into it as well

michaelbromley:
Established that the test `Should successfully refund the Mollie payment` will consistently fail when tested against postgres or mysql locally too.

michaelbromley:
Merged in https://github.com/vendure-ecommerce/vendure/commit/1aad00e34e456aeae0e33bebf68d7af6cc9f3d03

PR #2078, opened 2023-03-15
fix(docs): remove wrong example

seminarian:
`port` property is not listed in the `AssetServerOptions` (anymore?)


michaelbromley:
Ah thanks, at one time we ran it as a separate express instance but since 5db92d15 it runs as middleware.

PR #2084, opened 2023-03-17
feat(test): Add tokenMethod cookie support for TestServer

tianyingchun:
has testing in my local `esm` project.


michaelbromley:
谢谢!

PR #2088, opened 2023-03-20
fix(dev-server): Use env variables for db config

prasmalla:
replace hard-coded values with env variables for getDbConfig


michaelbromley:
Thanks and congrats on your first PR to Vendure! 🥳 🙌

prasmalla:
🙌 thank you)

PR #2091, opened 2023-03-21
fix(core): Slugify product created with fast importer service

### LINKED ISSUE #2053: Slugify mechanism does not work in the create method of productService, only in update
**Describe the bug**
The slugify mechanism is not execute in the create method of the ProductService

**To Reproduce**
```php
const productTranslation: ProductTranslationInput = {
            name: 'test product',
            slug: 'test product',
            description: 'test description'
        }
        const createProductInput: CreateProductInput = {
            translations: [productTranslation]
        };

        const createdProductId = await this.fastImporterService.createProduct(createProductInput);
```
This code will create product with `test product` name but also the slug will be `test product`.

Now on the update action:
```php 
        const productTranslation: ProductTranslationInput = {
            id: 1,
            name: 'test product',
            slug: 'test product',
            description: 'test description'
        }
        const updateProductInput: UpdateProductInput = {
            id: 1,
            translations: [productTranslation]
        }

        await this.productService.update(ctx, updateProductInput)
```
with the same parametrs like before, but now when we update the product the slug is now slugifed and it is `test-product` value.

**Expected behavior**
Slugify mechanism should be applied in both create and update method of ProductService

 - @vendure/core version:12.2.16
 - Nodejs version: v18.14.0
 - Database (mysql/postgres etc): mysql 5.7

prasmalla:
implements slug normalization and related test
fixes #2053

adds fallback values to dbConfig for dev-server to run scripts from cli

### DIFF
--- a/packages/core/e2e/fast-importer.e2e-spec.ts
+++ b/packages/core/e2e/fast-importer.e2e-spec.ts
@@ -1,0 +1,52 @@
+import { CreateProductInput, ProductTranslationInput } from '@vendure/common/lib/generated-types';
+import { createTestEnvironment } from '@vendure/testing';
+import path from 'path';
+
+import { testConfig, TEST_SETUP_TIMEOUT_MS } from '../../../e2e-common/test-config';
+import { initialData } from '../mock-data/data-sources/initial-data';
+import { FastImporterService, LanguageCode } from '../src';
+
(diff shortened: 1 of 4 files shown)


michaelbromley:
Thanks! 💯

PR #2092, opened 2023-03-22
feat(payments-plugin): allow partial payments

martijnvdbrug:


packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -156,334 +157,370 @@ describe('Mollie payments', () => {
...
-            },
-        );
-        order = addItemToOrder as TestOrderFragmentFragment;
-        // Add surcharge
-        const ctx = new RequestContext({
-            apiType: 'admin',
-            isAuthorized: true,
-            authorizedAsOwnerOnly: false,
-            channel: await server.app.get(ChannelService).getDefaultChannel(),
+    describe('Payment intent creation', () => {
```
martijnvdbrug, line 159: Sorry for the large diff, but I grouped the testcases for readability: 
image

packages/payments-plugin/src/mollie/mollie.helpers.ts
```diff
@@ -79,6 +79,13 @@ export function toAmount(value: number, orderCurrency: string): Amount {
     };
 }
 
+/**
+ * Return to number of cents
+ */
+export function amountToCents(amount: Amount): number {
+    return Math.round(Number(amount.value) * 100); // Round because floating point errors after Number() conversion
```
StampixSMO: I would suggest to use a dedicated library for any currency conversions/math to avoid rounding issues in general:
image

michaelbromley: @martijnvdbrug Interesting - I was wondering about this and I see from the [Mollie SDK](https://github.com/mollie/mollie-api-node) that `amount.value` is a string like `'10.00'`. Checked some values and indeed you do get rounding issues!
```ts
Number('10.21') * 100
// => 1021.0000000000001
```

In general @StampixSMO I agree we need to be cautious with math on currency - that's why Vendure only stores integers for currency values. In this case, I think the handling is fine because we're dealing here with an artifact of "floating-point numbers in binary" rather than an actual rounding error caused by multiplying by a non-integer, so the error is only ever so tiny as to be ignored.

StampixSMO: Mollie returns a string on purpose, it avoids FP errors due to frameworks along the way. It's meant to be used with a professional library like currency.js for math and parsing.

Thanks for confirming the Vendure strategy. With integers, you implicitly mean that you store the least significant currency unit, right? E.g. dollarcents, eurocents, ... instead of dollars and euros.

martijnvdbrug: Oke, the library is tiny, so I'll add it :-)

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -156,334 +156,370 @@ describe('Mollie payments', () => {
...
+                .post('/v2/payments/tr_mockPayment/refunds', body => {
+                    mollieRequest = body;
+                    return true;
+                })
+                .reply(200, { status: 'pending', resource: 'payment' });
+            const refund = await refundOrderLine(
+                adminClient,
+                order.lines[0].id,
+                10,
+                order.payments[2].id,
```
martijnvdbrug, line 436: Use the payments[2] now instead of [1] here, because we've added an extra test with partial payment

martijnvdbrug:
@michaelbromley Reallyyyy..., now its only the MySQL test that's failing. Same error. I might need your help on this, still don't quite understand whats going on under the hood with the multiple refunds.

packages/payments-plugin/src/mollie/mollie.helpers.ts
michaelbromley: @StampixSMO yes, we store the minor unit.

PR #2094, opened 2023-03-22
feat(payments-plugin): mollie plugin make `redirecturl` dynamic

### LINKED ISSUE #2093: PaymentsPlugin: make `redirectUrl` dynamic for a mollie payment method
The current version of the mollie payment plugin requires setting of a `redirectUrl` in the payment method through Admin UI.
Currently it's not possible to use one payment method (with one mollie API key) for different consumers of the Ecommerce Engine. E.g. Web Frontend, Mobile app, ...

Add a `redirectUrl` property to the input of the `createMolliePaymentIntent` mutation. This lets the consumer choose which URL to redirect to after successfully paying on the Mollie website.

Creating multiple payment methods per consumer (apps, web frontends, preview branches,..) 
Gets messy fast though, doesnt work for Vercel preview branches.

**Additional context**

seminarian:
closes #2093

It's my first contribution, best to check everything well.
Also didn't know if I had to add an `@since` tag somewhere and is the Breaking change in the commit message correct?

### DIFF
--- a/packages/payments-plugin/src/mollie/mollie.handler.ts
+++ b/packages/payments-plugin/src/mollie/mollie.handler.ts
@@ -35,6 +35,15 @@ export const molliePaymentHandler = new PaymentMethodHandler({
                 { languageCode: LanguageCode.en, value: 'This option only affects pay-later methods. Automatically capture payments immediately after authorization. Without autoCapture orders will remain in the PaymentAuthorized state, and you need to manually settle payments to get paid.' },
             ],
         },
+        redirectUrl: {
+            type: 'string',
+            required: false,
+            defaultValue: '',
+            label: [{ languageCode: LanguageCode.en, value: 'Redirect URL' }],
+            description: [
+                { languageCode: LanguageCode.en, value: 'Redirect the client to this URL after payment' },
+            ],
+        }
     },
     init(injector) {
         mollieService = injector.get(MollieService);
--- a/packages/payments-plugin/src/mollie/mollie.handler.ts
+++ b/packages/payments-plugin/src/mollie/mollie.handler.ts
@@ -56,7 +56,7 @@ export const molliePaymentHandler = new PaymentMethodHandler({
         metadata,
     ): Promise<CreatePaymentResult | CreatePaymentErrorResult> => {
         // Only Admins and internal calls should be allowed to settle and authorize payments
-        if (ctx.apiType !== 'admin') {
+        if (ctx.apiType !== 'admin' && ctx.apiType !== 'custom') {
             throw Error(`CreatePayment is not allowed for apiType '${ctx.apiType}'`);
         }
         if (metadata.status !== 'Authorized' && metadata.status !== 'Settled') {
--- a/packages/payments-plugin/src/mollie/mollie.handler.ts
+++ b/packages/payments-plugin/src/mollie/mollie.handler.ts
@@ -27,13 +27,6 @@ export const molliePaymentHandler = new PaymentMethodHandler({
             type: 'string',
             label: [{ languageCode: LanguageCode.en, value: 'API Key' }],
         },
-        redirectUrl: {
-            type: 'string',
-            label: [{ languageCode: LanguageCode.en, value: 'Redirect URL' }],
-            description: [
-                { languageCode: LanguageCode.en, value: 'Redirect the client to this URL after payment' },
-            ],
-        },
         autoCapture: {
             type: 'boolean',
             label: [{ languageCode: LanguageCode.en, value: 'Auto capture payments' }],
--- a/packages/payments-plugin/src/mollie/mollie.plugin.ts
+++ b/packages/payments-plugin/src/mollie/mollie.plugin.ts
@@ -21,6 +21,14 @@ export interface MolliePluginOptions {
      * This is used by Mollie to send webhook events to the Vendure server
      */
     vendureHost: string;
+
+    /**
+     * @description
+     * For backwards compatibility, by default set to false.
+     * When enabled, the `redirectUrl` can be passed via the `createPaymentIntent` mutation 
+     * instead of being configured in the Payment Method.
+     */
+    useDynamicRedirectUrl?: boolean;
 }
 
 /**
@@ -144,6 +152,7 @@ export class MolliePlugin {
      * @description
      * Initialize the mollie payment plugin
      * @param vendureHost is needed to pass to mollie for callback
+     * @param useDynamicRedirectUrl to indicate if the redirectUrl can be passed via the `createPaymentIntent` mutation, versus being configured in the Payment Method.
      */
     static init(options: MolliePluginOptions): typeof MolliePlugin {
         this.options = options;
--- a/packages/payments-plugin/src/mollie/mollie.plugin.ts
+++ b/packages/payments-plugin/src/mollie/mollie.plugin.ts
@@ -52,17 +52,18 @@ export interface MolliePluginOptions {
  *     ]
  *     ```
  * 2. Create a new PaymentMethod in the Admin UI, and select "Mollie payments" as the handler.
- * 3. Set the Redirect URL. This is the url that is used to redirect the end-user, e.g. `https://storefront/order`
- * 4. Set your Mollie apiKey in the `API Key` field.
+ * 3. Set your Mollie apiKey in the `API Key` field.
  *
  * ## Storefront usage
  *
  * In your storefront you add a payment to an order using the `createMolliePaymentIntent` mutation. In this example, our Mollie
- * PaymentMethod was given the code "mollie-payment-method".
+ * PaymentMethod was given the code "mollie-payment-method". The `redirectUrl``is the url that is used to redirect the end-user
+ * back to your storefront after completing the payment. The order code is appened to the `redirectUrl`.
  *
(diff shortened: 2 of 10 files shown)


martijnvdbrug:
> Hi, congrats on your first PR to Vendure! partying_face
> 
> So I'd prefer not to make this a breaking change if we can avoid that. How about we keep the `redirectUrl` config arg, and make the new GraphQL input field optional. So the logic would be:
> 
> 1. If the `redirectUrl` is passed in the `createMolliePaymentIntent` mutation, we use it.
> 2. If not, we use the configured `redirectUrl` from the PaymentMethod config args.
> 
> We discussed this via Slack and the conclusion is to:
> 
> 1. Keep the configArgs so it is not a breaking change
> 2. Add a new config arg to toggle whether the redirectUrl should come from the config arg or from the mutation input
> 3. Add validation to enforce this

Oops, this might have been my suggestion. My argument is: The same and more can be accomplished with the redirectUrl via graphql, so why maintain both ways?

seminarian:
Alright well I did the changes as Michael requested them. 
Once the final decision has been made I'll do a rebase to clean it up and select the version we want to keep.

About this version:
Only did the tests for `useDynamicRedirectUrl` set to false. Added some tests to test the validation.

Doing the tests for `useDynamicRedirectUrl` would probably have changed a too huge of a diff and make it impossible to merge with Martijns changes in https://github.com/vendure-ecommerce/vendure/pull/2092

Let me know what you think.

packages/payments-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -21,6 +21,14 @@ export interface MolliePluginOptions {
      * This is used by Mollie to send webhook events to the Vendure server
      */
     vendureHost: string;
+
+    /**
+     * @description
+     * For backwards compatibility, by default set to false.
+     * When enabled, the `redirectUrl` can be passed via the `createPaymentIntent` mutation 
+     * instead of being configured in the Payment Method.
+     */
```
michaelbromley, line 34: You can add 

```
@default false
@since 2.0.0
```

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -101,17 +103,32 @@ export class MollieService {
         if (!paymentMethod) {
             return new PaymentIntentError(`No paymentMethod found with code ${paymentMethodCode}`);
         }
+        if (this.options.useDynamicRedirectUrl == true) {
+            if (!input.redirectUrl) {
+                return new InvalidInputError(`Cannot create payment intent without redirectUrlspecified`);
```
michaelbromley: Missing a space "redirectUrlspecified"

michaelbromley:
> Oops, this might have been my suggestion. My argument is: The same and more can be accomplished with the redirectUrl via graphql, so why maintain both ways?

@martijnvdbrug yeah that's a valid point. My counter-argument was that with v2 there are already lots of breaking changes for people to deal with, so I wanted to avoid another one here. We can certainly deprecate the old way and remove it in a future version though. In fact, @seminarian it is probably worth starting that process now and mentioning that the configArg will be deprecated in the documentation.

@seminarian it looks good to me other than the couple of small comments.

packages/payments-plugin/src/mollie/mollie.handler.ts
```diff
@@ -27,13 +27,6 @@ export const molliePaymentHandler = new PaymentMethodHandler({
             type: 'string',
             label: [{ languageCode: LanguageCode.en, value: 'API Key' }],
         },
-        redirectUrl: {
-            type: 'string',
-            label: [{ languageCode: LanguageCode.en, value: 'Redirect URL' }],
-            description: [
-                { languageCode: LanguageCode.en, value: 'Redirect the client to this URL after payment' },
```
martijnvdbrug, line 34: Can/should we also add a deprecation message here?

michaelbromley:
@seminarian Martijn's other Mollie PR is now merged in and the `minor` branch includes it.

StampixSMO:
@michaelbromley @seminarian @martijnvdbrug I would suggest to not append the order code to the end of the incoming `redirectUrl`. It forces the storefront to be able to accept that order code as a last path parameter. Wouldn't it be better to have the storefront decide on what it can or cannot handle? I'm thinking: native mobile apps, less flexible frontend frameworks where query parameters are preferred, ...

So basically, the frontend would send the full `redirectUrl`, e.g.:
-  `https://storefront.tld/order/[code]/payment` 
- `storefront://check-order-payment?order=[code]` (native app)
- ...

packages/payments-plugin/src/mollie/mollie.plugin.ts
seminarian: It's being merged to minor right? But still only @since 2.0.0 ? Would expect that to be the next minor release actually?

michaelbromley: There won't be another minor release!

michaelbromley:
@StampixSMO Great point, I agree.

StampixSMO:
So while working on this, I bumped on another issue that is actually on Mollie's side.

**Issue**

If you don't specify a payment method, selecting a `failed` payment status on the Mollie hosted page results in being redirected to the payment selection screen and the redirection is never triggered.

I have attached a year-old screenshot from a support case where I asked about this bug and what they recommended as solution.

image

**Solution**

The bug is gone if you select a payment method upfront when making a test payment. For example `paypal` (`creditcard` is annoying as you need to enter actual credit card numbers).

You could have the storefront request a certain payment method already, using the exposed input variables. However, that's not ideal since the storefront has no idea whether Vendure+Mollie is running in test mode or not.

I would suggest that we force a payment method when the API key is a test key (can be checked using the `test_` prefix).

Let me know what you think.

seminarian:
@michaelbromley Could you enlighten me what's actually the role of `ArrangingPayment` status in Vendure:
- I noticed that the mollie plugin skips out on that, here you can see the docs: https://www.vendure.io/docs/typescript-api/payments-plugin/mollie-plugin/
- The storefront guide mentions `transitionToState` mutation calls: https://www.vendure.io/docs/storefront/shop-api-guide/#order-flow

Like, does the not maintaining the "correct" Order state have any implications? Is that something we would want to see streamlined/added to the Mollie plugin?

seminarian:
@martijnvdbrug I've managed to get the same payment multiple times listed in Vendure:
image

I was playing around with introducing exceptions in random places and ended up in place like this.

I'd like to add the `@Transaction()` decorator to the `mollie.controller.ts`:
```
    @Post('mollie/:channelToken/:paymentMethodId')
    @Transaction()
    async webhook(
        @Param('channelToken') channelToken: string,
        @Param('paymentMethodId') paymentMethodId: string,
        @Body() body: any,
    ): Promise<void> {
        if (!body.id) {
            return Logger.warn(` Ignoring incoming webhook, because it has no body.id.`, loggerCtx);
        }
        try {
            await this.mollieService.handleMollieStatusUpdate({ channelToken, paymentMethodId, orderId: body.id });
        } catch (error) {
            Logger.error(`Failed to process incoming webhook: ${error?.message}`, loggerCtx, error);
            throw error;
        }
    }
```

I dont' there's a reason not to right?

martijnvdbrug:
@seminarian That's a good improvement! 

Just make sure that we can't end up with a scenario where the actual payment has been done through Mollie successfully, but reverted in Vendure. As long as the error boils up to the controller, and thus returning a non-200 to Mollie, we should be good, because Mollie will retry the webhook

michaelbromley:
Hi @seminarian & @martijnvdbrug 
Apologies for the lack of response from my side - I was wrapped up in getting the beta release out and a bunch of other things.

> Could you enlighten me what's actually the role of ArrangingPayment status in Vendure:

The original intent is that in this state:

1. The order cannot be modified, so the price will not change
2. The order is not yet placed because payment has not completed yet.

It turns out that there are a wide variety of flows with different payment providers, so e.g. in the Mollie integration we don't require the storefront the do the transition as it is handled internally.

> I've managed to get the same payment multiple times listed in Vendure:

Is any further action needed - do you intend to add the `Transaction` decorator as suggested?

martijnvdbrug:
@seminarian let me know if there are any open points, could do a short call if needed

seminarian:
> It turns out that there are a wide variety of flows with different payment providers, so e.g. in the Mollie integration we don't require the storefront the do the transition as it is handled internally.
For our usecase, a photo site, it would sure come in handy to hook into `onTransitionStart` and perform validation if OrderLines are properly configured. 
Though we can just let our FE go to `ArrangingPayment` first I guess :-)

> Is any further action needed - do you intend to add the Transaction decorator as suggested?
I've pushed the changes!

@michaelbromley and @martijnvdbrug perhaps one of you guys could audit my changes or test it.. Thanks! :)

seminarian:
I've discovered a small validation issue. The Mollie library throws if `firstName`, `lastName` or `emailAddress` are empty strings. Will add a check for that.
image
image

seminarian:
@michaelbromley & @martijnvdbrug done!

martijnvdbrug (approved):
LGTM!

michaelbromley:
Excellent work @seminarian! This will go in the next v2 beta release.

PR #2095, opened 2023-03-22
fix(core): Filter collections by language

### LINKED ISSUE #2064: Collections not filterable by languageCode
The documentation lists `languageCode` as a filter field on a collections list.
When trying to use that option, I get an error (`error.invalid-filter-field`)

**To Reproduce**
Steps to reproduce the behaviour:
1. Query the shop API with this query:
```graphql
{
  collections(options: {
    filter: {
      languageCode: { eq: "en" }  
    }
  }) {
    items {
      name
    }
  }
}
```

**Expected behaviour**
Collections list, filtered by languageCode

 - Nodejs version: 16.19.0
 - Database (mysql/postgres etc): sqlite

prasmalla:
Fixes #2064

### DIFF
--- a/packages/core/src/service/helpers/list-query-builder/connection-utils.ts
+++ b/packages/core/src/service/helpers/list-query-builder/connection-utils.ts
@@ -17,12 +17,7 @@ export function getColumnMetadata<T>(connection: Connection, entity: Type<T>) {
 
     const translationRelation = relations.find(r => r.propertyName === 'translations');
     if (translationRelation) {
-        const commonFields: Array<keyof (Translation<T> & VendureEntity)> = [
-            'id',
-            'createdAt',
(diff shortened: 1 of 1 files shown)


michaelbromley:
I've been thinking about this more, and I'm not sure now whether this is the right approach. It might be more correct to just _disallow_ filtering by `languageCode` altogether.

The reason is that all requests can already take a `?languageCode=xx` query param. If present, this sets the language on the RequestContext that is then used throughout the code. The ListQueryBuilder already has mechanisms in place to detect and limit the results according to this `languageCode` query param, including handling fallbacks.

prasmalla:
have added notes to the [issue](https://github.com/vendure-ecommerce/vendure/issues/2064) closing this PR

PR #2102, opened 2023-03-28
feat(asset-server-plugin) s3 asset storage strategy

asonnleitner:
It is hard to test, because I am not really able to start the dev environment


asonnleitner:
Here are also some changes I would recommend to implement for the s3-asset-storage-strategy
- Switch from the aws sdk v2 compatible style to use the modular v3 sdk. As the v2 compatible style
  may be dropped in the future, and also this approach results in a smaller bundle size, if any bundler is used. More details in the blog post on [modular packages in AWS SDK for JavaScript](https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/)
- I would change the s3 config to use credentials from the @aws-sdk/credential-providers package instead of passing `accessKeyId` and `secretAccessKey` directly. This allows to use more credential providers like the `fromProcess` or `fromContainerMetadata` provider. More details in the [AWS SDK for JavaScript v3 Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html)

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
```diff
@@ -157,113 +157,114 @@ export function configureS3AssetStorage(s3Config: S3Config) {
...
-                {
-                    Bucket: this.s3Config.bucket,
-                    Key: fileName,
-                    Body: data,
-                },
-                this.s3Config.nativeS3UploadConfiguration,
-            )
-            .promise();
-        return result.Key;
+        const upload = new (await import('@aws-sdk/lib-storage')).Upload({
```
michaelbromley: Do you think it makes sense to also import the `Upload` class in the `init()` method so it does not need to be done every time these methods are called?

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
```diff
@@ -157,113 +157,114 @@ export function configureS3AssetStorage(s3Config: S3Config) {
...
+
+            return result.Key;
+        });
     }
 
     async readFileToBuffer(identifier: string): Promise<Buffer> {
-        const result = await this.s3.getObject(this.getObjectParams(identifier)).promise();
-        const body = result.Body;
+        const result = await this.s3.getObject(this.getObjectParams(identifier));
+        const body = result.Body as Readable | undefined;
```
michaelbromley: Why is the type assertion needed here? From [the s3 client docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/getobjectcommandoutput.html#body) I see it returns type `SdkStream<undefined | Readable | Blob | ReadableStream<any>>` Is no special handling needed for Blob?

asonnleitner: Because by default it also returns the types for the browser API types which are Blob and ReadableStream

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
asonnleitner: Did not think about it 😅 will change

asonnleitner: What about the other proposals I mentioned? Should I implement them?

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
michaelbromley: Ah ok, makes sense 👍

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
michaelbromley: Will answer in a comment to the main PR

michaelbromley:
> Switch from the aws sdk v2 compatible style to use the modular v3 sdk. 

Is this purely internal to the plugin? Would it then require any further changes for consumers of the plugin? 

> I would change the s3 config to use credentials from the @aws-sdk/credential-providers package instead of passing accessKeyId and secretAccessKey directly. 

Is it possible to support both? For example, I use this plugin with MinIO which is S3-compatible, but I'm not sure if the credentials stuff also works with it.

asonnleitner:
> > Switch from the aws sdk v2 compatible style to use the modular v3 sdk.
> 
> Is this purely internal to the plugin? Would it then require any further changes for consumers of the plugin?
> 
> > I would change the s3 config to use credentials from the @aws-sdk/credential-providers package instead of passing accessKeyId and secretAccessKey directly.
> 
> Is it possible to support both? For example, I use this plugin with MinIO which is S3-compatible, but I'm not sure if the credentials stuff also works with it.

About the v3 API, it is actually just internal to the plugin, I updated the PR with the changes.
For the credentials, it should be possible to use the `AwsCredentialIdentity` which should be compatible with MinIO, and `AwsCredentialIdentityProvider` which is for the aws credentials providers.

But I am still not able to test it myself because of this `Error: Cannot find module '@ardatan/aggregate-error'` error I am getting when starting the dev server.

michaelbromley:
> Error: Cannot find module '@ardatan/aggregate-error'

Try `rm -rf node_modules && rm yarn.lock && yarn` to fix that. I get it too.

michaelbromley:
Thank for your contribution 🥳

PR #2103, opened 2023-03-28
fix(asset-server-plugin): Change image format with no other transforms

### LINKED ISSUE #2098: Asset Server not respecting format query parameter without resize
**Describe the bug**
The asset server only respects the format parameter if w, h, or preset is supplied as a query parameter.

**To Reproduce**
Go to url of asset with a format different than native image. e.g. ?format=avif
See that it doesn't use correct format.
Adding a preset does change the format. e.g. ?format=avif&preset=medium

**Expected behavior**
The asset server should respect the requested image format, even if it isn't transforming the image otherwise

 - Nodejs version: 16.19.1
 - Database (mysql/postgres etc): postgres 14

jacobfrantz1:
Fixes #2098


michaelbromley:
Hi! Thanks for this :)

Could you make the PR to `master`, as it's not so much a new feature but rather fixing something that should have already worked that way.

jacobfrantz1:
Changing the base branch didn't work, so I'm going to recreate pull request

Issue #2044, opened 2023-02-21
EmailPlugin: Allow dynamic Email templates


martijnvdbrug:
**Is your feature request related to a problem? Please describe.**
Similar to #2043, I would like administrators to edit/create templates from the Admin UI.

**Describe the solution you'd like**
It would be great to allow async fetching of templates. Something like: 
```ts
    EmailPlugin.init({
      ...,
      templates: async (ctx, type) => {
        if (type === 'order-confirmation') {
          // Get, create or fetch my order-confirmation template from wherever I want
        }
        return {
          templateString: `<html>My handlebars template </html>`
        }
    })
```

This way we can get templates from the database based on the request context and event type, or even from external platforms (Some WYSIWYG handelbars SaaS? Just ideas...)

**Describe alternatives you've considered**
There currently is no way of doing this I think.

I would like to hear your thoughts on this. Is this something that should be included in Vendure, or is this a use case for a separate plugin?

## Additional context
This would also remove the need for `addTemplate()`, since the strategy pattern above also allows for selecting templates based on language and/or channels (and more). That would make it a breaking change though..

martijnvdbrug:
For future reference @is0utfitters

PR #2112, opened 2023-04-03
Feat/email plugin dynamic templates

dalyathan:
This PR is a continuation of a [previous](https://github.com/vendure-ecommerce/vendure/pull/2059) PR which implemented issues #2043 and #2044 involving having different SMTP settings per channel and dynamic email templates for the EmailPlugin. This specific PR builds on that change by updating EmailPlugin's docs to reflect the changes.


packages/email-plugin/src/handlebars-mjml-generator.ts
```diff
@@ -16,9 +21,11 @@ import { EmailPluginDevModeOptions, EmailPluginOptions } from './types';
...
  */
 export class HandlebarsMjmlGenerator implements EmailGenerator {
-    onInit(options: EmailPluginOptions | EmailPluginDevModeOptions) {
-        const partialsPath = path.join(options.templatePath, 'partials');
-        this.registerPartials(partialsPath);
+    async onInit(options: InitializedEmailPluginOptions) {
+        if (options.templateLoader.loadPartials) {
+            const partials = await options.templateLoader.loadPartials();
+            partials.forEach(({ name, content }) => Handlebars.registerPartial(name, content));
+        }
```
martijnvdbrug, line 28: @michaelbromley Just for clarification: The templatePath is still supported, but it's loaded as the default `FileBasedTemplateLoader` and moved to here https://github.com/vendure-ecommerce/vendure/pull/2112/files#diff-27105f8aa6052576d57a9c86ffd3f2953d970c4690e2b478c9b327775cee2c47R320

packages/email-plugin/src/template-loader.ts
```diff
@@ -1,20 +1,32 @@
...
+
+    async loadPartials(): Promise<Partial[]> {
+        const partialsPath = path.join(this.templatePath, 'partials');
+        const partialsFiles = await fs.readdir(partialsPath);
+        return Promise.all(partialsFiles.map(async (file) => {
+            return {
+                name: path.basename(file, '.hbs'),
+                content: await fs.readFile(path.join(partialsPath, file), 'utf-8')
+            }
+        }));
```
martijnvdbrug, line 30: This is the default template loader, based on files, i.e. based on how it currently works in 1.9

martijnvdbrug:
Thanks for your help on this @dalyathan ! Looks good

michaelbromley:
Hi, I'm slightly confused about this PR - it is described as "updated docs" but appears to be the full feature implementation. I guess I'm missing a bit of background on this - I know @martijnvdbrug had a PR that was closed for this - is this the continuation of that work?

martijnvdbrug:
Oops yes i skipped the description. Let me update that

packages/email-plugin/src/plugin.ts
```diff
@@ -80,6 +85,26 @@ import {
  *   ],
  * };
  * ```
+ * ### Channel aware SMTP settings
+ * If you would like to have different SMTP settings per channel, then you can configure the `transport` field to fetch the values.
+ * ```
+ * EmailPlugin.init({
+ *  transport: async (ctx) => {
+ *      // Get your SMTP settings based on the requestContext
+ *      const settings = await myService.getSmtpSettings(ctx);
```
michaelbromley: Is this example needed given the "Dynamic transport settings" section below? Maybe we can add this particular use-case to that section.

packages/email-plugin/src/plugin.ts
```diff
@@ -178,6 +211,31 @@ import {
...
+ * const config: VendureConfig = {
+ *   plugins: [
+ *     EmailPlugin.init({
+ *       handlers: defaultEmailHandlers,
+ *       templatePath: path.join(__dirname, 'static/email/templates'),
+ *       transport: (injector, ctx) => {
+ *         if (ctx) {
+ *           return injector.get(MyTransportService).getSettings(ctx);
+ *         } else {
+ *           return defaultSettings;
```
michaelbromley: People might get confused about this `defaultSettings` var that is not defined anywhere. Maybe it would be clearer to do something like:

```ts
return {
  type: 'smtp',
  host: 'smtp.example.com',
  // ... etc.
}
```

Also it add an import statement for `MyTransportService` at the top just to make clear it is coming from a local file.

packages/email-plugin/src/plugin.ts
```diff
@@ -91,6 +116,14 @@ import {
  * `node_modules/\@vendure/email-plugin/templates` to a location of your choice, and then point the `templatePath` config
  * property at that directory.
  *
+ * * ### Dynamic Email Templates
+ * Instead of passing a static value to `templatePath`, use `templateLoader` to define a template path.
+ * ```ts
+ *   EmailPlugin.init({
+ *    ...,
+ *    templateLoader: new FileBasedTemplateLoader(my/order-confirmation/template)
```
michaelbromley: The path should be a string, and also `my/order-confirmation/template` is a bit confusing because as far as I understand, this path should be the directory that houses _all_ the templates, rather than just 1 type as suggested by "order-confirmation". Am I understanding this correctly?

packages/email-plugin/src/plugin.ts
```diff
@@ -247,25 +305,35 @@ export class EmailPlugin implements OnApplicationBootstrap, OnApplicationShutdow
...
     static init(options: EmailPluginOptions | EmailPluginDevModeOptions): Type<EmailPlugin> {
-        this.options = options;
+        if (options.templateLoader) {
+            Logger.info(`Using custom template loader '${options.templateLoader.constructor.name}'`);
+        } else if (!options.templateLoader && options.templatePath) {
+            // TODO: this else-if can be removed when deprecated templatePath is removed, 
+            //       because we will either have a custom template loader, or the default loader with a default path
+            options.templateLoader = new FileBasedTemplateLoader(options.templatePath);
+        } else {
+            options.templateLoader = new FileBasedTemplateLoader('./vendure/static/email/templates');
```
michaelbromley: I am guessing you might have used this path as the default because the `templatePath` docs state "In a default Vendure installation,
the templates are installed to `<project root>/vendure/email/templates`."

These docs are wrong :D 

A project created with `@vendure/create` will actually use `path.join(__dirname, '../static/email/templates')`, where `__dirname` will point to the directory of the `vendure-config.ts` file.

I think we should perhaps throw an exception here, "You must either supply a `templatePath` or provide a custom `templateLoader`", otherwise I feel like whatever default we use will be potentially surprising. Better to force the dev to be explicit about this.

Related: in the `@vendure/create` [config template file](https://github.com/vendure-ecommerce/vendure/blob/master/packages/create/templates/vendure-config.hbs#L88) we should now default to adding:

```ts
templateLoader: new FileBasedTemplateLoader(path.join(__dirname, '../static/email/templates'))
```

rather than using the `templatePath` since it is now deprecated.

packages/email-plugin/src/types.ts
```diff
@@ -42,13 +42,21 @@ export interface EmailPluginOptions {
...
+     * 
+     * @deprecated Use `templateLoader` to define a template path: `templateLoader: new FileBasedTemplateLoader('../your-path/templates')`
      */
-    templatePath: string;
+    templatePath?: string;
+    /**
+     * @description
+     * An optional TemplateLoader which can be used to load templates from a custom location or async service.
+     * The default uses the FileBasedTemplateLoader which loads templates from `<project root>/vendure/email/templates`
+     */
```
michaelbromley, line 55: add
```
@since 2.0.0 
```

michaelbromley (changes requested):
Wow, this is a really, really great PR! Thank you both for the work you did on this.

I have added a few comments to things that I think need changing - most just docs stuff.

michaelbromley:
Thank you so much @dalyathan & @martijnvdbrug for this! It is getting released today :)

dalyathan:
Happy to have participated.

PR #2116, opened 2023-04-10
feat(core): allow specifying transaction isolation level

domsj:
The default isolation level in some databases (see e.g. for postgres https://www.postgresql.org/docs/current/transaction-iso.html - `READ COMMITTED` is default) allows several anomalies, leading to incorrect data. https://en.wikipedia.org/wiki/Isolation_%28database_systems%29 for more context (or do some googling).

I think it would make sense to at least easily allow users to opt in to stronger isolation levels, and preferably do so by default (picking a lower isolation level should only be done when really needed and when you understand what you're doing).

This PR aims to achieve the above. Note that I'm not a seasoned typescript developer, and I had some issues with `yarn` in this repo, so I had to bypass the pre-commit hook.

### DIFF
--- a/packages/core/src/api/decorators/transaction.decorator.ts
+++ b/packages/core/src/api/decorators/transaction.decorator.ts
@@ -34,11 +34,13 @@ export type TransactionMode = 'auto' | 'manual';
 export const TRANSACTION_ISOLATION_LEVEL_METADATA_KEY = '__transaction_isolation_level__';
 /**
  * @description
- * Transactions can be run at different isolation levels. The default is "SERIALIZABLE", which
- * is the only correct mode. See the documentation of your database for more information on available
- * isolation levels.
+ * Transactions can be run at different isolation levels. The default is undefined, which
+ * falls back to the default of your database. See the documentation of your database for more
+ * information on available isolation levels.
  * 
- * @default 'SERIALIZABLE'
+ * @default undefined
+ * @docsCategory request
+ * @docsPage Transaction Decorator
  */
 export type TransactionIsolationLevel = 'READ UNCOMMITTED' | 'READ COMMITTED' | 'REPEATABLE READ' | 'SERIALIZABLE';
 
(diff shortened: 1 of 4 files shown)


domsj:
> Hi,
> 
> thanks for the PR! I'm in favour of allowing control over this, but I am hesitant to change the default right now. As far as I understand, the tradeoff with the serializable level is that it is more expensive and can also have an effect on the number of deadlocks encountered. I'd rather have this as opt-in at least to start with until we can gather some real-world experience of any side-effects this change might have.

I see, I understand your hesitation. And I see the failing tests (specifically related to this).

I would expect the performance tradeoff to be okay, but the potential deadlocks are a bit annoying (even if after some time the database will detect them). I think the deadlocks can be avoided depending on how the sql is written, but that might work a bit against typeorm, and would require some changes here and there, so not something to just change quickly.

> 
> By the way, did you run into any anomalies due to the isolation level when working with Vendure so far?

I ran into an anomaly when adding some custom code. A test that verified concurrent behavior of the update noticed that 2 threads were able to successfully update some object while only 1 should've been able to.
So no Vendure specific faulty code to point to (at least not at this time).
I would prefer not needing to think about anomalies though :cold_sweat:, hence my preference stated above.

I updated the PR

packages/core/src/api/decorators/transaction.decorator.ts
```diff
@@ -31,6 +31,17 @@ export const TRANSACTION_MODE_METADATA_KEY = '__transaction_mode__';
...
 export type TransactionMode = 'auto' | 'manual';
 
+export const TRANSACTION_ISOLATION_LEVEL_METADATA_KEY = '__transaction_isolation_level__';
+/**
+ * @description
+ * Transactions can be run at different isolation levels. The default is "SERIALIZABLE", which
+ * is the only correct mode. See the documentation of your database for more information on available
+ * isolation levels.
+ * 
+ * @default 'SERIALIZABLE'
```
michaelbromley: Can you update the description to say that it defaults `undefined` which will fall back to the default of your database. Also add the following tags.

```
 * @docsCategory request
 * @docsPage Transaction Decorator
```

so that it appears in the documentation website on the correct page

michaelbromley (changes requested):
Thanks for the update! Just a small adjustment to the docs needed then we're good to go 👍

michaelbromley:
Thank you for your contribution!

PR #2130, opened 2023-04-19
search field added on administrators list on dashboard -> administrator.

Mizan3050:
search field added on administrators list for followeing fields.

- First name
- Last name
- Email


michaelbromley:
Thank you!

PR #2164, opened 2023-05-05
fix(dev-server): remove additional /package from admin-ui dependency

JMPJNS:
`yarn` command fails since it can't find the /package directory, removing that fixes it


chrislais:
I think we should not remove it, you should manually create a "package " folder there.

michaelbromley:
This is outdated now - I ended up doing this myself as I ran into the same issue. Thanks for the heads-up 👍

Issue #1828, opened 2022-10-07
Fulfillments: Setting tracking code after creation


etienneroudeix:
**Is your feature request related to a problem? Please describe.**
When an administrator wants to ship an order, it first creates the fulfillment, then actually sends it and needs to update the fulfillment with the tracking code before to set it as shipped.

**Describe the solution you'd like**
From an existing fulfillment, it should be possible to update tracking code.

michaelbromley:
Yep I agree this would be useful for supporting different workflows.

denz93:
Hello @michaelbromley,

I have come up with a solution, but there are little confusion whether or not to include some `generated-types*.ts` into my commit because I added some types into `*.graphql`. Just making sure before PR.


![fulfill-update-dialog](https://user-images.githubusercontent.com/128194978/236316725-1db7badb-7590-438a-87ac-b892e6b69488.png)
![fulfill-update-detail](https://user-images.githubusercontent.com/128194978/236316729-9afc71e2-49ac-4746-ab24-94b53dd0fd90.png)
![fulfill-update-shipped](https://user-images.githubusercontent.com/128194978/236316718-d0208b23-ef9c-40e4-81dd-033886c0b5f7.png)


![confusion-edit](https://user-images.githubusercontent.com/128194978/236318235-8218887b-8282-43e2-b728-d6cbc9bdfbef.png)

dlhck:
This feature request has been moved to GitHub Discussions for community voting.
Please upvote the discussion if you would like to see this implemented: https://github.com/vendurehq/vendure/discussions/4301

PR #2170, opened 2023-05-08
feat(core,admin-ui): Allow update tracking code while fulfillment pen…

denz93:
feat(core,admin-ui): Allow update tracking code while fulfillment pending

Related to #1828. This commit add a dialog for updating tracking code and expose an API


michaelbromley:
Hi! Thanks for working on this. I have some feedback on the API that I would suggest:

Rather than `updatePendingFulfillment`, I think it makes more sense to just have `updateFulfillment` because this leaves open the possibility to update fulfillments in different states than `Pending`. Since the fulfillment process is configurable, it may well be that a process allows updating in other states. And in this case, the error result would be `FulfillmentStateError`, which is more general and also matches the existing error results like `OrderStateError`, `RefundOrderStateError` etc.

michaelbromley:
I'm closing this as it is stale. If you (or someone else) wants to take up this task again, open a new PR. Thanks for taking the time to contribute.

PR #2173, opened 2023-05-10
fix(core): Trigger account registration event in any case

stefanfeldner:
Currently, the AccountRegistrationEvent is only published when a password for the customer is set.
It should trigger the Event in both cases tho, so it should fire if a password is set, and when no password is set because in both cases, a new account is registered.


michaelbromley:
I think the test failures [here](https://github.com/vendure-ecommerce/vendure/actions/runs/5465330432/jobs/10088843814?pr=2173#step:8:114) are due to the fact that in that test suite, the `sendEmailFn` mock function only gets defined during the `beforeAll()` function, so during the data-population step where we create customers, that `TestEmailPlugin` is trying to call the mock function before it is defined.

This could be remedied by making the function calls on lines 64-73 conditional like `sendEmailFn?.(event)`

michaelbromley:
Closed by #2316 - there was one more failing e2e test so I just fixed it and merged in that PR.

Thanks for your help on this!

PR #2186, opened 2023-05-20
fix(core): Missing closing parenthesis

ghost:
* fix(core): Missing closing parenthesis

* fix(core): Ensures that the value of args.ids is not undefined before calling the map method


michaelbromley:
Thanks for pointing this out! I happened to discover it myself when I tried to build the branch :D

PR #2192, opened 2023-05-24
fix(payments-plugin): Removed invalid error logging

martijnvdbrug:
* Removed error logging when Mollie sends a webhook with status `expired`, because this is not considered an error. It's fine if a Mollie session expires and a user tries to pay again later.
* Some linting fixes that occurred when trying to commit

### DIFF
--- a/packages/payments-plugin/src/mollie/mollie.service.ts
+++ b/packages/payments-plugin/src/mollie/mollie.service.ts
@@ -83,9 +83,15 @@ export class MollieService {
         if (!order) {
             return new PaymentIntentError('No active order found for session');
         }
-        await this.entityHydrator.hydrate(ctx, order,
-            { relations: ['customer', 'surcharges', 'lines.productVariant', 'shippingLines.shippingMethod', 'payments'] }
-        );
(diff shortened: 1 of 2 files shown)


packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -195,6 +227,10 @@ export class MollieService {
                 `Unable to find order ${mollieOrder.orderNumber}, unable to process Mollie order ${mollieOrder.id}`,
             );
         }
+        if (mollieOrder.status === OrderStatus.expired) {
+            // Expired is fine, a customer can retry the payment later
+            return;
+        }
```
martijnvdbrug, line 233: This is the actual fix that prevents unneeded error logs

martijnvdbrug:
Again not sure what is causing these e2e test fails? Flakiness? Already broken on major? Just some formatting changes and a single `if` statement have been added.

michaelbromley:
Thanks!
Yeah the tests are not in perfect shape on the major branch right now. Just working on this at the moment.

PR #2203, opened 2023-05-31
refactor: update generateToken method to use secure random number generation

asonnleitner:
This commit introduces a change to the `generateToken` method, improving the security of the token generation process. The previous implementation relied on `Math.random()`, which is not ideal for cryptographic purposes. The updated implementation now utilizes the `crypto.getRandomValues` function to generate an array of random bytes. These bytes are then converted into a standardized hexadecimal string representation, resulting in more secure and reliable tokens.

### DIFF
--- a/packages/admin-ui/src/lib/settings/src/components/channel-detail/channel-detail.component.ts
+++ b/packages/admin-ui/src/lib/settings/src/components/channel-detail/channel-detail.component.ts
@@ -226,7 +226,6 @@ export class ChannelDetailComponent
     }
 
     private generateToken(): string {
-        const randomString = () => Math.random().toString(36).substr(3, 10);
-        return `${randomString()}${randomString()}`;
+        return Array.from(crypto.getRandomValues(new Uint8Array(10))).map(b => b.toString(16).padStart(2, '0')).join('');
     }
 }


michaelbromley:
As a reference, here are some strings generated by the current method:

```
'12xhiz6zn8uriv08tqdx'
'0yf8o9470iqm9foubfaa'
'kzp1bb4vd50bubkggfl3'
```

and here are some by your proposed method:

```
'f90c00037226fa3b1ddb'
'9f767a4a956ebfc4be1c'
'fbb0711d4992b349d32b'
```

I think yours look nicer just because we are limiting to the 16 hex chars so it somehow looks more consistent.

asonnleitner:
> As a reference, here are some strings generated by the current method:
> 
> 
> 
> ```
> 
> '12xhiz6zn8uriv08tqdx'
> 
> '0yf8o9470iqm9foubfaa'
> 
> 'kzp1bb4vd50bubkggfl3'
> 
> ```
> 
> 
> 
> and here are some by your proposed method:
> 
> 
> 
> ```
> 
> 'f90c00037226fa3b1ddb'
> 
> '9f767a4a956ebfc4be1c'
> 
> 'fbb0711d4992b349d32b'
> 
> ```
> 
> 
> 
> I think yours look nicer just because we are limiting to the 16 hex chars so it somehow looks more consistent.

Yes, that's what I was going for! 😄

PR #2219, opened 2023-06-09
fix(docs): Change Uploading Files example files.map to Array.from

hialves:
I made this change because i was receiving an error based in the documentation for uploading files, files.map is undefined in the context of FileList and according to [FileList Documentation](https://developer.mozilla.org/pt-BR/docs/Web/API/FileList) we should use a for-loop to populate the sent array.

### DIFF
--- a/docs/content/developer-guide/uploading-files.md
+++ b/docs/content/developer-guide/uploading-files.md
@@ -42,15 +42,10 @@ function UploadFile() {
 
   function onChange(event) {
     const { target } = event;  
-    
     if (target.validity.valid) {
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,
Thanks for this fix! I would suggest maybe a nicer way to get the FileList items into an array could be to use the `Array.from()` method. What would you think about it like this?

```ts
input: Array.from(target.files).map(file => ({ file }))
```

michaelbromley:
Thanks for your contribution! 👍

PR #2232, opened 2023-06-14
fix(admin-ui): Update fr translation

jbanety:


michaelbromley:
Thank you so much! I was planning on trying ChatGPT for all the missing translations but a human is much better 😄

jbanety:
Haha.
> Pas de soucis !

PR #2239, opened 2023-06-20
fix for the facet resolution performance issue

goroka:
instead of one potentially large query, we first load the facets, then filter them for access to the appropriate channel.


michaelbromley:
Thank you for this contribution! 🙌

PR #2241, opened 2023-06-21
fix(core): Delete user sessions & token upon user-deletion

DanielBiegler:
This applies to both admins and regular users.

### DIFF
--- a/packages/core/src/service/services/user.service.ts
+++ b/packages/core/src/service/services/user.service.ts
@@ -1,4 +1,5 @@
 import { Injectable } from '@nestjs/common';
+import { ModuleRef } from '@nestjs/core';
 import { VerifyCustomerAccountResult } from '@vendure/common/lib/generated-shop-types';
 import { ID } from '@vendure/common/lib/shared-types';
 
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/user.service.ts
```diff
@@ -170,6 +172,10 @@ export class UserService {
     }
 
     async softDelete(ctx: RequestContext, userId: ID) {
+        // Dynamic import to avoid the circular dependency of SessionService
+        await this.moduleRef
+            .get((await import('./session.service.js')).SessionService)
+            .deleteSessionsByUser(ctx, new User({ id: userId }));
```
DanielBiegler, line 178: @michaelbromley For now I just constructed the User to be able to use the `deleteSessionsByUser` function - it might be cleaner to refactor? The deletion function only uses the userId anyway. Whats your more experienced-perspective on this? Cheers.

michaelbromley: Yes I agree it makes more sense for that `deleteSessionsByUser` to just take an ID rather than the User object. But since that's a public method, I would err on the side of caution and not change the signature in a breaking way. We _could_ overload it so it can take a User object _or_ an ID though, which would be a backward-compatible change.

PR #2243, opened 2023-06-22
feat(email-plugin): Expose template vars to template loader

### LINKED ISSUE #2242: Forward template variables to the template loader
We would like to customise template loading based on order data. For example, a different email for promotions.

Forwarding the `templateVars` to the `TemplateLoader` would make this possible.

No alternatives possible with the current API. We could use the `EmailGenerator` to override the already read `body`, but that is currently a synchronous function and hence that would have to change too.

**Additional context**
I'll open a PR for this.

StampixSMO:
Closes #2242 

Let me know if you'd like to pass more of the `data` object or if you somehow would hard-type the `templateVars`.


michaelbromley:
I was wondering whether it makes sense to pass the entire `data` object, but on reflection I think it is a safer approach to expose just the templateVars. In any case, the developer is able to put any arbitrary data in templateVars anyway, so that should be sufficient.

PR #2245, opened 2023-06-23
request to add 'where OR'

KushagraVasal:
Can you add OR operator
at this.listQueryBuilder.build(


michaelbromley:
Hi,

is the a feature request or a pull request? I don't fully understand what is meant here. Can you supply a detailed description?

KushagraVasal:
I just need sql 'OR' operator

https://stackoverflow.com/questions/53407236/typeorm-or-operator

you can look for the term `Querying with OR operator:` on
https://orkhan.gitbook.io/typeorm/docs/find-options

michaelbromley:
OK. I'm gonna close this then because it's more of a feature request. Feel free to open an issue for this, but please, you'll need to be much more specific that "I just need sql 'OR' operator".

- state the problem you are trying to solve
- state the exact GraphQL query or method call you would like to be able to make but currently cannot.

Without very well-defined specifics, I cannot really act on the request because I don't understand it well enough.

Issue #2257, opened 2023-06-30
Fix order list table query


cdiban:
**Describe the bug**
The search bar in order list component only search by code. However, the search bar indicates that you can search by name (lastname), code or transaction id.

**To Reproduce**
1. Go to Orders
2. Click on search orders by customer last name
4. You can't see results.

**Expected behavior**
Get order list by customer, code or transaction id.

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.2
 - Nodejs version: 16
 - Database (mysql/postgres etc): postgresql

PR #2258, opened 2023-06-30
fix(admin-ui): Fix query filter option in order list component

cdiban:
Relates to #2257. This commit add customerLastName and id filters


michaelbromley:
Thanks for the contribution! The original functionality (as in v1.x) was to filter by _transaction id_ rather than the _order id_ - the transaction ID being the id generated by the payment gateway. 

There is a filter input named `transactionId` for this purpose.

Could you update to use that filter rather than `id`?

cdiban:
> Thanks for the contribution! The original functionality (as in v1.x) was to filter by _transaction id_ rather than the _order id_ - the transaction ID being the id generated by the payment gateway.
> 
> There is a filter input named `transactionId` for this purpose.
> 
> Could you update to use that filter rather than `id`?

Of course, thank you for the clarification. I pushed a new commit.

Best.
Carlo.

michaelbromley:
Thank you! This will be part of the upcoming 2.0.3 patch release 👍

PR #2259, opened 2023-07-01
fix(admin-ui): Fix quantity label in order table

cdiban:
Fix quantity label in order table. 🙂


michaelbromley:
Oh good catch, thank you! :)

PR #2263, opened 2023-07-03
chore(deps): Update i18next

asonnleitner:


michaelbromley:
Thanks for the contribution! From the failing e2e tests, it looks like the variables are getting interpolated into error messages though...

asonnleitner:
> Thanks for the contribution! From the failing e2e tests, it looks like the variables are getting interpolated into error messages though...

Ah, thanks for pointing that out, let me fix that

michaelbromley:
Did the default method of interpolation change between v19 and v23? I checked the [migration guide](https://www.i18next.com/misc/migration-guide) and did not immediately notice anything about that.

If `{{` is now the standard way to do it, then yes I think it makes sense to update to that.

asonnleitner:
My bad, you are right. There was no change in the default interpolation prefix and suffix. I was confused by the example in the docs, which uses `{{` and `}}`.
I looked into `i18next-icu`, and the ICU message format changed.

> ### Quoting/Escaping
> If syntax characters occur in the text portions, then they need to be quoted by enclosing the syntax in pairs of ASCII apostrophes. A pair of ASCII apostrophes always represents one ASCII apostrophe, similar to `%%` in `printf` representing one `%`, although this rule still applies inside quoted text. (“This '{isn''t}' obvious” → “This {isn't} obvious”)

> Before ICU 4.8, ASCII apostrophes always started quoted text and had inconsistent behavior in nested sub-messages, which was a source of problems with authoring and translating message pattern strings.
> Starting with ICU 4.8, an ASCII apostrophe only starts quoted text if it immediately precedes a character that requires quoting (that is, “only where needed”), and works the same in nested messages as on the top level of the pattern. The new behavior is otherwise compatible; for details see the MessageFormat and MessagePattern (new in ICU 4.8) API docs.
> Recommendation: Use the real apostrophe (single quote) character ’ (U+2019) for human-readable text, and use the ASCII apostrophe `'` (U+0027) only in program syntax, like quoting in MessageFormat. See the annotations for U+0027 Apostrophe in The Unicode Standard.

So I changed the quotes to double quotes, and it works now. Additionally, there was also a change with the `{select}` syntax, which now requires an `other` case.

> **Note**: In a plural argument like in the example above, if the English message has both `=0` and `=1` (up to `=offset`+1) then it does not need a “`one`” variant because that would never be selected. It does always need an “`other`” variant.



For more details, you can check the [ICU User Guide on Quoting/Escaping](https://unicode-org.github.io/icu/userguide/format_parse/messages/#quotingescaping). Additionally, you can refer to the [ICU User Guide on Complex Argument Types](https://unicode-org.github.io/icu/userguide/format_parse/messages/#complex-argument-types) for more information on the required other case.

michaelbromley:
Thanks for the detailed explanation, and the extra work updating all the i18n strings 👍

asonnleitner:
Thanks for merging my PR! Looking forward to making many more contributions 🎉

PR #2271, opened 2023-07-06
fix(admin-ui): Add product variant name filter in product variant list

cdiban:
Since the v2.0.3 product variants list can't filter by name. In my opinion, a name filter is essential in business with thousands of product variants in each product.


cdiban:
@michaelbromley What do you thing about this filter?

Issue #2278, opened 2023-07-11
Populate mock data not working with sqlite


wpplumber:
**Describe the bug**
Setup the server to contribute to Vendure encountered a blocker while trying to populate the data.

**Environements:**
Mac OS Catalina

**To Reproduce**
1. cd packages/dev-server
2. DB=sqlite yarn populate
3. See error

```
DB=sqlite yarn populate
yarn run v1.22.19
$ node -r ts-node/register -r dotenv/config populate-dev-server.ts
Using sqlite connection
Clearing all tables...
Done!
info 7/12/23, 12:14 AM - [Vendure Server] Bootstrapping Vendure Server (pid: 50404)... 
info 7/12/23, 12:14 AM - [NestFactory] Starting Nest application... 
info 7/12/23, 12:14 AM - [InstanceLoader] ProcessContextModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] ServiceModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] CacheModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] TypeOrmModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] PluginModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] TypeOrmModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] PluginCommonModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] PluginModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] ConfigModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] BullMQJobQueuePlugin dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] I18nModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] AssetServerPlugin dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] TerminusModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] JobQueueModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] AppModule dependencies initialized 
info 7/12/23, 12:14 AM - [InstanceLoader] HealthCheckModule dependencies initialized 
[...]

PR #2279, opened 2023-07-12
fix(README): Redis must be installed to populate mock data

wpplumber:
See the [bug](https://github.com/vendure-ecommerce/vendure/issues/2278) I opened.


DanielBiegler (changes requested):
This is only relevant if your setup actually uses redis, you can just use the `DefaultJobQueuePlugin.init({})` and be done with it. Especially for local testing.

Then just including Mac instructions for a cross platform project is out of place.

If people have a hard time figuring out how to run the dev env locally you should probably change the `dev-config.ts` file and include a link to [documentation](https://docs.vendure.io/guides/developer-guide/worker-job-queue/#jobqueuestrategy) instead of adding insufficient install instructions

PR #2286, opened 2023-07-14
fix(admin-ui): Add custom detail component locationId and action bar items to draft order detail

cdiban:
Now we can add components and action bar items to complement draft orders.


michaelbromley:
Thank you for this improvement!

PR #2287, opened 2023-07-15
fix(admin-ui): all Arabic tokens translated (100%)

wpplumber:
> @tarik I created the branch admin-ui-rtl (https://github.com/vendure-ecommerce/vendure/tree/admin-ui-rtl) which has the basic parts wired up, and the 2 main tasks that would need to be done are:
> 
> Add all the AR translation strings
> Update CSS to display property in RTL. The main task would be to replace instances of margin-left and margin-right etc with the RTL-safe equivalents like margin-inline-start etc. This is covered well here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values

 ⬆️  by @michaelbromley

Part one done:
- ~~Add all the AR translation strings~~


wpplumber:
Why getting the following issue?

image

michaelbromley:
I failed to add an `isRTL()` method to the mock i18n service. 

it should be sufficient to add
```ts
isRTL(): boolean {
  return false;
}
```

to the class in i18n.service.mock.ts file.

wpplumber:
Thank you, Michael.

wpplumber:
I see the previous failure passed but now other one with mysql!

michaelbromley:
Don't worry - that failing e2e test suite is not related to your changes. Sometimes the e2e tests are a bit flaky because there are so many moving parts.

Do you also plan to incorporate the CSS changes in this PR or are you planning that in a separate PR?

wpplumber:
I prefer in other PR which makes it simple as possible.

I went through reading the [clarity.design foundation](https://core.clarity.design/foundation/app-layout/api/) but to avoid more time learning I plan to overwrite css styles which makes better UX i.e font size, margin between icon and text ...

michaelbromley:
OK, thanks so much for your work so far!

wpplumber:
Thank you too for helping in getting my first contribution.

Issue #2293, opened 2023-07-17
Vendure 2.0.1: The surcharge `price` and `priceWithTax` getters are not rounded


skid:
The surcharge entity can return a decimal number for the `price` and `priceWithTax` properties with the DefaultMoneyStrategy active. This happens because the `roundMoney` util is not used.

To Reproduce, just use a Surcharge and set tax rate that will guarantee a decimal number, then check those properties - one of them will most likely have decimals if your tax rate is not evenly 

 - @vendure/core version: 2.0.1 through 2.0.4

skid:
I'm opening a PR for this

PR #2294, opened 2023-07-17
fix(core): Use MoneyStrategy in Surcharges

skid:
This fixes https://github.com/vendure-ecommerce/vendure/issues/2293


michaelbromley:
Thanks for the quick fix!

PR #2307, opened 2023-07-21
feat(admin-ui): add channelTokenKey to AdminUiConfig

asonnleitner:
The `AdminUiConfig` now includes a `channelTokenKey` property. This property specifies the name of the header containing the channel token, and should match the `apiOptions.channelTokenKey` config option on the server.


michaelbromley:
Thanks! This was a key missing API 👍

PR #2308, opened 2023-07-21
Fix(payments-plugin): Only find payment methods in the current channel

### LINKED ISSUE #2290: Stripe plugin: webhook failing on multi-channel
Hi, we have a multi-channel setup, where every channel is connected to a different Stripe accounts with a dedicated payment method. 

Orders can transition to `Payment settled` only on one channel. When trying to pay with another user on another channel, the order remains stuck in `Arranging Payment`. In the Stripe dashboard it seems a problem of failing webhook, even though API keys and Webhook secrets are registered in the same way for every channel. 

On the main channel the plugin can set the payment intents without problems

`[StripePlugin] Stripe payment intent id pi_3NSxzoFltThU8k2X1wYxXkhH added to order N6Y2V94WXU7HMBKY` 

and the weebhook in the Stripe dashboard returns ok.

In the other channels however the webhook request fails, with this response.
```json
{
  "statusCode": 400,
  "message": "USER_INPUT_ERROR: error.payment-method-not-found",
  "timestamp": "2023-07-12T08:13:37.581Z",
  "path": "/payments/stripe"
}
```
Could it be because different Stripe accounts webhooks are registered to the same endpoint `https://api.myvendureinstance.com/payments/stripe` in the server? 🤔 

**To Reproduce**
1. Install vendure + payments-plugin + stripe
2. Create two channels
3. Add the ralative payment methods for each channel (with two diffrent stripe accounts on each payment handler)
4. setup the webhook on each stripe account

**Expected behavior**
we should receive payments on each channel/stripe account

 - @vendure/core version: 2.0.4
 - Nodejs version 18.16
 - Database postgres

martijnvdbrug:
Only search payment methods in the current channel if they have a stripe handler, not all payment methods of all channels.
Fixes #2290

### DIFF
--- a/packages/payments-plugin/src/stripe/stripe.controller.ts
+++ b/packages/payments-plugin/src/stripe/stripe.controller.ts
@@ -6,6 +6,7 @@ import {
     Order,
     OrderService,
     PaymentMethod,
+    PaymentMethodService,
     RequestContext,
     RequestContextService,
     TransactionalConnection,
@@ -27,6 +28,7 @@ const noPaymentIntentErrorMessage = 'No payment intent in the event payload';
 export class StripeController {
     constructor(
         private connection: TransactionalConnection,
+        private paymentMethodService: PaymentMethodService,
         private orderService: OrderService,
         private stripeService: StripeService,
         private requestContextService: RequestContextService,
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks Martijn, you are a hero!

Issue #2303, opened 2023-07-20
[V2.0.4] Should `getFulfillmentsLinesForOrderLine` will return entity `fulfillment` directly?


tianyingchun:
**Is your feature request related to a problem? Please describe.**

Should `getFulfillmentsLinesForOrderLine` will return entity `fulfillment` directly?

in `fullillmentSerivce `
```ts
    async getFulfillmentsLinesForOrderLine(ctx: RequestContext, orderLineId: ID): Promise<FulfillmentLine[]> {
        const fulfillmentLines = await this.connection
            .getRepository(ctx, FulfillmentLine)
            .createQueryBuilder('fulfillmentLine')
// this line use leftJoin without `select`
            .leftJoin('fulfillmentLine.fulfillment', 'fulfillment')
            .where('fulfillmentLine.orderLineId = :orderLineId', { orderLineId })
            .andWhere('fulfillment.state != :cancelledState', { cancelledState: 'Cancelled' })
            .getMany();

        return fulfillmentLines;
    }

```
<img width="535" alt="image" src="https://github.com/vendure-ecommerce/vendure/assets/2833251/476f038f-1f8c-4acc-8a46-19f58506c867">

in most case , we need to use this service to retrive fulfilments of `lines` like below
```ts
    const fulfillments = [];
    for (const line of order.lines) {
        const fulfillmentLines = await fulfillmentService.getFulfillmentsLinesForOrderLine(ctx, line.id);
        for (const fulfillmentLine of fulfillmentLines) {
            fulfillments.push(fulfillmentLine.fulfillment);
        }
    }
```
but `fulfillmentLine.fulfillment` return us `undefined`, we have to re invoke service to renew `fulilment`

**Describe the solution you'd like**
change `  .leftJoin('fulfillmentLine.fulfillment', 'fulfillment')` to   `.leftJoinAndSelect('fulfillmentLine.fulfillment', 'fulfillment')`
**Describe alternatives you've considered**

**Additional context**

tianyingchun:
or use `hydrate`?
```ts
await entityHydrator.hydrate(ctx, fulfillmentLine, {
          relations: ['fulfillment'],
        });
```

PR #2314, opened 2023-07-25
refactor(core): Add relations to the getFulfillmentsLinesForOrderLine method

alexisvigoureux:
Fixes https://github.com/vendure-ecommerce/vendure/issues/2303

This refactor allows us to get the fulfillmentLine entities with the relations and avoids calling `entityHydrator.hydrate` afterwards, which improves performance.


michaelbromley:
Thank you, this is a good improvement 👍

Issue #2323, opened 2023-07-27
Issue with long header empty table


BeepLoveKarki:
Why does this issue occur, only when the table is empty?
<img width="1440" alt="Screenshot 2023-07-27 at 22 18 33" src="https://github.com/vendure-ecommerce/vendure/assets/29729738/4f72251e-36d5-4a67-90e8-401cdb0861ed">

BeepLoveKarki:
yes

oidt:
when there are no results for that table the table got the layout fixed.

https://github.com/vendure-ecommerce/vendure/blob/cf20adee0415c1085c7ee8a30c24c7613ca4397d/packages/admin-ui/src/lib/catalog/src/components/collection-data-table/collection-data-table.component.html#L8

that is why.

PR #2325, opened 2023-07-28
Change the table-layout to inherit always for better design when have…

oidt:
… empty table (#2323) 
In case of a empty table for orders, collections or other data tables the table will be inherit his columns and not overwrite each other.

https://user-images.githubusercontent.com/17176500/256823830-8643b143-32a2-4be6-a25e-17c7081eccc2.png


michaelbromley:
Hi!

Thanks for the contribution. I agree with the change you are proposing, but I think that you can actually just delete the `[style.table-layout]="!items?.length ? 'fixed' : 'inherit'"` line entirely, since it should then revert to the default `auto` layout mode.

Issue #1770, opened 2022-09-07
[ElasticsearchPlugin]: Allow to configure elastic reindex chunk sizes


Izayda:
**Is your feature request related to a problem? Please describe.**
I have memory spikes and OOM while full reindex. 

**Describe the solution you'd like**
I want to able to configure reindex chunks to decrease consume of RAM.

**Describe alternatives you've considered**
No alternatives.

giosueDelgado:
There is this field theorycally: `batchSize`
But I search in the code but is not used, can be @michaelbromley ?

giosueDelgado:
I can made the change if needed

michaelbromley:
Hi @giosueDelgado, yes, if you have time I'd very much welcome a fix for this.

Min2who:
can i get assigned @michaelbromley

michaelbromley:
The PR has never been completed. I invite @monrostar to complete the PR if he has time. Otherwise I'd open it up to the community, who can build on the work he did. If someone else ends up claiming the bounty based largely on his work, I would offer to tip him on top of the bounty if he is interested (Algora has a tipping feature).

Min2who:
@michaelbromley what changes are you looking from this [pr](https://github.com/vendure-ecommerce/vendure/pull/2327) could you explain to me

michaelbromley:
At the bare minimum all tests passing. But read that entire PR thread - there is lots of info there.

giosueDelgado:
Hello, I complete the fix, I try to push today!

giosueDelgado:
Just to know in the pr there was a bug that we encountered when upgrade to vendure 2.0 I made a refactor and improve on elasticsearch plugin so in this pr I added only the fix needed on customfields.

So I think this issue can be closed with this.
I made a new version of elasticsearch plugin that we are using only on our production where customize the query, optimize some logics (we improve 5/6 times the time on top of the pr) and I open a new issue and new pr for that

giosueDelgado:
Thanks @michaelbromley 🥇

PR #2327, opened 2023-07-31
perf(Elasticsearch): optimize memory consumption during reindex.

monrostar:
Added two new variables to limit the amount of data that can be stored in memory during the reindex job

> Products chunk size for each loop iteration when indexing products.

`reindexProductsChunkSize` - default value is 2500

> Index operations are performed in bulk, with each bulk operation containing a number of individual index operations. This option sets the maximum number of operations in the memory buffer before a bulk operation is executed.

`reindexBulkOperationSizeLimit` - default value is 3000

Refs: https://github.com/vendure-ecommerce/vendure/issues/1770

BREAKING CHANGE: This update optimizes memory usage for large amounts of data and applies to products with many variants. With the help of the memory storage limit, memory consumption can be increased or decreased. The processing speed and memory consumption depends on the limits set in the new variables.
And batchSize was removed because it was a variable that was not used elsewhere

### DIFF
--- a/packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
+++ b/packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
@@ -1,18 +1,18 @@
-import { Client } from '@elastic/elasticsearch';
-import { Inject, Injectable, OnModuleDestroy, OnModuleInit } from '@nestjs/common';
+import type { Client } from '@elastic/elasticsearch';
+import type { OnModuleDestroy, OnModuleInit } from '@nestjs/common';
+import { Inject, Injectable } from '@nestjs/common';
 import { unique } from '@vendure/common/lib/unique';
 import {
     Asset,
-    asyncObservable,
-    AsyncQueue,
-    Channel,
     Collection,
-    ConfigService,
     EntityRelationPaths,
     FacetValue,
     ID,
-    InternalServerError,
     LanguageCode,
+    AsyncQueue,
+    Channel,
+    ConfigService,
+    InternalServerError,
     Logger,
     Product,
     ProductPriceApplicator,
@@ -27,9 +27,9 @@ import {
 import { Observable } from 'rxjs';
 import { In, IsNull } from 'typeorm';
 
-import { ELASTIC_SEARCH_OPTIONS, loggerCtx, VARIANT_INDEX_NAME } from '../constants';
-import { ElasticsearchOptions } from '../options';
-import {
+import { ELASTIC_SEARCH_OPTIONS, VARIANT_INDEX_NAME, loggerCtx } from '../constants';
+import type { ElasticsearchOptions } from '../options';
+import type {
     BulkOperation,
     BulkOperationDoc,
     BulkResponseBody,
@@ -46,6 +46,7 @@ import {
 
 import { createIndices, getClient, getIndexNameByAlias } from './indexing-utils';
 import { MutableRequestContext } from './mutable-request-context';
+import { asyncObservable } from "@vendure/core/src";
 
 export const defaultProductRelations: Array<EntityRelationPaths<Product>> = [
     'featuredAsset',
@@ -111,7 +112,7 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
     /**
      * Updates the search index only for the affected product.
      */
-    async updateProduct({ ctx: rawContext, productId }: UpdateProductMessageData): Promise<boolean> {
+    async updateProduct({ctx: rawContext, productId}: UpdateProductMessageData): Promise<boolean> {
         const ctx = MutableRequestContext.deserialize(rawContext);
         await this.updateProductsInternal(ctx, [productId]);
         return true;
@@ -120,12 +121,11 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
     /**
      * Updates the search index only for the affected product.
      */
-    async deleteProduct({ ctx: rawContext, productId }: UpdateProductMessageData): Promise<boolean> {
-        const operations = await this.deleteProductOperations(
+    async deleteProduct({ctx: rawContext, productId}: UpdateProductMessageData): Promise<boolean> {
+        await this.deleteProductOperations(
             RequestContext.deserialize(rawContext),
             productId,
         );
-        await this.executeBulkOperations(operations);
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


monrostar:
I think that such a loop can be added wherever a list is used to generate operations
```ts
if (operations.length >= this.options.reindexBulkOperationSizeLimit)
```

For example. Here we can have the same problem and we can fix it if we'll use limits for loops

```ts
    private async updateProductsInternal(ctx: MutableRequestContext, productIds: ID[]) {
        const operations = await this.updateProductsOperations(ctx, productIds); // potential problem with big productIds list
        await this.executeBulkOperations(operations);
    }
```

michaelbromley:
@monrostar thanks for the work on this so far! Regarding the last comment - do you plan to implement that additional optimization in this PR?

monrostar:
> @monrostar thanks for the work on this so far! Regarding the last comment - do you plan to implement that additional optimization in this PR?

Yes, I checked the generation during reindex and instead of 50 hours it took 200minutes to generate 18249972 documents. 
I will try to make this update as soon as possible

image

monrostar:
@michaelbromley Added optimization for all related operations

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
```diff
@@ -46,9 +46,7 @@ import {
 
 import { createIndices, getClient, getIndexNameByAlias } from './indexing-utils';
 import { MutableRequestContext } from './mutable-request-context';
-
-const REINDEX_CHUNK_SIZE = 2500;
-const REINDEX_OPERATION_CHUNK_SIZE = 3000;
+import { asyncObservable } from "@vendure/core/src";
```
michaelbromley: This import is causing issues due to importing from `/src`. Add the `asyncObservable` import back into the group of `@vendure/core` imports and hopefully this resolves the failing tests.

michaelbromley (changes requested):

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
monrostar: Yes, didn't notice the error, I already fixed

monrostar:
@michaelbromley  I need to fix some code, probably I missing something when coping with our code. I will fix e2e tomorrow

michaelbromley:
@monrostar Did you ever get a chance to finish this? Do you need any help?

monrostar:
Yes, I need help, I had a problem starting the project and I was never able to get back to fix this problem. I'll describe my problem in more detail later. I'm still in touch, just a lot of work

giosueDelgado:
We are having issue on reindex! If you can finish we can test ASAP @monrostar and give you a fast feedback

monrostar:
> We are having issue on reindex! If you can finish we can test ASAP @monrostar and give you a fast feedback

Ok, I will try to edit the PR within 2 days

giosueDelgado:
FYI @lorenzorottigni

giosueDelgado:
> > We are having issue on reindex! If you can finish we can test ASAP @monrostar and give you a fast feedback
> 
> Ok, I will try to edit the PR within 2 days

Cool! Thanks if you need help or some support tell us!

giosueDelgado:
Ok, no problem, I can start to check this Wednesday, I give you updates

giosueDelgado:
@monrostar Me and @lorenzorottigni try to check this today we give you updates as soon as we have!

giosueDelgado:
@monrostar I found a bug in the query you are using if there are relation on customfields.
I cannot push on your repository to make it work, in the meantime I create a copy in my local and make it works.

In my opinion the elasticsearch plugin the query to fetch the data (products and variant) can be extened in the config to allow a more performant way to manage the fetch based on the custom data of the installation what do you think? @michaelbromley

monrostar:
> > I'm not totally sure whether you can directly contribute to @monrostar's PR - maybe you can if he gives you write access to his fork.
> > > In my opinion the elasticsearch plugin the query to fetch the data (products and variant) can be extened in the config to allow a more performant way to manage the fetch
> > 
> > 
> > Maybe this can be a separate issue + PR?
> 
> Yes, in my local I just did all things, in the @monrostar pr I will just fix the query as soon as he can give to me the permission, after that I open a new issue + pr for allow custom query based on the specific needs.
> 
> I just develop all changes and also add some logging and some general improvements on the elasticsearch plugin I hope to contribute as soon as possible because it take a very important optimization and possibility to customize! :)

I can give you all access, this is not an issue if you already done this :)

monrostar:
> I'm not totally sure whether you can directly contribute to @monrostar's PR - maybe you can if he gives you write access to his fork.
> 
> > In my opinion the elasticsearch plugin the query to fetch the data (products and variant) can be extened in the config to allow a more performant way to manage the fetch
> 
> Maybe this can be a separate issue + PR?

In ES, we need to rewrite the index completely because mapping is not good for large amounts of data. It is necessary to rewrite the structure and also to split Product/ProductVariant into different indexes because they are responsible for different search levels and aggregation over such a large amount of data is a very costly procedure.

We can take a more sophisticated approach. We are working on this improvement now, because we couldn't handle an index size of 200Million documents

giosueDelgado:
Agree with you @monrostar but as you know the elasticsearch is not open source any more so I think (for our case) is better migrate to other solution opensource.
For now because the plugin was ready to use we start with elasticsearch.
If we need to create a plugin more performance I prefer to do with some opensource project what do you think?

I think we can build some plugin based on typesense, Solr, MeiliSearch (or the opensource fork of elasticsearch "OpenSearch") let me know what you think

monrostar:
> Agree with you @monrostar but as you know the elasticsearch is not open source any more so I think (for our case) is better migrate to other solution opensource. For now because the plugin was ready to use we start with elasticsearch. If we need to create a plugin more performance I prefer to do with some opensource project what do you think?
> 
> I think we can build some plugin based on typesense, Solr, MeiliSearch (or the opensource fork of elasticsearch "OpenSearch") let me know what you think

Such solutions depend on your business objectives and capabilities, so this choice does not always have to be made because it is opensource or not. We use ES in production on AWS and we have 5 million product variants, so we decided to use ES because it guarantees the performance and speed we need.

giosueDelgado:
Oh sure for this! I think on same performance guaranteed I prefer open source project, no?

giosueDelgado:
@monrostar I write to you on discord because I have a permission issue on the PR (I can clone, I made the little change but and error on push)

image


Let me know when you can give to me the permission so I can close this issue

giosueDelgado:
@michaelbromley can you check this pr? I check the changes and are very limited, the test not work also if I run from the start of the commit of this pr can you give me some suggestion?

Thanks

monrostar:
Wow, really :) Thanks a lot to our community for helping to finish this PR :) too bad I didn't have the energy and time to do it...

michaelbromley:
Because this branch was very old, I had to do quite a bit of manual merging to get it compatible with the current master branch.

The only major problem was that the `relationLoadStrategy: 'query'` which was added in 2 places of the indexing logic had to be removed. The reason is that it is buggy and breaks when used in combination with nested entity relations, which is how Vendure implements custom field relations. This is why almost all e2e tests were failing.

This has the downside of undoing some of the memory-pressure improvements, but is sadly necessary at this time until the TypeORM implementation is fixed.

Issue #2330, opened 2023-08-02
Maximum usage of a coupon


PKarics:
**Is your feature request related to a problem? Please describe.**
We would like to maximize coupon usage for a marketing campaign, where the first 100 user can use the coupon. Per customer coupon usage cannot handle that.

**Describe the solution you'd like**
Add usage limit to the promotions like per customer usage limit, but it counts the coupon usage globally.

michaelbromley:
Will be in v2.1. Thanks for the work on this!

PR #2331, opened 2023-08-02
feat(core): Add maximum coupon usage

PKarics:
This is the implementation of the feature request:
https://github.com/vendure-ecommerce/vendure/issues/2330


packages/admin-ui/src/lib/marketing/src/components/promotion-list/promotion-list.component.ts
```diff
@@ -73,11 +73,17 @@ export class PromotionListComponent
...
-                name: 'usageLimit',
+                name: 'perCustomerUsageLimit',
                 type: { kind: 'number' },
                 label: _('marketing.per-customer-limit'),
                 filterField: 'perCustomerUsageLimit',
             },
+            {
+                name: 'usageLimit',
+                type: { kind: 'number' },
+                label: _('marketing.limit'),
```
michaelbromley: The label here should match the translation token used in the detail form so that only 1 translation is created. I'd say `marketing.usage-limit` is better because it is more explicit.

packages/admin-ui/src/lib/marketing/src/components/promotion-detail/promotion-detail.component.html
```diff
@@ -92,6 +92,19 @@
                             formControlName="perCustomerUsageLimit"
                         />
                     </vdr-form-field>
+                    <vdr-form-field
+                        [label]="'marketing.usage-limit' | translate"
+                        for="usageLimit"
```
michaelbromley, line 99: Can we get a tooltip here where we can have a short explanation?

```
[tooltip]="'marketing.usage-limit-tooltip'"
```

We could also have a tooltip for `marketing.per-customer-limit` in the preceding form field too - this allows us to clearly describe the difference between these 2 similar fields in the UI.

michaelbromley (changes requested):
This is very good, thank you!

Just a couple of small points about the translation tokens.

michaelbromley:
Thank you! 
I plan to do a pre-release of the next minor version soon so this will be available to use.

Issue #2338, opened 2023-08-07
[Mollie Payments Plugin] Order line amounts are not refunded only shipping lines


seminarian:
**Describe the bug**
When doing a refund via Admin UI the `amount` sent to Mollie is not correct. Order line costs are not sent. Only shipping cost.

So there are two scenarios where things get dangerous:
1. No shipping cost gets refunded => refunding fails with message in Admin UI though the Order is cancelled. (Fails because a refund with amount 0 is sent to Mollie).
![image](https://github.com/vendure-ecommerce/vendure/assets/9881519/0e907064-6eea-4246-ba35-02914229617e)
```
[dev:server] {
[dev:server]   paymentId: 'tr_EurVpHQqdc',
[dev:server]   description: 'Customer request',
[dev:server]   amount: { value: '0.00', currency: 'EUR' }
[dev:server] }
```
2.  shipping cost + item lines are selected for refund => Refund seems to work but the amount is not correct
Only the shipping cost amount is refunded.
```
[dev:server] {
[dev:server]   paymentId: 'tr_g7BcVK8KZ6',
[dev:server]   description: 'Customer request',
[dev:server]   amount: { value: '3.99', currency: 'EUR' }
[dev:server] }
```

**To Reproduce**
1. Go to the Order you would like to refund
2. Click on Cancel order
4. Refund all except the shipping cost
5. See error popping up in Admin UI and no "refunds" are added

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.3
 - Nodejs version: v16.19.0
 - Database (mysql/postgres etc): postgres

@martijnvdbrug Would you have this issue as well?

seminarian:
@michaelbromley after thinking a bit more about it this might be something going wrong in the Vendure Core actually and not in the payment plugin?

martijnvdbrug:
Hi @seminarian,

Sounds like a big deal. However, I'm having some trouble reproducing your issue: Both canceling as well as refunding seem to work with the dev-server:

![image](https://github.com/vendure-ecommerce/vendure/assets/6604455/951692f7-a855-439b-a05e-fe3218f8a79f)

![image](https://github.com/vendure-ecommerce/vendure/assets/6604455/428558a1-5704-4d30-bd6c-dc43d0c007b8)

![image](https://github.com/vendure-ecommerce/vendure/assets/6604455/a305f3f3-dbae-4da6-9754-ae680d247a7a)


[...]

PR #2340, opened 2023-08-09
Fix/Mollie dev server

martijnvdbrug:
Did some fixes to the Mollie dev-server in an attempt to reproduce [this issue](https://github.com/vendure-ecommerce/vendure/issues/2338).


michaelbromley:
👍

PR #2348, opened 2023-08-17
Refactor core entity index and common index

cdiban:
Adding some export into core module, for example, the seller entity and calculated decorator. This improvement is for create plugins with this entity and decorator with more clean code.

Best.
Carlo.


michaelbromley:
Thanks!

PR #2349, opened 2023-08-21
fix(admin-ui): Add static title for add address dialog

### LINKED ISSUE #2311: Add customer address dialog form with reactive title!
**Describe the bug**
Adding a new customer address have a reactive title changes after changing `Street line 1` column!

**To Reproduce**
1. Go to [live demo](https://demo.vendure.io/admin/customer/customers/1)
2. Click on `create new address`.
3. Scroll down to the field Street line 1 and type something.
4. See title changed with the field.

**Expected behavior**
Having a static title

wpplumber:
See the [issue](#2311). (closes #2311)

### DIFF
--- a/packages/admin-ui/src/lib/customer/src/components/address-detail-dialog/address-detail-dialog.component.html
+++ b/packages/admin-ui/src/lib/customer/src/components/address-detail-dialog/address-detail-dialog.component.html
@@ -1,5 +1,6 @@
 <ng-template vdrDialogTitle>
-<span>{{ 'customer.create-new-address' | translate }}</span>
+<span *ngIf="addressForm.get('streetLine1')?.value">{{ 'customer.edit-address' | translate }}</span>
+<span *ngIf="!addressForm.get('streetLine1')?.value">{{ 'customer.create-new-address' | translate }}</span>
 </ng-template>
 
 <vdr-address-form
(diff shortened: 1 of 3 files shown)


packages/admin-ui/src/lib/customer/src/components/address-detail-dialog/address-detail-dialog.component.html
```diff
@@ -1,6 +1,5 @@
 <ng-template vdrDialogTitle>
-    <span *ngIf="addressForm.get('streetLine1')?.value as streetLine1">{{ streetLine1 }},</span>
-    <span *ngIf="addressForm.get('countryCode')?.value as countryCode"> {{ countryCode }}</span>
+<span>{{ 'customer.create-new-address' | translate }}</span>
```
michaelbromley: Hi - this only works for when you are creating a new address, but the same dialog is used when updating an existing address, so in that case the title is wrong.

michaelbromley (changes requested):
Thanks for the PR! This solution isn't quite complete though. Test it out when editing an existing address to see.

packages/admin-ui/src/lib/customer/src/components/address-detail-dialog/address-detail-dialog.component.html
tarikrital: Welcome, then I need to check the edit status and make other title which I think is `customer.edit-address`.

Issue #2344, opened 2023-08-11
Bi-directional Relations between CustomEntities and VendureEnteties


kefkalexis:
We want to create bi-directional relations between VendureEntities and our customEntites, for example:

````
@Entity()
export class MyCustomEntity extends VendureEntity {

  @ManyToMany(
    () => Collection,
    (c: Collection) => c.customFields?.theOtherSide,
  )
  public theOneSide: Collection[];
}

customFields.Collection = [
  {
    name: "theOtherSide",
    type: "relation",
    entity: MyCustomEntity ,
    list: true,
  },
]
````

We havn't found a solution to connet these both sides since typeOrm needs to know the jointable which is implicitly defined by vendure in `registerCustomFieldsForEntity`.

**Describe the solution you'd like**
We would propose to add the inverseSid-function used by typeorm as a field to the customFieldConfiguration:

````
customFields.Collection = [
  {
    name: "theOtherSide",
    type: "relation",
    entity: MyCustomEntity ,
    list: true,
    inverseSide: (x: MyCustomEntity ) => x.theOneSide,
  },
]
````

**Describe alternatives you've considered**
Another possibility would be to manipulate the typeOrm-metadata, but we would prefer to have it visible in the customField-definition

**Additional context**
If you agree with the proposal I could provide a pull request with the changes mentioned above

michaelbromley:
Hi,

Thanks for the report & proposal. Maybe this is a silly question, but why do you want to be able to define the inverse side? Is it so you can use the query builder API more conveniently when querying Collections and joining the custom relations?

I'd be happy to see a PR - if so make it against the `minor` branch please :)

kefkalexis:
Thanks for your response.

yes, it would make working with relations more convenient. 

I created a pull request here: https://github.com/vendure-ecommerce/vendure/pull/2365

PR #2365, opened 2023-08-27
feat(core): Support bi-directional relations in customFields

kefkalexis:
As mentioned in the issues (https://github.com/vendure-ecommerce/vendure/issues/2344) we would like to add the inversSide-Property for bidrection Relations on CustomFields.

With this it would be more convenientto work with a Custom Entities because it allows to query relations from both sides.


michaelbromley:
Thank you, this is an excellent PR 👍

Issue #2364, opened 2023-08-27
Bug in Stripe paymentIntent.create() triggered with express checkout + normal checkout


yasserlens:
**Describe the bug**
The stripe Service's createPaymentIntent() attaches metadata to stripe's paymentIntent.create() call. The metadata is not guaranteed to be the same for the same order at different stages of the checkout journey. This causes the following error:
Keys for idempotent requests can only be used with the same parameters they were first used with. Try using a key other than 'SVHLKDQ3DHMA6SPT_9000' if you meant to execute a different request.

The reason is as follows:
- Express checkout buttons are normally shown before customers fill in their contact and shipping details.
- To create the express checkout buttons, we need to create a payment intent.
- The call to do so tries to fetch customer information (email + shipping address), but this info is undefined at this stage
- This results in a request that contains metadata that contains undefined fields
- If the customer decides not to proceed with express shipping, they'll fill in their customer details (email + shipping info)
- Then they reach the payment page, we make another call to create a payment intent, which is needed to create the Stripe Element that collects card info.
- The new request to createPaymentIntent() has different metadata. 
- However, the idempotency key is the same (order code + total price).
- Due to that, Stripe will reject the call to create a payment intent, because the idempotency key is violated. The same idempotency key should have the same metadata (Express checkout call + later calls to create Stripe elements).
- Here's where the issue is in the Stripe Service:
https://github.com/vendure-ecommerce/vendure/blob/a83788073ac1b302f97bb0dc05bb9c9d68d776a3/packages/payments-plugin/src/stripe/stripe.service.ts#L43
- Any later call to create a payment intent for that order will fail (Stripe will reject it, which is expected).

**To Reproduce**
- Create a Payment intent before filling in customer details for an order (e.g. for express checkout)
- Try to create another payment intent after having customer details
- The 2nd attempt will fail, and all subsequent calls will fail.

```
Keys for idempotent requests can only be used with the same parameters they were first used with. Try using a key other than 'SVHLKDQ3DHMA6SPT_9000' if you meant to execute a different request.
```

**Expected behavior**
The idempotency key error shouldn't be triggered when calling createPaymentIntent multiple times for the same order. 

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.5
 - Nodejs version: v18.11.0
 - Database (mysql/postgres etc): Postgres
 - Stripe @ 8.222.0
 - @vendure/payments-plugin @ ^2.0.5

**Additional context**
[...]

PR #2366, opened 2023-08-27
Removed shipping address & customer email from payment intent metadata

yasserlens:
This should solve https://github.com/vendure-ecommerce/vendure/issues/2364

Removed shipping address and customer email from payment intent request metadata. This is because Stripe express payments require a payment intent, but at the point of doing an express payment we may not have customer details (they are fetched from the payment resolver, such as Google/Apple pay.

### DIFF
--- a/packages/payments-plugin/e2e/stripe-payment.e2e-spec.ts
+++ b/packages/payments-plugin/e2e/stripe-payment.e2e-spec.ts
@@ -204,7 +204,6 @@ describe('Stripe payments', () => {
             'metadata[channelToken]': E2E_DEFAULT_CHANNEL_TOKEN,
             'metadata[orderId]': '1',
             'metadata[orderCode]': activeOrder?.code,
-            'metadata[customerEmail]': customers[0].emailAddress,
         });
         expect(createStripePaymentIntent).toEqual('test-client-secret');
         StripePlugin.options.metadata = undefined;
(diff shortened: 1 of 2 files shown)


yasserlens:
Hey Michael

options.metadata() is deconstructed, generating the following pairs when an order has no customerEmail or shippingAddress attached:
```json
{
"customerEmail": "unknown",
"shippingAddress": "undefined undefined, undefined, undefined undefined, undefined",
}
```

Removing the call you referred to should omit the key-value pairs above. The rest are hard-coded:
```typescript
{ 
channelToken: ctx.channel.token,
orderId: order.id,
orderCode: order.code,
}
```

The unit tests don't check for shippingAddress, but they check for the customerEmail field, which is why I removed it from the checks to make the tests pass.

michaelbromley:
> options.metadata() is deconstructed, generating the following pairs when an order has no customerEmail or shippingAddress attached:

What I'm not understanding is that by default, the `options.metadata` function is undefined. This means that unless you yourself are defining a metadata function which returns the `customerEmail` and `shippingAddress` keys, how is this being included in the metadata that gets sent to Stripe?

yasserlens:
Great Catch - Michael. I actuall do define this metadata in Stripe's config when setting up (in vendure-config.ts).
The reason why they're their is to make sure customer details are attached to the payment, so we could find the customer payment in Stripe if needed given shipping info.

We can remove this and this should solve the issue without going with this change. 
However, keeping such metadata which can change from one call to another for the same order could result in this happening again to other Stripe plugin users. Do we want to keep this metadata in the paymentIntent.create() call?

Happy to discard the change if you think it's necessary to keep it.

michaelbromley:
Thanks for the clarification. I'm gonna close this because I don't think it's the right direction for a solution. I've made another suggestion in the original issue.

Issue #2375, opened 2023-09-01
Promotion disassociated from sub-channel when updating from default channel


seminarian:
**Describe the bug**
Imagine we have a promotion that is assigned to channel A. If somebody works in the admin UI and edits that promotion from within the default channel the result is that after saving suddenly the promotion isn't associated to channel A anymore.

**To Reproduce**
Imagine we have a promotion that is assigned to channel A. If somebody works in the admin UI and edits that promotion from within the default channel the result is that after saving suddenly the promotion isn't associated to channel A anymore.

**Expected behavior**
The assigned channels of the promotion stay intact

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.6
 - Nodejs version: v16.19.0
 - Database (mysql/postgres etc): postgres

**Additional context**
Will create PR.

PR #2376, opened 2023-09-01
fix: fix promotion channel association on update

seminarian:
closes https://github.com/vendure-ecommerce/vendure/issues/2375


michaelbromley:
Thanks!

PR #2378, opened 2023-09-01
feat: Added price field on variation dialog modal

JustDenP:
The "price" field has been added to the variation creation modal. Without it, a variation was always created with price 0.
Снимок экрана 2023-09-02 в 09 15 52


michaelbromley:
Thanks for this contribution! Good UX improvement for everyone :)

PR #2382, opened 2023-09-05
fix(core): Aligns items horizontally in the Chip component

alexisvigoureux:
## 🎯 Goal

Aligns the icon and label horizontally in the Chip component

## 📝 Approach

Before, the align items was set to baseline, so there was a horizontal shift between elements 
Capture d’écran 2023-09-05 à 09 28 03

After with the change to `center`
Capture d’écran 2023-09-05 à 09 28 26

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/chip/chip.component.scss
+++ b/packages/admin-ui/src/lib/core/src/shared/components/chip/chip.component.scss
@@ -51,7 +51,7 @@
     padding: 5px 8px;
     white-space: nowrap;
     display: flex;
-    align-items: center;
+    align-items: flex-start;
     gap: 2px;
 }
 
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks for the contribution. This change does fix some alignment issues with the chip component, but unfortunately messes up some other uses 🤦 

Example current:
image

with this change:
image

I agree that the case you highlighted does need fixing - but the solution need to account for the above scenario too somehow. Gotta love CSS 😆

alexisvigoureux:
Thanks for the catch-up 👀 , I didn't have this use case in mind. I have changed to `flex-start` which seems to work for our 2 use cases. Let me know if you think of another one
Capture d’écran 2023-09-07 à 16 41 00

Capture d’écran 2023-09-07 à 16 39 56

michaelbromley:
I want to do a release now so I just cherry-picked your initial commit: https://github.com/vendure-ecommerce/vendure/commit/19d3e515d4945c8fdc84334c2eb11d7cf2546b80

I'll close this now, thanks for your help!

PR #2402, opened 2023-09-19
fix(stripe): fix stripe payment handling

asonnleitner:
Implement transactional handling for Stripe Webhook

- Wrap all Stripe webhook-related operations inside a single transaction using
  `TransactionalConnection.withTransaction()`.
- Ensures that all database operations within the webhook use the "master" instance.

This change aims to solve the issue of database operations in the Stripe webhook not consistently
using the master instance, which led to inconsistencies in low-latency environments.


michaelbromley:
Thank you! 🙏

Issue #2353, opened 2023-08-22
OrderLine customField ui components are not shown in vendure v2


jacobfrantz1:
**Is your feature request related to a problem? Please describe.**
I have a very customization product that can't be represented by product variants. We solved that by adding string customFields to the order line with json to represent the customization.

In the V2 admin ui, the custom admin ui components we have do not show up, and only the text of the custom field is available as shown.
![Screenshot_20230822_141306](https://github.com/vendure-ecommerce/vendure/assets/69358280/3f59f5f6-9881-4716-a594-513b667fd827)

The V1 admin ui showed our custom components that we used to describe the customization, link to a 3D renderer with the customization, and show a preview image on the customized build
![Screenshot_20230822_141601](https://github.com/vendure-ecommerce/vendure/assets/69358280/e15eccf8-b6b5-4490-b194-d105e0466cc4)

**Describe the solution you'd like**
I think the custom admin ui components should show under the orderline when the custom fields are selected in the field selector like where they are placed in V1.

**Additional context**
Related Issue: #2347

Sorry for having to redact everything, bosses orders.

michaelbromley:
In v2.1 you'll be able to specify custom components for data table columns. You can test this on the `@next` release tag.

Docs here: https://beta-docs.vendure.io/guides/extending-the-admin-ui/custom-data-table-components/

jacobfrantz1:
@michaelbromley I tried this on the `@next` release, I don't think the custom data table components are implemented for the order-data-table, which is where I wish to use them.

michaelbromley:
Thanks for the fix!

PR #2420, opened 2023-09-29
feat(admin-ui): Implement custom components in order detail data table

jacobfrantz1:
Added custom data table components to the data table on the order detail page per my usecase: https://github.com/vendure-ecommerce/vendure/issues/2353

I was unable to get the dev-server compileUIExtensions to work so this is untested with a custom component defined, but I'm 90% sure it should work.


jacobfrantz1:
I tested it with a custom column component and it works.

My dev-server issue was that I didn't read the Readme :P

michaelbromley:
Thank you! I had overlooked this when implementing the custom components in the base data table2.

PR #2424, opened 2023-10-03
feat(translation): Added nepali translation

BeepLoveKarki:


michaelbromley:
Thanks! I noticed there are quite a few tokens not translated - 78% have been, but the entire "common" namespace is missing. Are you planning to do a 100% translation?

BeepLoveKarki:
Sure. We will be doing it.

michaelbromley:
OK, let me know when it is done and I can merge.

BeepLoveKarki:
@michaelbromley its ok now. 97%. Its max we can do in Nepali language.

PR #2435, opened 2023-10-10
fix(payments-plugin): List missing available Mollie payment methods for orders api

martijnvdbrug:
By default `mollie.methods.list()` lists payment methods for use with the `payments` API. However, the Mollie integration uses the `orders` integration. This resulted in some payment methods, like `Klarna pay later` not being listed.

This PR fixes that by passing `orders` to the mollie function: `mollie.methods.list({resource: 'orders'})`.


michaelbromley:
Thanks!

PR #2446, opened 2023-10-12
Fix/mollie klarna AutoCapture doesn't work!

martijnvdbrug:
Quite an impactful fix:

* Somewhere between V1.9 and v2.0.8 the relation `order.payments` got lost, resulting in Klarna Pay Later orders nver transitioning to PaymentSettled
* Without this fix orders stay in PaymentAuthorized, and shop owners never receive their money. Klarna has a policy that payments should be captured within 30 days, after that capturing payments (= receiving money) is not possible anymore.
* A test case is very useful here, but, time... I've added a Todo in the e2e file


michaelbromley:
Thanks for the fix!

PR #2447, opened 2023-10-12
fix(slug): Remove trademark symbols from normalized strings

gdarchen:
When importing a product, we resulted in a slug containing a `®` trademark symbol, which is invalid.

Therefore, in this PR, I made sure to remove the trademark symbols from the normalized strings, such as `©`, `®` and `™`.


michaelbromley:
Thank you!

Issue #2450, opened 2023-10-13
Error verifying Stripe webhook signature after upgrade to 2.1.0


lzzfnc:
**Describe the bug**
After upgrading to 2.1.0, arranging a payment with the stripe plugin crash the server. 

``` 
[server] error 10/13/23, 2:31 PM - [StripePlugin] Error verifying Stripe webhook signature t=1697207506,v1=4702b09d11c0388d6605a6e0f315b3db0e5e91b756cf347de16cc3c5d27e35e2,v0=b8a5a131f5ce8ab3dbfc26849737238b88e9675094c2b16527d94905ef67c569: No webhook payload was provided. 
[server] info 10/13/23, 2:31 PM - [StripePlugin] Stripe payment intent id pi_3O0mOlHJz38WEyqE0LpLgR0z added to order PYCG4ZY3N8EXSZM9 
[server] error 10/13/23, 2:31 PM - [Vendure Server] Cannot set headers after they are sent to the client
[server] Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
[server]     at new NodeError (node:internal/errors:387:5)
[server]     at ServerResponse.setHeader (node:_http_outgoing:644:11)
[server]     at ServerResponse.header (/usr/src/app/node_modules/express/lib/response.js:794:10)
[server]     at ServerResponse.send (/usr/src/app/node_modules/express/lib/response.js:174:12)
[server]     at StripeController.webhook (/usr/src/app/node_modules/@vendure/payments-plugin/package/stripe/stripe.controller.js:95:49)
[server]     at processTicksAndRejections (node:internal/process/task_queues:96:5) 
[server] node:internal/errors:478
2023[server]     ErrorCaptureStackTrace(err);
[server]     ^
[server] 
```

**To Reproduce**
1. Upgrade to 2.1.0
2. Send a payment with Stripe 
3. See error

**Expected behavior**
The Stripe plugin should be able to verify the webhook signature

**Environment (please complete the following information):**
 - @vendure/core version: 2.1.0
 - Nodejs version 18
 - Database: postgres etc
 - Deployed on Northflank

**Additional context**
Note that in the Stripe dashboard the payment arrives. 

Could probably be related to: https://github.com/vendure-ecommerce/vendure/issues/2444

lzzfnc:
The same procedure test in local still does not work, but it doesn't crash the server nor log any error

carathorys:
[...]

PR #2454, opened 2023-10-16
fix(payments-plugin): Stripe controller crashes server instance

carathorys:
This patch is slightly related to the issue [2450](https://github.com/vendure-ecommerce/vendure/issues/2450), but fixes only the crash side of it: 

I've put the last `response.status(HttpStatus.OK).send('Ok')` to the transaction block, otherwise it crashes the backend with the following error: 
```
...
[server] Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
[server]     at new NodeError (node:internal/errors:387:5)
[server]     at ServerResponse.setHeader (node:_http_outgoing:644:11)
[server]     at ServerResponse.header (/usr/src/app/node_modules/express/lib/response.js:794:10)
[server]     at ServerResponse.send (/usr/src/app/node_modules/express/lib/response.js:174:12)
[server]     at StripeController.webhook (/usr/src/app/node_modules/@vendure/payments-plugin/package/stripe/stripe.controller.js:95:49)
[server]     at processTicksAndRejections (node:internal/process/task_queues:96:5) 
[server] node:internal/errors:478
...
```

Please note that this does not solves the issue that the `rawBody` is not defined because the middleware is not executed, but at least the backend won't crash because of any unidentified issues.

### DIFF
--- a/packages/payments-plugin/e2e/stripe-payment.e2e-spec.ts
+++ b/packages/payments-plugin/e2e/stripe-payment.e2e-spec.ts
@@ -10,7 +10,9 @@ import { CREATE_PRODUCT, CREATE_PRODUCT_VARIANTS } from '@vendure/core/e2e/graph
 import { createTestEnvironment, E2E_DEFAULT_CHANNEL_TOKEN } from '@vendure/testing';
 import gql from 'graphql-tag';
 import nock from 'nock';
+import fetch from 'node-fetch';
 import path from 'path';
+import { Stripe } from 'stripe';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import { initialData } from '../../../e2e-common/e2e-initial-data';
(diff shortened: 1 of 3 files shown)


carathorys:
Ok, now I fixed the issue _(hope it will work for everybody else as well)_
The solution is strange *(and straightforward in the meantime)*: 
The `body-parser` needs a type parameter in order to execute, but if you're using `raw(...)`  middleware, it will put the `Buffer` into the `request.body` object. 
So if you use `json(...)` instead, and in the `verify(...)` method you put the `rawBody` into the `request` it seems like it will solve the issue. 

This is my solution now, should I put it into the same PR, or you plan to solve this somehow else? :slight_smile: 

```typescript
import { json } from 'body-parser';
import * as http from 'http';

import { RequestWithRawBody } from './types';

/**
 * Middleware which adds the raw request body to the incoming message object. This is needed by
 * Stripe to properly verify webhook events.
 */
export const rawBodyMiddleware = json({
    type: '*/*',
    verify(req: RequestWithRawBody, res: http.ServerResponse, buf: Buffer, encoding: string) {
        if (Buffer.isBuffer(buf)) {
            req.rawBody = Buffer.from(buf);
        }
        return true;
    },
});
```

carathorys:
Ok, the previos solution doesn't work _(thankfully, the E2E tests catched it in time!)_
I've reverted the `body-parser`, and made a change in the `StripeController`.
Now I'm trying to write at least an e2e test to the webhook verification, but I've ran into some issues with the order service: 
It is unable to execute the `findOneByCode(...)` function as it first executes a query to fetch an order by it's code, and then it tries to fetch the order by ID from the database. 

@michaelbromley do you have any tips how I can solve that?
Currently the test I wrote is failing in the in the `stripe.controller.ts:61` because of that issue. 
Also noticed that this function loads an order twice: 
First it fetches the order by it's `code`, with one relation (`Customer`), then if it founds the record, it fetches the order again, this time by `id`, and with the `relations` parameters.

carathorys:
Ok, now it seems to be working! :) 
Added some tests to validate wether the Stripe signature validation works or not. In the order payload it only works if I replace the `"T_"` from the `orderId`, I hope it's not a problem!

michaelbromley:
Thank you for the considerable work you must have put in for this investigation & fix! 🙏

PR #2460, opened 2023-10-17
fix(admin-ui): Do not load pending search index updates if permissions are insufficient

### LINKED ISSUE #2456: Job error of session for user with no product/category/job based permissions
When I create a new role with no permissions related to a product, category, or job, I get no pendingSearchIndexUpdates error and the admin logs it.

This didn't happen on prior versions.

jnugh:
The `pending-search-index-updates` executes a query that returns `forbidden` if neither `ReadCatalog`  nor `ReadProduct` permissions are available. This loads to an automatic logout in admin-ui. This change prevents the call by adding a new configuration property to the `configureAlert` method. If any permission passed via `requiredPermissions` exists the alert is being created. 
If not the alert will be registered if at some point the permission is added and `userStatus` updates.

Closes #2456

### DIFF
--- a/packages/admin-ui/src/lib/core/src/providers/alerts/alerts.service.ts
+++ b/packages/admin-ui/src/lib/core/src/providers/alerts/alerts.service.ts
@@ -103,13 +103,12 @@ export class AlertsService {
 
     configureAlert<T>(config: AlertConfig<T>) {
         this.hasSufficientPermissions(config.requiredPermissions)
-            .pipe(
-                filter(canReadPendingSearchIndexUpdates => canReadPendingSearchIndexUpdates),
-                first(),
-            )
-            .subscribe(() => {
-                this.alertsMap.set(config.id, new Alert(config));
-                this.configUpdated.next();
+            .pipe(first())
+            .subscribe(hasSufficientPermissions => {
+                if (hasSufficientPermissions) {
+                    this.alertsMap.set(config.id, new Alert(config));
+                    this.configUpdated.next();
+                }
             });
     }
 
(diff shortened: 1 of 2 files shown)


packages/admin-ui/src/lib/core/src/providers/alerts/alerts.service.ts
```diff
@@ -91,8 +102,26 @@ export class AlertsService {
     }
 
     configureAlert<T>(config: AlertConfig<T>) {
-        this.alertsMap.set(config.id, new Alert(config));
-        this.configUpdated.next();
+        this.hasSufficientPermissions(config.requiredPermissions)
+            .pipe(
+                filter(canReadPendingSearchIndexUpdates => canReadPendingSearchIndexUpdates),
```
michaelbromley: If I understand this code correctly: 

If the `hasSufficientPermissions()` observable emits `false`, this filter will block the emitted value from ever reaching the `subscribe()` function, and the subscription will stay open and not complete. 

One way to avoid this would be to remove the `filter()` clause and inside the subscriber you can inspect the value and only set the map if the value is `true`.

michaelbromley (changes requested):
Thanks for taking on this issue! I really like your solution - I just have one point that needs checking.

packages/admin-ui/src/lib/core/src/providers/alerts/alerts.service.ts
jnugh: I thought the user data might refresh periodically. This would eventually add the alert once permissions change. However the data is only loaded once - so you are obviously right.

I moved the check into the subscription as you suggested. The observable will now always close.
I need to skip the first `userStatus` value as permissions have not been fetched yet. I therefore added a filter for the `userStatus` stream:
```ts
filter(({ userStatus }) => userStatus.isLoggedIn),
```

michaelbromley:
Thanks for your help on this! 👍

PR #2461, opened 2023-10-17
feat(admin-ui): Add product slug in product multi selector dialog component

gdarchen:
### Initial issue

When trying to fill in the collections by manually selecting products, we sometimes have issues because products of different sellers have the exact same names.

For instance, we have a lot of `Amazon` products which we can't distinguish between all our vendors.

image

### Solution

In this PR, I enhanced the `product-multi-selector-dialog` Angular component so as to display the product `slug` in case the display mode is `product`.

Therefore, we know exactly what product corresponds to each card.

image

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/product-multi-selector-dialog/product-multi-selector-dialog.component.html
+++ b/packages/admin-ui/src/lib/core/src/shared/components/product-multi-selector-dialog/product-multi-selector-dialog.component.html
@@ -37,6 +37,12 @@
                 <span [title]="mode === 'product' ? item.productName : item.productVariantName">{{
                     mode === 'product' ? item.productName : item.productVariantName
                 }}</span>
+                <div *ngIf="mode === 'product'">
+                    <small>
+                        <span class="slug-label">slug:</span>
+                        <code>{{ item.slug }}</code>
+                    </small>
+                </div>
                 <div *ngIf="mode === 'variant'"><small>{{ item.sku }}</small></div>
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


packages/admin-ui/src/lib/core/src/shared/components/product-multi-selector-dialog/product-multi-selector-dialog.component.html
```diff
@@ -37,6 +37,12 @@
                 <span [title]="mode === 'product' ? item.productName : item.productVariantName">{{
                     mode === 'product' ? item.productName : item.productVariantName
                 }}</span>
+                <div *ngIf="mode === 'product'">
+                    <small>
+                        <span class="slug-label">slug:</span>
```
michaelbromley: This should be a translation token so that it works in other UI languages:

```html
<span class="slug-label">{{ 'common.slug' | translate }}:</span>
```

You could also consider using one of the built-in utility classes for the margin-right, e.g. `mr-1`, `mr-2` etc (whichever looks right to you). The benefit of this is that

1. no need to add another class to the stylesheet
2. we will be adding support to make those utility classes work for RTL languages like Arabic

michaelbromley (changes requested):
Thanks for the quick fix! Just a couple of points to take a look at.

packages/admin-ui/src/lib/core/src/shared/components/product-multi-selector-dialog/product-multi-selector-dialog.component.html
gdarchen: Thank you very much for the quick review, and thanks for the tips!
I applied the suggested changes in [#1b48276](https://github.com/vendure-ecommerce/vendure/pull/2461/commits/1b48276a7e2df53f5fee37e16be3bbb387f661e8).

michaelbromley:
Awesome! Thank you 🙏

Issue #2395, opened 2023-09-13
Vendure can not find collection by slug if 2 collection have the same slug in different channels.


ttournie:
**Describe the bug**
Vendure can not find collection by slug if 2 collection have the same slug in different channels.

**To Reproduce**
1. Create a channel A
2. Create a channel B
3. in the channel A create a collection with slug 'test'
4. in the channel B create a collection with slug 'test'
5. try to get the channel by slug test in channel B (or A if channel B works). For this you can use the collection query

**Expected behavior**
It should return the right collection

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.6
 - Nodejs version : 18.17
 - Database (mysql/postgres etc): postgres

**Additional context**

After investigation, it’s because the first query in CollectionService findOneBySlug, is not contextualised with the channel, but the last query is.
So it gets all the collection matching the criteria → if severals it take the first of the list, then try to get the collection by ID on the channel, but because the selected item in the first query is not from the right channel it does not finds it.

PR #2480, opened 2023-10-23
fix(core): Fix collection findBySlug issue2395 

ttournie:
Fix the : https://github.com/vendure-ecommerce/vendure/issues/2395 issue.

Return the first found collection matching the findBySlug criteria.

To Test :

1. Create 2 channel
2. Create a collection with slug 'test' in the first channel
3. Create a collection with the same slug the the other channel
4. Go to the admin-api playground (or shop-api).
5. Use the `collection` query with the header set to the first channel token -> you get the collection of this channel
6. Do the same with the header set to the other channel -> you should get the other collection (the one corresponding to the channel in the header).

### DIFF
--- a/packages/core/src/service/services/collection.service.ts
+++ b/packages/core/src/service/services/collection.service.ts
@@ -228,11 +228,26 @@ export class CollectionService implements OnModuleInit {
         if (!translations?.length) {
             return;
         }
-        const bestMatch =
-            translations.find(t => t.languageCode === ctx.languageCode) ??
-            translations.find(t => t.languageCode === ctx.channel.defaultLanguageCode) ??
-            translations[0];
-        return this.findOne(ctx, bestMatch.base.id, relations);
+
+        let bestMatches = translations.filter(t => t.languageCode === ctx.channel.defaultLanguageCode);
+
+        if (!(bestMatches.length > 0)) {
+            bestMatches = translations.filter(t => t.languageCode === ctx.channel.defaultLanguageCode);
+        }
+
+        // In case there is best matches return the first found in the channel.
+        let collection: Translated<Collection> | undefined;
+        if (bestMatches.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/collection.service.ts
```diff
@@ -228,11 +228,26 @@ export class CollectionService implements OnModuleInit {
...
-            translations[0];
-        return this.findOne(ctx, bestMatch.base.id, relations);
+
+        let bestMatches = translations.filter(t => t.languageCode === ctx.channel.defaultLanguageCode);
+
+        if (!(bestMatches.length > 0)) {
+            bestMatches = translations.filter(t => t.languageCode === ctx.channel.defaultLanguageCode);
+        }
+
+        // In case there is best matches return the first found in the channel.
```
gdarchen: 🔤 **typo:**

```suggestion
        // In case there are better matches, we return the first found in the channel.
```

michaelbromley:
Thanks for both the issue report and the fix! I just tried it out locally and it does indeed solve the issue. However, I was a little concerned about the fact that this solution has to make multiple `this.findOne()` calls in a for loop. I can imagine that in certain cases where a _lot_ of languages are used, this may have a negative perf impact.

I found that we can avoid this by directly asserting the channel in the initial query to get the translations a the start of the method. It would look like this:

```ts
const translations = await this.connection.getRepository(ctx, CollectionTranslation).find({
    relations: ['base'],
    where: {
        slug,
        base: {
            channels: {
                id: ctx.channelId,
            },
        },
    },
});
```

Unless I'm missing some issue with doing it this way, it seems preferable.

ttournie:
Oh sure, I did not know you could put a where constraint on the base channels Id like that. So if this works, it's better, yes !

michaelbromley:
Thanks!

PR #2482, opened 2023-10-23
feat(core): Enable to change the name of the cookie for Admin API

### LINKED ISSUE #2481: Name differently the cookies for the Admin API and the Shop API

We would like to be able to give different names to the cookies used for the Admin UI and the Shop API.

Indeed, we are logged out from the AdminUI whenever we log on the storefront and _vice-versa_.

We would love having the ability to override the name of the cookie set for the Admin API.
We thought at something in the Vendure config such as `authOptions.adminCookieName`.

We also considered using the [_Partitioned Cookies_](https://developer.mozilla.org/en-US/docs/Web/Privacy/Partitioned_cookies) to identify the host originating the request which persists the cookie, but this seems to be experimental so far.

**Additional context**

gdarchen:
### Issue

Fixes #2481 

### Description

In this PR, I added a new `authOptions.adminCookieName` optional parameter in the Vendure config. 
It makes it possible to give a custom name to the token set for the Admin API than the one given to the Shop API.

For now, here is the cookie naming behavior:
- If not setting the `authOptions.cookieOptions.name` parameter:
  - Both the Admin API and Shop API cookies names will be the default one: `'session'`
- If setting the `authOptions.cookieOptions.name` parameter to something like `'cookie-name'`:
  - Both the Admin API and Shop API cookies names will be the configured one: `'cookie-name'`

There is currently no way to give a custom name to the Admin API cookies, different from the Shop API ones.

With this changes, we can therefore set different cookies names for both APIs.

### How to test

1. In the `dev-config.ts`, change the `authOptions` as follows to name Shop API cookies `shop-api` and `shop-api.sign`, and the Admin API cookies `admin-api` and `admin-api.sig`:
   ```ts
    authOptions: {
        disableAuth: false,
        tokenMethod: ['bearer', 'cookie'] as const,
        requireVerification: true,
        customPermissions: [],
        cookieOptions: {
            name: "shop-api",
            secret: 'abc',
        },
        adminCookieName: 'admin-api'
    },
   ```
2. Start the `dev-server`
3. Open the GraphQL playground on the `/admin-api` and authenticate
4. You should see the following cookies
   image
5. Open the GraphQL playground on the `/shop-api` and authenticate
6. You should now see the following cookies
   image


michaelbromley:
I have a deep-seated aversion to adding new config properties to the Vendure config object. So here's another suggestion:

what if we change the type of the existing `cookieOptions.name` property to:

```ts
{
  name: string | { shop: string; admin: string; }
}
```

Internally the implementation could be similar, but just check the type of this property.

gdarchen:
> I have a deep-seated aversion to adding new config properties to the Vendure config object. So here's another suggestion:
> 
> what if we change the type of the existing `cookieOptions.name` property to:
> 
> ```ts
> {
>   name: string | { shop: string; admin: string; }
> }
> ```
> 
> Internally the implementation could be similar, but just check the type of this property.

Done in [`ead93f1` (#2482)](https://github.com/vendure-ecommerce/vendure/pull/2482/commits/ead93f1da7e30eacf7c1bc2f64b40b902a486f41)

I tested locally using the GraphQL Playgrounds and it seems to work as expected 😊

packages/core/src/bootstrap.ts
```diff
@@ -65,7 +66,21 @@ export async function bootstrap(userConfig: Partial<VendureConfig>): Promise<INe
         tokenMethod === 'cookie' || (Array.isArray(tokenMethod) && tokenMethod.includes('cookie'));
     if (usingCookie) {
         const { cookieOptions } = config.authOptions;
-        app.use(cookieSession(cookieOptions));
+        app.use(
```
michaelbromley: Can we wrap this whole part (from `const { tokenMethod } = ...` in a function `configureSessionCookies()`?

Then it can be exported and used inside the test server too to avoid the repetition.

michaelbromley (changes requested):
Thanks, it looks good. I just suggested a small refactor to avoid duplication in the testing package and make maintenance easier.

Also I think this should go on the `minor` branch because it's a new API being added.

packages/core/src/bootstrap.ts
gdarchen: Done in [`203fd90` (#2482)](https://github.com/vendure-ecommerce/vendure/pull/2482/commits/203fd900c3db382b76dc8ef8efa7ed1efe05afc8) 👍

michaelbromley:
Thank you!

gdarchen:
Hello @michaelbromley (and happy new year, it's allowed since we are still in January 🎉!),

Do you have an estimated release date of the future `vendure@2.2` including this change?
This is really awaited by our admins ☺️

michaelbromley:
Hi,
No fixed date, but I'd like to get it released in February. In the mean time you could consider using the `@next` release (currently `2.2.0-next.1`) which already includes this feature.

daniel-lxs:
Is this feature released? or is it still pending?

michaelbromley:
final release of v2.2 will be next week

Issue #2019, opened 2023-02-02
Customer group promotion condition not encoded correctly with uuid strategy


jacobfrantz1:
**Describe the bug**
When using uuid entity id strategy, the customer group promotion condition does not work. I think it has to do with encoding the condition in the create and update promotion mutations, as manually changing the database to remove the escaped quotes makes the promotion work as intended.

When trying to track down the issue in the code, a print inside the idsAreEqual function revealed the customerGroupId was encoded with an extra set of quotation marks
e.g. 53dc7cc8-2808-4f2c-90f5-aa4e8725f9fd != "53dc7cc8-2808-4f2c-90f5-aa4e8725f9fd"

**To Reproduce**
1. Use the UUID Entity Id Strategy
2. Create a Customer Group
3. Create a Promotion with the Customer Group Condition
4. Refresh Page, Customer group is now empty. Promotion is not applied to customer group.

**Expected behavior**
The promotion should apply to the customer group after creating or updating the promotion via 

**Environment (please complete the following information):**
 - @vendure/core version: 1.9.2
 - Nodejs version: 16.18.1
 - Database: Postgresql 14.5

jacobfrantz1:
This issue is still a problem, I ran into it today when adding a promotion, and it got worse in v2 with the admin-ui not able to set customer group condition with both uuid strategy and autoincrement id strategy.

I've tracked down the root of the issue, as well as a new issue introduced in v2.

I'll do a pull request in the next couple of days or so fixing this.

As for the new issue, the customer-group-form-input emits the group object instead of the id as of v2, and that doesn't work for the customer group promotion condition or another configurable operation I use in my server, as they expect ID/Array of IDs instead of Group/Array of Groups. This is stored in the db as "\"[Object object]\"".

This root issue I ran into is due to the configurable-operation-codec.

TLDR: ConfigurableOperationCodec's decodeConfigurableOperationIds shouldn't JSON.stringify the id if it is a string, and maybe shouldn't JSON.stringify the id at all (what I'm leaning towards), as it is JSON.stringified in the encodeConfigurableOperationIds function.

With the autoincrement id strategy, in decodeConfigurableOperationIds, the id is decoded to a number and JSON.stringified to "1" for example. In encodeConfigurableOperationIds, the id is encoded, a no-op as it was stringified earlier, and JSON.stringified to "\"1\"".

[...]

PR #2483, opened 2023-10-23
fix(core): fix configurable operation id decoding when using uuids

jacobfrantz1:
Fixes issue #2019 

Id's were getting JSON stringified when they shouldn't have been. It only worked for integer ids because `JSON.stringify(1) == 1`. For uuids, `JSON.stringify("XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX") != "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"`.

Later, when it uses the values in business logic: https://github.com/vendure-ecommerce/vendure/blob/218561f04cbaf738c0f18302b67b1b077a056a7f/packages/core/src/common/configurable-operation.ts#L459-L460 , JSON.parse() is not called on ids.

I can't guarantee this doesn't break anything as the change is in a critical area.

### DIFF
--- a/packages/core/src/api/common/configurable-operation-codec.ts
+++ b/packages/core/src/api/common/configurable-operation-codec.ts
@@ -43,8 +43,7 @@ export class ConfigurableOperationCodec {
                         const decodedIds = ids.map(id => this.idCodecService.decode(id));
                         arg.value = JSON.stringify(decodedIds);
                     } else {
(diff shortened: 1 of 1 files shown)


michaelbromley:
I tested this out locally and it solves the issue. The fact that none of the e2e tests are breaking is a good sign that it should be safe.

Thank you for taking the time to investigate and fix this!

PR #2489, opened 2023-10-27
Fix tanstack responseMiddleware

DanielBiegler:
Under successful conditions `errors` is undefined - Meaning you'd never go down this branch and persist your token like you'd actually want. :smile: 

Btw - on another note: I had to manually search for the file because the "Edit this page" buttons dont work - those point to the wrong branch "new-docs"


michaelbromley:
Thanks!

And also for the heads-up on the edit link - I'll get that fixed asap.

PR #2499, opened 2023-10-31
fix(mollie): Add missing `status` type

casperiv0:
# Description

Adds a missing `status` field to the `molliePaymentMethods` query. This type was previously not typed by Mollie. It is documented however. 

PR on Mollie's API plugin: https://github.com/mollie/mollie-api-node/pull/335

# Breaking changes

None.

# Checklist

:pushpin: Always:

:zap: Most of the time:

### DIFF
--- a/packages/payments-plugin/src/mollie/graphql/generated-shop-types.ts
+++ b/packages/payments-plugin/src/mollie/graphql/generated-shop-types.ts
@@ -1671,6 +1671,7 @@ export type MolliePaymentMethod = {
   image?: Maybe<MolliePaymentMethodImages>;
   maximumAmount?: Maybe<MollieAmount>;
   minimumAmount?: Maybe<MollieAmount>;
+  status?: Maybe<Scalars['String']['output']>;
 };
 
 export type MolliePaymentMethodImages = {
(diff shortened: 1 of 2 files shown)


casperiv0:
Not sure if the generated types have to be included, my guess is that they're formatting changes. `yarn format` seems to be configured incorrectly, unfortunately?

michaelbromley:
Hi! 
Thanks for the PR 👍 
@martijnvdbrug would you mind just quickly reviewing the addition of the field here? 

Regarding the generated type files - yes they _should_ be updated just to add this new field, but that should be a single line change. What happens if you run codegen without then formatting afterwards?

casperiv0:
> Regarding the generated type files - yes they should be updated just to add this new field, but that should be a single line change. What happens if you run codegen without then formatting afterwards?

Should be resolved now :)!

michaelbromley:
> It doesn't actually! It will also return ones with status null.

Is that a bug in the Mollie API or SDK? We're using the mollie client `methods.list` method:

https://github.com/vendure-ecommerce/vendure/blob/2f7a8d5f586fe22d67d414cc46c4048b019da60e/packages/payments-plugin/src/mollie/mollie.service.ts#L339

and the jsdoc for that method says:

> Retrieve all enabled payment methods. The results are not paginated.

whereas in the [mollie docs](https://docs.mollie.com/reference/v2/methods-api/get-method#response) it says that `null` signifies that the payment has not been requested 
image


so what doesn't make sense to me is how a "not requested" (null) method can at the same time be "enabled". Something doesn't add up.

michaelbromley:
OK, thanks for the clarifications. There's no harm adding this field even in future if Mollie fixes their probably bug and then it will just always be "activated"

martijnvdbrug:
Oh wow, good find. Then yes, in this case we should just proxy the `status` field. Thanks!

PR #2515, opened 2023-11-07
fix(core): Default search plugin query for non-default language fails

hans-rollingridges-dev:
# Description

The - default search plugin - search query for a non-standard language fails. If there is a product variant with non-standard translation, the standard language variant will also be included in the search results. I.e. resulting in search result records with duplicate product variant ID. Depending on the logic for sorting and grouping by, the outcome will be different.

# Breaking changes

# Screenshots

# Checklist

:pushpin: Always:

:zap: Most of the time:


michaelbromley:
We have a load testing setup described here: https://github.com/vendure-ecommerce/vendure/tree/master/packages/dev-server#load-testing

There's no script specifically for testing the search, but you could create one based on one of the existing scripts.

michaelbromley:
Thank you @hans-rollingridges-dev for your excellent work on this!

PR #2516, opened 2023-11-07
feat(mollie-plugin): Support extra parameters for listing methods

### LINKED ISSUE #2510: Mollie plugin returns unavailable payment methods for given currency
**Describe the bug**
The `molliePaymentMethods` query in the MolliePlugin, returns all enabled payment methods. However, it doesn't forward any currency data to Mollie, which therefore just returns all enabled payment methods. For a proper checkout flow, the plugin should leave out methods that are not supported for the current order's currency.

**To Reproduce**
1. Add the MolliePlugin
2. Make sure you have "Klarna" or "Bancontact" enabled in Mollie
3. Create an order with a `GBP` currency
4. Call `molliePaymentMethods`
5. Both "Klarna" and "Bancontact" will be returned.
6. Try to get a payment intent with either of above methods
7. An error will be thrown that this currency is not supported for the methods

**Expected behavior**
The `molliePaymentMethods` should not return unelligible payment methods for the current order's currency (or billing country, for that matter).

It would also be great if extra information (e.g. `locale`) is passed along, so payment methods can optionally be translated and/or sorted for better UX.

 - @vendure/core version: 2.1.1
 - Nodejs version: 18

casperiv0:
# Description

This PR includes:

- Allow consumers of the Mollie Plugin to pass through `billingCountry`, `amount` and `locale` as per Mollie's documentation: https://docs.mollie.com/reference/v2/methods-api/list-methods#parameters

# Breaking changes

Nope

# Checklist

:pushpin: Always:
- [x] I have [checked my own PR](#2516 "Fix typo's and remove unused or commented out code")

:zap: Most of the time:

closes #2510

### DIFF
--- a/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
+++ b/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
@@ -348,12 +348,8 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
         });
 
         it('Should get available paymentMethods', async () => {
-            const currency = 'USD';
-            const amount = '909.90';
-            const search = encodeURI(`amount[value]=${amount}&amount[currency]=${currency}&resource=orders`);
-
             nock('https://api.mollie.com/')
-                .get(`/v2/methods?${search}`)
+                .get('/v2/methods?resource=orders')
                 .reply(200, mockData.molliePaymentMethodsResponse);
             await shopClient.asUserWithCredentials(customers[0].emailAddress, 'test');
             const { molliePaymentMethods } = await shopClient.query(GET_MOLLIE_PAYMENT_METHODS, {
--- a/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
+++ b/packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
@@ -348,8 +348,12 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
         });
 
         it('Should get available paymentMethods', async () => {
+            const currency = 'USD';
+            const amount = '909.90';
+            const search = encodeURI(`amount[value]=${amount}&amount[currency]=${currency}&resource=orders`);
+
             nock('https://api.mollie.com/')
-                .get('/v2/methods?resource=orders')
+                .get(`/v2/methods?${search}`)
                 .reply(200, mockData.molliePaymentMethodsResponse);
             await shopClient.asUserWithCredentials(customers[0].emailAddress, 'test');
             const { molliePaymentMethods } = await shopClient.query(GET_MOLLIE_PAYMENT_METHODS, {
--- a/packages/payments-plugin/src/mollie/graphql/generated-shop-types.ts
+++ b/packages/payments-plugin/src/mollie/graphql/generated-shop-types.ts
@@ -1644,6 +1644,12 @@ export type MollieAmount = {
   value?: Maybe<Scalars['String']['output']>;
 };
 
+export type MollieAmountInput = {
+  __typename?: 'MollieAmountInput';
+  currency: Scalars['String']['output'];
+  value: Scalars['Int']['output'];
+};
+
 export type MolliePaymentIntent = {
   __typename?: 'MolliePaymentIntent';
   url: Scalars['String']['output'];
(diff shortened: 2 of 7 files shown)


StampixSMO:
I'm being undecisive if it makes more sense to have Vendure just calculate the amount/currency, instead of having a frontend pass it in via GQL inputs.

On one hand, it allows a for a broader usage of the query endpoint, on the other hand it allows for unintentional wrong currencies or values that are not equal to the active order one's.

Curious to hear input from @martijnvdbrug & @michaelbromley

martijnvdbrug:
I agree with @StampixSMO, I would say `amount`, `currency` and `billingCountry` should be passed to Mollie by Vendure. These are all properties of an order and should not be determined by the frontend.

I do see `locale` as a front end responsibility, though, so that would make sense to pass in to the eligible methods query.

StampixSMO:
Thanks for the input @martijnvdbrug

Shouldn't the locale also just be inferred from the `RequestContext` instead? In the end, that's also controlled by the frontend (using the proper query parameter) so it'd boil down to the same thing but aligns better with the rest of the core.

casperiv0:
> Shouldn't the locale also just be inferred from the RequestContext instead? In the end, that's also controlled by the frontend (using the proper query parameter) so it'd boil down to the same thing but aligns better with the rest of the core.

The locale used by Mollie will most likely be different then the one used by a storefront. Mollie only accepts certain locales :)

martijnvdbrug:
@StampixSMO Good point, I agree on that. But as @casperiv0 said, Mollie unfortunately only supports specific locales. For payment intent creation we also do a best guess of the locale ([see this line](https://github.com/vendure-ecommerce/vendure/blob/ea03d523de4b7c8fbffc2a8bf48625c6b88aa811/packages/payments-plugin/src/mollie/mollie.service.ts#L180)), so it would make sense to be able to overidde this from a storefront perspective.

This goes both for `eligibleMollieMethods` as well as for `createPaymentIntent` (@casperiv0 if you want the payment intent to be in the same locale, you'd have to add locale to that mutation as well`

casperiv0:
@martijnvdbrug, I've added the locale as an input type for the mutation. This PR is now ready for review :)

casperiv0:
Failing build doesn't look to be related to this PR

michaelbromley:
Thanks for this PR and for the feedback which helps me understand the issues involved.

I have a couple of questions:

1. Is there any way this could act as a breaking change? Could it happen that someone upgrades and then the methods being returned from Mollie changes in their checkout?
2. Is it possible a user might want to specifically omit one or more of these properties when listing methods? Would we need to consider some way to configure this? 
3. Is it possible a user might want to include _other_ properties such as `sequenceType` or `includeWallets`? 

The way we do it in the [Stripe Plugin](https://docs.vendure.io/reference/core-plugins/payments-plugin/stripe-plugin/#metadata) is that we allow the user to supply a function in the plugin `init()` method which defines what metadata we send to Stripe when interacting with their APIs. Could a similar pattern make sense here?

casperiv0:
> Thanks for this PR and for the feedback which helps me understand the issues involved.
> 
> I have a couple of questions:
> 
> 1. Is there any way this could act as a breaking change? Could it happen that someone upgrades and then the methods being returned from Mollie changes in their checkout?
> 2. Is it possible a user might want to specifically omit one or more of these properties when listing methods? Would we need to consider some way to configure this?
> 3. Is it possible a user might want to include _other_ properties such as `sequenceType` or `includeWallets`?
> 
> The way we do it in the [Stripe Plugin](https://docs.vendure.io/reference/core-plugins/payments-plugin/stripe-plugin/#metadata) is that we allow the user to supply a function in the plugin `init()` method which defines what metadata we send to Stripe when interacting with their APIs. Could a similar pattern make sense here?

That is possible that users would want to provide more information. I think adding that metadata option would be beneficial for all? I'll start implementing that soon.

casperiv0:
Implemented for the enabled payment methods query. Let me know if it should also be done for the payment intent :)!

michaelbromley:
@casperiv0 excellent work! 

So last thing: this should be made again the `minor` branch because it is a new feature.

And yes let's add support for the payment intent too but we'll get this one merged first and then do it as a separate PR.

casperiv0:
Thanks! @michaelbromley base branch has been updated!

PR #2531, opened 2023-11-17
feat(email-plugin): Allow currencies with formatMoney helper

### LINKED ISSUE #2530: Allow currency and locale with the `formatMoney` helper
Emails should list the prices for products, along with their currencies

The `formatMoney` function to accept a currency and optional locale, so it all doesn't have to be hardcoded.

Hardcoding, which is not very scalable since we shouldn't reinvent the wheel (i.e. what `Intl` has done for us).

StampixSMO:
# Description

Added 2 new parameters to the `formatMoney` helper: currency and locale

Closes #2530

# Breaking changes

No

# Checklist

:pushpin: Always:

:zap: Most of the time:

### DIFF
--- a/docs/docs/reference/core-plugins/email-plugin/index.md
+++ b/docs/docs/reference/core-plugins/email-plugin/index.md
@@ -108,7 +108,10 @@ Dynamic data such as the recipient's name or order items are specified using [Ha
 
 The following helper functions are available for use in email templates:
 
-* `formatMoney`: Formats an amount of money (which are always stored as integers in Vendure) as a decimal, e.g. `123` => `1.23`
+* `formatMoney`: Formats an amount of money (which are always stored as integers in Vendure) as a decimal, e.g. `123` => `1.23`.
+  * Also accepts two additional parameters (`currency` and `locale`), which will format according to the given currency and locale. For example:
+    * `{{ formatMoney 123 'USD' 'en-US' }}` => `$1.23`
+    * `{{ formatMoney 123 'EUR' 'de-DE' }}` => `1,23 €`
 * `formatDate`: Formats a Date value with the [dateformat](https://www.npmjs.com/package/dateformat) package.
 
 ## Extending the default email handlers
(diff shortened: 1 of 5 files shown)


StampixSMO:
MacOS CI failing doesn't look to be related to this PR

michaelbromley:
Thanks, this is a long-overdue feature! 

Everything looks good - nice work on the docs and tests 👍

I think it would also make sense to update the [default order confirmation template](https://github.com/vendure-ecommerce/vendure/blob/15f9b44fba0c31aa8ad932ec268b33f4a6925e52/packages/email-plugin/templates/order-confirmation/body.hbs#L104) to use the correct currencyCode too. Would you like to include that in this PR too?

StampixSMO:
> I think it would also make sense to update the [default order confirmation template](https://github.com/vendure-ecommerce/vendure/blob/15f9b44fba0c31aa8ad932ec268b33f4a6925e52/packages/email-plugin/templates/order-confirmation/body.hbs#L104) to use the correct currencyCode too. Would you like to include that in this PR too?

Yep, missed that. Done!

michaelbromley:
The test output indicates that the `onSend` mock function is not being called as expected for some reason. This part:

```
onSend.mock.calls[0][0].body
```
means "the first argument of the first call to `onSend`".

StampixSMO:
> The test output indicates that the `onSend` mock function is not being called as expected for some reason. This part:
> 
> ```
> onSend.mock.calls[0][0].body
> ```
> 
> means "the first argument of the first call to `onSend`".

Yep, my bad, I thought that issue was not due to my changes but it definitely was. The helper function threw an error and that was silently failing. All good now!

StampixSMO:
> > The test output indicates that the `onSend` mock function is not being called as expected for some reason. This part:
> > ```
> > onSend.mock.calls[0][0].body
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > means "the first argument of the first call to `onSend`".
> 
> Yep, my bad, I thought that issue was not due to my changes but it definitely was. The helper function threw an error and that was silently failing. All good now!

E2E tests seem to throw `Error: listen EADDRINUSE: address already in use :::3012`, which I'm pretty sure is not related to this PR?

michaelbromley:
> EADDRINUSE: address already in use :::3012

Yes that's just an intermittent flakiness of the e2e tests when 2 test suites running in parallel try to listen on the same port. This all looks good now, thanks for your contribution!

Issue #2413, opened 2023-09-26
configurable `ui-language-switcher-dialog.component.ts`


HoseinGhanbari:
**Is your feature request related to a problem? Please describe.**
some default values inside `ui-language-switcher-dialog.component.ts` are not configurable.

**Describe the solution you'd like**
I need to set/overwrite default value for `selectedCurrencyCode` & `currentLocale ` and also limit the items inside `availableLocales` array.

**Describe alternatives you've considered**
 The currency code used (USD) could be already fetched from the current channel's default - that would be a sensible default behaviour. The currentLocale can already be set using
```ts
AdminUiPlugin.init({
  adminUiConfig: {
    defaultLocale: 'XX',
  }
})
```

**Additional context**
By the way it is it really useful if we could customize the default values mentioned above like the ones inside vendure-config or any other way. Also note that when you target specific country or region for an e-commerce you are probably targeting the currency, locale, language etc so it would be nice to reflect these targets on the backend and admin-ui and we should not always consider whether the user browser has the right locale or has visited us before or not.

Iamshankhadeep:
@michaelbromley should I pick this up?

michaelbromley:
Hi @Iamshankhadeep - sorry I missed your message.

Yes, you are welcome to submit a PR for this issue 👍

PR #2550, opened 2023-11-27
configurable `ui-language-switcher-dialog.component.ts`

HoseinGhanbari:
# Description

Addressing the requirements described here https://github.com/vendure-ecommerce/vendure/issues/2413

# Breaking changes

⭕There might be a chance that `getAppConfig().availableLocales` don't filled with default value called `defaultAvailableLocales` when user didn't set any specific `availableLocales` on their `AdminUiConfig`.

⭕Actually currently I couldn't verify 100% that this is going to break current users admin-ui as they probably have not set any `availableLocales` but this might be the only breaking change.

# Screenshots

currently nothing applicable as a screenshot


HoseinGhanbari:
Also worth noting that during development session of this PR I figured out in some cases (at least during development) that if users didn't set `availableLanguages` OR `availableLocales` on **their configs** (vendure-ui-config.json), a method of `admin-ui-plugin` called `getAdminUiConfig` which is responsible for returning a complete config object is not using default values provided by `propOrDefault` helper.

I don't know if this is right to set configs inside `vendure-ui-config.json` during a development session or not but this **undefined value bug for default values** might not the case in production builds or official Vendure's releases.

packages/admin-ui/src/lib/core/src/components/app-shell/app-shell.component.ts
```diff
@@ -70,17 +70,23 @@ export class AppShellComponent implements OnInit {
...
             .pipe(
                 take(1),
-                switchMap(([currentLanguage, currentLocale]) =>
-                    this.modalService.fromComponent(UiLanguageSwitcherDialogComponent, {
+                switchMap(([currentLanguage, currentLocale]) => {
+                    const defaultLocale = getAppConfig().defaultLocale;
+                    const modalCurrentLocale = currentLocale ?? defaultLocale;
+
+                    const defaultLanguage = getAppConfig().defaultLanguage;
+                    const modalCurrentLanguage = currentLanguage ?? defaultLanguage;
```
michaelbromley: I don't understand this part - 

The setting of the `currentLanguage` is already handled here: https://github.com/vendure-ecommerce/vendure/blob/08ad982c6ef31fd53c5d95937c83ae8e313b9f32/packages/admin-ui/src/lib/core/src/core.module.ts#L94-L97

and the currentLocale is set here: https://github.com/vendure-ecommerce/vendure/blob/2cf8f094a98c65274dd4eb0bc26f54c84301bc16/packages/admin-ui/src/lib/core/src/data/client-state/client-defaults.ts#L8

why does this need to be repeated here?

packages/admin-ui/src/lib/core/src/components/ui-language-switcher-dialog/ui-language-switcher-dialog.component.ts
```diff
@@ -1,283 +1,60 @@
...
-        'EH',
-        'YE',
-        'ZM',
-        'ZW',
-        'AX',
-    ];
+    availableLocales: string[] = getAppConfig().availableLocales;
     availableCurrencyCodes = Object.values(CurrencyCode);
-    selectedCurrencyCode = 'USD';
+    selectedCurrencyCode;
```
michaelbromley: This should have a type `string`

packages/admin-ui/src/lib/static/vendure-ui-config.json
```diff
@@ -23,11 +24,264 @@
         "uk",
         "it",
         "fa",
-        "ne",
-        "hr"
+        "en"
+    ],
+    "availableLocales": [
```
michaelbromley, line 30: Rather than listing all the locales here, I think it would be better to default to an empty array, and adjust the logic in other parts so that if the array is empty, the full list from the shared constant is used. And then only if the developer wants to limit to specific locales, they then need to fill values in this array.

packages/common/src/shared-types.ts
```diff
@@ -283,6 +284,11 @@ export interface AdminUiConfig {
      * An array of languages for which translations exist for the Admin UI.
      */
     availableLanguages: LanguageCode[];
+    /**
+     * @description
+     * An array of locales to be used on Admin UI.
+     */
```
michaelbromley, line 293: This will go in the next minor release, so add this:
```
@since 2.2.0
```

michaelbromley (changes requested):
Overall looks good - I have a few points of feedback to address. Also, please make this PR against the `minor` branch as it is adding new functionality.

packages/admin-ui/src/lib/core/src/components/ui-language-switcher-dialog/ui-language-switcher-dialog.component.ts
HoseinGhanbari: done ✅

packages/common/src/shared-types.ts
HoseinGhanbari: done ✅

packages/admin-ui/src/lib/static/vendure-ui-config.json
HoseinGhanbari: done ✅
Actually did that because of non-empty `availableLanguages` array also I need to recheck the default values initialization mechanism and make sure that everything is going to run as expected.

HoseinGhanbari: I figured out that if we make `availableLanguages` or `availableLocales` empty, the admin-ui angular development mode (http://localhost:4200/admin/) will be broken with the following error message as nothing can be loaded as availableLanguages or availableLocales.

> sample error message
```
Error: The defaultLanguage "en" must be one of the availableLanguages []
```

But if we build the project and access the admin-ui (http://localhost:3000/admin/), there will be a new file created with default values filled and the admin-ui is working as expected.

> static/sample source-control tracked file
```
packages\admin-ui\src\lib\static\vendure-ui-config.json
```

> admin-ui-plugin generated file
```
packages\admin-ui-plugin\lib\admin-ui\vendure-ui-config.json
```

HoseinGhanbari: so if you are ok with this, we can pre-fill `availableLanguages` & `availableLocales` inside source-control tracked version of `vendure-ui-config.json` to make development mode of admin-ui work as expected and if this is not the desired change let me know and we can find a better solution.

packages/admin-ui/src/lib/core/src/components/app-shell/app-shell.component.ts
HoseinGhanbari: done ✅
Yeah, you are right 👌
calculation and assignment of  `modalCurrentLocale` & `modalCurrentLanguage` removed

packages/admin-ui/src/lib/static/vendure-ui-config.json
HoseinGhanbari: I just temporarily pre-filled those keys of `vendure-ui-config.json` to make sure that we are not breaking development mode of admin-ui, but I'm available if there are possible better solution that I can help.

michaelbromley:
Thank you for your excellent work on this!

HoseinGhanbari:
Anytime 🙏 I'm glad it was helpful !

Issue #2523, opened 2023-11-13
Refund always send `amount===0` to payment method `createRefund` function


alexisvigoureux:
**Describe the bug**

We are on version 2.1.1 of Vendure and we have the same problem as [here](https://github.com/vendure-ecommerce/vendure/issues/2302)

The refund amount sent to the `createRefund` method is equal to 0.

Wouldn't the [orderPlacedQuantity](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/services/payment.service.ts#L306) field already be 0?

![Capture d’écran 2023-11-13 à 11 05 50](https://github.com/vendure-ecommerce/vendure/assets/6134849/2e025d6c-6b40-4f9d-90de-f33340543fcd)
![Capture d’écran 2023-11-13 à 11 44 07](https://github.com/vendure-ecommerce/vendure/assets/6134849/6298e954-495c-4a67-8407-87f1fc4e9fe2)

**To Reproduce**
See https://github.com/vendure-ecommerce/vendure/issues/2302 and select "cancel order" in addition

**Expected behavior**
Send amount before cancellation

**Environment (please complete the following information):**
 - @vendure/core version: 2.1.1
 - Nodejs version: 18.18
 - Database (mysql/postgres etc): postgres

Issue #2302, opened 2023-07-20
[V2.0.4] Refund always send `amount===0` to payment method `createRefund` function


tianyingchun:
**Describe the bug**
Refund always send `amount===0` to payment method `createRefund` function

**To Reproduce**
<img width="1137" alt="image" src="https://github.com/vendure-ecommerce/vendure/assets/2833251/9839c93c-cb16-4d18-b01f-5ddd15d70df2">

2. choose check box `refund shipping (0.00)`
<img width="271" alt="image" src="https://github.com/vendure-ecommerce/vendure/assets/2833251/d19454c9-29e3-4c67-a913-4e758691c33c">

3.  in payment method 
```ts
   createRefund: async (ctx, input, amount, order, payment, args) => {
        const transactionId = payment.transactionId;
        try {
....
```
we always received `amount` equal to `zero`

**Expected behavior**
should correct return refund `amount`

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.4
 - Nodejs version 18
 - Database (mysql/postgres etc):mysql

**Additional context**

PR #2559, opened 2023-11-30
fix(core): send the correct amount to `refundOrder` on cancelled OrderLines

alexisvigoureux:
# Description

Fix issue https://github.com/vendure-ecommerce/vendure/issues/2523 related to https://github.com/vendure-ecommerce/vendure/issues/2302

Issue: The amount received by the `refundOrder` function is 0
After investigation, `orderPlacedQuantity` is always 0 [here](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/services/payment.service.ts#L306)

As explained in the [documentation](https://docs.vendure.io/guides/core-concepts/orders/#responding-to-a-state-transition
), the order is persisted later, and overrides the value set in each orderLine update.
So, this PR defines the correct value for `orderPlacedQuantity` of each orderLines when the order is placed to have the right value at persistence.

Work done in pairs with @ttournie 🙏 

# Screenshots

Capture d’écran 2023-11-30 à 17 59 33

# Checklist

:pushpin: Always:

:zap: Most of the time:

### DIFF
--- a/packages/core/src/config/order/default-order-process.ts
+++ b/packages/core/src/config/order/default-order-process.ts
@@ -412,13 +412,7 @@ export function configureDefaultOrderProcess(options: DefaultOrderProcessOptions
                 if (shouldSetAsPlaced) {
                     order.active = false;
                     order.orderPlacedAt = new Date();
-                    await Promise.all(
-                        order.lines.map(line =>
-                            connection
-                                .getRepository(ctx, OrderLine)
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi, thanks for the PR!

It looks like this change is causing some e2e test failures though - can you look into that?

michaelbromley:
Yeah the e2e tests flake out sometimes on port conflicts. Thanks for this impactful fix!

PR #2566, opened 2023-12-04
fix(core): Fix loading multiple customField relations

### LINKED ISSUE #2555: ListQueryBuilder parallelLoadRelations does not handle multiple customFields correctly
**Describe the bug**
When querying multiple entities with `customFields` `ListQueryBuilder.parallelLoadRelations` is being called. The method loads relations that are part of the original GraphQL query to make sure they don't need to be fetched sequentially. When querying multiple `customFields` relations only the first one will be populated due to the way the fetched `relationPaths` are being handled.

The issue seems to be in the area:
https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/helpers/list-query-builder/list-query-builder.ts#L486C9-L499C35

`results.map(r => ({ relation: relationPaths[0] as keyof T, entity: r })),` only writes the first relation with a common prefix to the `entitiesIdsWithRelations` array which is fine if the rest of the relation path are nested relations that are already part of the quer result.
In case of `customFields` the `relationsPaths` might look similar to this:
```json
["customFields.relationA", "customFields.relationB"]
```

`entitiesIdsWithRelations` will only contain `customFields.relationA` and thus only populate `customFields.relationA` event though `customFields.relationB` has been fetched as well.

**To Reproduce**
1. Create at least two customFields with relations
2. Use a list query and fetch both relations simultaneously
3. Inspect the database queries executed to resolve the second relation in customFields

**Expected behavior**
The relation is not fetched on a per-parent basis but in a single query that fetches all children of all parent entities at once.

 - @vendure/core version: 2.1.2
 - Nodejs version: v18.17.0

**Additional context**
A potential fix would be to write all relation paths into `entitiesIdsWithRelations` and add a loop during the assignment. Please let me know if you would like to accept a PR on this.

jnugh:
fixes #2555

# Description
If a list query contained multiple `customFields` relations, only one had been assigned to the target entity. This fix allows storing multiple relations in `customFields` to be fetched.

# Breaking changes
None

# Checklist

:pushpin: Always:

:zap: Most of the time:

### DIFF
--- a/packages/core/e2e/fixtures/test-plugins/list-query-plugin.ts
+++ b/packages/core/e2e/fixtures/test-plugins/list-query-plugin.ts
@@ -39,9 +39,25 @@ export class CustomFieldRelationTestEntity extends VendureEntity {
     parent: Relation<TestEntity>;
 }
 
+@Entity()
+export class CustomFieldOtherRelationTestEntity extends VendureEntity {
+    constructor(input: Partial<CustomFieldOtherRelationTestEntity>) {
+        super(input);
+    }
+
+    @Column()
+    data: string;
+
+    @ManyToOne(() => TestEntity)
+    parent: Relation<TestEntity>;
+}
+
 class TestEntityCustomFields {
     @OneToMany(() => CustomFieldRelationTestEntity, child => child.parent)
     relation: Relation<CustomFieldRelationTestEntity[]>;
+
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks, this is an excellent contribution!

PR #2568, opened 2023-12-05
feat(core): add the possibility of activating NestJS devtools

alexisvigoureux:
# Description

Add the possibility of activating [NestJS devtools](https://docs.nestjs.com/devtools/overview) in Vendure.
This will enable developers to analyse the bootstrap performance of their applications. This becomes very interesting when application have a large number of modules and plugins to find a bottleneck

> [!NOTE]
>  Activation is enabled via env vars. I am aware that it is not the best way, if you have any suggestions for doing it differently

# How to test
- In `packages/dev-server/index.ts` add 
```
bootstrap(devConfig, { nestApplicationOptions: { snapshot: true }})
...
```
- Launch the server `ENABLE_DEVTOOLS=true DB=postgres yarn start`
- Go https://devtools.nestjs.com/ (it's not free...)

# Breaking changes
No

# Screenshots
image

# Checklist

:pushpin: Always:

:zap: Most of the time:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -44,6 +44,7 @@
         "@nestjs/apollo": "^12.0.7",
         "@nestjs/common": "10.2.1",
         "@nestjs/core": "10.2.1",
+        "@nestjs/devtools-integration": "0.1.5",
         "@nestjs/graphql": "12.0.8",
         "@nestjs/platform-express": "10.2.1",
         "@nestjs/terminus": "10.0.1",
--- a/packages/core/src/app.module.ts
+++ b/packages/core/src/app.module.ts
@@ -12,6 +12,7 @@ import { I18nService } from './i18n/i18n.service';
 import { PluginModule } from './plugin/plugin.module';
 import { ProcessContextModule } from './process-context/process-context.module';
 import { ServiceModule } from './service/service.module';
+import { DevtoolsModule } from "@nestjs/devtools-integration";
 
 @Module({
     imports: [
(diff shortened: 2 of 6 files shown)


michaelbromley:
Hi,

Thanks for the suggestion & PR. I haven't tried the NestJS devtools but it looks very nice.

I _would_ like to add support, but I'd rather find a way that does not require us to add a direct dependency and couple our config to the implementation if possible.

From what I can see, the only thing preventing anyone from integrating the devtools is that we do not expose a way to modify the options passed to the `NestFactory.create()` method.

Let's imagine that in future, the Nest team decide to add more options that can be passed to the `create()` method - we would then need to expand our own VendureConfig interface to accommodate this.

I think a more general solution, which would cover this PR and other cases too, would be to add a second argument to the `bootstrap` and `bootstrapWorker` functions which allow custom options to be passed to the `NestFactory.create()` method.

In this way, integration with the devtools would be trivial and require no other changes to core.

alexisvigoureux:
Hi, in fact it's a better solution for maintaining the code over time. I made the change and update the "how to test" part

michaelbromley:
Yes this is much neater, I like it.

One final thing: I think I'd prefer to have the second argument be our own object that could potentially be extended in the future, like this:

```ts
interface BootstrapOptions {
  nestApplicationOptions: NestApplicationOptions;
}

interface BootstrapWorkerOptions {
 ...
}
```

because based on long experience it is quite likely that we eventually think "oh wouldn't it be great if we could also configure _this_ aspect of the bootstrap process", at which point we have an obvious place to do so.

alexisvigoureux:
Yes of course, it's done

michaelbromley:
Thank you!

PR #2571, opened 2023-12-07
chore: Update README.md

chrislaai:
Now that we no longer have to do this, I get no errors on my first build.


michaelbromley:
Thanks Chris 👍

PR #2582, opened 2023-12-16
chore(admin-ui): Login page dark mode color tweak

chrislaai:
# Description

Currently, when we log in and select dark mode and then log out, we do not have a corresponding dark mode login interface. This commit fixes it.

# Screenshots
Before:
image
After:
image


michaelbromley:
Thank you!

PR #2583, opened 2023-12-16
chore(admin-ui): Version info placement tweak

chrislaai:
* Delete the version and brand info under the login button.

* Add version info in `left-nav`.

# Description

The login page already has a lot of brand-related information, and the brand information under the login button seems unnecessary.
I saw elsewhere that hiding the version number is good for security, so should we delete it from the login page by default? In addition, I put it in `left-nav`, and we can still use `hideVersion` to hide it.

# Screenshots

Before:
image

After:
image
image


michaelbromley:
I think that layout looks good now. Last thing: I think that the devmode popup should always show the version. Let's say you have your admin ui configured to hide version info. In devmode I think we still want a convenient way to check the version, regardless of that setting.

michaelbromley:
Thanks!

PR #2584, opened 2023-12-16
chore(admin-ui): Tab-bar color tweak

chrislaai:
# Description

Vendure blue is great so I did a little color tweak to keep us with this clean feel, without too much color.

# Screenshots

Before:
image

After:
image

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/page-header-tabs/page-header-tabs.component.scss
+++ b/packages/admin-ui/src/lib/core/src/shared/components/page-header-tabs/page-header-tabs.component.scss
@@ -5,6 +5,11 @@
 
 .tab-bar {
     display: flex;
+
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hi Chris - thanks for the suggestion. I do think it looks good using the brighter blue but my main concern is the contrast ratio for accessibility.

With the current color of `--color-primary-800`, we have in the nav bar a contrast ratio of 6.2.
If we change it to `--color-primary-600` then this drops to 2.31

image
image

To me that is more important than the aesthetics of it.

So I would say we should leave that active color as-is. However, I do think that the `:hover` style for the tab is a great addition - the current hover color is totally wrong.

Could you amend the commit to keep just the change to the tab hover style, and revert the `--color-text-active` change?

PR #2586, opened 2023-12-17
fix(asset-server-plugin): Add the missing `format` attribute

chrislaai:
Add the missing `format` attribute to the `ImageTransformPreset` interface, and make `name`, `width`, and `height` optional.

### DIFF
--- a/packages/asset-server-plugin/src/plugin.ts
+++ b/packages/asset-server-plugin/src/plugin.ts
@@ -107,7 +107,7 @@ import { AssetServerOptions, ImageTransformPreset } from './types';
  * AssetServerPlugin.init({
  *   // ...
  *   presets: [
- *     { name: 'my-preset', width: 85, height: 85, mode: 'crop' },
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hi, I don't think this is the case. If we look at the code that actually uses the presets:

https://github.com/vendure-ecommerce/vendure/blob/5a0cbe61f3297119368c6aec76249e286760249b/packages/asset-server-plugin/src/transform-image.ts#L24-L29

we can see that only the `width`, `height`, and `mode` are used. 

I guess they technically _are_ optional, since if a width or height is not specified, the transform function will do something sensible as a default.

> you will get an error if you don't add format

What error do you get?

michaelbromley:
Yes, the errors you report are expected and in line with the intended usage of this feature. If you just want to specify "format" and "mode", you should just put the desired values in the query string when constructing your image src url from your client app.

I'm going to close this as "working as intended".

PR #2632, opened 2024-01-20
fix(core): Ensure deterministic sorting in case of duplicates

hans-rollingridges-dev:
# Description

Default search plugin fails to return deterministic results in case of sorting products with duplicate name, price or score results.

This behavior is observed using the Postgres search strategy.

Sorting by price with price duplicates, and pagination shows the search results of consecutive pages, i.e. skip and take queries are non deterministic.

This is perceived in the storefront as "duplicate or missing" products when viewing consecutive search result pages.

Solution:
Always ensure to finally sort by product variant id.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Amazing PR, thank you!

Issue #2363, opened 2023-08-25
Not Authorized error with External Authentication Strategy


Swampy469:
**Describe the bug**
I am using a custom class that implements "AuthenticationStrategy" interface for a custom authentication flow.
After the first access (when there is no ChannelToken saved in the browser) each initial APIs call return 403 error code.
I find out that, even if the ChannelToken is saved (after the login), the initial  (first 5 requests) APIs header requests doesn't contains "Vendure-Token".
After reloading the page (CTRL + R / F5) all is fine, each request contains the Vendure-Token as header.

<img width="907" alt="image" src="https://github.com/vendure-ecommerce/vendure/assets/46132382/61f49000-7b2a-40ae-bb23-2c0f671bfac7">

<img width="913" alt="image" src="https://github.com/vendure-ecommerce/vendure/assets/46132382/483fd6da-f0fa-43a9-8ba1-70ace1513082">
The first 5 requests after loading doen't contains the "Vendure-Token" inside the headers

**To Reproduce**
1. Make sure there is no Vendure data in the browser (Token, sessions, etc..)
2. Access the Admin UI using an External Authentication strategy
3. Go to any section
4. See "not authorized" error

**Expected behavior**
You must be authorized immediatly after the first login with an external authentication strategy.

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.9
 - Nodejs version: 18.16.1
 - Database (mysql/postgres etc): postgres

**Additional context**
Installed plugins:
- MultiVendor
- AssetServer
- Email

oliverqx:
I'm still working on it, will have a solution by eod tomorrow

michaelbromley:
Listing duplicates:

- https://github.com/vendure-ecommerce/vendure/issues/2580
- https://github.com/vendure-ecommerce/vendure/issues/1427

oliverqx:
took a bit longer than I thought cause i didnt have much free time this week

Issue #2580, opened 2023-12-15
Keycloak authentication - cannot mantain session


margamorais:
I was testing sso authentication for admins using the same code as in the [keycloak test plugin](https://github.com/vendure-ecommerce/vendure/tree/master/packages/dev-server/test-plugins/keycloak-auth).

On the first login it's creating the user. And after that I can retrieve the user, however when redirecting to '/admin' It's redirecting back to the login page because it's not mantaining the session.

What could be missing?

margamorais:
Hi there, Is there any more information we can provide to help assess this issue? @michaelbromley

michaelbromley:
Hi,
I have a vague idea of running into this in the past from someone else. I remember looking into it briefly and concluding that we need to make some change to properly support this in the way the Admin UI bootstraps. I cannot remember all the details right now.

I'm adding this to the backlog and will investigate further at the next opportunity. 

If you want to do so yourself in the mean time and share any findings or even open a PR, that will speed things up.

michaelbromley:
Closing in favour of https://github.com/vendure-ecommerce/vendure/issues/2363

michaelbromley:
Actually this is a distinct issue, and I think it is due to the fact that cookies/localStorage are scoped to a specific domain, so when you have the login page on a different domain, then redirect to the Admin UI app, the session token is not transferred.

PR #2636, opened 2024-01-22
fix(external-auth-strategy): Client implementation updated. (#2363)

oliverqx:
/claim #2363

# Description

External authentication strategy must set activeChannelToken on local storage if its missing to avoid admin ui initiation bugs.

# Breaking changes

No breaking changes.

# Screenshots

Its quite complex for just screenshots so I recorded a short video showing my solution:

https://www.loom.com/share/496c82c056034f8f87f1d1ce2affa4e8?sid=8cbb24d9-3a1b-4604-861c-893ed3164840

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dev-server/docker-compose.yml
+++ b/packages/dev-server/docker-compose.yml
@@ -84,6 +84,18 @@ services:
       - pgadmin_data:/var/lib/pgadmin
     links:
       - "postgres:pgsql-server"
+  keycloak:
+    image: quay.io/keycloak/keycloak
+    ports:
+      - "9000:8080"
+    environment:
+      KEYCLOAK_ADMIN: admin
+      KEYCLOAK_ADMIN_PASSWORD: admin
+    command:
+      - start-dev
+      - --import-realm
+    volumes:
+      - keycloak_data:/opt/keycloak/data
 volumes:
   postgres_data:
     driver: local
@@ -94,4 +106,6 @@ volumes:
   mysql_data:
     driver: local
   phpmyadmin_data:
+    driver: local
+  keycloak_data:
     driver: local
--- a/packages/dev-server/test-plugins/keycloak-auth/keycloak-auth-plugin.ts
+++ b/packages/dev-server/test-plugins/keycloak-auth/keycloak-auth-plugin.ts
@@ -1,3 +1,4 @@
+import { HttpModule } from '@nestjs/axios';
 import { MiddlewareConsumer, NestModule } from '@nestjs/common';
 import { PluginCommonModule, VendurePlugin } from '@vendure/core';
 import express from 'express';
@@ -17,7 +18,7 @@ import { KeycloakAuthenticationStrategy } from './keycloak-authentication-strate
  * Video demo of this: https://youtu.be/Tj4kwjNd2nM
  */
 @VendurePlugin({
-    imports: [PluginCommonModule],
+    imports: [PluginCommonModule, HttpModule],
     configuration: config => {
         config.authOptions.adminAuthenticationStrategy = [
             ...config.authOptions.adminAuthenticationStrategy,
--- a/packages/dev-server/test-plugins/keycloak-auth/keycloak-authentication-strategy.ts
+++ b/packages/dev-server/test-plugins/keycloak-auth/keycloak-authentication-strategy.ts
@@ -5,7 +5,8 @@ import {
     Injector,
     Logger,
     RequestContext,
-    RoleService,
+    Role,
+    TransactionalConnection,
     User,
 } from '@vendure/core';
 import { DocumentNode } from 'graphql';
@@ -29,13 +30,13 @@ export class KeycloakAuthenticationStrategy implements AuthenticationStrategy<Ke
     readonly name = 'keycloak';
     private externalAuthenticationService: ExternalAuthenticationService;
     private httpService: HttpService;
-    private roleService: RoleService;
+    private connection: TransactionalConnection;
     private bearerToken: string;
 
     init(injector: Injector) {
         this.externalAuthenticationService = injector.get(ExternalAuthenticationService);
         this.httpService = injector.get(HttpService);
-        this.roleService = injector.get(RoleService);
+        this.connection = injector.get(TransactionalConnection);
     }
 
     defineInputType(): DocumentNode {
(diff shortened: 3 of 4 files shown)


oliverqx:
# The issue

When logging in using an external authentication strategy, the vendure-token request header is not set. This makes important component initialization queries fail and requires end-users to reload the ui.  Ideally after authentication, reloading isnt necessary to see admin's data.

# Default vs External Authentication Strategy

So when logging in from the default login component, admin ui queries look like this:

```
✅ GetServerConfig (vendure-token set)
✅ GetCurrentUser (vendure-token set)
✅ GetActiveAdministrator (vendure-token set)
✅ GetGlobalSettings (vendure-token set)
✅ GetActiveChannel (vendure-token set)
✅ GetOrderSummary (vendure-token set)
✅ GetLatestOrders (vendure-token set)
✅ GetOrderChartData (vendure-token set)
✅ GetPendingSearchIndexUpdates (vendure-token set)

*go to products route*

✅ ProductListQuery (vendure-token set)
```

When logging in from an external authentication strategy component, admin ui queries look like this:

```
❌ GetServerConfig (no vendure-token set)
✅ GetCurrentUser (no vendure-token set)
        - authentication service runs authenticateToken which sets activeChannelToken using the localStorageService.
✅ GetActiveAdministrator (vendure-token set)
✅ GetGlobalSettings (vendure-token set)
...etc

*go to products route*

❓ ProductListQuery request is not sent at all.

*reload*

✅ GetServerConfig (vendure-token set)
✅ GetCurrentUser (vendure-token set)
...etc

*go to products route*

✅ ProductListQuery (vendure-token set)
```

External authentication strategy queries look as shown above only upon first login. Because on second login local storage activeChannelToken is set.

This means that:
1.  Default login component sets activeChannelToken upon successful login:
    - [login function calls authService.logIn](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/login/src/components/login/login.component.ts#L51)
    - [authService.login calls setChannelToken upon successful authentication](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/providers/auth/auth.service.ts#L36)
    - [authService.setChannelToken calls localStorageService.set('activeChannelToken', user.activeChannel)](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/providers/auth/auth.service.ts#L147)
3. GraphQL queries' Vendure-Token header gets set [set to the activeChannelToken saved on local storage or not at all](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/data/data.module.ts#L68).
4. Upon receiving GetCurrentUser, authentication service runs authenticateToken [which sets activeChannelToken](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/providers/auth/auth.service.ts#L109).

# Fix

Admin UI is configured to receive an activeChannelToken saved in localstorage. Set activeChannelToken before redirecting to /admin. 

The GraphQL authenticate mutation already returns the users channels:

Screenshot 2024-01-21 at 8 00 31 PM

So we can use that to make our custom client set the localStorage activeChannelToken:
```
mutation Authenticate($token: String!) {
                        authenticate(input: {
                          google: { token: $token }
                        }) {
                            ...on CurrentUser {
                                id
                                channels {
                                 code
                                }
                            },
                      }
                    }
                `,

localStorage.setItem('vnd__activeChannelToken', '"' + res.data.autheticate.channels[0].token + '"')
```

michaelbromley:
Hi,

Thank you for taking the time to look into this, and especially the video explainer - that was really helpful.

I'm still evaluating the status of this solution, as I feel like there may be a cleaner solution that manually manipulating the localstorage. I believe that solution will still fail if the the Google login page is on a different domain.

I'm also investigating this somewhat-related issue:

- https://github.com/vendure-ecommerce/vendure/issues/2580

Both that issue and this may have some common solution, but I'm still investigating.

oliverqx:
No worries. I totally understand, I also thought it was a bit weird to manually manipulate the local storage. But its actually the reason why vendure's default login page works. You can see it here:

1.  Default login component sets activeChannelToken upon successful login:
    - [login function calls authService.logIn](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/login/src/components/login/login.component.ts#L51)
    - [authService.login calls setChannelToken upon successful authentication](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/providers/auth/auth.service.ts#L36)
    - [authService.setChannelToken calls localStorageService.set('activeChannelToken', user.activeChannel)](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/providers/auth/auth.service.ts#L147)  

Both the authenticationService and localStorageService are actually running on the client.

The admin UI is expecting that local storage token and using it to add context to all GraphQL API calls. As [seen here](https://github.com/vendure-ecommerce/vendure/blob/master/packages/admin-ui/src/lib/core/src/data/data.module.ts#L68)

I've thought about this for some time and arrived at two solutions that would stop the client from needing to use localStorage: 

1. Adding a middleware to the GraphQL API endpoints and having that middleware look for the user's channel and add it to the headers. IMO this is suboptimal because it creates extra work on the server, and even though it could be cached, there are edge cases that wouldn't be able to cache it properly like serverless vendure. I normally go serverless with my work so I know I would always try to avoid server-side workload when its possible, this middleware change would incur in bandwith, computation and performance charges because its adding an extra db call to every GraphQL API query. 

2. Adding a cookie with activeChannel. It would be set and updated by the server.  I think this is way better. It doesn't add a significant amount of work to the server and its cleaner as the client wouldn't be required to update or set localStorage tokens anymore.

I can work on whichever solution you think its best. I would personally choose the second solution.

Also with regards to issue #2580, this is caused by cookies not being shared between domains and its expected behavior. Cookies are almost always same site (especially auth cookies) to make them as secure as possible, so domain A can not set cookies for doman B. This is enforced by most browsers. However subdomain A can set cookies for subdomain B. I have a live app that works like this, but there are some security concerns that must be addressed properly. I can look into drafting something that enables subdomain cookies if you want to take this route.

Issue #2171, opened 2023-05-08
customFields null on graphQl query collection


getwebM:
**Describe the bug**
customFields is null on graphQl query collection

Product with customFields:
![image](https://user-images.githubusercontent.com/53661314/236754856-0fab1ba8-92c6-4d40-9036-df6dbcb92363.png)

Query:
![image](https://user-images.githubusercontent.com/53661314/236754939-70542a0c-2218-4726-bc3b-32dc7796779a.png)

Response:
![image](https://user-images.githubusercontent.com/53661314/236755072-21891ed2-6be4-4443-aff1-40fc4b4b7e48.png)

**To Reproduce**
1. Create a product with custom Fields
2. Add Product to collection
3. Query Collection with productVariants
4. See customFields are empty

**Expected behavior**
List of customFields as JSON

**Environment (please complete the following information):**
 - @vendure/core version: 1.9.5
 - Nodejs version: 14.21.2
 - Database (mysql/postgres etc): 10.5.19-MariaDB-0+deb11u2 - Debian 11

michaelbromley:
I think the problem here (in both cases) is that the custom fields are being defined on the `Product` entity, but both the `collection` and (I am guessing) your `order` queries are actually querying `ProductVariant` rather than `Product`.

For example, to get the custom fields of the `Product` from the `activeOrder` query you would do:

```gql
query {
  activeOrder {
    id
    lines {
      productVariant {
        product {
          customFields {
            myCustomField
          }
        }
      }
    }
  }
}
```

daviddarx:
Hi @michaelbromley , 

Thanks for your quick feedback. Hum, I guess my query already was correct, see, if I query the product directly by ID, I get its `shortname`: 
<img width="952" alt="01_product" src="https://github.com/vendure-ecommerce/vendure/assets/10694809/92d5f2f6-81dc-4a2d-a47b-a7d5f48e095f">

On the other hand, If I have the product in my order lines with `order` or `orderByCode`, the `shortname` is there null. 
<img width="952" alt="02_order" src="https://github.com/vendure-ecommerce/vendure/assets/10694809/6daa2e28-db33-4510-84b7-bc72ae7143c5">
<img width="952" alt="03_orderByCode" src="https://github.com/vendure-ecommerce/vendure/assets/10694809/00488cde-c37e-4615-8622-6626446a10d4">

Am I missing something? 
Thanks in advance!

michaelbromley:
[...]

PR #2644, opened 2024-01-23
fix(product-variant-entity-resolver): Add product translation

oliverqx:
/claim #2171

# Description

The product variant entity resolver is not translating the product when its given productVariant argument has product already defined. 

productVariant [gets its product property from a call to listQueryBuilder](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/services/product-variant.service.ts#L214), which means its not translated.

(sorry about the double pr, wont happen again!)

# Breaking changes

No breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/resolvers/entity/product-variant-entity.resolver.ts
+++ b/packages/core/src/api/resolvers/entity/product-variant-entity.resolver.ts
@@ -80,8 +80,13 @@ export class ProductVariantEntityResolver {
         @Parent() productVariant: ProductVariant,
     ): Promise<Product | undefined> {
         if (productVariant.product) {
-            return productVariant.product;
+            return this.requestContextCache.get(
+                ctx,
+                `ProductVariantEntityResolver.product(${productVariant.productId})`,
+                () => this.productVariantService.translateProduct(ctx, productVariant.product),
+            );
         }
+
         return this.requestContextCache.get(
             ctx,
             `ProductVariantEntityResolver.product(${productVariant.productId})`,
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -364,6 +364,10 @@ export class ProductVariantService {
         return stockOnHand;
     }
 
+    translateProduct(ctx: RequestContext, product: Product) {
+        return this.translator.translate(product, ctx);
+    }
+
     async create(
         ctx: RequestContext,
         input: CreateProductVariantInput[],


packages/core/src/api/resolvers/entity/product-variant-entity.resolver.ts
```diff
@@ -80,8 +80,13 @@ export class ProductVariantEntityResolver {
         @Parent() productVariant: ProductVariant,
     ): Promise<Product | undefined> {
         if (productVariant.product) {
-            return productVariant.product;
+            return this.requestContextCache.get(
+                ctx,
+                `ProductVariantEntityResolver.product(${productVariant.productId})`,
+                () => this.productVariantService.translateProduct(ctx, productVariant.product),
```
michaelbromley: I don't think we need to create a brand new method `.translateProduct()` for this purpose. We can reuse `getProductForVariant()` which is already doing the translation step.

Which then would make both branches do the same thing, which also doesn't make sense. So there must be a logical branch in which this additional lookup/translation does not need to be done - that would be if the `product` is already translated. This can be tested by inspecting `productVariant.product.name` to see if it is defined. If so, we should be ok to just return it as-is like in the original version.

michaelbromley (changes requested):
See my comment on the resolver code. Also, do you think you'd be able to create a failing e2e test in [this suite](https://github.com/vendure-ecommerce/vendure/blob/9b90b961037b9bc921c1d6b6364cb948114582b5/packages/core/e2e/custom-fields.e2e-spec.ts) that demonstrates the bug & fix?

packages/core/src/api/resolvers/entity/product-variant-entity.resolver.ts
oliverqx: You're right, both branches are doing almost the same thing. I just pushed some changes. Also added the test.

To see the bug:

1. Check out commit [ac26c4a](https://github.com/vendure-ecommerce/vendure/pull/2644/commits/ac26c4a54b85a211d1efe950f72005b1f1d98802) 
2. `yarn build`
3. `yarn e2e`
4. test fails

Fix:

1. Checkout commit [368a9fc](https://github.com/vendure-ecommerce/vendure/pull/2644/commits/368a9fc1d325453fb99f5ef4b121b0a1046f38ce)
2. `yarn build`
3. `yarn e2e`
4. test does not fail

michaelbromley:
@oliverqx thank you! Can you contact me at m.bromley [at] vendure.io about the bounty payment?

oliverqx:
will do, thanks!

Issue #2652, opened 2024-01-29
Custom fields on ProductVariantPrice


HausTechTeam:
**Is your feature request related to a problem? Please describe.**
I would like to be able to connect other entities to ProductVariantPrice in order to implement custom pricing management. The use case is to enable customer-specific prices and campaign prices (discounted prices displayed in the product listing) via price lists, where the price list is connected to each individual variant price.

**Describe the solution you'd like**
Support for custom fields on ProductVariantPrice

**Describe alternatives you've considered**
Other entities that work separately but it seems redundant when Vendure already has support for multiple prices (in different currencies) that are also channel-aware.

HausTechTeam:
Here is a PR with an implementation of the requested feature: https://github.com/vendure-ecommerce/vendure/pull/2654

PR #2654, opened 2024-01-29
feat(core): Custom fields on ProductVariantPrice

HausTechTeam:
# Description

Support for custom fields on ProductVariantPrice.

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/config/custom-field/custom-field-types.ts
+++ b/packages/core/src/config/custom-field/custom-field-types.ts
@@ -165,6 +165,7 @@ export type CustomFields = {
     ProductOption?: CustomFieldConfig[];
     ProductOptionGroup?: CustomFieldConfig[];
     ProductVariant?: CustomFieldConfig[];
+    ProductVariantPrice?: CustomFieldConfig[];
     Promotion?: CustomFieldConfig[];
     Region?: CustomFieldConfig[];
     Seller?: CustomFieldConfig[];
--- a/packages/core/src/config/default-config.ts
+++ b/packages/core/src/config/default-config.ts
@@ -198,6 +198,7 @@ export const defaultConfig: RuntimeVendureConfig = {
         ProductOption: [],
         ProductOptionGroup: [],
         ProductVariant: [],
+        ProductVariantPrice: [],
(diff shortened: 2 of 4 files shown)


HausTechTeam:
Issue: https://github.com/vendure-ecommerce/vendure/issues/2652

michaelbromley:
Thank you!

HausTechTeam:
You're welcome :)

Do you happen to have a rough idea of when the next minor release might be available?

michaelbromley:
I'm aiming for February for the release, but before then I'll be doing a pre-release which will include this feature, probably by next week.

oidt:
I have a question about this. I can see the customfield on the admin-api after putting it in the config. ``ProductVariantPrice: [
            {
                name: 'costumerGroup',
                type: 'relation',
                entity: CustomerGroup,
            },
            { name: 'customerPrice', type: 'int' }
        ]``

but I can not find this on the shop-api. Neither I can find a input field on the admin-ui.

michaelbromley:
@oidt the `ProductVariantPrice` is not exposed to the Shop API by default.

And yes this PR does not include any Admin UI change that would add custom field inputs for the product variant price. Indeed, there is no GraphQL API part that even allows modification. As it stands the custom fields would need to be set in a custom mutation.

lzzfnc:
Hello! Sorry to bump into this, but what would it be the procedure to expose the admin ui for the custom fields? It is something doable with a plugin, or it requires changes to the core?

michaelbromley:
@lzzfnc A work-around solution would be to define a [Custom Detail Component](https://docs.vendure.io/guides/extending-the-admin-ui/custom-detail-components/) block on the product variant detail page as part of a UI extension, and in that block you can have a form which exposes the custom fields.

The main drawback here is that it will not be hooked up to the overall form save mechanism for the detail page, so changes made to the custom fields would need their own "save" button inside the custom detail component.

A more satisfactory solution would be to add built-in support which means changes to the admin ui package itself. You can open a separate feature request for this if you like.

PR #2657, opened 2024-01-31
feat(payments-plugin): Dont handle mollie webhook for any state after PaymentSettled

martijnvdbrug:
# Description

Occasionally, we receive webhooks from Mollie, while the Vendure order is already in state Shipped or Delivered. This results in a `ORDER_PAYMENT_STATE_ERROR`, because the order is already past its payment states. Even worse, a 500 is returned to Mollie, so Mollie will keep retrying this webhook.
While the root cause of these incoming hooks is not known, we should still not handle Mollie hooks when an order is already in PaymentSettled, Shipped or Delivered.  

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Thanks Martijn!

PR #2659, opened 2024-01-31
feat(core): add translations for new advanced search plugin version

oliverstreissi:
# Description

Add the error message translations for the new version of the advanced search plugin with the new facets API.

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Thanks for this, but thinking about it more, this approach will not be sustainable long-term. The core should have no knowledge of what translation strings are required by external plugins. 

There _is_ a mechanism for adding server-side translations for plugins, which is detailed here: https://docs.vendure.io/guides/developer-guide/translations/#server-message-translations

Let me know if this is sufficient, or if we need to make any other changes to core to make it easier to supply translations.

PR #2663, opened 2024-02-02
fix(core): update orderLine on applyPriceAdjustments

ankitdev10:
# Description

This patch fixes applyPriceAdjustment method for updated order lines.

# Breaking changes

NO

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1726,7 +1726,7 @@ export class OrderService {
             // a race condition where changing one or the other in parallel can
             // overwrite the other's changes.
             .save(omit(updatedOrder, ['shippingAddress', 'billingAddress']), { reload: false });
-        await this.connection.getRepository(ctx, OrderLine).save(updatedOrder.lines, { reload: false });
+        await this.connection.getRepository(ctx, OrderLine).save(updatedOrderLines, { reload: false });
         await this.connection.getRepository(ctx, ShippingLine).save(order.shippingLines, { reload: false });
         await this.promotionService.runPromotionSideEffects(ctx, order, activePromotionsPre);
 


JamesLAllen:
I'm not sure how this is any different from how it currently works, is it returning a different result?

```const updatedOrder = await this.orderCalculator.applyPriceAdjustments(
            ctx,
            order,
            promotions,
            updatedOrderLines ?? [],
        )```

The way I read this, `updatedOrder.lines` would equal `updatedOrderLines` or `[]`, right?

ankitdev10:
Yes, it was not working for some reason but it did when I changed exact those patched lines in this commit directly in my node_modules. I have a customField for discount on orderLines and i was using **applyPriceAdjustments** to trigger **OrderItemPriceCalculationStrategy** for price recalculation. 

To further extend, price was not updated in order_line table as well. And it works now after those changes.

michaelbromley:
Hi,

It is not clear to me what the exact error is that is being fixed by this PR. Would you be able to describe the issue in more detail?

- What you did
- What the expected result was
- What the actual (erroneous) result was

ankitdev10:
Hello,

So, all I was trying to do was add a customField for discount amount on orderLine because the discount depends upon the exemptions file uploaded by the customer. Admin will later apply the discount and the order price will be recalculated once discoutn is applied. I was failing to trigger price recalculation and while looking into **applyPriceAdjustments**  I had doubt that updated order lines are not saved correctly. All these despite **OrderItemPriceCalculationStrategy** getting triggered correctly. 
[here](https://discord.com/channels/1100672177260478564/1203025235025068092/1203773740798836736)  is the link for this disccusion on help thread in vendure discord server

michaelbromley:
Hi,

the scenario you describe sounds quite complex and I think I'd need to see an actual minimal reproduction in order to properly appreciate it.

BijanRegmi:
So in this loop https://github.com/vendure-ecommerce/vendure/blob/3dfb2d196f31b4277355fee0220054008780bc34/packages/core/src/service/services/order.service.ts#L1687-L1719 all the modification to the orderLine is applied to the `updatedOrderLines` parameter object. This mutated `updatedOrderLines` object is never assigned to `order.lines` so when saving `order.lines` as in here https://github.com/vendure-ecommerce/vendure/blob/3dfb2d196f31b4277355fee0220054008780bc34/packages/core/src/service/services/order.service.ts#L1733 the changes made to the `updatedOrderLines` won't be saved to the db. 


If you follow the call to `this.orderCalculator.applyPriceAdjustments` https://github.com/vendure-ecommerce/vendure/blob/3dfb2d196f31b4277355fee0220054008780bc34/packages/core/src/service/helpers/order-calculator/order-calculator.ts#L49-L106 you will find that the `updatedOrderLines` is never assigned to `order.lines`. Hence any changes made to orderLine is lost when you just save `order.lines`.

michaelbromley:
Thanks for the extra explanation.

> So, all I was trying to do was add a customField for discount amount on orderLine because the discount depends upon the exemptions file uploaded by the customer.

Can you please show me the actual code implementation of this part - I don't fully understand what is going on here.

BijanRegmi:
https://github.com/BijanRegmi/vendure-price-issue

Here's a minimal reproduction steps with a failing test case.

michaelbromley:
I'm going to close this as it became stale and I didn't manage to get my head around the issue it was solving. If a problem still exists, you can open a new PR but please include a test that fails before the change, and passes after it.

PR #2665, opened 2024-02-03
fix(admin-ui): Change fragment name to prevent export being ProductListQueryProductFragmentFragment

pevey:
Changes a fragment name to prevent export being ProductListQueryProductFragmentFragment

# Description

This is a very minor change to fix an inconsistently named fragment.  

The ProductListQueryFragment currently has to be imported like:

```ts
import { ProductListQueryProductFragmentFragment } from '@vendure/admin-ui/core'
```
Note the double `Fragment` in the name.

This changes that to be consistent with how other fragments are named.

# Breaking changes

- This could be a breaking change for users who are currently importing this fragment as ProductListQueryProductFragmentFragment

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/catalog/src/components/product-list/product-list.graphql.ts
+++ b/packages/admin-ui/src/lib/catalog/src/components/product-list/product-list.graphql.ts
@@ -1,7 +1,7 @@
 import { gql } from 'apollo-angular';
 
 const PRODUCT_LIST_QUERY_PRODUCT_FRAGMENT = gql`
-    fragment ProductListQueryProductFragment on Product {
+    fragment ProductListQueryProduct on Product {
         id
         createdAt
         updatedAt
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi Lacey,

I believe you would also need to run the `codegen` script in order to generate the actual exported document nodes which include the new names too right?

pevey:
Oh, that's good to know.  I was thinking for some reason it would be in the automated build step.

I ended up going with a different approach that I think may be better, using `GetProductListDocument` instead of the one from the product list component that I was trying to use.

So, I do:

```ts
products$: Observable<ItemOf<GetProductListQuery, 'products'>[]>
```
and then:
```ts
this.products$ = this.dataService.query(GetProductListDocument, { options: {}}).mapSingle(data => data.products.items)
```

There are similar imports for GetCustomerList, GetOrderList, and GetAdministratorList, so everything is consistent and satisfies my OCD. 👍 

I will close this PR.

PR #2670, opened 2024-02-07
fix(payments-plugin): Mollie ignore order states

martijnvdbrug:
# Description

In [this PR](https://github.com/vendure-ecommerce/vendure/pull/2657) we introduced order states that can be ingored when a Mollie webhook comes in, but we missed a few: Cancelled, PartiallyDelivered and partiallyShipped

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/payments-plugin/src/mollie/mollie.service.ts
+++ b/packages/payments-plugin/src/mollie/mollie.service.ts
@@ -12,6 +12,7 @@ import {
     Logger,
     Order,
     OrderService,
+    OrderState,
     OrderStateTransitionError,
     PaymentMethod,
     PaymentMethodService,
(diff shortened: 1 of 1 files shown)


michaelbromley:
This makes me wonder: what about when someone defines custom states? Will that be a potential issue? Is it better to enumerate the known-good states rather than the known-bad states?

martijnvdbrug:
@michaelbromley, yep, definitely. good point.

I guess that should make it 
```ts
if (
 !AddingItems &&
 !ArrangingAdditionalPayment &&
 !ArrangingPayment &&
 !PaymentAuthorized
 ){
  return
}
```

martijnvdbrug:
Updated :-)

PR #2675, opened 2024-02-12
feat(admin-ui): React useProseMirror hook

AleksanderBondar:
# Description

Adding `useProseMirror` for React plugins.

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

PS. I am not pretty sure how to add it into `docs` (not sure about section where it should live) but if someone will pass me some instructions I can add it too :). But we should inform about `missing` css which need to be added into plugin to make prosemirror looking like everywhere.

### DIFF
--- a/packages/admin-ui/src/lib/react/src/react-hooks/use-prose-mirror.ts
+++ b/packages/admin-ui/src/lib/react/src/react-hooks/use-prose-mirror.ts
@@ -1,0 +1,63 @@
+import { useEffect, useRef } from 'react';
+import { Injector } from '@angular/core';
+
+import { CreateEditorViewOptions, ProsemirrorService, ContextMenuService } from '@vendure/admin-ui/core';
+import { useInjector } from './use-injector';
+
+/**
+ * @description
+ * Provides access to the ProseMirror editor instance.
+ *
+ * @example
+ * ```ts
+ * import { useProseMirror } from '\@vendure/admin-ui/react';
+ * import React from 'react';
+ *
+ * export function Component() {
+ *     const { ref, editor } = useProseMirror({
+ *        attributes: { class: '' },
+ *        onTextInput: (text) => console.log(text),
+ *        isReadOnly: () => false,
+ *     });
+ *
+ *     return <div className="w-full" ref={ref} />
+ * }
+ * ```
+ *
+ * @docsCategory react-hooks
+ */
+export interface UseProseMirrorOptions extends Omit<CreateEditorViewOptions, 'element'> {
+    /**
+     * @description
+     * Control the DOM attributes of the editable element. May be either an object or a function going from an editor state to an object.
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,

Thanks for this great work!

Do you think it is best to expose this as a hook, rather than a component? I'm not really a React dev so I'm not familiar with what would be more idiomatic. But to me I would think of this more as a component that you can drop in your template and pass some props to. Is there an advantage to exposing it as a hook?

Also I think I'd prefer it to be named more generically, like `useRichTextEditor`, because
1. It better aligns with the Angular version, `<vdr-rich-text-editor />`
2. "prose mirror" is an implementation detail and I think it could be confusing or harder to find since I'd guess a lot of devs have no idea what "prose mirror" is.

AleksanderBondar:
Hello :D.
I think it can be better to keep it as `hook` it gives more flexibility for users. Because probably when someone will build some form to collect data it needs to be collected in some kind of useState. So we have easy access to store it. I think that is not a huge problem to just print `<div ref={ref} />` instead of importing a new component.

I agree with naming. Will change it today and try with docs.

If You would like to have a component too (I can create it). Maybe that is a good case to expose `Component` and `Hook` so people can decide what to use by their use cases.

michaelbromley:
The component would just be a thin wrapper around the hook right? Very much like the example component in the doc?

If so, then yeah I think it is nice to expose both, and include in the docs for each that there is an option depending on how much control is needed by the developer.

AleksanderBondar:
Yeap exactly it will be like thin wrapper around this hook. I will try to do this today and push it back to review :)

michaelbromley:
Thank you!

andriinuts:
I guess makes sense to move these styles to shared or also include them with the react component because for now, this component is not usable. If you will need any help give me know.

AleksanderBondar:
I have a bit of a problem with accessibility. The fastest I can fix it is sometime during the weekend. If You want to do that go ahead.

andriinuts:
I'm also not sure when I'll  have time to do it, if I'm free I'll let you know not to do the same things

AleksanderBondar:
+1 here. I will put a message here when I got time for that, but glad to hear some1 is using it :P

andriinuts:
and also probably there is a bug with `RichTextEditor ` as you are not setting the `name` property to input, so the form is not registering it.

AleksanderBondar:
Will check that

phuoymakara:
@AleksanderBondar Hello, how can I set the default value on RichTextEditor?

AleksanderBondar:
Hello there :). You should get `editor` instance
```
const { ref, editor } = useRichTextEditor({
   isReadOnly: () => false,
   onTextInput: (text) => updateSummaries({ main: text }),
});
```
Then I am setting default values by `useEffect`
```
useEffect(() => {
   editor.update("Some default");
}, []);
```
Probably this weekend I will get some time to update this hook with `css` and other fixes.

PR #2683, opened 2024-02-20
fix(core): Handle nullable relations in EntityHydrator

### LINKED ISSUE #2682: EntityHydrator crashes if a parent relation is null
**Describe the bug**
Let's say we have three entities (ProductVariant, MetaEntity, Configuration). MetaEntity is a nullable `customField` of `ProductVariant` and `Configuration` is related to `MetaEntity`. 

If I use the `EntityHydrator` to fetch `Configuration` I get an error in `getRelationEntityAtPath` at

```ts
const target = parent[part];
```

because `ProductVariant.customFields.metaEntity` was null `parent['configuration']` is not accessible.
If I understand correctly it would be okay to just return null here as nothing needs to be translated or calculated.

**To Reproduce**
Define a nullable custom field relation which contains another relation. Hydrate the nested relation while the first one is null.

**Expected behavior**
It should return null for the top level relation.

 - Nodejs version: 20
 - Database (mysql/postgres etc): postgres / sqlite

**Additional context**
I already implemented a fix for this, will create a PR as well.

jnugh:
Fixes #2682

# Description
I already described the issue in #2682. The EntityHydrator crashes if a nested relation should be hydrated but the parent is nullable.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/entity-hydrator.e2e-spec.ts
+++ b/packages/core/e2e/entity-hydrator.e2e-spec.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
+    Asset,
     ChannelService,
     EntityHydrator,
     mergeConfig,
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks for the report and PR!

Looks like a slight adjustment is needed to the e2e test for mysql/mariaDB though.

jnugh:
Right, I created the sub relation with `onDelete: 'SET NULL',` but did not make the column nullable. Let's see if this the new commit goes trough :)

PR #2685, opened 2024-02-20
feat(api): Not operator for filtering facet values when using 'search' and MySQL 

HausTechTeam:
# Description

Support for a "not" operator for filtering facet values on the `search` query when using MySQL. The "not" operator is used together with AND or OR, but not both at the same time - in the same way they currently work.

New GraphQL type:
```
input FacetValueNotFilterInput {
    and: ID
    or: [ID!]
}
```

The FacetValueFilterInput type is extended like so:
```
input FacetValueFilterInput {
    and: ID
    or: [ID!]
    not: FacetValueNotFilterInput 
}
```

Usage example:
```
{
  "input": { 
    "facetValueFilters": [
      {
        "not": {
          "or": [1,2,3]
        }      
      }  
    ]
  }
}
```
# Breaking changes

No.

# Screenshots

.

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Hi, thanks for bearing with me. I've now had some time to take a look at this PR.

To test the behaviour, I wrote some e2e tests which you can see here: https://github.com/vendure-ecommerce/vendure/commit/870dc705fc635abe8861409434b957a863209045

3 of the tests pass, but the [`testMatchFacetValueFiltersNotAndMultiple`](https://github.com/vendure-ecommerce/vendure/commit/870dc705fc635abe8861409434b957a863209045#diff-cf29290c4ebf9b451331243415145ce1df2d3204c57e2dc891cb966c156d3edeR354-R376) test fails.

It uses the following input:

```
facetValueFilters: [{ not: { and: 'T_1' } }, { not: { and: 'T_2' } }],
```

I am expecting this to be interpreted as "where the product does not have facet value ID T_1 AND does not have facet value ID T_2".

That is, return all the products (from [this csv file](https://github.com/vendure-ecommerce/vendure/blob/870dc705fc635abe8861409434b957a863209045/packages/core/e2e/fixtures/e2e-products-default-search.csv)) which do not have both the "electronics" (T_1) and "computer" (T_2) facet values.

The result I am getting upon running the tests when in the `src/packages/core` dir:

```
$ DB=mariadb yarn e2e default-search-plugin.e2e

 FAIL  e2e/default-search-plugin.e2e-spec.ts > Default search plugin > shop api > matches by FacetValueFilters NOT AND multiple
AssertionError: expected [ 'Road Bike', 'Skipping Rope', …(8) ] to deeply equal [ 'Instant Camera', …(13) ]

- Expected
+ Received

  Array [
-   "Instant Camera",
-   "Camera Lens",
-   "Tripod",
-   "Slr Camera",
    "Road Bike",
    "Skipping Rope",
    "Boxing Gloves",
    "Tent",
    "Cruiser Skateboard",
    "Football",
    "Running Shoe",
    "Spiky Cactus",
    "Orchid",
    "Bonsai Tree",
  ]

 ❯ testMatchFacetValueFiltersNotAndMultiple e2e/default-search-plugin.e2e-spec.ts:360:61
    358|         });
    359|
    360|         expect(result.search.items.map(i => i.productName)).toEqual([
       |                                                             ^
    361|             'Instant Camera',
    362|             'Camera Lens',
```

i.e., the 4 "photo" products are not being returned, even though they do not have both the "electronics" and "computer" facet values - they have "electronics" and "photo".

Can you take a look at this - is my assumption about the input for this query correct? Did I overlook anything?

HausTechTeam:
Hi! No problem, thank you for providing such a detailed description of the issue. I agree with your interpretation and that the behavior is unexpected.

I've looked into this by running our Vendure fork locally (v2.2.0-next3) and I see some strange results in the generated query. First of all, I don't have the facet value ids "T_1" or "T_2", only numeric values. I'm using `yarn populate` to populate the dev server, using the provided example data (which I believe is the same as the CSV you linked to).

What happens is that the parameters are set to "-1" if I provide a string value to the facet value id filter, for example:
```
"facetValueFilters": [    
    { "not": { "and": "T_1" } }, { "not": { "and": "T_2" } }
]
```

Parameters:
```
{ __1: -1 }
```
(just one parameter when it should be two)

But if I provide only numbers instead:
```
"facetValueFilters": [
    { "not": { "and": "1" } }, { "not": { "and": "2" } }
]
```

The parameters are set to:
```
{ _1: 1, _2: 2 }
```

I'm simply logging the values via `getParameters()` on the SelectQueryBuilder object.

Unless I've done something wrong with my local setup, perhaps this is the reason? In that case, any other passing tests may be false positives.

dlhck:
Closing this because it targets v2.2 which is not maintained anymore. Please feel free to open a new PR with the changes based on the current `minor` branch.

PR #2686, opened 2024-02-20
feat(core): Upgrade typeorm version from v0.3.11 to v0.3.20

monrostar:
# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -78,7 +78,7 @@
         "progress": "^2.0.3",
         "reflect-metadata": "^0.1.13",
         "rxjs": "^7.8.1",
-        "typeorm": "0.3.11"
+        "typeorm": "0.3.20"
     },
     "devDependencies": {
         "@types/bcrypt": "^5.0.0",
--- a/packages/core/src/service/helpers/list-query-builder/list-query-builder.ts
+++ b/packages/core/src/service/helpers/list-query-builder/list-query-builder.ts
@@ -271,11 +271,7 @@ export class ListQueryBuilder implements OnApplicationBootstrap {
             take,
             skip,
             where: extendedOptions.where || {},
-            // We would like to be able to use this feature
-            // rather than our custom `optimizeGetManyAndCountMethod()` implementation,
-            // but at this time (TypeORM 0.3.12) it throws an error in the case of
-            // a Collection that joins its parent entity.
-            // relationLoadStrategy: 'query',
+            relationLoadStrategy: 'query',
         });
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         FindOptionsUtils.joinEagerRelations(qb, qb.alias, qb.expressionMap.mainAlias!.metadata);
@@ -687,10 +683,7 @@ export class ListQueryBuilder implements OnApplicationBootstrap {
         }
 
         if (translationColumns.length && sortingOnTranslatableKey) {
-            const translationsAlias = qb.connection.namingStrategy.eagerJoinRelationAlias(
-                alias,
-                'translations',
-            );
+            const translationsAlias = qb.connection.namingStrategy.joinTableColumnName(alias, 'translations');
 
             qb.andWhere(
                 new Brackets(qb1 => {
--- a/packages/core/src/service/helpers/list-query-builder/parse-filter-params.spec.ts
+++ b/packages/core/src/service/helpers/list-query-builder/parse-filter-params.spec.ts
@@ -66,7 +66,7 @@ describe('parseFilterParams()', () => {
             },
         };
         const result = parseFilterParams(connection as any, Product, filterParams);
-        expect(result[0].clause).toBe('product_translations.name = :arg1');
+        expect(result[0].clause).toBe('productTranslations.name = :arg1');
         expect(result[0].parameters).toEqual({ arg1: 'foo' });
         expect(result[1].clause).toBe('product.id = :arg2');
         expect(result[1].parameters).toEqual({ arg2: '123' });
--- a/packages/core/src/service/helpers/list-query-builder/parse-filter-params.ts
+++ b/packages/core/src/service/helpers/list-query-builder/parse-filter-params.ts
@@ -67,7 +67,7 @@ export function parseFilterParams<
             if (columns.find(c => c.propertyName === key)) {
                 fieldName = `${alias}.${key}`;
             } else if (translationColumns.find(c => c.propertyName === key)) {
-                const translationsAlias = connection.namingStrategy.eagerJoinRelationAlias(
+                const translationsAlias = connection.namingStrategy.joinTableColumnName(
                     alias,
                     'translations',
                 );
--- a/packages/core/src/service/helpers/list-query-builder/parse-sort-params.spec.ts
+++ b/packages/core/src/service/helpers/list-query-builder/parse-sort-params.spec.ts
@@ -1,4 +1,5 @@
 import { Type } from '@vendure/common/lib/shared-types';
+import { fail } from 'assert';
 import { DefaultNamingStrategy } from 'typeorm';
 import { ColumnMetadata } from 'typeorm/metadata/ColumnMetadata';
 import { RelationMetadata } from 'typeorm/metadata/RelationMetadata';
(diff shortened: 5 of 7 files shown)


monrostar:
The first change that needed to be made was to replace `connection.namingStrategy.eagerJoinRelationAlias`
because it was removed in this commit [fix: resolve issue building eager relation alias for nested relations](https://github.com/typeorm/typeorm/commit/c6f608d3e8f9c28646240ac67e20e6567be1aab6 )

monrostar:
New update regarding core tests

image

asonnleitner:
@michaelbromley Could this PR also include changing the name of `connection` to `dataSource` and all related items, such as renaming `dbConnectionOptions` to `dataSourceOptions`, etc.? As mentioned here:

https://github.com/typeorm/typeorm/blob/83567f533482d0170c35e2f99e627962b8f99a08/src/data-source/DataSource.ts#L47-L54

Additionally, considering the naming convention, `TransactionalConnection` could be renamed to `TransactionalDataSource`, among other related changes. Both the `connection` and `dataSource` terminologies can be used for some time, with the `connection`-related items being annotated with `@deprecated` to indicate their phased removal.

monrostar:
Unfortunately, it turned out to be impossible to use DefaultNameStrategy from TypeOrm because inside SQL eager relations we have Alias' name with two `__` and in DefaultNameStrategy with only one `_`.

So we have a difference of one underscore `_`.

example:

```SQL
SELECT
"facet". "createdAt" AS "facet_createdAt",
"facet" "updatedAt" AS "facet_updatedAt", "facet" "isPrivate" AS "facet_isPrivate", "facet". "code" AS "facet_code", "facet". "id" AS "facet_id", "face
__translations". "createdAt" AS "facet__translations_createdAt" ,
"facet__translations" "updatedAt" AS "facet__translations_updatedAt", "facet__translations" "languageCode" AS "facet__translations_la
nguageCode",
"facet__translations". "name" AS "facet__translations_name", "facet__translations". "id" AS "facet__translations_id",
"facet__translations". "baseId" AS "facet__translations_baseId" FROM "
public". "facet" "facet" LEFT JOIN "public". "facet_translation" "facet__translations" ON "facet__translations". "baseId"="facet". "id" LEFT JOIN "public". "facet_channels_channel" "facet_lqb__channel"
ON "facet_lqb__channel". "facetId"="facet". "id" LEFT JOIN "public". "channel" "Zqb__channel" ON "Zqb__channel". "id"="facet_lqb__channel". "channelId" WHERE (facet_translations name ILIKE $1) AND "Zqb__
channel". "id" = $2 ORDER BY "facet" "createdAt" DESC
```

Progress in tests:
image

monrostar:
Progress in tests:

```bash
cross-env PACKAGE=core vitest --config ../../e2e-common/vitest.config.ts --run
```

image

monrostar:
This commit includes entity relationship changes because they are mostly related to @ManyToOne or other relationships that don't have a reverse side relationship for relationLoadStrategy: 'query'. Because the reverse relations were missing, Typeform cannot optimize queries on its own.

Progress in tests:

```bash
cross-env PACKAGE=core vitest --config ../../e2e-common/vitest.config.ts --run
```
image

monrostar:
Related issue in collection Tree decorators: https://github.com/typeorm/typeorm/issues/9936

monrostar:
https://github.com/vendure-ecommerce/vendure/pull/2686/commits/ddaaaf2c62f91c65be38b6944e10aaa77add239e

Special fixes and refactoring for a test case in e2e for nesting in collections
https://github.com/vendure-ecommerce/vendure/blob/39fe6102ba34809f958cd2e56443dcefd734cad5/packages/core/e2e/collection.e2e-spec.ts#L637

monrostar:
> [ddaaaf2](https://github.com/vendure-ecommerce/vendure/commit/ddaaaf2c62f91c65be38b6944e10aaa77add239e) A related issue in the Collection entity Tree decorators: [typeorm/typeorm#9936](https://github.com/typeorm/typeorm/issues/9936)
> 
> Special fixes and refactoring for the test case in e2e for nesting in collections
> 
> https://github.com/vendure-ecommerce/vendure/blob/39fe6102ba34809f958cd2e56443dcefd734cad5/packages/core/e2e/collection.e2e-spec.ts#L637
> 
> image
> @michaelbromley Please review this code. Do you agree with my solution?
> 
> The tests that remain to be fixed image

I would suggest adding property to the system to add a link to relation properties that have Tree decorators, then we can work around this bug until it is fixed. Like I added `collection` and `collection.children` and `collection.parent`

monrostar:
> > [ddaaaf2](https://github.com/vendure-ecommerce/vendure/commit/ddaaaf2c62f91c65be38b6944e10aaa77add239e) A related issue in the Collection entity Tree decorators: [typeorm/typeorm#9936](https://github.com/typeorm/typeorm/issues/9936)
> > Special fixes and refactoring for the test case in e2e for nesting in collections
> > https://github.com/vendure-ecommerce/vendure/blob/39fe6102ba34809f958cd2e56443dcefd734cad5/packages/core/e2e/collection.e2e-spec.ts#L637
> > 
> > image
> > @michaelbromley Please review this code. Do you agree with my solution?
> > The tests that remain to be fixed image
> 
> I would suggest adding property to the system to add a link to relation properties that have Tree decorators, then we can work around this bug until it is fixed. Like I was added `collection` and `collection.children` and `collection.parent`, but we can do it more dynamically This can be a temporary setting in vendureConfig with which you can identify entities that have these problems and just create joins according to the scheme I created


https://github.com/vendure-ecommerce/vendure/pull/2686/commits/32e0df7c5c5f7fcdb68fcd3c37dd99bcab4d4d0f
@michaelbromley 
This is what I was able to achieve when I made this function fully dynamic, now even if someone adds new Tree entities it won't be a problem.

image

monrostar:
https://github.com/vendure-ecommerce/vendure/pull/2686/commits/f9f5260a4e03f527db41ccc8c4639cf1a9bc94ec
Fixes for filtering on translations that were used in join with different aliases due to incorrect aliases
image
image

monrostar:
https://github.com/vendure-ecommerce/vendure/pull/2686/commits/2219b448c2ac2f0541b3cf8c9c2a2b888c9cc700

In this test, we have a preset of only 2 options. I don't understand why the test was for 6 calls if we have only 2 calls to the strategy

@michaelbromley It was your commit, so I need to make sure I didn't break any logic here that you were adding.
https://github.com/monrostar/vendure/commit/a09c2b2022a0380569f1c15dd5a11b1869fb8f63

michaelbromley:
@monrostar This is interesting. At first I thought the 6 calls was due to the fact that each variant is returning 3 price fields. But thinking about it further, the `price` and `priceWithTax` fields on the variant are derived from the one _true_ price value in `prices.price`.

So yes, 2 calls to the transformer does indeed make more sense. 

Yet, in the current version (pre your changes to TypeORM), it is indeed called 6 times. I logged the calls inside the transformer:

```ts
from: (databaseValue: string): number => {
  console.log(`transformerFromSpy: ${databaseValue}`);
  // ...
}
```

and got:

``` 
transformerFromSpy: 31
transformerFromSpy: 999999900
transformerFromSpy: 31
transformerFromSpy: 999999900
transformerFromSpy: 31
transformerFromSpy: 999999900
```

so it looks like the transformer gets called 3x for each variant. Totally strange. No idea why this would change now, but the change seems to be correct.

monrostar:
https://github.com/vendure-ecommerce/vendure/pull/2686/commits/b6b6663332e012fa65d7db1f243715e4010aacc9
image

I have now finalized this PR, I need you to check this code and also try to fix or find the problem in this test. I couldn't find a solution for this test, maybe it is the configuration that is the problem
https://github.com/vendure-ecommerce/vendure/blob/8f9fd9889622b36529af3fc48d5b694c0b738a6e/packages/core/e2e/list-query-builder.e2e-spec.ts#L1190
image

@michaelbromley

monrostar:
I also want to try this configuration at the driver level to find the remaining problematic queries related to collections

Issue #2690, opened 2024-02-21
Allow creating Mollie payment links for orders for draft orders


martijnvdbrug:
**Is your feature request related to a problem? Please describe.**
As an Administrator, I would like to be able to generate payment links for draft orders from the admin UI.

**Describe the solution you'd like**
After creating a draft order, I would like to be able to send my customer a payment link, to finalize the order. Currently, the shop-api mutation is depending on the active order, but for an admin it would be nice to pass an orderId.

* Add a dropdown menu item `createPaymentLink`
* Add a mutation on the admin API `createMolliePaymentIntent(orderId)`

martijnvdbrug:
@michaelbromley This can be assigned to me

martijnvdbrug:
@casperiv0 I know you are also using Mollie. I would like to hear your thoughts on the proposed implementation, since it also affects your project probably. 
(No worries, I will implement it :smile:  just a quick scan of the proposal is asked)

casperiv0:
We're indeed using the Mollie plugin, with the dynamic URLs feature 👀. The way we handle payments goes as following:

1. Go through checkout
2. Payment method list (`molliePayments` query)
3. Click on a payment method will execute the payment flow through Mollie, using the `createPaymentIntent`:
```js
await fetchClientGraphQL(
  createMolliePaymentIntentMutationDocument,
  {
    locale: options.locale,
    variables: {
      input: {
        molliePaymentMethodCode: options.code,
        paymentMethodCode: options.paymentMethodCode,
		// Generated by us using the storefront's URL and some additional information in search params & actual redirect URL
        redirectUrl: options.redirectUrl,
      },
    },
  },
);
```
4. Goes through Mollie flow
5. Redirects to the URL in the `redirectUrl` as a "successful payment" page.

--- 

I see no major breaking changes in your proposal and could possibly be useful for us too in the future.  

When handling a payment intent in the admin UI, I'm guessing it would then use the `fallbackRedirectUrl` if non is set for its mutation execution?

martijnvdbrug:
Thanks for the quick response :pray: It should indeed not break anything.

> When handling a payment intent in the admin UI, I'm guessing it would then use the fallbackRedirectUrl if non is set for its mutation execution?
Yep correct. The priority order is always like this, regardless of admin or shop api:
1. Use redirect url from mutation if given
2. Otherwise use fallbackRedirectURL

[...]

PR #2691, opened 2024-02-21
Feat(Mollie): Prevent duplicate payments

### LINKED ISSUE #2270: Locale context gets lost after paying with Mollie
**Describe the bug**
When integrating Mollie as payment provider, the webhook that processes the payment will trigger an email (assuming you use the `EmailPlugin`). However, the locale context gets lost here since the webhook that Mollie calls on Vendure's side does not have the `languageCode` parameter. This would cause emails to be sent out in the wrong locale.

**To Reproduce**
1. Integrate Mollie as payment provider
2. Enable the EmailPlugin with order confirmations
3. Place an order with a locale that is not the channel's default
4. Pay
5. Get an email in the channel default, instead of the one used while ordering

**Expected behavior**
The email should be sent out with the locale that was used to order.

 - @vendure/core version: 2.0.2
 - Nodejs version: 18

**Additional context**
This can be solved by creating custom fields in the order and reading the locale from there, but that's not working E2E yet due to issue #2269

### LINKED ISSUE #2620: Customers are able to pay twice for an order in Mollie
**Describe the bug**
We get reports from customer service that every now and then a customer pays twice for an order in Mollie.
In Mollie I do indeed see two payments created for that order. Whereas we would expect that this is prevented somehow. As in prevent the customers from starting a 2nd check-out. 

Screenshot 2024-01-11 at 10 47 53

**To Reproduce**
1. Add items to an order
2. Use createPaymentIntent mutation
3. Go to the returned check-out url
4. Use createPaymentIntent for the order again
5. A new url is returned
6. Go to the returned check-out url
7. Finish payments in both urls
8. You have done two payments which should be revealed in Mollie's dashboard as well. (See schreenshot)

**Expected behavior**
At step 4 an error is returned / user is warned that another payment has already been started.

 - @vendure/core version: 2.1
 - Nodejs version: 18

martijnvdbrug:
# Description

* Prevent duplicate payments by storing a reference to the already-open payment on an order. Closes #2620  
* Preserve original locale on Mollie webhook call. Closes #2270 
* Added `createMolliePaymentIntent(orderCode)` mutation on Admin API, as preparation on #2690 

# Breaking changes

* A DB migration is needed to add the `mollieOrderId` field on an order.

# Checklist

📌 Always:

👍 Most of the time:


packages/core/src/service/index.ts
```diff
@@ -12,6 +12,7 @@ export * from './helpers/order-merger/order-merger';
 export * from './helpers/order-modifier/order-modifier';
 export * from './helpers/order-splitter/order-splitter';
 export * from './helpers/order-state-machine/order-state';
+export * from './helpers/order-state-machine/order-state-machine';
```
martijnvdbrug, line 15: Without this, `this.injector.get(OrderStateMachine);` in the Mollie service throws an error.

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -324,6 +335,31 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
             });
         });
 
+        it('Should reuse payment url when amount is the same', async () => {
```
martijnvdbrug: The test if a new order is created when amounts are **not** the same, is already being done implicitly in the next test `Should get payment url with deducted amount if a payment is already made`

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -547,8 +596,8 @@ describe('Mollie payments (with useDynamicRedirectUrl set to true)', () => {
 
     it('Should add an unusable Mollie paymentMethod (missing redirectUrl)', async () => {
         const { createPaymentMethod } = await adminClient.query<
-            CreatePaymentMethod.Mutation,
-            CreatePaymentMethod.Variables
+            CreatePaymentMethodMutation,
+            CreatePaymentMethodMutationVariables
```
martijnvdbrug, line 611: These types were broken, but there is no type check or linting on test files. (Ideally there are, but that requires quite some time to set up...)

packages/payments-plugin/src/mollie/custom-fields.ts
```diff
@@ -0,0 +1,16 @@
+import { CustomFieldConfig, Order, CustomOrderFields } from '@vendure/core';
+
+export interface OrderWithMollieReference extends Order {
+    customFields: CustomOrderFields & {
+        mollieOrderId?: string;
+    };
+}
```
martijnvdbrug, line 7: Use a custom interface instead of a `*-types.d.ts` file, because that would also infer `customFields.mollieOrderId` in the Stripe, Paypal and braintree plugin.

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -96,30 +113,27 @@ export class MollieService {
...
+            }
         }
-        if (!order.customer.firstName.length) {
+        if (!order.customer?.firstName.length) {
             return new PaymentIntentError(
                 'Cannot create payment intent for order with customer that has no firstName set',
             );
         }
-        if (!order.customer.lastName.length) {
+        if (!order.customer?.lastName.length) {
```
martijnvdbrug, line 131: Other checks like `Order has order lines` have been removed, because they are checked by `canTransitionTo('ArrangingPayment')`.

martijnvdbrug:
`Vendure:cli` unit test is failing, is that expected on this branch, or did I break something?

```shell
Failed tasks:
   
   - @vendure/cli:test
```

martijnvdbrug:
Convert to draft. There is one thing I'd still like to check: If we can update the contents of an existing order. Then we wouldn't have to check if we can reuse the order and not cancel, we'd just always update the contents of the Mollie order with the Vendure order data.

packages/payments-plugin/src/mollie/mollie.helpers.ts
```diff
@@ -34,7 +34,7 @@ export function toMollieOrderLines(order: Order, alreadyPaid: number): CreatePar
         quantity: line.quantity,
         unitPrice: toAmount(line.proratedLinePriceWithTax / line.quantity, order.currencyCode), // totalAmount has to match unitPrice * quantity
         totalAmount: toAmount(line.proratedLinePriceWithTax, order.currencyCode),
-        vatRate: String(line.taxRate),
+        vatRate: line.taxRate.toFixed(2),
```
martijnvdbrug, line 37: Send as `20.00` to Mollie instead of `20`, otherwise comparison of new order vs existing order fails, because vat rate's dont equal.

packages/core/src/service/index.ts
michaelbromley: it throws an error if you use a deep import to this?

```
@vendure/core/dist/service/helpers/order-state-machine/order-state-machine
```

weird, but ok - it's worth exposing anyway.

packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
michaelbromley: yeah the e2e test suites could use a bit of maintenance re TS errors. Not a high prio though for me since the main function (telling us if we broke something) still works. But yeah I tend to fix errors I see whenever I touch an existing file.

packages/payments-plugin/src/mollie/custom-fields.ts
michaelbromley: OK, makes sense to encapsulate this implementation detail 👍

michaelbromley:
@martijnvdbrug thank you for this PR - clearly it took quite some time & careful effort to get right!

Issue #2718, opened 2024-03-05
Default Currency and Laguague are required while creating a new Channel, but UI misleads


arrrrny:
**Describe the bug**

**To Reproduce**
1. Go to 'Channel->Create New Channel on Admin UI'
2. Fill in Code, channel token and seller
3. User will be asked to fill in default tax zone and default shipping zone, 'Create' button on the right is greyed out
4. Select Default Shipping Zone and Tax Zone. 'Create' button is now 'blue' and user expects to create when clicked
5. Click on 'Create' and no response. No request is sent to server.
6. select 'Available Currency' and 'Available Language' and then select Default language and currency, now it creates the channel.

**Expected behavior**
There should be same type of warning like on the 'tax' for default language and currency. Craete button should stay greyed out until these fields are filled.

**Environment (please complete the following information):**
 - @vendure/core version: 2.1.7
 - Nodejs version v20
 - Database (mysql/postgres etc): postgres

<img width="1495" alt="Screenshot 2024-03-06 at 12 15 50 AM" src="https://github.com/vendure-ecommerce/vendure/assets/48218623/635a6b84-fa30-4155-87f2-de328b65744b">

PR #2722, opened 2024-03-07
fix(admin-ui): Require default currency and language for channels

oidt:
# Description

PR for this bug: https://github.com/vendure-ecommerce/vendure/issues/2718

# Screenshots

image


oidt:
I put default language under available language and the same for currencies.
Changed warn under default shipping zone to danger because it is a required field just like default tax zone.
could not found any e2e for this part, if we have please let me know I will update the PR.

michaelbromley:
Thanks!

Issue #2723, opened 2024-03-08
Custom field with entity `ProductVariant` throws `Cannot set properties of null (setting 'taxCategory')` when value is empty


andriinuts:
**Describe the bug**
When you are adding a custom field as  `ProductVariant` entity it throws an error ``Cannot set properties of null (setting 'taxCategory')` when you are leaving this value empty (for example when creating a new order line).

`config.customFields.OrderLine.push({
      name: 'relatedVariant',
      type: 'relation',
      entity: ProductVariant,
    });
`

**To Reproduce**
1. Go to the new draft order and add a new item without selecting `relatedVariant` value
2. Click `Add item to order`
3. See error

**Expected behavior**
I would expect that we will not receive this error because the field it not required in the configuration and it should be possible to skip it.

**Environment (please complete the following information):**
 - @vendure/core version: 2.1.8
 - Nodejs version:  v20.10
 - Database (mysql/postgres etc): any

PR #2727, opened 2024-03-10
fix(core): Fix custom field relation for ProductVariant when value is null

andriinuts:
# Description

Fixes https://github.com/vendure-ecommerce/vendure/issues/2723
The reason was that we are calling `applyVariantPrices` even when the entity is null. And it happens only for `ProductVariant`.

# Breaking changes

No

# Checklist

📌 Always:

👍 Most of the time:


andriinuts:
is it possible to restart checks? or only you can restart them?

michaelbromley:
Thank you!

PR #2730, opened 2024-03-11
Update type-ormhealth-check-strategy.md

gspoon666:
# Description

only added healthcheck to systemconfig

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Hi, thanks for this! The actual source of this md file is [here](https://github.com/vendure-ecommerce/vendure/blob/1579e49350e58549420df722790303a9eb9d8984/packages/core/src/health-check/typeorm-health-check-strategy.ts#L25-L30) - would you mind updating that part? Otherwise the md file will be overwritten next time the docs get generated.

PR #2743, opened 2024-03-16
perf(core): optimization for assignToChannels method

monrostar:
# Description

This is an optimization of the channel-aware update request for Entity. The method offers 3 separate queries without additional data processing by TypeORM. 
1. Get the object without unnecessary links and check that it exists.
2. Make a separate, more optimized query into the ManyToMany table to reduce the amount of data to process
3. Change the save method to `Remove` or `Add` (unbinds, binds) given value from entity relation.

```ts
relation('channels')
.of(entity.id)
.add(newChannelIds) or .remove(existingChannelIds);
```

# Breaking changes

No

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/channel.service.ts
+++ b/packages/core/src/service/services/channel.service.ts
@@ -11,17 +11,12 @@ import {
 import { DEFAULT_CHANNEL_CODE } from '@vendure/common/lib/shared-constants';
 import { ID, PaginatedList, Type } from '@vendure/common/lib/shared-types';
 import { unique } from '@vendure/common/lib/unique';
-import { FindOneOptions, FindOptionsWhere } from 'typeorm';
+import { FindOptionsWhere } from 'typeorm';
 
 import { RelationPaths } from '../../api';
 import { RequestContext } from '../../api/common/request-context';
 import { ErrorResultUnion, isGraphQlErrorResult } from '../../common/error/error-result';
-import {
-    ChannelNotFoundError,
-    EntityNotFoundError,
-    InternalServerError,
-    UserInputError,
-} from '../../common/error/errors';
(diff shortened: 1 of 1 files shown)


packages/core/src/service/services/channel.service.ts
```diff
@@ -124,6 +119,42 @@ export class ChannelService {
...
+        const metadata = repository.metadata;
+        const channelsRelation = metadata.findRelationWithPropertyPath('channels');
+
+        if (!channelsRelation) {
+            throw new InternalServerError(`Could not find the channels relation for entity ${metadata.name}`);
+        }
+
+        const junctionTableName = channelsRelation.junctionEntityMetadata?.tableName;
+        const junctionColumnName = channelsRelation.junctionEntityMetadata?.columns[0].databaseName;
+        const inverseJunctionColumnName = channelsRelation.junctionEntityMetadata?.inverseColumns[0].databaseName;
```
michaelbromley, line 145: With the latest updates we explicitly added inverse relations to the Channel entity. 

I'm wondering what would happen if someone defines a custom channel-aware entity which does not have the inverse relation on the Channel? Would this break?

monrostar: This is quite normal, there should be no errors because even without reverse side entity is still available
But if there are problems we replace them with pure SQL

> With the latest updates we explicitly added inverse relations to the Channel entity.
> 
> I'm wondering what would happen if someone defines a custom channel-aware entity which does not have the inverse relation on the Channel? Would this break?

monrostar: or just manually parse the column names in the table. This can also be done. We just need to test it on a real project

PR #2746, opened 2024-03-20
feat(core): Add replicationMode for ctx and getRepository

monrostar:
# Description

This PR includes improvements for connecting to the database if you are using multiple replicas and when you need to connect to Master without using TransactionManager

# Breaking changes

 - No

### DIFF
--- a/packages/core/src/connection/transactional-connection.ts
+++ b/packages/core/src/connection/transactional-connection.ts
@@ -11,6 +11,7 @@ import {
     ObjectType,
     Repository,
     SelectQueryBuilder,
+    ReplicationMode,
 } from 'typeorm';
 
 import { RequestContext } from '../api/common/request-context';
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hi,

Thanks, this looks good. Can you make a suggestion for some documentation on this feature? At least as doc blocks on the new method/property.

monrostar:
> Hi,
> 
> Thanks, this looks good. Can you make a suggestion for some documentation on this feature? At least as doc blocks on the new method/property.

Of course, I will. I can do it tomorrow

packages/core/src/connection/transactional-connection.ts
```diff
@@ -73,21 +74,32 @@ export class TransactionalConnection {
     getRepository<Entity extends ObjectLiteral>(
         ctx: RequestContext | undefined,
         target: ObjectType<Entity> | EntitySchema<Entity> | string,
+        replicationMode?: ReplicationMode,
     ): Repository<Entity>;
     getRepository<Entity extends ObjectLiteral>(
         ctxOrTarget: RequestContext | ObjectType<Entity> | EntitySchema<Entity> | string | undefined,
         maybeTarget?: ObjectType<Entity> | EntitySchema<Entity> | string,
+        replicationMode?: ReplicationMode,
```
monrostar: Here it's probably better to use an object instead of a strict arg

michaelbromley: For future extensibility?

monrostar: yes. For example, there may be several databases in the future

monrostar:
@michaelbromley 

I've added documentation to this code. We actively use the master connection in index update and other methods, such as email-send or SMS-send, to ensure that at the time of processing, workers take the most up-to-date data if this data was updated less than a second ago.

I'd still like to work on the Queue interface. We have identified a problem that some tasks can be executed before the transaction is completed on the server. 
I can make an issue or prepare a PR with an extended interface to be able to extend the functions of task registration depending on the service that is used

michaelbromley:
Hey, thanks for finding the time to complete this PR!

It now strikes me that we should have a page in our `developer guide > advanced topics` covering how to run Vendure with a replicated DB setup, so that others are aware of the considerations and potential pitfalls to this arrangement. Would you have time/interest in writing anything on this topic? Even just a list of bullet points that I can then flesh out. 

> I can make an issue or prepare a PR with an extended interface to be able to extend the functions of task registration depending on the service that is used

Yeah an issue outlining the problem & proposed solution would be a good start :+1

PR #2749, opened 2024-03-21
Fix aggregation buckets with taxes

Swampy469:
# Description

Fix a typo in ElasticSearch plugin on elasticsearch.service.ts where bucketsWithTax value was using prices value and not pricesWithTax value.

# Breaking changes

No


michaelbromley:
Thanks for the contribution!

PR #2754, opened 2024-03-27
feat(braintree): Allow generating a token before adding customer to order

pevey:
# Description

I'm working on a a single-page checkout using Braintree.  I found that it is not currently possible to generate a Braintree token/authorization code (which is needed to show the Braintree-web-dropin UI) before adding a customer to the active order.  However, for this particular use case, we want to have a single form that captures the customer name and email in step 1 but also shows the payment form at the same time.  This PR enables this. 

This may be better classified as a fix, because it looks like it may have been intended to allow generating tokens without a customer.  There is a boolean query argument, includeCustomerId.  However, this does not actually allow generating codes if there is no customer attached to the order yet because of [this line](https://github.com/vendure-ecommerce/vendure/blob/master/packages/payments-plugin/src/braintree/braintree.resolver.ts#L47).

# Breaking changes

I think this is done in a way that should not break existing functionality.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/payments-plugin/src/braintree/braintree.resolver.ts
+++ b/packages/payments-plugin/src/braintree/braintree.resolver.ts
@@ -44,8 +44,8 @@ export class BraintreeResolver {
             );
         }
         const order = await this.orderService.findOne(ctx, sessionOrder.id);
-        if (order && order.customer) {
-            const customerId = order.customer.customFields.braintreeCustomerId ?? undefined;
+        if (order) {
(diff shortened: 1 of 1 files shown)


pevey:
I'm not sure why the pipeline is failing.

pevey:
It seems like the failure had to do with perhaps two different PRs running through the checks at the same time?  Not sure why the port would have been in use.

michaelbromley:
Thanks! The e2e tests sometimes are flaky on the port. Not totally sure why, needs investigation.

PR #2766, opened 2024-04-04
Fix missing shipping lines in partial order after payment

Swampy469:
# Description
We encountered an issue regarding the absence of shipping methods in the order generated after payment completion. This problem occurred due to the duplication of shipping method items without assigning new IDs to the orderLines that had the shippingLineId attribute.


michaelbromley:
Thanks for the fix!

PR #2779, opened 2024-04-11
fix(core): Fix floating promise & missed eventBus

tianyingchun:
# Description

Fix floating promise & missed eventBus

# Breaking changes

no

# Screenshots

image


michaelbromley:
Wow great catch on that missing await in the map!

Issue #746, opened 2021-03-04
Is there a reason why creating an account with an email already registered doesn't return an error?


mariomacedo:
Is there a reason why creating an account with an email already registered doesn't return an error?

**Update - more information**

Running the following mutation multiple times with the same credentials always returns success.

```
mutation Register($input: RegisterCustomerInput!) {
  registerCustomerAccount(input: $input) {
    __typename
    ... on Success {
      success
    }
    ...ErrorResult
  }
}

fragment ErrorResult on ErrorResult {
  errorCode
  message
}
```

**Here are the two cases I could think of:**
1) _If the user registered, but has not yet verified his account._ Any subsequent registration attempt with the same credentials returns success. Also, multiple verification emails are sent. If different passwords were specified during the registration attempts, the first one is the one recorded in the database. You could see how this might confuse customers.

2) _If the account is verified_, any subsequent registration attempt returns success.

**Looking at the error codes currently used, it seems like it should return:**

- EMAIL_ADDRESS_CONFLICT_ERROR for case (2)

- Either EMAIL_ADDRESS_CONFLICT_ERROR or NOT_VERIFIED (subject to issue #500) for case (1), although I prefer the latter as it is more informative to the customer.

I'm now aware of commit b1ffa1e that makes the registration mutation silently fail to protect user data. You're trading off extra security for customer experience. As a user, I often forget if I have an account on a particular website - especially if I don't use it often. I'm curious to hear what you think.

_Originally posted by @marwand in https://github.com/vendure-ecommerce/vendure/issues/33#issuecomment-713831098_

michaelbromley:
Thanks for resurrecting this detailed comment as an issue. I somehow missed it the first time and never replied.

This is well worth considering newly. Regarding the possibility of exposing user data if we return different results for existing email addresses, perhaps this can be better mitigated by e.g. rate-limiting certain mutations rather than just hiding the results.

alibeknow:
Same problem how we can solve it ?

ssex-dev:
Isn't this a really big security problem? 
If I know the email address of a customer, I could change there password by registering again and would be able to receive there information like there orders, name and address, phone number. It would also be not so hard to automate this and simply scrape all shops I can find with a vendure backend for a list of known email addresses. 
I started to build a shop with vendure, please tell me that you fix this or that I'am wrong.

michaelbromley:
[...]

PR #2786, opened 2024-04-16
fix(core): Changes to Native Auth Register Flow

pevey:
# Description

This PR makes changes to the customer registration flow to address two issues:

- Possible email flooding (https://discord.com/channels/1100672177260478564/1229813869548146790)
- Possible data override of existing customers (https://discord.com/channels/1100672177260478564/1228399534745456742)

# Breaking changes

This PR DOES include breaking changes.

The existing sign-up flow checks first to see if a user with that email address exists.  If so, it pulls that data as a starting point.  

- If the existing user is verified, the user record (excluding password) is updated with the submitted data and 'Success' is returned by the mutation.  This is not ideal for two reasons:  The existing data should not be overridden without authentication, and there is no indication to the frontend that the user already exists.  This PR changes this flow to instead return a EmailAddressConflictError.

- If the existing user is not verified, currently a new verification token is generated and emailed to the user upon each sign up attempt.  This could be problematic if a user keeps entering the same (incorrect) email address as it generates spam.  It could also be used maliciously for this effect.  This PR changes the flow to first check to see whether the existing verification token is still valid.  If so, a NotVerifiedError is returned so that the frontend can provide a useful error message and prompt the user to check their email.  If the token is expired, it is possible the record is very old or from someone who initially entered an invalid email address.  For example, joel@example.com accidentally entered joe@example.com.  He realizes his error, corrects his email address, and verifies his joel@example.com address. Later, the real joe@example.com shows up and tries to register.  Only in this case, where the existing token is expired, will a new token be generated.  It will be treated as if the existing user record does not exist.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/graphql/generated-e2e-shop-types.ts
+++ b/packages/core/e2e/graphql/generated-e2e-shop-types.ts
@@ -6,215 +6,227 @@ export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K]
 export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
 export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
 export type MakeEmpty<T extends { [key: string]: unknown }, K extends keyof T> = { [_ in K]?: never };
-export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+export type Incremental<T> =
+    | T
+    | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
 /** All built-in and custom scalars, mapped to their actual values */
 export type Scalars = {
-  ID: { input: string; output: string; }
-  String: { input: string; output: string; }
-  Boolean: { input: boolean; output: boolean; }
-  Int: { input: number; output: number; }
-  Float: { input: number; output: number; }
-  DateTime: { input: any; output: any; }
-  JSON: { input: any; output: any; }
-  Money: { input: number; output: number; }
-  Upload: { input: any; output: any; }
+    ID: { input: string; output: string };
+    String: { input: string; output: string };
+    Boolean: { input: boolean; output: boolean };
+    Int: { input: number; output: number };
+    Float: { input: number; output: number };
+    DateTime: { input: any; output: any };
+    JSON: { input: any; output: any };
+    Money: { input: number; output: number };
+    Upload: { input: any; output: any };
 };
 
 export type ActiveOrderResult = NoActiveOrderError | Order;
 
-export type AddPaymentToOrderResult = IneligiblePaymentMethodError | NoActiveOrderError | Order | OrderPaymentStateError | OrderStateTransitionError | PaymentDeclinedError | PaymentFailedError;
+export type AddPaymentToOrderResult =
+    | IneligiblePaymentMethodError
+    | NoActiveOrderError
+    | Order
+    | OrderPaymentStateError
+    | OrderStateTransitionError
+    | PaymentDeclinedError
+    | PaymentFailedError;
 
 export type Address = Node & {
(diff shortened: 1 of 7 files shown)


pevey:
If you're good with this approach, I'll fix the auth-related e2e tests.

michaelbromley:
Hi,

My only concern with the first change is that it opens up the possibility of an account enumeration attack as discussed here:

- https://github.com/vendure-ecommerce/vendure/issues/746

Do you have any thoughts on that?

---

On the second part, I agree with this approach. One point I would change is to not reuse the `NotVerifiedError` since it is semantically a little different:

```gql
"""
Returned if `authOptions.requireVerification` is set to `true` (which is the default)
and an unverified user attempts to authenticate.
"""
type NotVerifiedError implements ErrorResult {
    errorCode: ErrorCode!
    message: String!
}
```

We can create a new error result especially for this case, e.g. `VerificationPendingError` (just a suggestion on the name).

pevey:
On the first topic, it is a good question and one that I also have wondered about.  Security vs UX is always a careful balance. 
 There was discussion of this with the author of the Lucia authentication library in this discord thread if you're interested.  https://discord.com/channels/1004048134218981416/1131337919957839972

In short, he changed my mind about this, and I no longer think feedback about whether an account exists should be hidden from users.  You can see now that pretty much all popular sites give feedback on whether an account exists.  Not doing it is a really poor user experience.  This is applicable to both the sign up and password reset flows.  

To mitigate enumeration, those same sites also use rate-limiting, via actual rate limiting and also bot detection.  Rate-limiting is not something Vendure currently supports out of the box, but it easily could since it is built on NestJS.  That is another PR I'd like to do.  Bot detection is something that everyone hosting ecom should be doing, for this reason but also mostly to protect against carding.  But that leaves the whole industry very reliant on the best bot protection, which is currently Cloudflare with no close second.

So bottom line, I think it is really important to provide the useful error.  Sometimes, you forget whether you created an account somewhere, especially an online store you may not purchase from frequently.  In my checkout flow, if the user is not logged in, I take the email address entered and try to set the customer for order.  If an account exists for that customer, it returns an EmailAddressConflictError.  Which it should, so that you can prompt the user to log in.  This is good UX.  But from this functionality, there already exists the possibility to enumerate accounts without other security protections in place.

pevey:
On the second issues, I did consider adding a new error type, and I'll go ahead and do that if you think it's the best approach.  But an alternative would be to slightly change the wording of the existing error.

pevey:
I'll add the new error and work on the tests.

I'll do a separate PR that implements rate limiting using the NestJS throttler.   https://docs.nestjs.com/security/rate-limiting

pevey:
While we are making changes to the auth flow, I want to bring up another question I have which is very related and should possibly be addressed by this PR.

- As you know, there is a config option for whether to require email verification with the native auth method
- Currently, the logic for adding native auth to a customer account goes like this:
    - If require verification is true, set verified to false, and generate verification token/send notification
    - If require verification is false, set verified to true <- POTENTIAL PROBLEM?
- This logic could be improved, I think.  The account is not really verified.  If we later change the setting to require verification, we have a bunch of accounts with verified set to true, but really they have never been verified.
- In the past, changing this would have broken the sign up flow, but here we are changing the sign up flow already.  Changing it will also break the authentication flow unless we make another small change that only throws the not verified error if the user is not verified AND the require verification is set to true.

Making these small changes would mean:
- The verified status of a user always reflects reality and is not dependent on whether require verification is enabled or disabled. and
- You could switch back and forth between requiring and not requiring verification without major issues.

pevey:
I think I need to generate new type files for the e2e tests, but I'm not sure how to generate them in a way that would include all the types for all plugins so that those tests don't fail.

pevey:
I've hit a wall for now on fixing the e2e tests.  I'm going to take a break from it for now and work on something else and look at this fresh in a couple days.

michaelbromley:
Agreed on the change to verification handling. We'll need to clearly document this change, but it does make sense to do it that way.

Let me know if you need help with the e2e test issues.

PR #2798, opened 2024-04-19
fix(core): Add surcharge taxLines to taxSummary

### LINKED ISSUE #2797: taxSummary on Order does not contain taxLines of surcharges
**Describe the bug**
The `taxSummary` in the `Order` entity does not include the `taxLines` of the orders surcharges. On the other side the `totalWithTax` field is calculated correctly with all taxes from the surcharges. I will submit a Pull request which should solve this issue.

**To Reproduce**
1. Create an order with one or more surcharges
2. Set some `taxLines` on the surcharges
3. Query the `taxSummary` of the order
4. `taxSummary` does not include the `taxLines` of the surcharges

**Expected behavior**
`taxSummary` should contain `taxLines` of surcharges

 - @vendure/core version: 2.1.7
 - Nodejs version: 18
 - Database (mysql/postgres etc): sqlite / postgres

tschortsch:
# Description

This change adds the `taxLines` of the orders surcharges to the `taxSummary`. Fixes #2797.

# Breaking changes

Since this fix changes the output of the `taxSummary` resolver it could be a breaking change for users which use the `taxSummary` in there storefront.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/entity/order/order.entity.spec.ts
+++ b/packages/core/src/entity/order/order.entity.spec.ts
@@ -5,6 +5,7 @@ import { beforeAll, describe, expect, it } from 'vitest';
 import { ensureConfigLoaded } from '../../config/config-helpers';
 import { createOrder, createRequestContext, taxCategoryStandard } from '../../testing/order-test-utils';
 import { ShippingLine } from '../shipping-line/shipping-line.entity';
+import { Surcharge } from '../surcharge/surcharge.entity';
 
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks for the fix & especially for including the test 👍 . I'd not class this as a breaking change in the sense that we are just adding data that was omitted on account of a bug that is now fixed.

Issue #2787, opened 2024-04-16
collection-duplicator throws error on asset association


mschipperheyn:
**Describe the bug**
When using the collection-duplicator that has assets associated to it, an error is thrown because of an odd [constructor](https://github.com/vendure-ecommerce/vendure/blob/09f25d39bf8510cc0c13c19396589492565eaf34/packages/core/src/service/services/asset.service.ts#L614) reference.

The root cause seems to be that the when updating a collection, the asset service update collection function is called with an empty array. The sort function then assumes there is at least 1 item and references the first one, which can be undefined.

```
error 4/16/24, 5:05 PM - [Vendure Server] Cannot read properties of undefined (reading 'constructor')
TypeError: Cannot read properties of undefined (reading 'constructor')
    at AssetService.createOrderableAssets (/Users/mschipperheyn/workspace/maos/node_modules/@vendure/core/src/service/services/asset.service.ts:614:70)
    at AssetService.updateEntityAssets (/Users/mschipperheyn/workspace/maos/node_modules/@vendure/core/src/service/services/asset.service.ts:275:40)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async CollectionService.create (/Users/mschipperheyn/workspace/maos/node_modules/@vendure/core/src/service/services/collection.service.ts:465:9)
    at async EntityDuplicator.duplicate [as duplicateFn] (/Users/mschipperheyn/workspace/maos/node_modules/@vendure/core/src/config/entity/entity-duplicators/collection-duplicator.ts:66:38)
    at async /Users/mschipperheyn/workspace/maos/node_modules/@vendure/core/src/service/helpers/entity-duplicator/entity-duplicator.service.ts:76:35
```

**To Reproduce**
1. Create a collection with associated assets
2. Duplicate the collection

**Expected behavior**
It should copy the collection and copy associated assets

**Environment (please complete the following information):**
 - @vendure/core version: 2.2.0
 - Nodejs version: 18.x
 - Database (mysql/postgres etc): postgres

mschipperheyn:
After further investigation I found that the cause of this issue is that when I use the collection import through the populator on a non default channel, the assets get imported on the default channel. [Here](https://github.com/vendure-ecommerce/vendure/blob/9c1cb1662e899c62d93f8741f9476f415c721998/packages/core/src/service/services/asset.service.ts#L270), it will subsequently return an empty array and fail

[...]

PR #2801, opened 2024-04-19
fix asset importer assigns asset on default channel when importing on non default channel

mschipperheyn:
# Description

Adds RequestContext to asset import for collections. 

The populator loads assets into collections. However, the RequestContext is not passed along. This means that if you import collections on any channel that is not default, the assets get assigned to the default channel only.

As a result a channel specific collection import fails because an empty asset array results that is assumed to be non empty, and referenced.

Resolves issue https://github.com/vendure-ecommerce/vendure/issues/2787

I adjusted test on the populate spec and was surprised that even on the 'populating a non-default channel' the test passed without my changes. On further inspection it seems that that test uses the default channel for its work instead of the intended channel-2. So, it seems to be not validating what it's supposed to validate. That might be because I'm running a filtered test or because the test has a somewhat unique setup that destroys the server after the additional test channel is created. Initially I thought it might a cache refresh issue but later I realized that channel-2 wasn't in the database during the test. So, writing the test became a bit of a rabbit hole and I bailed after a while. Perhaps someone can enlighten me on the test setup and how it's supposed to work.

BTW: after running npm install, testing doesn't work out of the box. had to remove package-lock.json to be able to run tests at all bc of darwin related errors on optional dependencies

# Breaking changes

None

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/data-import/providers/populator/populator.ts
+++ b/packages/core/src/data-import/providers/populator/populator.ts
@@ -125,7 +125,7 @@ export class Populator {
         for (const collectionDef of data.collections) {
             const parent = collectionDef.parentName && collectionMap.get(collectionDef.parentName);
             const parentId = parent ? parent.id.toString() : undefined;
-            const { assets } = await this.assetImporter.getAssets(collectionDef.assetPaths || []);
+            const { assets } = await this.assetImporter.getAssets(collectionDef.assetPaths || [], ctx);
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks for the fix. Regarding the e2e tests - if you like you can share a branch with the changes you made and when I have time I could look into it.

Issue #2804, opened 2024-04-22
Relation Custom Fields on StockLocation not Updating


dylviz:
**Describe the bug**
Update() method in stock location service file is missing call to update the relations. See code here:

https://github.com/vendure-ecommerce/vendure/blob/9c1cb1662e899c62d93f8741f9476f415c721998/packages/core/src/service/services/stock-location.service.ts#L89

**To Reproduce**
1. Create a custom relation on the StockLocation entity within your plugin
2. Run dev-server, open admin-ui
3. Modify the customField on StockLocation entity then click Update()
4. Navigate away from the page and back again and see that your update is not shown
5. This is because your update wasn't saved
6. verified in the DB that the updated relation is not modified/saved

**Expected behavior**
Updated relation should be shown in the admin-ui and in the DB StockLocation table

**Environment (please complete the following information):**
- Verified bug on V2.1.7 and V2.2.0
 - @vendure/core version: > 2.1.7+
 
 **Additional Context**
 Will link PR

PR #2805, opened 2024-04-22
fix: Update Relations on Stock Location update

dylviz:
The Stock Location update method was not updating relation customFields

Refs: #2804

# Description

Essentially just a few lines of code added to the `update` function on StockLocation service file. This will allow relation customField entities to be updated along with the StockLocation update. 

Here are the added lines. Instead of just saving, we now save, then update relation then return a promise<StockLocation> off of assertFound

```
await this.connection.getRepository(ctx, StockLocation).save(updatedStockLocation);
await this.customFieldRelationService.updateRelations(
    ctx,
    StockLocation,
    input,
    updatedStockLocation,
);
return assertFound(this.findOne(ctx, updatedStockLocation.id));
```

# Breaking changes

No Breaking Changes, Just additional steps included in `update()` function

# Screenshots
N/A

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Thank you!

PR #2791, opened 2024-04-17
fix(core): Remove original entityTable from channels sql request

monrostar:
# Description

This fisk is necessary because typeorm does not remove the original table from the `from` query when creating the qb. In order to do this it was necessary to use manager.

Now a SQL request looking like this 
```sql
‌SELECT "channel"."channelId" AS "channelId" FROM "public"."product_channels_channel" "channel" WHERE "channel"."productId" = $1
```
Prev:
```sql
SELECT "channel"."channelId" AS "channelId" FROM "public"."asset" "Asset", "public"."asset_channels_channel" "channel" WHERE "channel"."assetId" = $1
```

# Breaking changes

Does this PR include any breaking changes we should be aware of?
 * no

# Screenshots
telegram-cloud-photo-size-2-5258359367533714224-y

# Checklist

📌 Always:
- [ ] I have set a clear title
- [ ] My PR is small and contains a single feature
- [ ] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

PR #2810, opened 2024-04-25
fix(getAssignedEntityChannels): fix performance problems in v2.2

alexisvigoureux:
# Description

We have performance problems with the join queries for the `Channels` table.

Example with the join between `customer` and `channels`.
The [getAssignedEntityChannels](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/services/channel.service.ts#L138) function will generate this sql query:

```
SELECT "channel"."channelId" AS "channelId" FROM "customer" "Customer", "customer_channels_channel" "channel" WHERE "channel"."customerId" = $1
```

|  Env     |   Rows in the customer table    |    execution time   | 
|---    |:-:    |--:    |
|   test    |   1 500    |    625ms   |     
|   prod    |   1 700 000    |   >40s    |     

Here is the new query to filter the entity table and not retrieve all the rows, which causes performance problems

```
SELECT "channel"."channelId" AS "channelId" FROM "customer" "Customer", "customer_channels_channel" "channel" 
WHERE "channel"."customerId" = $1
and "Customer"."id" = $1;
```

|  Env     |   Rows in the customer table    |    execution time   | 
|---    |:-:    |--:    |
|   test    |   1 500    |    34ms   |     
|   prod    |   1 700 000    |   33ms    |     

In pairs with @ttournie & @gdarchen 

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/channel.service.ts
+++ b/packages/core/src/service/services/channel.service.ts
@@ -166,6 +166,7 @@ export class ChannelService {
             .select(`channel.${inverseJunctionColumnName}`, 'channelId')
             .from(junctionTableName, 'channel')
             .where(`channel.${junctionColumnName} = :entityId`, { entityId })
+            .andWhere(`${entityType.name}.id = :entityId`, { entityId })
             .execute();
     }
 


packages/core/src/service/services/channel.service.ts
```diff
@@ -166,6 +166,7 @@ export class ChannelService {
             .select(`channel.${inverseJunctionColumnName}`, 'channelId')
             .from(junctionTableName, 'channel')
             .where(`channel.${junctionColumnName} = :entityId`, { entityId })
+            .andWhere(`${entityType.name}.id = :entityId`, { entityId })
```
gdarchen, line 169: ⛏ **suggestion:** use the table name instead of the entity name

```suggestion
            .andWhere(`${metadata.tableName}.id = :entityId`, { entityId })
```

michaelbromley:
Thanks for the contribution. I suspect that this may be solving the same issue as

- https://github.com/vendure-ecommerce/vendure/pull/2791

Can you let me know what you think? @monrostar can you weigh in?

alexisvigoureux:
Thx for your answers, @monrostar your PR is better, I will close mine

PR #2811, opened 2024-04-25
fix(dev-server): Update postgres credentials in docker compose

davidhoeck:
# Description

This PR changes the Postgres database name and the credentials in the docker compose to be the same as in the `dev-config.ts`

# Breaking changes

none

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dev-server/docker-compose.yml
+++ b/packages/dev-server/docker-compose.yml
@@ -62,9 +62,9 @@ services:
     image: postgres:12.3
     container_name: postgres
     environment:
-      POSTGRES_DB: postgres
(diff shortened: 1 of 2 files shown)


packages/dev-server/docker-compose.yml
```diff
@@ -62,9 +62,9 @@ services:
     image: postgres:12.3
```
oidt, line 62: Isn't it better to go to latest 16 version for testing?

michaelbromley: Yes, good point but can be a separate PR 👍

PR #2814, opened 2024-04-26
feat(admin-ui): Migrate Clarity to Tailwind

dlhck:
# Description

This PR removes the styles of the Clarity design system and replaces them with Tailwind. Styling is done in a backwards compatible way, which means that all clarity classes will be removed. 

> Important: This is heavy WIP and more like a PoC right now. We are especially thinking about backwards compatibility. 

Relates to https://github.com/vendure-ecommerce/vendure/issues/2903

# Breaking changes

There could be some breaking changes, if extensions rely on specific Clarity classes which were not used in the Admin UI initially.

# Checklist

📌 Always:

👍 Most of the time:


nino-magazinovic:
@davidhoeck Cool, waiting for the release.

nino-magazinovic:
@davidhoeck The solution should have established marketing, business support, Figma design system, community, example pages and components to acceleration development on admin and storefront, integration to multiple frontend frameworks, fully expandable.

Default theme should be clean black/white like in shadcn.

Options:
- Flowbite
- daisyUI

Resources:
- https://www.tailwindresources.com/
- https://www.builtatlightspeed.com/category/tailwind

In any case, do not change the Nestjs and Angular stack 🥇

davidhoeck:
@nino-magazinovic thanks for the great input. We are already looking at Spartan, because it provides the same approach as Shadcn does. Only problem is that they are still in alpha, which makes it unclear for me if it is mature enough to use in our platform.

mschipperheyn:
As a general suggestion for the overall setup I suggest using the ability of Tailwind to set up aliases for your colors in line with Material ui, makes it easier to modify this down the line. It's also relevant to keep track of dark mode from day one. If you can use a library that offers all of that stuff out of the box, even better. 

We use `primary`, `secondary`, `warn`, `highlight` and `error` classes as aliases for our color scheme. We also use e.g. `primary-contrast: white` as a way to define the text colors that sit on top of buttons and such. That way you can set className="bg-primary-600" instead of className="bg-blue-600". You can easily set that up in tailwind.config.

One of the biggest challenges I found in this space is finding a good date-time picker. This is a great direction bc it has great keyboard support: https://shadcnui-expansions.typeart.cc/docs/datetime-picker but shadcn even lacks an out of the box year selector, even though there are some good suggestions to implement that in the issues.

davidhoeck:
@mschipperheyn yes, that's an absolute must. I also like the approach of descriptive color names way better, than just the color name itself.

PR #2816, opened 2024-04-27
Move `typescript` to dev dependencies

coopbri:
# Description

During the bootstrapping process of `@vendure/create`, the `typescript` package is installed as a regular dependency, I moved it to the dev dependencies. 

~~Regarding `dotenv`, it appears it's only used in the CLI, dev server, and in E2E tests for the payments plugin, please let me know if I missed something and it should be kept as a prod dep.~~

# Breaking changes

No

# Screenshots

N/A

# Checklist

### DIFF
--- a/packages/create/src/helpers.ts
+++ b/packages/create/src/helpers.ts
@@ -261,11 +261,14 @@ export function getDependencies(
         `@vendure/email-plugin${vendurePkgVersion}`,
         `@vendure/asset-server-plugin${vendurePkgVersion}`,
         `@vendure/admin-ui-plugin${vendurePkgVersion}`,
-        'dotenv',
         dbDriverPackage(dbType),
+    ];
+    const devDependencies = [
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,

The reason that dotenv is a regular dependency is that in the vendure-config.ts file, it is used to automatically load env vars which are used in the config. Moving it to a dev dependency would require a change to how env vars are handled there.

I _think_ moving typescript to a devDependency is fine though. I can't off the top of my head think of why that would be an issue.

coopbri:
> Hi,
> 
> The reason that dotenv is a regular dependency is that in the vendure-config.ts file, it is used to automatically load env vars which are used in the config. Moving it to a dev dependency would require a change to how env vars are handled there.
> 
> I _think_ moving typescript to a devDependency is fine though. I can't off the top of my head think of why that would be an issue.

Revised!

michaelbromley:
Thank you!

PR #2839, opened 2024-05-07
fix(cookie): Use the shop cookie name for default route

alexisvigoureux:
# Description

In addition to this PR
https://github.com/vendure-ecommerce/vendure/pull/2482

Here is the cookie naming behavior:

- If not setting the authOptions.cookieOptions.name parameter:
  -  Both the Admin API and Shop API cookies names will be the default one: 'session'
- If setting the authOptions.cookieOptions.name parameter to something like 'cookie-name':
  - Both the Admin API and Shop API cookies names will be the configured one: 'cookie-name'
- If setting the authOptions.cookieOptions.name parameter to something like `{shop: 'cookie-name', admin: 'admin-cookie-name'}`:
  - The Admin API cookie name will be 'admin-cookie-name'
  -  Shop API cookie name will be 'cookie-name'
  - And all the REST endpoints on the `/` route will use the default cookie: 'session' 
    -  context: some systems do not support Graphql and so we have a few REST endpoints. This PR allows to use the same shop cookie name for graphql and REST routes
    

# Breaking changes

I think it's an unintended behaviour that can be released in the next patch

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/bootstrap.ts
+++ b/packages/core/src/bootstrap.ts
@@ -250,7 +250,7 @@ function checkPluginCompatibility(config: RuntimeVendureConfig): void {
             if (!satisfies(VENDURE_VERSION, compatibility, { loose: true, includePrerelease: true })) {
                 Logger.error(
                     `Plugin "${pluginName}" is not compatible with this version of Vendure. ` +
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks!

PR #2855, opened 2024-05-16
feat(payments-plugin): Allow mollie orders with $0

### LINKED ISSUE #2821: Mollie doesn't allow orders with €0
**Describe the bug**
When the order total is €0, Mollie returns an error `Amount is lower than the minimum`, but it should transition to PaymentSettled. This can happen when a customers uses a gift card for example

**To Reproduce**
1. Create a promotion that discounts €10 on every order
2. Add an item to cart resulting in the order total being €0 after the discount is applied
3. call `createMolliePaymentIntent`

**Expected behavior**
The Mollie plugin should return the configured thank-you or order-confirmation url, and the order should have transitioned to PaymentSettled

 - @vendure/core version: 2.2
 - Nodejs version 20

**Additional context**
I can pick this up :-)

martijnvdbrug:
# Description

Transition orders to PaymentSettled when the total is $0. The payment intent returns the given thank-you/confirmation page instead of the link to Mollie, because we don't need to redirect the customer to the Mollie platform.

Closes #2821

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/payments-plugin/e2e/graphql/admin-queries.ts
+++ b/packages/payments-plugin/e2e/graphql/admin-queries.ts
@@ -114,3 +114,14 @@ export const CREATE_CHANNEL = gql`
         }
     }
 `;
+
+export const CREATE_COUPON = gql`
+    mutation CreatePromotion($input: CreatePromotionInput!) {
+        createPromotion(input: $input) {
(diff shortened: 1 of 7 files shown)


michaelbromley:
Thanks!

Issue #2864, opened 2024-05-23
Entity Hydrator randomly mixes nested relations


jnugh:
**Describe the bug**
This looks to be similar to #2013. I hope this is not too confusing - we spent two days figuring out the root cause of our issue:

When having an Entity that has two levels of nested 1:N relations, let's say for instance
`Order.lines[0].productVariant` and `Order.lines[0].sales[0]` we are able to mix the sub relations in a way that e.g. `Order.lines[0].sales[0]` would actually belong to `Order.lines[1].sales[0]` (in our case this happened with a custom entity but as far as I can tell).

**To Reproduce**
In order to do this the entity needs to be loaded with one of the references joined:
```ts
const order = await repo.findOne({
  where: { id: '123' },
  relations: ['lines', 'lines.sales']
});
```
Now we use the entity hydrator to fetch the other relation:
```ts
await entityHydrator.hydrate(ctx, order, {
  relations: ['lines', 'lines.sales', 'lines.productVariant']
});
```

the hydrator correctly identifies `productVariant` as a missing relation and generates a query that joins `lines` and `productVariant`. The result is then being merged using `mergeDeep`. https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts#L343

The merge reaches this branch:
```ts
 if (Array.isArray(value)) {
  (a as any)[key] = value.map((v, index) =>
    this.mergeDeep(a?.[key]?.[index], b[key][index]),
  );
}
```

which needs the items in `a` and `b` to be in the same order. In this case `a` and `b` would be the original and newly fetched `lines` arrays. The order is however not guaranteed to be the same, leading to randomly mixed relations inside `lines`.

**Expected behavior**
I wrote a quick test to simulate this. It should pass:
```ts
describe('mergeDeep()', () => {

    it('should sync the order of sub relations',  () => {
        const prefetched = new Order({
            lines: [{
                id: 'line1',
                sales: [new Sale({id: 'sale-of-line-1'})],
            }, {
                id: 'line2',
                sales: [new Sale({id: 'sale-of-line-2'})],
            }]
        })

        const hydrationFetched = new Order({
            lines: [{
                id: 'line2',
                productVariant: {id: 'variant-of-line-2'},
            }, {
                id: 'line1',
                productVariant: {id: 'variant-of-line-1'},
            }]
        });

        const merged = mergeDeep(prefetched, hydrationFetched);
[...]

PR #2865, opened 2024-05-23
fix(core): Keep order of nested relations during hydration (#2864)

jnugh:
# Description
Relations could get out of sync during entity hydration. See #2864 for a detailed description and example.

I moved `mergeDeep` out of the service to make testing easier. I think adding an e2e Test for this does not make sense as it would be very hard to control the result order of the hydration database query. So I decided to unit test `mergeDeep` instead.

The only actual change is:
```ts
                if (Array.isArray(value)) {
                    (a as any)[key] = value.map((v, index) =>
                        this.mergeDeep(a?.[key]?.[index], b[key][index]),
                    );
                } else if (isObject(value)) {
                    (a as any)[key] = this.mergeDeep(a?.[key], b[key]);
                }
```
changed to:
```ts
            if (Array.isArray(value) || isObject(value)) {
                (a as any)[key] = mergeDeep(a?.[key], b[key]);
            }
```

I don't see a case where calling mergeDeep would not work because mergeDeep still executes the same logic inside the recursion by iterating using `for (const [key, value] of Object.entries(b))` and assigning the values using `(a as any)[key] = b[key];`, event if `a` and `b` do not contain entities.

# Breaking changes
None

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
@@ -1,6 +1,5 @@
 import { Injectable } from '@nestjs/common';
 import { Type } from '@vendure/common/lib/shared-types';
-import { isObject } from '@vendure/common/lib/shared-utils';
 import { unique } from '@vendure/common/lib/unique';
 import { SelectQueryBuilder } from 'typeorm';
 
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks for the detailed write-up & fix! 👍

PR #2880, opened 2024-06-06
fix(cookies): Apply split cookie session middleware per route

gdarchen:
# Description

When using split cookies to have different cookie name for the `shop-api` and the `admin-api`, we used to apply the `cookieSession()` middleware globally.

However, [Nest Global Middlewares](https://docs.nestjs.com/middleware#global-middleware) can't be applied to specific routes.
By doing Global Middlewares, the last applied `app.use()` middleware was overriding the above-mentioned ones.

Instead, we have to use [Functional Middlewares](https://docs.nestjs.com/middleware#functional-middleware) for the specific paths we like.

As a consequence, for the case when the user wants to split the cookies, we do not do it in the `bootstrap()` function any longer but instead we do it in the `AppModule.configure()` method, where other middlewares route-specific are already applied.

# Breaking changes

There is no breaking change as we just fix a behaviour that was not working.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/app.module.ts
+++ b/packages/core/src/app.module.ts
@@ -1,4 +1,5 @@
 import { MiddlewareConsumer, Module, NestModule, OnApplicationShutdown } from '@nestjs/common';
+import cookieSession from 'cookie-session';
 
 import { ApiModule } from './api/api.module';
 import { Middleware, MiddlewareHandler } from './common';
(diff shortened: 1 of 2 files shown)


michaelbromley:
Thanks for the fix 👍

PR #2900, opened 2024-06-15
feat(product-duplicator): Copy prices in product duplicator

mschipperheyn:
# Description

When you copy products using the product-duplicator all variants get a price of 0. This PR fixes that by also copying the prices.

I'm not sure if the original behavior was intentional or a bug. I considered it the latter. But if it was intentional, I can create an additional flag to ensure backward. compatibility.

# Breaking changes

- Copied product variant prices are no longer 0

# Checklist

📌 Always:
- [ x] I have set a clear title

👍 Most of the time:


mschipperheyn:
I'm seeing some issues in the files I committed. Some changes occured in files I didn't touch. I will review and update the PR

davidhoeck:
As far as I can see, this would set the price of all variants to the price of the first variant in the array. I don't think that is good, as we are already looping through the variants at that point.

mschipperheyn:
I don't think that is the case. It's taking the price from the first price in the price array per variant. But I'll have a look.

davidhoeck:
True, everything ok! Mistake on my side 😁

michaelbromley:
Yes, the current behaviour is a bug indeed, thanks for picking this up! I'm wondering is we can make this even smarter: in this change, we take the 1st price in the `productVariantPrices` array. If we have a multi-channel setup, that price might not be associated with the current active channel, which may lead to an unexpected result.

We could do something like

```ts
const price = variant.productVariantPrices.find(p => idsAreEqual(p.channelId, ctx.channelId)?.price ?? /* fallback to 1st item */
```

Going even further, it could make sense to allow copying _all_ prices in the current channel (for the case that there are multiple currencies set up for this channel). But this is out-of-scope for this bugfix, and could be considered as a potential feature for a future release.

mschipperheyn:
I'll review and update. Thanks for the feedback!

PR #2905, opened 2024-06-16
fix(create): Supporting SSL enforced PostgreSQL databases

### LINKED ISSUE #2904: PostgreSQL Databases with SSL enforced not supported by create and populate
**Describe the bug**
If e.g. an external database is used which requires SSL, @vendure/create fails once connecting and when populating. 

**To Reproduce**
1. npx @vendure/create test
2. Select PostgreSQL
3. Add a db connection which requires ssl

**Expected behavior**
SSL should be an option

 - @vendure/core version: 2.2.5
 - @vendure/create version: 2.2.5
 - Nodejs version: 18/20

**Additional context**
I made a fork to fix this

tomh4:
# Description

If a db has mandatory ssl when connecting the current @vendure/create fails.
Implemented SSL Option for PostgreSQL
Added SSL option to populate function
fixes #2904 

# Breaking changes

No

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/create/src/gather-user-responses.ts
+++ b/packages/create/src/gather-user-responses.ts
@@ -76,7 +76,8 @@ export async function gatherUserResponses(
     const dbSSL =
         dbType === 'postgres'
             ? await select({
-                  message: 'Use SSL to connect to the database?',
+                  message:
+                      'Use SSL to connect to the database? (only enable if you database provider supports SSL)',
                   options: [
                       { label: 'no', value: false },
                       { label: 'yes', value: true },
--- a/packages/create/src/gather-user-responses.ts
+++ b/packages/create/src/gather-user-responses.ts
@@ -14,6 +14,7 @@ interface PromptAnswers {
     dbSchema?: string | symbol;
     dbUserName: string | symbol;
     dbPassword: string | symbol;
+    dbSSL?: boolean | symbol;
     superadminIdentifier: string | symbol;
     superadminPassword: string | symbol;
     populateProducts: boolean | symbol;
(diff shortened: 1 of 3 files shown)


michaelbromley:
Thanks for this PR - it's a good addition. 2 things:

1. Can you make the PR against the `minor` branch, since this is a new feature?
2. I'm wondering whether the question might become a source of confusion when people don't know whether they need SSL or not. They might think "oh SSL is more secure, I'd better set it to true", and then the connection will fail. What do you think?

tomh4:
Hello! 

1. Yes I will do that
2. That might happen, but I already made 'no' the default. I might add a note to only use SSL if the db requires it and also add a more detailed explanation to the readme?

michaelbromley:
Yeah, the "false" default is good, and yes probably just a note as suggested along the lines of "(only enable if you database provider supports SSL)" will be enough. 

A further enhancement (which you don't _need_ to add to this PR, but can if you like), would be to detect SSL connection errors during bootstrap of the server, and display an instruction to try again with the SSL option turned on.

tomh4:
I have added the error handling and description, now i need to find out on how to change the branch towards which I want to PR against 😄

michaelbromley:
Thanks! Does this help? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

tomh4:
@michaelbromley ah that easy, thanks 😄  I changed it, anything else I need to do?

michaelbromley:
Nope, we're all good! Thanks for the contribution. This will be made available in the next minor release (v2.3)

PR #2911, opened 2024-06-19
fix(cookies): Correctly apply global and per-route middlewares

gdarchen:
# Description

In a previous PR, I tried to fix an issue causing some per-route middlewares being broken when using the split cookies for the Admin and Shop APIs.
- #2880 

But since it was released in the `2.2.6`, the cookies split is no longer working at all.

After investigating, it turned out we were still badly applying the per-route middleware:
- We did not set a global middleware in the `bootstrap.ts` file
- We were overriding the per-route middlewares in the `app.modile.ts` file because of the configuration on the `/` route

In this PR, I made sure to:
1. Apply a global cookies middleware, either using the right cookie name:
   - If the cookies have to be split, use the `shop` cookie name
   - Else, use the shared cookie name if provided
   - If not provided, default to `session`
2. Only apply specific `cookieSession` middlewares on the `/shop-api` and `/admin-api` routes

# Breaking changes

No breaking change.

# Screenshots

https://github.com/vendure-ecommerce/vendure/assets/17927632/19fcbb66-513e-4072-bfe7-0b7d85eaa869

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/app.module.ts
+++ b/packages/core/src/app.module.ts
@@ -56,10 +56,6 @@ export class AppModule implements NestModule, OnApplicationShutdown {
                 handler: cookieSession({ ...cookieOptions, name: shopApiCookieName }),
                 route: shopApiPath,
             });
-            allMiddleware.push({
-                handler: cookieSession({ ...cookieOptions, name: shopApiCookieName }),
-                route: '/',
-            });
         }
 
         const consumableMiddlewares = allMiddleware.filter(mid => !mid.beforeListen);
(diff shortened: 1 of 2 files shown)


gdarchen:
> [!NOTE]
> @michaelbromley 
> I made a video to prove I could make it work properly on the `dev-server`.
> But that would be awesome if we could deploy like a _canary_ version from a pull request (based on a label?) to test this version on our actual projects before merging it.
> Do you think it's something doable?

gdarchen:
Hi @michaelbromley 👋 
I just tested a deployment on our _staging_ environment using the canary version you released and it seems to work as expected!

choim4389:
When will it be distributed to a stable version? This issue is an important feature of our product.

michaelbromley:
Likely this week.

Issue #2465, opened 2023-10-18
Customizing the RefundProcess


michaelbromley:
### Discussed in https://github.com/vendure-ecommerce/vendure/discussions/2433

<div type='discussions-op-text'>

<sup>Originally posted by **monrostar** October  9, 2023</sup>
When developing multivendors we faced the problem that we were limited in modifying refunds from multivendors’ channels, if we wanted to limit any refunds that involved local banks and required verification from the main channel. 

I think it would be a good idea to create the same refund process as the Order or Payment to customize the refund flow.

@michaelbromley @martijnvdbrug 

</div>

PR #2942, opened 2024-07-10
feat(core): Modified refundProcess to be extensible

dhrtn1006:
# Description

Unlike other state machines, the 'refund state' was previously implemented in a way that made extension impossible without modifying the core. I have modified it to allow for extension

- [Discord](https://discord.com/channels/1100672177260478564/1217663910572724335)
- [Issue](https://github.com/vendure-ecommerce/vendure/issues/2465)

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/config/config.module.ts
+++ b/packages/core/src/config/config.module.ts
@@ -14,7 +14,10 @@ import { ConfigService } from './config.service';
     exports: [ConfigService],
 })
 export class ConfigModule implements OnApplicationBootstrap, OnApplicationShutdown {
-    constructor(private configService: ConfigService, private moduleRef: ModuleRef) {}
+    constructor(
+        private configService: ConfigService,
+        private moduleRef: ModuleRef,
+    ) {}
 
     async onApplicationBootstrap() {
         await this.initInjectableStrategies();
@@ -106,6 +109,7 @@ export class ConfigModule implements OnApplicationBootstrap, OnApplicationShutdo
         const { entityIdStrategy } = this.configService.entityOptions;
         const { healthChecks, errorHandlers } = this.configService.systemOptions;
         const { assetImportStrategy } = this.configService.importExportOptions;
+        const { process: refundProcess } = this.configService.refundOptions;
         return [
             ...adminAuthenticationStrategy,
             ...shopAuthenticationStrategy,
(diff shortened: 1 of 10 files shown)


michaelbromley:
Hi!

Thank you, this is a _very_ good PR. Not an easy one either as your first Vendure contribution - it's dealing with one of the more complex parts of the Vendure internals. But you've done it exactly as I would.

I have _one_ reservation about merging this as-is though: I'm not really happy with creating a new `refundOptions` top-level config object just to store the refund process. I think I know why you did it that way - it is consistent with the other state machine configs `<type>Options.process`.

But logically I think that refund stuff belongs with the `paymentOptions` object - you cannot have a refund without a payment, they are part of the same larger process in fact. So in that respect I think it should go in `paymentOptions`. But then we have the problem of what to name the property. We of course can't use `process` because this is already taken by the payment process. 

In hindsight I think I should have named it `paymentProcess`, then it would be trivial to just add `refundProcess` and everything is nice and consistent. But as it stands, I still feel that having `process` and `refundProcess` - although not exactly consistent - is still preferable to having a totally separate `refundOptions` object.

What are your thoughts on this?

dhrtn1006:
@michaelbromley 

Hi!

Thank you for your positive feedback on this PR.

As you mentioned, this PR extends the refund process by creating a top-level config object called refundOptions, aiming to maintain consistency with other state machine configurations.

I agree with your opinion, and I also believe that we should name it refundProcess and apply this naming convention to other state machine configurations in the future.

However, as you mentioned, this is my first Vendure contribution. Therefore, I thought it would be a better choice to create a refundOptions config object rather than changing the convention based on my personal judgment.

michaelbromley:
OK here's what we do: remove the `refundOptions`, and add `paymentOptions.refundProcess`.

In a future release we will deprecate the `process` properties in favour of the more descriptive `paymentProcess` etc. which we can add at a later time.

michaelbromley:
Thank you!

dhrtn1006:
@michaelbromley 

As discussed, I have added a commit to move refundProcess under paymentOptions. Renaming other state processes might be burdensome, but I am available to help if you decide to proceed with those changes.

PR #2950, opened 2024-07-16
Feat/email load globalTemplateVars async

mschipperheyn:
# Description

This PR adds an async function call as an alternative to the static `globalTemplateVars` in the Email-Plugin. The function receives the `RequestContext` and the `Injector`. It has the objective to allow you to load theme variables that may be specific to e.g. a channel. Such as a footer text or color schemes, etc

I did get a whole bunch of eslint related warnings when I committed this and had to use --no-verify. But they didn't apply to my code. 

I'm assuming documentation is extracted from the inline docs I added. Not sure how that works.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:
- [ x] I have updated the README if needed


packages/email-plugin/src/types.ts
```diff
@@ -31,6 +31,41 @@ export type EventWithContext = VendureEvent & { ctx: RequestContext };
...
+ *          return theme;
+ *      }
+ *   [...]
+ * })
+ *
+ * ```
+ *
+ * @docsCategory core plugins/EmailPlugin
+ * @docsPage EmailPluginOptions
+ * @docsWeight 0
```
michaelbromley, line 62: Setting `docsWeight 0` will force this type to the top of the `EmailPluginOptions` docs page, which is not what we want. Better to have the main `EmailPluginOptions` interface at the top. So remove this to be safe add it to the `EmailPluginOptions` doc block to ensure it always stays at the top.

michaelbromley (changes requested):
Excellent work: clean code, good test case, excellent docs 👍 

One small change needed on a doc block, and also please target the `minor` branch rather than `master`.

michaelbromley:
Since the master will be merged into minor today anyway with the upcoming release, I decided to merge anyway, since the time-zone difference would delay things.

Thanks for the contribution!

PR #2959, opened 2024-07-18
fix(core): Truncate db aliases if necessary

### LINKED ISSUE #2899: EntityHydrator does not shorten jointable alias names anymore
**Describe the bug**
While upgrading our vendure project to vendure v2.2, we ran into a typeorm QueryFailedError that suggested there were tablenames in our queries that were specified more than once.

After some debugging i tracked the issue down to the following:

This [PR](https://github.com/vendure-ecommerce/vendure/pull/2742) changed the EntityHydrator to use typeorms SelectQueryBuilder instead of the Repository to hydrate entities. The SelectQueryBuilder does not [shorten](https://github.dev/typeorm/typeorm/blob/c4a36da62593469436b074873eba186f0f8b990d/src/util/StringUtils.ts#L76) alias names of joined relations, like the Repository does. When hydrating entities with very deeply nested relations (relation string resulting in a jointablename > 63 characters) now leads to alias names being cut off after 63 characters because of the alias string limitation of 63 characters in typeorms postgres driver. This leads to the errors we encountered in our project.

**To Reproduce**
1. Try using the EntityHydrator with a very long relations string (> 63 characters) and you should run into said typeorm QueryFailedError.

**Expected behavior**
The EntityHydrator should be usable, regardless of the size of relation strings it is given, or at least, if the behavior is not changeable, check for length and throw an error indicating relation strings that are too long, which would obviously limit the buildable datamodels in vendure drastically, if you still want to use the EntityHydrator regarding performance.

 - @vendure/core version: v2.2.5
 - Nodejs version: v20.13.1

jnugh:
Fixes #2899

# Description
Postgres truncates long identifiers making them equal if they share the same prefix. For instance if two tables are being joined with a long alias (the names need to be a lot longer in reality) `product_variant_relation_0_relation_1` and `product_variant_relation_0_relation_2` postgres will truncate both aliases to `product_variant_rel` and fail because it's not allowed to have two join statements with the same target alias.

# Breaking changes

None

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/entity-hydrator.e2e-spec.ts
+++ b/packages/core/e2e/entity-hydrator.e2e-spec.ts
@@ -22,7 +22,11 @@ import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 import { initialData } from '../../../e2e-common/e2e-initial-data';
 import { testConfig, TEST_SETUP_TIMEOUT_MS } from '../../../e2e-common/test-config';
 
-import { AdditionalConfig, HydrationTestPlugin } from './fixtures/test-plugins/hydration-test-plugin';
+import {
+    AdditionalConfig,
+    HydrationTestPlugin,
+    TreeEntity,
+} from './fixtures/test-plugins/hydration-test-plugin';
 import { UpdateChannelMutation, UpdateChannelMutationVariables } from './graphql/generated-e2e-admin-types';
 import {
     AddItemToOrderDocument,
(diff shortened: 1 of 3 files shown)


jnugh:

michaelbromley:
Hi,

Thanks for this contribution!

Since this is quite a low-level modification, and heavily subject to the TypeORM internals, I'm wondering if there is some way we can add a test for this that:

1. Fails before your modification
2. Passes after it
3. Will therefore detect regressions in future changes to TypeORM

Do you think this is something you can look into?

jnugh:
Sorry for the long delay. I implemented a test case, inside the entity hydrator test suite. Without the fix the test will fail with:

```
Serialized Error: { response: { errors: [ { message: 'table name "TreeEntity_childrenPropertyWithAVeryLongNameThatExceedsPostgres" specified more than once', locations: [ { line: 2, column: 3 } ], path: [ 'hydrateChannelWithVeryLongPropertyName' ] } ], data: { hydrateChannelWithVeryLongPropertyName: null }, status: 200 }, request: { query: 'query GetHydratedChannelNested($id: ID!) {\n  hydrateChannelWithVeryLongPropertyName(id: $id)\n}', variables: { id: 'T_1' } } }
```

michaelbromley:
Excellent work, thanks so much!

PR #2961, opened 2024-07-18
perf(core): Improve hydrator performance for customFields

### LINKED ISSUE #2696: EntityHydrator always refetches customFields
**Describe the bug**
The `EntityHydratorService` checks wether or not certain fields need to be fetched from the database or are already available locally. Therefore the relation path is split by dots and each segment is being checked for existence.

However there is a special case for `customFields`:
```ts
const parts = !relation.startsWith('customFields') ? relation.split('.') : [relation];
```

In case of custom fields the complete path would need to exist (e.g. `entity['customFields.myCustomField']` instead of `entity.customFields.myCustomField`.
I removed the special case and the code seems to be working just fine. Maybe this was related to a bug in a previous version of TypeORM?

**To Reproduce**
1. Create a custom field
2. Fetch the base entity including the custom field
3. Use entityHydrator to hydrate the custom field

**Expected behavior**
The first query should be sufficient however the relation is being fetched again.

 - Nodejs version: 20

**Additional context**
I also noticed that a relation will always marked as not fetched if there is a null value on the path because null would mean the relation has been fetched but does not exist (in contrast to undefined).

If this was indeed a workaround for an old bug I could create a PR for this.

jnugh:
Remove a workaround that led to refetching relations in customFields. Prevent refetching of customFields during hydration
Fixes #2696

# Description
See #2696. There was some sort of workaround that disabled the check which is in place to prevent reloading the same data over and over. `entity["customFields.xyz"]` will never be set because the data would be in `entity["customFields"]["xyz"]`.
I'm still wondering if there is some edge case that made this workaround necessary but we are running this patch in production for quite some time now and we are using a lot of `customFields`.

# Breaking changes
None I would be aware of. This might have been a workaround for a bug in typeorm maybe.

# Checklist

📌 Always:
- [x I have set a clear title

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
@@ -200,11 +200,16 @@ export class EntityHydrator {
         const missingRelations: string[] = [];
         for (const relation of options.relations.slice().sort()) {
             if (typeof relation === 'string') {
-                const parts = !relation.startsWith('customFields') ? relation.split('.') : [relation];
+                const parts = relation.split('.');
                 let entity: Record<string, any> | undefined = target;
                 const path = [];
                 for (const part of parts) {
                     path.push(part);
+                    // null = the relation has been fetched but was null in the database.
+                    // undefined = the relation has not been fetched.
+                    if (entity && entity[part] === null) {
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks!

Issue #2956, opened 2024-07-18
Add maxDiscountAmount to PromotionItemAction


BM-JiHoonPark:
**Is your feature request related to a problem? Please describe.**
I am implementing most of the discount features through promotions. I want to run various types of discount promotions, and I believe that product discount coupons with a maximum discount amount limit will be effective.

**Describe the solution you'd like**
Currently, for PromotionItemAction, the final discount amount is calculated by multiplying the return value of execute by the quantity.

If the return value of execute is divided by the quantity and returned, a fixed amount discount can be achieved.

However, this can cause issues if the return value of execute and the quantity do not divide evenly.

For example, if the maximum discount amount is `100000` and the quantity is `11`, when the return value of execute is `100000 / 11`, the amount in the adjustment is stored as `-99999.99999999999`.

I believe this could cause long-term issues and would prefer to store the maximum discount amount separately to be used in calculations.

**Describe alternatives you've considered**
Create a new column in the promotion table or add a new parameter to the execute method of PromotionItemAction. This will ensure that when calculating the amount of the adjustment in the applyOrderItemPromotions method, the value stored is `Math.min(adjustment.amount * line.quantity, maxDiscountAmount)` instead of `adjustment.amount * line.quantity`.

LavaSean:
retweet

PR #2971, opened 2024-07-25
feat(core): Create PromotionLineAction

dhrtn1006:
# Description
To address the issue mentioned in the previous PR, I modified the return type of the `execute` method in `PromotionItemAction` to return either a `Number` or `ExecutePromotionActionResult`, allowing the inclusion of `discountMode`.

However, to maintain code consistency, I have added `PromotionLineAction` and adjusted all `PromotionActions` to return a `Number` type.

Other tasks follow the contents of the attached previous PR.

- [Issue](https://github.com/vendure-ecommerce/vendure/issues/2956)
- [Previous PR](https://github.com/vendure-ecommerce/vendure/pull/2964)

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/entity/promotion/promotion.entity.ts
+++ b/packages/core/src/entity/promotion/promotion.entity.ts
@@ -159,9 +159,10 @@ export class Promotion
             if (promotionAction instanceof PromotionItemAction) {
                 if (this.isOrderItemArg(args)) {
                     const { orderLine } = args;
-                    amount +=
-                        roundMoney(await promotionAction.execute(ctx, orderLine, action.args, state, this)) *
-                        orderLine.quantity;
+                    amount += roundMoney(
+                        await promotionAction.execute(ctx, orderLine, action.args, state, this),
+                        orderLine.quantity,
+                    );
                 }
             } else if (promotionAction instanceof PromotionLineAction) {
                 if (this.isOrderLineArg(args)) {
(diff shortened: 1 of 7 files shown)


packages/core/src/entity/promotion/promotion.entity.ts
```diff
@@ -149,6 +158,13 @@ export class Promotion
             const promotionAction = this.allActions[action.code];
             if (promotionAction instanceof PromotionItemAction) {
                 if (this.isOrderItemArg(args)) {
+                    const { orderLine } = args;
+                    amount +=
+                        roundMoney(await promotionAction.execute(ctx, orderLine, action.args, state, this)) *
```
michaelbromley: the `roundMoney()` function accepts a 2nd argument for quantity, so that the configured MoneyStrategy can control the multiplication.

michaelbromley (commented):
Looks good - just 1 comment on the rounding. If there is a good reason to do it this was please let me know, else pass the quantity in to the `roundMoney()` function.

packages/core/src/entity/promotion/promotion.entity.ts
BM-JiHoonPark: Thank you for the review.
It seems the issue arose because the previous quantity calculation logic was simply transferred.
I have added the calculation logic as an argument to `roundMoney`.

Feelw00: Thank you for the review.
It seems the issue arose because the previous quantity calculation logic was simply transferred.
I have added the calculation logic as an argument to `roundMoney`.

michaelbromley:
Thank you very much for your contribution and continued input in developing this feature!

PR #2984, opened 2024-07-31
fix(admin-ui): Change max height to min height for channel assignment block

monrostar:
# Description

Updated the block height for the Channel block `.channel-assignment`. Because there was a fixed height there, a large number of channels simply did not fit and the button to add new channels was hidden

# Breaking changes

* no

# Screenshots
Before
image
After
image

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/catalog/src/components/product-detail/product-detail.component.scss
+++ b/packages/admin-ui/src/lib/catalog/src/components/product-detail/product-detail.component.scss
@@ -53,7 +53,7 @@ vdr-action-bar clr-toggle-wrapper {
 
 .channel-assignment {
     flex-wrap: wrap;
-    min-height: 144px;
+    min-height: 24px;
 }
 
 .pagination-row {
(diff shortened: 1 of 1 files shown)


monrostar:
@michaelbromley 

I don't know why this fix hasn't been made yet, but it's so annoying :) 
Even though we don't use this functionality, I would like to see all channels and the button :)

michaelbromley:
Does this still look reasonable with only eg 2 channels assigned? Not too much white space?

monrostar:
> Does this still look reasonable with only eg 2 channels assigned? Not too much white space?
Yeah, that's a good question :) I'll update again

monrostar:
> > Does this still look reasonable with only eg 2 channels assigned? Not too much white space?
> 
> Yeah, that's a good question :) I'll update again

24px will be fine. This is almost minimum

image

michaelbromley:
Thanks!

PR #2990, opened 2024-08-01
feat(core) Config to disable Anonymous session creation 

arrrrny:
# Description
Added config to disable Anonymous session creation. This allows only authenticated users to add to cart. It helps preventing bot/scrapers to make operations on shop-api, allows only authenticated users to view products etc.

Default is false

# Breaking changes

No

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/middleware/auth-guard.ts
+++ b/packages/core/src/api/middleware/auth-guard.ts
@@ -50,13 +50,15 @@ export class AuthGuard implements CanActivate {
             return true;
         }
         const authDisabled = this.configService.authOptions.disableAuth;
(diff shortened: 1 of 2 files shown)


arrrrny:
I agree with that. I am launching UserCheckout plugin today which I must have this feature conjuction with that. Let me see if I can make it with a plugin. other than that I did not see any side effects of this. I spent last 3 days on Vendure Auth and User checkout learnt a lot.

arrrrny:
Main hickup I had with the system that Permission.Authenticated does not aspply to users system wide. It is tied to roles Customer and Admins. Any external auth shoul implement its role and Push Autheticated to that role which I will do that on firebase. User entity is not much of a thing without being a customer.

Issue #3008, opened 2024-08-13
Money cannot represent non numberic value NaN and discount for order line with 0 quantity


LavaSean:
**Describe the bug**
When try to modify orderline quantity to 0, it show error Money cannot represent non numberic value NaN

<img width="1437" alt="Screenshot 2024-08-13 at 12 41 38 PM" src="https://github.com/user-attachments/assets/b6369f24-38c9-4a2d-8c14-6c35c378751b">

Demo Video:

https://github.com/user-attachments/assets/5b1ee212-f320-478f-98ee-6f9d37274855

**To Reproduce**
1. Go to Promotion, create a discount where the condition is 'If order total is greater than 0' with amount 0 and actions 'Discount order by 15%' and value is 15%
2. Create an order with 2 order line and make payment
3. Modify the order and change the quantity of the lines to become 0, click preview changes

**Expected behavior**
1. Error will be thrown 'Money cannot represent non numberic value NaN'

**Environment (please complete the following information):**
 - @vendure/core version: 3.0.0
 - Nodejs version: 18.20
 - Database (mysql/postgres etc):

**Additional context**
I am writing my own resolver that remove certain items when some condition are met (like, customer want to remove the item and the item is yet to be fulfilled). When doing this flow without any promotion that’s on the order level, it works as expected, the error arise when we want to remove the item. Which causes an error of NaN because i am running the modifyOrder without the dryRun.

I have discovered a potential root cause that might solve this issue. In the applyOrderPromotion in OrderCalculator, there’s a code to get the share of amount of discount that each order line get which is the `const shareOfAmount = distribution[i]`; However, I noticed that for order line that dont have any quantity, the distribution will result in the value of undefined, to solve this, I found out that adding `?? 0` after the `const shareOfAmount = distribution[i];` which becomes `const shareOfAmount = distribution[i] ?? 0;` solves the issue.

michaelbromley:
Hi,

Thanks for the detailed report. Are you interested in contributing a fix for this? If so, that would be very welcome. Let me know also if you need any help putting together an e2e test for this scenario.

btw I noticed you are based in Penang. I'm writing this reply on the train from Penang -> Ipoh! Greetings!

jyling:
> Hi,
> 
> Thanks for the detailed report. Are you interested in contributing a fix for this? If so, that would be very welcome. Let me know also if you need any help putting together an e2e test for this scenario.
> 
[...]

PR #3009, opened 2024-08-13
Fix: solve a empty order line with discount modification NaN error

jyling:
# Description

This PR solves https://github.com/vendure-ecommerce/vendure/issues/3008, where there's an order level discount and we are updating the order to remove an order line by setting the quantity to 0. Currently, it throws a NaN error.

The solution is just to have a fallback of 0, when the `distribution[i]` returns undefined.

# Breaking changes

should not break existing functionality.

# Screenshots

### Note: I'm not sure why this test failed on my local machine, my changes shouldn't affect this test
image

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/order-modification.e2e-spec.ts
+++ b/packages/core/e2e/order-modification.e2e-spec.ts
@@ -482,73 +482,6 @@ describe('Order modification', () => {
             expect(modifyOrder.errorCode).toBe(ErrorCode.ORDER_LIMIT_ERROR);
             await assertOrderIsUnchanged(order!);
         });
-
-        it('adjustOrderLines empty quantity with discounts', async () => {
-            const PercentDiscount15Percent = '15PERCENT';
-            await adminClient.query<
-                Codegen.CreatePromotionMutation,
-                Codegen.CreatePromotionMutationVariables
-            >(CREATE_PROMOTION, {
-                input: {
-                    enabled: true,
-                    couponCode: PercentDiscount15Percent,
-                    conditions: [
-                        {
-                            code: 'minimum_order_amount',
-                            arguments: [
-                                {
-                                    name: 'amount',
-                                    value: '0',
-                                },
-                                {
-                                    name: 'taxInclusive',
-                                    value: 'false',
-                                },
-                            ],
-                        },
-                    ],
-                    actions: [
-                        {
-                            code: orderPercentageDiscount.code,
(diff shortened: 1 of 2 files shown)


jyling:

michaelbromley (changes requested):
Thanks again - I reviewed this part of the code more closely and while you have indeed prevented the error from being thrown, I discovered another subtle bug that also needs to be addressed (my bug, not yours :D)

In the e2e test, if we add the following lines at the end, it will fail:

```ts
            // ensure correct adjustments applied
            // The first line should have a linePrice of 0 because it has zero quantity
            expect(modifyOrder.lines[0].linePriceWithTax).toBe(0);
            expect(modifyOrder.lines[0].proratedLinePriceWithTax).toBe(0);
            // The second line should have the proratedLinePriceWithTax discounted per the promotion
            expect(modifyOrder.lines[1].proratedLinePriceWithTax).toBe(
                modifyOrder.lines[1].discountedLinePriceWithTax / 2,
            );
```

The reason for this is subtle: on line 252 of `order-calculator.ts` we are filtering out zero-quantity order lines. This then means that the `weights` array (and the derived `distribution` array) will have fewer elements than the `order.lines` array.

So when on line 255 we loop over all the order lines, this is where we run into the error of trying to get an out-of-range index from the `distribution` array. Defaulting to `0` as in your fix will prevent the `NaN` error, but is still not correct, which is why the added e2e test assertions will fail.

The solution is to remove the `.filter()` call on line 252, to ensure that we do not end up with a smaller array and that `distribution.length === order.lines.length` at all times.

In summary, I recommend the following changes to make this fix correct:

1. Add the e2e assertions as above to your existing test
2. Remove the filter call on line 252
3. Change the map call to `.map(l => (l.quantity !== 0 ? l.proratedLinePriceWithTax : 0));`
4. At this point you can also remove the `?? 0` default since the `distribution` array will always have the correct length.

jyling:
> In summary, I recommend the following changes to make this fix correct:

Thank you for the detailed explanation of the code that caused this bug, I will update the code and rerun the test.

jyling:
I implemented the proposed fixes, which indeed solved the NaN issue, however i encountered another issue where the discounts is missing
image

image

I will look into this later, but just to update on this PR

michaelbromley:
Tested locally and it looks good to me 👍

PR #3011, opened 2024-08-14
fix(core): Roles were not resolved on User entity

arrrrny:
# Description

User Entity was missing the Roles field resolved

# Breaking changes
No
# Screenshots
Screenshot 2024-08-14 at 10 36 39 AM

# Checklist

📌 Always:

👍 Most of the time:


packages/core/src/api/resolvers/entity/user-entity.resolver.ts
```diff
@@ -32,4 +33,14 @@ export class UserEntityResolver {
             strategy: m instanceof ExternalAuthenticationMethod ? m.strategy : NATIVE_AUTH_STRATEGY_NAME,
         }));
     }
+
+    @ResolveField()
+    async roles(@Ctx() ctx: RequestContext, @Parent() user: User): Promise<Role[]> {
+        let roles: Role[] = [];
+        if (user.roles) {
+            roles = user.roles;
```
michaelbromley: here you can just have an early return to prevent the rest of the function from executing unnecessary code:

```ts
if (user.roles) {
  return user.roles;
}
```

michaelbromley (changes requested):
Small change to prevent unnecessary code execution

arrrrny:
done

michaelbromley:
Thanks!

PR #3014, opened 2024-08-15
Fix/mollie failing webhook

### LINKED ISSUE #2941: Mollie doesn't call webhook
:warning: Use `"@vendure/payments-plugin": "2.1.8",` until this is issue is closed!

**Describe the bug**
After upgrading the Payments Plugin to v2.2.7, we are experiencing issues where Mollie doesn't call the Vendure webhook anymore. This only happens for orders that have their corresponding Mollie orders updated (see below for details).

**To Reproduce**
1. Add items to order (doesn't matter what)
2. Call the `createMolliePaymentIntent` mutation, and follow the link to the Mollie hosted checkout page
3. Go back (don't pay) form the Mollie hosted checkout page
4. Add some more items to your order in Vendure 
5. Create another Mollie intent and go to the Mollie Checkout again. This updates the Mollie Order using the [Manage Order Lines endpoint](https://docs.mollie.com/reference/manage-order-lines) in a single request
6. Mark the order as 'Paid' on the hosted checkout
7. You will now see an error screen, and Mollie did not call your webhook, while the actual transaction is processed.

image

### Workaround
Install payments-plugin 2.1.8

### Suspected cause
* [This feature (Updating/reusing Mollie Orders)](https://github.com/vendure-ecommerce/vendure/pull/2691) was implemented **and tested against the Mollie platform** in March 2024
* We also tested this feature (which was included in Vendure 2.2.7) on our testing environment **Juni 26**
* After going live to production (July 7) we noticed the error above
* ** July 1st iDeal 2.0 was released** We suspect the combination of the update-mollie-order-feature and the release of iDeal 2.0 is causing this behaviour

**Expected behavior**
I would expect the payment to settled, and Mollie to call the webhook.

 - @vendure/core version:2.2.7
 - Nodejs version 18

martijnvdbrug:
# Description

Fixes #2941

* We create new Mollie orders on every payment intent. No changing Mollie orders anymore. This was suggested by the Mollie engineers and is also how Shopify handles Mollie payments.
* We now log an error when a customer still manages to pay twice for an order.

# Breaking changes

Removed `order.customFields.mollieOrderId`, because we dont need a reference to Mollie anymore, but a migration is not per se needed, because its only a field removal.

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:


packages/payments-plugin/e2e/mollie-payment.e2e-spec.ts
```diff
@@ -215,6 +216,34 @@ describe('Mollie payments', () => {
...
+                    {
+                        languageCode: LanguageCode.en,
+                        name: 'Mollie payment test',
+                        description: 'This is a Mollie test payment method',
+                    },
+                ],
+            },
+        });
+        expect(createPaymentMethod.code).toBe(mockData.methodCode);
+    });
```
martijnvdbrug, line 245: This was moved to top, because payment method creation should happen before creating orders

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -291,6 +264,19 @@ export class MollieService {
...
+                p => p.transactionId === mollieOrder.id && p.state === 'Settled',
+            );
+            if (!paymentWithSameTransactionId) {
+                // The order is paid for again, with another transaction ID. This means the customer paid twice
+                Logger.error(
+                    `Order '${order.code}' is already paid. Mollie order '${mollieOrder.id}' should be refunded.`,
+                    loggerCtx,
+                );
+                return;
+            }
```
casperiv0, line 278: Is there a way we can tap into this error log and automatically report this to Sentry?

martijnvdbrug: What we have done in one of our projects is to create a `SentryLogger` that is a wrapper around a `VendureLogger`, which sends all error logs to Sentry. You could use a similar approach and filter out only this message. We could add a prefix/constant in the log message to make sure that is always in the error message.

```ts
// Error function in our SentryLogger class

  error(message: string, context?: string | undefined, trace?: string | undefined,
  ): void {
    Sentry.captureMessage(message, "error");
    this.logger.error(message, context, trace);
  }

```

Any alternative suggestions welcome, but I dont think there is any error emitting way of working in Vendure currently?

The alternative is to throw an error that you can handle somewhere, but this will also make Mollie retry this webhook until we return a status 200

casperiv0: I think the wrapper logger would work well, since I also need it for some other logs. Thanks for the suggestion!

martijnvdbrug: The log message is checked in the test. I will add a comment that the message shouldn't be changed because monitoring can depend on it.

michaelbromley:
Thanks for the fix! I'll be publishing a patch this week.

PR #3017, opened 2024-08-17
Feat/add email event UI triggers

monrostar:
# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/email-plugin/src/handler/default-email-handlers.ts
+++ b/packages/email-plugin/src/handler/default-email-handlers.ts
@@ -30,6 +30,15 @@ export const orderConfirmationHandler = new EmailEventListener('order-confirmati
         event =>
             event.toState === 'PaymentSettled' && event.fromState !== 'Modifying' && !!event.order.customer,
     )
+    .setUIOptions({
+        entityType: Order,
+        handler: (ctx, _injector, entity, _languageCode) => {
+            return new OrderStateTransitionEvent('PaymentAuthorized', 'PaymentSettled', ctx, entity);
+        },
+        filter: (ctx, injector, entity, languageCode) => {
+            return true;
+        },
+    })
     .loadData(async ({ event, injector }) => {
         transformOrderLineAssetUrls(event.ctx, event.order, injector);
         const shippingLines = await hydrateShippingLines(event.ctx, event.order, injector);
--- a/packages/email-plugin/src/handler/event-handler.ts
+++ b/packages/email-plugin/src/handler/event-handler.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { Type } from '@vendure/common/lib/shared-types';
-import { Injector, Logger } from '@vendure/core';
+import { Customer, Injector, Logger, Order } from '@vendure/core';
 
 import { serializeAttachments } from '../attachment-utils';
 import { loggerCtx } from '../constants';
@@ -8,6 +8,7 @@ import { EmailEventListener } from '../event-listener';
 import {
     EmailAttachment,
     EmailTemplateConfig,
+    EventUIHandlerOptions,
     EventWithAsyncData,
     EventWithContext,
     IntermediateEmailDetails,
@@ -148,6 +149,7 @@ export class EmailEventHandler<T extends string = string, Event extends EventWit
         bcc?: string;
     };
     private _mockEvent: Omit<Event, 'ctx' | 'data'> | undefined;
+    private _uiOptions?: EventUIHandlerOptions;
 
     constructor(
         public listener: EmailEventListener<T>,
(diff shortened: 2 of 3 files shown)


packages/email-plugin/src/email-event.resolver.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (!isRequestedType) return false; // TODO add error
+
+        const canResend = await handler.resendOptions.canResend(ctx, new Injector(this.moduleRef), entity);
+        if (!canResend) return false; // TODO add error
+
+        const event = await handler.resendOptions?.createEvent(
+            ctx,
+            new Injector(this.moduleRef),
+            entity,
+            args.input.arguments,
```
michaelbromley: Note that there is a case that needs to be handled here for arguments of the type `ID`. Since Vendure has the concept of ID encoding/decoding, we also need to ensure that any ID arguments are correctly decoded before passing them to the `createEvent` function.

This can be done by injecting the `ConfigurableOperationCodec` helper and calling `decodeConfigurableOperationIds()` , passing the handler and args.

The function mutates the arg.value for any IDS. You can see an example of usage here: https://github.com/vendure-ecommerce/vendure/blob/c53929057eb420c0581de262cbfe2c9e61cac726/packages/core/src/api/resolvers/admin/promotion.resolver.ts#L79

packages/email-plugin/src/handler/default-email-handlers.ts
```diff
@@ -30,6 +32,57 @@ export const orderConfirmationHandler = new EmailEventListener('order-confirmati
...
+        entityType: Order,
+        label: [
+            {
+                value: 'Order confirmation.',
+                languageCode: LanguageCode.en,
+            },
+        ],
+        description: [
+            {
+                value: 'Order confirmation can be send only for specific reasons.',
```
michaelbromley: "can be sent"

packages/email-plugin/src/handler/default-email-handlers.ts
```diff
@@ -30,6 +32,57 @@ export const orderConfirmationHandler = new EmailEventListener('order-confirmati
         event =>
             event.toState === 'PaymentSettled' && event.fromState !== 'Modifying' && !!event.order.customer,
     )
+    .setResendOptions({
+        entityType: Order,
+        label: [
+            {
+                value: 'Order confirmation.',
```
michaelbromley: drop the `.`

michaelbromley (changes requested):

packages/email-plugin/src/email-event.resolver.ts
monrostar: I've added a value parser in this place but there seems to be no ID decode there...

monrostar:
What I didn't finish:
1. Only a temporary UI for the Order page is made. Not added EmailEventUI to customer
2. Not added Permissions, maybe it is possible to make permissions only for EmailEventResend...?
3. I have done args parsing, but I wanted to discuss how to add ID decode since EmailPlugin has no properties in VendureConfig and maybe you can find a better solution
4. No Errors have been added to the `email-event.resolver.ts`

packages/email-plugin/src/handler/default-email-handlers.ts
monrostar: done

packages/email-plugin/src/handler/default-email-handlers.ts
monrostar: done

michaelbromley:
Status update from my side:

I pulled this PR locally and tested it out. Basically it looks good but I have a few points I want to spend more time thinking about when I am a little less busy:

- We discussed extracting the UI parts to the EmailPlugin to avoid the coupling. Architecturally this would be correct. DX-wise it would mean more complexity to expose this functionality in the UI.
- In that regard it would be nice to be able to expose this functionality out-of-the-box without requiring to compile the Admin UI.

But in order to bundle this functionality into the Admin UI I'd really only want to do that if the functionality is not specific to the EmailPlugin. 

So I'm wondering whether in here there is a more generic concept such as "event replay" which would be more broadly applicable, and then the email re-send is a specific implementation of that.

monrostar:
> Status update from my side:
> 
> I pulled this PR locally and tested it out. Basically it looks good but I have a few points I want to spend more time thinking about when I am a little less busy:
> 
> * We discussed extracting the UI parts to the EmailPlugin to avoid the coupling. Architecturally this would be correct. DX-wise it would mean more complexity to expose this functionality in the UI.
> * In that regard it would be nice to be able to expose this functionality out-of-the-box without requiring to compile the Admin UI.
> 
> But in order to bundle this functionality into the Admin UI I'd really only want to do that if the functionality is not specific to the EmailPlugin.
> 
> So I'm wondering whether in here there is a more generic concept such as "event replay" which would be more broadly applicable, and then the email re-send is a specific implementation of that.

Do you need help integrating such functionality? And it would be cool to know when we can get such a feature

michaelbromley:
Just an update as I am going through old PRs this morning: this is still on our "under consideration" lane in the roadmap. For now we have other higher priority features we want to ship but I'm leaving this open because it is fundamentally a good feature.

dlhck:
I like the concept of "event replay" that @michaelbromley introduced - we we'll look into that. As this PR introduces a larger new feature to the Admin UI, I will close this PR for now - we won't add any _new_ features to the Admin UI as it was discontinued in favor of the new dashboard.

PR #3021, opened 2024-08-20
feat(asset-server-plugin): optimize S3 file handling

asonnleitner:
- Added Buffer import from 'node:buffer'.
- Replaced manual buffer concatenation with `transformToByteArray` for efficiency in `toBuffer`.
- Improved `readFileToStream` by utilizing `Readable.from` for the byte array.

These changes enhance the performance and readability of S3 file handling operations.

### DIFF
--- a/packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
+++ b/packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
@@ -2,6 +2,7 @@ import { PutObjectRequest, S3ClientConfig } from '@aws-sdk/client-s3';
 import { AwsCredentialIdentity, AwsCredentialIdentityProvider } from '@aws-sdk/types';
 import { AssetStorageStrategy, Logger } from '@vendure/core';
 import { Request } from 'express';
+import { Buffer } from "node:buffer";
 import * as path from 'node:path';
 import { Readable } from 'node:stream';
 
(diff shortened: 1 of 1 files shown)


packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
```diff
@@ -230,14 +226,15 @@ export class S3AssetStorageStrategy implements AssetStorageStrategy {
             });
         }
 
-        return body;
+        const typedArray = await body.transformToByteArray();
+        return Readable.from(typedArray)
```
michaelbromley, line 230: In this part, as far as I understand you are "consuming" the stream into a buffer and then creating a new stream from that buffer. That seems redundant to me? Or am I missing something?

dfernandesbsolus (approved):

packages/asset-server-plugin/src/s3-asset-storage-strategy.ts
```diff
@@ -210,13 +211,8 @@ export class S3AssetStorageStrategy implements AssetStorageStrategy {
...
             return Buffer.from('');
         }
-
-        const chunks: Buffer[] = [];
-        for await (const chunk of body) {
-            chunks.push(chunk);
-        }
-
-        return Buffer.concat(chunks);
+        const typedArray = await body.transformToByteArray()
```
dfernandesbsolus, line 214: Nice!

dfernandesbsolus (approved):

michaelbromley (commented):
Thanks for the contribution! Check my comment and let me know what you think.

michaelbromley:
Hi, I'm going to close this for now because the implementation was unclear to me, as per my comment above. Open to a new PR which addresses this. Thanks for your time 👍

PR #3024, opened 2024-08-20
Text message plugin

asonnleitner:
# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Hi!

Would it make sense to make text messaging a capability of the email plugin? I know the naming will then be wrong, but purely in terms of the technical implementation, will there be a large overlap there between the two?

asonnleitner:
@michaelbromley Will look into it, also looks like the way to go 😄

dlhck:
@michaelbromley @asonnleitner if we put text messaging capabilities into the email plugin we should think about renaming it to notification plugin. Notifications are more generic, they could be transferred as emails, text messages or even SMS and WhatsApp. What do you think?

michaelbromley:
@dlhck yes it would definitely need renaming in that case. Which is fine long-term. We'd need to think about how to minimize the breaking change for existing users of the package.

dlhck:
@michaelbromley we could think about renaming the current email plugin to `LegacyEmailPlugin`, so that APIs stay the same. We cloud start  then with  the `NotificationsPlugin` without thinking about backwards-compatibility.

michaelbromley:
@dlhck yes that's a possibility. We could take that opportunity to fix a bunch of other design issues with the current plugin - e.g. a better API for extending handlers, better i18n support in templates, etc. if we go with a clean break in a brand new package.

monrostar:
That's a great idea. We've already had to write a few extensions to send messages to Telegram or do scheduled Emails and Sms

monrostar:
> @dlhck yes that's a possibility. We could take that opportunity to fix a bunch of other design issues with the current plugin - e.g. a better API for extending handlers, better i18n support in templates, etc. if we go with a clean break in a brand new package.

That's a great idea. We've already had to write a few extensions to send messages to Telegram or do scheduled Emails and Sms

dlhck:
Closing this because of inactivity.

PR #3033, opened 2024-08-27
(feat): integrate Vendure Assets Picker with ProseMirror and add single image selection

dfernandesbsolus:
# Description

- Added functionality to allow the insertion of images from Vendure's Assets Picker into the ProseMirror editor.

- Implemented a restriction in the assets picker to allow only single image selection to support the ProseMirror's image handling capabilities.

# Breaking changes

No

# Screenshots

https://github.com/user-attachments/assets/f3dbdbfe-42fc-4c83-a760-5624c2f3a679

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/rich-text-editor/prosemirror/plugins/image-plugin.ts
+++ b/packages/admin-ui/src/lib/core/src/shared/components/rich-text-editor/prosemirror/plugins/image-plugin.ts
@@ -19,6 +19,7 @@ export const imageNode: NodeSpec = {
         title: { default: null },
         width: { default: null },
         height: { default: null },
+        dataExternal: { default: true },
     },
     group: 'inline',
     draggable: true,
@@ -32,13 +33,14 @@ export const imageNode: NodeSpec = {
                     alt: (dom as HTMLImageElement).getAttribute('alt'),
                     width: (dom as HTMLImageElement).getAttribute('width'),
                     height: (dom as HTMLImageElement).getAttribute('height'),
+                    dataExternal: (dom as HTMLImageElement).hasAttribute('data-external'),
                 };
             },
         },
     ],
     toDOM(node) {
-        const { src, alt, title, width, height } = node.attrs;
-        return ['img', { src, alt, title, width, height }];
+        const { src, alt, title, width, height, dataExternal } = node.attrs;
+        return ['img', { src, alt, title, width, height, 'data-external': dataExternal }];
     },
 };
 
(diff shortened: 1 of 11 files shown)


dfernandesbsolus:
@michaelbromley I put this pull request as a draft so you can analyze whether this was the best approach.

packages/admin-ui/src/lib/core/src/shared/components/rich-text-editor/prosemirror/plugins/image-plugin.ts
```diff
@@ -49,6 +67,7 @@ export function insertImageItem(nodeType: NodeType, modalService: ModalService)
                 })
                 .subscribe(result => {
                     if (result) {
+                        console.log(result);
```
michaelbromley: Needs removing :)

michaelbromley (commented):
Excellent contribution, thank you very much. Just a stray console.log that sneaked in there that needs removing. Other than that, perfect 👌

dfernandesbsolus:
> Excellent contribution, thank you very much. Just a stray console.log that sneaked in there that needs removing. Other than that, perfect 👌

Thanks, i was just thinking, when it's a URL that isn't from the domain, it doesn't make sense to show the preset select.

michaelbromley:
Yes, that's a very good point. Can you think of a way to implement that?

dfernandesbsolus:
> Yes, that's a very good point. Can you think of a way to implement that?

Can I validate if it is the url of the Vendure assets, if not I won't show it, can I access the url of the assets? Normally it is mysite.com/assets, correct?

michaelbromley:
Yes, it is `assets/` by default but that is configurable in the AssetServerPlugin `route` option. So not something we can safely assume. What if we set metadata on the inserted image tag to mark it as internal/external?

dfernandesbsolus:
It works, but I will have to put this information in the html to later retrieve something like:
 ``

michaelbromley:
In that case can we use a data attribute, since "internal" is not a valid attribute of `img`? 
Also maybe inverting it and marking "external" for outside images is better than "internal"?

So it would be ``

dfernandesbsolus:
I agree, I will make this change :D

dfernandesbsolus:
@michaelbromley What do you think? I blocked editing of the source when the image is internal to maintain the data-external logic, but I added the option to remove the image which enables the field again.


Uploading gravacao-de-tela-2024-09-06-as-135842_Z5BN3TC2.mp4…

michaelbromley:
That's very good, thank you!

Issue #3012, opened 2024-08-14
Rebuilding search index error when working with multiple channels


hsensh:
**Describe the bug**
I am working on a complex system with multiple channels and automatic product synchronization with an ERP, when I try to rebuild the search index, it gives me 

```
[worker] error 8/14/24, 2:38 PM - [Vendure Worker] Product not found for variant!
[worker] Error: Product not found for variant!
```

**To Reproduce**
It happens randomly but I cannot pinpoint why because of the complexity of my plugins, the database seems fine in terms of consistency. Product variant exists in default channel and channel 1
For some reason it's querying its product on channel 2

**Expected behavior**
It should not query products of variants of different channels.

**Environment (please complete the following information):**
 - @vendure/core version: 3.0.0
 - Nodejs version: 18
 - Database (mysql/postgres etc): mysql

**Additonal context**
I investigated the file responsible for this which is core/dist/plugin/default-search-plugin/indexer/indexer.controller.js specifically in the reindex function more specifically the updateVariantsById function:

in the index function:
```
const { variants } = await this.getSearchIndexQueryBuilder(ctx, {
    channels: channel],
    take: exports.BATCH_SIZE,
    skip: i * exports.BATCH_SIZE,
});
await this.saveVariants(ctx, variants);
```
for some reason the `channel` variable is not the same as the `ctx.channel` variable.

This is what's being called in the updateVariantsById:

```
product = await this.getProductInChannelQueryBuilder(ctx, variant.productId, ctx.channel);
```
Channel's id is different in both functions, resulting in the error, the product variant indeed has a product in the same channel so it shouldn't result in the error.

I believe the issue is here:
```
const channel = (_a = ctx.channel) !== null && _a !== void 0 ? _a : (await this.loadChannel(ctx, ctx.channelId));
```

And the weird thing is it's not even giving the default channel, just a different one.

hsensh:
After further investigation, I believe the issue is here:

```
for (const channel of variant.channels) {
                    ctx.setChannel(channel);
....
```
in the `saveVariants` method. It is swapping the channels and not swapping it back.

I added this line in the main loop:

```
async saveVariants(ctx, variants) {
        var _a, _b, _c, _d, _e;
        const originalChannel = ctx.channel;
...
for (const variant of variants) {
            ctx.setChannel(originalChannel);
```
And it didn't result in an error anymore

williamrijksen:
I experience the same issue. Can we create a fix for this?

[...]

PR #3041, opened 2024-09-02
fix(default-search-plugin): Fix channels swapping and not swapping back in saveVariants when working with multiple channels

hsensh:
Relates to #3012.

# Description

While indexing variants, a product for a specific variant is queried. However, somewhere in the code the channels are being swapped if multiple channels exist, and they are not being swapped back, which is resulting in the product not being found for that channel as they don't exist, which is making the indexing fail. that's why I have added a line in the code that swaps the channels back to the original one at the start of every loop, and a line at the end of the loop.


hsensh:

michaelbromley:
Thanks for the quick fix! 🙏

PR #3080, opened 2024-09-24
Check type of durations before passing to 'ms' function

### LINKED ISSUE #3078: AuthOptions.sessionDuration does not support actual milliseconds passed as number
The [docs](https://docs.vendure.io/reference/typescript-api/auth/auth-options#sessionduration) say that sessionDuration can be passed as a number or string, using the [ms](https://www.npmjs.com/package/ms) package to convert the string expression to milliseconds... when you actually pass milliseconds though, the `ms` function will convert from milliseconds to string expression which breaks session creation.

I would propose to only convert strings using the `ms` function or change the docs to explicitely say that milliseconds should not be passed as number but as string (as the `ms` package documentation says, strings that just contain a number and not an expression will just be converted to a number) 

https://github.com/vendure-ecommerce/vendure/blob/a735bdf9eaadb83251eec1a135b084569d10e333/packages/core/src/service/services/session.service.ts#L40

pujux:
# Description

Check if the type of `authOptions.sessionDuration` and `authOptions.verificationTokenDuration` is `string` before passing the values to the `ms` function for conversion from a string expression to milliseconds (number). If the type is not a string, it is already expected to be milliseconds so can be used right away.

Fixes #3078 

# Breaking changes

No breaking changes.

# Screenshots

Not applicable.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/verification-token-generator/verification-token-generator.ts
+++ b/packages/core/src/service/helpers/verification-token-generator/verification-token-generator.ts
@@ -28,15 +28,16 @@ export class VerificationTokenGenerator {
      * as specified in the VendureConfig.
      */
     verifyVerificationToken(token: string): boolean {
-        const duration =
-            typeof this.configService.authOptions.verificationTokenDuration === 'string'
-                ? ms(this.configService.authOptions.verificationTokenDuration)
-                : this.configService.authOptions.verificationTokenDuration;
+        const { verificationTokenDuration } = this.configService.authOptions;
+        const verificationTokenDurationInMs =
+            typeof verificationTokenDuration === 'string'
+                ? ms(verificationTokenDuration)
+                : verificationTokenDuration;
 
         const [generatedOn] = token.split('_');
         const dateString = Buffer.from(generatedOn, 'base64').toString();
         const date = new Date(dateString);
         const elapsed = +new Date() - +date;
-        return elapsed < duration;
+        return elapsed < verificationTokenDurationInMs;
     }
 }
--- a/packages/core/src/service/helpers/verification-token-generator/verification-token-generator.ts
+++ b/packages/core/src/service/helpers/verification-token-generator/verification-token-generator.ts
@@ -28,7 +28,11 @@ export class VerificationTokenGenerator {
      * as specified in the VendureConfig.
      */
     verifyVerificationToken(token: string): boolean {
-        const duration = ms(this.configService.authOptions.verificationTokenDuration as string);
+        const duration =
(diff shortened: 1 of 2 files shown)


pujux:

packages/core/src/service/helpers/verification-token-generator/verification-token-generator.ts
```diff
@@ -28,7 +28,11 @@ export class VerificationTokenGenerator {
      * as specified in the VendureConfig.
      */
     verifyVerificationToken(token: string): boolean {
-        const duration = ms(this.configService.authOptions.verificationTokenDuration as string);
+        const duration =
```
michaelbromley: just to ease the readability slightly, could we do:

```ts
const { verificationTokenDuration } = this.configService.authOptions;
```

to avoid repeating the full object path 3 times? Same with the other one too.

pujux: yeah definitely! I thought about doing so but wanted to keep it in one line so didn't do it then. Will update shortly.

pujux: what do you think about this?
```ts
verifyVerificationToken(token: string): boolean {
    let { verificationTokenDuration } = this.configService.authOptions;
    if (typeof verificationTokenDuration === 'string') {
        verificationTokenDuration = ms(verificationTokenDuration);
    }

    const [generatedOn] = token.split('_');
    const dateString = Buffer.from(generatedOn, 'base64').toString();
    const date = new Date(dateString);
    const elapsed = +new Date() - +date;
    return elapsed < verificationTokenDuration;
}
```

for sessionDuration I would do 

```ts
const { sessionDuration } = this.configService.authOptions;
this.sessionDurationInMs = 
    typeof sessionDuration === 'string' ? ms(sessionDuration) : sessionDuration;
```

michaelbromley: Rather than re-assign the verificationTokenDuration variable, how about you create a new variable named verificationTokenDurationInMs, and assign it with a ternary like in the sessionDuration example? That would be even more explicit.

michaelbromley:
Thank you! This looks perfect now. I just need to investigate the failing e2e tests before merging. I'm pretty sure it has nothing to do with your changes though.

pujux:
Alright! Do you want me to add the change to sessionCacheTTL discussed in #3078 in here or in another PR?

michaelbromley:
Yes, do that in a separate PR please 👍  It should also be done against the `minor` branch as it is technically a new feature rather than a fix.

PR #3088, opened 2024-09-25
Enhance AuthOptions.sessionCacheTTL by allowing string expressions (zeit/ms)

pujux:
# Description

- Support string expressions (as per `zeit/ms`) for `AuthOptions.sessionCacheTTL` just like for `sessionDuration` and `verificationTokenDuration` (conversion result needs to be divided by 1000 to receive seconds)

- Update documentation and jsDoc to better explain how these properties can be supplied as numbers (seconds/milliseconds) or string expressions (as per `zeit/ms`)

# Breaking changes

No

# Screenshots

Not applicable.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/session.service.ts
+++ b/packages/core/src/service/services/session.service.ts
@@ -146,8 +146,11 @@ export class SessionService implements EntitySubscriberInterface {
      * Serializes a {@link Session} instance into a simplified plain object suitable for caching.
      */
     serializeSession(session: AuthenticatedSession | AnonymousSession): CachedSession {
-        const expiry =
-            Math.floor(new Date().getTime() / 1000) + this.configService.authOptions.sessionCacheTTL;
+        const { sessionCacheTTL } = this.configService.authOptions;
+        const sessionCacheTTLSeconds =
+            typeof sessionCacheTTL === 'string' ? ms(sessionCacheTTL) / 1000 : sessionCacheTTL;
+
+        const expiry = Math.floor(new Date().getTime() / 1000) + sessionCacheTTLSeconds;
         const serializedSession: CachedSession = {
             cacheExpiry: expiry,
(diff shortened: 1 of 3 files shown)


packages/core/src/service/services/session.service.ts
```diff
@@ -146,8 +146,11 @@ export class SessionService implements EntitySubscriberInterface {
      * Serializes a {@link Session} instance into a simplified plain object suitable for caching.
      */
     serializeSession(session: AuthenticatedSession | AnonymousSession): CachedSession {
-        const expiry =
-            Math.floor(new Date().getTime() / 1000) + this.configService.authOptions.sessionCacheTTL;
+        const { sessionCacheTTL } = this.configService.authOptions;
+        const sessionCacheTTLSeconds =
+            typeof sessionCacheTTL === 'string' ? ms(sessionCacheTTL) / 1000 : sessionCacheTTL;
```
pujux, line 151: @michaelbromley should we `Math.floor` the whole expression assigned to `sessionCacheTTLSeconds` just like it's done with `new Date().getTime() / 1000`?

if so we should have some minimum for `sessionCacheTTLSeconds` like 1 to not run into the case of `sessionCacheTTLSeconds` being 0

if not, what is the reason to use `Math.floor` on `new Date().getTime() / 1000` anyway?

michaelbromley: Honestly I'm not sure if there is any good reason to do the Math.floor at all. I mean, it keeps the `cacheExpiry` value to a round number of seconds, but I don't think that really makes any functional difference. So let's just get rid of it.

dlhck (approved):

dlhck:
Thanks & welcome to the club of our contributors @pujux ! :)

Issue #3093, opened 2024-09-28
Custom Field number string value breaks if it's on select input


jyling:
**Describe the bug**
I'm trying to implement a simple input to my channel where I can set a list of supported calling codes, such as `1`, `60` and `61`. but I encounter an issue where the input from the admin UI will send a number type instead of a string to the backend on one of the values.

I tried searching for an issue with [this](https://github.com/vendure-ecommerce/vendure/issues?q=is%3Aissue+is%3Aopen+customfield+number), but I could not find anything related to this issue.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add this to your channel's custom field
```
config.customFields.Channel.push({
    name: "supportedNumbers"
    type: "string",
    label: [{ languageCode: LanguageCode.en, value: "Supported Phone Number" }],
    ui: { component: "select-form-input" },
    list: true,
    options: [

        // Usually we get this list from a library, but we 
        // can hard code it for now
        {
            value: "1",
            label: [{ languageCode: LanguageCode.en, value: "US" }]
        },
        {
            value: "60",
            label: [{ languageCode: LanguageCode.en, value: "MY" }]
        },
        {
            value: "61",
            label: [{ languageCode: LanguageCode.en, value: "SG" }]
        }
    ]
})
```
2. generate your migration
3. open the admin UI and navigate to any channel config page
4. you can see the custom field input looks weird
![image](https://github.com/user-attachments/assets/5225c5c7-f80b-41ae-81a0-bf3a753af6f0)
5. try add one input
6. update, there's no issue so far, only ui bug
![image](https://github.com/user-attachments/assets/d77c1fbb-8dee-45f5-8b79-a3a040b8a59c)
8. Now add another one, ui is still bugged
 ![image](https://github.com/user-attachments/assets/b35b8bba-dc60-40fb-9f09-711653f6d357)
9. press update after making the changes
10. See error (note: for testing purposes, I call my customfield as `e`)
![image](https://github.com/user-attachments/assets/249f6fef-73ca-47ff-a342-e6523fe538ed)
![image](https://github.com/user-attachments/assets/bfc37e6c-2569-4302-be26-55fb4d8f638e)

**Expected behavior**
It should allow the US, SG or MY, and should not make one of the inputs into a number, it should be a string. 

**Environment (please complete the following information):**
 - @vendure/core version: 3.0.3
 - Nodejs version: 18.2.3
 - Database (mysql/postgres etc): postgres 16

**Additional context**
[...]

PR #3094, opened 2024-09-28
fix: solve incorrect type when dealing with value in list

jyling:
# Description

Please include a summary of the changes and the related issues.
This fix solves the [issue 3093](https://github.com/vendure-ecommerce/vendure/issues/3093) where the output coming from the dynamic form input has the wrong type. This lead to weird behaviour when selecting the options and also issue with updating the record on the admin UI.

My fix consists of overriding the value with the correct type, I'm unsure if this is a acceptable solution, but it does solve the issue.

For example. 
I have a list of options from a custom field that contains numbers like calling code.
The options are meant to be strings of options. so it's configured like this
```typescript
{
    name: "supportedNumbers"
    type: "string",
    label: [{ languageCode: LanguageCode.en, value: "Supported Phone Number" }],
    ui: { component: "select-form-input" },
    list: true,
    options: [

        // Usually we get this list from a library, but we 
        // can hard code it for now
        {
            value: "1",
            label: [{ languageCode: LanguageCode.en, value: "US" }]
        },
        {
            value: "60",
            label: [{ languageCode: LanguageCode.en, value: "MY" }]
        },
        {
            value: "61",
            label: [{ languageCode: LanguageCode.en, value: "SG" }]
        }
    ]
}
```
When I select the first option, the output will become like this
```
["1"]
```

The issue arises when my form input is a list, and I want to add the second option, which will cause the first option to turn into a number, so it will become like this
```
[1, "60"].
```

if I change the first input again, it will become like this
```
["61", 60]
```

# Breaking changes

I'm not very sure, is there any test that I can run ?

# Screenshots
Before
image
After
image

# Checklist

📌 Always:
- [] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/dynamic-form-inputs/dynamic-form-input/dynamic-form-input.component.ts
+++ b/packages/admin-ui/src/lib/core/src/shared/dynamic-form-inputs/dynamic-form-input/dynamic-form-input.component.ts
@@ -187,6 +187,12 @@ export class DynamicFormInputComponent
         if (this.listItems) {
             for (const item of this.listItems) {
                 if (item.componentRef) {
+                    const { value } = item.control;
+                    const { type } = item.componentRef.instance.config || {};
+                    // fix a bug where the list item of string turns into number which lead to unexpected behavior
+                    if (typeof value === 'number' && type === 'string') {
+                        item.control.setValue(item.control.value?.toString(), { emitEvent: false });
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thank you for the detailed write-up and the fix! And also for your patience because it took me a while to get around to this :D

Issue #1726, opened 2022-08-19
Error when running queues in separate workers - Bullmq strategy


ttournie:
**Describe the bug**
When running 2 workers with separate queues I get errors from the server and the worker that is not supposed to run the queue still tries to do it.

**To Reproduce**
1. Create 2 worker, une dedicated to indexing, the other to apply-collection-filter
2. Run the project
3. Add/modify a product to trigger some jobs
4. See error in console

**Expected behavior**
No error should be displayed and jobs thaht does not concerne a worker should be ignored

**Environment (please complete the following information):**
 - @vendure/core version: 1.6.4
 - Nodejs version: 16.16
 - Database (mysql/postgres etc): postgres

michaelbromley:
Note, error seems to come from this part: https://github.com/vendure-ecommerce/vendure/blob/3835f8b22cb5b61f22c372a4f251907bc8f07a09/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts#L86-L97

in the case of worker instances configured to only handle a subset of jobs, the situation of having no defined `processFn` is actually _expected_ and not an exception situation.

michaelbromley:
Just spent a bit more time looking into this. Fundamentally the issue is that all Vendure job queues are stored in a single BullMQ `Queue` instance, which means there is 1 Redis list of job Ids which all workers will take jobs from.

BullMQ does not have any kind of "filtering" feature which allows a job to be conditionally picked up by a worker based on some function.

The closest thing I can find is that in the `workerOptions` there is a `stalledInterval` setting which controls how often checks are made for [stalled jobs](https://docs.bullmq.io/guide/workers/stalled-jobs). When a job is stalled, it will be returned to the "waiting" queue and can then be picked up by another worker.

Using this "stalled" mechanism however - if it could be made to work - would be a hack so it's not something I want to do in the plugin itself.

Perhaps we can create a subclass of the `Worker` class which overrides the [getNextJob method](https://api.docs.bullmq.io/classes/v4.Worker.html#getNextJob) and implements a filter for the job queue name.

PR #3108, opened 2024-10-05
refactor(job-queue-plugin): Use multiple BullMQ queues instead of one

### LINKED ISSUE #2419: `activeQueues` option not working as intended with BullMQ job queue strategy
**Description**
When using the `activeQueues` option in `JobQueueOptions`, the observed behavior with the BullMQ job queue strategy is inconsistent with expectations. All worker processes pull jobs from a single queue, even when different `activeQueues` are specified. Consequently, jobs not listed in the `activeQueues` option are marked as failed by workers.

<details>
<summary> `JobQueueOptions` </summary>

https://github.com/vendure-ecommerce/vendure/blob/7549aad738a0d8644de5a6df896b04fe86cbc31e/packages/core/src/config/vendure-config.ts#L872-L902

</details>

In the BullMQ strategy, only one queue is created:

https://github.com/vendure-ecommerce/vendure/blob/7549aad738a0d8644de5a6df896b04fe86cbc31e/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts#L64-L74

**Proposed Behavior**
I believe a more robust solution would be to create a separate queue for each entry in the `activeQueues` option. Each worker process would then pull jobs only from its corresponding queue. While it might be possible to use a single queue and have each worker check the `activeQueues` before processing, this approach could introduce delays in job processing.

When running multiple worker processes with defined `activeQueues`, jobs are pulled from the single queue. This causes jobs to fail if they are not defined in `activeQueues`.

**To Reproduce**
1. Set up Vendure with BullMQ and Redis.
2. Run multiple worker processes with different `activeQueues`.
3. Observe that all workers pull jobs from the same queue, causing jobs not defined in `activeQueues` to fail rather than be skipped.

**Expected behavior**
Each worker process should only process jobs from the queues listed in their `activeQueues` option. Ideally, separate queues should be created for each `activeQueues` entry to avoid unnecessary delays in job processing.

dlhck:
# Description

Refactors the `bullmq-job-queue-strategy` to use a separate BullMQ queue for every Vendure job queue. This should improve the stability and issues with horizontally scaled worker instances.   

Resolves #2419 and #1726

# Breaking changes

There is no breaking change.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/job-queue/job-queue.service.ts
+++ b/packages/core/src/job-queue/job-queue.service.ts
@@ -54,7 +54,10 @@ export class JobQueueService implements OnModuleDestroy {
         return this.configService.jobQueueOptions.jobQueueStrategy;
     }
 
-    constructor(private configService: ConfigService, private jobBufferService: JobBufferService) {}
+    constructor(
+        private configService: ConfigService,
+        private jobBufferService: JobBufferService,
+    ) {}
 
     /** @internal */
     onModuleDestroy() {
@@ -152,6 +155,13 @@ export class JobQueueService implements OnModuleDestroy {
      */
     flush(...forBuffers: Array<JobBuffer<any> | string>): Promise<Job[]> {
         return this.jobBufferService.flush(forBuffers);
+    }
+
+    /**
+     * @description Returns the raw objects representing the JobQueues.
+     */
+    getRawJobQueues() {
+        return this.queues;
     }
 
     /**
--- a/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
+++ b/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
@@ -7,6 +7,8 @@ import {
     InternalServerError,
     Job,
     JobData,
+    JobQueue,
+    JobQueueService,
     Logger,
     PaginatedList,
 } from '@vendure/core';
@@ -42,8 +44,8 @@ const DEFAULT_CONCURRENCY = 3;
 export class BullMQJobQueueStrategy implements InspectableJobQueueStrategy {
     private redisConnection: Redis | Cluster;
     private connectionOptions: ConnectionOptions;
-    private queue: Queue;
-    private worker: Worker;
+    private queues: Map<string, Queue> = new Map();
+    private workers: Map<string, Worker> = new Map();
     private workerProcessor: Processor;
     private options: BullMQPluginOptions;
     private queueNameProcessFnMap = new Map<string, (job: Job) => Promise<any>>();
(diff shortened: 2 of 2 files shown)


dlhck:
There is currently one major problem in the core implementation: Every job needs to have a unique ID. But with BullMQ the IDs are only unique within the sequence of a queue. This is the core issue that still needs to be resolved.

dlhck:
@michaelbromley we are running into an issue, that it is not possible to display a list of all jobs from all queues in all states (aka the default list view). This is because of limitations of the BullMQ library and the way how the Redis keys are structured. 

I see two options here: 
- Not support the default list view at all 
- Change the list view implementation to make it possible to disable the initial state (no filters) and automatically select the first queue (ergo only make it possible to view jobs filtered by queue)

michaelbromley:
I got sucked down a deep rabbithole on this today. I think I'm pretty close to done, but this is much more subtly complex than I first thought.

I established some e2e tests, but things are still buggy in manual testing and I don't have time to finish this now for v3.1.

alexisvigoureux:
Hello, we are really interested in this feature! It seems that the last sticking point is the display of queues. Why not display them via the [Bull board](https://github.com/felixmosh/bull-board) integration? We have done it for our vendure app via a plugin and it works well. I can help if this is a solution you want to consider
bullboard

dlhck:
@alexisvigoureux thanks for the input. The displaying the of queues is not really a problem anymore. We are looking into shipping this also with v3.2

michaelbromley:
I have realized that there is a simpler solution to a major practical problem solved by this PR: lack of priority mechanism.

Currently, (and this is a real world example) let's say we suddenly have 200k jobs created for a bulk price update. This might take 30-60 mins to process. In that time, any order placed will have their emails delayed until after that, due to the strict first-in-first-out (FIFO) method of handling jobs.

However, Bullmq supports priority, so if we had a mechanism for specifying a higher priority for the email jobs, they will get processed right away, without having to wait for all the "price update" jobs to complete.

This would be relatively simple to implement, basically using the exact same mechanism as the existing options `setRetries` and `setBackoff` (https://docs.vendure.io/reference/core-plugins/job-queue-plugin/bull-mqplugin-options#setretries)

There are _still_advantages to the rework proposed in this PR, but I will implement the priority support first.

yazfield:
nice to see this feature thanks 👍 

@alexisvigoureux I was planning to build a similar plugin, any chance you would open source it ?

dlhck:
Closing this in favor of a better, cleaner solution. A great improvement to prioritize jobs over others is already available.

PR #3119, opened 2024-10-08
fix(core): Deprecation of `lines`, `shipping` and `adjustment` fields in the `RefundOrderInput` input.

sonntag-philipp:
# Description

During the refund implementation in my PayPal integration, I noticed that the `RefundOrderInput` can be a little bit confusing for new developers. At first I did not understand how I am able to use the amount instead of the lines, adjustment and shipping properties.

There was no documentation for this specific input, so I reached out to the discord. In my [help thread](https://discord.com/channels/1100672177260478564/1293232023921491999) we came to false conclusions, as we thought it is not possible to just use the `amount` field by itself without `lines` etc.

After some further investigation, I noticed that deprecating the `lines`, `shipping` and `adjustment` fields in the GraphQL schema have no effects on the e2e tests. On this base, I removed those 3 fields and created this PR to make the input less confusing.

I understand that my changes can be a little bit controversial, especially as now no value is required in this input. Sadly I am not experienced enough with GraphQL to say if this is okay if checks still ensure at least one option is set. I understand that the fields may better be changed back.

# Breaking changes

- Deprecation of `lines`, `shipping` and `adjustment` fields in the `RefundOrderInput` input.
- Making the same fields optional and not required anymore.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/schema/admin-api/order.api.graphql
+++ b/packages/core/src/api/schema/admin-api/order.api.graphql
@@ -111,14 +111,12 @@ input CancelOrderInput {
 }
 
 input RefundOrderInput {
-    lines: [OrderLineInput!]!
-    shipping: Money!
-    adjustment: Money!
+    lines: [OrderLineInput!] @deprecated(reason: "Use the `amount` field instead")
+    shipping: Money @deprecated(reason: "Use the `amount` field instead")
+    adjustment: Money @deprecated(reason: "Use the `amount` field instead")
     """
-    If an amount is specified, this value will be used to create a Refund rather than calculating the
-    amount automatically. This was added in v2.2 and will be the preferred way to specify the refund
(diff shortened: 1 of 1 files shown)


packages/core/src/api/schema/admin-api/order.api.graphql
```diff
@@ -410,28 +408,28 @@ type ManualPaymentStateError implements ErrorResult {
 
 union TransitionOrderToStateResult = Order | OrderStateTransitionError
 union SettlePaymentResult =
```
sonntag-philipp, line 410: Automatic formatting. Will revert this if necessary.

michaelbromley (approved):
You're correct - these are indeed deprecated. It's good to have that made explicit in the schema 👍

Issue #3117, opened 2024-10-08
Streamline getting started experience


michaelbromley:
Our `@vendure/create` package is used to set up new Vendure projects. It is also used by developers looking to quickly try out Vendure for themselves.

It uses a series of prompts to collect the required information to scaffold a project directory with the correct configuration.

## Current flow

![image](https://github.com/user-attachments/assets/a31b84cf-14b4-4010-95b1-3e9f6fe04d60)

## Issues

- Sql.js is just not needed and can be removed to simplify the options & the code
- When using mysql/postgres, there are many steps to go through that add complexity to getting started:
  - You need to already have a DB server running somewhere
  - You need to manually create the database
  - You need to know the DB user credentials
- There is long pause after "Generated app scaffold" with no spinner. Some sync blocking work must be happening here.
- There are some reports of failures during install that seem related to native modules, especially Sharp. These are difficult to reproduce.

## Improved flow

We should aim to cut down the number of steps to an absolute minimum. Therefore we should have 2 flows:

- **Quick start**: select which DB to support, and that's it. Everything else is automated. We use Docker to create a DB server instance for you locally.
- **Manual config**: Like the current flow, with the possibility of adding other configurable options, eg ID strategy (it is hard to change later to UUIDs)

So the quick start would consist of just these steps:

1. `npx @vendure/create my-shop`
2. Select DB type
3. ...??
4. profit

## Docker detection

In the "quick start" flow:

We need to detect whether docker is running locally. Something like executing `docker -v` and testing the result / catching the error would probably suffice.

If docker is not detected, we can possibly try to start it automatically (research this), or at least prompt the user to start it and try again. If docker is simply not available, we could either 1) default to SQLite or 2) put user into the custom flow.

dlhck:
Let's remove the DB step also in that quick start - is it really necessary to choose a DB type to experience Vendure? My answer would be _no_. We generally tend to recommend to use Postgres, so that quick start should also use that. Everything else should be configurable via interactive prompts or non-interactive options (e.g. `--db=postgres`)

It should also automatically start server and worker after scaffolding the new Vendure installation.

michaelbromley:
Flow chart of the new process:

```mermaid
flowchart TD
    A[Start] --> B{Prompt: Quick Start or Manual Configuration}
    
    B -->|Quick Start| C{automated check: Is Docker running?}
    C -->|Yes| D[Continue with installation]
    C -->|No| E[Attempt to start Docker]
[...]

PR #3128, opened 2024-10-10
Improved getting started

michaelbromley:
Relates to #3117 

- Needs testing on macos & linux


dlhck (approved):
Ship it 🚀

packages/create/src/helpers.ts
```diff
@@ -383,6 +335,128 @@ async function checkPostgresDbExists(options: any, root: string): Promise<true>
...
+    const dockerSpinner = spinner();
+
+    function isDaemonRunning(): boolean {
+        try {
+            execFileSync('docker', ['stats', '--no-stream'], { stdio: 'ignore' });
+            return true;
+        } catch (e: any) {
+            return false;
+        }
+    }
```
DanielBiegler, line 356: Some setups will require root privileges for the `docker stats` command so this might fail even if technically the daemon is running.

packages/create/src/helpers.ts
```diff
@@ -383,6 +335,128 @@ async function checkPostgresDbExists(options: any, root: string): Promise<true>
...
+    const currentPlatform = platform();
+    try {
+        if (currentPlatform === 'win32') {
+            // https://stackoverflow.com/a/44182489/772859
+            execSync('"C:\\Program Files\\Docker\\Docker\\Docker Desktop.exe"');
+        } else if (currentPlatform === 'darwin') {
+            execSync('open -a Docker');
+        } else {
+            execSync('systemctl start docker');
+        }
```
DanielBiegler: Might be worth to `enable` instead of `start` to make docker auto-start on next reboot. Sidenote: Both starting and enabling require root privileges which might trip up some OS'

packages/create/src/helpers.ts
dlhck: That's true, but we won't be able to cover 100% of all setups.

packages/create/src/helpers.ts
dlhck: This command is running only on Linux based OS. In most cases the primary user has administrator rights. So again, covering 100% of setups (especially Linux-based) is virtually impossible. I think we can iterate here based on user feedback

PR #3131, opened 2024-10-13
Fix broken product image URLs in order confirmation email

### LINKED ISSUE #3130: Product image urls is broken in order confirmation email
I think this change break image links in order confirmation email - https://github.com/vendure-ecommerce/vendure/commit/e3d6c21ea37a5df8513b88eea0ac472054a35247#diff-ddc432c4392054b0d16b826194ca9e408576cee61292c9176f678919af6deb6fL1667

Standard order confirmation template expects order.lines.featuredAsset, but it is not loaded as relations in changed code
featuredAsset

const order = await this.findOne(
            ctx,
            orderId,
            relations ?? [
                'lines',
                'lines.productVariant',
                'lines.productVariant.productVariantPrices',
                'shippingLines',
                'surcharges',
                'customer',
            ],
        );

nirzaf:
Fixes #3130

Update the order service and resolver to include the `lines.featuredAsset` relation.

* **Order Service**
  - Add `lines.featuredAsset` to the relations array in the `findOne` method in `packages/core/src/service/services/order.service.ts`.

* **Order Resolver**
  - Add `lines.featuredAsset` to the relations array in the `order` query in `packages/core/src/api/resolvers/admin/order.resolver.ts`.

### DIFF
--- a/packages/core/src/api/resolvers/admin/order.resolver.ts
+++ b/packages/core/src/api/resolvers/admin/order.resolver.ts
@@ -69,7 +69,15 @@ export class OrderResolver {
         @Relations({ entity: Order, omit: ['aggregateOrder', 'sellerOrders'] })
         relations: RelationPaths<Order>,
     ): Promise<Order | undefined> {
-        return this.orderService.findOne(ctx, args.id, relations);
+        return this.orderService.findOne(ctx, args.id, relations ?? [
+            'lines',
+            'lines.productVariant',
+            'lines.productVariant.productVariantPrices',
+            'lines.featuredAsset', // Add this line
+            'shippingLines',
+            'surcharges',
+            'customer',
+        ]);
     }
 
     @Transaction()
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1678,6 +1678,7 @@ export class OrderService {
                 'lines',
                 'lines.productVariant',
                 'lines.productVariant.productVariantPrices',
+                'lines.featuredAsset',
                 'shippingLines',
                 'surcharges',
                 'customer',


HerclasNido:
Hi,
In the orderConfirmationHandler, when calling the function transformOrderLineAssetUrls(), toAbsoluteUrl() is used but seems to return the absolute url with an extra '/'.

Below, when I log the result and generate a test email I get: "http://localhost:3000/assets//mailbox/placeholder-image"
Maybe the bug comes from assetStorageStrategy.toAbsoluteUrl? I'm not sure how to approach this problem.

```
export function transformOrderLineAssetUrls(
    ctx: RequestContext,
    order: Order,
    injector: Injector
): Order {
    const { assetStorageStrategy } = injector.get(ConfigService).assetOptions;
    if (assetStorageStrategy.toAbsoluteUrl) {
        const toAbsoluteUrl =
            assetStorageStrategy.toAbsoluteUrl.bind(assetStorageStrategy);
        for (const line of order.lines) {
            if (line.featuredAsset) {
                const absolute = toAbsoluteUrl(ctx.req as Request, line.featuredAsset.preview);
                console.log(absolute);
                line.featuredAsset.preview = absolute;
                // line.featuredAsset.preview = toAbsoluteUrl(
                //     ctx.req as Request,
                //     line.featuredAsset.preview
                // );
            }
        }
    }
    return order;
}
```

jsfallon:
Just wanted to check in on the status of this issue. The broken product image URLs in the order confirmation email are a major issue, especially for new developers evaluating or onboarding with Vendure. It's one of the first things people notice during testing, and it can give the impression that key features are not fully functional.

Is there any timeline or plan for when this might be addressed? This is a critical feature for production use, so any update would be greatly appreciated.

packages/core/src/api/resolvers/admin/order.resolver.ts
```diff
@@ -69,7 +69,15 @@ export class OrderResolver {
...
-        return this.orderService.findOne(ctx, args.id, relations);
+        return this.orderService.findOne(ctx, args.id, relations ?? [
+            'lines',
+            'lines.productVariant',
+            'lines.productVariant.productVariantPrices',
+            'lines.featuredAsset', // Add this line
+            'shippingLines',
+            'surcharges',
+            'customer',
+        ]);
```
dlhck, line 80: Why do we need to join all the other relations too? Why not only `lines` and `lines.featuredAsset`

PR #3145, opened 2024-10-18
NativeAuth Changes

d46:
# Description

Follow up PR for the #2786 

# Breaking changes

`registerCustomerAccount` now returns `EmailAddressConflictError` if the customer has no user and no nativeAuth method. The method logic has been simplified with early termination. I removed the token refreshing logic from the first PR, as I believe it should not be the concern of the registration method.

For enumeration attacks, several types of enumeration attacks are possible. Addressing them at the application-wide or software level may not be the best approach. Handling them at the load balancer level could be a more effective solution.

# Checklist

📌 Always:

👍 Most of the time:


d46:

d46:
recheck

PR #3146, opened 2024-10-18
fix(chore): Return account conflict error for NativeAuth registration…

d46:
# Description

Follow up PR for the #2786 

# Breaking changes

`registerCustomerAccount` now returns `EmailAddressConflictError` if the customer has no user and no nativeAuth method. The method logic has been simplified with early termination. I removed the token refreshing logic from the first PR, as I believe it should not be the concern of the registration method.

For enumeration attacks, several types of enumeration attacks are possible. Addressing them at the application-wide or software level may not be the best approach. Handling them at the load balancer level could be a more effective solution.

# Checklist

📌 Always:

👍 Most of the time:


e2e-common/vitest.config.mts
```diff
@@ -10,6 +10,9 @@ export default defineConfig({
          * automatically fail for going over the 5 second default timeout.
          */
         testTimeout: process.env.E2E_DEBUG ? 1800 * 1000 : process.env.CI ? 30 * 1000 : 15 * 1000,
+        sequence: {
+            concurrent: false,
```
d46, line 14: Concurrency should be implicitly false otherwise tests are 100% failing

michaelbromley:
Hi,
Thanks for this contribution! Just an update: right now I am focusing on the next tasks for the v3.1 release, but I definitely intend to properly review this and work with you as soon as that is out of the way. Thanks for your patience!

dlhck:
@coderabbitai review

dlhck:
@d46 could you please fix the conflicts so that we can do a proper manual review? Thank you!

PR #3151, opened 2024-10-19
fix: Refund order dialog is showing the wrong field for prorated unit…

### LINKED ISSUE #3150: Refund order dialog "prorated unit price" display is incorrect (shows wrong field)
**Describe the bug**

Prorated unit price in the refund order dialog is showing the wrong value. Here is the code from `refund-order-dialog.component.html`:

```html
<vdr-dt2-column id="prorated-unit-price" [heading]="'order.prorated-unit-price' | translate">
            <ng-template let-line="item">
                {{ line.unitPriceWithTax | localeCurrency : order.currencyCode }}
```

It *should* be showing `line.proratedUnitPriceWithTax`

**To Reproduce**
Use Vendure (latest master as of commit dcf7b0a023532)

**Expected behavior**
It should show the correct field

Not much more to say – it's a pretty self evident issue. I'm looking at an order where I can see the values returned from graphql for the prorated unit price with tax and it's not what is displayed, so I looked at the template and found it's incorrect as well. It does use the correct value for the refund amount, but it was quite confusing when I couldn't see why it was giving me that value.

(actually I'm still trying to figure out why it thinks that's the right value, but that's a different problem)

taxilian:
Fixes #3150

### DIFF
--- a/packages/admin-ui/src/lib/order/src/components/refund-order-dialog/refund-order-dialog.component.html
+++ b/packages/admin-ui/src/lib/order/src/components/refund-order-dialog/refund-order-dialog.component.html
@@ -50,7 +50,7 @@
         </vdr-dt2-column>
         <vdr-dt2-column id="prorated-unit-price" [heading]="'order.prorated-unit-price' | translate">
             <ng-template let-line="item">
-                {{ line.unitPriceWithTax | localeCurrency : order.currencyCode }}
+                {{ line.proratedUnitPriceWithTax | localeCurrency : order.currencyCode }}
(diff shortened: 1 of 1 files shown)


dlhck:
@taxilian great catch - thanks. Could you please sign the CLA (description in the comment above), so that we can merge it 💪🏼

taxilian:
---
I have read the CLA Document and I hereby sign the CLA.

---

dlhck:
recheck

michaelbromley:
Thanks for the fix!

Issue #3153, opened 2024-10-19
Order merge with UseGuestStrategy delete all lines existing in both orders


gkielwasser:
**Describe the bug**
If a product exists in the guest order and in the customer order, it will be deleted.

**To Reproduce**
1. Create an order with a customer
2. Add product A
3. Logout
4. Add product A ( as a guest )
5. Login with the customer

Observe: the cart is empty

Or add an e2e test. This won't pass and will return an empty array.:
```
it('UseGuestStrategy with conflicting lines', async () => {
        const result = await testMerge({
            strategy: new UseGuestStrategy(),
            customerEmailAddress: customers[3].emailAddress,
            existingOrderLines: [
                { productVariantId: 'T_5', quantity: 1 },
                { productVariantId: 'T_3', quantity: 1 },
            ],
            guestOrderLines: [{ productVariantId: 'T_5', quantity: 3 }],
        });

        expect(
            result.lines.sort(sortById).map(line => ({
                productVariantId: line.productVariant.id,
                quantity: line.quantity,
            })),
        ).toEqual([{ productVariantId: 'T_5', quantity: 3 }]);
    })
```

**Expected behavior**
The guest order lines shouldn't be deleted and appears in the final order.

**Additional**
https://github.com/vendure-ecommerce/vendure/blob/dcf7b0a023532e815b7aff0995389db4f9ef6bb7/packages/core/src/service/services/order.service.ts#L1617
The mergedLines lines returned by the strategy itself is correct but then there is a conflict between linesToInsert and linesToDelete.
The lines seems to be inserted and then deleted.

**Environment (please complete the following information):**
 - @vendure/core version: 3.0.5
- @vendure/core version: 2.2.7

PR #3155, opened 2024-10-19
fix(core): Fix merging order with conflicting products

gkielwasser:
# Description
Merging the guest with the existing order delete any products that would be in both orders. A relevant test was missing.
[related issue](https://github.com/vendure-ecommerce/vendure/issues/3153)

# Breaking changes

Don't think so but we have to ensure it does not affects other merging strategies.

# Checklist

📌 Always:

👍 Most of the time:


gkielwasser:

michaelbromley (approved):
Thanks for the fix & for adding a test for this 👍

PR #3156, opened 2024-10-20
feat(core): Enable GraphQL subscriptions

### LINKED ISSUE #2369: GraphQL Subscriptions
As a headless platform, I believe vendure would greatly benefit from an idiomatic way to register new graphql subscriptions.

Some users have managed to get them working https://github.com/vendure-ecommerce/vendure/discussions/1879 but my personal recent attempts were unfruitful.

What seems to be the approach that is most aligned with vendure's coding standards / architecture would be to use the nest resolver decorators so `@Subscription` can sit along `@Mutation` & `@Query` like in the example below

```ts
@Resolver('Entity')
export class EntityResolver {

  @Query('entity')
  @Transaction()
  async entity(@Args('id') id: number) {
    ...
  }

  @Mutation('deleteEntity')
  @Transaction()
  async deleteEntity(@Args('id') id: number) {
    ...
  }

  @Subscription('entityDeleted')
  async entityDeleted() {
    ...
  }
}
```

dlhck:
# Description

This PR introduces the native support for GraphQL subscriptions. They enable real-time updates via web sockets. 

Closes #2369 

# Breaking changes

Not right now (to be updated)


dlhck:
Currently I am working on the problem that in a websocket context we don't have the usual `req` and `res` objects available in our execution context. This results in lot of problems in different areas of Vendure, mostly in the `auth-guard` and `vendure request context`.

d46:
### Problem Context:

Vendure's GraphQL API supports subscriptions, which operate over WebSockets. Unlike standard HTTP requests where authentication (like a bearer token or session cookie) is sent with every request, WebSockets establish a persistent connection. Authentication information is typically only sent during the initial handshake (connectionParams).

This presents a challenge: while the initial connection might be authenticated, subsequent messages (events pushed from the server via the subscription) might lack the necessary user context within Vendure's RequestContext system, as the standard token extraction mechanisms (checking HTTP headers/cookies via extractSessionToken) won't find a token on these subsequent internal pushes over the established socket.

### Proposed Solution:

To ensure consistent authentication and user context for GraphQL subscriptions, we propose leveraging the persistent nature of the WebSocket connection itself.

Client-Side: The client application connecting via WebSocket should include the user's Vendure-Token within the connectionParams object during the initial WebSocket handshake.

Server-Side: Modify Vendure's core authentication logic to recognize and extract the Vendure-Token from these connectionParams if present.

This allows the server to identify the authenticated user associated with that specific WebSocket connection and correctly populate the RequestContext for any operations related to that subscription.

### Implementation Details:

The core change involves modifying the extractSessionToken function located at packages/core/src/api/common/extract-session-token.ts. This function is used by the AuthGuard (packages/core/src/api/middleware/auth-guard.ts) to determine the session token for incoming requests.

The modification will involve:

Adding logic at the beginning of extractSessionToken to check if the incoming request object (req) contains connectionParams (indicating a WebSocket context) and if those params include a Vendure-Token.

If the Vendure-Token is found in connectionParams, it will be returned immediately as the valid session token for that connection.

If no token is found in connectionParams, the function will proceed with its existing logic to check for the token in HTTP cookies or the Authorization: Bearer header, based on the configured tokenMethod.


### Result

1. WebSocket Connection: Client connects, sending Vendure-Token in connectionParams (initial connection).

2. Subscription/Query/Mutation Resolver: When a GraphQL subscription/query/mutation operation starts (e.g., subscription { onChannelUpdate }), the main resolver function decorated with @Subscription()/@Query/@Mutation is called.

3. AuthGuard: The AuthGuard intercepts this call before the resolver code runs.

4. Token Extraction: The AuthGuard calls getSession(), which calls our modified extractSessionToken(). This function now successfully finds the token in req.connectionParams['Vendure-Token'] for the socket connection.

5. Session Retrieval: getSession() uses the token to retrieve the correct CachedSession via sessionService.getSessionFromToken().

6. RequestContext Creation: The AuthGuard then calls requestContextService.fromRequest(), passing the retrieved session. This service builds the fully populated RequestContext (including the apiType, channel, user, session, permissions, etc.).

7. Context Storage: The AuthGuard stores this newly created RequestContext internally, associating it with the request (internal_setRequestContext(req, requestContext, context)).

8. Resolver Execution: The AuthGuard allows the execution to proceed to the actual subscription resolver function.

9. Context Injection: Inside the resolver/query/mutation, decorators like @Ctx() are used to inject the RequestContext. This decorator retrieves the context that the AuthGuard just created and stored in step 7.

PR #3159, opened 2024-10-22
Fix crash when Administrator has custom field that is `internal` but not `readonly`

### LINKED ISSUE #3158: Specifying `internal` Administrator custom field without also specifying `readonly` causes crash
**Describe the bug**

When specifying an internal custom field for the Administrator entity, Vendure crashes on startup. The error is:

```
Error: Unknown type "UpdateAdministratorCustomFieldsInput". Did you mean "UpdateAdministratorInput" or "CreateAdministratorInput"?
    at assertValidSDLExtension (/home/projects/vendure-starter/node_modules/graphql/validation/validate.js:149:11)
    at extendSchema (/home/projects/vendure-starter/node_modules/graphql/utilities/extendSchema.js:63:43)
    at addActiveAdministratorCustomFields (/home/projects/vendure-starter/node_modules/@vendure/core/src/api/config/graphql-custom-fields.ts:299:24)
    at buildSchemaForApi (/home/projects/vendure-starter/node_modules/@vendure/core/src/api/config/configure-graphql-module.ts:171:56)
    at async createGraphQLOptions (/home/projects/vendure-starter/node_modules/@vendure/core/src/api/config/configure-graphql-module.ts:93:25)
    at async InstanceWrapper.useFactory [as metatype] (/home/projects/vendure-starter/node_modules/@nestjs/graphql/dist/graphql.module.js:89:48)
    at async Injector.instantiateClass (/home/projects/vendure-starter/node_modules/@nestjs/core/injector/injector.js:369:37)
    at async callback (/home/projects/vendure-starter/node_modules/@nestjs/core/injector/injector.js:65:34)
    at async Injector.resolveConstructorParams (/home/projects/vendure-starter/node_modules/@nestjs/core/injector/injector.js:144:24)
    at async Injector.loadInstance (/home/projects/vendure-starter/node_modules/@nestjs/core/injector/injector.js:70:13) 
```

**To Reproduce**

1. Add the following to `customFields` in vendure-config.ts:
```ts
Administrator: [{
      name: 'testField',
      type: 'string',
      public: false,
      internal: true
}]
```
2. (Optional; the problem happens either way) Run `npx vendure migrate` and generate a migration for the new custom field.
3. Run `npm run dev`
4. It will crash.

Here is a reproduction in Stackblitz, forked from an existing Vendure example and with all the Vendure dependencies updated:
https://stackblitz.com/edit/vendure-starter-kspu5w?file=src%2Fvendure-config.ts

**Expected behavior**
The app would start and the custom internal field would be available on the Administrator entity (exclusively within plugins, in TypeScript.)

 - @vendure/core version: 3.0.5
 - Nodejs version: v18.20.3 and v20.14.0
 - Database (mysql/postgres etc): It happens at least with SQLite and Postgres

**Additional context**

It works as soon as I take away `internal: true`, so I assume that that's the problem. It seems like when the custom field is internal, the `UpdateAdministratorCustomFieldsInput` type is not created in GraphQL, which makes sense, but some code somewhere still expects it to exist just because there is a custom field.

More specifically, it seems like the problem is here:

https://github.com/vendure-ecommerce/vendure/blob/37775552b38b7d3f7158875756074b34a091fcd0/packages/core/src/api/config/graphql-custom-fields.ts#L291-L299

Just because there is a non-readonly custom field, the `UpdateActiveAdministratorInput` type is being extended with the assumption that the `UpdateAdministratorCustomFieldsInput` exists. However, because the field is `internal`, which from my reading of the docs should act as a superset of `readonly`, the administrator custom fields input type was apparently not created, so the schema is invalid.

I have not checked if the same problem occurs with entities other than `Administrator`.

toBeOfUse:
# Description

This fixes #3158.

Previously, the `addActiveAdministratorCustomFields` function, which adds the custom fields for the Administrator entity into the input type for the `updateActiveAdministrator` mutation, would assume that the `UpdateAdministratorCustomFieldsInput` type exists if there are any non-`readonly` custom fields, even if those custom fields are `internal`. This was incorrect, since `internal` custom fields will not appear in mutation input types, any more than ones that are `readonly`. So, the schema produced would reference that input type, causing a crash if it did not exist, which happened every time there were `internal`, non-`readonly` custom fields.

I also added a test to confirm that the `UpdateAdministratorCustomFieldsInput` type is not referenced when the Administrator entity has only an `internal` custom field.

Also, an irrelevant typo is fixed.

# Breaking changes

Nope.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/config/__snapshots__/graphql-custom-fields.spec.ts.snap
+++ b/packages/core/src/api/config/__snapshots__/graphql-custom-fields.spec.ts.snap
@@ -250,6 +250,15 @@ scalar JSON
 scalar DateTime"
 `;
 
+exports[`addGraphQLCustomFields() > uses JSON scalar in UpdateActiveAdministratorInput if only internal custom fields defined on Administrator 1`] = `
+"scalar JSON
+
+input UpdateActiveAdministratorInput {
+  placeholder: String
+  customFields: JSON
+}"
+`;
+
 exports[`addOrderLineCustomFieldsInput() > Modifies the schema when the addItemToOrder & adjustOrderLine mutation is present 1`] = `
 "type Mutation {
   addItemToOrder(id: ID!, quantity: Int!, customFields: OrderLineCustomFieldsInput = null): Boolean
--- a/packages/core/src/api/config/graphql-custom-fields.spec.ts
+++ b/packages/core/src/api/config/graphql-custom-fields.spec.ts
@@ -4,6 +4,7 @@ import { describe, expect, it } from 'vitest';
 import { CustomFieldConfig, CustomFields } from '../../config/custom-field/custom-field-types';
 
 import {
+    addActiveAdministratorCustomFields,
     addGraphQLCustomFields,
     addOrderLineCustomFieldsInput,
     addRegisterCustomerCustomFieldsInput,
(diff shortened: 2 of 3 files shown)


michaelbromley:
Thanks for this contribution. Everything is perfect - a very good, detailed issue which was easy for me to reproduce, a PR which includes tests, & a concise PR summary. Very well done 🙌

PR #3171, opened 2024-10-28
fix(core/utils): Allow null on idsAreEqual function

twlite:
# Description

I've been getting errors when following multivendor example due to this function. ID sometimes seems to be `null` type, causing unexpected errors. This PR adds null support for `idsAreEqual` function to prevent this issue.

# Breaking changes

This PR shouldn't break existing features.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:


ghimirebibek:
This is an issue which I also faced several times. It would be really nice if this is adapted into vendure.

michaelbromley:
I'm guessing you are seeing nulls due to the configuration of your graphql codegen. I can't see any reason not to also handle nulls, so this looks fine to me. Thanks for the contribution 👍

PR #3176, opened 2024-10-31
feat(admin-ui): Preset filters should preserve query parameters

taxilian:
# Description

Any page that uses data-table2 with presets, when the presets change they lose any query parameters. This includes the currently set "display per page". This can be very frustrating if you use presets to switch quickly between groups of items which number more than 10 because you have to keep moving down and turning it back up.

This fix causes it to keep any existing query parameters except the "filters" and "page" parameters, which are the ones that it uses.

# Breaking changes

No, just changed behavior

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have added or updated test cases - Not applicable, AFAIK
- [x] I have updated the README if needed - Not applicable, AFAIK

### DIFF
--- a/packages/admin-ui/src/lib/core/src/shared/components/data-table-filter-presets/data-table-filter-presets.component.html
+++ b/packages/admin-ui/src/lib/core/src/shared/components/data-table-filter-presets/data-table-filter-presets.component.html
@@ -19,7 +19,7 @@
             </div>
             <a
                 [routerLink]="['./']"
-                [queryParams]="preset.value === serializedActiveFilters ? {} : { filters: preset.value, page: 1 }"
+                [queryParams]="getQueryParamsForPreset(preset.value, serializedActiveFilters)"
             >
                 <div>{{ preset.name }}</div>
             </a>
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hi!

This is a good improvement but it seems to not work reliably in all cases, according to my testing.

1. go to product list
2. create a filter preset (exact type does not matter)
3. navigate away from product list to another page
4. return to product list
5. set items per page to 25
6. click the preset

You will see that the "perPage" query param is lost.

This can be solved by adding this to the `onInit()` method:

```ts
        // When any query param changes, we want to trigger change detection
        // so that the links for each preset are updated.
        this.route.queryParamMap
            .pipe(takeUntil(this.destroy$))
            .subscribe(() => this.changeDetectorRef.markForCheck());
```

taxilian:
pushed a fix; thanks for taking the time to look at and test this. It's really slowing down some of our workflows =]

My last real experience with Angular was about 6 years ago, so I'm beyond rusty dealign with rxjs nonsense. I'm more of a vue fan myself =]

PR #3177, opened 2024-10-31
fix(core): It is possible for featuredAsset to resolve to {id: ID}

taxilian:
# Description

We experienced an issue with trying to add products to an existing order; when we added a new product which was not previously in the order, the featuredAsset would resolve to an object with an ID and nothing else.  We have this fix locally as a resolver, but it's simple and innocuous enough I thought I'd submit it for your consideration to help others.

If this isn't the right way to do this I apologize; we're still new-ish to this project. Let me know how to do it better and I'll fix it =]

# Breaking changes

None

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:
       - (this might not be a bad idea, but I don't know how to reproduce it with a test case)
- [ ] I have updated the README if needed N/A

### DIFF
--- a/packages/core/src/api/resolvers/entity/order-line-entity.resolver.ts
+++ b/packages/core/src/api/resolvers/entity/order-line-entity.resolver.ts
@@ -31,10 +31,13 @@ export class OrderLineEntityResolver {
         @Ctx() ctx: RequestContext,
         @Parent() orderLine: OrderLine,
     ): Promise<Asset | undefined> {
-        if (orderLine.featuredAsset !== undefined) {
-            return orderLine.featuredAsset;
-        } else {
+        // In some scenarios (e.g. modifying an order to add a new item), orderLine.featuredAsset is an object
+        // with only an `id`. Since the resolver expects the featuredAsset to be a full Asset object, we need to
+        // fetch the full Asset object if it's not already populated.
+        if (!orderLine.featuredAsset?.preview) {
             return this.assetService.getFeaturedAsset(ctx, orderLine);
+        } else {
+            return orderLine.featuredAsset;
         }
     }
 


taxilian:
Hmm. These failing tests certainly suggest this isn't the correct solution for backend, but I'm at a bit of a loss since it works fine in my code.  I'm attempting to reproduce the failure on my system so I can hopefully address it.

taxilian:
I am unable to reproduce this issue running e2e tests on my system and would appreciate any guidance.

packages/core/src/api/resolvers/entity/order-line-entity.resolver.ts
```diff
@@ -32,6 +34,18 @@ export class OrderLineEntityResolver {
         @Parent() orderLine: OrderLine,
     ): Promise<Asset | undefined> {
         if (orderLine.featuredAsset !== undefined) {
```
michaelbromley: I believe the tests are failing because you are doing an exact `!==` check for `undefined`. So if `orderLine.featuredAsset` is `null`, then the "if" statement will be true, and the program will go to line 40 and attempt to access `featuredAsset.preview`, which results in the error you see in the e2e tests: 

```
Error: Cannot read properties of null (reading 'preview')
```

So the solution should be to use `!=`, which will check for _both_ `null` and `undefined`.

taxilian:
we'll give that a try... if we're just checking for "truthyness" I'd rather it was specifically doing that, rather than the != which does the same thing but doesn't look like it.

packages/core/src/api/resolvers/entity/order-line-entity.resolver.ts
```diff
@@ -31,7 +33,19 @@ export class OrderLineEntityResolver {
         @Ctx() ctx: RequestContext,
         @Parent() orderLine: OrderLine,
     ): Promise<Asset | undefined> {
-        if (orderLine.featuredAsset !== undefined) {
+        if (!!orderLine.featuredAsset) {
```
michaelbromley: I thought more about this and I think it can be simplified even further to:

```ts
if (!orderLine.featuredAsset?.preview) {
  return orderLine.featuredAsset;
} else {
  return this.assetService.getFeaturedAsset(ctx, orderLine);
}
```

reasoning:
1. less code
2. fewer logical paths == easier to reason about
3. if any future refactors occur to how we fetch a featured asset, they will automatically apply to this code path == easier maintenance.

The comment is still worth having though to explain why we explicitly check for the presence of the `preview` property.

michaelbromley (changes requested):
Nice, test are all passing now. I have one further suggestion to simplify things.

packages/core/src/api/resolvers/entity/order-line-entity.resolver.ts
taxilian: Yeah, that makes sense; I was stuck on "change the least you need to" thinking that if featuredAsset is not truthy then we should avoid calling getFeaturedAsset, but I trust that you know better than I if there are any legitimate cases where that would happen, so I'll make and push the change.

taxilian: unless I'm reading wrong, though, you have your logic backwards, and it should be:

```typescript:
if (!orderLine.featuredAsset?.preview) {
    return this.assetService.getFeaturedAsset(ctx, orderLine);
} else {
    return orderLine.featuredAsset;
}
```

michaelbromley: yes you're right!

michaelbromley:
thank you!

Issue #3173, opened 2024-10-30
registerPageTab does not render component when location is order-list


kkerti:
**Describe the bug**
Following the Page Tab rendering guide, custom component cannot be rendered under location `order-list`.
![demo_vendure](https://github.com/user-attachments/assets/b65b9adf-e6aa-4b7c-a80c-179e2b87628e)

**To Reproduce**
1. Download minimal reproduction of issue: https://github.com/kkerti/vendure-tab-creation-issue
2. Go to `/admin/orders`
3. Click on "Greeting" page tab
4. The **Hello!** text is not visible
5. Refresh the page on url `/admin/orders/greeter`, header **Order** disappears

**Expected behavior**
The custom component should be visible correctly on the page tab route.

**Environment:**
 - @vendure/core version: 3.0.5
 - Nodejs version: v21.7.3
 - Database (mysql/postgres etc): 16-alpine

**Additional context**
Similar issue was mentioned under https://github.com/vendure-ecommerce/vendure/issues/2220, could not reproduce that one.

kkerti:
Solved by merging #3187

PR #3187, opened 2024-11-05
fix(admin-ui): Make registerPageTab work on 'order-list' location (#3173)

kkerti:
# Description

I got into the issue where I wasn't able to add a custom component to location `order-list` with the `registerPageTab` provider. Found out that the culprit is the router configuration in admin-ui for the order part. Under the more complex paths with subroutes `pathMatch: full` is not used, only on more basic pages like _login_ and _dashboard_.

Issue reference is here: #3173

# Breaking changes

If a user has the same order ID like the tab name, that could cause a race condition I assume, but I think this would be unlikely.

# Checklist

📌 Always:

👍 Most of the time:


kkerti:

michaelbromley:
Thanks for the fix!

Issue #1671, opened 2022-07-21
Not able to perform file upload e2e vendure testing


BeepLoveKarki:
I run a file upload mutation on vendure and it always crashes with fatal javascript size issue.

I tried with node version 18.0.0 and v16.3.1 and both had same issue whenever I call "fileUploadMutation" of vendure testing admin or shop client and add files name to filePaths, I get this issue. I have attached the code snippet and 

`const filesToUpload = [
            path.join(__dirname, 'fixtures/assets/dummy.txt'),
        ];`

        `const createdIt = await adminClient.fileUploadMutation({
            mutation:CREATE_MERCHANT,
            filePaths: filesToUpload,
            mapVariables: filePaths => ({
                input: {
                    adminFirstName: "Biplab",
                    adminLastName: "Karki",
                    adminEmail: "biplab@d.co",
                    adminPassword: "abcd",
                    adminPhoneNumber: "1122334",
                    companyCode:"MENZZ",
                    companyName: "Menzz Co. Ltd.",
                    companyAddress: "Kirtipur-2, Kathmandu",
                    companyDescription: "All about men's grooming",
                    customerContactEmail:"info@menzz.co",
                    customerContactPhone:"00000",
                    enabled: true,
                    document: filePaths.map(() => ({ file:null })),
                }
            }),
        })`

The following is my input schema for this:

```
`input CreateMerchantInput {
    adminFirstName: String!
    adminLastName: String!
    adminEmail: String!
    adminPassword: String!
    adminPhoneNumber: String
    companyCode: String!
    companyName: String!
    companyAddress: String
    companyDescription: String
    customerContactEmail: String
    customerContactPhone: String
    enabled: Boolean
    document: Upload
}

createMerchant(input: CreateMerchantInput!): Merchant!

`
```
<img width="1440" alt="Screen Shot 2022-07-21 at 17 19 16" src="https://user-images.githubusercontent.com/29729738/180212349-1fc4587e-c6a8-4dd3-800f-e23ac583320c.png">

michaelbromley:
Hi, thanks for the report. Would you be able to create an actual git repo which I could simply clone and run in order to see the issue? It will make it much more likely that I find the time to look into this sooner.

BeepLoveKarki:
Sure here it is.

https://github.com/BeepLoveKarki/merchant-repo

toBeOfUse:
[...]

PR #3188, opened 2024-11-06
Make test client's `fileUploadMutation` work for more input variable shapes

toBeOfUse:
# Description

In the testing package, Vendure provides a `SimpleGraphQLClient` that has a function called `fileUploadMutation`. This function works for the built-in GraphQL mutations that accept file uploads, but if you add a custom mutation that accepts file inputs, as the [docs](https://docs.vendure.io/guides/developer-guide/uploading-files/#custom-upload-mutations) show you how to do, you will run into problems, since it assumes that it can find the `Upload` inputs in the GraphQL variables just by looking at the first property of each nested object. So, for example, the existing code that creates the FormData body for the upload will go into an infinite loop if there are any string fields before the Upload fields in your GraphQL mutation input. I detailed the problem that I ran into with this in #1671.

So, this PR makes the `fileUploadMutation` method work with mutation inputs with arbitrary shapes.

Also, the two tests for this functionality that exist seem to be written for Jest, and there is no `test` command inside the `testing` package.json to run them. I converted the test file to Vitest and added the `test` command so that they could be run.

# Breaking changes

None: the `fileUploadMutation` should work the same as it always had for the mutations it previously worked with, as shown by the two existing tests passing.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -30,7 +30,8 @@
         "build": "tsc -p ./tsconfig.build.json",
         "watch": "tsc -p ./tsconfig.build.json -w",
         "lint": "eslint --fix .",
-        "ci": "npm run build"
+        "ci": "npm run build",
+        "test": "vitest --config vitest.config.mts --run"
     },
     "bugs": {
         "url": "https://github.com/vendure-ecommerce/vendure/issues"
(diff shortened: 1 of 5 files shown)


michaelbromley:
Wow, I think this is my favourite PR of the week. Tricky investigation, complex code to fix, and even getting previously-lost tests back up and running again! Bravo and thanks so much for your continued contributions 🙌

PR #3189, opened 2024-11-06
fix(admin-ui): Fix Variant Quick-Select Menu

### LINKED ISSUE #3143: Quick jump to variant select has no options after upgrading to 3.0.4
**Describe the bug**
"Quick jump to variant" select has no options.

**To Reproduce**
1. Go to a product variant details page
2. Click on the "Quick jump to variant" select
3. See "No options" placeholder

**Expected behavior**
List of variants should appear in the select

 - @vendure/core version: 3.0.4
 - Nodejs version 20
 - Database (mysql/postgres etc): posgres

**Additional context**
Downgrading vendure to 3.0.3 fixes the issue, presenting the regression in 3.0.4. Probably in the UI component, since the Graphql query `GetProductVariantsQuickJump` is present when inspecting the network tab

toBeOfUse:
# Description

Fixes #3143, in which the product variant quick select menu was always empty. It appears that this was because the Observable that yielded the variants that should be shown in that menu was created with `mapSingle`, but it needed to be unpacked multiple times by the `async` pipe in the component template.

If there is some test case that should be added for it, please let me know.

# Breaking changes

Nope.

# Screenshots

Here it is, working:

image

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/catalog/src/components/product-variant-quick-jump/product-variant-quick-jump.component.ts
+++ b/packages/admin-ui/src/lib/catalog/src/components/product-variant-quick-jump/product-variant-quick-jump.component.ts
@@ -31,14 +31,17 @@ export class ProductVariantQuickJumpComponent implements OnInit {
     @Input() productId: string;
     selectedVariantId: string | undefined;
     variants$: Observable<NonNullable<GetProductVariantsQuickJumpQuery['product']>['variants']>;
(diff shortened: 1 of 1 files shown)


michaelbromley:
This must have regressed in recent refactoring done at the data layer. Thanks for the fix!

PR #3195, opened 2024-11-06
feat(payments-plugin): prevent false positive logging

martijnvdbrug:
# Description

Errors were also logged when Mollie  `expired` hooks are coming in. However, `expired` should state should be ignored.

# Breaking changes

# Screenshots

image
Lowest message was first

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Thank you!

PR #3201, opened 2024-11-07
Channel awareness for FacetService findByCode

d46:
# Description

This PR introduces channel awareness to the findByCode method in the FacetService. The method now uses a channel-aware query to ensure that facets are retrieved in the context of the specified channel. This change enhances the accuracy and relevance of the data returned by the service.

# Breaking changes

This PR does not introduce any breaking changes. The existing functionality is preserved, and backward compatibility is maintained by supporting the old method signature.

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Hi, 
I'm not sure I understand the use case here, because the [Facet.code column](https://github.com/vendure-ecommerce/vendure/blob/bdf2329b9bc929b051cfb6d4cfc68bfd9c9537b4/packages/core/src/entity/facet/facet.entity.ts#L36-L37) is set to be unique. 

Do you have some situation where there is more than 1 version of the same code on different channels?

mschipperheyn:
> Hi, I'm not sure I understand the use case here, because the [Facet.code column](https://github.com/vendure-ecommerce/vendure/blob/bdf2329b9bc929b051cfb6d4cfc68bfd9c9537b4/packages/core/src/entity/facet/facet.entity.ts#L36-L37) is set to be unique.
> 
> Do you have some situation where there is more than 1 version of the same code on different channels?

Yes. We have a custom populate scenario where we import a csv for a 1 or more demo channels. A number of facets are created on the default channel and shared between channels. The FacetValue are channel specific and created on a per channel basis referencing the central facet and all have the same code. We don't want `age-5-7-channel-code` or anything like that. It has to be the same code age-5-7. the `findByCode` gets `all` the facetValues of all the channels. That has created issues for us when we want to import a csv with facet values inside. 

I figured in general all the facet - facetValue relation retrieval should result in a channel specific result and that it would likely solve a lot of hard to diagnose issues we had during the import process.

d46:
Yes, as @mschipperheyn mentioned, that is the same case we are struggling with. We need full channel-aware support for all entities when we have two distinct channels  customers.

dlhck:
@d46 could you please add tests that cover this channel-aware case?

d46:
@dlhck Please use this PR

https://github.com/vendure-ecommerce/vendure/pull/3605

dlhck:
@d46 your linked PR is about product slugs and this is already implemented.

Issue #3219, opened 2024-11-20
Cleanup expired Sessions / Document Session Cleanup Possibilities


pujux:
**Is your feature request related to a problem? Please describe.**

As described in #1425, there currently is no way of cleaning up the expired sessions Vendure creates which can lead to alot of rows in the database over time that just aren't of any use. 

**Describe the solution you'd like**
It should be clearly documented, that there is no periodic cleanup of expired sessions by Vendure and at least there should be functionality exposed for users to call a method that cleans up all expired sessions.

Furthermore, documentation on how to setup a custom CLI script that will call this method would be good. Even better, this custom CLI script could be shipped with the [@vendure/create](https://github.com/vendure-ecommerce/vendure/tree/master/packages/create) tool so that users can easily call this script periodically with their preferred way of scheduling script execution.

**Describe alternatives you've considered**
Writing a script on my own to run a DB query and execute a delete query against all expired sessions.

michaelbromley:
The docs could be added to the deployment section under the heading "running scheduled tasks" which includes the general approach as well as the specific example of cleaning up sessions.

michaelbromley:
This is covered since v3.3.0: there's a built-in `cleanSessionsTask` that's pre-registered in the default config and runs once the `DefaultSchedulerPlugin` is added to your plugins. The scheduled-tasks guide in the docs covers the setup, including the migration needed for the scheduler tables. Closing as completed.

PR #3221, opened 2024-11-20
Document ways to cleanup expired sessions

pujux:
# Description

Issue: #3219

This pull request adds a method to the SessionService that will cleanup expired sessions. Furthermore it adds a note to the SessionService docs to explain that sessions are not automatically cleaned up which references the new How-to guide demonstrating how to use Stand-alone CLI scripts to automate the process of cleaning up sessions.

Also updated the `docusaurus-theme-search-typesense` package version to make docs generation work again.

# Breaking changes

N/A

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:


packages/core/src/service/services/session.service.ts
```diff
@@ -298,6 +314,28 @@ export class SessionService implements EntitySubscriberInterface {
...
+     * @since 3.1.0
+     */
+    async cleanupExpiredSessions(ctx: RequestContext): Promise<void> {
+        const now = new Date();
+        const sessions = await this.connection
+            .getRepository(ctx, Session)
+            .createQueryBuilder('session')
+            .select('session.id')
+            .where('session.expires <= :now', { now })
+            .getMany();
```
michaelbromley: For existing apps that have been in prod for a while, there might be millions of rows being selected here. I see we only select the id which is good, but I'm wondering whether there could still be issues here?

docs/docs/guides/how-to/expired-session-cleanup/index.md
```diff
@@ -0,0 +1,49 @@
+---
```
michaelbromley, line 1: I think we should have a more general topic of "how to run scheduled tasks" with a general explanation of the problem and solution and then the session cleanup as the main example. In future we can then add more examples like cleaning up abandoned orders.

packages/core/src/service/services/session.service.ts
pujux: I could change this to just run a DELETE query for every row where `session.expires <= :now` 
-> that should not pose any issues I would say, but we also wouldn't be able to retrieve the sessionIds thereafter and thus not be able to delete them from the sessionCacheStrategy -> which I think wouldn't even be necessary since they are already expired so should not be in the cache anymore anyway (?) - haven't had a look at the internals of the sessionCacheStrategy

docs/docs/guides/how-to/expired-session-cleanup/index.md
pujux: good idea! should I put that into the "Developer Guide" section under "Advanced Topics" just like the "Migrating from v1" topic? 
image

michaelbromley: Yes, good idea 👍

packages/core/src/service/services/session.service.ts
michaelbromley: The Redis session cache uses the built-in Redis TTL mechanism to evict old entries, and the default session cache is in-memory anyway, so yeah that should not be an issue.

pujux:
I'll pick this back up ASAP but I'm currently not able to dedicate time to it.

michaelbromley:
This is superseded by https://github.com/vendure-ecommerce/vendure/pull/3463 - in v3.3 we will ship with a built-in scheduled task for cleaning sessions.

Thanks for your input though!

Issue #3095, opened 2024-09-29
Assign Permission to Users per Channel


Draykee:
**Is your feature request related to a problem? Please describe.**
Currently, assigning permissions in our system is rigid and inefficient because roles are directly tied to users and channels independently. This setup forces us to create multiple roles for each channel even if they share similar permissions. For instance, if I want to assign 'admin' (role) permissions to a user for Channel X and only 'support' (role) permissions for the same user on Channel Y, I must create duplicate roles for each channel, when I want to have each admin and support roles for the channels. This results in a complex and error-prone permission management process once you work with multiple channels and roles.

**Describe the solution you'd like**
I propose a more flexible permission assignment system where roles can be assigned based on a combination of user and channel. Instead of binding a role solely to a user or a channel, we would assign roles to a pair of user + channel (`UserChannelRole`). This approach would allow for precise control over permissions on a per-channel basis without the need to duplicate roles across different channels. For example, a user could be assigned the 'admin' role on Channel X and 'support' on Channel Y with just a single assignment operation per channel.

![image](https://github.com/user-attachments/assets/e3e0645c-9229-42ad-baa5-d76d7a848fec)

**Describe alternatives you've considered**
I can't think of any simpler alternative. The current alternative is to create new roles every time.

**Additional context**
Implementing this feature would significantly streamline our permission management process, reducing administrative overhead and the potential for errors.

DanielBiegler:
Yes please. For anyone that wants a concrete example, see the POC multi vendor plugin inside [the service](https://github.com/vendure-ecommerce/vendure/blob/1b4c50fffcac08d0f256eff875ddd1023499a580/packages/dev-server/example-plugins/multivendor-plugin/service/mv.service.ts#L138-L172).

https://github.com/vendure-ecommerce/vendure/blob/1b4c50fffcac08d0f256eff875ddd1023499a580/packages/dev-server/example-plugins/multivendor-plugin/service/mv.service.ts#L138-L172)

You gotta create `${shopCode}-admin` roles per seller which becomes a PITA quite quick once you need to add/remove a permission for all "*-admin" roles. Or is there an ergonomic way to deal with it right now @Draykee ? I dont think so?

Draykee:
I also had to solve it like this as well. I will most likely run a database query to add/remove permissions to roles.

We already have 50+ roles, and we didn't even fully launch the feature yet. In the future we also need admin, moderator, etc. roles per channel, so the amount of roles will at least 3x.

[...]

PR #3222, opened 2024-11-20
feat(core): Refactor how permissions get serialized for sessions into using a new strategy

DanielBiegler:
# Preamble

This is a proof of concept for #3095 

# Description

Since people are asking for more details on how this works I'll copy the docs here as quick reminder:

```ts
/**
 * @description
 * This plugin extends the Admin-API and provides a {@link ChannelRolePermissionResolverStrategy} which
 * fundamentally changes how Vendure resolves {@link Permission}s for {@link User}s.
 *
 * By default with the included {@link DefaultRolePermissionResolverStrategy}
 * Vendure stores and resolves permissions by assigning {@link User}s
 * `N` {@link Role}s of which each role can relate to `M` {@link Channel}s.
 *
 * This works well, but it also means you cannot share roles between channels while restricting
 * your users exclusively to their own channel, because channels are tied to the role.
 *
 * Imagine a scenario where you have two users called `UserA`, `UserB`, two channels called
 * `ChannelA`, `ChannelB` and a `CatalogManager`-Role which has permissions for creating,
 * updating and deleting products. You want `UserA` to only be able to work on products that
 * belong to `ChannelA` and `UserB` should only be allowed to work on `ChannelB`.
 *
 * Sharing of a general `CatalogManager`-Role between these channels is not possible under the
 * default strategy because once you relate both channels to the role, both users will have
 * permissions to access each others channels, looking like so:
 *
 * ```
 * UserA ═══➤ CatalogManager ═╦═➤ ChannelA
 *                            ╚═➤ ChannelB
 * UserB ═══➤ CatalogManager ═╦═➤ ChannelA
 *                            ╚═➤ ChannelB
 * ```
 *
 * This means in order to isolate them from each other you must duplicate and maintain two separate
 * `CatalogManager`-Role like so:
 *
 * ```
 * UserA ═══➤ CatalogManagerA ═══➤ ChannelA
 * UserB ═══➤ CatalogManagerB ═══➤ ChannelB
 * ```
 *
 * This scales badly and makes it hard to maintain for marketplaces with lots of administrators.
 *
 * This plugin replaces the default strategy with {@link ChannelRolePermissionResolverStrategy}
 * which allows you to share roles between channels, while still being able to restrict users
 * to their dedicated channels. Using the previously mentioned example, adding the plugin lets you
 * share the permissions from a role granularly for any channel, like so:
 *
 * ```
 * UserA ═══➤ CatalogManager ═══➤ ChannelA
 * UserB ═══➤ CatalogManager ═══➤ ChannelB
 * UserC ═══➤ CatalogManager ═══➤ ChannelA
 * UserC ═══➤ CatalogManager ═══➤ ChannelB
 * UserD ═══➤ CatalogManager ═══➤ ChannelD
 * ... etc. ...
 * ```
 *
 * This is especially useful for multi vendor marketplaces where each vendor has their own channel with `N` roles respectively.
 *
 * @docsCategory auth
 * @docsPage ChannelRolePlugin
 * @docsWeight 0
 * @since 3.2.0
 */

```

# Breaking changes

Nothing should change with the default strategy but once you use the new strategy there is a migration needed because the plugin introduces a new Entity (`ChannelRole`) and existing roles will stop working. This makes this feature purely opt-in for people that are interested in multi-vendor setups.

# Screenshots

# ToDo

- [ ] New UI dropdown depending on the strategy, on admin creation and update page
- [ ] Fix/answer remaining TODO comments
- [ ] Migration
- [ ] Docs
- [ ] Unit tests for new strategy
- [ ] E2E tests for new strategy

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

- New Features
  - Introduces a Channel Role Plugin for channel-scoped administrator roles and permissions.
  - Adds GraphQL mutations to create/update channel administrators.
  - Makes the role-permission resolver strategy configurable with a new default.

- Tests
  - Adds e2e test scaffolding for channel-role permission resolution.

- Documentation
  - Inserts a placeholder in the docs sidebar for a future Channel-Role plugin.

- Chores
  - Dev server: adds Reviews plugin and lowers log verbosity to Debug.


packages/core/src/api/resolvers/base/base-auth.resolver.ts
```diff
@@ -143,11 +142,13 @@ export class BaseAuthResolver {
     /**
      * Exposes a subset of the User properties which we want to expose to the public API.
      */
-    protected publiclyAccessibleUser(user: User): CurrentUser {
+    protected async publiclyAccessibleUser(user: User): Promise<CurrentUser> {
         return {
             id: user.id,
             identifier: user.identifier,
-            channels: getUserChannelsPermissions(user) as CurrentUserChannel[],
+            channels: (await this.configService.authOptions.rolePermissionResolverStrategy.resolvePermissions(
```
kevmtt: When `as` is required I usually assume that this line can be improved

packages/core/src/config/auth/channel-role-permission-resolver-strategy.ts
```diff
@@ -0,0 +1,39 @@
...
+        const channelRolePermissions = new Array<UserChannelPermissions>(channelRoleEntries.length);
+        for (let i = 0; i < channelRoleEntries.length; i++) {
+            channelRolePermissions[i] = {
+                id: channelRoleEntries[i].channel.id,
+                token: channelRoleEntries[i].channel.token,
+                code: channelRoleEntries[i].channel.code,
+                permissions: channelRoleEntries[i].role.permissions,
+            };
+        }
+        channelRoleEntries.sort((a, b) => (a.id < b.id ? -1 : 1));
```
kevmtt: What do we gain by sorting `channelRoleEntries` here? Or is this just for readability while debuggen?

packages/core/src/entity/channel-role/channel-role.ts
```diff
@@ -0,0 +1,34 @@
...
+@Entity()
+export class ChannelRole extends VendureEntity implements HasCustomFields {
+    constructor(input?: DeepPartial<ChannelRole>) {
+        super(input);
+    }
+
+    @Column(type => CustomChannelRoleFields)
+    customFields: CustomChannelRoleFields;
+
+    @ManyToOne(type => User)
```
kevmtt: Shall we configure the inverse sides as well? + cascade delete

DanielBiegler: Yes. Forgot it in the beginning but its there now. :)

packages/core/src/config/auth/channel-role-permission-resolver-strategy.ts
DanielBiegler: Added that because this is currently how its done inside `getChannelPermissions`. Sorting got added [here](https://github.com/vendure-ecommerce/vendure/commit/71b8f4cb944e8967206235c3ef1a4bf0aa36ef57#diff-f094d80e8041fc0c9be6dfcdf711415e6b1577e72a6efe55ef1579d885fa28b1R37) 

According to Michael:

> fix(core): Fix inconsistencies in behaviour between DB drivers
> E.g. Postgres & MySQL differ in how they order arrays by default.

packages/core/src/api/resolvers/base/base-auth.resolver.ts
DanielBiegler: Agreed, hate it. Turns out typescript is smart enough to understand the overlap between the types. :bulb:  Removed it now.

packages/core/src/config/auth/role-permission-resolver-strategy.ts
```diff
@@ -0,0 +1,10 @@
+import { InjectableStrategy } from '../../common';
+import { User } from '../../entity';
+import { UserChannelPermissions } from '../../service/helpers/utils/get-user-channels-permissions';
+
+/**
+ * @description TODO
```
kevmtt:
```TS
/**
 * @description
 * A RolePermissionResolverStrategy defines how role-based permissions for a user should be resolved.
 * This strategy is used to determine the permissions assigned to a user based on their roles per channel.
 *
 * By default {@link DefaultRolePermissionResolverStrategy} is used. However, for more complex environments using
 * multiple channels and roles {@link ChannelRolePermissionResolverStrategy} is recommended.
 * 
 * :::info
 *
 * This is configured via the `authOptions.rolePermissionResolverStrategy` properties of your VendureConfig.
 *
 * :::
 *
 * @docsCategory auth
 * @since 3.3.0
 */
 ```

Comment (DanielBiegler) [reply packages/core/src/config/auth/role-permission-resolver-strategy.ts]: I like it, thanks. :+1:  Will add comments once we finalize the PR and everything is set.

File: packages/core/src/service/helpers/utils/get-user-channels-permissions.ts [hunk 5]
```diff
@@ -14,6 +14,7 @@ export interface UserChannelPermissions {
 
 /**
  * Returns an array of Channels and permissions on those Channels for the given User.
+ * @deprecated See `RolePermissionResolverStrategy`
```
martijnvdbrug: Do we really need to deprecate this first? This helper isn't exported, so I would say it's ok to just mention in the changelog that you should use `configService.authOptions.rolePermissionResolverStrategy.resolvePermissions` from now on.

packages/core/src/config/auth/role-permission-resolver-strategy.ts
```diff
@@ -0,0 +1,24 @@
...
+     */
+    persistUserAndTheirRoles(
+        ctx: RequestContext,
+        user: User,
+        /* channelIds: ID[], */ roleIds: ID[],
+    ): Promise<void>;
+    /**
+     * @param user User for which you want to retrieve permissions
+     */
+    resolvePermissions(user: User): Promise<UserChannelPermissions[]>;
```
martijnvdbrug: Maybe a bit nitpicky, but more consistent with the rest of the codebase: `saveUserRoles` and `getPermissionsForUser`, instead of resolve/persists?

martijnvdbrug: Also, the `channelIds: ID[]` seems pretty crucial for this PoC to work? But the PR is still in draft, right?

martijnvdbrug:
I like it already. One question, to mess everything up: Is there any functional benefit of useing `DefaultRolePermissionResolverStrategy` over the new `ChannelRolePermissionResolverStrategy`, or would the latter always be better and more flexible?

I know it makes this PR less complex and less breaking, but if there is no functional benefit over the `DefaultRolePermissionResolverStrategy`, is it worth maintaining for ever? Or would it be better to do a more painful migration now, and be done with it forever?

mohdbk:
> I like it already. One question, to mess everything up: Is there any functional benefit of useing `DefaultRolePermissionResolverStrategy` over the new `ChannelRolePermissionResolverStrategy`, or would the latter always be better and more flexible?
> 
> 
> 
> I know it makes this PR less complex and less breaking, but if there is no functional benefit over the `DefaultRolePermissionResolverStrategy`, is it worth maintaining for ever? Or would it be better to do a more painful migration now, and be done with it forever?

Great addition to Vendure. I've been waiting for this change for a while now, and it is required in order to go to production. To be honest, I'm not sure but I do think replacing this with the default role permission resolver strategy would be better.

Appreciate your support and contribution to this, it will help us a lot 😃

DanielBiegler:
FYI: Questions have been answered in private but I'll resurface them here to be visible for future interested parties (cc @mohdbk )

> I'm not sure but I do think replacing this with the default role permission resolver strategy would be better

and

> Or would it be better to do a more painful migration now, and be done with it forever?

Talked about it with Michael and our consensus is that breaking existing Vendure instances is a non starter, especially since technically nothing is "wrong" with the existing handling of Roles/Permissions, it just doesnt fit marketplaces specifically. Marketplaces are a niche and not the norm, so it feels more correct to have this as purely opt in.

There might come a point in the future where this new channel role strategy becomes the default, like for example the next major Vendure version, but for now keeping it a strict opt-in is a lot nicer to the existing user base. :-)

mohdbk:
Appreciate your efforts @DanielBiegler, waiting for this to complete the development of our marketplace, any updates on this?

DanielBiegler:
@mohdbk current status is: Generally the idea and logic in itself works, but because I'm trying to fit everything into an optional plugin some details get finicky. If you'd edit Vendures Source things could get fixed rather easily but currently with the plugin approach two things are annoying:

When bootstrapping and populating a new Vendure instance `ensureSuperAdminExists` gets called, this is especially important inside of test suites so we can test the plugin properly.

Normally in `ensureSuperAdminExists` roles just get added to the roles array 

```ts
  createdAdministrator.user.roles.push(superAdminRole);
  this.connection.getRepository(ctx, User).save(createdAdministrator.user, { reload: false });
```

We have to replace that with the strategy so it gets persisted correctly but probably (?) due to execution order between `OnApplicationBootstrap` and `OnModuleInit` the connection inside the strategy is undefined. Not sure how to work around that.

Example:

```ts
private async ensureSuperAdminExists() {
  // ... skip

  const { id } = await this.connection.getRepository(ctx, Administrator).save(administrator);
  const createdAdministrator = await assertFound(this.findOne(ctx, id));
  
  const superAdminRole = await this.roleService.getSuperAdminRole();
  const defaultChannel = await this.channelService.getDefaultChannel();
  
  // --- The following fails due to undefined connection inside the strategy
  await this.configService.authOptions.rolePermissionResolverStrategy.saveUserRoles(
      ctx,
      createdAdministrator.user,
      [{ channelId: defaultChannel.id, roleId: superAdminRole.id }],
  );

  // ... skip rest
```

Thats a blocker right now, I need a clever solution here. Feel free to point a PR at my fork if you want to help out. Maybe its enough to pass in the `this.connection` into the strategy directly? Could that have unwanted sideeffects?

Secondly its quite annoying that you cant remove fields from a graphql type when extending the API. So currently we're extending the `CreateAdministratorInput` like so:

```gql
input ChannelRoleInput {
    channelId: ID!
    roleId: ID!
}
extend input CreateAdministratorInput {
    channelRoles: [ChannelRoleInput!]!
    # Sadly we cant just overwrite it like this :-(
    # roleIds: [ID!] @deprecated(reason: "Use \`channelRoles\` due to ChannelRolePlugin")
}
```

And `roleIds` has the type `[ID!]!` which means even after using the plugin you must always pass in an empty roles array. Yeah not super critical but annoying because it can lead to confusion. Maybe we should overwrite both default resolvers to throw, and make two new ones to be very explicit?

DanielBiegler:
So I investigated why the `RolePermissionResolverStrategy` fails to initialize injected properties like the transactional connection.

1. I did include the strategy in the config modules `private getInjectableStrategies()` - This is important because the strategies get initialized in the above `initInjectableStrategies()` which gets called in the config modules `onApplicationBootstrap` hook. This is correct, but keyword being this specific lifecycle hook.
2. The `InitializerService` which is responsible for calling `await this.administratorService.initAdministrators();` hooks into `onModuleInit` which runs **BEFORE** `onApplicationBootstrap` hooks.
3. So after nest boots up and starts calling all the needed hooked in modules, `InitializerService` starts initializing multiple things like channels, roles and importantly admins **BEFORE** the config module was able to call `initInjectableStrategies()` - which only gets called after, at application bootstrap.
4. This results in this PRs introduced `DefaultRolePermissionResolverStrategy` not having its `init(injector: Injector)` function called in time, rendering the connection as undefined.

I quickly skimmed through all the initializer functions and this bug has been hiding until today because none of the essential services used injectable strategies in their initialization. So I believe this is a general problem even outside of this specific PR. As soon as Vendure introduces some strategy, thats needed when initializing, this bug will occur again.

dlhck:
@coderabbitai review

mohdbk:
> @mohdbk current status is: Generally the idea and logic in itself works, but because I'm trying to fit everything into an optional plugin some details get finicky. If you'd edit Vendures Source things could get fixed rather easily but currently with the plugin approach two things are annoying:
> 
> When bootstrapping and populating a new Vendure instance `ensureSuperAdminExists` gets called, this is especially important inside of test suites so we can test the plugin properly.
> 
> Normally in `ensureSuperAdminExists` roles just get added to the roles array 
> 
> ```ts
>   createdAdministrator.user.roles.push(superAdminRole);
>   this.connection.getRepository(ctx, User).save(createdAdministrator.user, { reload: false });
> ```
> 
> We have to replace that with the strategy so it gets persisted correctly but probably (?) due to execution order between `OnApplicationBootstrap` and `OnModuleInit` the connection inside the strategy is undefined. Not sure how to work around that.
> 
> Example:
> 
> ```ts
> private async ensureSuperAdminExists() {
>   // ... skip
> 
>   const { id } = await this.connection.getRepository(ctx, Administrator).save(administrator);
>   const createdAdministrator = await assertFound(this.findOne(ctx, id));
>   
>   const superAdminRole = await this.roleService.getSuperAdminRole();
>   const defaultChannel = await this.channelService.getDefaultChannel();
>   
>   // --- The following fails due to undefined connection inside the strategy
>   await this.configService.authOptions.rolePermissionResolverStrategy.saveUserRoles(
>       ctx,
>       createdAdministrator.user,
>       [{ channelId: defaultChannel.id, roleId: superAdminRole.id }],
>   );
> 
>   // ... skip rest
> ```
> 
> Thats a blocker right now, I need a clever solution here. Feel free to point a PR at my fork if you want to help out. Maybe its enough to pass in the `this.connection` into the strategy directly? Could that have unwanted sideeffects?
> 
> Secondly its quite annoying that you cant remove fields from a graphql type when extending the API. So currently we're extending the `CreateAdministratorInput` like so:
> 
> ```gql
> input ChannelRoleInput {
>     channelId: ID!
>     roleId: ID!
> }
> extend input CreateAdministratorInput {
>     channelRoles: [ChannelRoleInput!]!
>     # Sadly we cant just overwrite it like this :-(
>     # roleIds: [ID!] @deprecated(reason: "Use \`channelRoles\` due to ChannelRolePlugin")
> }
> ```
> 
> And `roleIds` has the type `[ID!]!` which means even after using the plugin you must always pass in an empty roles array. Yeah not super critical but annoying because it can lead to confusion. Maybe we should overwrite both default resolvers to throw, and make two new ones to be very explicit?

Hi @DanielBiegler 

Would you be interested in the below?
I’m planning to submit a PR to Vendure core shortly to get some feedback on this.

The proposal introduces two new schema directives to make GraphQL extension more flexible for plugins:

```graphql
directive @vendureRemove on INPUT_FIELD_DEFINITION | FIELD_DEFINITION
directive @vendureMakeNullable on INPUT_FIELD_DEFINITION | FIELD_DEFINITION
```

These directives allow plugins to modify or relax existing core types directly in SDL.
For example:

```graphql
extend input CreateAdministratorInput {
  channelRoles: [ChannelRoleInput!]!
  roleIds: [ID!]! @vendureRemove @deprecated(reason: "Use `channelRoles` due to ChannelRolePlugin")
}
```

This removes or makes nullable core fields during schema build, solving issues like mandatory `roleIds` when a plugin replaces that logic.

What else do you think is still needed to make your role/permission PR possible once this is in place?

DanielBiegler:
Heya @mohdbk ! :wave: 

Interesting. I dont have researched opinion yet so please anyone feel free to correct me, but my gut feeling says this sounds pretty dangerous. Underlying services expect input types to be a certain way to safely function.

Are there more usecases for this? If it's just this single purpose I feel like this is more of a costly bandaid that comes with a complex maintenance burden, especially if we design new "channel-roles" to be entirely opt-in via plugin, because then its not a problem to demand users just use the new mutations.

New roles model is a quite complex and critical issue not sure yet.

Mhh, I'd need to think about this further to give a proper opinion. Currently focused on other tasks.

mohdbk:
> Heya @mohdbk ! :wave: 
> 
> Interesting. I dont have researched opinion yet so please anyone feel free to correct me, but my gut feeling says this sounds pretty dangerous. Underlying services expect input types to be a certain way to safely function.
> 
> Are there more usecases for this? If it's just this single purpose I feel like this is more of a costly bandaid that comes with a complex maintenance burden, especially if we design new "channel-roles" to be entirely opt-in via plugin, because then its not a problem to demand users just use the new mutations.
> 
> New roles model is a quite complex and critical issue not sure yet.
> 
> Mhh, I'd need to think about this further to give a proper opinion. Currently focused on other tasks.

I agree with you on this, and have decided this would be pretty dangerous.

As this will be an opt in plugin, we can use the new mutations 👍

DanielBiegler:
Quite a bit of time has passed with more and more conflicts creeping in, so I'll close this PR for now. This PR has a lot of useful details and ideas and should be referenced when revisiting this but personally, I currently dont have the bandwidth for such a critical feature. A new fresh PR shall pick this up. Cheers.

PR #3239, opened 2024-11-27
Feat/braintree plugin multi currency

kkerti:
Note: this PR has been remade, as it was originally branched off from master and made against master.

# Description

As briefly described in #3218, this PR adds new feature to the core **BraintreePlugin**, extending the plugin options with an optional object named `merchantAccountIds`.

1. Usage of Braintree Merchant Account Ids are optional, as without them, the Braintree account's default currency (and Merchant Id) is used for authorising transactions.
```ts

// src/braintree/types.ts
export interface BraintreePluginOptions {
  //...other defs
  merchantAccountIds?: BraintreeMerchantAccountIds;
}

// e2e/braintree-dev-server.ts
// Plugin initialisation with currency-merchantAccountId key and value pairs.
BraintreePlugin.init({
    //...other config
    merchantAccountIds: {
        USD: process.env.BRAINTREE_MERCHANT_ACCOUNT_ID_USD,
        EUR: process.env.BRAINTREE_MERCHANT_ACCOUNT_ID_EUR,
    },
}),
```

2. Two of braintree's nodejs API use `merchantAccountId`, the `gateway.clientToken.generate()` and `gateway.transaction.sale()`. Both of these methods accept the `merchantAccountId` with type `string | undefined`. To look up the correct `merchantAccountId` to use, we are using the active `order.currencyCode` and the plugin options.  I did not implement heavy input error checks as the plugin option validity is evaluated runtime thanks to typescript.

```ts
// src/braintree/braintree.handler.ts
// used in clientToken generation and processPayment
const merchantAccountId = lookupMerchantAccountIdByCurrency(
    options.merchantAccountIds,
    order.currencyCode,
);

// src/braintree/braintree-common.ts
export function lookupMerchantAccountIdByCurrency(
    merchantAccountIds: BraintreeMerchantAccountIds | undefined,
    currencyCode: CurrencyCode,
): string | undefined {
    if (!merchantAccountIds || !currencyCode) {
        return undefined;
    }
    const merchantAccountIdForCurrency = merchantAccountIds[currencyCode];
    return merchantAccountIdForCurrency;
}
``` 
3. There was no development server setup in the Braintree plugin, so I took inspiration from the Mollie and Stripe examples. The heavy lifting is done here by using simply a custom payment handler, `clientToken` generation and relying on the braintree-drop-in UI client side. The `braintree-dev-server` sets up an Order, logs info for the developer and exposes the `shopClient` and generated `clientToken` for a test plugin found under `/fixtures/braintree-checkout-test.plugin.ts`.
4.  The additional test plugin exposes a controller with the checkout UI, similar to the Stripe fixture plugin. Results of the attempted payments are dumped in `<div id="result"/>`.
image

6. I added a BraintreeTestPlugin, but I did not add any automated tests, as the `nonce` generated by the client [drop-in UI doesn't have fake-currency-mismatch-nonce test value](https://developer.paypal.com/braintree/docs/reference/general/testing/node#payment-method-nonces). The drop-in UI in the browser has to be manually completed and the value sent to the vendure server. I think the development server introduced here helps a lot in this, but I'm open for suggestions to make complete e2e here. 

# Possible improvement
I think it's not optimal to add the `merchantAccountId` values during plugin initialisation, this would be better off as args for the payment handler itself. But I would rather wait for the new `struct` type to create the currency-merchantAccountId pairs as handler arguments to make it bullet-proof.

# Breaking changes

As far as I'm concerned, there are none. 

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have added or updated test cases - added braintree-dev-server for manual testing


michaelbromley:
Hi!

Thanks for this contribution. I'll need to find some time to properly review & test this - unfortunately it's unlikely in the next week due to workload.

michaelbromley:
Thanks for the nudge :)

This is an amazing PR - great work on all the supporting parts for testing 👏 

OK so it's now May 2025 and we have the `struct` type for custom fields - you mention above that you'd like to use that rather than hard-coded config. Do you want to go ahead with that?

You could add a flag to the init options, like `multipleMerchantAccountSupport: boolean` so that existing users don't need to do a migration for the custom field. Then you can check the value of that in the plugin's `configuration` function:

```ts
configuration: config => {
  if (BraintreePlugin.options.multipleMerchantAccountSupport === true) {
    config.customfields.PaymentMethod.push(...)
  }
  return config;
}
```

kkerti:
@michaelbromley I have played around with `struct`, but I found the current implementation is although stricter, it feels safer through type checking, while the plugin user adds the `CurrencyCode: MERCHANT_STRING` pairs to plugin options. I would leave it as-is. No breaking change either.

kkerti:
@michaelbromley started to patch package this feature set, do you need anything from me to get this into core payments plugins?

I get that you were working 24/7 on the dashboard - which came together fabolously -, still would love to see this merged.

packages/payments-plugin/e2e/fixtures/braintree-checkout-test.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+        const { addPaymentToOrder } = await exposedShopClient.query<
+            AddPaymentToOrderMutation,
+            AddPaymentToOrderMutationVariables
+        >(ADD_PAYMENT, {
+            input: {
+                method: 'braintree-payment-method',
+                metadata: body,
+            },
+        });
+        console.log(addPaymentToOrder);
```
dlhck: remove console log

dlhck:
@kkerti I can't find the flag in the plugins options regarding the support for multiple merchant support, that Michael mentioned in his [comment](https://github.com/vendure-ecommerce/vendure/pull/3239#issuecomment-2886774409). Can you please take a look at that?

kkerti:
> @kkerti I can't find the flag in the plugins options regarding the support for multiple merchant support, that Michael mentioned in his [comment](https://github.com/vendure-ecommerce/vendure/pull/3239#issuecomment-2886774409). Can you please take a look at that?

The plugin in this PR does not add any new customFields to vendure. I opted to define the `merchantAccountIds` object as a typed plugin option. The current implementation in my opinion doesn't need a feature flag, the new option property can be just omitted. Using `undefined` will use the user's braintree account default merchantAccountId.

PR #3245, opened 2024-11-29
fix(payments-plugin): False positive error logging fix

martijnvdbrug:
# Description

Another fix to prevent false positive error logging for Mollie:

* We log an error when 2 transactions with different ID's in state `Setted` are added to the order
* However, Klarna payments are `Authorized` first, sometimes, with autocapture, the order is marked as placed already, but the attached Payment on the order is still in state `Authorized`. It only happens occasionally, most of the times Vendure is quick enough to update `order.placedAt` and the corresponding `payment.state=settled`. Some race condition somewhere...

This PR fixes that. Should be a patch, but unsure if there will be a patch before 3.1?

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/payments-plugin/src/mollie/mollie.service.ts
+++ b/packages/payments-plugin/src/mollie/mollie.service.ts
@@ -279,15 +279,21 @@ export class MollieService {
                 `Unable to find order ${mollieOrder.orderNumber}, unable to process Mollie order ${mollieOrder.id}`,
             );
         }
-        if (mollieOrder.status === OrderStatus.expired) {
-            // Expired is fine, a customer can retry the payment later
+        const mollieStatesThatRequireAction: OrderStatus[] = [
+            OrderStatus.completed,
+            OrderStatus.authorized,
+            OrderStatus.paid,
+        ];
+        if (!mollieStatesThatRequireAction.includes(mollieOrder.status)) {
+            // No need to handle this mollie webhook status
+            Logger.info(
+                `Ignoring Mollie status '${mollieOrder.status}' from incoming webhook for '${order.code}'`,
+                loggerCtx,
+            );
             return;
         }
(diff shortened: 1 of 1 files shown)


packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -279,15 +279,21 @@ export class MollieService {
...
+            OrderStatus.completed,
+            OrderStatus.authorized,
+            OrderStatus.paid,
+        ];
+        if (!mollieStatesThatRequireAction.includes(mollieOrder.status)) {
+            // No need to handle this mollie webhook status
+            Logger.info(
+                `Ignoring Mollie status '${mollieOrder.status}' from incoming webhook for '${order.code}'`,
+                loggerCtx,
+            );
```
martijnvdbrug, line 292: These are all the states we care about, any other incoming state can be ignore, to prevent further unexpected logic being executed.

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -297,14 +303,14 @@ export class MollieService {
...
-        const statesThatRequireAction: OrderState[] = [
+        const vendureStatesThatRequireAction: OrderState[] = [
             'AddingItems',
             'ArrangingPayment',
             'ArrangingAdditionalPayment',
             'PaymentAuthorized',
             'Draft',
         ];
-        if (!statesThatRequireAction.includes(order.state)) {
+        if (!vendureStatesThatRequireAction.includes(order.state)) {
```
martijnvdbrug, line 313: Refactor, because we now have `mollie states that require action` and `vendure states that require action`

michaelbromley:
Thanks! Yes I'll do a small patch before v3.1

Issue #3244, opened 2024-11-29
built-in graphql playground has buggy tooltips


TronkIshere:
**Describe the bug**
In http://localhost:3000/admin-api has a bug that makes code mirror-hint show up and doesn't disappear, and It stays there forever

**To Reproduce**
1. Go to http://localhost:3000/admin-api
2. Use mirror-hint and there it is

**Expected behavior**
I think It should be disappear

**Environment (please complete the following information):**
 - @vendure/core version: 3.0.0
 - Nodejs version: 10.9.0
 - Database (mysql/postgres etc): firebase
 
![Screenshot 2024-11-29 120819](https://github.com/user-attachments/assets/3eb8637f-000b-434d-84aa-9c778d99f4f7)

toBeOfUse:
The GraphQL sandbox that Vendure uses by default is deprecated. Change your vendure-config.ts file to look like this to use the more modern Apollo GraphQL Studio:

```ts
import { ApolloServerPluginLandingPageLocalDefault } from '@apollo/server/plugin/landingPage/default';

export const config: VendureConfig = {

  // ... other stuff ...
  apiOptions: {
    // ... other stuff ...
    shopApiPlayground: false,
    adminApiPlayground: false,
    apolloServerPlugins: IS_DEV
      ? [ApolloServerPluginLandingPageLocalDefault()]
      : [],
    }
  },

  // ... more other stuff ...
};
```

TronkIshere:
@toBeOfUse It worked, tks man

michaelbromley:
Thanks for providing that work-around. I think ultimately we need to consider moving to that in the core. But I'm not sure whether I like that particular one provided by Apollo, because the whole app is served from Apollo's servers. I'd prefer something self-contained.

Ideally, we can use [graphiql](https://github.com/graphql/graphiql) and wrap it into an Apollo plugin that implements the [renderLandingPage](https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages) method.

toBeOfUse:
It's true that the third-party cookies and prompts to log into Apollo are annoying - I'll look into this tonight

toBeOfUse:
[...]

PR #3246, opened 2024-11-30
Replace API Playgrounds with GraphiQL

toBeOfUse:
# Description

The GraphQL playground that Vendure has been using [is an abandoned project](https://github.com/graphql/graphql-playground/issues/1366); Apollo warns us when running `npm install` that it is no longer actively supported, and it has UI issues in modern browsers where tooltips never disappear (#3244). Therefore, it was proposed to replace it with [GraphiQL](https://github.com/graphql/graphiql/), which is a modern equivalent of this playground that does not have its problems.

It was proposed to do this by bundling GraphiQL and serving it from an Apollo plugin that implements `renderLandingPage`. Sadly, this function can literally only return a single HTML string; the Apollo documentation says that this is an intentional limitation, and that if it annoys you, you should implement something in your web framework to serve a playground (so, in this case, a NestJS controller, presumably.) So, this PR adds an Apollo plugin that serves GraphiQL as a single 1.5mb HTML string.

I am not 1000% confident that this is the way to go, this time. For one thing, bundling a package like GraphiQL into a single HTML string is unusual. For another, this adds some quite heavy dependencies to the @vendure/core package: GraphiQL depends on react and react-dom, and then you need some kind of bundler to produce the final result. I added Webpack and some loaders for this purpose, since Webpack is already in use for the Admin UI plugin. Then, I committed the bundled GraphiQL file to this branch, so that it wouldn't have to be re-built by everyone who installed @vendure/core for the first time (so the aforementioned dependencies end up being dev dependencies only.)

So, my questions are: Would it be better to implement this as a NestJS controller, and serve a directory of files for GraphiQL instead of a string? And, would it be better to implement GraphiQL as, perhaps, a separate package, instead of adding React and Webpack to the core, which previously was largely backend code?

I welcome feedback on this PR. But it does solve the problem.

# Breaking changes

This changes the pages that show up when you go to `/admin-api` or `/shop-api` with the playgrounds enabled in vendure-config.ts. That said, this shouldn't break any code or disrupt anyone's workflow, since GraphiQL provides the same basic features.

Except, I didn't implement the playground configuration options, which could be used to configure the Apollo playground:

```ts
export interface RenderPageOptions extends MiddlewareOptions {
  version?: string
  cdnUrl?: string
  env?: any
  title?: string
  faviconUrl?: string | null
}
```

You can see the other settings [here](https://github.com/graphql/graphql-playground/blob/91ed7d8b1c1b76f1bf738b7c97a666360aaab516/packages/graphql-playground-html/src/render-playground-page.ts). Some of them would be very difficult to try to emulate with GraphiQL: for example, the Apollo playground offered very detailed customization of colors through the `EditorColours` interface, but GraphiQL just allows for colors to be changed with predefined CodeMirror themes or CSS variables. Some decision needs to be made regarding how much backward-compatibility to provide.

Also, making this change would perhaps imply that the screenshots of the old playground that exist in the docs should be updated. I could do this; there don't seem to be a ton.

# Screenshots

This was checked using the included dev-server package.

image

When not signed into the Admin API:

image

After signing in, via the Admin UI at `/admin`:

image

When the playground is turned off by setting `adminApiPlayground` (or `shopApiPlayground`) to `false` in the Vendure config:

image

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Wow great! Thanks for taking the time to investigate. I don't mind the react dev dependencies because we already have those as in the admin ui package. I'm wondering whether we can use vite or rollup for the bundling, since we already use them elsewhere for bundling the UI devkit.

Yes, this should be made against the minor branch.

And I have a slight concern that the was it resolves the html string might be brittle in situations where the server code is being itself bundled, which happens when using Nx for example.

It might be a good idea to initially make this plugin opt-in, eg via a `playground: 'graphiql'`  option.

toBeOfUse:
Sorry for the lack of progress on this. I can make the above changes (Vite, making it opt-in) on Monday. However, I haven't worked with Nx before, and I will have to figure out how to test it with server-bundled code

dlhck:
@michaelbromley I think we should directly replace GraphQL playground with GraphiQL and not make it optional. GraphQL Playground is kind of abandoned / unmaintained and doesn't properly work.

dlhck:
Closing this in favor of #3495 - we introduced a new `GraphiqlPlugin`, so that the code is not bundled into core. The  `adminApiPlayground` and `shopApiPlayground` options on the `apiOptions` property of the Vendure config have been deprecated and will be removed in the next major version.

PR #3248, opened 2024-11-30
fix(e2e): Use VITEST_WORKER_ID to generate test server ports

### LINKED ISSUE #3247: Occasional Port Conflicts in CI tests
**Describe the bug**

Tests in CI in Vendure will occasionally fail due to port conflicts:

image

This makes it harder to detect actual errors in CI that might be caused by one particular Node.js version or one particular database being used.

As discussed on Discord, this is probably due to port assignment process, in which a temporary file called `ports.json` is created to store which ports are already in use. If two test suites read this file at the same time, they'll both come up with the same "next available port", and will both try to claim it.

**To Reproduce**
Open a PR and look at the automatic CI tests. Some of them will fail with `EADDRINUSE`.

**Expected behavior**

 - @vendure/core version: 3.0.6
 - Nodejs version: 18.x, 20.x, 22.x
 - Database (mysql/postgres etc): All

toBeOfUse:
# Description

Closes #3247. Instead of using unsynchronized access to a temporary file to determine the next available port, it uses the [`VITEST_WORKER_ID`](https://vitest.dev/guide/migration#envs) environment variable, which should be unique in each test suite, to figure out what port it should be using.

# Breaking changes

Nope, it only applies to the internal test configuration used in Vendure's e2e self-tests.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ docs/static/intro.css*
 docs/public
 docs/data/build.json
 yarn-error.log
-e2e-common/ports.json
 
 # Sensitive or high-churn files:
 .idea/**/dataSources/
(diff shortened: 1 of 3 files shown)


toBeOfUse:
Well, this failed miserably. Apparently learn runs the asset server e2e tests, the elasticsearch plugin e2e tests, and the payments plugin e2e tests in parallel, in separate vitest processes. In a normal repository, the `VITEST_WORKER_ID` approach works; however, in a monorepo where multiple vitest instances are started in parallel, it does not.

PR #3253, opened 2024-12-02
perf(core): disable lastLogin update as an option

alexisvigoureux:
# Description

Disable the last login update (prevent a save on user entity) when a user logs in.

# Breaking changes
No

# Screenshots

context: our users need to be authenticated to view our products.
We receive a lots of request on the authenticate endpoint, and we have a bottleneck on this endpoint.

When we have too much traffic, we have locks on the ‘user’ table

Capture d’écran 2024-11-29 à 15 47 49

Capture d’écran 2024-11-29 à 15 48 16

Here is the request `UPDATE "user" SET "lastLogin" = $1, "updatedAt" = CURRENT_TIMESTAMP WHERE "id" IN ($2)`
This query is not useful for us. This PR proposes a solution to deactivate it and avoid a lock on the user table.
In terms of volume, this represents more than 250k updates of the lastLogin field per day, often at the same connection time.

Not sure where to write the tests, please let me know. 

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/config/vendure-config.ts
+++ b/packages/core/src/config/vendure-config.ts
@@ -473,6 +473,13 @@ export interface AuthOptions {
      * @default DefaultPasswordValidationStrategy
      */
     passwordValidationStrategy?: PasswordValidationStrategy;
+    /**
+     * @description
+     * Disable the last login update (prevent a save on user entity) when a user logs in.
+     * This can be useful when there is a lot of authentication in a short period of time and get lock on the user table.
+     * @since 3.0.7
+     */
+    disableLastLoginUpdate?: boolean
 }
 
 /**
--- a/packages/core/src/config/vendure-config.ts
+++ b/packages/core/src/config/vendure-config.ts
@@ -476,7 +476,7 @@ export interface AuthOptions {
     /**
      * @description
      * Disable the last login update (prevent a save on user entity) when a user logs in.
-     * This can be useful when there is a lot of authentication in a short period of time and get lock on the user table.
+     * This can be useful when there are a lot of authentication in a short period of time and there are locks on the user table.
      * @since 3.0.7
      */
     disableLastLoginUpdate?: boolean
--- a/packages/core/src/service/services/auth.service.ts
+++ b/packages/core/src/service/services/auth.service.ts
@@ -97,8 +97,10 @@ export class AuthService {
         if (ctx.session && ctx.session.activeOrderId) {
             await this.sessionService.deleteSessionsByActiveOrderId(ctx, ctx.session.activeOrderId);
         }
-        user.lastLogin = new Date();
-        await this.connection.getRepository(ctx, User).save(user, { reload: false });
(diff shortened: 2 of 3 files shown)


packages/core/src/service/services/auth.service.ts
```diff
@@ -97,8 +97,10 @@ export class AuthService {
         if (ctx.session && ctx.session.activeOrderId) {
             await this.sessionService.deleteSessionsByActiveOrderId(ctx, ctx.session.activeOrderId);
         }
-        user.lastLogin = new Date();
-        await this.connection.getRepository(ctx, User).save(user, { reload: false });
+        if (user.roles || !this.configService.authOptions.disableLastLoginUpdate) {
+            user.lastLogin = new Date();
+            await this.connection.getRepository(ctx, User).save(user, { reload: false });
+        }
```
DanielBiegler, line 103: I'm curious, why include `user.roles` in the condition? Especially if it specifically gets set above, doesnt that result in this always being cast to `true` since even empty arrays are `true` ?

dlhck: @alexisvigoureux could we please remove the `user.roles` condition? I think it is not necessary. If that's a way how do differentiate between a customer and an administrator, we should do it differently.

packages/core/src/config/vendure-config.ts
```diff
@@ -473,6 +473,13 @@ export interface AuthOptions {
      * @default DefaultPasswordValidationStrategy
      */
     passwordValidationStrategy?: PasswordValidationStrategy;
+    /**
+     * @description
+     * Disable the last login update (prevent a save on user entity) when a user logs in.
+     * This can be useful when there are a lot of authentication in a short period of time and there are locks on the user table.
+     * @since 3.0.7
```
dlhck, line 480: We should use `v3.4.0` here

dlhck (changes requested):
Thank you. Please see my comments. This could go into `v3.4.0`, so please also change the target to `minor`

biggamesmallworld:
Closing in favor of a fresh take-over PR that rebases this onto current `minor`, applies the review feedback (drop the `user.roles` condition, correct `@since`), and adds coverage.

Thank you @alexisvigoureux for the original work and for documenting the user-table lock contention under high auth traffic — you'll be credited as the original author on the replacement PR.

biggamesmallworld:
Replacement PR: https://github.com/vendurehq/vendure/pull/5184

biggamesmallworld:
Replacement PR: https://github.com/vendurehq/vendure/pull/5184

biggamesmallworld:
Replacement PR: https://github.com/vendurehq/vendure/pull/5184

biggamesmallworld:
Replacement PR: https://github.com/vendurehq/vendure/pull/5184

biggamesmallworld:
Replacement PR: https://github.com/vendurehq/vendure/pull/5184

PR #3263, opened 2024-12-06
Add ProductVariantPrice custom fields ui inputs

### LINKED ISSUE #3020: Expose admin-ui for custom fields on ProductVariantPrice.

Custom fields for `ProductVariantPrice` at the moment don't expose an admin-ui.

Custom fields are rendered as it happens for other entities. 

@michaelbromley  mentioned a workaround here: 

> A work-around solution would be to define a [Custom Detail Component](https://docs.vendure.io/guides/extending-the-admin-ui/custom-detail-components/) block on the product variant detail page as part of a UI extension, and in that block you can have a form which exposes the custom fields. The main drawback here is that it will not be hooked up to the overall form save mechanism for the detail page, so changes made to the custom fields would need their own "save" button inside the custom detail component.

https://github.com/vendure-ecommerce/vendure/pull/2654#issuecomment-2296083586

**Additional context**

I'm trying to setup a system to show discounts on product variants in the catalog. As far as I understood, In order to display the discounted prices in the catalog this cannot be made with a promotion, since it gets applied during checkout, so I'm going with custom fields.

At the moment I managed to implement a single channel version adding a Recommended Retail Price custom field to the ProductVariant, the only problem is that this custom field seems to be not channel aware, meaning that I cannot have the same product with different discounts in different channels. It seems that is possible to set the custom field to ProductVariantPrice intead, and that would be channel-aware apparently. However it doesn't expose any field in the admin UI.

kevmtt:
# Description

- Added the custom field component to the ProductVariantPrice UI
- Introduced CreateProductVariantPriceInput type as well, for possible future UI updates
Closes: #3020

# Breaking changes

The admin-api schema needs to be updated. The Typescript should not break though, so it should be safe to update.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


packages/admin-ui/src/lib/catalog/src/components/product-variant-detail/product-variant-detail.component.html
```diff
@@ -195,12 +195,24 @@
...
                     />
+
+                    <vdr-card [title]="'common.custom-fields' | translate" *ngIf="customPriceFields.length">
+                        <vdr-tabbed-custom-fields
+                            entityName="ProductVariantPrice"
+                            [customFields]="customPriceFields"
+                            [customFieldsFormGroup]="price.get(['customFields'])"
+                            [readonly]="!(updatePermissions | hasPermission)"
+                        />
+                        {{ price.get(['customFields'])?.value | json}}
```
kevmtt: Only for debugging - remove later!

Draykee:
image
TODO: Improve the card and blend it into the current design

michaelbromley:
Hi Kevin!

Let's get this one merged. You'll just need to:

1. resolve the merge conflicts
2. regenerate the gql types but without the custom fields

I don't know why the CLA bot is unhappy but you [definitely signed it](https://github.com/vendure-ecommerce/vendure/blob/7d462a0f84cbb1a78ffe174c9cb01e73e991a5de/license/signatures/version1/cla.json#L332) so that won't be a blocker.

dfernandesbsolus:
Hi @Draykee, thanks for you work, i need this feature, can i fix your pull request? :D

Draykee:
I just can't get it working, because of husky. Crazy how simple git merges always turn out to be the most troublesome.

Draykee:
> I just can't get it working, because of husky. Crazy how simple git merges always turn out to be the most troublesome.

Ups.. @michaelbromley I think I made it worse 😅

PR #3282, opened 2024-12-19
Fix Dynamic smtp config require server restart to reflect changes

Ankish76:
# Description

Fixes the issue 

When using dynamic SMTP config per channel, if SMTP config is changed from admin portal.
Nodemailer was not using updated config. 
Had to restart the server to reflect the changes.


Ankish76:

packages/email-plugin/src/sender/nodemailer-email-sender.ts
```diff
@@ -76,22 +76,29 @@ export class NodemailerEmailSender implements EmailSender {
     }
 
     private getSmtpTransport(options: SMTPTransportOptions) {
-        if (!this._smtpTransport) {
+        if (!this._smtpTransport ||
+            JSON.stringify(this._smtpTransport?.options) !== JSON.stringify(options)
```
michaelbromley: I believe there is a theoretical risk here with calling `JSON.stringify` on the `options` object could throw an error if there is a circular data structure in the object. I'm not sure but the options object can potentially include some complex structures like `connection` and `socket` and `tls`.

So it would be safer to extract a function for comparing these stringified objects which wraps the stringification in a try-catch block.

michaelbromley (changes requested):

michaelbromley:
Thanks for your contribution!

PR #3285, opened 2024-12-20
feat(graphql): Allow interceptors on field resolvers

mschipperheyn:
# Description

Allows you to use graphql interceptors on field resolvers. This is by default possible for Queries and Mutations. However, in order for this to work for field resolver, an additional setting is necessary.

Current workaround for some scenarios would be to use a guard. 

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/config/configure-graphql-module.ts
+++ b/packages/core/src/api/config/configure-graphql-module.ts
@@ -120,7 +120,7 @@ async function createGraphQLOptions(
         typeDefs: printSchema(builtSchema),
         include: [options.resolverModule],
         inheritResolversFromInterfaces: true,
-        fieldResolverEnhancers: ['guards'],
+        fieldResolverEnhancers: ['guards', 'interceptors'],
         resolvers,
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,

This looks fine - can you just give me an example use case for how you intend to use field-level interceptors so I can understand it better?

michaelbromley:
I'm hesitant to merge this due to the warning in the NestJS docs: https://docs.nestjs.com/graphql/other-features#execute-enhancers-at-the-field-resolver-level

I'm concerned about any possible performance issues that might be introduced here. From my understanding, any globally-defined interceptors will then execute on all field resolvers for a given query/mutation.

If an app has a number of global interceptors, this may have a noticeable performance impact if returning large responses.

They could code around it, by altering their interceptor implementation to check whether this is a top-level or a field resolver. But that's then not something I want to roll out in a minor.

Therefore I'm going to remove this from the 3.5.1 milestone and it could potentially go in a minor or major.

It would help ease my mind if anyone has time to do any benchmarking on this performance aspect.

Issue #3127, opened 2024-10-10
Add columnId entity references for all relations


mschipperheyn:
**Is your feature request related to a problem? Please describe.**
One of things I often find annoying in Vendure is that many entity many to one relations don't expose the `columnId`. This often requires you to execute an additional query to retrieve the entity in services and generally leads to more query overhead and processing.

Vendure does this in some places but it's more an exception than a rule. 

This would also be a great to have within the context of `customFields`

**Describe the solution you'd like**
Any entity in Vendure should expose a Many to One relation as both a reference and a column

``` 
  @Column()
  customerId: ID

  @Index()
  @ManyToOne(() => Customer, { nullable: false, onDelete: 'CASCADE' })
  customer: Customer
```

**Describe alternatives you've considered**
The alternative in most scenario's where this is an issue, is to retrieve the entity relation through a query.

mschipperheyn:
This should also be combined with an optimizer for customFields entity relations. If you are just querying for the entity.id and no other field, the automatic entity resulution should not query for the relation and just return the `{ id }`.

dlhck:
This feature request has been moved to GitHub Discussions for community voting.
Please upvote the discussion if you would like to see this implemented: https://github.com/vendurehq/vendure/discussions/4241

PR #3291, opened 2024-12-23
Add productVariantId,orderId to OrderLine api, orderId column to OrderLine

mschipperheyn:
# Description

I cannot tell you how much time I spent on performance related issues caused by needing access to the `orderId` from an `OrderLine` without having access to the root `order`. 

Vendure strongly assumes that orderLines and payments are accessed through the order. And normally, that's the case. However, there are a number of scenario's where you might want to access the `orderId` or `productVariantId` without loading those entities. And you might not have the root order entity handy. It's a scenario common to us.

Also, Vendure should IMHO optimize for getting all ManyToOne columns easily through GraphQL and code, without having to load the entire entity if all you need is the id. Which IS a common scenario.

I want to push another PR which addreses ALL of the columns and ManyToOne relations, but let's see how this one goes first :-)

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:
- [x ] My PR is small and contains a single feature

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/schema/common/order.type.graphql
+++ b/packages/core/src/api/schema/common/order.type.graphql
@@ -122,6 +122,7 @@ type OrderLine implements Node {
     createdAt: DateTime!
     updatedAt: DateTime!
     productVariant: ProductVariant!
+    productVariantId: ID!
     featuredAsset: Asset
     "The price of a single unit, excluding tax and discounts"
     unitPrice: Money!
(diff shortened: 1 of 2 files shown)


michaelbromley:
I 100% agree with you that we should support getting IDs everywhere without loading the relations.

Are the tests failing for you locally?

(relates to #3127)

PR #3313, opened 2025-01-10
Recommend global installation for Vendure CLI in Documentation

wpplumber:
# Description

This PR updates the installation instructions for the Vendure CLI. Previously, the documentation suggested installing the Vendure CLI locally in each project. However, I recommend installing the Vendure CLI globally to streamline the developer workflow and avoid redundant installations.

### Key Changes:

**Global Installation**: Installing the Vendure CLI globally via `npm install -g @vendure/cli` eliminates the need to install it in each individual project. This makes the CLI available for use across all projects on the system, improving the developer experience.

  **Local Installation** (Optional): Local installation instructions remain available for users who prefer a project-specific setup, but the global installation option should be the default for most use cases.

# Breaking changes

No

# Screenshots

No

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/docs/docs/guides/developer-guide/cli/index.md
+++ b/docs/docs/guides/developer-guide/cli/index.md
@@ -17,20 +17,20 @@ codebase to integrate new functionality.
 The Vendure CLI comes installed with a new Vendure project by default from v2.2.0+
 :::
 
-To manually install the CLI, run:
+To install the Vendure CLI globally, run:
 
 <Tabs groupId="package-manager">
 <TabItem value="npm" label="npm" default>
 
 ```bash
-npm install -D @vendure/cli
+npm install -g @vendure/cli
 ```
 
 </TabItem>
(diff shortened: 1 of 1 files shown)


wpplumber:

michaelbromley:
Hi,

Thanks for the suggestion, but I think that it is preferable to install the CLI locally to the project in question. Why?

- Installing globally ties it to the global node version. Many people use node version managers like nvm to swap between different versions. The global version is only available for the version on which it was installed. This leads to the developer needing to globally install it multiple times when they switch node versions.
- When installed locally, the CLI version is tied to the local Vendure version. So we eliminate the potential risk of having e.g. an outdated global CLI version which is generating scaffold which is either incompatible or just outdated in a newer project.
- You may also have multiple Vendure projects running multi versions. But there can only be a _single_ globally installed CLI version.

PR #3319, opened 2025-01-16
fix(core): Clear previous line adjustments before testing order item promotions

putermancer:
# Description

`OrderCalculator#applyOrderPromotions` clears out all existing `DISTRIBUTED_ORDER_PROMOTION`s before testing promotions. `applyOrderItemPromotions` does not currently do the same, the result of which is that when conditions / actions are encountered, previous adjustment data may be present on the line items in question.

My specific use case was when creating a promotion of the type "apply a bulk purchase discount, but only to items that aren't already discounted," e.g. when there are 10+ of a product and they aren't already discounted.

**Without this change**: upon reaching 10 items, the line items become discounted. Upon reaching 11 items, the test for "are there adjustments applied" returns true (the previous adjustment), resulting in the discount being removed. Bump the quantity to 12 and it applies again.

**With this change**: it works properly when reaching 10, 11, 12, and on.

# Breaking changes

No.

# Checklist

📌 Always:

👍 Most of the time:
- [x] (N/A) I have updated the README if needed

### DIFF
--- a/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
+++ b/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
@@ -1182,6 +1182,71 @@ describe('OrderCalculator', () => {
                     expect(order.subTotalWithTax).toBe(5719);
                 });
             });
+
+            it(`clear previous promotion state before testing`, async () => {
+                const noLineDiscountsCondition = new PromotionCondition({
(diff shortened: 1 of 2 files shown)


putermancer:

putermancer:
Closing this to clean up the commit -- an instance of `order` should be `_order`

PR #3327, opened 2025-01-21
Add ProductVariantPrice custom fields ui inputs

### LINKED ISSUE #3020: Expose admin-ui for custom fields on ProductVariantPrice.

kevmtt:
# Description
2nd attempt of: https://github.com/vendure-ecommerce/vendure/pull/3263

- Added the custom field component to the ProductVariantPrice UI
- Introduced CreateProductVariantPriceInput type as well, for possible future UI updates
Closes: #3020

# Breaking changes

The admin-api schema needs to be updated. The Typescript should not break though, so it should be safe to update.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


Draykee:

packages/asset-server-plugin/e2e/graphql/generated-e2e-asset-server-plugin-types.ts
kevmtt, line 1: No clue where there are changes here.. codegen did it.

PR #3340, opened 2025-01-30
Feat: Allow skipping complexity check in Harden Plugin

martijnvdbrug:
# Description

We are using the Harden Plugin, but would like to be able to skip the complexity check for storefront SSR builds. 

This PR introduces an optional `skip` function, inspired by [express-rate-limit](https://express-rate-limit.mintlify.app/reference/configuration#skip), that allows the consumer to skip certain requests.

Screenshots below we're tested with the following config:
```ts
  skip: (ctx) => {
    return !!ctx.request.http?.headers.get("x-skip-complexity")
  }
```
This skips the complexity check when the header `x-skip-complexity` is present. In production use, you would check for a token or something.

# Breaking changes

No

# Screenshots

Complex query without header
image

The same complex query with custom header
image

And, another test without a `skip` function specified.
image

# Checklist

📌 Always:

👍 Most of the time:


martijnvdbrug:
@michaelbromley  Not sure why tests are failing, HardenPlugin doesn't seem to be included in any e2e tests. Someone else broke the minor branch maybe?

packages/harden-plugin/src/middleware/query-complexity-plugin.ts
```diff
@@ -20,11 +20,15 @@ import { HardenPluginOptions } from '../types';
 export class QueryComplexityPlugin implements ApolloServerPlugin {
     constructor(private options: HardenPluginOptions) {}
 
-    async requestDidStart({ schema }: GraphQLRequestContext<any>): Promise<GraphQLRequestListener<any>> {
+    async requestDidStart(ctx: GraphQLRequestContext<any>): Promise<GraphQLRequestListener<any>> {
         const maxQueryComplexity = this.options.maxQueryComplexity ?? 1000;
         return {
             didResolveOperation: async ({ request, document }) => {
-                if (isAdminApi(schema)) {
+                if (this.options.skip?.(ctx)) {
```
twlite: It would be great if this supported `Awaitable<boolean>` in addition to `boolean`. Supporting promises would allow developers to include external resources, such as making a third-party validation request.

martijnvdbrug: I purposely did not do that, because it can have severe performance impact if you do anything async in every request.

But I guess we can move that responsibility to the consumer.... Let me add it.

martijnvdbrug: It has been added. And thanks for the review!

packages/harden-plugin/src/types.ts
```diff
@@ -79,4 +80,16 @@ export interface HardenPluginOptions {
      * @default 'prod'
      */
     apiMode?: 'dev' | 'prod';
+    /**
+     * @description
+     * Allows you to skip the complexity check for certain requests.
```
michaelbromley, line 85: I think we should note that the `ctx` is _not_ the `RequestContext` object here, since I think a lot of people will make that assumption based on typical Vendure patterns.

Maybe even rename it to `context` to make it even more explicit.

michaelbromley (commented):
Great addition. I made a suggestion on the naming/documentation.

packages/harden-plugin/src/types.ts
martijnvdbrug: @michaelbromley  Valid concern. Fixed :-)

michaelbromley (approved):

michaelbromley:
Thanks!

Issue #3344, opened 2025-02-03
Bump @sentry-node dependency in @vendure/sentry-plugin to 8.x


kamui:
**Is your feature request related to a problem? Please describe.**
We can't upgrade to Sentry 8 because @vendure/sentry-plugin has a dependency on @sentry/node v7. 
**Describe the solution you'd like**
Bump the @sentry/node dependency in @vendure/sentry-plugin to v8 and go through the migration documentation to upgrade.

**Describe alternatives you've considered**
None

**Additional context**
v8 was released about 9 months ago. It was overhauled to be powered by OpenTelemetry under the hood. It also added automatic integration with NestJS, express, graphql, postgres, etc.

michaelbromley:
Hi,

Would you be interested in contributing a PR for this?

kamui:
Added a PR, but I'm not a 100% the graphql trace stuff is working as intended since I never really used the built in graphql resolver tracing.

michaelbromley:
This ended up being resolved by the bigger migration in issue 3744, which moved the plugin from @sentry/node to @sentry/nestjs (released in v3.4.4 — note it was a breaking change with a new two-step setup). The sentry-plugin has since moved out of this repo to vendurehq/community-plugins, so any further Sentry SDK work belongs over there. Closing as resolved.

PR #3359, opened 2025-02-10
feat(sentry-plugin): Upgrade sentry-plugin to Sentry v8

kamui:
# Description

Currently `@sentry/node` is on v9, `@vendure/sentry-plugin` depends on v7. This PR makes an incremental upgrade to v8 before v9. I tried going to v9 directly but ran into more breakages, so I thought I'd go to v8 first before v9.

Because v8 doesn't work the same way in v7, I had to make some guesses about where to put things and definitely want a gut check on that.

Related Issue: https://github.com/vendure-ecommerce/vendure/issues/3344

# Breaking changes

`@sentry/node` v9 makes a big [breaking change](https://github.com/getsentry/sentry-javascript/blob/develop/docs/v8-new-performance-apis.md) with the way they modeled `Transaction` and span methods: `startTransaction` and `startChild()`. 

In the old way, there was a `Transaction` root object, then there would be a nested tree of spans, which is why you would call `startTransaction` and `startChild()`. In the new way, everything is just a span, no more Transactions.

There shouldn't be any breaking change if you just passed in `request.operationName` to override the name. If you had previously accessed the root `Transaction` via:

```
const transaction = contextValue.req[SENTRY_TRANSACTION_KEY];
```

This will no longer work since there's no longer a concept of `Transaction` in v8. Since you don't chain spans off the transaction anymore, I added a request variable called `startInactiveSpan` so you can basically call the span function off the root `Sentry` object anywhere. Before you would call `startChild()`, now yoou call `startInactiveSpan()`.

```
const startInactiveSpan: StartInactiveSpanFunction | undefined =
            contextValue.req[SENTRY_START_SPAN_INACTIVE_KEY];
```

and you can use it like:

```
span = startInactiveSpan({
    name: 'GraphQLResolver',
    op: 'resolver',
});
span?.setAttribute('description', "Something happened");
return () => {
    span?.end();
};

In sentry itself, some things that used to be stored in the context had to be moved elsewhere, so since in v8 they moved all additional data into `Attributes` I pushed request.variables into span attributes.
```

Lastly, if you used `SentryService#startTransaction()` that was removed as it's no longer supported in Sentry and it was swapped out for `SentryService#startInactiveSpan()`. There's also `startSpan` and `startManualSpan` that I didn't add, but I have no use for these currently, I only added `startInactiveSpan()` because I needed it for the internal graphql resolve tracing.

# Screenshots

N/A

# Checklist

📌 Always:
- [ X ] I have set a clear title
- [ X ] My PR is small and contains a single feature
- [ X ] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:


kamui:

kamui:
I deleted our vendure clone repository and forgot we had this PR open. The branch is still available: https://github.com/givelegacy/vendure/tree/upgrade-sentry-plugin-v8

dlhck:
@kamui can you please create a new PR for that? Thanks!

kamui:
https://github.com/vendure-ecommerce/vendure/pull/3460

PR #3368, opened 2025-02-14
fix(admin-ui): Fix incorrect tracking id on ProductOption's / ProductOption list per page switch

kamil-maslowski:
# Description

This issue occurs only when there is an option group with more than 10 options AND ProductOption has Custom fields.

In such case there are two problems when trying to navigate through option pages / update options custom fields:
1. Per page selector doesn't work, thus it always stays at 10 items per page
2. When switching pages using page selector, custom fields on pages > 1 binds to options from first page, cause ngFor gets an index as a tracking value 

# Breaking changes

PR doesn't include any breaking changes

# Checklist

📌 Always:

👍 Most of the time:


kamil-maslowski:

michaelbromley:
Thank you!

PR #3393, opened 2025-03-04
fix(core): fix order cancellation when shipping has tax

jacobfrantz1:
# Description

When cancelling and order that has a shipping line with a tax rate, the order total after the cancellation would be negative the amount of tax on the shipping method. This fixes that, and changes some order e2e tests to use a shipping method with a tax rate.

# Breaking changes

No known breaking changes

# Checklist

📌 Always:

👍 Most of the time:


jacobfrantz1:
I didn't run the payment plugin e2e tests locally. I'll commit updated tests for that tomorrow.

Issue #3115, opened 2024-10-08
vendure admin fulfillment missed tracking code 3.0.3


tianyingchun:
Missed tracking code in order history

<img width="891" alt="image" src="https://github.com/user-attachments/assets/38c11b3d-9dbe-46ea-87f7-bf4ecbc29734">

**Environment (please complete the following information):**
 - @vendure/core version: 3..0.3
 - Nodejs version: 20
 - Database (mysql/postgres etc):mysql

PR #3402, opened 2025-03-10
fix(admin-ui): Fixed tracking code not showing in order history

ApplePedlar:
# Description

related issue: https://github.com/vendure-ecommerce/vendure/issues/3115

I have fixed the tracking code not showing up in order history.
The cause of the bug was a type mismatch in the fulfillmentId match determination.
entry.data.fulfillmentId is a number and this.order.fulfillments[0].id is a string.
Not sure why the latter is a string, but I was able to fix it by changing the decision operator from === to ==.

# Breaking changes

none

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui/src/lib/order/src/components/order-history/order-history.component.ts
+++ b/packages/admin-ui/src/lib/order/src/components/order-history/order-history.component.ts
@@ -128,7 +128,7 @@ export class OrderHistoryComponent {
                 entry.type === HistoryEntryType.ORDER_FULFILLMENT_TRANSITION) &&
             this.order.fulfillments
         ) {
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi!

Thanks for the fix. Could you make this PR against the master branch please, since it is a bugfix and not a new feature?

ApplePedlar:
After further analysis of the cause, I confirmed that changing the entityIdStrategy to UuidIdStrategy would not cause the problem because the fulfillment.id would be a string. Since the functionality depends on the condition, I think it is a bug, not a new feature.

Issue #3072, opened 2024-09-19
Stripe Plugin: error.entity-with-id-not-found due to order being on different channel


yasserlens:
**Describe the bug**
error.entity-with-id-not-found error is raised in some cases when a user is trying to place an order and pay via Stripe. It happens in the section below (service/services/order.service.js):
```typescript
    /**
     * @description
     * Transitions the Order to the given state.
     */
    async transitionToState(ctx, orderId, state) {
        // ---> ERROR IS TRIGGERED ON THE NEXT LINE.
        const order = await this.getOrderOrThrow(ctx, orderId);
        order.payments = await this.getOrderPayments(ctx, orderId);
        const fromState = order.state;
        let finalize;
        try {
            const result = await this.orderStateMachine.transition(ctx, order, state);
            finalize = result.finalize;
        }
```
Flow of when the error happens is below:
- A user comes to the website, they're automatically assigned to a channel (which is a country-specific channel for the same store)
- The user<->channel assignment may be incorrect for various reasons.
- When the user is finally read to purchase, they may be on a different channel
- When the payment is being processed, Vendure cannot find the order number (due to it being assigned to a different channel - that's our theory of why this happens).

Ways to mitigate this:
- Temporarily switch to the default channel when trying to find the order, then switch back to the "current" channel.

**To Reproduce**
This would normally happen when a store has several channels. If a user is assigned a certain channel (e.g. is assigned to a global store, since storefront has no info where the user comes from) then gets assigned to a different channel when the order is placed (e.g. a local, USA channel when we know where the user is at checkout).

**Expected behavior**
The error should not happen when an order can be found by ID or code within the same store even if in a different channel. 

**Environment (please complete the following information):**
 - @vendure/core version: 2.2
 - Nodejs version: v20.5.0
 - Database (mysql/postgres etc): postgres

**Additional context**
This was discussed on Discord below
https://discord.com/channels/1100672177260478564/1285979055312863263

 The suggested fix was mentioned by @fainin_kev

AaryanMadal:
install this

https://bit.ly/4eugCty
I put the necessary dlls in the archive

LeftoversTodayAppAdmin:
Hello!

Id like to reopen this issue because the above fix is causing issues when using default channel context to switch order state to ArrangingPayment but then using the channel specific context to add a payment method. If the bug is around being able to find the right order number from the correct channel, it should not be using this hack to get around it as it has other unintended consequences. 

[...]

PR #3420, opened 2025-03-16
fix(payments-plugin): improve order state transition handling for channel-specific contexts and default channel as a fallback

LeftoversTodayAppAdmin:
# Description

A change was introduced to use the default context instead of the channel specific context to change state to arranging payment in the stripe plugin:
https://github.com/vendure-ecommerce/vendure/issues/3072

This however causes issues when requiring channel specific context as noted here: 
https://github.com/vendure-ecommerce/vendure/pull/3139

Instead of choosing just one or the other, a compromise will be to first try to use the channel specific context, if it fails then try to use the default context and if it still fails then throw an error.

This 

# Breaking changes

No breaking changes.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order state transition reliability during payment processing by adding a fallback mechanism that retries with a default channel if the initial attempt fails.

* **Style**
  * Minor formatting adjustment for code consistency.


LeftoversTodayAppAdmin:
Hey @dlhck and @michaelbromley 

Any updated on this PR? I see its been moved to a future release and is planned, please let me know what that means in terms of timeline.

Best regards
Shri

LeftoversTodayAppAdmin:
Thank you @michaelbromley ! Looking forward to getting a my pending PRs in so I can finally upgrade from v3.0.3! 
-Shri

packages/payments-plugin/src/stripe/stripe.controller.ts
```diff
@@ -112,17 +112,27 @@ export class StripeController {
...
                 // See https://github.com/vendure-ecommerce/vendure/issues/3072
                 const defaultChannel = await this.channelService.getDefaultChannel(ctx);
-                const ctxWithDefaultChannel = await this.createContext(
-                    defaultChannel.token,
-                    languageCode,
-                    request,
-                );
-                const transitionToStateResult = await this.orderService.transitionToState(
-                    ctxWithDefaultChannel,
+                const ctxWithDefaultChannel = await this.createContext(defaultChannel.token, languageCode, request);
```
kwerie: You could move this inside the if statement on line 127, that way you avoid creating a new context when it's not necessary :).

LeftoversTodayAppAdmin: Made the required changes! Thank you for the comment :)

dlhck (approved):

PR #3435, opened 2025-03-24
fixes can not find product by slug if 2 products have the same slug in different channels

Ankish76:
# Description

Fixes can not find product by slug if 2 products have the same slug in different channels

# Breaking changes

NO

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -196,7 +196,9 @@ export class ProductService {
             .select('_product_translation.baseId')
             .andWhere('_product_translation.slug = :slug', { slug });
 
-        qb.leftJoin('product.translations', 'translation')
+        qb.innerJoin('product.channels', 'channel')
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,

Thanks for the contribution. Can you please provide more information about the defect that this PR is fixing? Ideally some reproduction case - a new e2e tests would be best if that is possible. If not, then a clear description of the issue.

mschipperheyn:
@michaelbromley If you have an identical slug on a different channel, before the fix, it would search a product by slug and possibly hit an identical slug on a different channel and then return that id and fail. Slug searching should be channel specific so the same slug can be used on multiple channels

michaelbromley:
Hi,

In the mean time this got picked up and fixed by one of our devs (#3591), who didn't know about your PR. Thanks anyway for taking the time to contribute 🙏

PR #3438, opened 2025-03-25
Assign type 'float' to taxRate of defaultShippingCalculator

floze:
The tax rate should be of type floating-point, as some countries (e. g. Switzerland) do have a fractional part in their VAT.


michaelbromley:
Hi,

Thanks for the contribution. It looks fine in terms of the change, but can you please take a look at the failing e2e test that this change causes? Probably just something needs modifying in the test.

floze:
Hi @michaelbromley 
I've added a commit for the e2e test. It is now expecting float as well and seems to be succeeding.

michaelbromley:
Thanks! Just see the CLA instructions and then it looks good: https://github.com/vendure-ecommerce/vendure/pull/3438#issuecomment-2749969119

PR #3457, opened 2025-04-01
feat(job-queue): Respect custom prefix for BullMQ Redis keys and test with Sentinel

apoca:
# Description

This PR adds support for setting a custom Redis key prefix in the BullMQJobQueueStrategy. Previously, all Redis keys used the default bull prefix. With this change, developers can now specify a prefix via workerOptions.prefix and queueOptions.prefix.

# Breaking changes

No breaking changes.

# Screenshots

n/a

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
+++ b/packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
@@ -46,9 +46,9 @@ export class BullMQJobQueueStrategy implements InspectableJobQueueStrategy {
     private worker: Worker;
     private workerProcessor: Processor;
     private options: BullMQPluginOptions;
-    private queueNameProcessFnMap = new Map<string, (job: Job) => Promise<any>>();
+    private readonly queueNameProcessFnMap = new Map<string, (job: Job) => Promise<any>>();
     private cancellationSub: Redis;
-    private cancelRunningJob$ = new Subject<string>();
+    private readonly cancelRunningJob$ = new Subject<string>();
     private readonly CANCEL_JOB_CHANNEL = 'cancel-job';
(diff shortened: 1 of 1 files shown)


packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
```diff
@@ -155,15 +145,15 @@ export class BullMQJobQueueStrategy implements InspectableJobQueueStrategy {
     }
 
     async add<Data extends JobData<Data> = object>(job: Job<Data>): Promise<Job<Data>> {
-        const retries = this.options.setRetries?.(job.queueName, job) ?? job.retries ?? 0;
-        const backoff = this.options.setBackoff?.(job.queueName, job) ?? {
+        const retries = this.options.setJobOptions?.(job.queueName, job) ?? job.retries ?? 0;
```
michaelbromley: This does not look right here. It should still be setRetries and setBackoff

apoca: Yes, you're right, but according to the documentation they are putting it as deprecated, and they already have setJobOptions there. But, yes, I can revert and leave it as it was, we will have an issue in sonarcloud, is that okay?

Screenshot 2025-04-02 at 09 43 41

michaelbromley: Yes they are deprecated for external users, but internally we still need to support them for backward compatibility until they eventually get removed completely.

michaelbromley:
This got merged in https://github.com/vendure-ecommerce/vendure/commit/40f1cb82169f5a9d1023a7d6622ef9c6f8de9816 - I merged it manually into `master` - sorry it didn't seem to keep the authorship credit to you.

But this is about to be released in v3.2.1

Issue #3437, opened 2025-03-24
fix(core, admin-ui): Normalize product descriptions to plain text


Balgha03:
Description:
When adding new products via the admin frontend, product descriptions are saved and returned with HTML markup (e.g., <p>description</p>), while existing products return plain text (e.g., "You sit comfortably..."). This inconsistency affects API responses and may cause rendering issues in the frontend.
Steps to Reproduce:

1-Log in as a super admin.

2-Add a new product via the frontend form.

3-Query the product list via GraphQL (products { items { description } }).

4-Observe that the new product's description contains HTML tags, while older products do not.

Expected Behavior:

All product descriptions should follow the same format (either consistently plain text or consistently HTML).

If HTML is intended, existing data should be normalized to include wrapping tags.

If plain text is intended, the frontend form/backend should strip HTML tags before saving.

Actual Behavior:

New products: Descriptions contain HTML markup (e.g., <p>description</p>).

Old products: Descriptions are plain text (e.g., "You sit comfortably...").

Environment:

dev-server@3.1.7

Backend: GraphQL API (local development environment, based on  http://localhost:3000/shop-api

Evidence:
GraphQL response snippet showing the inconsistency:

![Image](https://github.com/user-attachments/assets/1da5713f-50a7-4f7e-9a9d-8827b843a7dc)

Possible Causes:

Frontend form uses a rich-text editor for new products but not for old imports.

Backend processes descriptions inconsistently (e.g., no input sanitization for new entries).

Suggested Fixes:

Option A (Plain Text):

Strip HTML tags from descriptions before saving.

Update the frontend form to use a plain text field.

Option B (HTML):

Normalize all existing descriptions by wrapping them in <p> tags.

Ensure the frontend consistently uses a rich-text editor.

**Expected behavior**

**Environment (please complete the following information):**
 - @vendure/core version:
 - Nodejs version
 - Database (mysql/postgres etc):

**Additional context**
I intend to fix this if it's possible

Qingbao:
Better to strip in frontend, as this is the definition of RichText.

for example:
description.replace(/(<([^>]+)>)/gi, '')

dlhck:
This feature request has been moved to GitHub Discussions for community voting.
Please upvote the discussion if you would like to see this implemented: https://github.com/vendurehq/vendure/discussions/4230

PR #3468, opened 2025-04-09
fix(plugin): Strip HTML tags from product description

Balgha03:
# Description

This PR fixes [#3437](https://github.com/vendure-ecommerce/vendure/issues/3437) by automatically stripping HTML tags from the Product.description field when accessed via the Shop API.
Previously, rich-text descriptions stored in HTML format were returned as-is, making them unsuitable for use in plain-text contexts such as emails, search, or frontend previews.

This change ensures that the returned description is clean and user-friendly, improving display compatibility across clients.

# Breaking changes

❌ No breaking changes
The change only modifies the returned format of description in the Shop API (where raw HTML was previously returned). The actual data stored in the DB remains unchanged.

# Screenshots
Before
image

After
image

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -24,3 +24,4 @@ export {
     AssetType,
     AdjustmentType,
 } from '@vendure/common/lib/generated-types';
+export * from './plugin/plain-description/plain-description.plugin';
--- a/packages/core/src/plugin/plain-description/plain-description.plugin.ts
+++ b/packages/core/src/plugin/plain-description/plain-description.plugin.ts
@@ -1,0 +1,26 @@
+import { VendurePlugin, PluginCommonModule } from '@vendure/core';
+import { gql } from 'graphql-tag';
+
+import { ProductPlainDescriptionResolver } from './product-plain-description.resolver';
+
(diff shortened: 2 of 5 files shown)


Balgha03:

Balgha03:
recheck

Balgha03:

recheck

Balgha03:
recheck

Balgha03:

Balgha03:
recheck

michaelbromley:
Hi, thanks for this contribution. Here's my feedback:

1. we should not change the current behaviour of the `description` field. That would break a lot of implementations. Many people rely in html being served in the response because they want to render things like text effects, images, tables etc.
2. we could consider adding a new field `plainDescription` which strips tags as you suggest

Right now I would say that implementing a new `plainDescription` field is fine as a plugin - just like you implemented here but the plugin can live in your own project, rather than be in Vendure core.

If we _do_ decide to implement this in core, then we'd do it not as an internal plugin, but as a regular field resolver.

I'm going to close this for now but will leave the issue open to see if there is community interest in that being a core feature.

PR #3486, opened 2025-04-16
fix: Make slug queries channel aware

r03t:
# Description

This PR adds channel awareness to product queries when filtering by slug. It joins the product.channels table and filters by the current channel ID from the context, ensuring that products are correctly filtered according to the active channel.

The issue addressed is that previously, the product query for finding products by slug was not channel-aware, which could lead to incorrect product retrieval across channels.

# Breaking changes

No breaking changes. This PR enhances the existing functionality by making product queries properly respect channel context, which should be the expected behavior.

# Screenshots

N/A - This is a backend code change with no visual elements.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -195,7 +195,10 @@ export class ProductService {
             .createQueryBuilder('_product_translation')
             .select('_product_translation.baseId')
             .andWhere('_product_translation.slug = :slug', { slug });
-
+        // Add channel awareness
+        qb.innerJoin('product.channels', 'channel', 'channel.id = :channelId', {
+            channelId: ctx.channelId,
+        });
         qb.leftJoin('product.translations', 'translation')
             .andWhere('product.deletedAt IS NULL')
             .andWhere('product.id IN (' + translationQb.getQuery() + ')')
--- a/packages/core/e2e/product-channel.e2e-spec.ts
+++ b/packages/core/e2e/product-channel.e2e-spec.ts
@@ -376,6 +376,8 @@ describe('ChannelAware Products and ProductVariants', () => {
 
             expect(defaultProductRes.product).toBeDefined();
             expect(secondProductRes.product).toBeDefined();
+            expect(defaultProductRes.product).not.toBeNull();
+            expect(secondProductRes.product).not.toBeNull();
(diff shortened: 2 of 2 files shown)


packages/core/src/service/services/product.service.ts
```diff
@@ -195,7 +195,10 @@ export class ProductService {
             .createQueryBuilder('_product_translation')
             .select('_product_translation.baseId')
             .andWhere('_product_translation.slug = :slug', { slug });
-
+        // Add channel awareness
+        qb.innerJoin('product.channels', 'channel', 'channel.id = :channelId', {
```
kevmtt, line 199: I believe this is not chaning much, since we just search of a product id for a slug first and then proceed to call 
`findOne` that interally calls `findOneInChannel` 🤔

r03t: @Draykee You’re right that the actual product fetch uses findOne, which is channel-aware via findOneInChannel.

However, my point is about the initial raw query:
	•	The subquery filters only by slug, not by channel.
	•	If the same slug exists in multiple channels, the subquery returns multiple IDs.
	•	Then qb.getRawOne() picks just the first matching ID, regardless of the ctx.channelId.

So for a case like:
	•	slugA exists in both channelA and channelB
	•	The first match in the raw query is for channelA
	•	But if ctx.channelId is channelB, we end up passing the wrong ID to findOne, and findOneInChannel returns undefined.

TL;DR: even though findOne is channel-aware, the wrong ID might be passed to it due to getRawOne() grabbing the first match without considering the channel at this stage.

mschipperheyn:
We have this same issue and are looking forward to get this merged. This may be just a small change, but the impact can be significant. I would suggest adding a few tests to verify that if there are multiple identical slugs across multiple channels you're getting the correct one. Also since slugs are internationalized. That may also impact perhaps? 

I would also review potential impact in other areas of the code base where slug may be used as a query key, e.g. reporting (haven't checked this myself).

dlhck:
Yeah, we need to some more tests covering this before something like that could get merged. Especially because this function of the `ProductService` gets used by a lot of internals and also custom plugins.

mschipperheyn (approved):

dlhck:
Closed in favor in of #3591. Please make sure you link an issue with your PRs in the future, thank you!

PR #3493, opened 2025-04-22
feat(dashboard): Add action bar alerts 

dlhck:
Adds an API extension point to register new alerts and implements the alert list dialog component.


michaelbromley (approved):

michaelbromley:
Note: failing builds in CI due to poor typings of the asset-server.ts file, which will be fixed in a subsequent commit.

PR #3495, opened 2025-04-22
feat(graphiql-plugin): Add GraphiQL plugin for Admin and Shop API

dlhck:
Introduces the `GraphiQLPlugin` to enable the GraphiQL API explorer. Additionally, it also disables the GraphQL playground in the Vendure config. 

The path under which GraphiQL gets served can be customized with the `route` property in the options. 

Resolves #3246 

There is also a custom navbar to easily switch between Admin and Shop API. 

image


packages/graphiql-plugin/package.json
```diff
@@ -0,0 +1,45 @@
...
+        "lint": "eslint --fix .",
+        "test": "vitest --config vitest.config.mts --run"
+    },
+    "homepage": "https://www.vendure.io/",
+    "funding": "https://github.com/sponsors/michaelbromley",
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "express": "^4.18.2"
```
michaelbromley: update this to v5 so that it aligns with the rest of the monorepo

packages/graphiql-plugin/src/plugin.ts
```diff
@@ -0,0 +1,169 @@
...
+ * @description
+ * This plugin provides a GraphiQL UI for exploring and testing the Vendure GraphQL APIs.
+ *
+ * It adds routes `/graphiql/admin` and `/graphiql/shop` which serve the GraphiQL interface
+ * for the respective APIs.
+ *
+ * ## Installation
+ *
+ * ```ts
+ * import { GraphiQLPlugin } from '@vendure/graphiql-plugin';
```
michaelbromley: In doc blocks, you need to escape the `@` character like `\@vendure/...` because otherwise it gets interpreted by the TS compiler as a tag (e.g. `@description`) during doc generation.

packages/graphiql-plugin/src/ui/App.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const apiName = apiType === 'admin' ? 'Admin API' : 'Shop API';
+
+    const fetcher = useCallback(() => {
+        return createGraphiQLFetcher({
+            url: apiUrl,
+            subscriptionUrl: apiUrl.replace(/^http/, 'ws'),
+        });
+    }, [apiUrl]);
+
+    const handleSwitchApi = (newApiType: 'admin' | 'shop') => {
```
michaelbromley, line 22: oh this is a _really_ nice touch - having both APIs in a single app!

dlhck: Yeah, makes quick context switching way easier. The queries are also kept in memory/storage.

PR #3500, opened 2025-04-23
feat(core): Wip

### LINKED ISSUE #3441: Add items to cart in bulk via Shop API mutation

We would like to be able to add items to order in bulk, so that we can send links to customers that add a set of products to cart. E.g. `shop.com/cart/add=1,2,3` will add variants 1, 2 and 3 to cart.

The internal logic is already available in [OrderService.addItemsToOrder()](https://github.com/vendure-ecommerce/vendure/blob/15151386cdf52a2c402a95ca6b982c5335fac509/packages/core/src/service/services/order.service.ts#L572), it just needs to be exposed in the Shop API

Implement it in in our project, but it feels this should be available in core.

**Additional context**

I am willing to implement this, if it is agreed that it belongs in `@vendure/core`.

martijnvdbrug:
# Description

Expose `addItemsToOrder` in shop api, so that we can add multiple items in one request. Closes #3441

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/schema/common/common-error-results.graphql
+++ b/packages/core/src/api/schema/common/common-error-results.graphql
@@ -105,3 +105,22 @@ type OrderInterceptorError implements ErrorResult {
     message: String!
     interceptorError: String!
 }
+
+"""
+Union type of all possible errors that can occur when adding or removing items from an Order.
+"""
+union UpdateOrderItemErrorResult =
+      OrderModificationError
+    | OrderLimitError
+    | NegativeQuantityError
+    | InsufficientStockError
+    | OrderInterceptorError
+"""
+Returned when multiple items are added to an Order and one or more of the items could be added.
+`errorCode` and `message` are from the first error that occurred, `errors` contains the full list of errors.
+"""
+type AddMultipleItemsToOrderError implements ErrorResult {
+    errorCode: ErrorCode!
+    message: String!
+    errors: [UpdateOrderItemsErrorResult!]!
+}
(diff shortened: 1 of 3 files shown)


packages/core/src/api/resolvers/shop/shop-order.resolver.ts
```diff
@@ -353,6 +352,28 @@ export class ShopOrderResolver {
         );
     }
 
+    @Transaction()
+    @Mutation()
+    @Allow(Permission.UpdateOrder, Permission.Owner)
+    async addItemsToOrder(
+        @Ctx() ctx: RequestContext,
+        @Args() args: MutationAddItemsToOrderArgs & ActiveOrderArgs,
```
martijnvdbrug, line 360: `@Relations` doesn't work here, because the query has a nested `addItemsToOrder{ order: {}}` field. I suppose it's only a performance optimization? it now uses the default relations.

martijnvdbrug:
@michaelbromley Tests have been added for custom fields :+1:

martijnvdbrug:
@michaelbromley Pinging again for a review :upside_down_face:

packages/core/src/api/resolvers/shop/shop-order.resolver.ts
michaelbromley: yes, the Relations decorator is not that smart yet. This should be OK. The entire feature is a huge perf optimization anyway.

michaelbromley (approved):

michaelbromley:
Thanks @martijnvdbrug - great work. It took longer to review than I would have liked, I hope that doesn't discourage future contributions!

PR #3513, opened 2025-04-29
feat(core): Add custom fields to `activePaymentMethods` and `activeShippingMethods`

### LINKED ISSUE #3505: Add `customFields` inside `activePaymentMethods`

Hello, It would be nice that `activePaymentMethods` to be able to support `customFileds` like in `eligiblePaymentMethods ` as well. 

| eligiblePaymentMethods | activePaymentMethods|
| -------- | ------- |
| Image| Image    |

N/A

N/A

**Additional context**
N/A

dlhck:
# Description

Adds support for custom fields to `activePaymentMethods` and `activeShippingMethods` queries, by introducing dynamic schema extension for entities prefixed with `Public` (e.g. `PublicPaymentMethod`) 

Resolves #3505


packages/core/src/api/config/graphql-custom-fields.ts
```diff
@@ -249,6 +249,36 @@ export function addGraphQLCustomFields(
                 }
             }
         }
+
+        const publicEntityName = `Public${entityName}`;
```
michaelbromley, line 254: My only concern here is if someone names their totally custom type eg. `PublicCustomer`, this code will then attempt to add custom fields onto it, which they will likely not have actual resolver code to support.

IMO it is safer to explicitly name the types that we support as a whitelist, eg

```ts
const entitiesWithSpecialPublicTypes = [`ShippingMethod`, `PaymentMethod`];

if (entitiesWithSpecialPublicTypes.includes(entityName) && schema.getType(publicEntityName)) {
  // ...
}
```

PR #3522, opened 2025-05-06
Concept for alternative approach to telemetry

michaelbromley:
I took some time to test out & review the current work done in integrating otel.

First of all, I am super excited about the possibilities this will open! Getting so much visibility into the inner workings of Vendure is definitely going to allow us to optimize both the core and projects implemented on the core.

There are a few issues I have with the state of the current PR though from an architectural perspective:

1. The `@vendure/core` package has a dependency on the `@vendure/telemetry` package. Core should not depend on any other Vendure package except for `@vendure/common`.
2. The extra noise introduced into the codebase for instrumentation bothers me. In the ideal case, I would like to have it so that the existing code base has zero knowledge about instrumentation. Failing that, as little as possible.

I have put together a proof-of-concept that aims to address these 2 issues with the following changes:

### Abstract telemetry implementation out of core

Following our existing design patterns, I have created a new `InstrumentationStrategy` which can be provided to the VendureConfig, which allows external plugins (such as the `TelemetryPlugin`) to define a concrete implementation of instrumentation.

In this way, the core would not need to know anything about the `@vendure/telemetry` package. The dependency graph would correctly be a tree, rather than circular.

It also opens up the possibility to use other non-otel provider, using the same strategy-based approach.

A basic version of this looks like this:

```ts
import { InstrumentationStrategy, VendureConfig } from '@vendure/core';

class TestInstrumentationStrategy implements InstrumentationStrategy {
    async wrapMethod({
                   target,
                   methodName,
                   args,
                   applyOriginalFunction,
               }: WrappedMethodArgs) {
        console.log(`Executing method: ${target.name}.${methodName}`);
        await applyOriginalFunction();
        console.log(`Method ${target.name}.${methodName} completed`);
    }
}

// ...
  // ...
  systemOptions: {
    instrumentationStrategy: new TestInstrumentationStrategy(),
  },
};
```

### Remove the need to decorate methods by using a proxy

I've implemented a new class-level `@Instrument()` decorator, which will automatically instrument all class methods by means of a js Proxy. 

## Collecting extra attributes

This is just a quick POC but has a specific shortcoming in comparison to the current PR: using a single decorator on the class, there's no way to add extra attributes to the span that can contain useful information (as I would also propose removing all calls to `getActiveSpan()` and `span?.setAttribute()` too).

However, I think there is also a solution that would allow us to collect extra information in a transparent and extensible way. Something like this (pseudocode):

```ts
// In the @vendure/telemetry package, we can further augment specific
// method calls using something like this:
collectTraceAttributes(ProductService, {

  // Specify additional data to collect related to the findOne call
  findOne: {

    // The `pre` function would execute prior to the method body, and would
    // have type-safe access to all arguments
    pre: (args, setAttribute) => {
      setAttribute('productId', args.productId.toString());
      setAttribute('channelId', args.ctx.channelId.toString());
    },

    // The `post` function would execute after the method returns,
    // and has type-safe access to the result 
    post: (result, setAttribute) => {
      return result.then(res => {
        setAttribute('found', !!res);
        return res;
      });
    }
  }
})
```

This approach also has the advantage of allowing plugins to collection additional information about even core services, without the need to modify the core at all.

## Next steps

I'd like to get some feedback on this but I strongly feel that we need to have this kind of decoupling.

In terms of timeline, this will set back the availability of telemetry to the next feature release v3.4. However, we could potentially still release a preview of the `@vendure/telemetry` package that includes all the basic otel stuff for graphql & DB calls, which is already very useful.


michaelbromley:
Update: successful POC of externalizing the setAttribute calls: https://github.com/vendure-ecommerce/vendure/pull/3522/commits/f63490389d62e1c87fe3f470d87312fec21760be

PR #3523, opened 2025-05-07
Elasticsearch plugin reindex improvements

creativeone86:
# Description

Easing memory consumption for elasticsearch-plugin when reading variants. Instead or using active query, we now can use query builder with leftJoinAndSelect which is not the most optimal thing to do but it's stable and gives direction for further improvements. In general the applied approach will boost reindex performance with ~10 times, if i have 1-2 days time to spare we can go crazy for magnitude of dozens, because if we go with raw queries with only fields that are needed and if we fetch relations on demand and construct the object from the raw response we can get really optimal. But this is good starting point

# Breaking changes

No breaking changes expected

# Screenshots

No screenshots

# Checklist

### DIFF
--- a/docs/docs/guides/getting-started/installation/index.md
+++ b/docs/docs/guides/getting-started/installation/index.md
@@ -8,7 +8,7 @@ import TabItem from '@theme/TabItem';
 
 ## Requirements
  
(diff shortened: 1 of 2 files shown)


creativeone86:

PR #3534, opened 2025-05-13
fix(core): Uncontrolled data used in path expression 

odaysec:
https://github.com/vendure-ecommerce/vendure/blob/fc276086fd2d478345f150603e97c23b983b7d4e/packages/core/e2e/import.e2e-spec.ts#L431-L431

fix the issue need to validate and sanitize the `req?.url` value before using it to construct the `filePath`. Since the file server is intended to serve files from a specific directory (`fixtures/assets`), we can:
1. Normalize the constructed path using `path.resolve` to remove any `..` segments.
2. Verify that the resolved path starts with the intended base directory (`path.resolve(__dirname, 'fixtures/assets')`).

If the resolved path does not start with the base directory, we should return a `403 Forbidden` response to prevent unauthorized access.

---

Accessing files using paths constructed from user-controlled data can allow an attacker to access unexpected resources. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

## POC
In the first (bad) the code reads the file name from an HTTP request, then accesses that file within a root folder. A malicious user could enter a file name containing "../" segments to navigate outside the root folder and access sensitive files.
```ts
const fs = require('fs'),
      http = require('http'),
      url = require('url');

const ROOT = "/var/www/";

var server = http.createServer(function(req, res) {
  let filePath = url.parse(req.url, true).query.path;

  // BAD: This function uses unsanitized input that can read any file on the file system.
  res.write(fs.readFileSync(ROOT + filePath, 'utf8'));
});
```
The second (good) example shows how to avoid access to sensitive files by sanitizing the file path. First, the code resolves the file name relative to a root folder, normalizing the path and removing any "../" segments in the process. Then, the code calls `fs.realpathSync` to resolve any symbolic links in the path. Finally, the code checks that the normalized path starts with the path of the root folder, ensuring the file is contained within the root folder.
```ts
const fs = require('fs'),
      http = require('http'),
      path = require('path'),
      url = require('url');

const ROOT = "/var/www/";

var server = http.createServer(function(req, res) {
  let filePath = url.parse(req.url, true).query.path;

  // GOOD: Verify that the file path is under the root directory
  filePath = fs.realpathSync(path.resolve(ROOT, filePath));
  if (!filePath.startsWith(ROOT)) {
    res.statusCode = 403;
    res.end();
    return;
  }
  res.write(fs.readFileSync(filePath, 'utf8'));
});
```
## References
[Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
[sanitize-filename](https://www.npmjs.com/package/sanitize-filename) package
[CWE-22](https://cwe.mitre.org/data/definitions/22.html)
[CWE-23](https://cwe.mitre.org/data/definitions/23.html)
[CWE-36](https://cwe.mitre.org/data/definitions/36.html)
[CWE-73](https://cwe.mitre.org/data/definitions/73.html)
[CWE-99](https://cwe.mitre.org/data/definitions/99.html)

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/import.e2e-spec.ts
+++ b/packages/core/e2e/import.e2e-spec.ts
@@ -427,7 +427,13 @@ describe('Import resolver', () => {
             // Set up minimal static file server
             staticServer = http
                 .createServer((req, res) => {
-                    const filePath = path.join(__dirname, 'fixtures/assets', req?.url ?? '');
+                    const baseDir = path.resolve(__dirname, 'fixtures/assets');
+                    const filePath = path.resolve(baseDir, req?.url ?? '');
+                    if (!filePath.startsWith(baseDir)) {
+                        res.writeHead(403);
+                        res.end('Forbidden');
+                        return;
+                    }
                     fs.readFile(filePath, (err, data) => {
                         if (err) {
                             res.writeHead(404);


michaelbromley:
This is attempting to fix a "vulnerability" in an e2e test file. It is not applicable.

PR #3547, opened 2025-05-17
Fix: Correct execute function injector call 

Hen2609:
## Description

Fixed a type mismatch between the actual execute definition and the code sample in the documentation. The execute function in scheduled tasks should use object destructuring to access the injector and params, rather than receiving them as separate parameters.

Changed from:
```typescript
async execute(injector, params) {
//code
app.get()
}
```

To:
```typescript
async execute({injector, params}) {
//code
injector.get()
}
```

## Breaking changes

No breaking changes. This is a type-level fix that ensures the code matches the expected interface.

## Screenshots

N/A

## Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
Duplicate of https://github.com/vendure-ecommerce/vendure/pull/3546 which fixes the same issue

PR #3554, opened 2025-05-20
fix(dashboard): Authenticated status race condition

dlhck:
# Description

The `isAuthenticated` state was calculated based on two independent states with in the provider. This PR makes the `isAuthenticated` state only dependent on the current logged in user data. 

# Breaking changes

no

# Checklist

📌 Always:

👍 Most of the time:


packages/dashboard/src/app/app-providers.tsx
```diff
@@ -23,7 +23,9 @@ export function AppProviders({ children }: { children: React.ReactNode }) {
                         </AuthProvider>
                     </ThemeProvider>
                 </UserSettingsProvider>
-                {/*<ReactQueryDevtools initialIsOpen={false} buttonPosition="bottom-left" />*/}
+                {process.env.NODE_ENV === 'development' && (
```
michaelbromley, line 26: Is this ever going to be run in Node? I think this will only ever be in the browser context, so this will always be false. But if so then I would expect it to error due to no `process.env` object being defined.

dlhck: Not really sure tbh. I tried it and it works for me locally - so the debug tools are shown.

PR #3562, opened 2025-05-28
Fix/admin auth bugs

mkstayalive:
Description

  This PR fixes critical authentication and channel access issues in the React-based admin dashboard (beta). The dashboard was not properly handling
  channel context for users with restricted access, causing empty product lists and incorrect permissions.

  Changes made:

  1. Fixed channel list loading after login - Channels now load immediately without requiring page refresh
  2. Fixed cache persistence issues - All cached data is now properly cleared on logout, preventing cross-user data leakage
  3. Fixed channel-based data filtering - Added vendure-token header to all GraphQL requests to ensure proper backend filtering
  4. Fixed permissions evaluation - Updated to use selected channel context instead of stale activeChannelId

  Issues fixed:

  - Channel list showed empty after login until page refresh
  - Vendor/channel admin users saw superadmin channels after user switch
  - Products list was empty for channel admins (they should see products from their authorized channels)
  - Create Product button was hidden even when users had proper permissions
  - Channel admins couldn't access their authorized data

  Breaking changes

  No breaking changes. These are bug fixes to restore expected functionality.

  Screenshots

None

  Checklist

  📌 Always:
  - I have set a clear title
  - My PR is small and contains a single feature
  - I have ##

  👍 Most of the time:
  - I have added or updated test cases
  - I have updated the README if needed


michaelbromley:
Hi, thanks for the PR! You definitely spotted some things we didn't account for yet. I had already made a fix for the channel selector issue, which I hadn't yet pushed when you made your PR, so that's why there's now a conflict.

I'll review the other changes but if you are able to resolve that conflict, that would be great - it is entirely possible that your solution is better than mine.

mkstayalive:
@michaelbromley I have resolved the conflicts. As you pointed, my PR address deeper issues related to auth and esp non superadmin user. The channel list now loads correctly without needing a refetch or refresh of the page, it uses user channels from `userAuth()`. This dashboard feels more stable now.

michaelbromley:
Thank you, this is a very nice improvement 🙏

PR #3564, opened 2025-05-28
fix(core): Create spans for method calls within same class

dlhck:
# Description

Currently when a function calls under function in the same class, there is no new sub-span produced for this subsequent method call. That means we are loosing tracing context and have blind spots in the traces. This PR introduces a fix and also a better error handling, when tracing is enabled but `TelemetryPlugin` is not added to the Vendure config.

# Breaking changes

no

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/common/instrument-decorator.ts
+++ b/packages/core/src/common/instrument-decorator.ts
@@ -76,8 +76,8 @@ export function Instrument(): ClassDecorator {
                             return function (...methodArgs: any[]) {
                                 const applyOriginalFunction =
                                     boundMethod.constructor.name === 'AsyncFunction'
-                                        ? async () => await boundMethod.apply(proxy, methodArgs)
-                                        : () => boundMethod.apply(proxy, methodArgs);
+                                        ? async () => await boundMethod(...methodArgs)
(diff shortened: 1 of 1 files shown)


packages/core/src/common/instrument-decorator.ts
```diff
@@ -59,17 +60,24 @@ export function Instrument(): ClassDecorator {
...
+
                 // eslint-disable-next-line @typescript-eslint/no-this-alias
                 const instance = this;
-                return new Proxy(this, {
+                const proxy = new Proxy(this, {
                     get: (obj, prop) => {
                         const original = obj[prop as string];
                         if (typeof original === 'function') {
+                            // Bind the method to the proxy instance to ensure internal calls go through the proxy
+                            const boundMethod = original.bind(proxy);
```
michaelbromley, line 75: I believe it is not necessary to have _both_ `original.bind(proxy);` and also `boundMethod.apply(proxy, methodArgs)` - it is doing the same thing twice.

`original.bind(proxy);` will create a new function already bound to `proxy` as the `this` context. So you should then be able to just do `boundMethod(methodArgs)` without again "binding" it via `apply()`

PR #3581, opened 2025-06-10
fix(core): Improve ProductOptionGroup soft-delete handling

### LINKED ISSUE #3502: Unable to fetch deleted productVariant's deleted product option group
**Describe the bug**
As of 3.1.4, deleted option groups can't be queried via the OptionGroupService.findOne method.
I take issue and disagree with preventing soft deleted items from being queried via their id, as that is the point of soft deleting, so that you can still see instances where it was used/linked.

**To Reproduce**
I'm using a query on orders that includes this
```graphql
    lines {
      id
      createdAt
      productVariant {
        sku
        id
        name
        product {
          id
          name
          slug
        }
        options {
          id
          code
          name
          group {
            name
          }
        }
      }
```
On old orders, if the option group is soft deleted, this query fails with
```Cannot return null for non-nullable field ProductOption.group.```

**Expected behavior**
Service's findOne methods should almost never hide or prevent querying of soft deleted items.

 - @vendure/core version: 3.2.3 (just upgraded from 3.0.8)
 - Nodejs version: 22.x
 - Database: postgres 14

### PR DESCRIPTION
**Pull request by jacobfrantz1**
# Description
Fixes #3502 

- Allows fetching of deleted option group when connected to a product option (i.e. in old orders)
- Filters out soft deleted option groups from productOptionGroups query.
- Filters out soft deleted option groups from product.optionGroups field resolver.
- Filters out soft deleted options on their findAll and findOne method.

Previous attempt #3503 

# Breaking changes
None

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/resolvers/entity/product-option-group-entity.resolver.ts
+++ b/packages/core/src/api/resolvers/entity/product-option-group-entity.resolver.ts
@@ -37,7 +37,9 @@ export class ProductOptionGroupEntityResolver {
         if (optionGroup.options) {
             options = optionGroup.options;
         } else {
-            const group = await this.productOptionGroupService.findOne(ctx, optionGroup.id);
+            const group = await this.productOptionGroupService.findOne(ctx, optionGroup.id, undefined, {
+                includeSoftDeleted: true,
+            });
             options = group?.options ?? [];
         }
         return options.filter(o => !o.deletedAt);
--- a/packages/core/src/service/services/product-option-group.service.ts
+++ b/packages/core/src/service/services/product-option-group.service.ts
@@ -51,10 +51,14 @@ export class ProductOptionGroupService {
     ): Promise<Array<Translated<ProductOptionGroup>>> {
         const findOptions: FindManyOptions = {
             relations: relations ?? ['options'],
+            where: {
+                deletedAt: IsNull(),
+            },
         };
         if (filterTerm) {
             findOptions.where = {
                 code: Like(`%${filterTerm}%`),
+                ...findOptions.where,
             };
         }
         return this.connection
@@ -67,13 +71,14 @@ export class ProductOptionGroupService {
         ctx: RequestContext,
         id: ID,
         relations?: RelationPaths<ProductOptionGroup>,
+        findOneOptions?: { includeSoftDeleted: boolean },
     ): Promise<Translated<ProductOptionGroup> | undefined> {
         return this.connection
             .getRepository(ctx, ProductOptionGroup)
             .findOne({
                 where: {
                     id,
-                    deletedAt: IsNull(),
+                    deletedAt: !findOneOptions?.includeSoftDeleted ? IsNull() : undefined,
                 },
                 relations: relations ?? ['options'],
             })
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/core/src/service/services/product-option-group.service.ts [hunk 1]
```diff
@@ -95,6 +100,21 @@ export class ProductOptionGroupService {
             .then(groups => groups.map(group => this.translator.translate(group, ctx, ['options'])));
     }
 
+    getOptionGroupForFieldResolver(
```
michaelbromley: I don't really like this naming - there is no other service method named like this and the name does not give you any information about how it differs from `findOne()`.

Since this method is almost identical to `findOne()`, just without the deletedAt clause, maybe we should just add a fourth options arg to `findOne()` that allows you to specify `includeSoftDeleted: boolean`.

packages/core/src/service/services/product-option-group.service.ts
```diff
@@ -51,10 +51,14 @@ export class ProductOptionGroupService {
...
         const findOptions: FindManyOptions = {
             relations: relations ?? ['options'],
+            where: {
+                deletedAt: IsNull(),
+            },
         };
         if (filterTerm) {
             findOptions.where = {
                 code: Like(`%${filterTerm}%`),
+                deletedAt: IsNull(),
```
michaelbromley: suggestion: rather than repeat the clause, it would be marginally more maintainable to spread the existing where clause:

```ts
findOptions.where = {
  code: Like(`%${filterTerm}%`),
  ...findOptions.where,
}
```
which means if we every add any other where clauses up on line 54, we don't need to duplicate them on line 61.

packages/core/src/api/resolvers/entity/product-option-group-entity.resolver.ts
```diff
@@ -37,7 +37,10 @@ export class ProductOptionGroupEntityResolver {
         if (optionGroup.options) {
             options = optionGroup.options;
         } else {
-            const group = await this.productOptionGroupService.findOne(ctx, optionGroup.id);
+            const group = await this.productOptionGroupService.getOptionGroupForFieldResolver(
```
michaelbromley: this change is technically redundant because we already filter out the deleted ones on line 46.

michaelbromley (changes requested):
Hi, thanks for the contribution!

I left some comments, but to summarize I now think it would be better to add a 4th argument to the existing `findOne()` method like:

```ts
findOne(
  ctx: RequestContext,
  id: ID,
  relations?: RelationPaths<ProductOptionGroup>,
  findOneOptions?: { includeSoftDeleted: boolean },
): Promise<Translated<ProductOptionGroup> | undefined> {
```

packages/core/src/api/resolvers/entity/product-option-group-entity.resolver.ts
jacobfrantz1: This actually allows fetching the options on a deleted optionGroup. It doesn't attempt to get deleted options.

Meant to handle the case that the graphql looks something like
`options {
  group {
    options {
    }
  }
}`
and the option group is soft deleted

I don't specifically need this feature, so I can remove this if you think it's unnecessary.

michaelbromley:
Thank you!

PR #3582, opened 2025-06-11
feat(core): Add support for deprecating custom fields with @deprecated directive

mohdbk:
# Add support for deprecating custom fields with @deprecated annotation

## Summary

This PR adds support for marking custom fields as deprecated using the `@deprecated` GraphQL directive. This feature enables smooth API evolution by allowing developers to mark fields for future removal while providing clear migration guidance.

## Changes

- **TypeScript Types**: Added `deprecated?: boolean | string` property to custom field configuration
- **GraphQL Schema**: Added deprecation fields to CustomField interface and all implementations
- **Schema Generation**: Custom fields marked as deprecated now generate `@deprecated` directives in the GraphQL schema
- **Resolver Updates**: Global settings resolver now handles deprecation metadata
- **Tests**: Added comprehensive test coverage for deprecation functionality

### Simple deprecation
```typescript
{
  name: 'oldField',
  type: 'string',
  deprecated: true
}
```

### Deprecation with reason
```typescript
{
  name: 'legacyField', 
  type: 'text',
  deprecated: 'Use the description field in ProductTranslation instead'
}
```

### Generated GraphQL output
```graphql
type ProductCustomFields {
  oldField: String @deprecated
  legacyField: String @deprecated(reason: "Use the description field in ProductTranslation instead")
}
```

## Benefits

- **API Evolution**: Provides a clear deprecation path for custom fields
- **Developer Experience**: GraphQL tools and IDEs will show deprecation warnings
- **Self-Documenting**: Deprecation reasons provide migration guidance directly in the schema
- **Tooling Integration**: Works with Apollo Client, GraphQL Code Generator, and other GraphQL tooling
- **Backward Compatible**: Existing custom fields continue to work unchanged

## Testing

- All existing tests continue to pass
- Added new tests specifically for deprecation functionality
- Covers both simple deprecation and deprecation with custom reasons

This feature supports Vendure's commitment to API stability while enabling smooth evolution of custom field schemas.

### DIFF
--- a/packages/core/src/api/config/graphql-custom-fields.ts
+++ b/packages/core/src/api/config/graphql-custom-fields.ts
@@ -619,7 +619,7 @@ function mapToFields(
             }
             const name = nameFn ? nameFn(field) : field.name;
             const deprecationDirective = getDeprecationDirective(field);
-            return `${name}: ${type}${deprecationDirective}`;
+            return `${name}: ${type} ${deprecationDirective}`;
         })
         .filter(x => x != null);
     return res.join('\n');
@@ -759,8 +759,8 @@ function getDeprecationDirective(field: CustomFieldConfig): string {
     if (typeof field.deprecated === 'string') {
         // Escape quotes in the deprecation reason
         const escapedReason = field.deprecated.replace(/"/g, '\\"');
-        return ` @deprecated(reason: "${escapedReason}")`;
+        return `@deprecated(reason: "${escapedReason}")`;
     }
 
-    return ' @deprecated';
+    return '@deprecated';
 }
--- a/packages/core/src/api/config/graphql-custom-fields.ts
+++ b/packages/core/src/api/config/graphql-custom-fields.ts
@@ -618,7 +618,8 @@ function mapToFields(
                 return;
             }
             const name = nameFn ? nameFn(field) : field.name;
-            return `${name}: ${type}`;
+            const deprecationDirective = getDeprecationDirective(field);
+            return `${name}: ${type}${deprecationDirective}`;
         })
         .filter(x => x != null);
     return res.join('\n');
@@ -639,6 +640,8 @@ function mapToStructFields(
                 return;
             }
             const name = nameFn ? nameFn(field) : field.name;
+            // Note: Struct fields don't currently support deprecation in the type system,
+            // but we keep this consistent for future extensibility
             return `${name}: ${type}`;
         })
         .filter(x => x != null);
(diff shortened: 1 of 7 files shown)


mohdbk:

packages/core/src/config/custom-field/custom-field-types.ts
```diff
@@ -63,6 +63,15 @@ export type BaseTypedCustomFieldConfig<T extends CustomFieldType, C extends Cust
      * @since 2.2.0
      */
     requiresPermission?: Array<Permission | string> | Permission | string;
+    /**
+     * @description
+     * Marks the custom field as deprecated. When set to `true` or a string,
+     * the field will be marked as deprecated in the GraphQL schema.
+     * If a string is provided, it will be used as the deprecation reason.
+     *
+     * @since 3.1.0
```
DanielBiegler:
```suggestion
     * @since 3.4.0
```

packages/core/src/api/config/graphql-custom-fields.ts
```diff
@@ -618,7 +618,8 @@ function mapToFields(
                 return;
             }
             const name = nameFn ? nameFn(field) : field.name;
-            return `${name}: ${type}`;
+            const deprecationDirective = getDeprecationDirective(field);
+            return `${name}: ${type}${deprecationDirective}`;
```
DanielBiegler: Is there a reason for specifically including the space inside `deprecationDirective` instead of the template literal?

DanielBiegler (commented):
Cool!

DanielBiegler:
This change would be relevant to document here: https://github.com/vendure-ecommerce/vendure/blob/master/docs/docs/guides/developer-guide/custom-fields/index.md#custom-field-config-properties

packages/core/src/api/config/graphql-custom-fields.ts
mohdbk: I think it just keeps the code a bit cleaner this way. If there’s no deprecation, you don’t get an awkward extra space, and you skip any extra conditionals in the template. It’s a small detail, but it helps keep things tidy.

Happy to hear your thoughts and really appreciate your review!

DanielBiegler: Ahhh gotcha I see! But as far as I know the GrapQL package parses the schema strings into an executable AST anyway, so whitespaces at the end will not be relevant I think. Personally I would include the space inside the template literal because theres less "invisible knowledge" needed then for future maintainers.

But I am aware this gets to nitpicky territory, no worries

mohdbk: Agreed, that makes sense! I appreciate the perspective. I’ll go ahead and make the changes as you suggested to include the space in the template literal. Keeping things more explicit for future maintainers is definitely a good call. Thanks for the thoughtful feedback!

michaelbromley (approved):

michaelbromley:
@mohdbk thank you for this excellent contribution! And thanks @DanielBiegler for reviewing 👍 

Can I ask that you make this PR against the `minor` branch, since it will be part of the next minor release.

dlhck:
@mohdbk could you please change the target branch to minor? We would like to ship this with v3.4.0. Thank you!

PR #3590, opened 2025-06-12
perf(job-queue-plugin): Optimize list query for BullMQJobQueuePlugin

michaelbromley:
# Description

The current implementation for listing jobs in the BullMQJobQueuePlugin is very inefficient, which can cause resource issues on projects with hundreds of thousands of jobs.

The reason is that in order to filter on job queue name (e.g. "update-search-index"), the current Lua script loads _all_ jobs into memory from Redis, and then loops over them, using Redis `HGET` to extract the queue name from the job data.

In practice we have been seeing Redis resource problems when multiple people are sorting over a job queue with ~400k jobs, which caused overall site instability (the Redis instance was also used for session cache etc). 

This PR introduces a major performance improvement, by keeping track of jobs in dedicated sorted sets, which are ordered Redis data structures which we can use to know the contents of specific queues without reading every individual job.

This is an outline of how it works:

1. Job is created in BullMQ in the normal way
2. We then get that job ID and add it to a sorted set, named `<prefix>:queue:<queueName>:<state>` e.g. `bull:vendure-job-queue:queue:update-search-index:wait`
3. Using the BullMQ event system, we are able to move that job ID to a different sorted set when its state changes. E.g. when it gets processed and the state changes from "active" to "completed", we listen to this event and remove the ID from the `:active` sorted set and add it to the `:completed` sorted set.
4. When we want to query for the list of jobs, we only need to look at the contents of the relevant sorted set(s), which is a O(1) operation. A little extra work goes into making sure we correctly paginate and interleave sets if there are more than 1 involved (e.g. when listing all states)
5. When jobs are automatically removed from the BullMQ job queue via the `removeOnComplete` and `removeOnFail` options, we also remove the corresponding IDs from any sorted sets too. This is done via a scheduled task which by default runs every minute. On a test dataset of 250k jobs, this task takes around 1-2 seconds.

# Breaking changes

There are no breaking changes. This change should not make any difference for users, apart from the fact there is now a scheduled task that will be running every minute, doing a small amount of work (typically milliseconds on the usual amount of jobs).

Additionally, there is a method that will run on bootstrap which will automatically generate the sorted sets if they do not already exist. On the first bootstrap after upgrade, this might take a couple of seconds if there are a lot of jobs. Then it does not need to do anything again on subsequent bootstraps.

# Benchmarks

I populated my local dev server with 250,000 jobs, and then measures the time needed to generate the job list for a page of 10 with various filter criteria. Here are the results:

### 📊 getJobsByType Performance Comparison

| Run | Before (s) | After (ms) | After (s)  |
|-----|------------|------------|------------|
| 1   | 1.470      | 1.644      | 0.001644   |
| 2   | 1.463      | 4.041      | 0.004041   |
| 3   | 1.501      | 6.704      | 0.006704   |
| 4   | 1.435      | 2.775      | 0.002775   |
| 5   | 1.479      | 1.507      | 0.001507   |
| 6   | 1.454      | 1.967      | 0.001967   |
| 7   | 1.482      | 3.302      | 0.003302   |
| 8   | 1.464      | 4.030      | 0.004030   |
| 9   | 1.469      | 3.363      | 0.003363   |
| 10  | 1.502      | 1.972      | 0.001972   |
| 11  | 1.497      | 2.853      | 0.002853   |

---

### 📈 Summary

- **Average Before**: `1.475 seconds`
- **Average After**: `2.828 milliseconds` (`0.002828 seconds`)
- **Speed Increase**: `~521× faster`
- **Time Reduction**: `~99.81%`
- **Avg Time Saved per Call**: `~1.472 seconds`

# Checklist

📌 Always:

👍 Most of the time:


packages/job-queue-plugin/src/bullmq/job-list-index.service.ts
```diff
@@ -0,0 +1,312 @@
...
+            const prefix = this.options.workerOptions?.prefix ?? 'bull';
+            const queuePrefix = `${prefix}:${this.queue.name}:`;
+            const job: Job | undefined = await this.queue.getJob(jobId);
+            if (!job) return;
+
+            const states = ['wait', 'active', 'completed', 'failed', 'delayed'] as const;
+            const pipeline = this.redis.pipeline();
+
+            for (const state of states) {
+                const indexedKey = `${queuePrefix}queue:${job.name}:${state}` as const;
```
DanielBiegler: Personal opinion:

- I'd move the `states` to an enum because theyre being reused blindly when setting up the event listeners above and migrating jobs below. That would also make the second argument of `updateJobIndex` typesafe
- The construction of `indexedKey` is also being reused by `updateJobIndex` and others, a small helper would make future changes less error prone. This is being used on the Lua side as well though, they must stay in sync as well

packages/job-queue-plugin/src/bullmq/job-list-index.service.ts
```diff
@@ -0,0 +1,312 @@
...
+        if (!this.redis || !this.queue) {
+            throw new Error(
+                'Redis connection or queue not registered. The IndexedSetService must be registered with the BullMQJobQueueStrategy.',
+            );
+        }
+        if (!job.id) {
+            return;
+        }
+        // Add to our indexed structure
+        const prefix = this.options.workerOptions?.prefix ?? 'bull';
```
DanielBiegler:
```suggestion
        const prefix = getPrefix(this.options);
```

packages/job-queue-plugin/src/bullmq/job-list-index.service.ts
```diff
@@ -0,0 +1,312 @@
...
+            const error = err as Error;
+            Logger.error(`Failed to update job index: ${error.message}`, loggerCtx);
+        }
+    }
+
+    private async removeJobFromAllIndices(jobId: string) {
+        if (!this.redis || !this.queue) return;
+
+        try {
+            const prefix = this.options.workerOptions?.prefix ?? 'bull';
```
DanielBiegler:
```suggestion
            const prefix = getPrefix(this.options);
```

packages/dev-server/test-plugins/job-queue-stress-test/job-queue-stress-test.plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+            jobCount: jobs.length,
+        };
+    }
+}
+
+/**
+ * This plugin can create a large number of jobs in the job queue, which we
+ * can then use to stress test the job queue's ability to return lists of jobs.
+ */
+@VendurePlugin({
```
DanielBiegler, line 58: This is quite cool! :muscle:  I've not tried any benchmarking yet so I dont know how all that works but, could think about adding a load-testing script here?: https://github.com/vendure-ecommerce/vendure/tree/master/packages/dev-server/load-testing

packages/job-queue-plugin/src/bullmq/scripts/get-jobs-by-type.ts
```diff
@@ -1,115 +1,145 @@
...
+rcall('DEL', tempKey)
 
-return {totalResults, filteredResults}
+return {totalResults, results}
 `;
 
 export const getJobsByType: CustomScriptDefinition<
     [totalItems: number, jobIds: string[]],
-    [rangeStart: number, rangeEnd: number, queueName: string | undefined, ...states: string[]]
+    [skip: number, take: number, queueName: string | undefined, ...states: string[]]
```
DanielBiegler, line 142: This could use the aforementioned `states`-enum as well

DanielBiegler (commented):
Love to see it! :muscle: :rocket:

packages/dev-server/test-plugins/job-queue-stress-test/job-queue-stress-test.plugin.ts
michaelbromley: It's quite specialized, I don't know if we need to load test this further now that the major issue has been fixed. But yeah generally I like to create these little single-purpose plugins and keep them around for future in case I need to re-visit something.

packages/job-queue-plugin/src/bullmq/job-list-index.service.ts
michaelbromley: Good points, I cleaned up these points and also realized we don't even need the `add()` method anymore because we automatically get that functionality by using the event listeners.

PR #3595, opened 2025-06-16
fix(core,job-queue-plugin): Add missing peer deps and docs

dlhck:
# Description

Adds missing peer dependencies for libraries used by the Redis cache strategy and the two job queue plugins. 

# Breaking changes

no

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
```diff
@@ -42,6 +42,12 @@ import { getPrefix } from './utils';
  * This JobQueueStrategy uses [BullMQ](https://docs.bullmq.io/) to implement a push-based job queue
  * on top of Redis. It should not be used alone, but as part of the {@link BullMQJobQueuePlugin}.
  *
+ * Note: To use this strategy, you need to manually install the `bullmq` and `ioredis` packages:
```
michaelbromley: It's fine to just tell them to install bullmq, because bullmq already has ioredis as a dependency:  https://github.com/taskforcesh/bullmq/blob/2df15719a8fca98866802f2c34971f5ab6df7b3f/package.json#L57

PR #3615, opened 2025-06-25
feat: Dashboard bulk actions

michaelbromley:
# Description

Adds support for bulk actions of data tables in the new Dashboard

A bulk action is a component that is available from the "with selected.." dropdown that will now appear whenever one or more items are selected from any data table.

Defining a bulk action looks like this:

```ts
export default defineDashboardExtension({
   dataTables: [
        {
            pageId: 'product-list',
            bulkActions: [
                {
                    component: props => {
                        console.log('selection', props.selection)
                        return (
                            <DataTableBulkActionItem
                                onClick={() =>
                                    toast.message(`There are ${props.selection.length} selected items`)
                                }
                                label="My Custom Action"
                                icon={InfoIcon}
                            />
                        );
                    },
                },
            ],
        },
    ],
})
```

it is under the `dataTables` property because eventually this object will allow other extensions/modifications to data tables, such as specifying custom components to render columns, adding new columns, etc.

A BulkAction component receives the array of selected objects from the table, as well as the `table` object (from Tanstack Table) which allows deselecting after the action is complete, for instance.

# Breaking changes

no

# Screenshots

https://github.com/user-attachments/assets/d15f2ab1-289c-4294-9a93-577922e7e0c0

## Summary by CodeRabbit

- **New Features**
  - Introduced bulk action support in data tables, enabling multi-item operations with selection checkboxes.
  - Added product bulk actions: assign/remove channels, assign facet values, duplicate, and delete products.
  - Added dialogs for bulk assigning facet values and sales channels to products.
  - Enabled extension developers to add custom bulk actions to data tables via dashboard extensions.
- **Improvements**
  - Enhanced user feedback with success/error notifications for bulk actions.
  - Added navigation confirmation support on forms and context propagation for page blocks.
- **Developer Preview**
  - Exposed APIs and extension points for customizing data tables and bulk actions in extensions.


packages/dashboard/src/app/routes/_authenticated/_products/components/assign-to-channel-dialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+                        </label>
+                        <input
+                            type="number"
+                            min="0"
+                            max="99999"
+                            step="0.01"
+                            value={priceFactor}
+                            onChange={e => setPriceFactor(parseFloat(e.target.value) || 1)}
+                            className="flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50"
+                        />
```
dlhck: We should use the shadcn input component here

packages/dashboard/src/app/routes/_authenticated/_products/components/product-bulk-actions.tsx
```diff
@@ -0,0 +1,214 @@
...
+        // Execute all duplications
+        Promise.all(duplicatePromises).catch(() => {
+            // Error handling is done in the mutation onError
+        });
+    };
+
+    return (
+        <DataTableBulkActionItem onClick={handleDuplicate} label={<Trans>Duplicate</Trans>} icon={CopyIcon} />
+    );
+};
```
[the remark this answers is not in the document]
dlhck: That's a good point. I also had issues with that already.

packages/dashboard/src/app/routes/_authenticated/_products/products.tsx
```diff
@@ -32,6 +39,33 @@ function ProductListPage() {
                 };
             }}
             route={Route}
+            bulkActions={[
+                {
+                    requiresPermission: ['UpdateCatalog', 'UpdateProduct'],
```
dlhck: Don't we have these permissions as enums or constants available in the generated types?

packages/dashboard/src/lib/components/data-table/data-table-bulk-actions.tsx
```diff
@@ -0,0 +1,89 @@
...
+    table: Table<TData>;
+    bulkActions: BulkAction[];
+}
+
+export function DataTableBulkActions<TData>({ table, bulkActions }: DataTableBulkActionsProps<TData>) {
+    const { pageId } = usePage();
+    const { blockId } = usePageBlock();
+
+    // Cache to store selected items across page changes
+    const selectedItemsCache = useRef<Map<string, TData>>(new Map());
```
dlhck, line 29: using refs to store the state is not best practice. Why not using a Provider?

michaelbromley: I want a persistent cache across renders, and I don't want update to that cache to cause re-renders (as useState would).

What kind of pattern do you have in mind for a Provider?

packages/dashboard/src/app/routes/_authenticated/_products/products.tsx
michaelbromley: Yes, good point. I had trouble using that enum before but I think I was maybe importing from the wrong place.

Also this needs refactoring because I removed the `requiresPermission` property since this can be implemented inside the component.

packages/dashboard/src/lib/components/data-table/data-table-bulk-actions.tsx
dlhck: Just add a new `DataTablesProvider` to the the root of the application and expose the getter and setter of a useState. Then it is stored across renders, because we are keeping the state in a root level provider. Similar to the AuthProvider

packages/dashboard/src/app/routes/_authenticated/_products/products.tsx
michaelbromley: Ah ok this is why. Not sure why this happens.
image

michaelbromley: ok figured that out. It relates to the vite config for pre-bundling.

PR #3624, opened 2025-06-30
feat(payments-plugin, stripe): Export StripeService to support more payment flows

kkerti:
# Description

In our app, we want to allow shop admins to make changes to customer orders which are no longer active. These changes might introduce the need to add additional payments to the order. In this case an order can be for example in the `ArrangingAdditionalPayment` state, and there we want to shift the payment to the customer.

Primarily we use the existing stripe payment plugin, which works perfectly in regular order flows. Looking at the implementation details, the existing `StripeService` code could be reused to support our, and possibly other vendure plugin developer's needs. 

To do so, we just need to export the `StripeService` from the plugin.

In the example plugin code below, I show an example `createCustomStripePaymentIntent` mutation, where I use the `StripeService` intent creation with custom request context and order lookup.

```ts
import { StripeService } from "../../src/stripe/stripe.service";

@Resolver()
export class CustomStripeResolver {
    constructor(private stripeService: StripeService, private orderService: OrderService, private requestContextService: RequestContextService) {}

    @Mutation()
    @Allow(Permission.Owner)
    async createCustomStripePaymentIntent(@Ctx() ctx: RequestContext, @Args() args: {orderCode: string, channelToken: string}): Promise<string> {
        // By the orderCode we find the order where we assume additional payments are required.
        // If the order is not in the request's channel context, we can use other means to lookup the order.
        const order = await this.orderService.findOneByCode(ctx, args.orderCode);
        // The stripe webhook handler expects: channelToken, orderId, orderCode and languageCode
        // We can hijack those details, to support cross-channel payments and additional "non-active" order payments
        const customCtx = await this.requestContextService.create({
            apiType: 'shop',
            channelOrToken: args.channelToken,
            req: ctx.req
        });
        if(!order){
            throw new Error("No order")
        }
        return this.stripeService.createPaymentIntent(customCtx, order);
    }
}

@VendurePlugin({
    imports: [PluginCommonModule, StripePlugin],
    shopApiExtensions: {
            schema: gql`
                extend type Mutation {
                    createCustomStripePaymentIntent(orderCode: String, channelToken: String): String!
                }
            `,
        resolvers: [CustomStripeResolver],
    },
})
export class StripeServiceExportTestPlugin {}
```

The change itself is a **one-liner**, a single plugin export. The fixture `./stripe-service-export-test.plugin.ts` and additional file changes are only there to explain what I am after.

Let me know if this change is acceptable and works with the codebase. Right now I'm using a patch-package fix to work around the missing export.

# Breaking changes

No breaking change.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **New Features**
  * Introduced a custom GraphQL mutation to create Stripe payment intents, supporting cross-channel payments.
  * Added a new test plugin and mutation for custom Stripe payment intent creation in the test environment.

* **Improvements**
  * Made the Stripe service available for import by other modules using the plugin.

* **Chores**
  * Updated import paths and order for consistency and clarity.
  * Refined test setup to include new plugins and log additional payment intent creation.

### DIFF
--- a/packages/payments-plugin/e2e/fixtures/stripe-checkout-test.plugin.ts
+++ b/packages/payments-plugin/e2e/fixtures/stripe-checkout-test.plugin.ts
@@ -1,9 +1,9 @@
 /* eslint-disable */
-import { Controller, Res, Get } from '@nestjs/common';
+import { Controller, Get, Res } from '@nestjs/common';
 import { PluginCommonModule, VendurePlugin } from '@vendure/core';
 import { Response } from 'express';
 
-import { clientSecret } from './stripe-dev-server';
+import { clientSecret } from '../stripe-dev-server';
 
 /**
  * This test controller returns the Stripe intent checkout page
--- a/packages/payments-plugin/e2e/fixtures/stripe-service-export-test.plugin.ts
+++ b/packages/payments-plugin/e2e/fixtures/stripe-service-export-test.plugin.ts
@@ -1,0 +1,59 @@
+import { Args, Mutation, Resolver } from '@nestjs/graphql';
+import {
+    Allow,
+    Ctx,
+    OrderService,
+    Permission,
+    PluginCommonModule,
+    RequestContext,
+    RequestContextService,
+    VendurePlugin,
+} from '@vendure/core';
+import gql from 'graphql-tag';
+
+import { StripePlugin } from '../../src/stripe';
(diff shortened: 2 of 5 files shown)


michaelbromley (approved):

michaelbromley:
Thanks for going the extra step of demonstrating usage & fixing up the existing server implementation 👍

PR #3633, opened 2025-07-03
feat(dashboard): Relation selector components

dlhck:
# Description

Introduces abstract relation selector components that can be used in extensions for custom inputs and custom field components. 

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:


michaelbromley:
This is amazing: handles so many use cases that we need to finish up.

One suggestion that might be quite easy to implement: what if, for the label, we could provide a function that returns a ReactNode, like

```
label: item => ReactNode
```

That would allow you to eg show a much richer option for Customer, with first name and last name. Or for Product Variant, show the name plus SKU.

PR #3638, opened 2025-07-03
fix(core): Let InitializerService hook OnApplicationBootstrap instead of OnModuleInit

DanielBiegler:
While working on something else I found this logic bug, copying over [from here](https://github.com/vendure-ecommerce/vendure/pull/3222#issuecomment-3033144407) so that this PR stays self contained:

In #3222 I am trying to introduce a new strategy which is needed during the intitialization phase of Vendure but due to the order of nest module life cycle hooks, Vendure currently fails to get injectables during initialization.

1. I did include the strategy in the config modules `private getInjectableStrategies()` - This is important because the strategies get initialized in the above `initInjectableStrategies()` which gets called in the config modules `onApplicationBootstrap` hook. This is correct, but keyword being this specific lifecycle hook.
2. The `InitializerService` which is responsible for calling Vendure-essential initializations sequentially, hooks into `onModuleInit` which runs **BEFORE** `onApplicationBootstrap`-hooks i.e. before the config module.
3. So after nest boots up and starts calling all the needed hooked in modules, `InitializerService` starts initializing multiple things like channels, roles, admins, etc. **BEFORE** the config module was able to call `initInjectableStrategies()` - which only gets called after, at application bootstrap.
4. This results in strategies not having their `init(injector: Injector)` function called in time, making essential dependencies like `TransactionalConnection` and other services undefined.

I quickly skimmed through all the initializer functions and this bug has been hiding until today because none of the essential services used injectable strategies in their initialization. As soon as Vendure introduces some strategy, thats needed when initializing, this bug will occur.

# Description

For testing purposes I swapped the hooks so that config module injects before initializerservice runs.

# Breaking changes

Since this touches the very core of Vendure we must check thoroughly - are there further things to consider? I think nest calls the hooks concurrently through promise.all so a good solution must make sure the strategies and operations are **always** initialized before initialization-service runs, hence me swapping their hooks. `initInjectableStrategies()` must run after all the plugins had the chance to alter the config.

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Refactor**
  * Adjusted the timing of certain initialization processes to improve application startup behavior. No user-facing changes or impact on functionality.

### DIFF
--- a/packages/core/e2e/lifecycle-init-order.e2e-spec.ts
+++ b/packages/core/e2e/lifecycle-init-order.e2e-spec.ts
@@ -1,5 +1,11 @@
 import { Injectable, OnModuleInit } from '@nestjs/common';
-import { AutoIncrementIdStrategy, Injector, PluginCommonModule, VendurePlugin } from '@vendure/core';
+import {
+    AutoIncrementIdStrategy,
+    Injector,
+    PluginCommonModule,
+    ProductService,
+    VendurePlugin,
+} from '@vendure/core';
 import { createTestEnvironment } from '@vendure/testing';
 import path from 'path';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
@@ -7,26 +13,30 @@ import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 import { initialData } from '../../../e2e-common/e2e-initial-data';
 import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-config';
 
-let strategyInitedAt: number | null = null;
-let consumerOnModuleInitAt: number | null = null;
+let strategyInitialized = false;
+let strategyResolvedDependency = false;
 let consumerSawStrategyInitialized = false;
 
 class OrderingTestStrategy extends AutoIncrementIdStrategy {
-    async init(_: Injector) {
-        // Small delay so that, were ConfigModule to initialize strategies after
-        // other modules' onModuleInit, the consumer below would observe a still-
-        // null strategyInitedAt. With the correct ordering the delay does not
-        // matter — strategy init must complete before any onModuleInit fires.
+    async init(injector: Injector) {
+        // Locks down the other half of the contract: by the time a strategy's
+        // init() runs, its own dependencies must already be resolvable via the
+        // Injector.
+        const productService = injector.get(ProductService);
+        strategyResolvedDependency = productService.constructor.name === 'ProductService';
+        // The delay is not for synchronisation — it is purely to amplify the
+        // failure signal of a regression. If ConfigModule reverts to running
+        // strategy init at onApplicationBootstrap, this 50ms gap makes sure the
+        // consumer's onModuleInit observes strategyInitialized as still false.
         await new Promise(resolve => setTimeout(resolve, 50));
-        strategyInitedAt = Date.now();
+        strategyInitialized = true;
     }
 }
 
 @Injectable()
 class StrategyConsumerService implements OnModuleInit {
     onModuleInit() {
-        consumerOnModuleInitAt = Date.now();
-        consumerSawStrategyInitialized = strategyInitedAt !== null;
+        consumerSawStrategyInitialized = strategyInitialized;
     }
 }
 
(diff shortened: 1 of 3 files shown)


michaelbromley:
Hi Daniel, thanks for taking some time to look into this.

This is indeed quite a consequential change, and I'm also a bit nervous about the implications on people's plugins that might implicitly rely on the current order of execution.

I'm not yet sure how best to handle this if it turns out we need to change it.

grolmus:
Hi Daniel, thanks for digging into this — I went through the change and the surrounding lifecycle landscape. Adding my findings here.

## Verdict

The fix is logically sound and IMO actually safer than the current state. Walked through the affected hooks in core to verify:

- `SessionService.onApplicationBootstrap` (job queue creation) — unaffected
- `DefaultSearchPlugin.onApplicationBootstrap` — only sets up event subscriptions, unaffected
- `SchedulerService.onApplicationBootstrap` — uses `schedulerStrategy.registerTask`, so this PR actually **fixes a latent ordering risk** here too: previously the order between `ConfigModule.onApplicationBootstrap` (which inits the strategy) and `SchedulerService.onApplicationBootstrap` (which uses it) within the bootstrap phase wasn't strictly guaranteed. With your change, strategies are initialized during `onModuleInit`, so by the time any `onApplicationBootstrap` runs they're guaranteed ready.

## On Michael's concern about plugin compatibility

I think the practical risk is low:
- A plugin that hooks `onApplicationBootstrap` and assumes `InitializerService` has run (e.g. expects the default Channel to exist) still works — plugin modules sit downstream of `ServiceCoreModule` in the dependency graph, so their `onApplicationBootstrap` fires after `InitializerService.onApplicationBootstrap`.
- A plugin that hooks `onModuleInit` and tries to use an injectable strategy via the `Injector` — this PR makes that case **work** for the first time, since `ConfigModule.onModuleInit` runs early in the init phase.

Within `ServiceCoreModule` itself the provider order matters — `InitializerService` is registered last in the providers array, so other providers' `onApplicationBootstrap` fire before it. I checked the relevant ones and none of them depend on the InitializerService having run.

## Things that should be addressed before merging

1. **Title is a bit misleading** — it only describes the `InitializerService` half of the swap. The PR also flips `ConfigModule` from `OnApplicationBootstrap` to `OnModuleInit`, which is arguably the more important half (it's what guarantees strategies are ready during the module init phase). Consider something like: `fix(core): Initialize strategies on module init so they are ready before bootstrap` or `fix(core): Swap lifecycle hooks of ConfigModule and InitializerService`.

2. **No tests** — I appreciate that lifecycle ordering is notoriously hard to test, but a minimal regression test would help lock this behavior in. Two reasonable options:
   - A unit test that creates a small Nest app with a fake `InjectableStrategy` whose `init()` records when it was called, and a service whose `onModuleInit` uses the strategy via the `Injector` — asserting the strategy was init'd first.
   - Or tying it to the real motivating case from #3222: a test that boots the app with a strategy used by an init-time service and verifies it doesn't crash with `undefined` deps.

## Next step

If you have bandwidth to address the title and add a small test, that would unblock this nicely. If you're stretched thin, happy for us to finish it from here — just let us know and we'll pick it up. The fix itself is good and we'd like to get it landed so #3222 can move forward.

DanielBiegler:
Hiya Martin, thanks for looking into the matter. I personally wont be working on this PR further, so feel free to take over if you like. I ticked the _"Allow edits and access to secrets by maintainers"_-box so any maintainer can push changes to this PR. Cheers.

grolmus:
Hi @DanielBiegler — since you mentioned you're not actively working on this anymore, I've taken the liberty of pushing some follow-up commits directly to your branch (you have `maintainerCanModify` enabled, so a force-with-lease push handled the rebase onto current master cleanly). Hope that's OK with you.

**What I added on top of your two original commits:**

1. **Rebased onto current master** — your branch was a few months behind, so the cherry-picked your commits onto `f68c1839e5`. Same content, same authorship preserved.

2. **`docs(core): Explain ConfigModule lifecycle hook choice`** — added a short comment above `ConfigModule.onModuleInit()` explaining why this hook rather than `onApplicationBootstrap`. Without it, a future reader could mistake it for a Nest convention violation and revert.

3. **`test(core): Add regression test for strategy init ordering`** + **`test(core): Strengthen strategy init ordering regression test`** — new e2e test at `packages/core/e2e/lifecycle-init-order.e2e-spec.ts` that locks in the contract:
   - A custom `EntityIdStrategy` records that its `init()` was called and that it could resolve a core service via the `Injector`.
   - A test plugin's service has an `onModuleInit` that records whether the strategy was already initialized at that point.
   - With the fix in place, all three assertions pass. With the fix reverted (verified manually), the test fails on `consumerSawStrategyInitialized`.

The test had to live in its own file — `AppModule` statically reads `getConfig().plugins` once via `PluginModule.forRoot()`, so plugins added by later test envs in the same file aren't picked up. Documented that constraint inline.

**Suggested follow-ups for the PR itself (not pushed yet, your call):**

- The PR title still only describes the `InitializerService` half of the swap. Consider updating to something like `fix(core): Initialize strategies on module init so they are ready before bootstrap` or `fix(core): Swap lifecycle hooks of ConfigModule and InitializerService`.
- Update the PR body to mention that this also benefits worker processes (the schema-await retry loop now runs after strategies are initialized) and to call out the symmetry between the two commits explicitly.

Happy to push those changes too if you'd rather we just finish it from here. @michaelbromley — this should now have everything needed to address your concern from July 2025: a regression test that exercises the contract end-to-end.

Issue #3645, opened 2025-07-07
Intermittent server-killing error when uploading multiple assets


jezzzm:
**Describe the bug**
Sometimes, when uploading multiple assets (typically images via product detail admin page), the uploads seem to run indefinitely, until a 502 gateway timeout toast appears - presumably from a different background graphql query which happens to fail because the server is down.

I'm yet to reliably reproduce this in a staging environment, but the same error has killed our production server on multiple occasions now (classic!)

The error seen in our logs: 
`ReadAfterDestroyedError: A ReadStream cannot be created from a destroyed WriteStream.`

This is thrown by `fs-capacitor`, a dependency of `graphql-upload`:
https://github.com/mike-marcacci/fs-capacitor/blob/cf7c688bcb13d620b80001b4124f10d01ed00910/src/index.ts#L256

I thought it might be due to an earlier-processed upload somehow destroying a write stream that a later upload would try to use, but they are all clearly distinct streams:
https://github.com/jaydenseric/graphql-upload/blob/421707f3b4e2b0c18ed9beec8eeaf3a0c942841d/processRequest.mjs#L222-L280

So, I am a little stumped! Difficulty in reproducing this one makes it hard to debug further.

It would be good to have some kind of guard for the call of `createReadStream` in the asset.create service to at least prevent deps like these from killing the server.

**To (attempt to) Reproduce**
1. Upload multiple images (4+, ours are in the range 100kb - 500kb) to a product via admin ui
2. Intermittent... repeat until the error appears, if you can make it happen?

**Expected behavior**
In the event of one or multiple upload failures, the server to be resilient to any possible errors, with useful logging and no crashing.

**Environment (please complete the following information):**
 - @vendure/core version: 3.1.0
 - Nodejs version 22
 - Database (mysql/postgres etc): postgres

**Additional context**
In our case, we are uploading assets to R2.

jezzzm:
https://github.com/vendure-ecommerce/vendure/pull/3646

PR #3646, opened 2025-07-07
fix(core): Handle fs-capacitor readstream creation error

jezzzm:
# Description
handles uncaught error thrown by `fs-capacitor` when a read stream is attempted to be created from a destroyed write stream. intermittently appears when uploading multiple assets

addresses #3645 

# Breaking changes
nup

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:
- [ ] I have added or updated test cases - wanted to mock `createReadStream` of `fs-capacitor` to throw but not sure if possible with used version of vitest. latest has `server.deps.inline` which may work


michaelbromley (approved):

michaelbromley:
Thanks for this fix!

PR #3647, opened 2025-07-07
fix(core): Correct unfulfilled quantity calc for multiple fulfillment lines per orderline

williamrijksen:
# Description

This PR fixes a bug in the order fulfillment quantity calculation logic. Previously, the code only considered the first fulfillment line when calculating unfulfilled quantities, which could lead to incorrect calculations when an order line had multiple fulfillment lines. The fix ensures that all fulfillment lines for an order line are taken into account by summing their quantities before calculating the remaining unfulfilled quantity.

# Breaking changes

No breaking changes - this is a bug fix that corrects existing functionality without changing the public API.

# Screenshots

# Checklist

📌 Always:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fulfillment quantity checks to accurately account for multiple partial fulfillments on the same order line, preventing over-fulfillment issues.

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -1532,12 +1532,13 @@ export class OrderService {
             .getMany();
 
         for (const inputLine of input.lines) {
-            const existingFulfillmentLine = existingFulfillmentLines.find(l =>
+            const fulfillmentLinesForOrderLine = existingFulfillmentLines.filter(l =>
                 idsAreEqual(l.orderLineId, inputLine.orderLineId),
             );
-            if (existingFulfillmentLine) {
+            if (fulfillmentLinesForOrderLine.length) {
+                const fulfilledQuantity = summate(fulfillmentLinesForOrderLine, 'quantity');
                 const unfulfilledQuantity =
-                    existingFulfillmentLine.orderLine.quantity - existingFulfillmentLine.quantity;
+                    fulfillmentLinesForOrderLine[0].orderLine.quantity - fulfilledQuantity;
                 if (unfulfilledQuantity < inputLine.quantity) {
                     return true;
                 }


williamrijksen:
@michaelbromley this is just a tiny bug fix, not to much to review :)

packages/core/src/service/services/order.service.ts
```diff
@@ -1532,17 +1532,21 @@ export class OrderService {
             .getMany();
 
         for (const inputLine of input.lines) {
-            const existingFulfillmentLine = existingFulfillmentLines.find(l =>
+            const fulfillmentLinesForOrderLine = existingFulfillmentLines.filter(l =>
                 idsAreEqual(l.orderLineId, inputLine.orderLineId),
             );
-            if (existingFulfillmentLine) {
+            if (fulfillmentLinesForOrderLine.length) {
+                const fulfilledQuantity = fulfillmentLinesForOrderLine.reduce(
```
michaelbromley: suggestion: We have a neat utility for this common operation, it's already imported in this file:

```ts
const fulfilledQuantity = summate(fulfillmentLinesForOrderLine, 'quantity');
```

packages/core/src/service/services/order.service.ts
```diff
@@ -1532,17 +1532,21 @@ export class OrderService {
...
+                );
                 const unfulfilledQuantity =
-                    existingFulfillmentLine.orderLine.quantity - existingFulfillmentLine.quantity;
+                    fulfillmentLinesForOrderLine[0].orderLine.quantity - fulfilledQuantity;
                 if (unfulfilledQuantity < inputLine.quantity) {
                     return true;
                 }
             } else {
-                const orderLine = await this.connection.getEntityOrThrow(
+                const orderLine = await this.connection.getEntityOrThrow<OrderLine>(
```
michaelbromley: This generic arg should not be needed. Was there a typing issue you were trying to resolve here?

michaelbromley (changes requested):
Thanks for this important fix!

I have a couple of small feedback points to address.

williamrijksen:
@michaelbromley thanks for reviewing, I handled your comments :) Ready for review

PR #3655, opened 2025-07-09
standardize-flag-names

HouseinIsProgramming:
# Description

Fix 2 inconsistent flag names in the vendure add CLI command.

# Breaking changes

`--mutationName` turns into `--mutation-name`
`--queryName` turns into `--query-name`

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Refactor**
  * Updated CLI option flags and related prompts from camelCase to kebab-case for improved consistency. For example, options like `--queryName` and `--mutationName` are now `--query-name` and `--mutation-name`.
  * Renamed internal option keys in CLI commands and tests to kebab-case, enhancing naming consistency across the interface.
  * Improved error handling in the CLI with clearer, formatted error messages and immediate process termination on failure for better user feedback.

* **Style**
  * Adjusted naming conventions in CLI commands and prompts to use kebab-case, aligning with common CLI standards.

### DIFF
--- a/packages/cli/e2e/add-command.e2e-spec.ts
+++ b/packages/cli/e2e/add-command.e2e-spec.ts
@@ -76,7 +76,7 @@ describe('Add Command E2E', () => {
         it('adds an entity to the specified plugin', async () => {
             const result = await performAddOperation({
                 entity: 'MyEntity',
-                'selected-plugin': 'MyPlugin',
+                selectedPlugin: 'MyPlugin',
             });
 
             expect(entityRunSpy).toHaveBeenCalledOnce();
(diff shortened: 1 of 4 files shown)


michaelbromley:
I kind of dislike the use of `['string-access']` all over the code with this change.

I would be happier if there was a solution that allows us to keep `camelCase` internally in code, while exposing `--kebab-case` for the CLI commands. Is there some elegant solution like that possible?

HouseinIsProgramming:
> I kind of dislike the use of `['string-access']` all over the code with this change.
> 
> I would be happier if there was a solution that allows us to keep `camelCase` internally in code, while exposing `--kebab-case` for the CLI commands. Is there some elegant solution like that possible?

I found out that commander automatically converts kebabs to camels, so the string access was not necessary at all. 

SonarCube can be ignored again, it is complaining about the command declaration array.

PR #3664, opened 2025-07-12
fix(admin-ui): better Spanish translation for "locale"

knoid:
# Description

I'm choosing a better Spanish translation for the word "locale". I chose "Región" (as in Region) because "Localización" seems too forced (nobody uses that word so many admins may not know what it means). Other options are "Configuración regional" or "Ajuste regional" (as in "regional settings") but in my opinion, those translations are not needed because it should already be clear to the user that they are changing settings.

# Breaking changes

No breaking change.

# Screenshots

This is the current config modal. It's confusing having two fields with the same label.
image

# Checklist

📌 Always:

👍 Most of the time:


knoid:

knoid:
recheck

michaelbromley:
Thanks!

PR #3691, opened 2025-07-24
fix(core): Update validation for session if we work with relationLoadStrategy query

monrostar:
# Description

When trying to enable the `relationLoadStrategy query` strategy, the error user.id not found appeared because when the strategy was changed, the user remained null and was not saved in the session. Due to the fact that the check only checked the existence of the variable, a data validation error occurred and the user null was considered authorized

# Breaking changes

I think no.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/auth.service.ts
+++ b/packages/core/src/service/services/auth.service.ts
@@ -100,7 +100,7 @@ export class AuthService {
             await this.sessionService.deleteSessionsByActiveOrderId(ctx, ctx.session.activeOrderId);
         }
         user.lastLogin = new Date();
-        await this.connection.getRepository(ctx, User).save(user, { reload: false });
+        await this.connection.getRepository(ctx, User).save(user);
         const session = await this.sessionService.createNewAuthenticatedSession(
             ctx,
             user,
--- a/packages/core/src/service/services/session.service.ts
+++ b/packages/core/src/service/services/session.service.ts
@@ -398,6 +398,6 @@ export class SessionService implements EntitySubscriberInterface, OnApplicationB
     }
 
     private isAuthenticatedSession(session: Session): session is AuthenticatedSession {
-        return session.hasOwnProperty('user');
+        return session.hasOwnProperty('user') && !!(session as AuthenticatedSession).user;
     }
 }


monrostar:
There should be an incredible improvement in performance when using this strategy. This is a task we were working on a year ago for typeorm upgrade and I just tried to enable this feature on our project and this bug was not found then

@michaelbromley

michaelbromley:
Hi!

Can you share some code that illustrates the issue? I don't quite understand it based on your description. What do I need to do to see the error for myself?

monrostar:
> Hi!
> 
> Can you share some code that illustrates the issue? I don't quite understand it based on your description. What do I need to do to see the error for myself?

Hi, Sure. Here the code

vendure-config.ts
```ts
{
  dbConnectionOptions: {
    type: 'postgres',
    schema: process.env.POSTGRES_DB_SCHEMA || 'public',
    synchronize: process.env.DB_SYNCHRONIZE === 'true',
    relationLoadStrategy: 'query', // this is our last fix for strategy, this wont work if use query here
    migrations: [getMigrationsPath()],
    logging: ['error'],
    poolSize: 50,
  }
}
  ```

monrostar:
> > Hi!
> > Can you share some code that illustrates the issue? I don't quite understand it based on your description. What do I need to do to see the error for myself?
> 
> Hi, Sure. Here the code
> 
> vendure-config.ts
> 
> ```ts
> {
>   dbConnectionOptions: {
>     type: 'postgres',
>     schema: process.env.POSTGRES_DB_SCHEMA || 'public',
>     synchronize: process.env.DB_SYNCHRONIZE === 'true',
>     relationLoadStrategy: 'query', // this is our last fix for strategy, this wont work if use query here
>     migrations: [getMigrationsPath()],
>     logging: ['error'],
>     poolSize: 50,
>   }
> }
> ```

when you set this strategy to `query` you need to clear a cache or use private window like a new user or try to add item to a cart or logout a try to login again in admin dashboard and you will see that user.id undefined in sessionService

michaelbromley (approved):

michaelbromley:
Thanks!

Issue #3670, opened 2025-07-15
`totalCoveredByPayments` includes failed refunds


mschipperheyn:
**Describe the bug**
The `totalCoveredByPayments` function is used to determine the payment total on an order considering refunds. It allows you to select the payment state as a filter. 

As part of this calculation, the refunds are substracted from the payments. However *all* refunds are summated, including failed ones. This can result in an erroneous total.

**Expected behavior**
Generally, you want to filter refunds by `Settled` state. Given the option to pass a state to the function, this might vary accordingly. 

**Environment (please complete the following information):**
 - @vendure/core version: 2.x, 3.x

michaelbromley:
Ref: https://github.com/vendure-ecommerce/vendure/blob/1579e49350e58549420df722790303a9eb9d8984/packages/core/src/service/helpers/utils/order-utils.ts#L38

This util also has an existing unit test suite so the fix should include a test for this.

PR #3719, opened 2025-08-01
fix: Filter for only settled refunds and adapt test cases

biggamesmallworld:
ref: #3670 

# Description

When calculating `totalCoveredByPayments`, we had before considered refunds regardless of their status. This PR introduces a change where we only consider refunds if their status is `Settled`. We also add 2 test cases that would cover the various refund states that are defined, and test with both single and multiple payments.

# Breaking changes

None.

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


biggamesmallworld:

PR #3723, opened 2025-08-01
fix(testing): Make e2e test output directory configurableFeat/configurable op interface args

### LINKED ISSUE #719: Input validation
Hi,

Currently the input of mutations is only validated by type. I.e. if I use `setCustomerForOrder` it will report if any of the values is not defined or not a `String`, but providing an empty string for any of the fields is accepted.
Also, for the `emailAddress` field, a non-email string is also accepted.

I did a quick search and it seems that `class-validator` is used with TypeORM for input validations: 
https://github.com/typeorm/typeorm/blob/master/docs/validation.md

The resulting database error when inserting non-valid data could be wrapped in a `ValidationError` result such that the frontend also knows what went wrong.
I also noticed that the customer service does a manual check for existing email addresses. Isn't it easier to insert the new customer and check for a unique conflict error by adding the [@Unique()](https://typeorm.io/#/decorator-reference/unique) decorator?

RubenDarioGuerreroNeira:
The hard-coded path to the test output directory in the cleanup
script was causing issues in different monorepo setups.

This commit introduces a `globalSetup` script for Vitest that reads
the `outDir` path from `e2e-common/tsconfig.e2e.json` and uses it
to clean the test output directory before tests are run. This makes
the e2e test setup more robust and portable.

Closes #719


RubenDarioGuerreroNeira:

dlhck:
Closing this due to multiple reasons:
- Does not resolve the issue mentioned 
- Destroys formatting of files 
- Adds a lot of useless files 
- Looks a lot of like AI generated code without any purpose 

Feel free to improve the quality of the PR and and reopen it again.

PR #3726, opened 2025-08-02
fix(dashboard): Enable dashboard React JSX compilation

hans-rollingridges-dev:
# Description

Dashboard compilation is failing because React JSX templates from Vendure plugins are not being emitted during the build process. This prevents the dashboard from loading the necessary React JSX modules.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript compilation settings to improve React JSX code output.

### DIFF
--- a/packages/dashboard/vite/utils/compiler.ts
+++ b/packages/dashboard/vite/utils/compiler.ts
@@ -154,6 +154,7 @@ async function compileTypeScript({
         target: ts.ScriptTarget.ES2020,
         module: ts.ModuleKind.CommonJS,
         moduleResolution: ts.ModuleResolutionKind.Node10, // More explicit CJS resolution
+        jsx: ts.JsxEmit.ReactJSX, // Emit React JSX optimized for production
         experimentalDecorators: true,
         emitDecoratorMetadata: true,
         esModuleInterop: true,


hans-rollingridges-dev:

dlhck:
This should not be necessary, as we are processing the JSX differently. The compilation in that case is only for the Vendure config and plugins discovery. Can you please open an issue and add a minimal reproduction repo of your setup, where it is not able to find the extensions?

hans-rollingridges-dev:
You're right, this shouldn't be necessary. However, the dashboard compilation is failing because it tries to process React JSX from the Vendure plugins themselves, not from the dashboard's own code. For example, plugins that use React for rendering rich-text emails or PDF invoices are causing the build to fail.

Should a reproduction still be needed, I can create a minimal example. I'll use a clean `npx @vendure/create my-shop` setup and add a simple plugin with a React render-to-PDF component.

hans-rollingridges-dev:
I removed the word "extensions" from the title because it is rather confusing in this context.

dlhck:
Thank you for the detailed information, that helps us _a lot_! We'll look into that!

hans-rollingridges-dev:
FYI, I only tested the proposed compile option `jsx: ts.JsxEmit.ReactJSX`, by directly modifying the installed Vendure dashboard package within the `node_modules` directory.

hans-rollingridges-dev:
It's also worth noting that in my test project, I've overridden the `react` dependency in the package.json to match the version used in `@vendure`.

biggamesmallworld:
Confirmed on current `master`. The react-pdf / react-email case you described is the one that matters: a plugin importing a `.tsx` emits raw JSX into the `.js` output with no diagnostic, and the config import then fails with the misleading `Could not find a variable exported as VendureConfig`. Your fix is correct.

`compiler.ts` was rewritten since you opened this (#4561 swapped `ts.createProgram` for per-file `ts.transpileModule`), so the diff no longer applies. Please rebase onto `master` and make the four changes below — everything you need is written out, so it should be mostly copy-paste.

**1.** In `compileTypeScript`, add to the `compilerOptions` object:

```ts
// Plugin import graphs can transitively reach .tsx files (e.g. server-side
// React for PDF or email rendering). Without this, transpileModule emits raw
// JSX into the .js output and the compiled config fails to import.
jsx: ts.JsxEmit.ReactJSX,
```

**2.** Add `packages/dashboard/vite/tests/fixtures-jsx/` — four files:

`package.json`
```json
{
    "type": "commonjs",
    "name": "jsx",
    "version": "0.0.1",
    "main": "index.ts"
}
```

`vendure-config.ts`
```ts
import { VendureConfig } from '@vendure/core';

import { MyPlugin } from './my-plugin/src/my.plugin.js';

    apiOptions: { port: 3000 },
    authOptions: { tokenMethod: 'bearer' },
    dbConnectionOptions: { type: 'postgres' },
    paymentOptions: { paymentMethodHandlers: [] },
    plugins: [MyPlugin],
};
```

`my-plugin/src/invoice.tsx`
```tsx
/**
 * Simulates server-side React usage in a plugin, e.g. rendering PDF invoices
 * or email templates. Imported by the plugin, so it ends up in the config
 * import graph that the dashboard compiler walks.
 */
export const Invoice = () => <div>Invoice</div>;
```

`my-plugin/src/my.plugin.ts`
```ts
import { PluginCommonModule, VendurePlugin } from '@vendure/core';

import { Invoice } from './invoice.js';

@VendurePlugin({
    imports: [PluginCommonModule],
    providers: [],
    dashboard: './dashboard/index.tsx',
})
export class MyPlugin {
    static invoice = Invoice;
}
```

**3.** Add `packages/dashboard/vite/tests/jsx-compilation.spec.ts`:

```ts
import { readFile, rm } from 'node:fs/promises';
import { join } from 'node:path';
import { describe, expect, it } from 'vitest';

import { compile } from '../utils/compiler.js';
import { debugLogger, noopLogger } from '../utils/logger.js';

describe('compiling plugins which import .tsx files', () => {
    for (const module of ['commonjs', 'esm'] as const) {
        it(`should transform JSX in ${module} mode`, { timeout: 60_000 }, async () => {
            const tempDir = join(__dirname, `./__temp/jsx-${module}`);
            await rm(tempDir, { recursive: true, force: true });

            const result = await compile({
                outputPath: tempDir,
                vendureConfigPath: join(__dirname, 'fixtures-jsx', 'vendure-config.ts'),
                logger: process.env.LOG ? debugLogger : noopLogger,
                module,
            });

            const emitted = await readFile(join(tempDir, 'my-plugin', 'src', 'invoice.js'), 'utf-8');
            expect(emitted).toContain('react/jsx-runtime');
            expect(emitted).not.toContain('<div');

            expect(result.pluginInfo).toHaveLength(1);
            expect(result.pluginInfo[0].name).toBe('MyPlugin');
        });
    }
});
```

**4.** Update the PR title and description. The current wording points at dashboard extension bundling, which Vite handles — the actual bug is `.tsx` leaking into the config / plugin-discovery compile. Suggested:

Title:

```
fix(dashboard): Transform JSX when compiling plugin config imports
```

Description:

```
The dashboard's config compiler walks the Vendure config's import graph and
transpiles each file with ts.transpileModule(). No `jsx` option was set, so
any .tsx file reachable from the config — e.g. server-side React used for PDF
invoice or email template rendering — was emitted with its JSX untransformed.

TypeScript reports no diagnostic for this, so the failure only surfaces when
the compiled config is imported, as a misleading
"Could not find a variable exported as VendureConfig".
```

## Steps

```bash
# from your fork, on the PR branch
git remote add upstream https://github.com/vendure-ecommerce/vendure.git   # skip if already added
git fetch upstream master
git checkout fix/dashboard-extensions-build
git rebase upstream/master

# apply changes 1-3, then verify from packages/dashboard/
cd packages/dashboard && bunx vitest run vite/tests

git add -A
git commit -m "fix(dashboard): Transform JSX when compiling plugin config imports"
git push --force-with-lease
```

The new spec should fail if you drop the `jsx` line and pass with it — worth checking both ways, since that's the whole point of the test.

Once that's up, I'll approve and merge. If you'd rather not pick it back up after this long, no problem at all — just say the word and we'll land it ourselves with you credited as co-author. Either way, thanks for tracking this down and for the clear explanation back in August; it made the diagnosis straightforward on our end.

biggamesmallworld:
Closing this in favour of #5152, which ships your fix.

You diagnosed this a year ago and the diagnosis was the hard part: the dashboard config compiler calls `ts.transpileModule()` with no `jsx` option, so any `.tsx` file reachable from the Vendure config is emitted with its JSX untransformed. TypeScript raises no diagnostic, and the only symptom is the misleading `Could not find a variable exported as VendureConfig`.

The reason this is a new PR rather than a merge of yours: #4561 replaced `ts.createProgram` with `ts.transpileModule`, so this branch no longer applies. #5152 is the same one-line change rebased onto the current compiler, plus a fixture and a spec covering commonjs and esm, which fail with exactly the error you reported when the `jsx` line is removed. You are credited as co-author on the commit.

Sorry it sat this long, and thanks for tracking it down.

Issue #3698, opened 2025-07-27
Issue in draft order for first time product addition to cart


BeepLoveKarki:
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/bd6208ec-553e-4580-80db-22d8eb90fb8e" />

So whenever I add product for the first time, I get this issue even if custom fields are nullable: true.

ronakmaheshwari:
can i work on it

dlhck:
@ronakmaheshwari sure! Just create a PR, I will assign this issue to you now.

ronakmaheshwari:
Hey @dlhck, I’ve addressed the issue related to invalid order state mutations by updating the logic in addItemsToOrder. The problem should now be resolved. Could you kindly review the PR when you get a chance? 

https://github.com/vendure-ecommerce/vendure/pull/3729

IdhayaBastine15:
Hey @dlhck, @BeepLoveKarki, I see this is assigned but PR #3729 was closed unmerged back in Aug 2025. Noticed it also just got tagged T2: Elevated risk a month ago, and #3729 was mentioned again a week ago, I wanted to check before assuming it's dormant: is this actively being worked, or open for someone else to pick up?
 
If it's open: I don't think the previous PR's theory (order state transition) was the actual bug, OrderService.addItemsToOrder on current main already has the correct state guard, extension-point calls, and price recalculation, matching what the reviewer asked for. The real issue looks client-side: the screenshot shows a GraphQL variable-coercion error (Boolean cannot represent a non boolean value:  "", etc.) for non-string OrderLine custom fields (Boolean/Int/Struct types),  this fails at schema validation, before any resolver runs, so a server-side fix can't touch it. Points to the legacy Admin UI's order-editor "add item" custom-fields form (packages/admin-ui/src/lib/or der/src/components/order-editor/order-editor.component.ts) sending untouched form controls as "" instead of omitting them/sending null.

Given the elevated-risk tag, happy to write this up in more detail or pair with someone before opening anything, just don't want to step on toes if it's already spoken for.

PR #3729, opened 2025-08-03
Fix order state check before mutating order lines

ronakmaheshwari:
# Description

This PR ensures that the order state is checked and transitioned to 'Modifying' before any order lines are created or mutated via addItemsToOrder.

# Breaking changes
This PR only changes internal flow control within OrderService.addItemsToOrder and does not affect external API behavior or existing API consumers.

# Screenshots
N/A – this is a backend logic fix and does not affect the UI.

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation to ensure items can only be added to orders in the correct state, with clear error messaging if not allowed.
  * Enhanced handling of custom fields and item quantities when adding items to an order.
  * Ensured the order reflects all latest changes after items are added.

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -581,7 +581,7 @@ export class OrderService {
         } else {
             return result.order;
         }
-    }
+    }   
 
     /**
      * @description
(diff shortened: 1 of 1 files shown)


ronakmaheshwari:
> I have read the CLA Document and I hereby sign the CLA

recheck

ronakmaheshwari:
recheck

ronakmaheshwari:

ronakmaheshwari:
recheck

ronakmaheshwari:
recheck

biggamesmallworld:
Thanks for taking a look at #3698, and apologies for the long delay in getting back to you.

I'm going to close this one, as the current diff can't be merged as-is. The main issues:

**Doesn't build.** `IllegalOperationError` is thrown but never imported into `order.service.ts`. Separately, `'Modify'` isn't a valid `OrderState`, the state is `'Modifying'`.

**The state guard blocks the normal path.** Per `default-order-process.ts`, `Modifying` is only reachable from `PaymentSettled`, `PartiallyShipped`, `Shipped`, `PartiallyDelivered` and `Delivered`. `addItemsToOrder` runs on orders in `AddingItems` or `Draft`, so the guard would reject every real call. `assertAddingItemsState()`, already called a few lines below, is the correct check for this path.

**Some key logic was removed:**
- The `orderInterceptors` / `willAddItemToOrder` block. That's a documented public extension point, so plugins relying on it would silently stop being called.
- `applyPriceAdjustments()`, replaced with a re-fetch via `getOrderOrThrow()`. Re-reading the row doesn't recalculate promotions, taxes or totals, so the returned order would have the new lines but stale amounts.
- The quantity branching. `constrainQuantityToSaleable()` already returns `quantity + existingOrderLineQuantity` in the unconstrained case, so adding `existingOrderLine.quantity` on top double counts it.

Also, passing `relations` into the initial `getOrderOrThrow(ctx, orderId, relations)` overrides the default relations, which include `lines`. A client that doesn't select `lines` would hit a runtime error on `order.lines.filter(...)`.

On the underlying bug: the custom field merge added here is a no-op, since `getExistingOrderLine()` already matches on custom fields and `getOrCreateOrderLine()` already runs `updateRelations()`. The nullable custom field problem in #3698 is most likely in that custom field comparison path, so a fix probably wants to start there rather than in `addItemsToOrder`.

If you'd like to take another run at it, please open a fresh PR from a topic branch with an e2e case covering the scenario from #3698 (draft order, first add of a variant with a nullable custom field). Happy to review it.

PR #3733, opened 2025-08-04
feat(core): Add @Override() decorator to assert method overrides

### LINKED ISSUE #2774: @Override decorator for overriding methods in core services
This question has come up before, and it is also something I'm running up against now.

In Java, there is the @Override decorator for overriding methods in superclasses.  There is no similar built-in concept in Javascript or Nestjs.

I found this Github issue for Nestjs that describes a way someone created and added an @Override decorator. https://github.com/nestjs/nest/issues/3457

Is this something we might want to add to Vendure core to make it possible for plugins to override certain parts of core services?  

It's not the ideal way to make the vast majority of customizations, but sometimes it might be necessary.  For instance, I am still banging my head against the wall a bit on the best way to achieve some inventory process changes I want to make, where the currently available injectable strategies are just not enough.

RubenDarioGuerreroNeira:
Closes #2774

# Description

This PR addresses [issue #2774](https://github.com/vendure-ecommerce/vendure/issues/2774), which highlights the use of hard-coded credentials in the default SQLite configuration.

To resolve this:

I replaced the hard-coded username and password values in the SQLite config file with environment variables (process.env.DB_USERNAME and process.env.DB_PASSWORD).

I updated the relevant configuration to fall back to sensible defaults only if the environment variables are not defined.

This change improves security and aligns with best practices for handling sensitive information, especially when using version control or deploying to different environments.

## Summary by CodeRabbit

* **New Features**
  * Introduced a decorator to ensure that a method properly overrides a superclass method, providing runtime validation in addition to compile-time checks.

* **Chores**
  * Updated TypeScript configuration for improved output management and clearer inclusion/exclusion of files.

### DIFF
--- a/packages/core/src/common/decorators/override.decorator.ts
+++ b/packages/core/src/common/decorators/override.decorator.ts
@@ -1,0 +1,47 @@
+/**
+ * @description
+ * A method decorator which asserts that the decorated method is overriding a method of the superclass.
+ * This is primarily to guard against cases where a method is intended to override a core service
+ * method, but a typo in the method name means it does not.
+ *
+ * This decorator should be used in conjunction with the TypeScript `override` keyword for full
+ * compile-time and run-time safety.
+ *
+ * @example
+ * ```ts
+ * import { Injectable } from '@nestjs/common';
+ * import { ProductService, ID, RequestContext, Override } from '@vendure/core';
+ *
+ * @Injectable()
+ * export class MyCustomProductService extends ProductService {
+ *
+ *   @Override()
+ *   override async findOne(
+ *     ctx: RequestContext,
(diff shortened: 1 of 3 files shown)


RubenDarioGuerreroNeira:
Thank you for maintaining this great project!
Looking forward to your feedback.
Best regards,
Rubén Darío Guerrero Neira
GitHub: [@RubenDarioGuerreroNeira](https://github.com/RubenDarioGuerreroNeira)

michaelbromley:
This is the 3rd PR from you that has more or less the same changes, with completely different descriptions.

Either this is AI gone wrong, or you're confused. Can you explain?

PR #3734, opened 2025-08-04
fix(core): Make payment state transitions idempoten

RubenDarioGuerreroNeira:
Problem Description
A race condition can occur during payment and refund processing, especially with gateways that use webhooks for transaction confirmation. The sequence of events is as follows:

A storefront mutation (e.g., addPaymentToOrder) is called.
Vendure communicates with the external payment gateway.
The gateway processes the transaction and sends a webhook notification to a Vendure endpoint.
The webhook handler receives this notification and transitions the Payment state to Settled (or Refunded).
Almost simultaneously, the original mutation from step 1 resolves and also attempts to transition the Payment to the Settled state.
Because the Payment is already in the Settled state, the state machine throws a PaymentStateTransitionError when attempting a Settled -> Settled transition. This results in an error being returned to the storefront, even though the payment was successfully processed by the gateway.

Solution
This fix makes the payment state transition logic idempotent.

In the PaymentService, the transitionToState method now checks if the fromState and toState are the same. If they are, the operation is treated as a successful "no-op" (no operation) and simply returns the Payment object without throwing an error.

This elegantly resolves the race condition: the first process to arrive (either the webhook or the mutation response) will change the state. The second process will find the state is already correct and will continue its execution flow successfully, preventing the error and ensuring a smooth user experience.

### DIFF
--- a/packages/core/src/common/decorators/override.decorator.ts
+++ b/packages/core/src/common/decorators/override.decorator.ts
@@ -1,0 +1,47 @@
+/**
+ * @description
+ * A method decorator which asserts that the decorated method is overriding a method of the superclass.
+ * This is primarily to guard against cases where a method is intended to override a core service
+ * method, but a typo in the method name means it does not.
+ *
+ * This decorator should be used in conjunction with the TypeScript `override` keyword for full
+ * compile-time and run-time safety.
+ *
+ * @example
+ * ```ts
+ * import { Injectable } from '@nestjs/common';
+ * import { ProductService, ID, RequestContext, Override } from '@vendure/core';
+ *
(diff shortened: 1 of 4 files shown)


RubenDarioGuerreroNeira:
Thank you for reviewing this PR and for maintaining such a well-structured project.
Happy to keep contributing and learning from this great community.

Best regards,
Rubén Darío Guerrero Neira
GitHub: [@RubenDarioGuerreroNeira](https://github.com/RubenDarioGuerreroNeira)

michaelbromley:
Hi, this PR does not seem to be making any relevant changes to the issue described.

PR #3756, opened 2025-08-16
fix(email-plugin): Escape recipient's HTML entities in dev mailbox

knoid:
# Description

This fixes the recipient property on the development mailbox so the name and email shows property in all kinds of accepted formats like `"John Doe" <john.doe@test.com>`.

# Breaking changes

# Screenshots

Before:
before

After:
after

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Bug Fixes**
  * Special characters in recipient fields are now escaped, preventing unintended HTML rendering in the dev mailbox. Fix applies to both the message list and the email detail view, avoiding visual glitches or injected markup from crafted recipient values.

* **Chores**
  * Added a centralized HTML-escaping helper to standardize recipient sanitization across the dev mailbox.

### DIFF
--- a/packages/email-plugin/dev-mailbox.html
+++ b/packages/email-plugin/dev-mailbox.html
@@ -155,15 +155,10 @@
             .then((res) => renderList(res));
     }
 
-    const ESCAPE_ENTITIES = {
-      '&': '&amp;',
-      '<': '&lt;',
-      '>': '&gt;',
-      '"': '&quot;',
-    };
-    
+    const ESCAPE_ENTITIES = { '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;' };
     const escapeChar = char => ESCAPE_ENTITIES[char] || char;
     const escapeHTMLEntities = value => String(value).replace(/[<>"&]/g, escapeChar);
+
     function renderList(items) {
         const list = document.querySelector('.list');
         list.innerHTML = '';


packages/email-plugin/dev-mailbox.html
```diff
@@ -155,6 +155,12 @@ <h1 class="heading">Vendure Dev Mailbox</h1>
             .then((res) => renderList(res));
     }
 
+    function escapeHTMLEntities(value) {
+        const str = String(value ?? '');
+        const map = { '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;', "'": '&#39;' };
+        return str.replace(/[&<>"']/g, (ch) => map[ch]);
+    }
+
```
DanielBiegler:
```suggestion
    const ESCAPE_ENTITIES = {
      '&': '&amp;',
      '<': '&lt;',
      '>': '&gt;',
      '"': '&quot;',
    };
    
    const escapeChar = char => ESCAPE_ENTITIES[char] || char;
    const escapeHTMLEntities = value => String(value).replace(/[<>"&]/g, escapeChar);
```

Small optimization because theres no need to reconstruct on every call, single quotes are fine and I'd show `null`s and `undefined`s for clarity.

michaelbromley (approved):

PR #3757, opened 2025-08-16
fix(admin-ui): Administrator's customField don't show up in profile if requiresPermission is Owner

knoid:
# Description

A custom field in the Administrator entity wouldn't show up in the Administrator's profile if the requiresPermission included `Permission.Owner`. This fix assumes that whoever is in the profile view is the owner so checking for the Owner permission should return true.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **New Features**
  * Profile: Custom fields now honor owner-specific permissions. Users can view and edit fields designated for “Owner” when managing their own profile, while non-owners won’t see these fields.
  * Permissions: Enhanced permission checks to support owner-aware access across custom fields, expanding what eligible users can see without affecting existing non-owner visibility.

### DIFF
--- a/packages/admin-ui/src/lib/core/src/common/base-detail.component.ts
+++ b/packages/admin-ui/src/lib/core/src/common/base-detail.component.ts
@@ -150,10 +150,13 @@ export abstract class BaseDetailComponent<Entity extends { id: string; updatedAt
         }
     }
 
-    protected getCustomFieldConfig(key: Exclude<keyof CustomFields, '__typename'>): CustomFieldConfig[] {
+    protected getCustomFieldConfig(
+        key: Exclude<keyof CustomFields, '__typename'>,
+        isOwner = false,
+    ): CustomFieldConfig[] {
         return this.serverConfigService.getCustomFieldsFor(key).filter(f => {
             if (f.requiresPermission?.length) {
-                return this.permissionsService.userHasPermissions(f.requiresPermission);
+                return this.permissionsService.userHasPermissions(f.requiresPermission, isOwner);
             }
             return true;
         });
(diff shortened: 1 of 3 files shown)


packages/admin-ui/src/lib/core/src/common/base-detail.component.ts
```diff
@@ -150,10 +150,13 @@ export abstract class BaseDetailComponent<Entity extends { id: string; updatedAt
         }
     }
 
-    protected getCustomFieldConfig(key: Exclude<keyof CustomFields, '__typename'>): CustomFieldConfig[] {
+    protected getCustomFieldConfig(
+        key: Exclude<keyof CustomFields, '__typename'>,
+        isOwner = false,
+    ): CustomFieldConfig[] {
```
DanielBiegler, line 156: This function gets called for basically all detail components, so changing the function signature for one single case makes me think twice. I only looked at it briefly so I dont have a better solution right now and you might be totally right, but I think having a closer look at the root cause might be worth it.

In any way, thanks for making a PR!

knoid: You are right that this is just for one case but I couldn't find any other way the admin-ui knows what entities "belong" to the active user. I think this is the most straightforward solution without implementing that belonging logic in the admin-ui that I could think of.

grolmus:
Thanks for the contribution, @knoid.

This targets the legacy Angular Admin UI (`packages/admin-ui`), which we're no longer actively maintaining — all new work is going into the React-based Dashboard (`@vendure/dashboard`). We're not taking further fixes to the old admin-ui, so I'm closing this one.

If the issue still applies in the new Dashboard, a PR against that would be very welcome. Appreciate you taking the time to look into it. 🙏

PR #3802, opened 2025-09-05
fix(create): Fix resolving path in repos using hoisted node_modules

### LINKED ISSUE #3506: @vendure/create is failing with "Failed to copy email templates"
**Describe the bug**
I am trying to create a new project from scratch and, whether I use manual or quick installation, it fails with the following

```
➜  packages git:(main) ✗ npx @vendure/create my-example
┌  Let's create a Vendure App :sparkles: v3.2.3
│
◇  How should we proceed?
│  Manual Configuration
│
◇  Using port 3000
│
◇  Which database are you using?
│  Postgres
│
◇  What's the database host address?
│  localhost
│
◇  What port is the database listening on?
│  5432
│
◇  What's the name of the database?
│  vendure
│
◇  What's the schema name we should use?
│  public
│
◇  Use SSL to connect to the database? (only enable if your database provider supports SSL)
│  no
│
◇  What's the database user name?
│  postgres
│
◇  What's the database password?
│  postgres
│
◇  What identifier do you want to use for the superadmin user?
│  superadmin
│
◇  What password do you want to use for the superadmin user?
│  superadmin
│
◇  Populate with some sample product data?
│  no
│
◇  Created package.json
│
◇  Successfully installed 6 dependencies
│
◇  Successfully installed 3 dev dependencies
│
◐  Generating app scaffold   Failed to copy email templates.
◇  Generated app scaffold
│
▲  Something went wrong
```

**To Reproduce**
1. Run `npx @verdure/create my-example

**Expected behavior**
A project is initiated

 - @vendure/core version: not sure
 - Nodejs version: 23.7.0

**Additional context**
I am attempting to create the project in a freshly created monorepo

HouseinIsProgramming:
Fixes #3506 

# Description
This PR adds a new helper function that uses `require.resolve()` to find the package directory instead of using `root + /path-to-package`.

This solves the problem where node_modules are hoisted (moved up the directory tree), causing the hardcoded path approach to fail when packages aren't in the expected location.

**Before:** `root + '/node_modules/package-name'`
**After:** `require.resolve('package-name')` to dynamically locate the package

Also adds `process.exit(0)` on errors to prevent "locking the console".

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

- New Features
  - None

- Bug Fixes
  - Added top-level error handling for the app-creation flow with clearer non-zero exit on failure.
  - Email template copying now reports failures and stops instead of failing silently.

- Refactor
  - Replaced hard-coded module location logic with dynamic package-resolution for improved portability.

- Chores
  - Added a helper to robustly resolve package root directories.

### DIFF
--- a/packages/create/src/create-vendure-app.ts
+++ b/packages/create/src/create-vendure-app.ts
@@ -499,8 +499,6 @@ async function createDirectoryStructure(root: string) {
 async function copyEmailTemplates(root: string) {
     const emailPackageDirname = resolveDirName('@vendure/email-plugin');
     const templateDir = path.join(emailPackageDirname, 'templates');
-    log(pc.bgRedBright(emailPackageDirname));
-    log(pc.bgRedBright(templateDir));
     try {
         await fs.copy(templateDir, path.join(root, 'static', 'email', 'templates'));
     } catch (err: any) {
--- a/packages/create/src/create-vendure-app.ts
+++ b/packages/create/src/create-vendure-app.ts
@@ -302,9 +302,9 @@ export async function createVendureApp(
 
     let superAdminCredentials: { identifier: string; password: string } | undefined;
     try {
-        const { populate } = await import(path.join(resolveDirName('@vendure/core'), '/cli/populate'));
+        const { populate } = await import(path.join(resolveDirName('@vendure/core'), 'cli', 'populate'));
         const { bootstrap, DefaultLogger, LogLevel, JobQueueService, ConfigModule } = await import(
-            path.join(resolveDirName('@vendure/core'), '/dist/index')
+            path.join(resolveDirName('@vendure/core'), 'dist', 'index')
         );
         const { config } = await import(configFile);
         const assetsDir = path.join(__dirname, '../assets');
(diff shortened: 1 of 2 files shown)


packages/create/src/create-vendure-app.ts
```diff
@@ -294,13 +298,13 @@ export async function createVendureApp(
 
     // register ts-node so that the config file can be loaded
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    require(path.join(root, 'node_modules/ts-node')).register();
+    require(path.join(resolveDirName('ts-node'))).register();
```
michaelbromley: `path.join` is redundant here

packages/create/src/helpers.ts
```diff
@@ -532,3 +532,30 @@ export function cleanUpDockerResources(name: string) {
         log(pc.yellow(`Could not clean up Docker resources`), { level: 'verbose' });
     }
 }
+
+export function resolveDirName(packageName: string) {
```
michaelbromley: Can we rename this to `resolvePackageRootDir`? The current name is slightly too generic.

Issue #3708, opened 2025-07-31
Schedule Task Stuck & Not Marked Stopped/Locked


BeepLoveKarki:
**Describe the bug**
If during any scheduled task run the server or worker gets stuck — or any similar issue occurs that halts the task — the task never runs again.
Even manual attempts to trigger it fail.
The only way to make it run again is to change the task ID and redeploy. Issue is in any version above 3.3.0.

**To Reproduce**

- Schedule a recurring task.
- During its execution, cause the server or worker to become stuck (e.g., simulate a crash or stall).
- Allow the task to halt midway.
- Attempt to run the task again manually.

**Expected behavior**
The task should be able to resume or run again (either on its next schedule or manually) after a stall without requiring a new ID or redeploy.

***Environment**

@vendure/core version: 3.3.2
Nodejs version: 22.22.0
Database: pgsql
Operating System: macOS
Browser: Chrome
Package manager: npm

**Actual behavior**
Once a task has stalled, it never runs again — even with manual triggering — unless the task ID is changed and the application is redeployed.

**Workaround**
Change the task ID and redeploy the application to make the task run again.

michaelbromley:
Hi @BeepLoveKarki 

I know we discussed this on Discord, but _please_ always use the issue template and supply **full context**. It is quite likely someone other than myself will be assigned to work on this, and the current issue description will make it very hard for them to know what to fix.

For your reference, here's the template. Please use it!

---

**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:

1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error

**Expected behavior**
A clear and concise description of what you expected to happen.

**Actual behavior**
A clear and concise description of what actually happened.

**Screenshots/Videos**
If applicable, add screenshots or videos to help explain your problem.

**Error logs**
If applicable, paste any relevant error messages or stack traces:

```
Paste error logs here
```

**Environment (please complete the following information):**

- @vendure/core version:
- Nodejs version:
- Database (mysql/postgres etc):
- Operating System (Windows/macOS/Linux):
- Browser (if applicable):
- Package manager (npm/yarn/pnpm):

**Configuration**
If relevant, share your Vendure configuration (remove sensitive data):

```typescript
// Your configuration here
```

**Minimal reproduction**
If possible, provide a minimal code example or repository that reproduces the issue.

**Workaround**
If you found a temporary workaround, please describe it here.

[...]

PR #3804, opened 2025-09-08
fix: Clear stale locks when past the scheduled task interval

biggamesmallworld:
ref: #3708 

# Description

This PR introduces a method which handles locked tasks in the case that the Vendure worker goes down. The approach is as follows - whenever a job runs, we loop through the existing scheduled tasks, check if we are outside of the interval that the task has run on, and if so we set the lockedAt value to null. 

## Caveat
This assumes that in most cases, scheduled tasks should not take longer than their intervals. In the case that this happens, there is a chance that the worker would pick up a task that isn't finished yet and start it again. This would result in additional system load, but probably no significant impact or data loss. We would assume that the devs scheduling the tasks would be aware of roughly how long this would take

# Breaking changes

None

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

- New Features
  - Scheduler now clears stale task locks before executing tasks, reducing stuck executions and improving stability.
  - A new TaskService is provided and exposed via the public service API to support programmatic lock cleanup and observability.

- Chores
  - Dev server config includes an example long‑running task plugin to aid local testing of extended task durations.


michaelbromley:
Thanks Will! I made a couple of refactors and simplified things a little, but this is solid 👍

PR #3815, opened 2025-09-16
Add api key authentication strategy

### LINKED ISSUE #3412: API Keys

Vendure doesn't have nor currently supports creation and use of API Keys as means of auth. This can be cumbersome for server side scripting where you'd like to one-off mutate something.

API Keys are generally nice for server side scripting. For example when I need to interact with any payment provider I dont have to login either, I provide the api key in the header and thats it. A simple curl request is enough to do something.

If I just want to do a one-off mutation I'd have to login, wait, parse and get the session token, do the request and logout again. If I want to do get rid off this round trip, I'd have to manage state and take care of storing and keeping my token up to date and what happens if my token gets invalidated in between multiple requests, now I have to care about refreshing and retrying failed requests. This makes the job quite a bit more complicated than it needs to be.  A long lived API key easily gets around all of this and makes it simpler.

One usecase is doing stuff on the backend that is not directly tied to a user like importing, fetching or mutating custom entities that should be behind admin permissions but not really behind a specific user. It makes the distinction between users and scripts very clear for example for cronjobs that sync/create entities.

Imagine your storefront wants to update some custom entities behind the scenes to keep track of something but its not tied to the user who did the action. Right now you must have some user for that and log in, do the thing, log out.

There are some other benefits like not polluting the session table, per key ratelimiting and usage insights, revokation/rerolling and more.

Long lived API Keys that I can scope to specific tasks that server side scripts use to do their work.

.

**Additional context**

- Medusa Docs about API keys: [Link](https://docs.medusajs.com/resources/commerce-modules/api-key)
- Stripe Docs about API keys: [Link](https://docs.stripe.com/keys)
- Mollie Docs about API keys: [Link](https://docs.mollie.com/reference/authentication)

DanielBiegler:
# Description

Adds long lived API keys as authentication method for easier scripting

## ToDos

- [x] Authentication method works (after creating a DB row manually)
- [ ] GraphQl mutation for assigning keys to users
- [ ] Currently you can logout an api key session, should that even be possible? (see `base-auth.resolver.ts@logout`, we could add a simple check in `auth.service.ts@destroyAuthenticatedSession`
- [ ] UI
- [ ] Tests
- [ ] Point branch to minor
    - Havent done so yet because its not up to date and would show many irrelevant commits
- [ ] Docs (also look for mentions of other auth methods)
- [ ] Remove all TODOs

Fixes #3412 

## Misc

Example login via new auth method for quick reference:

# Breaking changes

A migration is needed because the `AuthenticationMethod` entity got extended, see `ApiKeyAuthenticationMethod`

Typeorm automatically makes the `apiKey` database column nullable due to the table inheritance of `AuthenticationMethod` so the migration simply adds an "empty" column, should be fast enough for any setup.  

# Screenshots

--

# Checklist

📌 Always:

👍 Most of the time:


mohdbk:
@DanielBiegler I've implemented the complete workflow of api keys with service accounts (tests, entities, strategy, dashboard ui, etc.). Let me know if you would like to discuss. I can share the codebase for you to review.

image

packages/core/src/api/schema/admin-api/api-key.api.graphql
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Generates a new API-Key and attaches it to an Administrator.
+    Returns the generated API-Key.
+    API-Keys cannot be viewed again after creation.
+    """
+    createApiKey(input: CreateApiKeyInput!): CreateApiKeyResult!
+    "Updates an API-Key"
+    updateApiKey(input: UpdateApiKeyInput!): ApiKey!
+    "Deletes an API-Key"
+    deleteApiKey(input: DeleteApiKeyInput!): DeletionResponse!
```
DanielBiegler: Mh this doesnt follow the pattern of Vendures other deletion mutations, something like this would follow more closely:

```graphql
deleteApiKeys(ids: [ID!]!): [DeletionResponse!]!
```

Initially when scaffolding the PR I just liked this cleaner look of specific input types, might wanna change it.

packages/core/src/api/schema/admin-api/api-key.api.graphql
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Replaces the old with a new API-Key.
+    This is a convenience method to invalidate an API-Key without
+    deleting the underlying roles and permissions.
+    """
+    rotateApiKey(input: RotateApiKeyInput!): RotateApiKeyResult!
```
DanielBiegler: Similarly to the deletion mutation, might wanna potentially change the input type to something similar to:

```graphql
rotateApiKeys(ids: [ID!]!): [RotateApiKeyResult!]!
# or
rotateApiKeys(ids: [ID!]!): [CreateApiKeyResult!]!
```

Just thinking out loud, this'd lend itself more to bulk actions :shrug:

DanielBiegler: ehh lets not do bulk for now

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Returns the appropriate {@link ApiKeyAuthorizationOptions} based on the {@link ApiType}.
+     * This is needed because the admin and shop ApiKey options may be configured differently.
+     */
+    getApiKeyAuthOptionsByApiType(apiType: ApiType) {
+        return apiType === 'admin'
+            ? this.configService.authOptions.adminApiKeyAuthorizationOptions
+            : this.configService.authOptions.shopApiKeyAuthorizationOptions;
+    }
```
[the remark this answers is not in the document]
DanielBiegler: Generally this should be infallible I think, since Vendure default config provides this through the default config (?) A bit of defensive programming couldnt hurt though, true.

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,344 @@
...
+                channelPermissions.id,
+                channelPermissions.permissions,
+            );
+
+            if (!isAllowed)
+                throw new UserInputError('error.active-user-does-not-have-sufficient-permissions');
+        }
+
+        return roles;
+    }
```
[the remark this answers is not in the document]
DanielBiegler: Same behavior in admin service -- If we decide to be more strict here add it to admin service too.

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Helper, intended for the AuthGuard to quickly update the lastUsedAt timestamp
+     */
+    async updateLastUsedAtByLookupId(lookupId: NonNullable<ApiKey['lookupId']>): Promise<UpdateResult> {
+        return this.connection.rawConnection
+            .getRepository(ApiKey)
+            .update({ lookupId }, { lastUsedAt: new Date() });
+    }
```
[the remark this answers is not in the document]
DanielBiegler: Disagree. This function should generally only be called for keys that get used, so worst case if there is a bug, updating the soft deleted one might help someone spot said bug.

packages/common/src/generated-types.ts
```diff
@@ -5832,6 +5997,12 @@ export type RoleSortParameter = {
   updatedAt?: InputMaybe<SortOrder>;
 };
 
+export type RotateApiKeyResult = {
+  __typename?: 'RotateApiKeyResult';
+  /** The generated API-Key. API-Keys cannot be viewed again after creation! */
+  apiKey: Scalars['String']['output'];
+};
+
```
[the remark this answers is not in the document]
DanielBiegler: What do you want from me?

packages/core/src/config/config.service.ts
```diff
@@ -43,7 +42,7 @@ export class ConfigService implements VendureConfig {
         return this.activeConfig.apiOptions;
     }
 
-    get authOptions(): Required<AuthOptions> {
+    get authOptions() {
         return this.activeConfig.authOptions;
     }
```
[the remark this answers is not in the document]
DanielBiegler: No, removed it on purpose due to the new type.

packages/core/src/entity/api-key/api-key-translation.entity.ts
```diff
@@ -0,0 +1,29 @@
...
+import { VendureEntity } from '../base/base.entity';
+import { CustomApiKeyFieldsTranslation } from '../custom-entity-fields';
+
+import { ApiKey } from './api-key.entity';
+
+@Entity()
+export class ApiKeyTranslation extends VendureEntity implements Translation<ApiKey>, HasCustomFields {
+    constructor(input?: DeepPartial<Translation<ApiKeyTranslation>>) {
+        super(input);
+    }
```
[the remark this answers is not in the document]
DanielBiegler: No, its fixed and conforming to the other entities like product, facets, collections, etc.

packages/core/src/api/resolvers/admin/api-key.resolver.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    @Transaction()
+    @Mutation()
+    @Allow(Permission.DeleteApiKey)
+    async deleteApiKeys(
+        @Ctx() ctx: RequestContext,
+        @Args() { ids }: MutationDeleteApiKeysArgs,
+    ): Promise<DeletionResponse[]> {
+        return Promise.all(ids.map(id => this.apiKeyService.softDelete(ctx, id)));
+    }
```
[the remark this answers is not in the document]
DanielBiegler: This pattern is used throughout basically all resolvers. Doesnt seem to be an issue for several years now (?)

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -59,7 +63,10 @@ export class AuthGuard implements CanActivate {
         if (targetIsFieldResolver) {
             requestContext = internal_getRequestContext(req);
         } else {
-            const session = await this.getSession(req, res, hasOwnerPermission);
+            const apiKeyHashingStrategy = this.apiKeyService.getApiKeyAuthOptionsByApiType(
```
michaelbromley: Do this inside this.getSession() and don't pass it as a arg

packages/core/src/config/api-key-strategy/api-key-generation-strategy.ts
```diff
@@ -0,0 +1,32 @@
...
+ * - `authOptions.adminAuthenticationStrategy`
+ * - `authOptions.shopAuthenticationStrategy`
+ *
+ * of your VendureConfig.
+ * :::
+ *
+ * @docsCategory auth
+ * @since 3.6.0
+ */
+export interface ApiKeyGenerationStrategy extends InjectableStrategy {
```
michaelbromley: Let's consolidate generateApiKey and lookupId into a single `ApiKeyStrategy`
Also include the hashing interface in it so we only need to supply the strategy in the AuthOptions property directly - no nested options object.

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -59,7 +63,10 @@ export class AuthGuard implements CanActivate {
...
+                getApiType(info),
+            ).hashingStrategy;
+            const session = await this.getSession(req, res, hasOwnerPermission, apiKeyHashingStrategy);
```
michaelbromley: We need to be able to handle the scenario where the API Key session got deleted for some reason (eg manual "clean up" of session table)

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,344 @@
...
+        return roles;
+    }
+
+    private generateApiKeyUserIdentifier(lookupId: string): string {
+        // Too long identifiers could be an issue depending on the underlying DB
+        return `apikey-user-${lookupId}`;
+    }
+
+    /**
+     * Creates a new API-Key for the given User
```
michaelbromley, line 116: Put @description on all of these doc blocks

packages/core/src/config/vendure-config.ts
```diff
@@ -398,6 +405,20 @@ export interface AuthOptions {
...
+     * @default 'vendure-api-key-lookup'
+     */
+    apiKeyLookupHeaderKey?: string;
+    /**
+     * @description
+     * Defines which header will be used to read the API-Key when using the 'api-key' token method.
+     *
+     * @default 'vendure-api-key'
+     */
+    apiKeyHeaderKey?: string;
```
michaelbromley, line 414: Let's just use a single string for the API key which incorporates the lookup and uses a delimiter to separate them. The delimiter will therefore need to be configurable since the lookup & apiKey generation is also configurable. A sensible default might be `:`

packages/core/src/api/middleware/auth-guard.ts
DanielBiegler: The reason for that initially was the extract-session-token function, but I reworked it now to be way simpler and this also got rid off the argument passing! :+1:

packages/core/src/config/api-key-strategy/api-key-generation-strategy.ts
DanielBiegler: Done! Similar to the `StockLocationStrategy` which introduces a `BaseStockLocationStrategy` in order to have shared functions between implementations, there now exists a `BaseApiKeyStrategy` which makes introducing new api key strategies even simpler.

packages/core/src/api/middleware/auth-guard.ts
DanielBiegler: Done! After we make sure that the provided token is valid but we're unable to find a session for it, we re-create it. Tested it locally and it works but I left a comment which we can discuss tomorrow. For ref. see `auth-guard.ts@getSessionFromToken`

packages/core/src/config/vendure-config.ts
DanielBiegler: Done. See new `RandomBytesApiKeyStrategy`, `BaseApiKeyStrategy` and `ApiKeyStrategy`

packages/core/src/config/config.module.ts
```diff
@@ -158,6 +160,8 @@ export class ConfigModule implements OnApplicationBootstrap, OnApplicationShutdo
             ...(instrumentationStrategy ? [instrumentationStrategy] : []),
             ...orderInterceptors,
             schedulerStrategy,
+            adminApiKeyStrategy,
+            shopApiKeyStrategy,
```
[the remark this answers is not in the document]
DanielBiegler: You seem to be hallucinating. Bad bot.

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,380 @@
...
+     *
+     * If you allow users to specify these IDs, they may leak existing User IDs via thrown errors.
+     *
+     * @throws {EntityNotFoundError} When either Owner or ApiKeyUser cannot be found
+     * @throws {UserInputError} When the User tries to grant a role which they themselves dont have
+     */
+    async create(
+        ctx: RequestContext,
+        input: CreateApiKeyInput,
+        userIdOwner: ID,
```
martijnvdbrug, line 129: Maybe a little nitpicky, but I had some trouble understanding the methods description in combination with the `userIdOwner` param. Would it make sense to name this something like `callerUserId`, so that it is clear that we need the callers userId to check for privilege escalation?

packages/core/src/config/api-key-strategy/random-bytes-api-key-strategy.ts
```diff
@@ -0,0 +1,91 @@
...
+export interface RandomBytesApiKeyStrategyOptions {
+    /**
+     * @description
+     * The number of bytes to generate. The size must not be larger than 2**31 - 1.
+     * This is the input size for `crypto.randomBytes`
+     *
+     * @default 32
+     * @see {@link DEFAULT_SIZE_RANDOM_BYTES_KEY}
+     */
+    sizeKey?: number;
```
michaelbromley: let's do `keySize` etc

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -168,4 +179,65 @@ export class AuthGuard implements CanActivate {
...
+        }
+
+        const strategy = this.apiKeyService.getApiKeyStrategyByApiType(apiType);
+        const parseResult = strategy.parse(extracted.token);
+        if (!parseResult) {
+            return;
+        }
+
+        const apiKey = await this.apiKeyService.findOneByLookupId(
+            RequestContext.empty(),
```
michaelbromley: RequestContextService.create() // specify channelId & languageCode

channel can be found using ChannelService.getChannelFromToken(req)

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,380 @@
...
+              })
+            : await this.userService.createApiKeyUser(
+                  ctx,
+                  roles,
+                  this.generateApiKeyUserIdentifier(lookupId),
+              );
+
+        const secret = await strategy.generateSecret(ctx);
+        const apiKey = strategy.constructApiKey(lookupId, secret);
+        // TODO(Dan): should we only hash the secret or the entire apikey?
```
michaelbromley: fine, keep it as is

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -168,4 +179,65 @@ export class AuthGuard implements CanActivate {
...
+        }
+
+        // At this point we may assert:
+        // 1. The token came from the api-key header
+        // 2. The hash matches
+        // 3. There is no session
+        // We can conclude that the API-Key is actually broken.
+        // For example someone could have deleted the session manually in the DB.
+        // We must create a new session, otherwise the API-Key is unusable.
+        const newSession = await this.sessionService.createNewAuthenticatedSession(
```
michaelbromley: remove unused variable

packages/core/e2e/api-key.e2e-spec.ts
```diff
@@ -0,0 +1,225 @@
...
+        expect(result.updateApiKey.translations.find(t => t.languageCode === LanguageCode.de)?.name).toBe(
+            'Neuer Eintrag',
+        );
+    });
+
+    it('deleteApiKey', async ({ expect }) => {
+        const result = await adminClient.query(DeleteApiKeyDocument, { ids: [createdApiKeyId] });
+        expect(result.deleteApiKeys?.[0]?.result).toBe(DeletionResult.DELETED);
+        // Should not be found anymore
+        const { apiKey } = await adminClient.query(API_KEY, { id: createdApiKeyId });
```
michaelbromley, line 81: use the generated doc

packages/core/src/api/schema/admin-api/api-key.api.graphql
```diff
@@ -0,0 +1,81 @@
...
+
+    translations: [CreateApiKeyTranslationInput!]!
+}
+
+type CreateApiKeyResult {
+    """
+    ID by which you can look up the API-Key.
+    Also helps you identify keys without leaking the underlying secret API-Key.
+    """
+    lookupId: String!
```
michaelbromley: don't need to return this

michaelbromley (dismissed):
Great work!

Address the few remaining comments, and investigate failing tests on postgres order merge suite

packages/core/src/service/services/api-key.service.ts
DanielBiegler: The owner basically only becomes relevant for plugins where customers can own a key, so I think in line with the `Permission.Owner` - Owner is clearer than caller.

packages/core/e2e/api-key.e2e-spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+`;
+
+export const ROTATE_API_KEY = gql`
+    mutation RotateApiKey($id: ID!) {
+        rotateApiKey(id: $id) {
+            apiKey
+        }
+    }
+`;
```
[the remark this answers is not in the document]
DanielBiegler: No, they are the basis for code generation.

packages/core/src/api/schema/admin-api/api-key.api.graphql
```diff
@@ -0,0 +1,76 @@
+type Query {
+    apiKey(id: ID!): ApiKey
+    apiKeys(options: ApiKeyListOptions): ApiKeyList!
+}
+
+type Mutation {
+    """
```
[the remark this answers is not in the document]
DanielBiegler: Thats a larger refactoring task that shouldnt be fixed in this PR.

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,376 @@
...
+            relations: [...(relations ?? []), 'channels'],
+            where: {
+                lookupId,
+                deletedAt: IsNull(),
+                channels: { id: ctx.channelId },
+            },
+        });
+        if (!entity) return null;
+        return this.translator.translate(entity, ctx);
+    }
```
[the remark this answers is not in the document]
DanielBiegler: Fixed in 2e64929 by constructing the context from the request

packages/core/src/service/services/session.service.ts
```diff
@@ -108,21 +113,29 @@ export class SessionService implements EntitySubscriberInterface, OnApplicationB
...
                 : undefined;
         const existingOrder = await this.orderService.getActiveOrderForUser(ctx, user.id);
         const activeOrder = await this.orderService.mergeOrders(ctx, user, guestOrder, existingOrder);
+
+        const expires = this.getExpiryDate(
+            authenticationStrategyName === API_KEY_AUTH_STRATEGY_NAME
+                ? API_KEY_AUTH_STRATEGY_DEFAULT_DURATION_MS
+                : this.sessionDurationInMs,
+        );
+
```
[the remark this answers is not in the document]
DanielBiegler: Going with the defaults where api keys have 100 year long sessions and regular sessions are 1 year long, this edge case would technically only occur after 99.5 years. And due to the new re-generation of sessions in the auth guard an invalidated session should become valid again.

Would of course still be good to fix it just in case someone misconfigures their sessions somehow.

After checking `this.isAuthenticatedSession(session)` -- is the `authenticationStrategy` field even populated?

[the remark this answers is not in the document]
DanielBiegler: OK I tested it locally and it works.

packages/core/src/api/middleware/auth-guard.ts
```diff
@@ -168,4 +180,68 @@ export class AuthGuard implements CanActivate {
...
+        if (!apiKey) {
+            return;
+        }
+
+        const isHashMatching = await strategy.hashingStrategy.check(extracted.token, apiKey.apiKeyHash);
+        if (!isHashMatching) {
+            return;
+        }
+
+        this.apiKeyService
```
kevmtt: Investiage an alternative solution to make `lastUsedAt` session based, meaning:
- shorter API-Key sessions
- `lastUsedAt` is only updated on session creation (when it expired)

Therefore we have less database write operations in case of API-key usecases that will fire many requests at once or in a short time range.

packages/core/src/api/schema/admin-api/api-key.type.graphql
```diff
@@ -0,0 +1,44 @@
+type ApiKey implements Node {
+    id: ID!
+    createdAt: DateTime!
+    updatedAt: DateTime!
+
+    """
+    ID by which we can look up the API-Key.
+    Also helps you identify keys without leaking the underlying secret API-Key.
+    """
+    lookupId: String!
```
kevmtt, line 10: Is this something like a "public key"?
Here we could think about a better name imo

packages/core/src/api/schema/admin-api/api-key.type.graphql
```diff
@@ -0,0 +1,44 @@
...
+    "Helps you identify unused keys"
+    lastUsedAt: DateTime
+
+    """
+    Usually the user who created the ApiKey but could also be used as the basis for
```
kevmtt, line 15: I think we can improve the documenation to really explain the difference between `owner` and `user`. Or maybe it's worth to rename one to `creator`? Otherwise let's just ask an AI to explain it to a 5 year old haha

packages/core/src/entity/api-key/api-key-translation.entity.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    @Index()
+    @ManyToOne(type => ApiKey, base => base.translations, { onDelete: 'CASCADE' })
+    base: ApiKey;
+
+    @Column(type => CustomApiKeyFieldsTranslation)
+    customFields: CustomApiKeyFieldsTranslation;
+
+    @Column()
+    name: string;
```
kevmtt, line 28: My inner Monk is somehow yelling at me, that all Localized Columns should be on the top.

packages/core/src/service/services/api-key.service.ts
```diff
@@ -0,0 +1,376 @@
...
+    async create(
+        ctx: RequestContext,
+        input: CreateApiKeyInput,
+        userIdOwner: ID,
+        /**
+         * Optionally allow overriding the creation of a separate User.
+         * This is an advanced use case for plugin-authors to allow impersonation.
+         * You are responsible for avoiding privilege escalation by verifying this ID.
+         */
+        userIdApiKeyUser?: ID,
```
kevmtt, line 135: how about `apiKeyUserId`?

packages/core/src/api/schema/admin-api/api-key.type.graphql
DanielBiegler: Not quite, it has nothing to do with encryption, it is an ID thats needed for looking up the api key hash. This is because when a request comes in and you're using a hashing method with automatic salting, like bcrypt/argon/.. , it'd create a differing hash for the same input on every request.

Some projects "solve" this by simply fetching multiple hashes into memory and comparing the input to all of them until one matches... Yikes

The proper solution, which for example also Shopware uses, is to have some identifier for looking up the hash, hence lookup-id. Shopware calls it "access key" though which seems fair enough too. The PR description has some other comparisons. Cheers.

packages/core/src/api/middleware/auth-guard.ts
DanielBiegler: Done! By configuring the newly introduced time interval field, lets say "15 minutes", we only execute the update query once per key per time interval. Theres only one place where we need to update the usage timestamp (auth guard) and there we already have the api key in memory, so we can simply compare the last used date and check for the interval. This may result in drastically fewer db writes.

I added a detailed comment on the api key strategy which explains why and how this works. 

Heres the commit: https://github.com/vendure-ecommerce/vendure/pull/3815/commits/08db7b61455336fd2339a88cd6d76eb04e0816fb

PR #3816, opened 2025-09-16
feat(core,dashboard, docs): API keys (create, rotate, revoke, expire) with settings dashboard UI

### LINKED ISSUE #3412: API Keys

mohdbk:
# Description
  
Fixes #3412. This PR adds header‑based Admin (and optional Shop) API key authentication, persisted sessions linked to keys, per‑scope configuration, and a dedicated API Keys UI in the React Dashboard. Sessions
created with API keys align to key expiry by default, with overrides available.

### Core

- API key model and API
- New ApiKey entity (hash at rest, one‑time raw reveal, expiresAt required, lastUsedAt, status: 'active'|'revoked', optional notes, scoped admin|shop).
- Admin API: new GraphQL types and mutations:
- `createApiKey(name: String!, expiresAt: DateTime!, notes?: String)`
- `rotateApiKey(id: ID!)` (revokes old key + invalidates its sessions; returns new raw key)
- `revokeApiKey(id: ID!)` (invalidates sessions)
- `deleteApiKey(id: ID!): DeletionResponse!` (invalidates sessions + delete)
- `invalidateApiKeySessions(id: ID!): Int!`

Server guards: rotate/revoke/invalidate allowed only for active keys.

Header auth as middleware
- New ApiKeyHeaderAuthMiddleware validates Authorization: ApiKey <KEY>, reuses/creates a persisted session linked to apiKeyId, and sets the session token for the request.
Unified strategy for the authenticate pipeline
- New ApiKeyAuthenticationStrategy (unified for Admin/Shop); enforces scope by ctx.apiType.
- Legacy admin/shop‑specific strategies are removed.

Session TTL policy
- Default: API‑key sessions use the key’s remaining lifetime (cap to expiresAt).
- Overrides: authOptions.apiKey.admin.sessionDuration, authOptions.apiKey.shop.sessionDuration (still capped by key expiry).
- Refresh clamped to never extend beyond key expiry.

Config updates
- authOptions.apiKey now:
- `generationStrategy?: ApiKeyGenerationStrategy`
- `admin?: { enabled?: boolean; sessionDuration?: string|number }` (default enabled)
- `shop?: { enabled?: boolean; sessionDuration?: string|number }` (default disabled)

Middleware registers only when scope is enabled and ApiKeyAuthenticationStrategy is present in that API’s strategies.

Session linkage and helpers
- Persisted sessions link to apiKeyId. Added findActiveSessionByApiKeyId() for reuse and invalidateSessionsByApiKeyId() for targeted invalidation.

### Dashboard

- New API Keys UI (React Dashboard)
- List with search/sort/pagination; Status faceted filter (active/revoked).
- Create (required expiry; no “no expiration”), Rotate (shows new raw key), Revoke, Invalidate Sessions, Delete; all with confirmation dialogs.
- Toasts for success and error.
- Actions disabled for non‑active keys.
- Standard actions column; Delete uses built-in confirm pattern.

### Docs

- New guide at docs/docs/guides/core-concepts/auth/api-keys.md
- Header auth + strategy requirement; admin/shop enable flags.
- Default TTL (key expiry) + per‑API overrides.
- Security best practices.

# Breaking changes
No breaking changes.

 # Checklist
  
- [x] Clear title
- [x] Self‑review (typos/dead code)
- [ ] Tests updated where relevant
- [x] Docs updated (API keys guide, config, security)

### DIFF
--- a/packages/core/src/api/middleware/api-key-header-auth.middleware.ts
+++ b/packages/core/src/api/middleware/api-key-header-auth.middleware.ts
@@ -25,8 +25,18 @@ export class ApiKeyHeaderAuthMiddleware implements NestMiddleware {
         try {
             const authHeader = req.get('Authorization');
             if (!authHeader) return next();
-            const apiKeyMatch = authHeader.trim().match(/^apikey\s+(.+)$/i);
-            if (!apiKeyMatch) return next();
+            // Parse scheme and token without complex regex to avoid potential backtracking DoS.
+            // Accept: 'ApiKey <token>' (case-insensitive).
+            let token: string | undefined;
+            const trimmed = authHeader.trim();
+            if (trimmed.length >= 6 && trimmed.slice(0, 6).toLowerCase() === 'apikey') {
+                const rest = trimmed.slice(6);
+                // Require at least one whitespace after scheme
+                if (/^\s+/.test(rest)) {
+                    token = rest.trim();
+                }
+            }
+            if (!token) return next();
 
             const path = ((req.baseUrl ?? '') + (req.path ?? '')).toLowerCase();
             const adminBase = `/${this.configService.apiOptions.adminApiPath}`.toLowerCase();
@@ -50,7 +60,7 @@ export class ApiKeyHeaderAuthMiddleware implements NestMiddleware {
             if (!apiType) return next();
 
             const tmpCtx = await this.requestContextService.create({ apiType, req });
-            const result = await this.apiKeyService.verifyRawKey(tmpCtx, apiKeyMatch[1]);
+            const result = await this.apiKeyService.verifyRawKey(tmpCtx, token);
             if (!result) return next();
             if (apiType === 'admin' && result.apiKey.scope !== 'admin') return next();
             if (apiType === 'shop' && result.apiKey.scope !== 'shop') return next();
--- a/packages/core/src/service/services/session.service.ts
+++ b/packages/core/src/service/services/session.service.ts
@@ -394,7 +394,8 @@ export class SessionService implements EntitySubscriberInterface, OnApplicationB
             .andWhere('session.invalidated = false')
             .andWhere('session.expires > :now', { now: new Date() })
             .orderBy('session.updatedAt', 'DESC');
-        return qb.getOne() ?? undefined;
+        const found = await qb.getOne();
+        return found ?? undefined;
     }
 
     /**
--- a/packages/dev-server/dev-config.ts
+++ b/packages/dev-server/dev-config.ts
@@ -12,7 +12,7 @@ import {
     SettingsStoreScopes,
     VendureConfig,
     NativeAuthenticationStrategy,
-    ApiKeyAdminAuthStrategy,
+    ApiKeyAuthenticationStrategy,
 } from '@vendure/core';
 import { DashboardPlugin } from '@vendure/dashboard/plugin';
 import { defaultEmailHandlers, EmailPlugin, FileBasedTemplateLoader } from '@vendure/email-plugin';
@@ -56,7 +56,7 @@ export const devConfig: VendureConfig = {
         cookieOptions: {
             secret: 'abc',
         },
-        adminAuthenticationStrategy: [new NativeAuthenticationStrategy(), new ApiKeyAdminAuthStrategy()],
+        adminAuthenticationStrategy: [new NativeAuthenticationStrategy(), new ApiKeyAuthenticationStrategy()],
     },
     dbConnectionOptions: {
         synchronize: false,


mohdbk:
**Note:** A draft PR was opened for this issue as well (#3815). Since we both worked on this at the same time, @DanielBiegler, I’d appreciate your feedback on these changes and whether you see anything that needs adjustment.

@michaelbromley, please let me know if this overall approach looks good.

DanielBiegler:
Cool, thanks for taking the time to work on this, I'd love to have API keys finally! I'll be taking a look shortly when I get some time. :+1:

mohdbk:
> This current approach fundamentally misses how API keys generally are supposed to work.
> 
> ## 1. One of the key requirements is not having to login when scripting or running one-off requests via `curl`.
> With this approach scripts still need to call the mutation `authenticate` and retrieve a new "short-lived" session, not meeting the requirement and making it impossible to run one-off HTTP requests.
> 
> Look at other services like payment providers like Stripe, Mollie, .. where you just provide the key either as user or bearer token:
> 
> ```shell
> curl -X GET https://api.stripe.com/v1/balance -u "<YOUR_API_KEY_HERE>:"
> curl -X GET https://api.mollie.com/v2/balances/primary -H "Authorization: Bearer <YOUR_API_KEY_HERE>"
> ```
> 
> With the current approach longer running scripts still need to worry about session invalidation happening in between requests. This PR provides a new way to login administrators, namely, without a username+password combo. But thats not why we need API Keys to begin with, API keys must enable authorization without needing to establish a new session. This PR does not enable this.
> 
> ## 2. API keys are not channel aware and its service is vulnerable to privilege escalation and hence critically dangerous
> Channel awareness is critical because in multi-vendor setups any single random admin can with the current approach abuse data they shouldnt be able to.
> 
> Example: A must have feature for multi-vendor setups is that other vendors on their own channels can create API keys for themselves and their own sub-administrators. So it's reasonable to assume to give them the introduced `ServiceAccount` permissions but this is extremely dangerous. There are no privilege checks, meaning any admin with Read permissions can read all Apikey entities of ANY admin of this vendure instance regardless of channel, if they know/guess the Admin-ID.
> 
> Any admin with the Create permission can create API keys for ANY other admin because it blindly accepts the ID of the input. Worst case example would be if a hacker can see the superadmin ID once, they can attach a new api key to this superadmin, authenticate and now they control the entire vendure instance.
> 
> ## 3. `.isServiceAccount` seems to unnecessarily obfuscate admins on channels for no functional benefit?
> It's only being used for filtering the `administrators`-query and no real functionality and actually obfuscates existing administrators on channels. I don't quite see a need nor benefit of introducing this? In multi-tenant Vendure instances where each Channel is a separate vendor, on first glance it seems weird to hide existing administrators from shop owners, especially since they can't do anything about it due to the hardcoded SuperAdmin permission. If the superadmin wants to do work on those channels via scripting, they have all the permissions anyway, so having a hidden admin means nothing and would only confuse channel-admins seeing references to an un-findable secret admin.
> 
> In Vendure an "Administrator" in itself doesnt really mean much beyond holding some identifying data, because the underlying authentication (auth-method) and authorization (permissions) entirely depends on the User-entity so marking specifically Administrators as "service-accounts" seems unnecessary too because API Keys could, and most likely should, also work for Customer-Users.
> 
> ## 4. Prefixes, validation and key length
> Generally these prefixes exist for services like stripe and mollie because you are not hosting the platform yourself e.g. without separate test keys you literally cant test anything. This isnt the case for Vendure, local and prod are already separate so having those prefixes be part of the data model seems unnecessary.
> 
> I do really like the idea of having different keys for different environments to make sure they can never overlap, so how about simplifying this approach with some type of short "api-key-generation-strategy", then users can make their own to fit their needs exactly, shorter/longer keys with or without prefixes. In prod you could for example default to "DefaultApiKeyGenerationStrategy" and locally you could have a "LocalApiKeyGenerationStrategy" with a `local_` prefix and shorter lengths. Customizability is a core tenet of Vendure.
> 
> The "validation" in `validateRawKey` seems problematic, why read every single api key of the entire vendure instance into memory first to then just compare the hashes via `passwordCipher.check()` one by one. We already have the `rawKey` here, so we can simply hash it locally and look up the hash directly in the DB. There could be thousands of keys so iterating in memory is problematic. Also this isnt really "validating" the key itself, its verifying that the key exists in the DB so I'd rename the function similar to native-auths `verifyUserPassword` to be clearer.
> 
> There are some smaller issues like magic strings, missing & wrong types and tests but those can be reviewed and fixed after the fundamentally important stuff is out of the way.

Agreed with all the points. I'll look into them and see how we can implement them here. Thanks for your valuable input @DanielBiegler.

PR #3828, opened 2025-09-22
feat(core): Add read/write permission support for settings store

dlhck:
## Summary

Adds dedicated read and write permission support to the settings store, allowing fine-grained control over who can read vs write specific settings fields.

### Changes

- **New `RwPermissionDefinition` class**: Creates separate Read and Write permissions (e.g., `ReadDashboardSavedViews`, `WriteDashboardSavedViews`)
- **Enhanced `requiresPermission` property**: Now supports both:
  - Backward compatible: Single permission (applies to both read and write)
  - New granular control: Object with `read` and `write` properties
- **New service methods**: `hasReadPermission()` and `hasWritePermission()` with smart fallback logic
- **GraphQL resolver updates**: Query operations use read permissions, mutations use write permissions
- **Full backward compatibility**: Existing configurations continue to work unchanged

### Usage Examples

```typescript
// Backward compatible - applies to both read and write
requiresPermission: Permission.UpdateSettings

// Separate read/write permissions
requiresPermission: {
  read: Permission.ReadSettings,
  write: Permission.UpdateSettings
}

// Using custom RwPermissionDefinition
const dashboardSavedViews = new RwPermissionDefinition('DashboardSavedViews');
requiresPermission: {
  read: dashboardSavedViews.Read,
  write: dashboardSavedViews.Write
}
```

### Test Coverage

- Comprehensive e2e tests covering all usage patterns
- Tests for backward compatibility
- Tests for the new `RwPermissionDefinition` class
- Validation of permission resolution logic

All existing tests continue to pass, confirming backward compatibility is maintained.

### DIFF
--- a/packages/core/src/config/settings-store/settings-store-types.ts
+++ b/packages/core/src/config/settings-store/settings-store-types.ts
@@ -65,8 +65,38 @@ export interface SettingsStoreFieldConfig {
      * @description
      * Permissions required to access this field. If not specified,
      * basic authentication is required for admin API access.
+     *
+     * Can be either:
+     * - A single permission or array of permissions (applies to both read and write)
+     * - An object with `read` and `write` properties for granular control
+     *
+     * @example
+     * ```ts
+     * // Single permission for both read and write
+     * requiresPermission: Permission.UpdateSettings
+     *
+     * // Separate read and write permissions
+     * requiresPermission: {
+     *   read: Permission.ReadSettings,
+     *   write: Permission.UpdateSettings
+     * }
+     *
+     * // Using custom RwPermissionDefinition
+     * requiresPermission: {
+     *   read: dashboardSavedViews.Read,
+     *   write: dashboardSavedViews.Write
+     * }
+     * ```
+     * @since 3.4.0 - Added support for object with read/write properties
      */
-    requiresPermission?: Array<Permission | string> | Permission | string;
+    requiresPermission?:
+        | Array<Permission | string>
+        | Permission
+        | string
+        | {
+              read?: Array<Permission | string> | Permission | string;
+              write?: Array<Permission | string> | Permission | string;
+          };
 
     /**
      * @description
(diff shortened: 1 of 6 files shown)


packages/core/src/config/settings-store/settings-store-types.ts
```diff
@@ -65,8 +65,38 @@ export interface SettingsStoreFieldConfig {
...
+     *   write: Permission.UpdateSettings
+     * }
+     *
+     * // Using custom RwPermissionDefinition
+     * requiresPermission: {
+     *   read: dashboardSavedViews.Read,
+     *   write: dashboardSavedViews.Write
+     * }
+     * ```
+     * @since 3.4.0 - Added support for object with read/write properties
```
michaelbromley: since 3.5.0

packages/core/src/service/helpers/settings-store/settings-store.service.ts
```diff
@@ -460,20 +460,85 @@ export class SettingsStoreService implements OnModuleInit {
...
+     * Check if the current user has permission to read a field.
+     * @since 3.4.0
+     */
+    hasReadPermission(ctx: RequestContext, key: string): boolean {
+        try {
+            const fieldConfig = this.getFieldConfig(key);
+            const requiredPermissions = fieldConfig.requiresPermission;
+
+            if (requiredPermissions) {
+                if (
```
michaelbromley: I would suggest pulling out this duplicated logic into a helper method, lines 482 - 484 are duplicated on 521-523, and is a little complex so if housed in a well-named function will make it nicer to read & understand.

packages/core/src/service/helpers/settings-store/settings-store.service.ts
```diff
@@ -460,20 +460,85 @@ export class SettingsStoreService implements OnModuleInit {
...
+                    if (readPerms) {
+                        const permissions = Array.isArray(readPerms) ? readPerms : [readPerms];
+                        return ctx.userHasPermissions(permissions as any);
+                    }
+                    // If no read permission specified but write is, fall back to authenticated
+                    if (requiredPermissions.write) {
+                        return ctx.userHasPermissions([Permission.Authenticated]);
+                    }
+                } else {
+                    const permissions = Array.isArray(requiredPermissions)
```
michaelbromley: Also this default case is the same here and in the `hasWritePermissions` so could also be extracted into a method and the just 

```
return this.checkSimplePermissions(ctx, permissions);
```

michaelbromley (approved):

PR #3832, opened 2025-09-23
feat(core, dashboard): Slug strategy and slug input

dlhck:
# Description

Adds a new slug generation strategy to Vendure core and a corresponding slug input component for the dashboard. This feature enables automatic and manual slug generation for entities based on configurable source fields.

## Core Features
- New `SlugStrategy` interface and `DefaultSlugStrategy` implementation for customizable slug generation
- `slugForEntity` GraphQL query to generate unique slugs for any entity
- Support for both translatable and non-translatable slug fields
- Automatic uniqueness checking with entity exclusion for updates

## Dashboard Component  
- **SlugInput**: A smart form component that:
  - Auto-generates slugs when the field is empty and the watched field changes
  - Provides manual regeneration via a refresh button for existing slugs
  - Allows toggling between auto-generation mode and manual editing
  - Automatically resolves whether watched fields are translatable or not
  - Shows loading states during slug generation

## Usage Example
```tsx
<SlugInput
    {...field}
    entityName="Product"
    fieldName="slug"
    watchFieldName="name"
    entityId={productId}
/>
```

# Breaking changes

None

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

- **New Features**
  - Auto-generated, editable slugs in Product, Collection, and Facet forms with toggle and regenerate controls; tied to the name field and respects manual edits.
  - Admin API endpoint for on-demand slug generation used by the dashboard.
  - Configurable slug strategy with a sensible default to customize slug rules.
- **Tests**
  - Extensive end-to-end tests covering slug generation, edge cases, uniqueness, translations, and error handling.

### DIFF
--- a/packages/core/src/config/entity/slug-strategy.ts
+++ b/packages/core/src/config/entity/slug-strategy.ts
@@ -39,7 +39,7 @@ export interface SlugGenerateParams {
  * ```
  *
  * @docsCategory configuration
- * @since 3.x.x
+ * @since 3.5.0
  */
 export interface SlugStrategy extends InjectableStrategy {
     /**
--- a/packages/core/src/service/helpers/entity-slug.service.ts
+++ b/packages/core/src/service/helpers/entity-slug.service.ts
@@ -36,7 +36,7 @@ export interface GenerateSlugFromInputParams {
  * and handling conflicts by appending numbers.
  *
  * @docsCategory services
- * @since 3.x.x
+ * @since 3.5.0
  */
 @Injectable()
 export class EntitySlugService {
--- a/packages/core/src/service/helpers/slug.service.ts
+++ b/packages/core/src/service/helpers/slug.service.ts
@@ -8,7 +8,7 @@ import { ConfigService, SlugGenerateParams } from '../../config';
  * A service that handles slug generation using the configured SlugStrategy.
  *
  * @docsCategory services
- * @since 3.x.x
+ * @since 3.5.0
  */
 @Injectable()
 export class SlugService {
--- a/packages/core/src/config/entity/default-slug-strategy.ts
+++ b/packages/core/src/config/entity/default-slug-strategy.ts
@@ -20,7 +20,7 @@ import { SlugGenerateParams, SlugStrategy } from './slug-strategy';
  * ```
  *
  * @docsCategory configuration
- * @since 3.x.x
+ * @since 3.5.0
  */
 export class DefaultSlugStrategy implements SlugStrategy {
     generate(ctx: RequestContext, params: SlugGenerateParams): string {


packages/core/src/service/helpers/slug.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @description
+ * A service that handles slug generation using the configured SlugStrategy.
+ *
+ * @docsCategory services
+ * @since 3.x.x
+ */
+@Injectable()
+export class SlugService {
```
michaelbromley, line 14: I think this service is a bit redundant - I would just directly use the `ConfigService` inside the `EntitySlugService` rather than splitting into 2 services.

packages/core/src/service/helpers/entity-slug.service.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * @description
+ * A service that handles slug generation for entities, ensuring uniqueness
+ * and handling conflicts by appending numbers.
+ *
+ * @docsCategory services
+ * @since 3.x.x
+ */
+@Injectable()
+export class EntitySlugService {
```
michaelbromley, line 42: This service should go in the `/service/helpers` dir because it's not a CRUD-type service.

packages/core/src/config/entity/slug-strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export interface SlugStrategy extends InjectableStrategy {
+    /**
+     * @description
+     * Generates a slug from the input string.
+     *
+     * @param input The input string to be converted to a slug
+     * @returns A URL-friendly slug string
+     */
+    generate(input: string): string;
```
michaelbromley: I think it would be useful to pass the whole input object (entityName, fieldName, value) to this method - e.g. a strategy might want to alter the algorithm depending on what entity type it is generating a slug for.

Also we should pass `ctx` too because who knows - maybe they want to do something that relies on it, plus it bring it into line with most other strategy methods.

Lastly it should be able to return a `string | Promise<string>` to handle async methods.

packages/core/src/config/entity/slug-strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+ *     return input
+ *       .toLowerCase()
+ *       .replace(/[^a-z0-9]+/g, '-')
+ *       .replace(/^-+|-+$/g, '');
+ *   }
+ * }
+ * ```
+ *
+ * @docsCategory configuration
+ * @since 3.x.x
```
michaelbromley: 3.5.0

packages/core/src/service/helpers/entity-slug.service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        throw new UserInputError(`error.entity-has-no-field`, {
+            entityName,
+            fieldName,
+        });
+    }
+
+    private async fieldValueExists(
+        ctx: RequestContext,
+        repository: any,
```
michaelbromley: No `any`

packages/core/src/service/helpers/entity-slug.service.ts
```diff
@@ -0,0 +1,141 @@
...
+     * @param ctx The request context
+     * @param entityName The name of the entity (base entity, e.g., 'Product', 'Collection')
+     * @param fieldName The name of the field to check for uniqueness (e.g., 'slug', 'code')
+     * @param inputValue The value to generate the slug from
+     * @param entityId Optional ID of the entity being updated
+     * @returns A unique slug string
+     */
+    async generateSlugFromInput(
+        ctx: RequestContext,
+        entityName: string,
```
michaelbromley: Let's make the args into an input object -> more readable from call-sites and easier to extend in future if needed.

packages/core/src/service/helpers/entity-slug.service.ts
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param entityName The base entity name
+     * @param fieldName The field name to find
+     * @returns Object containing entityMetadata and actualEntityName
+     */
+    private findEntityWithField(
+        entityName: string,
+        fieldName: string,
+    ): {
+        entityMetadata: any;
```
michaelbromley: No `any`

packages/core/src/service/helpers/entity-slug.service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        // Check if the field exists on the base entity
+        const baseEntityHasField = entityMetadata.columns.find(col => col.propertyName === fieldName);
+
+        if (baseEntityHasField) {
+            return { entityMetadata, actualEntityName: entityName };
+        }
+
+        // If field doesn't exist on base entity, try the translation entity
+        const translationEntityName = `${entityName}Translation`;
```
michaelbromley: I think there might be a safer way to find a potential translation entity than a simple name-based lookup.

Given the base `entityMetadata`, we can introspect the relations of that type in order to find the translation. 

This would prevent accidental false results due to unfortunate naming collisions.

Update: we have an existing util function that will give you exactly this: https://github.com/vendure-ecommerce/vendure/blob/1579e49350e58549420df722790303a9eb9d8984/packages/core/src/service/helpers/list-query-builder/connection-utils.ts#L12

michaelbromley (approved):

Issue #3845, opened 2025-09-29
[DashBoard] perPage in url is dynamic


oidt:
**Describe the bug**
The rows per page on the overview page for example Products you can change it with 10,20,30,40 and 50.
When you change that the URI is changed.

**To Reproduce**

1. Go to https://demo.vendure.io/admin/products and login
2. Click on Rows per page, you get as example: https://demo.vendure.io/admin/products?sort=&page=1&perPage=30
3. Change the url to https://demo.vendure.io/admin/products?sort=&page=1&perPage=1000 for example
4. 1000 products are loaded instead of max 50 (maybe with the harden plugin it will not work, not sure)

**Expected behavior**
Max of Total 100 for example just as the admin-ui.

**Actual behavior**
Change it by hand.

**Screenshots/Videos**
![image](https://github.com/user-attachments/assets/3c5a8e5f-cef6-457c-997f-ab7da4b92fea)

PR #3873, opened 2025-10-10
fix(dashboard): Limit perPage and page uri changes (#3845)

oidt:
# Description

The rows per page on the overview page for example Products you can change it with 10,20,30,40 and 50.
When you change that the URI is changed. https://github.com/vendure-ecommerce/vendure/issues/3845

This pull request improves the handling and validation of pagination settings across the dashboard, ensuring a consistent and user-friendly experience. The main focus is on centralizing pagination logic, validating user preferences, and syncing page size settings throughout the application.

**Pagination logic and validation improvements:**

- Added a new utility function `validatePerPageValue` in `pagination.ts` to ensure that only allowed per-page values are used, defaulting to `DEFAULT_PER_PAGE` if an invalid value is provided.
- Updated the `DataTablePagination` and `DataTable` components to use `ALLOWED_PER_PAGE_VALUES` and `DEFAULT_PER_PAGE` from constants, and to validate and sync pagination state based on user settings and URL parameters.

**User settings enhancements:**

- Extended the `UserSettings` context and provider to include a global `itemsPerPage` preference, with validation and proper initialization from both local storage and server settings.

**Page and table integration:**

- Updated the `ListPage` component to use the new validated `itemsPerPage` setting, ensuring consistency between URL parameters, user preferences, and table state. The page change handler now updates the global `itemsPerPage` instead of per-table settings.

These changes together ensure that pagination is robust, user preferences are respected and validated, and the UI remains consistent across navigation and reloads.

# Breaking changes

Not that I know of

## Summary by CodeRabbit

- New Features
  - Persistent user preference for "Rows per page" with validated allowed options and a sensible default.

- Bug Fixes
  - Validates and bounds page and per-page values to prevent invalid or out-of-range pagination.
  - Keeps pagination state synchronized across UI, URL, and user settings to avoid mismatches.

- Style
  - More compact "Rows per page" selector and tighter page indicator layout.

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_facets/components/facet-values-table.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_facets/components/facet-values-table.tsx
@@ -3,7 +3,6 @@ import { PaginatedListDataTable } from '@/vdb/components/shared/paginated-list-d
 import { Button } from '@/vdb/components/ui/button.js';
 import { addCustomFields } from '@/vdb/framework/document-introspection/add-custom-fields.js';
 import { graphql } from '@/vdb/graphql/graphql.js';
-import { useUserSettings } from '@/vdb/hooks/use-user-settings.js';
 import { validatePerPageValue } from '@/vdb/utils/pagination.js';
 import { useUserSettings } from '@/vdb/hooks/use-user-settings.js';
 import { Trans } from '@lingui/react/macro';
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hi,
Since you made this PR, I added some extra logic to limit the max items per page that will be fetched.

Do you want to bring this PR up to date with current master? 

One thing I notice is that it looks like you move the perPage from the table-specific setting into a global value. Is that right? What's the idea behind that?

oidt:
I’ll look into that.

The perPage global setting was more my own expected behavior. When I change it on one place to 25 I expected it to be 25 everywhere. I don’t usually spend much time in dashboards like this, so if the per-table setup is the intended behavior, that makes sense. Let me know.

michaelbromley:
Hi,
Let's keep the per-page user setting table-specific I think.

PR #3878, opened 2025-10-15
List elements should be batched

floze:
When the list elements grow larger than 8000 items, Lua obscurely fails with "too many results to unpack".

Found the fix here:
https://github.com/taskforcesh/bullmq/issues/422

## Summary by CodeRabbit

- **Performance**
  - Optimized job retrieval by type to handle very large lists more efficiently via batched operations, reducing memory overhead and improving responsiveness.
- **Bug Fixes**
  - Prevented failures and errors that could occur when querying or merging extremely large job lists, enhancing stability for high-volume workloads.


michaelbromley:
Thanks for this, and sorry it sat so long. The too many results to unpack failure is now fixed on master via #5014, which rewrote get-jobs-by-type.ts to read bounded ranges from the indexed sorted sets rather than copying the full wait list through a temp key — so there’s no unpack call left to overflow. There’s a regression test seeding 9000 waiting jobs covering exactly this case. Closing as fixed, but the bullmq issue link you provided was useful in confirming the root cause.

PR #3907, opened 2025-10-26
fix(dashboard): Properly convert price to minor units before saving product variants

### LINKED ISSUE #3905: React Dashboard: When creating product variants - price gets divided by 100 after submitting the form
**Describe the bug**

When creating product variants, the value stored in the database after form submission is divided by 100.
For example when I create a product variant with the price of 300 CZK, it becomes 3 CZK instead.

This is not specific to the CZK currency, I tested this issue with USD and EUR and it behaves the same.

**To Reproduce**

1. Go to Catalog > Products
2. Create a new product
3. Click on "Create Variants"
4. Fill in price
5. Submit the form
6. See the incorrect price in the variants listing - divided by 100

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.

**Screenshots/Videos**

After filling 100 CZK as the product variant price in the form:

Image

When submitted, a value of 1 CZK is stored in the database.

(The incorrect currency displayed in the listing has been reported as issue #3901)

Image

- Nodejs version: 25.0.0

**Workaround**

Multiply the value by 100 when creating product variants

jirkavrba:
# Description

Fixes #3905. Prices are not multiplied by the scaling factor when creating product variants from the React dashboard.

# Breaking changes

This is not a breaking change.

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Enhancements**
  * Product variant price input now uses the active channel's currency and a dedicated money input control for numeric handling.
* **Behavioral change**
  * When the active channel has no configured currency, the price input may show an unspecified currency (previous default of USD removed).


michaelbromley:
Hi! Thanks for the PR!

The better way to handle this would be to use the MoneyInput component like you see [here](https://github.com/vendure-ecommerce/vendure/blob/0320150b80b24474580c5ac740a56fe4d67dbe0f/packages/dashboard/src/app/routes/_authenticated/_product-variants/product-variants_.$id.tsx#L283) - this performs the conversion internally.

jirkavrba:
I see, thanks. I will update the PR later today with the MoneyInput component :+1:

packages/dashboard/src/app/routes/_authenticated/_products/components/create-product-variants-dialog.tsx
```diff
@@ -32,6 +33,7 @@ export function CreateProductVariantsDialog({
     onSuccess?: () => void;
 }) {
     const { activeChannel } = useChannel();
+    const { toMinorUnits } = useLocalFormat();
```
biggamesmallworld:
```suggestion
```
this can be removed as it is unused later in the component

packages/dashboard/src/app/routes/_authenticated/_products/components/create-product-variants.tsx
```diff
@@ -262,16 +265,12 @@ export function CreateProductVariants({
...
-                                                                    {currencyCode}
-                                                                </span>
-                                                                <Input
+                                                                <MoneyInput
                                                                     {...field}
-                                                                    className="pl-12"
-                                                                    placeholder="0.00"
+                                                                    value={Number(field.value) || 0}
+                                                                    onChange={value => field.onChange(value.toString())}
+                                                                    currency={activeChannel?.defaultCurrencyCode ?? 'USD'}
```
biggamesmallworld:
```suggestion
currency={activeChannel?.defaultCurrencyCode}                                       
```

The money input itself has a fallback to the `activeChannel`'s default currency. I think we can either remove the USD string or remove this line entirely so that we don't need to fall back to the USD string in multiple places. I believe that by default USD is the channel's default currency from looking through the codebase, and it might be safer to rely on other components to handle this. What do you think @michaelbromley?

jirkavrba:
Thanks for the review, I've addressed the currency fallback and removed the unused hook/
import

PR #3919, opened 2025-10-29
fix(dashboard): Fix healthcheck view when api endpoint is auto

SiebelsTim:
# Description

When the API endpoint is configured to be `'auto'` (which is the default), the healthcheck view reports that some resources are down.
This is due to the fact that the dashboard tries to request `/dashboard/auto/health`, where `auto`  is not the host but rather a special configuration action indicating that it should request on the same domain.

# Breaking changes

n/a

# Screenshots

Screenshot from 2025-10-29 15-23-51

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Bug Fixes**
  * Health checks now reliably resolve the API base URL, correctly handling automatic host detection by falling back to the current page location.
* **Chores**
  * Centralized API base URL resolution so all dashboard API calls use the same derived base URL.

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_system/healthchecks.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_system/healthchecks.tsx
@@ -1,12 +1,12 @@
 import { Button } from '@/vdb/components/ui/button.js';
 import { Card, CardContent, CardHeader, CardTitle } from '@/vdb/components/ui/card.js';
 import { Page, PageActionBar, PageTitle } from '@/vdb/framework/layout-engine/page-layout.js';
+import { getApiBaseUrl } from '@/vdb/utils/config-utils.js';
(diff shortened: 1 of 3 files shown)


dlhck (changes requested):
I think we should introduce a utility function that builds the `schemeAndHost` as we have the same logic in here too: https://github.com/vendure-ecommerce/vendure/blob/715b905beaa41e333b9da6b8b50becbe54c8c8f9/packages/dashboard/src/lib/graphql/api.ts#L11 

Can you please do that and implement the utility function in both places?

SiebelsTim:
I tried to add a test, however I got stuck trying to mock the module `virtual:vendure-ui-config` multiple times in a test. Since it is an object, I couldn't use `vi.fn` to replace the return value each time how I would do when using a function.

dlhck (approved):

Issue #3925, opened 2025-10-30
Allow specifying encoding options for AssetStorageStrategy


pujux:
**Is your feature request related to a problem?**

In our current project, we have found that especially with the `LocalAssetStorageStrategy` implementation, we are unable to specify an encoding option for the `AssetStorageStrategy.readFileToStream` method and it always uses `"binary"` when calling `fs.createReadStream()`

Of course it's the same with `AssetStorageStrategy.writeFileFromStream`.

**Describe the solution you'd like**
I would like to update the `AssetStorageStrategy` interface to allow implementations to specify an option for users to choose an encoding without enforcing a specific encoding. While doing it I would also update the `LocalAssetStorageStrategy` implementation of the interface to allow for specifying the encoding.

To allow for backwards compatibility, not specifying this option could still result in the use of "binary", since `fs.createReadStream()` defaults to `utf8` encoding, specifying `uft8` would be the same as passing nothing to `fs.createReadStream()`.

**Describe alternatives you've considered**
Until that is in place, we will create our own implementation of the `AssetStorageStrategy` interface which doesn't have that limitation

**Additional context**
https://github.com/vendure-ecommerce/vendure/blob/7274b230c5d28f480aeabe810e3f36d24f17cb13/packages/asset-server-plugin/src/config/local-asset-storage-strategy.ts#L31

https://github.com/vendure-ecommerce/vendure/blob/7274b230c5d28f480aeabe810e3f36d24f17cb13/packages/asset-server-plugin/src/config/local-asset-storage-strategy.ts#L59

dlhck:
This feature request has been moved to GitHub Discussions for community voting.
Please upvote the discussion if you would like to see this implemented: https://github.com/vendurehq/vendure/discussions/4224

PR #3926, opened 2025-10-30
feat(asset-server-plugin): Allow specifying encoding for AssetStorageStrategy

pujux:
# Description

Updates `AssetStorageStrategy` and it's implementation `LocalAssetStorageStrategy` to allow passing an additional parameter `encoding` which accepts any option valid for https://nodejs.org/api/buffer.html#class-buffer (`BufferEncoding` type)

Passing null will result in `undefined` being passed meaning no specified encoding will be used.

The issue for this PR is: #3925 

# Breaking changes

No breaking changes as not specifying anything will result in the current behavior 

# Checklist

📌 Always:

## Summary by CodeRabbit

* **New Features**
  * Asset storage now accepts an optional encoding when reading or writing streamed files, allowing explicit text/binary handling without changing return behavior.
* **Documentation**
  * Updated docs to describe the new encoding parameter and note that passing null disables specific encoding.


michaelbromley:
Thanks!

PR #3865, opened 2025-10-06
fix(dashboard): Render multiple blocks per anchor

brett-rand:
## Summary
  - Fixes vendure-ecommerce/vendure#3848 by collecting dashboard extension blocks per anchor so every before/after/replace registration renders in order.
  - Reuses the processed block list on mobile so extensions appear on narrow viewports.
  - Adds focused Vitest coverage for multi-block insertion, replacement, and mobile behaviour.
  - Extends the dashboard ESLint ignore list to cover generated `.temp` output and the Vite `fake_node_modules` fixture so `npm run lint --workspace packages/dashboard` runs cleanly.

## Breaking changes
  None.

  ## Testing
  - `npm run lint --workspace packages/dashboard`
  - `npm run test --workspace packages/dashboard -- --run src/lib/framework/layout-engine/page-layout.spec.tsx`
  - Manual: Ran the dev server with a small dashboard extension (`DashboardMultiBlockPlugin`) and confirmed stacked blocks render in the Customers list. (IMPORTANT: blocks currently appear beneath the table because the `list-table`
  anchor is full-width; a follow-up PR will introduce additional anchors so header placement is possible.)

 ## Screenshots
  N/A

  ## Checklist
  - [x] Tests pass
  - [ ] Documentation updated (N/A for this change)

  ## Follow-up
  - Introduce named anchors for page title/action bar/filter regions to support richer layouts.
  - Add explicit “remove” semantics for page blocks.
  - Update documentation and provide a sample plugin once the follow-up work lands.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for PageLayout on desktop and mobile.
  * Verifies ordering of extension blocks (before/after), replacement of original blocks, and preservation of original placement on mobile.
  * Improves reliability of the layout engine by validating multiple extension scenarios and rendering outcomes.
  * Ensures clean test isolation with proper setup and teardown.
  * No user-facing changes or public API modifications.

PR #3937, opened 2025-11-03
fix(dashboard): only first custom page block renders on same anchor

### LINKED ISSUE #3848: React Dashboard: Only first custom page block renders when multiple blocks target the same anchor
**Describe the bug**
When registering more than one custom dashboard page block against the same anchor (e.g. the list-table block on a list page), only the first block renders. Additional blocks that target the same anchor do not appear, which prevents stacking multiple UI components on the same page area. This behavior contradicts the documented guidance to reuse existing block IDs discovered through Developer Mode.

**To Reproduce**
1. Enable Developer Mode in the new React dashboard and open Customers so you can discover the list-table anchor.
2. Register two page blocks in an extension, both targeting the same anchor (`plugin-name/dashboard/index.tsx`):
```
     defineDashboardExtension({
       pageBlocks: [
         {
           id: 'customer-acquisition-metrics',
           location: {
             pageId: 'customers-list',
             column: 'main',
             position: { blockId: 'list-table', order: 'before' },
           },
           component: ({ context }) => <AcquisitionMetrics context={context} />,
         },
         {
           id: 'customer-retention-metrics',
           location: {
             pageId: 'customers-list',
             column: 'main',
             position: { blockId: 'list-table', order: 'before' },
           },
           component: ({ context }) => <RetentionMetrics context={context} />,
         },
       ],
     });
```
3. Restart the server and dashboard
4. Navigate to Admin → Customers. Only one of the two custom blocks appears above the list.

**Expected behavior**
Both blocks should render, respecting their declared order (before/after/replace), so that the user can view multiple complementary widgets above the customers list. In the example provided here, AcquisitionMetrics and RetentionMetrics components should both show before (on top of) the customers-list table.

**Actual behavior**
Only the first registered block renders. All subsequent blocks targeting the same blockId are not shown. Additionally, on narrow viewports (< 769 px), the mobile branch of PageLayout returns only the original children, so no extension blocks render at all.

**Screenshots/Videos**
N/A -- the second block never renders (customer-retention-metrics), only the first one renders (customer-acquisition-metrics)

**Error logs**
Javascript console in Chrome Devtools shows ```{pageId:`` 'customers-list', form: undefined, entity: undefined}```

- @vendure/core version: 3.4.2 (also reproducible on current minor)
- Nodejs version: 20.19.0
- Database: SQLite (should be storage agnostic)
- Operating System: macOS 26.0.1 (issue is framework-level)
- Browser (if applicable): Chrome 141 and Safari 26.0.1
- Package manager (npm/yarn/pnpm): npm 11.6.0

**Configuration**
Relevant dashboard extension snippet shared previously (see To Reproduce)

**Minimal reproduction**
- Start from the React dashboard example or any Vendure project using the new dashboard.
- Add the extension above (or any two blocks sharing an anchor such as list-table on customers/products pages).
- Run the dashboard and open the relevant page: only the first block renders.

**Workaround**
- [WORKAROUND] Combine multiple widgets into a single page block, losing composability of UI components.
- [PROPOSED FIX] Update packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx in Vendure to iterate extensionBlocks.filter(...) rather than .find(...), so multiple matches are inserted. Consider also adjusting the mobile branch to render the merged array.

**Additional context**
- The documentation encourages targeting existing blockIds via the Developer Mode inspector, so the proposed fix (allow all registered blocks per anchor) aligns with the documented extension model.
- The issue stems from PageLayout using extensionBlocks.find(...) (minor branch line ~237) which returns only the first match. Replacing it with an iteration over all matches would preserve ordering (before/after/replace) and unlock richer page customizations.
- It would also be great to keep extensions visible on small viewports, since the current mobile branch skips them entirely.

DeltaSAMP:
# Description

Fixes #3848

Addresses these 3 Issues:

- Fixes adding multiple custom pageBlocks with the same "anchor" element under the position option would only show the first element
- Order of the blocks did not reflect how they were defined in the dashboardExtensions
- Custom pageBlocks were not shown in the mobile version

also added the test created in this pr #3865

# Breaking changes

No

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive unit test suite for page layout covering block registration, ordering, replacement semantics, and mobile rendering.

* **Bug Fixes**
  * Improved extension handling to support multiple extensions per block with correct before/replace/after ordering and conditional rendering.
  * Enhanced desktop and mobile layout rendering to preserve correct block sequence and insertion behavior.

### DIFF
--- a/docs/docs/reference/dashboard/list-views/bulk-actions.md
+++ b/docs/docs/reference/dashboard/list-views/bulk-actions.md
@@ -71,14 +71,15 @@ interface DataTableBulkActionItemProps {
 
 ## DataTableBulkActionItem
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table-bulk-action-item.tsx" sourceLine="66" packageName="@vendure/dashboard" since="3.4.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table-bulk-action-item.tsx" sourceLine="67" packageName="@vendure/dashboard" since="3.4.0" />
 
 A component that should be used to implement any bulk actions for list pages & data tables.
 
 *Example*
 
 ```tsx
-import { DataTableBulkActionItem, Trans } from '@vendure/dashboard';
+import { Trans } from '@lingui/react/macro';
+import { DataTableBulkActionItem, BulkActionComponent } from '@vendure/dashboard';
 import { Check } from 'lucide-react';
 
 export const MyBulkAction: BulkActionComponent<any> = ({ selection, table }) => {
--- a/packages/dashboard/src/app/routes/_authenticated/_customers/components/customer-address-form.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_customers/components/customer-address-form.tsx
@@ -1,3 +1,4 @@
+import { CustomFieldsForm } from '@/vdb/components/shared/custom-fields-form.js';
 import { Button } from '@/vdb/components/ui/button.js';
 import { Checkbox } from '@/vdb/components/ui/checkbox.js';
 import {
@@ -284,6 +285,8 @@ export function CustomerAddressForm({ address, onSubmit, onCancel }: Readonly<Cu
                     />
                 </div>
 
+                {/* Custom Fields */}
+                <CustomFieldsForm entityType="Address" control={form.control} />
                 {/* Default Address Checkboxes */}
                 <div className="grid grid-cols-1 md:grid-cols-2 gap-4 pt-2">
                     <FormField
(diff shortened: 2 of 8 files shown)


packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
```diff
@@ -234,59 +234,100 @@ export function PageLayout({ children, className }: Readonly<PageLayoutProps>) {
...
-                if (typeof extensionBlock?.shouldRender === 'function') {
-                    extensionBlockShouldRender = extensionBlock.shouldRender(page);
+            // Get all extension blocks with the same position blockId
+            const matchingExtensionBlocks = extensionBlocks.filter(
+                block => block.location.position.blockId === blockId,
+            );
+
+            // sort the blocks to make sure we have the correct order
+            const arrangedExtensionBlocks = matchingExtensionBlocks.sort((a, b) => {
+                const orderPriority = { 'before': 1, 'replace': 2, 'after': 3 };
```
michaelbromley: This const is being defined for every invokation of sort inside filter, so it's n^2. Move it up to the top level scope. A minor optimization but worth doing.

michaelbromley:
Nice work Alex!

PR #3963, opened 2025-11-12
fix: Product graphql query - fixed lang approach on method findOneBySlug

grolmus:
## Fix: Product slug lookup returning wrong product when same slug exists in different languages

  ### Problem
  When two different products have the same slug in different languages (e.g., Product A has slug "phone" in EN, Product B has slug "phone" in DE), the `findOneBySlug()` method would return the wrong product based on the first match found, ignoring the requested language context.

  ### Solution
  Added language code filter to the translation subquery in `ProductService.findOneBySlug()` that:
  - Searches for products with the slug in either the requested language OR the default channel language (preserves fallback)
  - Uses existing CASE statement to prioritize results by language match (requested language = priority 2, default = priority 1)
  - Returns the product with highest priority via `getRawOne()`

  ### Changes
  - `packages/core/src/service/services/product.service.ts` (lines 202-209)

  ### Testing
  All existing e2e tests pass (141 tests):
  - ✅ product.e2e-spec.ts (99 tests)
  - ✅ product-channel.e2e-spec.ts (20 tests)
  - ✅ shop-catalog.e2e-spec.ts (18 tests)
  - ✅ translations.e2e-spec.ts (4 tests)

  No regressions detected. Language fallback mechanism remains intact.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved product lookup to properly handle language preferences by checking the requested language first, then falling back to the channel's default language.


michaelbromley:
Unfortunately this is breaking some of the multi-language product tests.

PR #3967, opened 2025-11-13
fix(dashboard): cannot save global languages when selecting norwegian

### LINKED ISSUE #3943: React Dashboard: Cannot save Global Languages when selecting Norwegian; GraphQL error: value "no" not in LanguageCode enum
**Description**

While configuring languages in the Admin UI, attempting to add Norwegian to Global Languages results in a GraphQL validation error when saving. The UI appears to send the ISO code "no", which is not a valid value in Vendure’s LanguageCode enum, and the mutation rejects the update.

**Steps to Reproduce**

1. Open Admin UI.
2. Go to Settings -> Channels -> Create new channel (or edit an existing one).
3. In the “Manage Languages” modal, under “Select Available Languages,” choose “Norwegian” among other languages.
4. Click “Save Changes.”

**Expected Behaviour**

- The selection should save successfully.
- Norwegian should be added to the list of global languages.
- No GraphQL errors should be thrown.

**Actual Behaviour**

- A toast/error banner appears: “Failed to update global settings: GraphQL Request Error: Variable '$input' got invalid value 'no' at 'input.availableLanguages[12]'; Value 'no' does not exist in 'LanguageCode' enum. Did you mean the enum value 'bo', 'co', 'eo', 'fo', or 'ko'?”
- Changes are not saved.

**Screenshots/Videos**

All languages selected:

Image

All languages, except Norwegian selected:

Image

**Error Logs**

N/A

- Nodejs version: 20.19.4
- Database (mysql/postgres etc): Postgres 17
- Operating System (Windows/macOS/Linux): Linux (Kubuntu 25.10)

**Configuration**

N/A

**Minimal Reproduction**

N/A

**Workaround**

None

**Additional Context**

N/A

DeltaSAMP:
# Description

Fixes #3943
removes "no" (Norwegian) and adds "nb" (Norwegian Bokmål) and "nn" (Norwegian Nynorsk) to the `ALL_LANGUAGE_CODES` array used in the `manage-languages-dialog`

+also fixes a small nested scroll issue where the dropdown select was not scrollable in the dialog

Also maybe to avoid some confusion in the future a possible improvement would be to use the same languages codes that are used in the global settings located [here](https://github.com/vendure-ecommerce/vendure/blob/c68120a4e7f574fa96116c9ccaf4edeca1228619/packages/dashboard/src/app/routes/_authenticated/_global-settings/utils/global-languages.ts#L4)

Since those two selects look the same and (as far as I can tell) are intended for the same use with the only difference being that in the `manage-languages-dialog` there are less options to choose from

# Breaking changes

no

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **New Features**
  * Added support for Norwegian (Bokmål) and Norwegian (Nynorsk) language options.

* **Bug Fixes**
  * Improved multi-select component to prevent unintended scrolling when using the mouse wheel.


michaelbromley:
Hi Alex, your suggestion is so good I am wondering if you have time to actually make that change?

We should have a single global const containing the language codes. It must contain the exact same values as the server side LanguageCode enum. The only reason we don't directly use that enum is I think there were issues trying to import it from the common package.

But having just 1 const in the dashboard is also fine.

michaelbromley:
> We should have a single global const containing the language codes. It must contain the exact same values as the server side LanguageCode enum. The only reason we don't directly use that enum is I think there were issues trying to import it from the common package.

We decided this will follow in another PR by Alex.

PR #4017, opened 2025-12-02
 fix(dashboard): Support localized decimal separator in tax rate input

starry-osean:
## Summary

Fixes an issue in the Admin UI where tax rates could not be entered as decimal values when using certain locales (e.g. German).

### What was the problem?

In the new Admin UI, the tax rate field only accepted values entered with a dot (`.`) as the decimal separator via an `<input type="number">`. This caused the following issues:

- With **German** UI language:
  - Entering `5.5` resulted in: “Please enter a valid value” (not localized).
  - Entering `5,5` resulted in: “Expected number, received nan” (also not localized).
- As a result, it was impossible to save a tax rate with a decimal value.

### What has been changed?

1. **[PercentageSuffixInputComponent](cci:2://file:///e:/kai/vendure/packages/admin-ui/src/lib/core/src/shared/components/affixed-input/percentage-suffix-input.component.ts:8:0-75:1)**
   - Changed the internal input type from `number` to `text` to allow localized characters (e.g. comma).
   - Added automatic detection of the current locale’s decimal separator using `Intl.NumberFormat`.
   - Implemented parsing & formatting logic:
     - User input is parsed using the locale-specific decimal separator.
     - The internal form control value is always stored as a proper `number`.
     - On blur / initialization the value is formatted back using the current locale.

2. **Tax rate detail form**
   - Replaced the raw `<input type="number">` field with the enhanced `<vdr-percentage-suffix-input>` component for the tax rate (`value`) field.

### Expected behaviour after this change

- With **English** locale:
  - `5.5` is accepted and saved as a decimal tax rate.
- With **German** locale:
  - `5,5` is accepted and saved as the same decimal tax rate.
- No validation errors such as “Please enter a valid value” or “Expected number, received nan” should occur for valid localized decimal inputs.
- Error messages remain properly localizable via the existing i18n system.

### Testing

- Set the Admin UI language to **German**.
- Navigate to `Settings → Tax Rates`.
- Create or edit a tax rate.
- Try entering:
  - `5.5` and `5,5` as the rate.
- Confirm that:
  - The form allows saving the tax rate.
  - The stored value is correct.
  - No erroneous validation messages are shown.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes issues with tax rate entry—prevents invalid values/NaN and keeps displayed value synchronized with saved data on edit and blur.

* **Improvements**
  * Locale-aware numeric input: supports local decimal separators during typing and parsing.
  * Supports configurable prefix/suffix (e.g., “%”) and preserves step/min/max behavior for smoother, predictable rate entry.


starry-osean:

starry-osean:
@michaelbromley .Thanks for the review! I’ve addressed both concerns:
1. Removed the Chinese comments; everything is now in English.
2.Implemented localized decimal handling in the shared `NumberInput` (text input with locale-aware `.`/`,` parsing), so any page using `NumberInput` benefits without per-page special code. The tax rate detail view now just uses `NumberInput` with a `%` suffix.

michaelbromley:
@starry-osean thanks for quickly responding to the feedback! @gabriellbui from our team will be reviewing and testing this in the coming week 👍

gabriellbui:
Hi @starry-osean! Thanks for the PR! I tested it locally and it fixes the original issue.

While reviewing it, I ran into a few concerns around the implementation of the number input (higher cognitive load, allowing arbitrary characters, etc ...). I’ve opened a separate [PR](https://github.com/vendure-ecommerce/vendure/pull/4052) to address those points with a simpler approach. I’d prefer to move forward with that solution instead.

Still great job on this and thanks again for the effort!

Issue #3897, opened 2025-10-24
Shop API mutation to update Order currency code


dlhck:
Need mutation to update currency code on Order and trigger recalculation of order items and dependencies.

Current limitation: currency code update only possible via state-changing operations (e.g. adding items).

Proposed mutation: `setCurrencyCodeForOrder` (or similar).

ahmed-e-elhor:
Hi @dlhck 
I would love to work on this issue.

Can you please share more details about the expected behavior of the new mutation (what is the expected result), and can you please clarify what you mean by `dependencies` mentioned in the description?

Thanks.

dlhck:
@ahmed-e-elhor the mutation should basically update the currency code and run all the calculations of the active order again - similar to what happens when you add an item to the cart.

aelhor:
Hello @dlhck ,
Thanks for the clarification. I have worked on the issue and pushed PR  #4033 

Looking forward to your review

_This is my personal GitHub account here and on the PR_ 

Thanks

michaelbromley:
Closed by #4466

PR #4033, opened 2025-12-08
Feet(core)  shop api mutation to update order currency code

aelhor:
# Description

Added new mutation to update order currency and trigger price recalculations
Added e2e tests to cover the new added mutation

Fixes issue #3897

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **New Features**
  * Add mutation to set an order's currency; active order pricing is recalculated in the new currency.
  * Order history gains a new entry type for currency changes (ORDER_CURRENCY_UPDATED).
  * Currency changes are validated against the current channel; invalid requests return a clear error.

* **Tests**
  * Added end-to-end tests for updating order currency, channel currency validation, and resulting order/line updates.

### DIFF
--- a/packages/core/src/api/resolvers/shop/shop-order.resolver.ts
+++ b/packages/core/src/api/resolvers/shop/shop-order.resolver.ts
@@ -28,7 +28,7 @@ import {
     UpdateMultipleOrderItemsResult,
     UpdateOrderItemsResult,
 } from '@vendure/common/lib/generated-shop-types';
-import { CurrencyCode, QueryCountriesArgs } from '@vendure/common/lib/generated-types';
+import { QueryCountriesArgs } from '@vendure/common/lib/generated-types';
 import { unique } from '@vendure/common/lib/unique';
 
 import { ErrorResultUnion, isGraphQlErrorResult } from '../../../common/error/error-result';
@@ -389,12 +389,7 @@ export class ShopOrderResolver {
             args[ACTIVE_ORDER_INPUT_FIELD_NAME],
             true,
         );
-        return this.orderService.setCurrencyCodeForOrder(
-            ctx,
-            order.id,
-            args.currencyCode as CurrencyCode,
-            relations,
-        );
+        return this.orderService.updateOrderCurrency(ctx, order.id, args.currencyCode, relations);
     }
 
     @Transaction()
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -586,16 +586,6 @@ export class OrderService {
             return result.order;
         }
     }
-    async setCurrencyCodeForOrder(
-        ctx: RequestContext,
-        orderId: ID,
-        currencyCode: CurrencyCode,
-        relations?: RelationPaths<Order>,
-    ): Promise<ErrorResultUnion<UpdateOrderItemsResult, Order>> {
-        const result = await this.updateOrderCurrency(ctx, orderId, currencyCode, relations);
-
-        return result;
-    }
 
     async updateOrderCurrency(
         ctx: RequestContext,


aelhor:

packages/core/e2e/order.e2e-spec.ts
```diff
@@ -146,7 +150,19 @@ describe('Orders resolver', () => {
...
+                // channels: [
+                //     {
+                //         token: 'default',
+                //         defaultLanguageCode: LanguageCode.en,
+                //         // Set the base currency ()
+                //         currencyCode: CurrencyCode.USD,
+                //         availableLanguageCodes: [LanguageCode.en],
+                //         availableCurrencyCodes: [CurrencyCode.USD, CurrencyCode.EGP],
+                //     },
+                // ],
```
DanielBiegler: lil leftover thingy? :)

packages/core/src/api/resolvers/shop/shop-order.resolver.ts
```diff
@@ -374,6 +375,28 @@ export class ShopOrderResolver {
...
+            ctx,
+            args[ACTIVE_ORDER_INPUT_FIELD_NAME],
+            true,
+        );
+        return this.orderService.setCurrencyCodeForOrder(
+            ctx,
+            order.id,
+            args.currencyCode as CurrencyCode,
+            relations,
+        );
```
DanielBiegler: I'm aware that you're following the current pattern of passing in the orderId instead of the order object because thats how the other methods work, and it's good to follow conventions, but I want to pose a general question to the vendure team here that we could get rid of database queries by replacing the IDs with a real order objects. Food for thought.

packages/core/e2e/order.e2e-spec.ts
aelhor: Nice catch, sorry for this

packages/core/src/api/resolvers/shop/shop-order.resolver.ts
aelhor: Can you explain how the order object will be represented without hitting the DB? Does it exist in the context?

aelhor: And will this postpone this PR?

DanielBiegler: > Can you explain how the order object will be represented without hitting the DB?

Sorry I dont mean to get rid off all the requests, just some. The `.getActiveOrder` function may hit the DB but then we only pass the ID to `.setCurrencyCodeForOrder` which fetches the order again. This holds true for other methods which also re-fetch orders. We could change function signatures to accept the order object directly, getting rid off this re-fetching.

Like I said, its correct of you to follow this convention, I'm posing this new shift as a general question, no worries

> And will this postpone this PR?

No probably not

packages/core/src/service/services/order.service.ts
```diff
@@ -583,6 +586,65 @@ export class OrderService {
             return result.order;
         }
     }
+    async setCurrencyCodeForOrder(
```
michaelbromley: I don't see the value of having this method which just wraps `updateOrderCurrency`. I would recommend just having the `updateOrderCurrency` one, as it is consistent with a few of the existing methods naming-wise.

packages/core/src/api/resolvers/shop/shop-order.resolver.ts
michaelbromley: @DanielBiegler it's a good point and one we can revisit in some future performance work I think.

packages/core/src/service/services/order.service.ts
aelhor: That's a good point 
I will fix this and directly use the updateOrderCurrency, use it in the resolver ,and delete setCurrencyCodeForOrder

aelhor:
@michaelbromley 
I'm not sure I fully understand the last comment 
Does this mean I have to change the base branch from master to a minor branch?

Regarding the e2e test and codegen replacement, could you please clarify what examples you are referring to? And does this require a change in the PR?

dlhck:
@aelhor yes, please rebase to minor and apply the new way how GraphQL codegen works there. Thank you!

michaelbromley:
Hi @aelhor - just let us know if you'll have time to make the changes we mentioned above. No problem if you don't - but it'd be good to know so we know how to proceed. Thanks!

aelhor:
@michaelbromley, Yes, I'd love to finish the work here.
But would you please clarify what examples you refering at from you previos comment here?
> "Check the other examples to see how it works".

michaelbromley:
@aelhor I mean that you can look at some other of the `.e2e-spec` files on the minor branch, and see how they no longer use the `gql` template literal from `graphql-tag` to define their graphql documents, but use a `graphql()` function instead, which is part of the gql.tada library we moved to on the minor branch.

Let me know if you need help on that point.

michaelbromley:
Superceded by 

- #4466

PR #4039, opened 2025-12-09
docs: Use web app instead of Angular app

AbderrahimDz24:
# Description

Use web app instead of Angular app since the Admin UI is built using ReactJs now.

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [ ] I have checked my own PR

👍 Most of the time:

## Summary by CodeRabbit

* **Documentation**
  * Updated Admin UI deployment guidance with refined terminology for improved clarity.

### DIFF
--- a/docs/docs/guides/deployment/deploying-admin-ui.md
+++ b/docs/docs/guides/deployment/deploying-admin-ui.md
@@ -43,7 +43,7 @@ const config: VendureConfig = {
 
 ## Deploying a stand-alone Admin UI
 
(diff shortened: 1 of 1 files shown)


michaelbromley:
Thanks for bringing this issue up. Actually the proper thing we need to do is have a totally separate new docs page for deploying the new React dashboard, since the current page is very much specific to the Angular Admin UI.

If you are interested in drafting one, let me know. It's a much more involved task than just changing the existing doc though. I understand if you don't want to do it - then we'll handle it internally.

AbderrahimDz24:
You're welcome.

You are absolutely correct; that makes perfect sense. I didn't notice the deployment guide was specific to the Angular Admin UI, even though the link to the Angular guide was right there.

Since it's a significant effort, I'll pass on drafting the new documentation page for the React dashboard right now, but thanks for the offer!

PR #4045, opened 2025-12-11
fix(dashboard): Use null instead of undefined when clearing single relation selector

### LINKED ISSUE #4032: dashboard: Single relation selectors fail to clear: onChange(undefined) should be onChange(null)
#### **Description**  
When using built-in Vendure relation selectors in **single-select mode** (e.g., `SingleRelationInput` with `customerRelationConfig` as shown in the [official docs](https://docs.vendure.io/guides/extending-the-dashboard/custom-form-components/relation-selectors#built-in-configurations)), clicking the × (remove) button triggers `onChange(undefined)`. However, `undefined` is **not serialized in JSON** and is **ignored by Vendure’s GraphQL API**, causing the previously selected value to persist on the server after form submission.

#### **To Reproduce**
1. Use the documented example:
   ```tsx
   import {
       customerRelationConfig,
       SingleRelationInput,
       DashboardFormComponent,
      CustomFormComponentInputProps, // not imported 
   } from '@vendure/dashboard';

   export function CustomerSingleSelector({ value, onChange }: DashboardFormComponent) {
       return <SingleRelationInput value={value} onChange={onChange} config={customerRelationConfig} />;
   }
   ```
2. Select a customer.
3. Click × to clear the selection.
4. Submit the form.
5. Refresh — the old value is restored.

#### **Expected Behavior**  
After clearing, `onChange(null)` should be called (not `undefined`), so the payload includes `"customerId": null`, and the relation is properly unset on the server.

#### **Actual Behavior**  
`onChange(undefined)` is called → field disappears from JSON payload → server retains the old value.

#### **Root Cause**  
The internal `RelationSelector` component (used by `SingleRelationInput`) calls `onChange(undefined)` in `handleRemove` for `multiple: false`. This is incompatible with Vendure’s expected input shape for optional relations (`ID` → `string | null`, not `undefined`).

#### **Workaround**  
Wrap `onChange` to convert `undefined` → `null`:
```tsx
export function CustomerSingleSelector({ value, onChange }: DashboardFormComponent) {
    return (
        <SingleRelationInput
            value={value}
            onChange={newValue => {
                onChange(newValue ?? null); // ✅ Ensures null is passed
            }}
            config={customerRelationConfig}
        />
    );
}
```

#### **Suggested Fix**  
Update `RelationSelector` (and/or `SingleRelationInput`) so that in single-select mode, `handleRemove` calls `onChange(null)` instead of `onChange(undefined)`. This aligns with GraphQL conventions and ensures clean interoperability with Vendure core.

#### **Impact**  
Affects all built-in and custom single-select relation components (`productRelationConfig`, `collectionRelationConfig`, etc.) wherever `multiple: false` is used.

---

iamgerwin:
## Summary

- Fixed the `RelationSelector` component to use `onChange(null)` instead of `onChange(undefined)` when clearing a single-select relation
- Updated the `onChange` type signature from `string | string[] | undefined` to `string | string[] | null`
- Updated `SingleRelationInput` to properly type the onChange callback

## Problem

When clearing a single-select relation field in the Dashboard, the component was calling `onChange(undefined)`. Since `undefined` values don't serialize in JSON, the GraphQL API would ignore the clearing request, causing previously selected values to persist on the server.

## Solution

Changed `onChange(undefined)` to `onChange(null)` which properly serializes to JSON and allows the GraphQL API to receive and process the null value, effectively clearing the relation field.

## Files Changed

- `packages/dashboard/src/lib/components/data-input/relation-selector.tsx` - Changed handleRemove to use `null` instead of `undefined`, updated type signature
- `packages/dashboard/src/lib/components/data-input/relation-input.tsx` - Updated SingleRelationInput onChange typing

## Test Plan

- [x] Create a custom field with a single-select relation (e.g., Customer relation on Product)
- [x] Select a relation value
- [x] Click the × button to clear the selection
- [x] Save the form and refresh the page
- [x] Verify the relation is properly cleared and doesn't reappear

Closes #4032

## Summary by CodeRabbit

* **Refactor**
  * Updated internal handling of empty relation selection states in the dashboard for improved consistency.

### DIFF
--- a/packages/dashboard/src/lib/components/data-input/relation-input.tsx
+++ b/packages/dashboard/src/lib/components/data-input/relation-input.tsx
@@ -1,5 +1,6 @@
 import { DashboardFormComponent, DashboardFormComponentProps } from '@/vdb/framework/form-engine/form-engine-types.js';
 import { graphql } from '@/vdb/graphql/graphql.js';
+
 import { createRelationSelectorConfig, RelationSelector } from './relation-selector.js';
 
 /**
@@ -35,7 +36,7 @@ export function SingleRelationInput<T>({
             config={singleConfig}
             value={value}
             selectorLabel={selectorLabel}
-            onChange={newValue => onChange(newValue as string)}
+            onChange={newValue => onChange(newValue as string | null)}
             disabled={disabled}
             className={className}
         />
--- a/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
+++ b/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
@@ -48,7 +48,7 @@ export interface RelationSelectorProps<T = any> {
      */
     selectorLabel?: React.ReactNode;
     value?: string | string[];
-    onChange: (value: string | string[] | undefined) => void;
+    onChange: (value: string | string[] | null) => void;
     disabled?: boolean;
     className?: string;
 }
@@ -344,7 +344,7 @@ export function RelationSelector<T>({
         } else {
             // Clear cache for single select
             setSelectedItemsCache([]);
-            onChange(undefined);
+            onChange(null);
         }
     };
 


michaelbromley:
Please can you explain to me what this is fixing. Please do not use any AI assistant when answering.

grolmus (approved):
Correct fix. `undefined` doesn't serialize in JSON so the clear request was silently dropped. `null` serializes properly and the GraphQL API processes it. Type signatures updated consistently.

grolmus:
Thanks for this fix @iamgerwin, and apologies it sat open so long! In the meantime the exact same change — clearing single relations with `null` instead of `undefined` — landed on `master` via #4495, so `relation-selector.tsx` already calls `onChange(null)` and the `onChange` type already accepts `null`. That's also why this PR now shows conflicts: it collides with the code that already carries the fix.

Since the fix is already in, I'll close this as resolved by #4495. Really appreciate you catching and fixing it — hope to see more contributions! 🙏

PR #4071, opened 2025-12-20
fix(core): apply pessimistic locking to prevent order modification race conditions #3398

### LINKED ISSUE #3398: Race condition when performing parallel order actions
## Description
When making parallel requests to update different parts of an order, the data gets inconsistently applied. This issue affects not only basic order details, but all order operations including customer information, addresses, shipping methods, order state transitions, and payment processing.

## Steps to Reproduce
1. Create an order by adding an item to the cart
2. Make parallel requests to any combination of:
   - Set customer information with `setCustomerForOrder`
   - Set shipping address with `setOrderShippingAddress`
   - Set billing address with `setOrderBillingAddress`
   - Set shipping method with `setOrderShippingMethod`
   - Transition order state with `transitionOrderToState`
   - Add payment with `addPaymentToOrder`
3. Retrieve the order data with `activeOrder`

## Expected Behavior
All updates should be properly applied to the order in a consistent manner, with all information present in the final order regardless of which operations were performed in parallel.

## Actual Behavior
The parallel requests result in inconsistent order state. Specifically:

- The individual mutation responses show success for their operations
- Each response shows only its own changes applied correctly
- The final `activeOrder` query shows an inconsistent state where some updates are applied and others are lost
- In our test case, customer information was completely lost despite a successful response from `setCustomerForOrder`

## Technical Details
The issue appears to be a race condition in the backend when handling concurrent order modifications. Each mutation seems to operate on its own version of the order entity, and the last one to commit wins, overwriting some fields from earlier mutations.

When a client performs multiple parallel operations on the same order, the backend processes might be retrieving and updating different instances of the same order entity without proper synchronization, leading to lost updates.

In our test logs:
1. `setCustomerForOrder` response shows customer data but empty addresses
2. `setOrderShippingAddress` response shows shipping and billing addresses but null customer
3. `setOrderBillingAddress` response shows only billing address
4. Final `activeOrder` contains shipping and billing addresses but null customer

This issue likely affects all order mutation operations including `setOrderShippingMethod`, `transitionOrderToState`, and `addPaymentToOrder`, which would be particularly problematic during checkout flows where multiple operations might be performed in quick succession.

## Environment
- Vendure version: v2.2.0
- NodeJS version: 22
- Database: postgres

## Suggested Fix
Consider implementing:
1. Transaction isolation or locking mechanisms for order entities
2. Atomic operations that can handle multiple order updates in a single request
3. Optimistic concurrency control using version numbers
4. A queue system for order operations to ensure they're processed sequentially

This issue requires attention as it could lead to incomplete orders, payment inconsistencies, and poor customer experience during checkout flows.

RubenDarioGuerreroNeira:
### Description
This PR addresses a critical race condition ("lost update" anomaly) that occurs when multiple concurrent requests attempt to modify the same Order simultaneously.

For example, if two `addItemToOrder` requests arrive at the same time for the same order, they both read the initial state, calculate the new state, and save it. The last transaction to commit overwrites the changes made by the first one, leading to incorrect order totals or missing items.

### Solution
I have implemented **Pessimistic Locking** (`SELECT ... FOR UPDATE`, specifically `pessimistic_write` mode) in the `OrderService`.

- Updated `getOrderOrThrow` (and the underlying `findOne`) to accept a `lock` option.
- Applied this lock to all critical `OrderService` methods that modify order state (e.g., `addItemsToOrder`, `adjustOrderLines`, `applyCouponCode`, `setShippingAddress`, etc.).
- This ensures that when a transaction fetches an order for modification, it locks the row in the database. Any concurrent transaction attempting to read the same order for modification will block and wait until the first transaction releases the lock.

### Fixes
Fixes #3398)

This test simulates concurrent `addItemToOrder` operations and asserts that the final order quantity is consistent with the number of requests sent.

- [x] Verified that pessimistic locking is applied to all modification paths in `OrderService`.
- [x] Tested with `race-condition.e2e-spec.ts` locally.

### DIFF
--- a/packages/core/e2e/race-condition.e2e-spec.ts
+++ b/packages/core/e2e/race-condition.e2e-spec.ts
@@ -1,0 +1,77 @@
+import { createTestEnvironment } from '@vendure/testing';
+import * as path from 'path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import { initialData } from '../../e2e-common/e2e-initial-data';
+import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../e2e-common/test-config';
+
+describe('Order race conditions', () => {
+    const { server, adminClient, shopClient } = createTestEnvironment(testConfig);
+
+    beforeAll(async () => {
+        await server.init({
+            initialData,
+            productsCsvPath: path.join(__dirname, '../../e2e-common/fixtures/e2e-products-full.csv'),
+            customerCount: 1,
+        });
+        await shopClient.asUserWithCredentials('test@vendure.io', 'test');
+    }, TEST_SETUP_TIMEOUT_MS);
+
+    afterAll(async () => {
+        await server.destroy();
+    });
+
+    it('handles parallel modifications to the order correctly', async () => {
+        const { products } = await shopClient.query(`
+            query GetProducts {
+                products(options: { take: 1 }) {
+                    items {
+                        variants {
+                            id
+                        }
+                    }
+                }
+            }
+        `);
+        const variantId = products.items[0].variants[0].id;
+
+        const ADD_ITEM_TO_ORDER = `
+            mutation AddItemToOrder($productVariantId: ID!, $quantity: Int!) {
+                addItemToOrder(productVariantId: $productVariantId, quantity: $quantity) {
+                    ... on Order {
+                        id
+                        totalQuantity
+                        total
+                    }
+                    ... on ErrorResult {
+                        errorCode
+                        message
+                    }
+                }
+            }
+        `;
+
+        const quantityPerRequest = 1;
+        const concurrency = 5;
+
+        // Executing 5 simultaneous requests to add an item
+        const promises: Array<Promise<any>> = [];
+        for (let i = 0; i < concurrency; i++) {
+            promises.push(
+                shopClient.query(ADD_ITEM_TO_ORDER, {
+                    productVariantId: variantId,
+                    quantity: quantityPerRequest,
+                }),
+            );
+        }
+
+        await Promise.all(promises);
+
+        const { activeOrder } = await shopClient.query(`
+            query GetActiveOrder { activeOrder { totalQuantity } }
+        `);
+
+        // If there is a race condition, the total will be less than 5 (some writes overwritten)
+        expect(activeOrder.totalQuantity).toBe(concurrency * quantityPerRequest);
+    });
+});
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import {
     AddPaymentToOrderResult,
     ApplyCouponCodeResult,
+    CurrencyCode,
     PaymentInput,
     PaymentMethodQuote,
     RemoveOrderItemsResult,
@@ -12,7 +13,6 @@ import {
     AddFulfillmentToOrderResult,
     AddManualPaymentToOrderResult,
     AddNoteToOrderInput,
-    AdjustmentType,
     CancelOrderInput,
     CancelOrderResult,
     CancelPaymentResult,
@@ -40,13 +40,13 @@ import {
 import { omit } from '@vendure/common/lib/omit';
 import { ID, PaginatedList } from '@vendure/common/lib/shared-types';
 import { summate } from '@vendure/common/lib/shared-utils';
-import { In, IsNull } from 'typeorm';
+import { EntityManager, In, IsNull, LockNotSupportedOnGivenDriverError } from 'typeorm';
 import { FindOptionsUtils } from 'typeorm/find-options/FindOptionsUtils';
 
 import { RequestContext } from '../../api/common/request-context';
 import { RelationPaths } from '../../api/decorators/relations.decorator';
 import { RequestContextCacheService } from '../../cache/request-context-cache.service';
-import { CacheKey } from '../../common/constants';
+import { CacheKey, TRANSACTION_MANAGER_KEY } from '../../common/constants';
 import { ErrorResultUnion, isGraphQlErrorResult, JustErrorResults } from '../../common/error/error-result';
 import { EntityNotFoundError, InternalServerError, UserInputError } from '../../common/error/errors';
 import {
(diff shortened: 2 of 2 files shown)


michaelbromley:
Hi!
Thank you for looking into this. I notice a lot of extra workflow-related changes in the PR, which seem to be causing the workflows to fail.

Can you review that part and make sure this PR contains only those changes that are absolutely necessary to address the issue at hand?

Lennix:
Just wanted to chip in, as we are also facing concurrency issues. We could fix most issues with the pessimistic_write lock, but it could still happen that concurrent `addItemToOrder` hit a session where no active order exists. In that case the lock did not work, because multiple orders were created (race condition happened there).
Our current workaround is to create an order every time a session gets created, which is not optimal.

RubenDarioGuerreroNeira:
> Just wanted to chip in, as we are also facing concurrency issues. We could fix most issues with the pessimistic_write lock, but it could still happen that concurrent `addItemToOrder` hit a session where no active order exists. In that case the lock did not work, because multiple orders were created (race condition happened there). Our current workaround is to create an order every time a session gets created, which is not optimal.


Thanks for the insight! That is an excellent point regarding the creation race condition.

I agree that this PR mainly solves the concurrency issue when an Order already exists (via pessimistic_write). The 'creation race condition' seems like a distinct challenge that might require a different approach (like unique constraints on active order per session or locking the Session entity) to avoid the suboptimal 'create empty order everywhere' strategy.

Given that this PR is already stabilizing the CI and fixing the modification race conditions, I suggest we proceed with this fix as a first step. I can then open a separate issue/discussion to tackle the creation edge-case specifically. WDYT?

grolmus:
Hi @RubenDarioGuerreroNeira, thanks for tackling this — race conditions on order modification are a real and important problem. I went through the change carefully and have a number of concerns that I'd want to see addressed before this can land. Sharing them here so you have full context.

## Blockers

### 1. Pessimistic locking is not supported on sqljs (default test DB)

TypeORM throws `LockNotSupportedOnGivenDriverError` when `setLock('pessimistic_write')` (or the equivalent `lock` find option) is used on the `sqljs`/`sqlite` driver. Every modification path in `OrderService` now unconditionally requests this lock, which will crash on sqljs.

This matters for two reasons:
- The CI matrix includes `DB: sqljs`, but the e2e workflow run that should catch this didn't actually execute on this PR (probably because workflow changes from forks need approval).
- Many users and contributors run the e2e suite locally with the default sqljs config — they'll hit hard crashes on practically every shop mutation.

The right pattern here is the one already established in `packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts:242-289` — branch on `connection.rawConnection.options.type` and only apply pessimistic locking on `postgres`/`mysql`/`mariadb`, with a sensible fallback for other drivers (or a no-op, since sqljs is single-connection so the race can't happen anyway).

### 2. `leftJoin('order.channels')` → `innerJoin('order.channels')` is a silent behavior change

```ts
.leftJoin('order.channels', 'channel')   →   .innerJoin('order.channels', 'channel')
```

This was likely needed because TypeORM's pessimistic locking complains about left-joined tables, but it changes semantics: any Order without an associated channel row in the join (whatever the cause — partial migrations, customizations, edge cases) now silently disappears as `EntityNotFoundError` instead of being returned. This is a regression that's not mentioned in the PR description.

If the join change is required for locking to work, please document the rationale and verify there's no scenario in core where an Order exists without a channel row in the join table.

### 3. Scope creep — three unrelated changesets in one PR

The PR mixes:

- **The actual fix** — adding `lock` option to `getOrderOrThrow`/`findOne` and applying it across `OrderService` (the core change, ~25 lines in `order.service.ts`)
- **CI workflow stabilization** — 58 added lines in `build_and_test.yml` (DB env vars, separate `Wait for services to be ready` step, Elasticsearch tweaks, healthcheck changes)
- **Test config connection-pool tweaks** — 38 added lines in `test-config.ts` (connection limits, timeouts, init SQL, `keepConnectionAlive`)
- **`docker-compose.yml`** — switching MariaDB to InnoDB by command override

Please split these into separate PRs. The CI/connection-pool changes are infrastructure work that deserves its own review trail; the actual locking fix should be its own focused diff so it's easy to reason about and revert if needed. As a bonus the locking-only PR will be a lot easier to land.

### 4. Always-on pessimistic locking has performance and deadlock implications

Currently every single shop mutation that touches an Order (`addItemToOrder`, `adjustOrderLine`, `applyCouponCode`, `setShippingAddress`, etc.) takes a `SELECT ... FOR UPDATE` even when there's no concurrent access. Two consequences:

- **Performance**: every cart update now serializes on the order row even when there's only one client. This is fine in a low-traffic shop but adds non-trivial overhead at scale.
- **Deadlocks**: any flow that touches multiple Orders, or touches the Order plus another locked resource, can deadlock if a concurrent flow takes the locks in a different order. With locks now spread across ~17 methods, the surface for deadlock detection increases.

Worth considering alternatives or making this configurable:

- **Optimistic locking** via TypeORM's `@VersionColumn` on the `Order` entity, with retry on `OptimisticLockVersionMismatchError`. Less contention, but requires retry logic at a higher level.
- **Application-level mutex** keyed by order id, similar to how some other Vendure systems handle exclusion.
- **Opt-in via config** — a `OrderOptions.useStrictOrderLocking: boolean` so users who don't have a contention problem don't pay the cost.

Worth getting maintainer input on the preferred direction here before locking in a specific approach.

## Other concerns (non-blocking)

- **Test coverage is thin.** The new `race-condition.e2e-spec.ts` only exercises `addItemToOrder`. Locking was added to ~17 methods; ideally we'd verify at least a couple of representative paths (e.g. `applyCouponCode`, `adjustOrderLine`).
- **Test contains a Spanish comment** (`Si hay condición de carrera, el total será menor a 5 (algunas escrituras se sobrescribieron)`). Please translate to English to match the repo convention.
- **Test uses a hard-coded variant id** (`T_1`). Brittle; better to look it up via the admin API or a fixture.
- **`updateCustomFields` and `addNoteToOrder`** receive `pessimistic_write` locks too, even though they're admin-only and carry low risk of concurrency conflict. Worth considering whether these need to participate in the locking.
- **Docs/breaking-change note**: this is a behaviour change for anyone extending `OrderService` or building on top of it. A note in the release notes / changelog would be valuable.
- **Currently `CONFLICTING` with master** — needs a rebase.

## Summary

The intent is right and the issue this fixes (#3398) is real, but the current implementation has at least one hard blocker (sqljs crash), one silent behavior change (innerJoin), and a scope problem that makes the PR hard to review and land. If you have time, the most useful next step would be:

1. Split out the CI/test-config/docker changes into a separate PR.
2. Add the DB-type guard so sqljs/sqlite gets a no-op fallback (mirror the scheduler-strategy pattern).
3. Justify or revert the leftJoin → innerJoin change.
4. Discuss with the maintainers whether always-on pessimistic locking is the right default vs. an opt-in / optimistic approach.

If you don't have bandwidth to take this all the way, let us know and we can pick it up — happy to land this so the underlying race condition gets fixed.

RubenDarioGuerreroNeira:
Hi @grolmus, thank you for taking the time to go through this in such detail — and sorry for the slow response.

Agreed on all four points, particularly the sqljs crash, that needs fixing regardless of which direction we take. Before I rework this, I'd like your input on point 4: would you prefer I pursue pessimistic locking (scoped to postgres/mysql/mariadb only, per your suggestion) or would optimistic locking via @VersionColumn be a better fit for this codebase long-term?

Once we're aligned on approach, I'll split this into a focused PR (just the OrderService change) and open the CI/test-config changes separately, as you recommended. Happy to close this PR and start clean if that's easier to review.

RubenDarioGuerreroNeira:
Per the review feedback, I have created a **new, clean PR** that addresses all 4 blocking issues from the review in a single focused change.

RubenDarioGuerreroNeira:
Closing this PR as the clean version has been submitted in https://github.com/RubenDarioGuerreroNeira/vendure/pull/5 which addresses all 4 blocking issues from the review.

PR #4075, opened 2025-12-21
fix(dashboard): Channel switcher not scrollable if list exceeds screen height

lucatk:
# Description

I have noticed that if there's more channels than can fit into the viewport height in the Channel switcher, the switcher overflows instead of providing proper scrolling.

Therefore, the channel switcher was limited in height to the viewport (with some margin), and scrolling with a sticky header for the current channel was added.

# Breaking changes

No

# Screenshots

Short demonstration video:

https://github.com/user-attachments/assets/1bd3952e-cf60-440c-a78e-32e9b1ac45ce

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
+++ b/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
@@ -13,6 +13,7 @@ import {
     DropdownMenuTrigger,
 } from '@/vdb/components/ui/dropdown-menu.js';
 import { SidebarMenu, SidebarMenuButton, SidebarMenuItem, useSidebar } from '@/vdb/components/ui/sidebar.js';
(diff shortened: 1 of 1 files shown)


dlhck (changes requested):
Could you please try to use the [Scroll Area](https://ui.shadcn.com/docs/components/scroll-area) component?

lucatk:
@dlhck I tried, idk if this is how you would imagine it?
Without the explicit `overflow-y-auto` I wasn't able to get it to scroll (despite not finding this requirement documented), and I wasn't happy with the way the right padding on DropdownMenuContent moved the scrollbar inside the container.

michaelbromley:
Thanks!

Issue #4005, opened 2025-11-27
localeString custom fields not displayed in Vendure Admin Dashboard


thochoertler:
Custom fields of type `localeString` do not appear in the Vendure Admin Dashboard.  
The field is defined correctly, the Admin UI is rebuilt successfully, and no errors are shown — but the field is completely missing from the entity detail view and create/update form.

**To Reproduce**  

1. Add a `localeString` custom field to an entity (e.g. Product):
   
```typescript
   {
     name: "shortName",
     type: "localeString",
     nullable: true,
     label: [
       { languageCode: LanguageCode.de, value: "Kurztitel" },
       { languageCode: LanguageCode.en, value: "Short name" },
     ],
     ui: { component: "text-form-input" },
   }
```

2. Restart the server.
3. Rebuild the Admin UI (`yarn build` / `pnpm build`).
4. Open the entity in the Admin Dashboard.
5. Observe that the field is not rendered.

**Expected behavior**
The Admin Dashboard should display a localized text input for `localeString` fields — respecting the selected language and the configured UI component (`text-form-input`).

**Actual behavior**

* The field is not visible anywhere in the Admin UI.
* It's missing in the detail view, create/update form, and list view.
* No warnings or errors appear in the browser console or server logs.
* The field exists in the GraphQL schema and is returned normally through the API.

**Screenshots/Videos**
(If you need, I can prepare an example reproduction video.)
<img width="1081" height="221" alt="Image" src="https://github.com/user-attachments/assets/d850dfd5-4589-45dc-888e-f5765f0c95d2" />

<img width="722" height="129" alt="Image" src="https://github.com/user-attachments/assets/a1c73ace-a03e-4e4b-98d9-ce5ffd775424" />

**Environment (please complete the following information):**

* @vendure/core version: 3.5.1
* Nodejs version: 22
* Database (MySQL/Postgres/etc): postgres
* Operating System: Mac
* Browser: Sarai
* Package manager (npm/yarn/pnpm): npm

**Configuration**

```typescript
customFields: [
  Product: [
    {
      name: "shortName",
      type: "localeString",
      nullable: true,
      label: [
        { languageCode: LanguageCode.de, value: "Kurztitel" },
        { languageCode: LanguageCode.en, value: "Short name" },
      ],
      ui: { component: "text-form-input" },
    },
  ],
],
```

BibiSebi:
Hello @thochoertler,

[...]

PR #4081, opened 2025-12-23
Fixes the issue where localeString custom fields do not appear in the Admin Dashboard despite being correctly defined in the configuration.

starry-osean:
Summary
When localeString/localeText custom fields exist but a query doesn’t explicitly select translations, the dashboard failed to request translations.customFields, so localized custom fields wouldn’t render.
In add-custom-fields.ts, I now  a translations selection set (with languageCode) and inject translations.customFields. whenever localized custom fields are present. This guarantees the data is fetched and the fields render.
Changes
packages/dashboard/src/lib/framework/document-introspection/add-custom-fields.ts
If localized fields are detected, ensure a translations selection set exists and append customFields selections for all localized custom fields.
Validation
Install deps (ensure @vendure/core is available), then: npx vite build --emptyOutDir.
In the admin UI, create/edit an entity with a localeString custom field (e.g., Product.shortName), switch languages, and confirm the field appears and is editable.
CI Status Note
The CI test failure appears to be an environment issue (port conflicts: EADDRINUSE: address already in use :::3000 and Vite port mismatch). The dashboard tests timeout because:
Server port 3000 is already in use
Vite runs on port 5175 instead of the expected 5173.  This is unrelated to the code changes. Local validation confirms the fix works correctly. The build and type checks pass locally, and the functionality has been manually verified.
Related Issue: Fixes the issue where localeString custom fields do not appear in the Admin Dashboard despite being correctly defined in the configuration.


starry-osean:
Hi，@BibiSebi ！Thank you for taking the time to push forward the resolution of this issue. The current issue is a test timeout: e2e tests are timing out in the MySQL environment, possibly due to:
  1. Long test execution time.vitest.
  2. CI environment resource constraints.
  3. Slow execution of individual test cases.
I have a solution: we can increase the CI environment test timeout.Like  this：
  vitest.config.mts Line 12:
  testTimeout: process.env.E2E_DEBUG ? 1800 * 1000 : process.env.CI ? 60 * 1000 : 15 * 1000（The timeout for the CI environment has been increased from 30 seconds to 60 seconds.）

BibiSebi:
Hello @starry-osean thank you for taking your time to submit a PR on this. 🙏 

Could you add some screen-recording of a before and after behavior? I believe the issue #4005 is just that the `text-form-input` was not globally registered and that is why it never rendered, even when the custom field was not a localized one. It might be that what you fixed is a separate thing, and that is why I would like to reproduce it. 


Thank you 🙏

BibiSebi:
Just to give you an overview of what I tried:

I extended the config with the following values. That rendered an input field and it was possible for me to add translatable content (see recording).

```
customFields: {
        Product: [
            {
                name: 'testName',
                type: 'localeString',
                nullable: true,
                label: [
                    { languageCode: LanguageCode.en, value: 'Test Name' },
                ]
            }
}

```


https://github.com/user-attachments/assets/17f6c062-bd92-4354-a1e2-aefd3242813f

starry-osean:
Hi，@BibiSebi  ！This picture  is the repaired version.
修
And indeed, your approach is more efficient and simple !

BibiSebi:
Thank you for the quick response @starry-osean, in that case I would close this PR and proceed by merging mine. I will keep an eye out on the issue and if something similar/connected re-appears I will revisit your solution.

Issue #4091, opened 2026-01-01
Insights page shows NaN or empty values in Latest Orders table when filtering


raidsobhi:
## **Describe the bug**

On **Admin Dashboard → Insights → Latest Orders**, **all optional order fields** (e.g. totals, customer, state, etc.) render as **empty values or invalid data** when table filters are applied.

The same fields display correctly on other Admin UI pages (Order List, Order Details).

---

## **To Reproduce**

1. Go to **Admin UI**
2. Navigate to **Dashboard → Insights**
3. Scroll to the **Latest Orders** widget
4. Enable any **optional column** (e.g. customer, total with tax, state)
5. Observe empty cells or invalid values in those columns

---

## **Expected behavior**

All visible table columns should render correctly and continue to work when filters or sorting are applied.

---

## **Actual behavior**

After filtering:

* Optional columns display **empty values**
* Monetary fields may render as `NaN`
* Customer or state columns may render blank

No errors are thrown; the data is simply missing from the query result.

---

## **Screenshots**

<img width="1920" height="892" alt="Image" src="https://github.com/user-attachments/assets/d6ac9d6f-013f-43ac-9409-30edbe6da053" />

---

## **Error logs**

No console or server errors are logged.

---

## **Environment**

* **@vendure/core version:** 3.5.2
* **Node.js version:** 22
* **Database:** PostgreSQL
* **Operating System:** Linux
* **Browser:**  Firefox
* **Package manager:** npm

---

## **Configuration**

Default Vendure configuration.
No custom Admin UI or dashboard widgets.

---

## **Minimal reproduction**

1. Install Vendure with default Admin UI
2. Navigate to **Dashboard → Insights**
3. Apply filters in **Latest Orders**
4. Observe missing data in optional columns

Issue #4085, opened 2025-12-24
Insights page shows "Total with tax" as NaN in Latest Orders


Ryrahul:
## Describe the bug
On the **Admin Dashboard → Insights** page, in the **Latest Orders** section, the **Total with tax** value is displayed as `$NaN` instead of a valid amount.

Order totals (including tax) are calculated and displayed correctly in other places such as the order details page and the order list page.

---

## To Reproduce

1. Create an order with tax applied
2. Open the Vendure Admin UI
3. Go to **Dashboard / Insights**
4. Check the **Latest Orders** section
5. Observe the **Total with tax** column

---

## Expected behavior
The **Total with tax** value in the **Latest Orders** section should display a valid numeric amount with the correct currency.

---

## Actual behavior
The **Total with tax** column displays `$NaN`.

---

## Screenshots/Videos

<img width="645" height="666" alt="Image" src="https://github.com/user-attachments/assets/6a38504a-6b82-47f4-8f1e-0ef6575f7064" />

---

## Environment
- **@vendure/core version:** `3.5.2`

---

PR #4092, opened 2026-01-02
fix(dashboard): Prevent NaN or empty values in Latest Orders table w…

raidsobhi:
## Description

Related issue: #4091 & #4085

This PR fixes an issue where the **Latest Orders** widget on the **Insights** page shows **NaN or empty values** for optional table columns **when filtering**.

The widget uses a minimal list query while rendering optional columns that do not explicitly declare their required data dependencies. When filtering is applied, the table attempts to access fields that were not fetched, resulting in missing or invalid values.

This change ensures that column dependencies are properly declared so the query remains minimal but complete, and all visible columns continue to render correctly when filtering.

---

## Breaking changes

No breaking changes.

---

## Screenshots

image

---

## Checklist

📌 Always:

👍 Most of the time:

* [ ] I have added or updated test cases
* [ ] I have updated the README if needed


michaelbromley:
Thanks for this fix! 🙏

PR #4094, opened 2026-01-04
fix(dashboard): Re-sync channel token when activeChannelId persists but token is cleared

niko91i:
## Summary

  This PR fixes a bug that causes a **black screen** in the Admin UI for users with limited channel access (e.g., multi-tenant/marketplace vendors).

  ## Problem

  After logout, `localStorage` state becomes desynchronized:
  - `vendure-selected-channel-token` is cleared
  - `vendure-user-settings.activeChannelId` persists

  On next login, `selectedChannelId` is initialized from `activeChannelId`, so the existing check `if (!selectedChannelId)` never triggers token restoration. API requests are sent without the `vendure-token` header, resulting in `FORBIDDEN` errors and a blank screen.

  ## Solution

  Added a check in the `useEffect` of `ChannelProvider` to ensure the localStorage token stays in sync with the selected channel, re-setting it if missing or mismatched.

  ## Testing

  - Manually verified fix by removing `vendure-selected-channel-token` from localStorage
  - Confirmed token is automatically restored on page reload
  - Dashboard renders correctly after fix

  ## Checklist

  - [x] I have read the [Contributing Guidelines](https://github.com/vendure-ecommerce/vendure/blob/master/CONTRIBUTING.md)
  - [x] I have performed a self-review of my own code
  - [x] My changes follow the code style of this project

### DIFF
--- a/packages/dashboard/src/lib/providers/channel-provider.tsx
+++ b/packages/dashboard/src/lib/providers/channel-provider.tsx
@@ -223,12 +223,12 @@ export function ChannelProvider({ children }: Readonly<{ children: React.ReactNo
     // Find the selected channel from the list of channels
     const selectedChannel = activeChannelData?.activeChannel;
 
-    const refreshChannels = () => {
+    const refreshChannels = React.useCallback(() => {
         refreshCurrentUser();
         queryClient.invalidateQueries({
             queryKey: ['channels', isAuthenticated],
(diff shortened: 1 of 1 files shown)


niko91i:

michaelbromley:
Thank you for the fix!

PR #4104, opened 2026-01-09
Mollie: Allow forcefully updating payment status in case webhooks are delayed

martijnvdbrug:
# Description

This PR introduces the functionality to:
* Fetch payment status from Mollie and update the Vendure order status according to the payments found in Mollie. This can be done using the mutation `syncMolliePaymentStatus`, in case webhook delivery from Mollie is delayed.
* An emergency switch `disableWebhookProcessing` has been introduced. This makes the Mollie plugin ignore any incoming webhooks.

## Why?

In November 2025 [an incident](https://status.mollie.com/cmilqy36b0a935cby38iwie3q) occurred with Mollie webhooks that resulted in:
1. Webhooks being severely delayed, sometimes longer than 60 minutes. This resulted in orders staying in `AddingItems` state, even though the customers paid.
2. In some cases, duplicated webhooks for the same order were received simultaneously. This resulted in duplicate `OrderPlacedEvent's`. The plugin's webhook mechanism should be idempotent, but it can't handle parallel/concurrent webhooks for the same order. This is why the option `disableWebhookProcessing` was introduced: it allows the consumer to resort to `syncMolliePaymentStatus`, and ignore webhooks altogether.

# Breaking changes

None :+1: 

# Additional Context

We have considered implementing some kind of locking mechanism for concurrent webhooks, but this feels overly complex for something that is definitely an anomaly: This was the first time in 5 years webhooks came in simultaneously.

# Checklist

📌 Always:

👍 Most of the time:

## Summary by CodeRabbit

* **New Features**
  * Added capability to manually synchronize Mollie payment status when webhooks experience delays
  * Introduced configuration option to optionally disable webhook processing for alternative payment flow setups

* **Tests**
  * Added comprehensive end-to-end test coverage for manual payment status synchronization

* **Refactor**
  * Improved internal payment status handling and method creation flows

* **Chores**
  * Removed unused import


packages/payments-plugin/e2e/mollie-dev-server.ts
```diff
@@ -61,34 +56,30 @@ async function runMollieDevServer() {
         }
     `);
     // Create method
-    await adminClient.query<CreatePaymentMethodMutation, CreatePaymentMethodMutationVariables>(
```
martijnvdbrug, line 64: Not sure why, but these types got lost somewhere in the migration to Tada. It's just the dev-server, so fine with me...

packages/payments-plugin/src/mollie/mollie.plugin.ts
```diff
@@ -166,27 +172,49 @@ export interface MolliePluginOptions {
...
+ * }
+ * ```
+ *
+ * You should wait for an incoming webhook first, because due to technical limitations on the Mollie API, the `syncMolliePaymentStatus`
+ * mutation will iterate through the last 500 Mollie payments to find the payments for the given order.
+ * Hence, it is not very performant, and should only be used as a fallback when a webhook
+ * was not received for ~10 seconds.
  *
- * ## Pay later methods
+ * ### Pay later methods
```
martijnvdbrug, line 200: Fixed heading levels to fix the ToC 
image

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -413,31 +417,88 @@ export class MollieService {
...
+    ): Promise<[MollieClient, PaymentMethod]> {
+        const paymentMethod = await this.getPaymentMethod(ctx, paymentMethodCode);
+        if (!paymentMethod) {
+            throw Error(`No Mollie payment method found`);
+        }
+        const apiKey = paymentMethod?.handler.args.find(arg => arg.name === 'apiKey')?.value;
+        if (!apiKey) {
+            throw Error(`No apiKey configured for payment method ${paymentMethod.code}`);
+        }
+        return [createMollieClient({ apiKey }), paymentMethod];
```
martijnvdbrug, line 521: The method was changed to also return the payment method, because that can contain channel-specific configuration.

martijnvdbrug:
The only failing test is from `vendure/core`, only for Node 24 with MySQL:

```shell
@vendure/core:  ✓ e2e/entity-prefix.e2e-spec.ts (1 test) 13320ms
@vendure/core: ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
@vendure/core:  FAIL  e2e/job-queue.e2e-spec.ts > JobQueue > subscribable that times out
@vendure/core: AssertionError: expected '' to be 'Job subscription timed out. The job m…' // Object.is equality
@vendure/core: - Expected
@vendure/core: + Received
@vendure/core: - Job subscription timed out. The job may still be running
@vendure/core:  ❯ e2e/job-queue.e2e-spec.ts:158:37
@vendure/core:     156| 
@vendure/core:     157|         expect(result.status).toBe(200);
@vendure/core:     158|         expect(await result.text()).toBe('Job subscription timed out. …
@vendure/core:        |                                     ^
@vendure/core:     159|         const jobs = await getJobsInTestQueue(JobState.RUNNING);
@vendure/core:     160|         expect(jobs.items.length).toBe(1);
```

Since it's a timeout, and no files in core were changed, I am going to assume it is not related to this PR.

packages/payments-plugin/src/mollie/mollie.service.ts
```diff
@@ -413,31 +417,88 @@ export class MollieService {
...
+        Logger.info(
+            `Synced status for order '${order.code}' from '${originalOrderState}' to '${
+                order.state
+            }' based on Mollie payment(s) ${processedPaymentIds.join(',')}`,
+            loggerCtx,
+        );
+        if (!(await this.configService.orderOptions.orderByCodeAccessStrategy.canAccessOrder(ctx, order))) {
+            throw new ForbiddenError(LogLevel.Verbose);
+        }
+        return order;
```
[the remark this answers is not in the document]
martijnvdbrug: That is ok, because the order status is updated based on the actual mollie payment. That is fine, even when the callee isn't authorized for the order.

packages/payments-plugin/src/mollie/mollie.common-resolver.ts
```diff
@@ -22,6 +22,14 @@ export class MollieCommonResolver {
         return this.mollieService.createPaymentIntent(ctx, input);
     }
 
+    @Mutation()
+    async syncMolliePaymentStatus(
+        @Ctx() ctx: RequestContext,
+        @Args('orderCode') orderCode: string,
+    ): Promise<Order | undefined> {
+        return this.mollieService.syncMolliePaymentStatus(ctx, orderCode);
```
michaelbromley, line 31: Can you walk me through the security model here? I see there's no specific permission needed here - is there any risk of unauthorised usage?

martijnvdbrug: @michaelbromley It should be the same as `orderByCode`, but I see that that one does have `Permission.owner` set. To my understanding, the OrderByCode access is defined by `orderByCodeAccessStrategy.canAccessOrder`, so what does the `Permission.owner` do here? Do I also need that on the Mollie mutation?

From a functional perspective: It is OK if the order status is updated by someone who is unauthorised, as long as we don't return the actual order.

Edit: I have added an additional order state check **before** fetching payments, to prevent unnecessary Mollie API calls.

michaelbromley: Permission.owner actually does not provide _any_ control over access, it's more of a hint that you need to implement some business logic to limit access to the logical "owner" of the resource in question. In this case, the owner would be the user who is related to this order. `orderByCode` is expressly designed for the guest checkout scenario though so that's why it defaults to a time-based access control.

martijnvdbrug: Ahh ok, in that case I will add that decorator as well. 

Other that, we expect this sync mutation to be called from the confirmation page, as a fallback to `orderByCode`, thus it should also allow unauthenticated calls.

It works like this:
1. Customer paid at the Mollie platform (a redirect to mollie.com)
2. Mollie redirects to myshop.com/order-confirmation
3. On the confirmation page, we poll `orderByCode` for 10 seconds
4. If `orderByCode` does not give us the order within 10 seconds, we resolve to calling `syncMolliePaymentStatus` once, because apparently the Mollie webhook is delayed.
5. `syncMolliePaymentStatus` fetches payments from Mollie, and updates the Vendure order if a payment has been found.
6. Only if the `OrderByCodeAccessStrategy` allows it, do we return the actual order. (We can not do the "canAccess" check before hand, because the order will not be settled yet).

Issue #3373, opened 2025-02-18
core: Unused code with important comment


Draykee:
I just want to mention it once, that I found unused code with a comment, that let it sound little important:
https://github.com/vendure-ecommerce/vendure/blob/98a4087ed9544d978eb33ca8211069a7f657c837/packages/core/src/service/services/order.service.ts#L304

```TS
const effectiveRelations = (relations ?? ['lines', 'customer', 'channels', 'shippingLines']).filter(
            r =>
                // Don't join productVariant because it messes with the
                // price calculation in certain edge-case field resolver scenarios
                !r.includes('productVariant'),
        );
```

The problem is that `effectiveRelations` is never used here.

Shall we just remove it, or use `effectiveRelations` instead?

PR #4118, opened 2026-01-15
fix(core): Use previously unused relations filter in findByCustomerId

DeltaSAMP:
# Description

Addresses Issue #3373

effectiveRelations was first added in this [commit](https://github.com/vendurehq/vendure/commit/b98220b6a) where it does seem like it was just accidentally not used.

I also think I found the original related Issue that lead to this change.
https://github.com/vendurehq/vendure/issues/1508

Included in the fix to the issue this [commit](https://github.com/vendurehq/vendure/commit/5061dd923f25bcae43fccf4602022d790a790795) removed the relations from the fallback which of course still leads to the issue that the passed relations can include the productVariant relation which was addressed in the commit mentioned above.

Now while testing before and after adding `effectiveRelations` I never found anything wrong with the price calculations like mentioned in the comment but I also don't know what the specific edge case is so I would say it still makes sense to add it to make sure the issue doesn't come up again.

# Breaking changes

no

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -335,7 +335,7 @@ export class OrderService {
         );
         return this.listQueryBuilder
             .build(Order, options, {
-                relations: relations ?? ['lines', 'customer', 'channels', 'shippingLines'],
+                relations: effectiveRelations,
                 channelId: ctx.channelId,
                 ctx,
             })


michaelbromley (approved):

michaelbromley:
Thanks for the research on this Alex! 👍

Issue #4010, opened 2025-12-01
Potential performance issue: N+1 Query warnings in Sentry for CollectionList query when listing the collections in dashboard


robinjoh:
**Describe the bug**
Getting N+1 Query warnings in Sentry from this query in the Admin API when browsing collections in the dashboard:

`query CollectionList($options: CollectionListOptions) { collections(options: $options) { items { id createdAt updatedAt featuredAsset { ...Asset } name slug description breadcrumbs { id name slug } children { id name } position isPrivate parentId productVariants { totalItems } } totalItems } } fragment Asset on Asset { id createdAt updatedAt name fileSize mimeType type preview source width height focalPoint { x y } }
`

Not sure this is a big issue, but since this can be a potential performance problem i thought i would report it.

**To Reproduce**
I get this warning when browsing individual collections from the collections overview.

**Expected behavior**
GraphQL query to be performed in bulk instead of doing multiple fetches to the DB.

**Actual behavior**
Getting warning about the following query being repeated over and over:
```
SELECT DISTINCT "distinctAlias"."productvariant_id" AS "ids_productvariant_id"
FROM (
  SELECT "productvariant"."createdAt" AS "productvariant_createdAt", "productvariant"."updatedAt" AS
    "productvariant_updatedAt", "productvariant"."deletedAt" AS "productvariant_deletedAt",
    "productvariant"."enabled" AS "productvariant_enabled", "productvariant"."sku" AS
    "productvariant_sku", "productvariant"."outOfStockThreshold" AS "productvariant_outOfStockThreshold",
     "productvariant"."useGlobalOutOfStockThreshold" AS "productvariant_useGlobalOutOfStockThreshold",
    "productvariant"."trackInventory" AS "productvariant_trackInventory", "productvariant"."id" AS
    "productvariant_id", "productvariant"."featuredAssetId" AS "productvariant_featuredAssetId",
    "productvariant"."taxCategoryId" AS "productvariant_taxCategoryId", "productvariant"."productId" AS
    "productvariant_productId", "productvariant"."customFieldsShortdescription" AS
    "productvariant_customFieldsShortdescription", "productvariant"."customFieldsLongdescription" AS
    "productvariant_customFieldsLongdescription", "productvariant"."customFieldsDeliverytime" AS
    "productvariant_customFieldsDeliverytime", "productvariant"."customFieldsProducttype" AS
    "productvariant_customFieldsProducttype", "productvariant"."customFieldsBrand" AS
[...]

PR #4132, opened 2026-01-19
feat: Implement productVariantCount in collections query

biggamesmallworld:
# Description
Thi PR addressed issue #4010
Using the following structure in the collections query: 

```
query CollectionList($options: CollectionListOptions) {
  collections(options: $options) {
    items {
      id
      createdAt
      updatedAt
      featuredAsset {
        ...Asset
      }
      name
      slug
      description
      breadcrumbs {
        id
        name
        slug
      }
      children {
        id
        name
      }
      position
      isPrivate
      parentId
      productVariants {
        totalItems
      }
    }
    totalItems
  }
}
```

resulted in an N+1 query due to fetching all productVariants within the collection just for the totalItems field. This PR adds a new field that uses a simple count query that eliminates the need to fetch each variant. The dashboard now uses this gql document. This approach was preferred over a field resolver for totalItems due to the reduction of complexity. Although it may have been possible to keep the document the same, this is the simplest solution that touches the least amount of code.

Additionally, we set a weak map on the request cache which stores every collection id we are fetching in order to perform only one query for the `productVariantCount` field on each collection. This is filled with the promise for the count query, and awaited if retrieved. This allows us to easily access the response in the map, and await just the single promise that captures the counts.  

# Breaking changes

None

# Checklist

📌 Always:

### DIFF
--- a/packages/core/src/api/common/is-field-in-selection.ts
+++ b/packages/core/src/api/common/is-field-in-selection.ts
@@ -1,8 +1,9 @@
-import { FieldNode, GraphQLResolveInfo } from 'graphql';
+import { FieldNode, GraphQLResolveInfo, SelectionNode } from 'graphql';
 
 /**
  * Checks if a specific field is requested in the GraphQL query selection set.
  * Looks for the field within the 'items' selection of a paginated list.
+ * Supports direct field selections, fragment spreads, and inline fragments.
  */
 export function isFieldInSelection(
     info: GraphQLResolveInfo,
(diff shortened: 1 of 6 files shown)


packages/core/src/api/common/is-field-in-selection.ts
```diff
@@ -0,0 +1,18 @@
+import { FieldNode, GraphQLResolveInfo } from 'graphql';
+
+/**
+ * Checks if a specific field is requested in the GraphQL query selection set.
+ * Looks for the field within the 'items' selection of a paginated list.
+ */
+export function isFieldInSelection(
```
michaelbromley, line 8: I'd like to see a test suite for this helper - ensuring it will handle various types of query document, including where the type is a fragment spread.

packages/core/src/api/resolvers/admin/collection.resolver.ts
```diff
@@ -56,8 +60,15 @@ export class CollectionResolver {
...
         })
         relations: RelationPaths<Collection>,
+        @Info() info: GraphQLResolveInfo,
     ): Promise<PaginatedList<Translated<Collection>>> {
-        return this.collectionService.findAll(ctx, args.options || undefined, relations);
+        const collections = await this.collectionService.findAll(ctx, args.options || undefined, relations);
+        // Only cache collection IDs if productVariantCount is requested in the query
+        if (isFieldInSelection(info, 'productVariantCount')) {
+            const collectionIds = collections.items.map(c => c.id);
+            this.requestContextCache.set(ctx, 'CollectionService.collectionIds', collectionIds);
```
michaelbromley: Almost. Rather than just caching the IDs here we should be caching the promise generated by `collectionService.getProductVariantCounts()`. Then in the `Collection.productVariantCount` resolver, you just need to await that promise from the RequestContextCache (if it exists), and then extract the correct count from the result.

As it stands, we're not benefitting from firing off the count query to the DB ahead of time.

packages/core/src/api/common/is-field-in-selection.ts
biggamesmallworld: added a test for this 👍

packages/core/src/api/resolvers/admin/collection.resolver.ts
biggamesmallworld: Ok got it, I thought it would be enough to fill the map with the promises one step down the tree but I've adapted it so that it is handled here in the collection resolver

PR #4151, opened 2026-01-23
fix(dashboard): Hide bulk action bar when no selected items are visible

### LINKED ISSUE #4007: Action selection box remains visible after filtering selected items out of the list
When selecting multiple entries in a list view and then applying a filter that hides those selected entries, the action selection box (“X selected – actions…”) remains visible. It only disappears after making the previously selected entries visible again (e.g., by removing filters) and manually deselecting them.

**To Reproduce**

1. Go to any list view in the Vendure Admin UI (e.g., Products, Orders, Customers).
2. Select multiple entries using the checkboxes.
3. Apply a filter that hides these entries.
4. Observe that the action selection box at the top remains visible.
5. Remove the filter so the items reappear.
6. Deselect the items → only now does the action box disappear.

---

### **Primary expected behavior**

The action selection box should automatically disappear as soon as the selected items are no longer visible in the filtered list.

### **Alternative expected behavior**

The “X selected entries” box provides a clear UI option to *manually deselect all selected items* directly within the box — even if the items are no longer visible.
Examples could include:

* A “Deselect all” button
* A small “Clear selection” icon
* A text link like “Remove selection”

---

## **Actual behavior**

The action selection box stays visible even though none of the selected entries are currently displayed, leading to confusion and risk of accidentally performing bulk actions on invisible items.

---

* @vendure/core version: 3.5.1
* Nodejs version: 22
* Database (mysql/postgres etc): postgres
* Operating System: macOS
* Browser: Safari
* Package manager: npm

jantokic:
Fixes #4007

## Problem

When users select items in a data table and then apply a filter that hides those items, the floating action bar ("X selected") remains visible even though none of the selected items are displayed. This creates a confusing UX.

## Solution

Hide the bulk action bar when no selected items are currently visible in the filtered view, while preserving the selection state internally.

This approach is better than clearing selection on filter change because it preserves a common workflow: selecting items from different parts of a large list by using filters to navigate. For example, with 260 countries, a user might select Argentina first, then filter to "Zimbabwe" to avoid scrolling—clearing the selection would break this.

## Changes

- Added `visibleSelectedCount` derived from `table.getRowModel().rows` to track how many selected items are currently visible
- Use `visibleSelectedCount` instead of `selection.length` for the floating action bar visibility check

## Demo

https://github.com/user-attachments/assets/0e2f9b20-7e84-46a6-8459-db4bbd5fb48b

### DIFF
--- a/packages/dashboard/src/lib/components/data-table/data-table-bulk-actions.tsx
+++ b/packages/dashboard/src/lib/components/data-table/data-table-bulk-actions.tsx
@@ -27,6 +27,7 @@ export function DataTableBulkActions<TData>({
     // Cache to store selected items across page changes
     const selectedItemsCache = useRef<Map<string, TData>>(new Map());
     const selectedRowIds = Object.keys(table.getState().rowSelection);
+    const visibleSelectedCount = table.getRowModel().rows.filter(row => row.getIsSelected()).length;
 
     // Get selection from cache instead of trying to get from table
     const selection = selectedRowIds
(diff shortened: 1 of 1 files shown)


jantokic:

PR #4157, opened 2026-01-26
fix(dashboard): Fix entity creation when using a non default language

tbouliere-datasolution:
# Description

Link to https://github.com/vendurehq/vendure/issues/4144

When the Dashboard UI language is set to a language other than English, creating a new product causes the form to crash. As soon as the user starts typing in any input field, the form becomes invalid and the UI breaks.


tbouliere-datasolution:

michaelbromley:
Thank you for the fix! 🙏

Issue #4161, opened 2026-01-27
Create draft order button doesn't offer any information about what is stopping the draft from being created.


HouseinIsProgramming:
**Describe the bug**

While trying to create a draft order, it is somewhat unintuitive to figure out what the offer is still missing for it to be created. 

**To Reproduce**

1. Go to the Admin Dashboard
2. Navigate to Sales > Orders, then click on "+ Draft Offer" 
3. Fill out **just** the customer and products.
4. Try to click the create draft offer

**Expected behavior**
Be warned about missing fields, eg, Shipping method, shipping, and billing address in this case, to avoid confusion

**Actual behavior**
Create draft offer is greyed out without any indication of what the form still requires to be filled out, be confused.

**Screenshots/Videos**

![Image](https://github.com/user-attachments/assets/fecfb75a-4930-4e5d-9854-b89f7ac29e3d)

**Environment (please complete the following information):**

- @vendure/core version: any version with the new dashboard

**Minimal reproduction**
A plain vendure installation/dev server

mohamed7-dev:
I’ve started working on this UX issue for draft orders.

Currently, the Complete draft button is disabled when required data is missing (customer, order lines, shipping address/method, or non‑Draft state), but the UI doesn’t clearly explain why.

**My approach is**:

- Keep the existing validation logic for when a draft can be completed.
- Compute a single, human‑readable reason explaining what’s missing (e.g. “Select a customer to continue”, “Add at least one item to the order”, “Set a shipping address and select a shipping method”).
- Show this reason in a prominent alert at the top of the draft‑order page while the button is disabled.

This avoids changing the underlying business rules, but makes it obvious to the user what they still need to fill in before the draft can be completed. I’ll open a PR with this change shortly.

michaelbromley:
Fixed in https://github.com/vendurehq/vendure/pull/4163

PR #4163, opened 2026-01-27
fix(dashboard): Clarify draft order completion UX and add ar translations

mohamed7-dev:
# Description

This PR improves the UX of completing draft orders in the new dashboard.
- On the draft order page, when the **Complete draft** button is disabled, a prominent alert is shown at the top of the page explaining *why* the draft cannot be completed.
- The alert surfaces the first unmet requirement, e.g.:
  - `Select a customer to continue`
  - `Add at least one item to the order`
  - `Set a shipping address and select a shipping method`
  - `Only draft orders can be completed`
- The underlying completion rules are unchanged: a draft can only be completed when a customer, at least one line, a shipping address + method, and the `Draft` state are all present.
- Adds Arabic (`ar`) translations for the new draft-order messages.

**Related issue**: `#4161`.

# Breaking changes

- No.  
  This is a purely UX-level change in the dashboard:
  - It does not alter API contracts, DB schema, or order state transitions.
  - It only adds UI messaging and translations on top of existing behavior.

# Screenshots

- **Desktop – draft not ready**
  - Draft order with missing customer / items / shipping
  - “Draft order status” block in side column shows: “Order draft isn't ready to be completed”
  - “Complete draft” button is disabled
error-status

- **Desktop – draft ready**
  - Draft order with customer, items, and shipping set
  - “Draft order status” block shows: “Order draft is ready to be completed”
  - “Complete draft” button is enabled
success-status

- **Mobile**
small-screen

# Checklist

📌 Always:

👍 Most of the time:


mohamed7-dev:
I am totally comfortable making these changes, i am going to code them and share a screenshot after that

mohamed7-dev:
I’ve updated the screenshots section to show the new draft-order status behaviour on desktop.

One limitation I ran into: 
- [PageLayout] currently controls how [PageBlock]s are arranged on mobile vs desktop. 
- There isn’t a built-in way to move a specific side-column block (like the status alert) to the very top of the page on small screens while still letting [PageLayout] manage the blocks.

**A naive workaround** would be to duplicate the status block and hide/show each copy with Tailwind breakpoints, but that can’t be perfectly synced with [PageLayout]'s own breakpoint logic.

**A more robust** architectural option (for a follow-up) would be:

- Expose the [PageLayout] breakpoint via a hook (e.g. [usePageLayoutBreakpoint] and/or small helpers like [PageLayoutDesktopOnly] / [PageLayoutMobileOnly].
- Use those helpers in page components or small wrappers to decide when to render certain [PageBlock]s, keeping them aligned with [PageLayout]’s internal breakpoint.

**I haven’t implemented** this architectural change in this PR, but I wanted to flag it as a potential direction if you think it would be useful more broadly.

packages/dashboard/src/app/routes/_authenticated/_orders/orders_.draft.$id.tsx
```diff
@@ -289,6 +290,31 @@ function DraftOrderPage() {
...
+        completeDraftDisabledReason = t`Select a customer to continue`;
+    } else if (!hasLines) {
+        completeDraftDisabledReason = t`Add at least one item to the order`;
+    } else if (!hasShippingMethod) {
+        completeDraftDisabledReason = t`Set a shipping address and select a shipping method`;
+    } else if (!isDraftState) {
+        completeDraftDisabledReason = t`Only draft orders can be completed`;
+    }
+
+    let alertVariant: 'default' | 'destructive' = 'default';
```
michaelbromley: There is a bunch of unused variables here. Actually I have an architectural suggestion I'd like to see:

1. The `StatusAlert` should be removed from `/lib/components` - let's keep it local to this route since it's not used anywhere else and we don't want to start prematurely abstracting components.
2. Move it into `packages/dashboard/src/app/routes/_authenticated/_orders/components` and put all the logic about selecting the `completeDraftDisabledReason` message inside that component. Name the component `DraftOrderStatus`.
3. The `DraftOrderStatus` would accept the `hasCustomer` etc as props.
4. Make the "success" icon a checkmark rather than info icon. Also make it green ("success" class I think).

mohamed7-dev:
I have applied all the changes requested: 

- Moved status logic into a local DraftOrderStatus component.
- Removed the shared StatusAlert.
- Placed the status block first in PageLayout so it appears first on small screens.
- Switched to a green check icon and title for the ready state.

i think it's ready now, i would be happy if you asked for further tweaking naming or placement.

mohamed7-dev:
- i have resolved all conflicts in en, and ar locales files.
- reset all other locales to the master branch.

I don't know if there is anything else missing in order for this PR to be merged correctly.

michaelbromley:
Thanks again for the quick action! I'm happy we can release this in the next patch! 🥳

PR #4165, opened 2026-01-27
chore: Add Claude Code rules for dashboard i18n

HouseinIsProgramming:
## Summary

Adds `.claude/rules/dashboard.md` with i18n guidelines for the React dashboard.

This addresses an issue I noticed multiple times where Claude Code misses adding translations in obvious spots where they should be used. There are also existing places in the codebase where translations are already missing.

The rules file:
- Uses path filtering to only load when working with `packages/dashboard/**` files
- Documents the `<Trans>` and `t\`...\`` patterns from `@lingui/react/macro`
- Reminds to never hardcode English strings in config values (dropdowns, reason lists, etc.)

## Test plan

- [ ] Verify `.claude/rules/dashboard.md` is loaded when editing dashboard files

### DIFF
--- a/.claude/rules/dashboard.md
+++ b/.claude/rules/dashboard.md
@@ -1,0 +1,28 @@
+---
+paths:
+  - "packages/dashboard/**/*.tsx"
+  - "packages/dashboard/**/*.ts"
+---
+
+# Dashboard Development
+
+## i18n / Translations
+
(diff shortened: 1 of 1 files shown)


HouseinIsProgramming:
Closing in favor of a new PR rebased on master

PR #4170, opened 2026-01-28
feat(dashboard): Add disabled option to exclude columns in ListPage

### LINKED ISSUE #3972: feat(dashboard) Disable columns in ListPage component
The ListPage component for displaying data results will take the graphql definition to generate what columns should belong to the list. And if you want extra data for, lets say calculations. That data will be added and displayed in the columns. 

And the only way now is to hide those columns by using 'defaultVisibility'. And it will also force filters to be available for the extra data. "but filters can be disabled by using 'additionalColumns'"

I would like the possibility / option to have more control of extra data that is provided. But that i don't want to be automatically displayed or filtered

Naming for properties are only for the examples

Option 1:
Have an argument in the the ListPage component, that will define what columns should be created for the table:
```
<ListPage
    listQuery={queryList} 
    listQueryColumns={[
        'col1', 'col2' // Allow list for columns that should be used from the listQuery
    ]}
/>
```

Option 2:
Add extra property to the argument 'customizeColumns' to will allow to disable the column :
```
<ListPage
    listQuery={queryList} 
    customizeColumns={{
        col1: { disabled: true } // Add property disabled
    }}
/>
```

BibiSebi:
# Description

Adds a `meta.disabled` option to `customizeColumns` in the `ListPage` component. When set to `true`, the column is completely excluded from the table and the column visibility toggle.

This is useful when you need to fetch certain fields in the query for use in custom cell renderers, but don't want those fields to appear as their own columns.

This PR also includes documentation updates for the new feature.

Fixes #3972

# How it was tested

Tested manually using the reviews test plugin in dev-server with `meta: { disabled: true }` on the `downvotes` columns.

https://github.com/user-attachments/assets/0b030d9c-ff9c-44a9-8522-7ad79c1d4285

# Breaking changes

None

# Checklist

📌 Always:

👍 Most of the time:


BibiSebi:
I explored adding Storybook interaction tests to the CI pipeline, but this would be better addressed in a separate PR to keep this one focused on the disabled columns feature.

I still extended the storybook stories but there is no automated tests/checks atm.

michaelbromley:
💪  Good stuff!

PR #4176, opened 2026-01-29
fix(core): Correctly publish ChangeChannelEvent with new channel IDs.

### LINKED ISSUE #4174: core: ChangeChannelEvent Fired Incorrectly When No New Channels Are Assigned
**Describe the bug**
The `ChangeChannelEvent` is being fired in the `assignToChannels` method even when no new channels are actually assigned to an entity. The method filters out channels that are already assigned (creating a `newChannelIds` array), but still publishes the event with the original `channelIds` parameter regardless of whether `newChannelIds` is empty or not.

**Expected behavior**
The `ChangeChannelEvent` should only be published when channels are actually added to or removed from an entity. If `newChannelIds` is empty (meaning all requested channels are already assigned), no event should be fired.

**Actual behavior**
The `ChangeChannelEvent` is published with the full `channelIds` array regardless of whether any actual assignment occurred. This happens at line 213 in `channel.service.ts`:

https://github.com/vendurehq/vendure/blob/5296c674e4e61338e36bd9dda2a720eb7467a319/packages/core/src/service/services/channel.service.ts#L202-L213

The database operation correctly uses newChannelIds, but the event is published with channelIds unconditionally.

kevmtt:
# Description

Correctly publish ChangeChannelEvent with new channel IDs.
Fixes #4174

# Breaking changes

No - Except someone was relying on this event to actually track `assignToChannels` function calls instead of new channels being assigned

# Checklist

📌 Always:

👍 Most of the time:

## Release Notes

* **Bug Fixes**
  * Channel event payloads now accurately reflect only the channels that were actually modified.

* **Chores**
  * Optimized channel assignment operations.


michaelbromley:
Thanks!

PR #4192, opened 2026-01-31
feat(core): Add anonymous telemetry collection module

dlhck:
## Summary

Adds a telemetry module that collects anonymous usage data on Vendure application startup and sends it to the Vendure telemetry endpoint. This helps the Vendure team understand how the framework is being used and prioritize development efforts.

### Key behaviors
- **Fire-and-forget on bootstrap** - Non-blocking, 5-second timeout
- **Opt-out via `VENDURE_DISABLE_TELEMETRY=true`**
- **Auto-skip in CI environments** - Detects common CI env vars
- **Fails silently** - No logs, no errors visible to users
- **Only runs from server process** - Not from worker
- **Telemetry status logged on startup** - Users are informed whether telemetry is enabled or disabled in the server bootstrap banner
- **Database-backed installation ID** - Persists across container restarts
- **Hardcoded endpoint** - The telemetry endpoint cannot be overridden

### Installation ID resolution flow

The installation ID is resolved using a DB-primary strategy with filesystem fallback. This ensures a stable ID in containerized deployments (Docker, K8s) where the filesystem is ephemeral.

```
                        collect()
                           |
                    +------v------+
                    | cached ID?  |
                    +------+------+
                      yes / \ no
                     /       \
                return     +--v--------------+
                           | DB initialized? |
                           +--+-----------+--+
                         yes /             \ no
                            /               \
                  +---------v---------+      |
                  | read from DB      |      |
                  | (SettingsStore)    |      |
                  +---------+---------+      |
                   found  / \ not found      |
                         /   \               |
                  cache &     +---+----------+
                  return          |
                           +------v-----------+
                           | read from        |
                           | filesystem       |
                           | (.vendure/)      |
                           +------+-----------+
                            found / \ not found
                                 /   \
                    +-----------v+    +v-----------+
                    | save to DB |    | generate   |
                    | (migrate)  |    | new UUID   |
                    | cache &    |    +---+--------+
                    | return     |        |
                    +------------+   +----v--------+
                                     | save to DB  |
                                     | save to FS  |
                                     | cache &     |
                                     | return      |
                                     +-------------+

  Notes:
  - All DB operations are best-effort (wrapped in try/catch)
  - If DB read fails, falls through to filesystem
  - If DB write fails, continues silently
  - If FS write fails, continues silently
  - Once cached, subsequent calls skip all I/O
```

### Startup banner output

When enabled:
```
=========================================================
   Vendure server (v3.5.2) now running on port 3000
---------------------------------------------------------
Shop API:   http://localhost:3000/shop-api
Admin API:  http://localhost:3000/admin-api
=========================================================
Anonymous telemetry is enabled to help us improve Vendure.
To disable, set VENDURE_DISABLE_TELEMETRY=true.
```

When disabled:
```
=========================================================
   Vendure server (v3.5.2) now running on port 3000
---------------------------------------------------------
Shop API:   http://localhost:3000/shop-api
Admin API:  http://localhost:3000/admin-api
=========================================================
Anonymous telemetry is disabled.
```

### Data collected
- Installation ID (random UUID, persisted in database via SettingsStore with filesystem fallback)
- Vendure version, Node version, platform
- Database type and entity count ranges (using buckets like `1-100`, `1k-10k`)
- Official plugin names (custom plugin names are NOT collected, only count)
- Configuration strategy class names
- Deployment environment (container, cloud provider, serverless)

### Privacy guarantees
- Installation ID is a random UUID, not derived from system info
- Custom plugin names are never collected (only count)
- Entity counts use range buckets, not exact numbers
- No PII (no hostnames, IPs, user data)
- Telemetry endpoint is hardcoded and cannot be overridden via environment variables

### Configuration
- `VENDURE_DISABLE_TELEMETRY=true` - Disable telemetry

## Test plan
- [x] Start dev server, verify telemetry is sent (check server logs)
- [x] Set `VENDURE_DISABLE_TELEMETRY=true`, verify no request is made
- [x] Set `CI=true`, verify no request is made
- [x] Verify installation ID is stored in database (SettingsStore)
- [x] Verify filesystem ID is migrated to database on first read
- [x] Verify new installation generates UUID saved to both DB and filesystem
- [x] Verify DB-unavailable fallback reads from filesystem
- [x] Verify DB-error fallback reads from filesystem
- [x] Verify telemetry endpoint is hardcoded and not overridable
- [x] Restart server, verify same installation ID is reused
- [x] Verify telemetry status message appears in bootstrap banner
- [x] Verify "disabled" message when `VENDURE_DISABLE_TELEMETRY=true`
- [x] All 678 unit tests pass

Closes OSS-55

### DIFF
--- a/packages/core/src/telemetry/collectors/config.collector.ts
+++ b/packages/core/src/telemetry/collectors/config.collector.ts
@@ -94,9 +94,10 @@ export class ConfigCollector {
  * then falls back to `constructor.name`. Returns 'unknown' if the name
  * appears to be minified (single char or empty).
  */
-function getStrategyName(strategy: { name?: string; constructor?: { name?: string } }): string {
-    if (typeof strategy.name === 'string' && strategy.name.length > 1) {
-        return strategy.name;
+function getStrategyName(strategy: object): string {
+    const name = (strategy as any).name;
+    if (typeof name === 'string' && name.length > 1) {
+        return name;
     }
     const ctorName = strategy.constructor?.name;
     if (ctorName && ctorName.length > 1) {
--- a/packages/core/src/telemetry/collectors/config.collector.spec.ts
+++ b/packages/core/src/telemetry/collectors/config.collector.spec.ts
@@ -7,7 +7,7 @@ import { ConfigCollector } from './config.collector';
 
 describe('ConfigCollector', () => {
     let collector: ConfigCollector;
-    let mockConfigService: Partial<ConfigService>;
+    let mockConfigService: Record<string, any>;
 
     beforeEach(() => {
         mockConfigService = {
--- a/packages/core/src/telemetry/collectors/database.collector.spec.ts
+++ b/packages/core/src/telemetry/collectors/database.collector.spec.ts
@@ -8,7 +8,7 @@ import { DatabaseCollector } from './database.collector';
 
 describe('DatabaseCollector', () => {
     let collector: DatabaseCollector;
-    let mockConfigService: Partial<ConfigService>;
+    let mockConfigService: Record<string, any>;
     let mockConnection: Partial<TransactionalConnection>;
     let mockRepository: { count: ReturnType<typeof vi.fn> };
 
--- a/packages/core/src/telemetry/collectors/deployment.collector.spec.ts
+++ b/packages/core/src/telemetry/collectors/deployment.collector.spec.ts
@@ -10,9 +10,9 @@ import { CLOUD_PROVIDERS, DeploymentCollector, SERVERLESS_ENV_VARS } from './dep
 vi.mock('fs');
 describe('DeploymentCollector', () => {
     let collector: DeploymentCollector;
-    let mockProcessContext: Partial<ProcessContext>;
-    let mockConfigService: Partial<ConfigService>;
-    let mockJobQueueService: Partial<JobQueueService>;
+    let mockProcessContext: Record<string, any>;
+    let mockConfigService: Record<string, any>;
+    let mockJobQueueService: Record<string, any>;
     let mockFs: typeof import('fs');
 
     // Collect all env vars that need to be cleaned (from source of truth)
--- a/packages/core/src/telemetry/collectors/plugin.collector.spec.ts
+++ b/packages/core/src/telemetry/collectors/plugin.collector.spec.ts
@@ -7,7 +7,7 @@ import { PluginCollector } from './plugin.collector';
 
 describe('PluginCollector', () => {
     let collector: PluginCollector;
-    let mockConfigService: Partial<ConfigService>;
+    let mockConfigService: Record<string, any>;
     const addedCacheKeys: string[] = [];
 
     beforeEach(() => {
--- a/packages/core/src/telemetry/telemetry.service.spec.ts
+++ b/packages/core/src/telemetry/telemetry.service.spec.ts
@@ -17,7 +17,7 @@ vi.mock('./helpers/ci-detector.helper', () => ({
 
 describe('TelemetryService', () => {
     let service: TelemetryService;
-    let mockProcessContext: Partial<ProcessContext>;
+    let mockProcessContext: Record<string, any>;
     let mockInstallationIdCollector: Partial<InstallationIdCollector>;
     let mockSystemInfoCollector: Partial<SystemInfoCollector>;
     let mockDatabaseCollector: Partial<DatabaseCollector>;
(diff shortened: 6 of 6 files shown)


dlhck:
@coderabbitai review

packages/core/src/telemetry/collectors/database.collector.ts
```diff
@@ -0,0 +1,120 @@
...
+    private getDatabaseType(): SupportedDatabaseType {
+        const dbType = this.configService.dbConnectionOptions.type;
+        if (dbType === 'better-sqlite3' || dbType === 'sqlite') {
+            return 'sqlite';
+        }
+        if (dbType === 'postgres' || dbType === 'mysql' || dbType === 'mariadb') {
+            return dbType;
+        }
+        return 'postgres';
+    }
```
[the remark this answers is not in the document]
biggamesmallworld: @dlhck I think this is a valid comment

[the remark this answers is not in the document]
dlhck: @biggamesmallworld already fixed.

packages/core/src/job-queue/job-queue.service.ts
```diff
@@ -52,6 +52,14 @@ export class JobQueueService implements OnModuleDestroy {
     private queues: Array<JobQueue<any>> = [];
     private hasStarted = false;
 
+    /**
```
biggamesmallworld, line 55: question: is this relevant to the PR?

dlhck: yes. We need to send telemetry after the job queue has started to determine if the project runs an integrated or separate worker/server setup.

packages/core/src/telemetry/telemetry.service.ts
```diff
@@ -0,0 +1,140 @@
...
+            this.sendTelemetry().catch(() => {
+                // Silently ignore all errors
+            });
+        }, 5000);
+    }
+
+    /**
+     * Checks if telemetry is disabled via environment variable or CI detection.
+     */
+    private isTelemetryDisabled(): boolean {
```
biggamesmallworld: nit: is this wrapper necessary?

biggamesmallworld (approved):
This looks great. The only question I have remaining is this; due to us capturing these metrics on each startup, is the application the data is being sent to intended to or capable of detecting changes or trends over time? or will it just update an existing record to reflect the current state of the existing installations?

Obviously not blocking, but I'm interested in how to best use this data.

dlhck:
> This looks great. The only question I have remaining is this; due to us capturing these metrics on each startup, is the application the data is being sent to intended to or capable of detecting changes or trends over time? or will it just update an existing record to reflect the current state of the existing installations?
> 
> Obviously not blocking, but I'm interested in how to best use this data.

We are de-duplicating based on the payload. If there are changes we capture them into separate events which are uniquely identified by the installation ID which should be the same.

packages/core/src/telemetry/telemetry.service.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        // Skip if disabled or CI environment
+        if (isTelemetryDisabled()) {
+            return;
+        }
+
+        // Delay telemetry collection to allow user bootstrap code to complete
+        // This ensures JobQueueService.start() has been called (if it will be)
+        // before we check worker mode
+        setTimeout(() => {
```
michaelbromley: I think the handle of the timeout should be saved and then explicitly cleared in `onApplicationShutdown()`. Leaving the 5s timeout might force a delay during graceful shutdown.

packages/core/src/telemetry/helpers/is-telemetry-disabled.helper.ts
```diff
@@ -0,0 +1,10 @@
+import { isCI } from './ci-detector.helper';
+
+/**
+ * Checks if telemetry is disabled via the VENDURE_DISABLE_TELEMETRY environment
+ * variable or CI environment detection.
+ */
+export function isTelemetryDisabled(): boolean {
+    const disableEnv = process.env.VENDURE_DISABLE_TELEMETRY;
+    return disableEnv === 'true' || disableEnv === '1' || isCI();
```
michaelbromley, line 9: Might be worth calling `.toLowerCase()` on the `disableEnv`, to catch `'True'` and `'TRUE'` as well (i know, not documented to work with those... but someone will surely try it)

packages/core/src/telemetry/collectors/config.collector.ts
```diff
@@ -0,0 +1,89 @@
...
+        } catch {
+            return 'unknown';
+        }
+    }
+
+    private getEntityIdStrategy(): string {
+        try {
+            const strategy =
+                this.configService.entityOptions.entityIdStrategy ?? this.configService.entityIdStrategy;
+            return strategy?.constructor.name ?? 'unknown';
```
michaelbromley: I wonder if this would be affected by code that gets minified, e.g. an Nx project that bundles the server code with Webpack?

dlhck: I don't know. That's why it needs to be tested in a production environment with bundling etc.

PR #4195, opened 2026-02-02
feat(dashboard): Improve product variant handling by providing enhanced "creation assistant" dialog

lucatk:
# Description

Right now, product option (groups) and variant creation and management is a little confusing and error-prone UX-wise.
It's easy to create duplicate option groups and difficult to delete these, while it's difficult to efficiently create product variants for new options added at a later time, when some variants exist already.

The functionality of the buttons "Create Variant" and "Manage Variants" is confusing - the former only appearing if no variants exist, the latter only appearing if there are variants.
A video is attached displaying some of the issues with the dialog and what happens if option groups existed and then all variants are deleted and readded.
The Manage Variants page on the other hand doesn't provide any added value regarding variants, it rather enables managing the option groups and options.
The Manage Variants page also provides another dialog for adding a variant, adding to the confusion.

Therefore, I want to propose:
- enhancing the "Create Variant" dialog into an "assistant" displaying already existing option groups, enabling creation of new options (groups), and intelligently creating variants for new option combinations
- replacing /variants page with /options page, only keeping the option group management and moving the "manage" button to the product options block
- adding the missing delete action for option groups in /options
- moving the other "manually add variant" dialog to the product detail page (if even desired to keep)

# Screenshots

Current state:
[Untitled.webm](https://github.com/user-attachments/assets/4852c66c-da21-42d3-96c9-d784cd00685c)

Updated:
image
image
image

# Checklist

📌 Always:
- [ ] [Tbh, not sure here]  My PR is small and contains a single feature

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_products/components/generate-variants-panel.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_products/components/generate-variants-panel.tsx
@@ -110,22 +110,43 @@ export function GenerateVariantsPanel({
     productId,
     productName,
     optionGroups,
+    existingVariants,
     onSuccess,
     onBack,
+    additionalActions,
 }: Readonly<{
     productId: string;
     productName: string;
     optionGroups: OptionGroup[];
+    existingVariants?: Array<{ options: Array<{ id: string }> }>;
     onSuccess?: () => void;
     onBack?: {
         handler: () => void;
         confirmation?: { title: string; description: string };
     };
+    additionalActions?: React.ReactNode;
 }>) {
     const { t } = useLingui();
     const { activeChannel } = useChannel();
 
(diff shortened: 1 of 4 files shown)


biggamesmallworld:
Closing this — the functionality has since landed, though not through this PR, and that's on us rather than you.

`GenerateMissingVariantsDialog` (#4983, on `minor`) covers the core of what you proposed: after adding a new option value it creates only the combinations that don't exist yet, with the existing ones shown for context. The option-group removal you added here landed separately in #4876 on `master`.

Two things worth saying. First, you identified this problem in February and it took us until July to fix it — in a separate PR, without linking back here. Second, you rebased in June and #4876 then landed in July and overwrote two of your four files, which wasn't your doing.

The one part not covered anywhere is the `/variants` → `/options` split you originally proposed and then stripped out. If you still think that's the right structure, open an issue for it and we'll give you a direction before you write code this time.

PR #4198, opened 2026-02-03
fix(dashboard): Fix required-field validation on the channel create form

### LINKED ISSUE #4173: Creating a new channel in the dashboard with missing fields give unformatted gql errors
## Describe the bug
A user-facing error occurs when creating a new channel in the admin UI and returns a raw gql error.
Missing fields are also not highlighted on failure (would be good UX) 

## To Reproduce
1. Go to Settings > Channels
2. Click "New channel"
3. Fill in some of the required fields
4. Attempt to save/create the channel
5. See error 

## Expected behavior
Error should be readable and missing fields should be obvious

## Actual behavior
A raw gql error is toasted

## Screenshots/Videos

Image

## Environment
Any

biggamesmallworld:
# Description

Fixes #4173

Creating a channel with missing fields toasted a raw GraphQL error, and the offending fields were not highlighted.

## Root cause

`CreateChannelInput` declares `defaultTaxZoneId: ID!` and `defaultShippingZoneId: ID!`. The generated form seeds them with `''`, which passes the generated `z.string()`. `removeEmptyIdFields` then **strips** them from the payload, so the mutation is sent without two required variables and fails during server-side GraphQL variable coercion — hence the raw toast. Because nothing failed Zod, no field was marked invalid either. Both halves of the issue come from this one cause.

## Why there is no blanket `min(1)`

An earlier revision of this PR applied `z.string().min(1)` to every non-nullable `String`/`ID`/`DateTime`. That was wrong, and it regressed the collection, facet, promotion and payment-method create forms (thanks @gabriellbui for catching it). Required-ness simply cannot be inferred from the GraphQL type:

- **`String!` means "not null", not "not empty".** `CollectionTranslationInput.description` and `ConfigArgInput.value` are legitimately empty.
- **A non-nullable `ID!` is not necessarily user-supplied.** `CreateFacetValueInput.facetId` is injected by the page in `transformCreateInput`, so it is legitimately `''` in the form. (Restoring `min(1)` for non-nullable IDs alone still breaks facet-value creation — the e2e suite catches this.)
- Applying `min(1)` before the nullable wrap also rejects the `''` that `getDefaultValueFromField` seeds into *nullable* fields such as a translation's `id`, which is what made every create form with a translation block permanently invalid.

So `getZodTypeFromField` is unchanged from `master`: the generated schema stays permissive.

## The fix

Pages declare the fields they actually require, via a new optional `extendSchema` option on `useDetailPage` / `useGeneratedForm`:

```ts
extendSchema: schema =>
    schema.extend({
        code: z.string().min(1, { message: t`This field is required` }),
        token: z.string().min(1, { message: t`This field is required` }),
        ...(creatingNewEntity
            ? {
                  defaultTaxZoneId: z.string().min(1, { message: t`This field is required` }),
                  defaultShippingZoneId: z.string().min(1, { message: t`This field is required` }),
              }
            : {}),
    }),
```

This is the "explicit field-level required metadata" @dlhck asked for. The zones are only required on create, since `UpdateChannelInput` declares them as nullable `ID` and omits what isn't sent.

The channel Save button is also no longer gated on `formState.isValid`. With required fields, an untouched invalid form would otherwise leave the button silently disabled with no indication of what is missing — the same "submit button is always disabled" symptom seen earlier in this PR. `submitHandler` already validates and bails, so clicking Create now surfaces the inline field errors.

**Trade-off, stated plainly:** required-field validation is now opt-in per page. Other detail pages with user-facing required `ID!` relations keep the latent bug until they adopt `extendSchema`. That is inherent — the metadata needed to do it automatically does not exist in the introspected schema.

## Changed since the last review

- Merged `master`. **Dropped the theme refactor entirely** — dashboard themes now come from `@vendure-io/design-tokens` (#4575), so the block this PR edited no longer exists upstream. The destructive-contrast concern @dlhck raised is real, but it is pre-existing and lives in that package (`destructive[500] = oklch(0.55 0.2 25)` for both themes), so it is out of scope here.
- **Dropped the unrelated `docs/package-lock.json` changes.**
- Folded the vestigial `case 'ID'` back in with `String`/`DateTime` (@grolmus).
- The i18n catalogs show 28 changed files, but the only real addition is one new string (`msgid "This field is required"`). The rest is `i18n:extract` regenerating `#:` source-reference comments.

# Breaking changes

None. `extendSchema` is a new optional option on an existing public API.

# Testing

- **Unit** (`form-schema-tools.spec.ts`): specs corrected to assert the generated schema *accepts* `''` for non-nullable `String`/`ID`/`DateTime` (deliberate, per the above), plus a schema-vs-defaults contract test asserting a create form's own default values satisfy its own schema — the guard that would have caught the original regression — and a test for `extendSchema`.
- **E2E** (`e2e/tests/settings/channels.spec.ts`): reproduces #4173 — fill only `Code`, click Create, assert inline "This field is required" on Token / Default tax zone / Default shipping zone, and assert **no** error toast. Plus a test that the error clears once the field is filled.
- Full dashboard e2e suite: 289 passed / 0 failed on `master`, 291 passed / 0 failed on this branch (the 2 new tests). Collection, facet, promotion and payment-method create forms all pass.

# Checklist

📌 Always:

👍 Most of the time:


packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts
```diff
@@ -408,7 +421,9 @@ export function getZodTypeFromField(field: FieldInfo): ZodTypeAny {
         case 'String':
         case 'ID':
         case 'DateTime':
-            zodType = z.string();
+            zodType = field.nullable
+                ? z.string()
+                : z.string().min(1, { message: VALIDATION_MESSAGES.REQUIRED });
```
gabriellbui: I believe this change might have broken other forms. Like facets, promotion or collection creations as you can see in the [dashboard e2e tests](https://github.com/vendurehq/vendure/actions/runs/23003168235/job/66792740825?pr=4198)
I've tried out manually, the submit button is always disabled for the above mentioned forms, even if the required fields are filled out

biggamesmallworld:
Addressed @gabriellbui's feedback. Root cause: `ID` type fields (e.g. `translations[0].id`) default to `''` for new entities and are never user-filled — they're server-assigned. Applying `min(1)` to them caused create forms (facets, collections, promotions) to become permanently invalid once any field was touched.

Fixed by removing `ID` from the `min(1)` case — only `String` and `DateTime` fields get required validation now. Updated the test that expected `ID` to reject empty strings, and fixed the contract test to use `LanguageCode` type for the `languageCode` field (matching real-world usage).

packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts
biggamesmallworld: Thanks for catching this! You were right — the blanket `min(1)` on all non-nullable `String`/`DateTime` fields was too aggressive. In GraphQL, `String!` means "not null", but an empty string is still a valid non-null value. Fields like `description: String!` in `CollectionTranslationInput` and `value: String!` in `ConfigArgInput` legitimately default to empty strings, so the validation was keeping the Create button permanently disabled.

This has been addressed in 2a8c49dc7 — the `min(1)` validation has been removed from the auto-generated form schema for non-nullable String/DateTime fields.

biggamesmallworld:
@gabriellbui looking good now 🤝

gabriellbui (commented):
I tried it out, I'm not sure if the original issue is fixed, could be that it's only on my side. Could you check again?

https://github.com/user-attachments/assets/c4c5b718-0d7c-4c33-af0a-1fbddfe84f37

grolmus (dismissed):
Thanks @biggamesmallworld — the destructive-colour brightness bump for error visibility is a nice touch, and the `COLORS` / `generateCssVariables` extraction in the theme plugin reads well. But the PR as it stands doesn't match what it says on the tin, so I can't approve it yet.

### The PR no longer does what the title/description claim

Title: *"Improve required field validation"*; description: *"Add `min(1)` validation for non-nullable `String`/`ID`/`DateTime` fields, so required fields show inline 'This field is required' errors"* and *"Fixes #4173"*.

But the code does the **opposite**:
- `getZodTypeFromField` produces a plain `z.string()` for non-nullable `String`/`ID`/`DateTime` — no `min(1)` anywhere.
- The new spec **locks that in**: *"should accept empty strings for non-nullable String fields"*, *"…ID fields"*, *"…DateTime fields"* all assert `.parse('')` does **not** throw.
- `VALIDATION_MESSAGES` has `pattern`/`min`/`max`/`dateAfter`/`dateBefore` but **no `required`** message — the "This field is required" the description promises doesn't exist.

So **#4173 isn't fixed by this branch** — required fields still accept empty strings. I suspect the `min(1)` was dropped while resolving @gabriellbui's (valid) point that server-assigned `ID` fields default to `''`. That's the right instinct for `ID`, but it looks like it took `String`/`DateTime` down with it.

**Could you clarify the intent?**
- If accepting empty strings is now the deliberate outcome (the "GraphQL `String!` = not-null, not not-empty" argument is defensible), then please **retitle / rewrite the description and drop `Fixes #4173`** — the tests are then good regression coverage for that decision, but the issue stays open.
- If required validation is still the goal, please **restore `min(1)` for `String`/`DateTime` (excluding `ID`)**, add a `required` entry to `VALIDATION_MESSAGES`, and flip the specs to assert empty is rejected for those.

### Also before merge

1. **`docs/package-lock.json` (+8 / −539)** is unrelated to this change and shouldn't ride along — please drop it from the PR.
2. **Vestigial `ID` case.** `case 'ID': zodType = z.string()` is split out from `String`/`DateTime` but produces the identical result, so the split is currently dead. Either fold it back in, or (if `min(1)` returns for `String`/`DateTime`) let the separation actually carry the ID exception.
3. Minor: the theme refactor is unrelated to form validation — not a blocker, but it would be cleaner as its own PR (or at least call out the two concerns in the description).

Happy to re-review quickly once the intent is settled. Thanks!

dlhck (dismissed):
Thanks for the work here. I am requesting changes because the current head still does not resolve #4173, and the theme change introduces a separate accessibility regression.

## The required-field fix is no longer present

`getZodTypeFromField()` returns a plain `z.string()` for non-nullable `String` and `DateTime` fields, and the new specs explicitly assert that `''` is accepted for both. As a result, an incomplete channel form can still pass client-side validation, reach the mutation, and surface the raw GraphQL error reported in #4173.

The earlier blanket `min(1)` approach was also too broad: GraphQL nullability does not tell us whether an empty string is invalid user input. Fields such as descriptions may legitimately be empty even when represented as `String!`. Please add targeted validation for the actual required channel fields, or introduce explicit field-level required/non-empty metadata, and add a dashboard test covering the #4173 reproduction. Restoring `min(1)` globally for every non-nullable `String`/`DateTime` would reintroduce the collection, facet, promotion, and payment-method regressions already found on this PR.

## The destructive theme pair does not meet the claimed contrast

In the dark theme, `destructive: oklch(0.75 0.22 25)` is paired with the near-white `destructive-foreground`. Destructive buttons render that pair via `bg-destructive text-destructive-foreground`; its contrast is approximately 2.75:1, well below the 4.5:1 WCAG AA threshold for normal text. The light destructive-button pairing is also only approximately 4.26:1. Please validate the foreground/background token pairs as they are actually consumed, rather than only checking the destructive color against the page background.

## Cleanup before merge

- Revert the unrelated `docs/package-lock.json` changes.
- Drop or separate the unrelated theme refactor. This PR now conflicts with `master`, where dashboard themes are sourced from `@vendure-io/design-tokens`.
- Update the title and description if the scope changes; the current branch does not support the stated `Fixes #4173` claim.

biggamesmallworld:
Updated according to @dlhck and @grolmus feedback. Here is a screen recording of the form validation working correctly: 


https://github.com/user-attachments/assets/294f1bea-ce99-4108-803e-0cea4a900bf6

I've added some additional functionality on the zod validation to improve UX for the available currencies and default currencies. When one default currency is selected, it is automatically made an available currency. If the user selects multiple currencies, then they are required to select one as the default.

dlhck (dismissed):
Thanks, Will — the original review feedback is addressed well now: validation is scoped to the channel page, the issue has direct E2E coverage, and the unrelated theme and lockfile changes are gone. The full dashboard suite is green too.

There is one remaining blocker in the new currency flow, so I cannot approve this head yet.

## Keep the default currency inside the available currencies

The new UI allows this sequence:

1. Leave Available currencies empty.
2. Pick USD as Default currency from the unrestricted selector.
3. Add EUR to Available currencies.
4. Submit.

`transformCreateInput` now preserves the non-empty available list, so the mutation receives `defaultCurrencyCode: USD` and `availableCurrencyCodes: [EUR]`. `ChannelService.create()` uses a supplied list verbatim and, unlike `update()`, does not verify that it contains the default. The channel is therefore saved in an inconsistent state where its default currency is unavailable.

This also means the feedback comment that selecting a default automatically makes it available is not true once the user subsequently makes a non-empty available selection.

Please either synchronize the available list so it always contains the selected default, or add schema validation which prevents submission until the default is included. The same invariant should be considered for the language pair. Please add an E2E case covering a default selected before a different non-empty available list; the current new test only covers leaving the list empty, where the server fallback works correctly.

biggamesmallworld:
@dlhck I've implemented the changes from your feedback, when selecting a default currency we now add that to the list of available currencies if it is not yet in the list. Below is a screen recording. We still validate that if multiple available currencies are chosen, we force the user to select a default. I believe this addresses your feedback. 


https://github.com/user-attachments/assets/a6ad63d7-4b3f-4642-a74a-8d22674031b0

dlhck:
@biggamesmallworld I think the available currencies selected of a channel should restrict the items available in the default currency for channel. Same for the language. A user should not be able to select a currency as as default which isn't available for the channel

dlhck (approved):

biggamesmallworld:
Fair point, rebuilt this so that the default selector options are populated from the available list and we have a new error message that accurately represents the flow of requiring to select the available first, so there is at least one option, and then select the default


https://github.com/user-attachments/assets/fbfbb508-cdba-4cea-9ecb-d43db1f9fd41

PR #4201, opened 2026-02-04
feat(core,job-queue-plugin): Add per-queue concurrency configuration

biggamesmallworld:
# Description

Adds support for per-queue concurrency configuration across all job queue plugins. Previously, concurrency was a single number that applied to all queues. Now it can be a function that receives the queue name and returns the concurrency limit for that specific queue.

This is useful for limiting concurrency on resource-intensive queues like `apply-collection-filters` while allowing higher throughput on lighter queues.

The implementation includes an e2e test validating the job concurrency numbers. 

## Affected Packages

- `@vendure/core` - `DefaultJobQueuePlugin` (SQL-based strategy)
- `@vendure/job-queue-plugin` - `BullMQJobQueuePlugin` and `PubSubJobQueuePlugin`

## Usage Example

```typescript
// DefaultJobQueuePlugin
DefaultJobQueuePlugin.init({
  concurrency: (queueName) => {
    if (queueName === 'apply-collection-filters') {
      return 1; // Resource-intensive, limit to 1
    }
    return 3; // Default for other queues
  },
})

// BullMQJobQueuePlugin
BullMQJobQueuePlugin.init({
  concurrency: (queueName) => {
    if (queueName === 'apply-collection-filters') {
      return 1;
    }
    return 5;
  },
})

// PubSubJobQueuePlugin
PubSubJobQueuePlugin.init({
  concurrency: (queueName) => {
    if (queueName === 'apply-collection-filters') {
      return 1;
    }
    return 20;
  },
  queueNamePubSubPair: new Map([...])
})
```

## Implementation Details

- **Backwards compatible**: Passing a number still works as before
- **SQL strategy**: Each `ActiveQueue` instance now resolves its own concurrency value at construction time
- **BullMQ strategy**: When using function-based concurrency, creates separate workers per concurrency level (note: workers are grouped by concurrency value, not queue name - see JSDoc for details)
- **Pub/Sub strategy**: Applies concurrency per-queue via `flowControl.maxMessages`

# Breaking changes

None. The `concurrency` option type changed from `number` to `number | ((queueName: string) => number)`, which is backwards compatible.

# Screenshots

Manual testing was performed by adding the following to `packages/dev-server/dev-config.ts`:

```typescript
DefaultJobQueuePlugin.init({
    pollInterval: 200,
    concurrency: (queueName: string) => {
        const concurrency = queueName === 'apply-collection-filters' ? 1 : 3;
        console.log(`[Concurrency] Queue "${queueName}" → ${concurrency}`);
        return concurrency;
    },
}),
```

Console output when triggering collection filter jobs:

```
[Concurrency] Queue "apply-collection-filters" → 1
[Concurrency] Queue "update-search-index" → 3
```
Screenshot 2026-02-05 at 14 57 19

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have updated the README if needed (documentation is via JSDoc comments which generate to docs site)


michaelbromley:
Nice feature! Since this is a `feat` rather than a `fix`, could you retarget this PR to the `minor` branch? Thanks!

biggamesmallworld:
@michaelbromley sure! retargeted to minor

packages/job-queue-plugin/src/bullmq/bullmq-job-queue-strategy.ts
```diff
@@ -284,32 +292,84 @@ export class BullMQJobQueueStrategy implements InspectableJobQueueStrategy {
...
+        // If concurrency is a function, we create workers grouped by concurrency value.
+        // Note: Workers are stored in `this.workers` keyed by concurrency number, not queue name.
+        // All Vendure job types share a single BullMQ queue (`QUEUE_NAME`), so any worker can
+        // process any job type. This means multiple Vendure queues returning the same concurrency
+        // will share a worker, and the concurrency limit applies to total jobs processed by that
+        // worker—not strictly per Vendure queue. For strict per-queue isolation, use BullMQ Pro
+        // Groups or create separate BullMQ queues per Vendure queue.
+        if (typeof this.options.concurrency === 'function') {
+            const concurrency = this.options.concurrency(queueName);
+            if (!this.workers.has(concurrency)) {
```
michaelbromley: This section to me looks like unnecessary duplication. I would like to avoid this "modal" approach of having "this.worker" for one mode and "this.workers" for another mode.

Why not just always use this.workers, and in the case that there is no concurrency function, key the map using the default concurrent value?

biggamesmallworld: @michaelbromley removed the duplication and ran the tests, looking good for now 👍

PR #4202, opened 2026-02-04
fix(dashboard): Validate required args in ConfigurableOperationInput

### LINKED ISSUE #4148: Dashboad: Required flag in FulfillmentHandler is ignored in Dashboad
```
export const digitalFulfillmentHandler = new FulfillmentHandler({
  code: 'digital-fulfillment',
  args: {
    apiKey: { 
      type: 'string', 
      required:true // <--- This is ignored
     },
});
```
Expected behavior
The Dashboard UI should validate that the field is not empty before allowing the fulfillment transition, consistent with the required: true property.

grolmus:
## Summary

The Dashboard UI was ignoring the `required: true` flag on configurable operation args (e.g., FulfillmentHandler, ShippingCalculator, PaymentHandler, PromotionConditions, etc.), allowing form submission with empty required fields.

Fixes #4148

### Core Implementation
- **`ConfigurableOperationInput`** - Added validation logic that checks if all required (non-list) args have values, with `onValidityChange` callback to notify parent components. Also added visual red asterisk (`*`) indicator for required fields.
- **`ConfigurableOperationSelector`** - Propagates `onValidityChange` to child components
- **`ConfigurableOperationMultiSelector`** - Aggregates validity from multiple operations using `code+index` as map key (handles add/remove/reorder correctly)

### Wrapper Selectors (all now support `onValidityChange`)
- `ShippingCalculatorSelector`
- `ShippingEligibilityCheckerSelector`
- `PaymentHandlerSelector`
- `PaymentEligibilityCheckerSelector`
- `PromotionActionsSelector`
- `PromotionConditionsSelector`
- `CollectionFiltersSelector`

### Detail Pages (all now validate required args before submit)
- Shipping methods (`checker` + `calculator`)
- Payment methods (`checker` + `handler`)
- Promotions (`conditions` + `actions`)
- Collections (`filters`)
- Fulfill order dialog (`handler`)

## Technical Notes

- **List args exemption**: List-type args are exempt from required validation (matching legacy Angular admin-ui behavior) because an empty array `[]` is considered a valid value.
- **Multi-selector validity tracking**: Uses `code+index` as the validity map key instead of just index. This correctly handles cases where operations are removed from the middle (indices shift) or when multiple operations of the same type exist.
- **FulfillmentHandlerSelector**: Only stores handler CODE on shipping methods page - the actual args are entered during fulfillment action in `fulfill-order-dialog.tsx`.

## Code Review Summary

The implementation was reviewed using automated code review (Nigel). Key findings:

✅ **Approved for merge**
- Validity tracking correctly uses `code+index` as map key
- Validity map clears when operations change (prevents stale state)
- Consistent propagation pattern through component hierarchy
- Visual asterisk indicator is good UX
- Proper cleanup on dialog cancel

Minor observations (not blocking):
- `setTimeout` in `handleOpen` for dialog initialization (works but could theoretically race)
- Unused internal `useForm` in `ConfigurableOperationInput` (harmless)

## Test Plan

- [ ] Create a FulfillmentHandler/ShippingCalculator/PaymentHandler with `required: true` arg
- [ ] Open corresponding detail page in Dashboard
- [ ] Verify asterisk (`*`) appears next to required field label
- [ ] Verify submit/update button is disabled when required field is empty
- [ ] Verify button enables when required field has value
- [ ] Test multi-selector: add multiple conditions, remove one from middle, verify form still validates correctly
- [ ] Test fulfill order dialog with handler that has required args

### DIFF
--- a/packages/dashboard/src/lib/components/shared/configurable-operation-selector.tsx
+++ b/packages/dashboard/src/lib/components/shared/configurable-operation-selector.tsx
@@ -114,6 +114,7 @@ export function ConfigurableOperationSelector({
 
     const onOperationRemove = () => {
         onChange(undefined);
+        onValidityChange?.(true);
     };
 
     const operationDef = operations?.find((op: ConfigurableOperationDefFragment) => op.code === value?.code);


grolmus:
perfect, i am glad that it is working :)

PR #4204, opened 2026-02-04
feat(core): Introduce BootstrappedEvent to signal server readiness

### LINKED ISSUE #4203: Separate bootstrap event to signal server readiness

During the development of features with extensive database preloading (e.g., checking roles, configurations, fetching tokens, etc.), issues arose with Kubernetes liveness probes. The bootstrap process took so long that liveness probes frequently timed out before the application was ready to receive requests.

The problem is that the current `InitializerEvent` that I used is triggered before the server is actually listening for requests. This causes time-consuming initialization logic that responds to this event to delay the availability of the server, leading to failed health checks and deployment issues in containerized environments.

Introduce a new `BootstrappedEvent` that is triggered after `app.listen()` when the server is already capable of receiving HTTP requests. This allows:

- (still) Critical initialization logic to run before the `BootstrappedEvent` with the `InitializerEvent` (before the server accepts requests)
- Non-critical, async, time-consuming initialization logic to be deferred to the `BootstrappedEvent` (after the server is already available)
- Kubernetes liveness probes to succeed while background initialization is still running
- Better separation of concerns between "application initialized" (services/config available) and "application ready to serve traffic"

The new event would be published in the `bootstrap()` function right after the server starts listening:

- For k8s I can just increase the timeout durations for the liveness probes...
- I could use timers to load things delayed

kevmtt:
# Description
Fixes: #4203

This PR introduces a new `BootstrappedEvent` that is emitted after the server finishes bootstrapping and starts listening for HTTP requests. This addresses issues with Kubernetes liveness probes timing out when heavy initialization logic needs to be performed during startup.

# Breaking changes

No breaking changes. This is a new feature that adds an event without modifying existing behavior.

# Screenshots

N/A - Backend event system change

# Checklist

📌 Always:

👍 Most of the time:


packages/core/src/event-bus/events/bootstrapped-event.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @description
+ * This event is fired when vendure finished bootstrapping.
+ * For the server and worker process, this happens after the welcome message is logged.
+ *
+ * Use this event to preload data into the cache or perform other startup logic that would otherwise
+ * block the bootstrapping process and slow down server start time.
+ *
+ * @docsCategory events
+ * @docsPage Event Types
+ * @since 3.4.4
```
kevmtt: Maybe 3.5.0?

dlhck: As this goes into master (and I think it is correct to do so) this would be shipped with v3.5.4

dlhck (approved):

dlhck (changes requested):
Version in the doc block needs to be changed but that's it

dlhck (commented):
Also please run docs:build to regenerate the references 😄

kevmtt:
@dlhck The docs build command caused many merge conflicts because of updates formatting. I assume we can just accept or deny them all and generate it again before deployment?

The event is present in the docs though:
image

dlhck:
@Draykee looks like you're working on an older version of `master`. Can you pull the latest upstream and run it again? It should only generate .mdx files, not .md

dlhck:
> @dlhck The docs build command caused many merge conflicts because of updates formatting. I assume we can just accept or deny them all and generate it again before deployment?
> 
> The event is present in the docs though: image

these are the old docs. You need to update your branch to the latest `master` and then run `npm run docs:build`

kevmtt:
@dlhck I'm a bit confused, because still so many files changed. Is it because I ran the command on windows?

Also I believe we should add these to gitignore:
PowerToys_Paste_20260210105023

kevmtt:
Replaced by: https://github.com/vendurehq/vendure/pull/4498

Issue #4162, opened 2026-01-27
Changing quantity to 0 while creating and Offer Draft entirely removes the product.


HouseinIsProgramming:
**Describe the bug**

In the Create draft offer page, while adding products, setting the quantity to 0 or deleting the value removes the product. 

I also expect Enter while changing the quantity to "submit the value", right now it does nothing and the value gets updated anytime it's changed and the user is notified with a toast. 

When setting the value to, for example, 1000, the user would see 3 toasts, "order line updated" toasts, 1 for every key stroke.  

**To Reproduce**

1. Go to the dashboard
2. navigate to Sales > Orders and click on "+ Draft order" 
3. Add a product and try to change the value, then set it to 0.
4. Product is removed.

**Expected behavior**
Product is not removed, and the value can be null, and a proper placeholder is shown, and only when pressing enter while the value is null, would the product be removed.

**Actual behavior**
Product is removed. 

**Screenshots/Videos**

![Image](https://github.com/user-attachments/assets/f02732f0-84b4-40c6-b1ce-486208d36403)

**Minimal reproduction**
Any vendure installation/dev server

latifniz:
Hi, I’d like to work on this issue. I plan to create a PR addressing it. 
Is anyone already working on it?

PR #4208, opened 2026-02-06
fix(admin-dashboard): defer draft order line updates to Enter key, support empty input

latifniz:
# Description

**Related issue:** [[#4162](https://github.com/vendurehq/vendure/issues/4162)](https://github.com/vendurehq/vendure/issues/4162)

While creating a Draft Order, changing the quantity of a product to `0` previously removed the product immediately. Additionally, editing quantities triggered backend updates on every key stroke, causing multiple toasts and an inconsistent UX.

**This PR changes the behavior as follows:**

* Quantity input updates **only on Enter key press**.
* Empty input shows a proper placeholder (`—`).
* Product is removed **only if Enter is pressed while the input is empty**.
* Reduces unnecessary backend calls and toast notifications.
* Fully scoped to `EditOrderTable` component in the admin dashboard.

# Breaking changes

No breaking changes. This change is fully backward compatible.

# Screenshots

https://github.com/user-attachments/assets/3f9593f0-31d7-4f33-aa1f-1d154c118323

# Checklist

📌 Always:


dlhck:
To be honest, this is on purpose. Setting 0 as a quantity means "I want to remove this item" and that's exactly what the system does. What is the benefit of having items with 0 quantity?

latifniz:
Thanks for the feedback .

I implemented this so that the backend is only called when the user presses Enter, rather than on every key press. This prevents unnecessary backend calls and avoids multiple toasts firing for each keystroke, improving both performance and user experience.  

The input still shows a placeholder when empty, and the product is only removed on Enter if the field is empty or 0, following the requested UX in #4162.  

Happy to adjust if the maintainers prefer the strict default behavior where quantity=0 removes the product immediately.

dlhck:
Ok, this makes sense. Pressing "Enter" is a good workflow there from a UX perspective. Please fix the formatting of the code and we can review it afterwards. Thanks.

latifniz:
yes sure.

latifniz:
I have formatted it and just changed it from onChange to onKeyDown, so now it only calls the backend when the user hits enter. and the product is not automatically removed when the quantity is 0 or null, when you press Enter than it's removed

latifniz:

dlhck:
@latifniz your commits are not connected to any GitHub user. Please check the email you have locally configured and the ones in your GitHub profile.

latifniz:
Oh let me check .

latifniz:
recheck

packages/dashboard/src/app/routes/_authenticated/_orders/components/edit-order-table.tsx
```diff
@@ -115,19 +116,22 @@ export function EditOrderTable({
...
                             <Input
                                 type="number"
-                                value={row.original.quantity}
                                 min={0}
-                                onChange={e => {
-                                    const value = Number.isNaN(e.target.valueAsNumber)
-                                        ? 0
-                                        : e.target.valueAsNumber;
+                                defaultValue={row.original.quantity}
+                                placeholder="—"
```
dlhck: Please change the placeholder to `1`

dlhck (changes requested):

packages/dashboard/src/app/routes/_authenticated/_orders/components/edit-order-table.tsx
latifniz: sure

latifniz:
I have changed the placeholder.

packages/dashboard/src/app/routes/_authenticated/_orders/components/edit-order-table.tsx
```diff
@@ -115,19 +116,22 @@ export function EditOrderTable({
...
-                                onChange={e => {
-                                    const value = Number.isNaN(e.target.valueAsNumber)
-                                        ? 0
-                                        : e.target.valueAsNumber;
+                                defaultValue={row.original.quantity}
+                                placeholder="1"
+                                onKeyDown={e => {
+                                    if (e.key !== 'Enter') return;
+                                    const input = e.currentTarget.value.trim();
+                                    const quantity = Number(input);
```
jyling, line 125: It maybe better to also include back the isNaN logic to avoid the quanity to become undefined. if for some reason the `input` is undefined, the quanity would be `undefined`.

Something like this would work
```
  const value = e.currentTarget.value.trim();
  const input = Number(value)
  const quantity = Number.isNaN(input) ? 0 : input;
```


edit: my suggestion have bug lmao

jyling (changes requested):

HouseinIsProgramming:
Thanks for working on this back in February. Closing it now as superseded: the underlying issue (#4162) was fixed by #4345, which merged in July. It implements the same behaviour this PR proposed: the quantity input keeps the typed value in local state, commits on Enter or blur, and pressing Enter on an empty input removes the line.

Since the review feedback here was never picked up and the code has since moved on, there is nothing left for this branch to add. Thanks again for the contribution.

PR #4334, opened 2026-02-10
feat(core): Add ChannelRole entity and RolePermissionResolverStrategy

biggamesmallworld:
# Description

Introduces a `ChannelRole` bridge entity and `RolePermissionResolverStrategy` to decouple role definitions from channel assignments, enabling per-user, per-channel role assignments.

**Key changes:**
- New `ChannelRole` entity that associates a User with a Role on a specific Channel
- `RolePermissionResolverStrategy` interface (InjectableStrategy) with a built-in `ChannelRolePermissionResolverStrategy` implementation
- `createChannelAdministrator` / `updateChannelAdministrator` mutations for managing admins with per-channel role assignments
- `channelRoles` field on the `User` type in the admin API
- Updated session serialization and `RequestContextService` to delegate permission resolution to the configured strategy
- Dashboard UI components for channel-role assignment on the administrator detail page

**Motivation:** In multi-vendor marketplace scenarios, the same Role definition (e.g. "vendor-manager") should be reusable across different users on different channels without duplicating the role per channel. The existing Role → Channel many-to-many means a role's channel scope is global to all users who have it. The new `ChannelRole` entity allows per-user, per-channel role assignments.

Example query: 

```
query AdminChannelRoles($id: ID!) {
    administrator(id: $id) {
      id
      firstName
      lastName
      emailAddress
      user {
        id
        channelRoles {
          id
          channel {
            id
            code
          }
          role {
            id
            code
            description
            permissions
          }
        }
      }
    }
  }
```

Response: 
```
{
  "data": {
    "administrator": {
      "id": "2",
      "firstName": "Will",
      "lastName": "Nahmens",
      "emailAddress": "w.nahmens@vendure.io",
      "user": {
        "id": "12",
        "channelRoles": [
          {
            "id": "5",
            "channel": {
              "id": "2",
              "code": "channel-2"
            },
            "role": {
              "id": "3",
              "code": "administrator",
              "description": "Administrator",
              "permissions": [
                "Authenticated",
                "CreateCatalog",
                "ReadCatalog",
                "UpdateCatalog",
                "DeleteCatalog",
                "CreateSettings",
                "ReadSettings",
                "UpdateSettings",
                "DeleteSettings",
                "CreateCustomer",
                "ReadCustomer",
                "UpdateCustomer",
                "DeleteCustomer",
                "CreateCustomerGroup",
                "ReadCustomerGroup",
                "UpdateCustomerGroup",
                "DeleteCustomerGroup",
                "CreateOrder",
                "ReadOrder",
                "UpdateOrder",
                "DeleteOrder",
                "CreateSystem",
                "ReadSystem",
                "UpdateSystem",
                "DeleteSystem"
              ]
            }
          },
          {
            "id": "6",
            "channel": {
              "id": "1",
              "code": "__default_channel__"
            },
            "role": {
              "id": "4",
              "code": "order-manager",
              "description": "Order manager",
              "permissions": [
                "Authenticated",
                "CreateOrder",
                "ReadOrder",
                "UpdateOrder",
                "DeleteOrder",
                "ReadCustomer",
                "ReadPaymentMethod",
                "ReadShippingMethod",
                "ReadPromotion",
                "ReadCountry",
                "ReadZone"
              ]
            }
          }
        ]
      }
    }
  }
}
```

# Breaking changes

This would be a breaking change when users move to the new strategy. The default `RolePermissionResolverStrategy` preserves existing behavior if that strategy is used. This allows devs who run multi vendor marketplaces to opt in to this permission resolver strategy.

# Screenshots
https://github.com/user-attachments/assets/a4acf4b1-d5d1-4cb6-ba43-059d7a1eec5c

UI for determining the role per channel
https://github.com/user-attachments/assets/f5737ad7-0450-4b2b-b002-7cf4b86e163f

# Checklist

📌 Always:

👍 Most of the time:


grolmus (commented):
## Code Review

Reviewed the full PR — the design direction is solid and the marketplace use case makes sense. However, there are a couple of blocking issues and several things worth addressing before merge.

### Blocking

**1. Privilege escalation — `createChannelAdministrator` / `updateChannelAdministrator` skip role permission checks**

`administrator.service.ts:223-241` — Neither mutation calls `checkActiveUserCanGrantRoles`. The existing `create`/`update` methods both guard against this (line 135, 165), preventing an admin from granting permissions they don't hold. The new mutations skip this entirely. An admin with `CreateAdministrator` permission could assign roles with `SuperAdmin` permissions.

**2. Session cache not invalidated on ChannelRole changes**

`session.service.ts:91-102` — `clearSessionCacheOnDataChange` only clears the session cache when `Channel` or `Role` entities change. `ChannelRole` is not included. After `updateChannelAdministrator` changes assignments, the user's existing session continues using stale permissions until cache TTL expires or they re-authenticate. The e2e test passes because `asUserWithCredentials` forces a fresh login.

### High

**3. Destructive replace-all in `setChannelRolesForUser()`**

`channel-role.service.ts:115-137` — Deletes ALL existing entries then recreates them one by one. Two concerns:
- Race condition under concurrent `updateChannelAdministrator` calls for the same user
- N+1 pattern: each `create()` does 3 `getEntityOrThrow` lookups + 1 save. For 10 assignments = 30 lookups + 10 inserts. Consider a diff-based approach or bulk insert.

**4. Dual code paths for strategy fallback**

`base-auth.resolver.ts`, `session.service.ts`, `role.service.ts`, `request-context.service.ts` — Every call site does:
```ts
const channelPermissions = strategy
    ? await strategy.resolvePermissions(user)
    : getUserChannelsPermissions(user);
```
But `default-config.ts` always sets `DefaultRolePermissionResolverStrategy`, so the fallback never executes. Either make the config property non-optional and trust the strategy, or remove the default — the current approach is dead code that looks important.

**5. `instanceof` check on concrete class**

`administrator.service.ts:285-292` and `global-settings.resolver.ts:70` — `instanceof ChannelRolePermissionResolverStrategy` breaks custom strategy implementations. Consider a marker method on the interface (e.g. `supportsChannelRoles(): boolean`) so third-party strategies can also opt in.

**6. Strategy mutates `user.roles` on the passed-in entity**

`channel-role-permission-resolver-strategy.ts:64` — `user.roles = roles` mutates the entity passed in, which could cause subtle bugs if the same User object is referenced elsewhere in the request lifecycle.

### Medium

**7. `allChannelsUsed` UI logic contradicts data model**

`channel-role-assignment.tsx:69` — Assumes one role per channel, but the unique constraint is on `(user, channel, role)`, allowing multiple roles per channel. The "Add channel-role" button gets disabled prematurely.

**8. `channelRoles` resolver runs unconditionally**

`user-entity.resolver.ts` — The `channelRoles` field queries the DB even when the default strategy is active and no ChannelRole records exist. Consider short-circuiting when channel-role strategy is not active.

**9. Orphaned ChannelRole entries on administrator soft delete**

`onDelete: CASCADE` on the entity fires on DB-level hard delete only. Vendure uses soft delete for administrators, so ChannelRole entries for deleted admins will persist.

### Low / Nice to Have

- Duplicated form fields between Default and ChannelRole admin detail pages — could extract shared component
- Roles query hardcoded `take: 100` without pagination in `channel-role-assignment.tsx`
- `findByUserIdRaw` naming is unclear — consider `findByUserIdWithoutContext` or making ctx optional
- No error wrapping for unique constraint violations in `create()` — raw DB error surfaces to the client

biggamesmallworld:
@michaelbromley if memory serves me correctly, you were working on this functionality with another community member, was this ever implemented? If so I can just close this PR, lmk!

BibiSebi:
Picking this up as part of #3095

Issue #3725, opened 2025-08-02
Product Search Extremely Slow Due to Custom Field Joins


jumarp:
# Product Search Extremely Slow Due to Custom Field Joins

### Describe the bug
Vendure product search is extremely slow when a high number of custom fields are defined on the `Product` entity. Each custom field adds an additional join to the SQL query, causing significant performance issues in the Admin UI when searching or filtering products. This setup is necessary to support the complex requirements of our client (El Gaucho), but it results in unacceptable delays in basic product search operations.

### To Reproduce
1. Go to the Admin UI  
2. Navigate to **Products**  
3. Enter a search term or apply any filter  
4. Observe the long delay before search results appear

### Expected behavior
Search results should load within 1–2 seconds, even when multiple custom fields are present.

### Actual behavior
Search takes 5–10+ seconds or more. The underlying SQL queries generated by Vendure include a large number of `JOIN`s corresponding to each custom field on the `Product` entity, causing the performance degradation.

### Screenshots/Videos
- Video showing the delay: [`VENDURE_SLOWW.mp4`  ](https://drive.google.com/file/d/1WaaI-f55M-66-Uc8SUc_bCy24B0_mnDp/view?usp=drive_link)
- Screenshot of custom fields UI:  
  
<img width="987" height="238" alt="Image" src="https://github.com/user-attachments/assets/0c905b8b-5687-4356-8952-cf881e5a5e7b" />

### Error logs
No application-level error logs are thrown.  
Performance impact is visible in the UI and correlates with the complexity of the generated SQL joins.

```

N/A – no specific error messages, just observable performance degradation.

````

### Environment (please complete the following information):
- **@vendure/core version**: 3.1.3  
- **Node.js version**: v22.17.0  
- **Database**: PostgreSQL 16.9 (via Supabase)  
- **Operating System**: Any (Linux/macOS/Windows)  
- **Browser**: Any  
- **Package manager**: npm 10.9.2  

### Minimal reproduction

1. Set up a Vendure instance with a PostgreSQL backend.
2. Add 20+ custom fields to the `Product` entity.
3. Populate with 1,000+ products.
4. Perform a search or apply a filter in the Admin UI and observe the delay.

### Workaround

No reliable workaround currently. Considering external search plugins (e.g., Elasticsearch) or restructuring the schema.

### Additional context

* Issue started as the number of custom fields grew.
* Happens consistently across all environments.
* Search slowness is directly tied to the SQL join overhead from custom fields.

[...]

PR #4340, opened 2026-02-11
fix(core): Optimize list query builder for many custom fields

RubenDarioGuerreroNeira:
This optimizes SQL generation by using EXISTS subqueries instead of LEFT JOINs for ManyToOne custom field filters, resolving performance issues with large numbers of custom fields.

#3725

# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/benchmark-list-query.e2e-spec.ts
+++ b/packages/core/e2e/benchmark-list-query.e2e-spec.ts
@@ -1,0 +1,112 @@
+/* eslint-disable no-console */
+import { FacetValue, VendureConfig } from '@vendure/core';
+import { createTestEnvironment } from '@vendure/testing';
+import { gql } from 'graphql-tag';
+import path from 'path';
+import { afterAll, beforeAll, describe, it } from 'vitest';
+
+import { initialData } from '../../../e2e-common/e2e-initial-data';
+import { testConfig } from '../../../e2e-common/test-config';
+
+describe('ListQueryBuilder Optimization Benchmark', () => {
+    const capturedQueries: string[] = [];
+    const baseConfig = testConfig();
+
+    const benchmarkConfig: VendureConfig = {
+        ...baseConfig,
+        customFields: {
+            Product: [
+                {
+                    name: 'testRelation',
+                    type: 'relation',
+                    entity: FacetValue,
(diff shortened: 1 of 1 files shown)


RubenDarioGuerreroNeira:
Hi! I've performed some benchmarking to quantify the impact of the changes, specifically focusing on the scenario you mentioned with multiple custom field relations.

## Findings

1. **Functionality Gap:** In the current master branch, filtering by relation IDs on Custom Fields (e.g., testRelationId: { eq: '1' }) is not fully supported or fails to generate the correct filters in the GraphQL schema. This PR enables this functionality.
2. **SQL Optimization:** I've added an E2E benchmark (packages/core/e2e/benchmark-list-query.e2e-spec.ts) that simulates a Product list with multiple Many-to-Many/Many-to-One filters.

## Before (Master/Traditional Join approach)

Normally, TypeORM would attempt to add multiple LEFT JOINs to the same junction table for each filter criteria. With 10-20 custom fields, this leads to a massive Cartesian product, significantly degrading performance or even hitting database join limits.

## After (This PR with EXISTS optimization)

The generated SQL now uses independent EXISTS subqueries for each filter. Here is an example of the generated SQL for a query with multiple filters on the same relation:

```sql
 SELECT DISTINCT "product"."id" FROM "product"
 WHERE ("product"."deletedAt" IS NULL)
 AND (
   EXISTS (
     SELECT 1 FROM "product_facet_values" "lqb_exists_1"
     WHERE "lqb_exists_1"."productId" = "product"."id" AND "lqb_exists_1"."facetValueId" = 1

   AND EXISTS (
     SELECT 1 FROM "product_facet_values" "lqb_exists_2"
     WHERE "lqb_exists_2"."productId" = "product"."id" AND "lqb_exists_2"."facetValueId" = 2


13 LIMIT 1000;

Metrics (SQLite E2E Environment)

Execution Time: ~95ms for a query with multiple _and filters on relations.

Complexity: O(N) subqueries instead of O(N) joins. In PostgreSQL, EXISTS is highly optimized for this "has-a" relationship check, especially when supported by indexes on foreign keys.

This approach ensures that the main query remains "slim" regardless of the number of custom field filters applied, avoiding the performance cliff of multi-join queries.

PR #4343, opened 2026-02-12
fix(core): Persist customFields in updateGlobalSettings mutation

### LINKED ISSUE #4342: updateGlobalSettings does not persist scalar custom fields when a relational custom field exists
# Bug: `updateGlobalSettings` mutation does not persist scalar custom fields when a relational custom field exists

## Describe the bug

The `updateGlobalSettings` mutation does not persist updates made to scalar `customFields` when **at least one custom field on `GlobalSettings` is of type `relation`**.

The mutation executes successfully without errors, but scalar custom field values are not saved in the database if a relational custom field is defined on `GlobalSettings`.

This appears to be caused by the interaction between `customFieldRelationService.updateRelations()` and `settings.customFields` inside `GlobalSettingsService.updateSettings()`.

---

## To Reproduce

1. Add custom fields to `GlobalSettings`:
   - At least one **scalar custom field** (e.g. `string`, `boolean`, etc.)
   - At least one **relational custom field**
2. Call the `updateGlobalSettings` mutation and update the scalar custom field value.
3. The mutation resolves successfully.
4. Query `globalSettings`.
5. Observe that the scalar custom field value is not updated in the database.

---

## Expected behavior

All custom field values (both scalar and relational) provided in the `updateGlobalSettings` mutation input should be correctly saved and persisted.

---

## Actual behavior

When a relational custom field exists on `GlobalSettings`, scalar custom field updates are not persisted.

- The mutation completes successfully.
- No runtime errors are thrown.
- The returned result appears correct.
- However, the database value remains unchanged.

---

Ryrahul:
Closes #4342

### Issue

The `updateGlobalSettings` mutation did not persist updates to `customFields`.

When updating global settings, the custom field values were being overwritten during the update flow, resulting in changes not being saved correctly.

### Root Cause

The update flow caused `customFields` to be reset before the entity was persisted, leading to loss of the updated values.

### Fix

Adjusted the update order in `GlobalSettingsService.updateSettings()` so that custom field updates are preserved and correctly persisted when saving the entity.

### Impact

- No breaking changes
- No schema changes
- Ensures consistent and expected behavior for `updateGlobalSettings`

### DIFF
--- a/docs/docs/reference/dashboard/detail-views/use-generated-form.mdx
+++ b/docs/docs/reference/dashboard/detail-views/use-generated-form.mdx
@@ -2,7 +2,7 @@
 title: "UseGeneratedForm"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx" sourceLine="86" packageName="@vendure/dashboard" since="3.3.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx" sourceLine="91" packageName="@vendure/dashboard" since="3.3.0" />
 
 This hook is used to create a form from a document and an entity.
 It will create a form with the fields defined in the document's input type.
@@ -38,7 +38,7 @@ Parameters
 
 <MemberInfo kind="parameter" type={`<a href='/reference/dashboard/detail-views/use-generated-form#generatedformoptions'>GeneratedFormOptions</a><T, VarName, E>`} />
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx" sourceLine="24" packageName="@vendure/dashboard" since="3.3.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx" sourceLine="29" packageName="@vendure/dashboard" since="3.3.0" />
 
 Options for the useGeneratedForm hook.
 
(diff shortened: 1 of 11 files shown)


Ryrahul:
> Thanks for the PR — good diagnosis of the root cause!
> 
> The fix needs a couple of adjustments to match established patterns:
> 
> ### 1. Use the standard `patchEntity → save → updateRelations` pattern
> Every other service (`zone.service.ts`, `channel.service.ts`, etc.) saves the entity between patching and updating relations. This ensures `updateRelations()` re-fetches the already-persisted scalar values. The method should look like:
> 
> ```ts
> async updateSettings(ctx: RequestContext, input: UpdateGlobalSettingsInput): Promise<GlobalSettings> {
>     const settings = await this.getSettings(ctx);
>     patchEntity(settings, input);
>     await this.eventBus.publish(new GlobalSettingsEvent(ctx, settings, input));
>     await this.connection.getRepository(ctx, GlobalSettings).save(settings);
>     await this.customFieldRelationService.updateRelations(ctx, GlobalSettings, input, settings);
>     return settings;
> }
> ```
> 
> Note: the event should be published **after** `patchEntity` so listeners receive the updated entity, not the pre-change version.
> 
> ### 2. Add a test case
> The existing GlobalSettings test in `custom-field-relations.e2e-spec.ts` only asserts relational custom fields. Please add a case that verifies scalar custom fields persist alongside relational ones.

Sure, completly missed the event publishing part

Ryrahul:
Do we also need to assert the multi relation here? I didn’t add it since similar tests for other entities don’t assert multi in this case, to keep things consistent.

PR #4345, opened 2026-02-12
fix(dashboard): Prevent accidental line removal when editing draft order quantity

### LINKED ISSUE #4162: Changing quantity to 0 while creating and Offer Draft entirely removes the product.

biggamesmallworld:
# Description

The quantity input in the draft order table was firing a GraphQL mutation on every keystroke, causing two problems:

1. **Accidental line removal** — when clearing the field (value passes through 0), the backend auto-removes the order line
2. **Toast spam** — typing a multi-digit number like "1000" would fire 3 separate "order line updated" toasts

Extracts a `QuantityCell` component that uses local state for the input. The value is only committed on **Enter** (removes the line if 0/empty) or **blur** (reverts to previous quantity if 0/empty).

Fixes #4162

# Breaking changes

None.

# Screenshots

https://github.com/user-attachments/assets/8090a794-b969-4f3e-840c-69926cdaac63

# Checklist

📌 Always:

👍 Most of the time:

## Release Notes

* **New Features**
  * Enhanced order line editing with keyboard (Enter) and blur commit, plus a trash button for quick line removal.
  * Improved custom fields integration in order tables.

* **Documentation**
  * Updated form components and page layout documentation with accurate reference information.


grolmus (dismissed):
Tested locally — the core fix works correctly. Clearing the field, typing 0, blur/Enter all behave as expected. Good UX decision on the Enter vs blur asymmetry.

A couple of nice-to-haves for a future pass:

1. **`useMemo` empty deps (line 203)** — the column definitions close over `currencyCode` and `t` but the dependency array is `[]`. Not an issue in practice since currency doesn't change on draft orders, but the correct deps would be `[currencyCode, t]`.

2. **`OrderLineCustomFieldsForm` placement (lines 85-96)** — the custom fields popover is rendered inside `QuantityCell`, which is semantically a quantity concern. Ideally it would be its own column or rendered at the row level. Works fine as-is though.

grolmus (approved):
Re-approving after the master sync — the only changes since my earlier review are `master` merges (no code changes to the fix itself), and CI is green across the board (build, all four dashboard e2e shards, `all-passed`). The core behaviour I tested in March still holds: clearing the field, typing `0`, and blur/Enter all behave correctly, and the Enter-vs-blur asymmetry is a good call.

The two nits from before are still open but remain non-blocking, noting for whenever this area is next touched:
- `columns` `useMemo` still has `[]` deps while closing over `currencyCode` / `t` — harmless here (neither changes on a draft order) but technically the deps should be `[currencyCode, t]`.
- `OrderLineCustomFieldsForm` still lives inside `QuantityCell` rather than its own column/row-level slot.

Neither blocks merge. Approving. Thanks @biggamesmallworld.

PR #4349, opened 2026-02-13
fix(create): Use explicit localhost API host in vite config template

dlhck:
## Summary
- Changed `api: { host: 'auto', port: 'auto' }` to `api: { host: 'http://localhost', port: 3000 }` in the `@vendure/create` vite config template
- The `auto` value doesn't resolve correctly when running the Vite dev server separately, since it runs on its own port and not on `localhost:3000`

## Test plan
- [ ] Run `npx @vendure/create` and verify the generated `vite.config.ts` contains the explicit host/port
- [ ] Confirm the dashboard dev server correctly connects to the Vendure backend at `localhost:3000`


michaelbromley:
Note: this reverts the change introduced in 

- https://github.com/vendurehq/vendure/pull/4102

PR #4352, opened 2026-02-13
fix(core): Emit post-update entity in VendureEntityEvent for updated events

### LINKED ISSUE #4351: fix(core): VendureEntityEvent emits pre-update entity state for several services
## Description

Several services emit `VendureEntityEvent` (or subclasses) with `type: 'updated'` where `event.entity` contains the entity state **before** the update mutation is applied, rather than the **after** state. This causes downstream consumers (like audit trail plugins) to record stale data.

## Evidence

When updating a Promotion's `enabled` field from `true` to `false`:

```
input.enabled: false    ← what the user submitted (correct)
state.enabled: true     ← entity BEFORE the update was applied (wrong — should be false)
```

The `event.entity` and `event.input` are inverted — `input` has the new values, `entity` has the old values.

### Wrong variable passed to event (3 services)

| Service | Entity | Line | Issue |
|---|---|---|---|
| `promotion.service.ts` | Promotion | L191 | Passes `promotion` (pre-fetch) instead of `updatedPromotion` (translatableSaver return) |
| `shipping-method.service.ts` | ShippingMethod | L182 | Passes `shippingMethod` (pre-fetch) instead of `updatedShippingMethod` (translatableSaver return) |
| `administrator.service.ts` | Administrator | L207 | Passes `administrator` (pre-fetch) instead of `updatedAdministrator` (reassigned after `assignRole()`) |

### Event fires before final save (1 service)

| Service | Entity | Lines | Issue |
|---|---|---|---|
| `payment-method.service.ts` | PaymentMethod | L145-146 | Event published on L145, but final `.save()` on L146 happens after |

## Root Cause

The bug pattern occurs in services using `translatableSaver.update()`, which returns a **new entity object** from `repo.save()`. The surrounding code keeps a reference to the original entity fetched at the top of the method and passes that stale reference to the event instead of the updated one.

## Fix

- For wrong-variable bugs: pass the post-update variable to the event constructor
- For ordering bug: swap the event publish and save lines

michaelbromley:
## Summary

Several services emitted `VendureEntityEvent` (or subclasses) with `type: 'updated'` where `event.entity` contained the entity state **before** the update was applied. This caused downstream consumers (audit trail plugins, webhooks, etc.) to receive stale data.

### Fixed services

| Service | Entity | Bug | Fix |
|---|---|---|---|
| `promotion.service.ts` | Promotion | Passed pre-fetch `promotion` to event | Use `updatedPromotion` (return of `translatableSaver.update()`) |
| `shipping-method.service.ts` | ShippingMethod | Passed pre-fetch `shippingMethod` to event | Use `updatedShippingMethod` (return of `translatableSaver.update()`) |
| `administrator.service.ts` | Administrator | Passed `administrator` (stale after `assignRole()` reassigns `updatedAdministrator`) | Use `updatedAdministrator` |
| `payment-method.service.ts` | PaymentMethod | Event published before final `.save()` | Swap event publish and save lines |

### Root cause

Services using `translatableSaver.update()` receive a **new entity object** from `repo.save()`, but the event was passed the original entity fetched at the top of the method. For the administrator service, the `updatedAdministrator` variable gets reassigned to a new object when roles are updated via `assignRole()`, making `administrator` definitively stale.

## Test plan

- [x] Added `entity-event-update-state.e2e-spec.ts` with 11 tests covering all 4 services
- [x] Tests subscribe to EventBus events and verify `event.entity` reflects post-update state
- [x] Tests confirmed to **fail** on the old code and **pass** after the fix

Fixes #4351

### DIFF
--- a/packages/core/e2e/entity-event-update-state.e2e-spec.ts
+++ b/packages/core/e2e/entity-event-update-state.e2e-spec.ts
@@ -1,4 +1,3 @@
-import { Type } from '@vendure/common/lib/shared-types';
 import {
     AdministratorEvent,
     defaultShippingCalculator,
@@ -11,12 +10,11 @@ import {
     PromotionEvent,
     PromotionOrderAction,
     ShippingMethodEvent,
-    VendureEvent,
 } from '@vendure/core';
 import { createTestEnvironment } from '@vendure/testing';
 import gql from 'graphql-tag';
 import path from 'path';
-import { Subscription } from 'rxjs';
+import { firstValueFrom } from 'rxjs';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import { initialData } from '../../../e2e-common/e2e-initial-data';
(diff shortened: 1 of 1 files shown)


packages/core/e2e/entity-event-update-state.e2e-spec.ts
```diff
@@ -0,0 +1,480 @@
...
+    eventType: new (...args: any[]) => T,
+): {
+    eventReceived: Promise<T>;
+    subscription: Subscription;
+} {
+    let resolveFn: (value: T) => void;
+    const eventReceived = new Promise<T>(resolve => {
+        resolveFn = resolve;
+    });
+    const subscription = eventBus.ofType(eventType).subscribe(event => {
```
michaelbromley: can we not use rxjs firstValueFrom here wrapped around the observable instead of this whole function?

packages/core/e2e/entity-event-update-state.e2e-spec.ts
```diff
@@ -0,0 +1,480 @@
...
+const testPromoAction = new PromotionOrderAction({
+    code: 'test_event_action',
+    description: [{ languageCode: LanguageCode.en, value: 'Test event action' }],
+    args: {},
+    execute: (_order, _args) => {
+        return 0;
+    },
+});
+
+const PAYMENT_METHOD_FRAGMENT = gql`
```
michaelbromley: in other e2e tests we tend to define the the gql docs at the end of the file, we should do the same here

packages/core/e2e/entity-event-update-state.e2e-spec.ts
```diff
@@ -0,0 +1,480 @@
...
+    const eventReceived = new Promise<T>(resolve => {
+        resolveFn = resolve;
+    });
+    const subscription = eventBus.ofType(eventType).subscribe(event => {
+        resolveFn(event);
+    });
+    return { eventReceived, subscription };
+}
+
+describe('Entity event updated state', () => {
```
michaelbromley, line 58: add a comment referencing the issue that this suite is testing, as we do in other e2e test suites

PR #4361, opened 2026-02-17
feat(dashboard): Add bulk cancel action and human-readable duration t…

### LINKED ISSUE #4331: Bulk actions not available for selected jobs in Job Queue
## Description

When selecting multiple jobs in the Job Queue admin UI, the bulk action dropdown shows "No actions available" even when actions should be applicable to all selected jobs. Also the time should be formatted in days, hours, seconds etc. not only ms 

## Steps to Reproduce

1. Navigate to **Job Queue** in the Admin UI
2. Filter by a specific queue (e.g., `generate-docs`) and state (e.g., `Running`)
3. Select multiple running jobs using the checkboxes
4. Click the "With selected..." dropdown

## Expected Behavior

The dropdown should show available bulk actions that apply to all selected jobs. For running jobs, I would expect at minimum a **Cancel** action to be available.

## Actual Behavior

The dropdown displays "No actions available" despite having 11 running jobs selected.

## Screenshot

Image

## Additional Context

The jobs shown have been running for extended periods (ranging from ~8.5 days to ~16.5 days based on the duration values), which suggests they may be stuck. Being able to bulk cancel these jobs would be essential for queue management.

Ryrahul:
## Title
 Add bulk cancel action and human-readable duration to job queue

---

## Summary

- Add a bulk cancel action to the Job Queue list page, allowing users to cancel multiple running/pending jobs at once via the **"With selected..."** dropdown.
- Format the duration column in human-readable units (e.g. `2m 5s`, `16d 12h`) instead of raw milliseconds (e.g. `125230ms`).

---

## Context

Currently, cancelling stuck or long-running jobs requires cancelling them one by one through the per-row dropdown. This PR adds a bulk action so multiple jobs can be selected and cancelled together  essential for queue management when many jobs are stuck.

The duration column previously displayed raw milliseconds, which becomes unreadable for long-running jobs (e.g. `1425600000ms`). It now displays durations in days, hours, minutes, and seconds.

---

Fixes #4331


Ryrahul:
The formatDuration() helper is currently defined in the same job-queue.tsx file since it's only used there. Happy to move it to a shared utility file (e.g. src/lib/utils/format-duration.ts) if you'd prefer  let me know!

grolmus (dismissed):
## Review Summary

The duration formatting change is clean and correct — nice improvement. The bulk cancel action is a useful feature but needs a few changes before merge.

### Must Fix

**Missing `requiresPermission` on bulk action**

The `cancelJob` mutation requires `Permission.DeleteSettings` or `Permission.DeleteSystem` (see `job.resolver.ts:77`). The bulk action doesn't pass `requiresPermission` to `DataTableBulkActionItem`, so any logged-in admin sees the cancel option regardless of their permissions. Every other bulk action in the codebase includes this. Fix:

```tsx
<DataTableBulkActionItem
    requiresPermission={['DeleteSettings', 'DeleteSystem']}
    onClick={cancelSelectedJobs}
    ...
```

### Should Fix

**1. Not using `useMutation` — no double-click protection**

The bulk action uses raw `async/await` with `api.mutate()` instead of `useMutation` from `@tanstack/react-query`. This means there's no loading state — the button stays clickable during execution, so a double-click fires duplicate cancellations. The canonical pattern is in `delete-bulk-action.tsx`. The `DataTableBulkActionItem` component already supports a `disabled` prop that can be wired to `isPending`.

**2. Duplicate cancellable jobs filter**

The filter `selection.filter(job => job.state === 'RUNNING' || job.state === 'PENDING')` is computed twice with identical logic (lines 14 and 33). Compute it once and share the result:

```tsx
const cancellableJobs = selection.filter(job => job.state === 'RUNNING' || job.state === 'PENDING');
const cancellableCount = cancellableJobs.length;
```

**3. `job(s)` pattern doesn't work for i18n**

```tsx
toast.success(t`Successfully cancelled ${fulfilled} job(s)`);
```

The `(s)` pattern breaks for languages with complex pluralization rules. Lingui has `Plural` component and `plural` macro for this. At minimum, follow the existing `${count} jobs` pattern from `delete-bulk-action.tsx`.

### Nitpicks (non-blocking)

- Extra blank line after `CancelJobsBulkAction` import in `job-queue.tsx`
- `formatDuration` could live in a shared utility file, but since it's only used here that's fine for now

biggamesmallworld:
## Merge Conflict Resolution

There's a conflict in `packages/dashboard/src/app/routes/_authenticated/_system/job-queue.tsx`. Here's what happened:

**Master** added a `getJobStateBadgeVariant()` function at the same location where **this branch** added a `formatDuration()` function. Git sees them as conflicting because they were both inserted at the same spot in the file.

### How to resolve

You need **both** functions — they do different things. After rebasing/merging, the file should have both:

```ts
import { CancelJobsBulkAction } from './components/cancel-jobs-bulk-action.js';
import { PayloadDialog } from './components/payload-dialog.js';
import { cancelJobDocument, jobListDocument, jobQueueListDocument } from './job-queue.graphql.js';

// Keep this from master
function getJobStateBadgeVariant(state: string) {
    switch (state) {
        case 'PENDING':
        case 'RETRYING':
            return 'warning';
        case 'COMPLETED':
            return 'success';
        case 'FAILED':
        case 'CANCELLED':
            return 'destructive';
        default:
            return 'secondary';
    }
}

// Keep this from the PR branch
function formatDuration(ms: number): string {
    if (ms < 1000) {
        return `${ms}ms`;
    }

    const seconds = Math.floor(ms / 1000);
    const minutes = Math.floor(seconds / 60);
    const hours = Math.floor(minutes / 60);
    const days = Math.floor(hours / 24);

    const parts: string[] = [];

    if (days > 0) {
        parts.push(`${days}d`);
    }
    if (hours % 24 > 0) {
        parts.push(`${hours % 24}h`);
    }
    if (minutes % 60 > 0) {
        parts.push(`${minutes % 60}m`);
    }
    if (seconds % 60 > 0) {
        parts.push(`${seconds % 60}s`);
    }

    return parts.join(' ');
}
```

In short: accept both sides, keep both functions. The rest of the changes (import of `CancelJobsBulkAction`, the `bulkActions` prop, and the `formatDuration` usage in the duration cell) are additive and should merge cleanly once the conflict marker is resolved.

biggamesmallworld:
@Ryrahul this is approved, please resolve the merge conflicts and we will merge as soon as that is finished

Ryrahul:
> @Ryrahul this is approved, please resolve the merge conflicts and we will merge as soon as that is finished

Addressed

biggamesmallworld:
@Ryrahul one last thing, you'll need to run `npm run i18n:extract ` due to this file introducing some translated strings, can you please run the extract and commit? That should re-trigger the jobs and allow us to merge

Ryrahul:
> @Ryrahul one last thing, you'll need to run `npm run i18n:extract ` due to this file introducing some translated strings, can you please run the extract and commit? That should re-trigger the jobs and allow us to merge

Next steps:
1. Copy the content of missing-translations.txt to Claude or another LLM
2. Save the translated output to a file (e.g., translations.txt)
3. Run: node 118n-tool.js apply translations-file>

Is this also needed after extraction ? @biggamesmallworld

biggamesmallworld:
No you just need to run the extract, all good

PR #4369, opened 2026-02-17
fix(dashboard): add entity to DetailPage

LucidityDesign:
# Description

When adding a page block via `defineDashboardExtension` the `context.entity` is received from the page component. Without this change `context.entity` is undefined.

# Breaking changes

None

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


LucidityDesign:
Sidenote: I have a special usecase where the create mutation returns more data than the get query. So it would be nice to have a way to skip the "get", because the create mutation already returns all the data that I need.

E.g.
```
mutation createApiKey {
   id
   ...
   unhashedKey
}

query apiKey {
   id
   ...
   # unhashedKey is only available during create
}
```

LucidityDesign:

biggamesmallworld (approved):
Reviewed the change — this is correct and straightforward.

The `Page` component already accepts `entity` and passes it to `PageContext.Provider`, but `DetailPage` wasn't forwarding it. Every hand-written detail page in `src/app/routes/` already passes `entity={entity}` to `<Page>` — this was simply an oversight in the generic `DetailPage` abstraction.

No risk of regression: `entity` is optional on `PageProps`, and when creating a new entity it will be `undefined`, which `PageContext` consumers already handle gracefully (e.g. `EntityInfoDropdown` guards with `if (!entity || !entity.id) return null`).

LGTM — thanks for the contribution!

biggamesmallworld:
**Testing performed:**
- [x] Nigel deep code review (correctness, architecture, edge cases, performance, security)
- [ ] Local build & type check
- [ ] Unit / e2e tests run locally
- [ ] Manual testing against running dev server

One-line prop forwarding fix with no regression risk — code review sufficient for this change.

LucidityDesign:
I tried to run `npm run docs:build` but get this error message:
```
> @vendure/docs@0.0.0-202601281114 generate-dates
> tsx scripts/generate-dates.ts


node:internal/modules/run_main:107
    triggerUncaughtException(
    ^
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@vendure-io/docs-provider' imported from /path/to/vendure/docs/scripts/generate-dates.ts
```

PR #4374, opened 2026-02-18
fix: Discounts visibility in order details table

### LINKED ISSUE #4326: Dashboard: incorrect display of discounts on the detailed order page in the OrderLine
Image

Ryrahul:
---

## Description

Fix `[object Object]` being displayed in the "Discounts" column of the order detail table in the Dashboard.

The order line table in `order-table.tsx` had custom cell renderers for most columns (featuredAsset, productVariant, unitPrice, quantity, etc.) but was missing one for `discounts`. This caused the auto-generated column renderer in `DefaultDisplayComponent` to call `.join(', ')` on an array of Discount objects, which produced `[object Object]` via `Object.prototype.toString()`.

Added a custom `discounts` column renderer that displays each discount's description along with the formatted amount in a stacked layout, consistent with how other price columns (like Unit Price and Total) display gross/net values.

**File changed:**  
`packages/dashboard/src/app/routes/_authenticated/_orders/components/order-table.tsx`

---

## Breaking changes

None. This is a display-only fix scoped to the order detail table in the Dashboard.

---

## Screenshots

image

---

## Checklist

---
Fixes #4326

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_orders/components/order-table.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_orders/components/order-table.tsx
@@ -124,6 +124,25 @@ function createCustomizeColumns(currencyCode: string) {
                 <Money currencyCode={currencyCode} value={row.original.discountedLinePriceWithTax} />
             ),
         },
+        discounts: {
+            header: () => <Trans>Discounts</Trans>,
+            cell: ({ row }: { row: any }) => {
+                const discounts = row.original.discounts;
+                if (!discounts?.length) return null;
+                return (
+                    <div className="flex flex-col gap-1">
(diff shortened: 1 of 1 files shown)


gabriellbui:
Hi @Ryrahul! Nice work on this! Two small things I've noticed

1. The other price columns use `MoneyGrossNet` to show both gross and net amounts. To keep it consistent I would use that here as well instead of `Money`
2. I think the root problem is in `DefaultDisplayComponent` as you have also mentioned. However, it's still gonna call `.join(',  ')` on similar object arrays without a custom renderer resulting in the same `[object Object]`. I think we should also consider that.

I created a [PR](https://github.com/vendurehq/vendure/pull/4378) that would address both of these

PR #4376, opened 2026-02-19
feat(core): Add configurable OrderTaxSummaryCalculationStrategy

colinpieper:
# Description

Introduces a strategy pattern for how order-level tax totals and tax summaries are calculated, allowing different rounding approaches.

The default strategy (DefaultOrderTaxSummaryCalculationStrategy) rounds tax per line then sums, preserving existing Vendure behavior.

The new OrderLevelTaxSummaryCalculationStrategy groups net subtotals by tax rate and rounds once per group, eliminating per-line rounding accumulation. This is required by certain jurisdictions and ERP systems that expect tax calculated on the subtotal per rate.

Configurable via taxOptions.orderTaxSummaryCalculationStrategy.

Relates to #4375

# Breaking changes

No breaking changes. Defaults to the existing behavior.

# Checklist

📌 Always:

👍 Most of the time:

--

This is my first feature contribution to Vendure, so please let me know if I missed anything. Feedback is welcome.

### DIFF
--- a/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
+++ b/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
@@ -233,6 +233,7 @@ describe('OrderCalculator', () => {
             mockConfigService.taxOptions = {
                 taxZoneStrategy: new DefaultTaxZoneStrategy(),
                 taxLineCalculationStrategy: new DefaultTaxLineCalculationStrategy(),
+                orderTaxCalculationStrategy: new DefaultOrderTaxCalculationStrategy(),
             };
 
             const ctx = createRequestContext({ pricesIncludeTax: false });


colinpieper:
@coderabbitai review

colinpieper:
@coderabbitai review

packages/core/src/config/tax/order-tax-summary-calculation-strategy.ts
```diff
@@ -0,0 +1,67 @@
...
+ * your VendureConfig.
+ *
+ * :::
+ *
+ * @docsCategory tax
+ * @docsPage OrderTaxSummaryCalculationStrategy
+ * @docsWeight 0
+ * @since 3.6.0
+ */
+export interface OrderTaxSummaryCalculationStrategy extends InjectableStrategy {
```
michaelbromley: I think the name could be made a bit more descriptive of what it actually does - since it handles more than just the tax summary.

I would suggest either `OrderTaxCalculationStrategy` _or_ we could consider splitting this into 2 distinct strategies - `OrderTaxSummaryCalculationStrategy` which contains only the `calculateTaxSummary` method, and then a `OrderTotalsCalculationStratety` which contains `calculateOrderTotals`.

Do you have any input on that?

colinpieper: Makes sense. I'd opt for renaming (OrderTaxCalculationStrategy) over splitting as the two methods are cohesive as they must use identical grouping logic. The splitting of the methods is a hot-path optimization.

I'll go ahead and rename it.

PR #4383, opened 2026-02-20
feat(core): Provide app instance before app starts listening

LucidityDesign:
# Description

I'm trying to set up swagger/openAPI.

Bootstrapping swagger requires the instance to app before `app.listen` has been called (see https://docs.nestjs.com/openapi/introduction#bootstrap). So currently I can't find an easy way to set up swagger for my REST endpoints in Vendure.

@michaelbromley What do you think about this solution? If you like it I can add documentation and finish this PR.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


packages/core/src/bootstrap.ts
```diff
@@ -70,6 +70,27 @@ export interface BootstrapOptions {
      * @since 3.1.0
      */
     ignoreCompatibilityErrorsForPlugins?: Array<DynamicModule | Type<any>>;
+
+    /**
+     * @description
+     * A function which is called before the app starts listening. This can be used to perform any final configuration of the Nest application.
+     * E.g., to set global filters, interceptors or pipes.
```
michaelbromley: I wouldn't use the example of global middleware, because we already have an established pattern for that documented here: https://docs.vendure.io/current/core/developer-guide/the-api-layer#global-nestjs-middleware

Your original use case of supporting Swagger is a better example, because there's literally no other way to cleanly do that, than use this API.

Could you change the example? Also then run `npm run docs:build` to have it generate the mdx docs from your changes.

PR #4384, opened 2026-02-20
fix(dashboard): Add missing dashboard dependency

### LINKED ISSUE #4382: Missing direct dependency in dashboard
**Describe the bug**
The dashboard [imports `@radix-ui/react-direction`](https://github.com/vendurehq/vendure/blob/12d1ef84b93a8a81023469128daa825e9efa6cd6/packages/dashboard/src/app/main.tsx#L17), but it's not listed in its dependencies which leads to problems when used in monorepos that have apps with different React versions.

**To Reproduce**
No reproduction required as [the correct practice](https://pnpm.io/typescript#:~:text=This%20is%20actually%20an%20issue%20on%20antd's%20end) is to list all direct dependencies in the package's `package.json` to avoid this exact problem.

**Expected behavior**
The vendure dashboard should correctly behave in a PNPM workspace with packages that use multiple React versions by rendering without errors in the browser console.

**Actual behavior**
The dashboard doesn't render any screen, and the browser console contains the following error:

```
Error: A React Element from an older version of React was rendered. This is not supported. It can happen if:
- Multiple copies of the "react" package is used.
- A library pre-bundled an old copy of "react" or "react/jsx-runtime".
- A compiler tries to "inline" JSX instead of using the runtime.
```

- @vendure/core version: master
- Nodejs version: N/A
- Database (mysql/postgres etc): N/A
- Operating System (Windows/macOS/Linux): N/A

**Workaround**

Add this to your `pnpm-workspace.yaml`:

```yaml
packageExtensions:
  '@vendure/dashboard':
    'dependencies':
      '@radix-ui/react-direction': '*'
```

Then delete `node_modules` (not sure whether you need to delete all of them across all packages, but that's what I did) and `pnpm-lock.yaml`, run `pnpm install` and run the app.

**Additional context**
I proposed the PR #4384 to fix it.

fcFn:
# Description

Fixes #4382

# Breaking changes

No

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/license/signatures/version1/cla.json
+++ b/license/signatures/version1/cla.json
@@ -1031,6 +1031,14 @@
       "created_at": "2026-02-17T13:15:11Z",
       "repoId": 136938012,
       "pullRequestNo": 4369
+    },
+    {
+      "name": "fcFn",
+      "id": 8431256,
+      "comment_id": 3934161767,
+      "created_at": "2026-02-20T12:47:26Z",
+      "repoId": 136938012,
+      "pullRequestNo": 4384
     }
   ]
 }


fcFn:

michaelbromley:
Thanks!

PR #4402, opened 2026-02-24
feat(core): add previous to collection update-events

flobacher:
# Description

In addition to the updated CollectionEntity, there are scenarios where it comes in handy to have the state of the collection before the event happened.

# Breaking changes

Nope. Only an additional optional property on the CollectionEvent

# Checklist

📌 Always:

### DIFF
--- a/packages/core/src/event-bus/events/collection-event.ts
+++ b/packages/core/src/event-bus/events/collection-event.ts
@@ -21,6 +21,7 @@ export class CollectionEvent extends VendureEntityEvent<Collection, CollectionIn
         entity: Collection,
         type: 'created' | 'updated' | 'deleted',
         input?: CollectionInputTypes,
+        public readonly previousEntity?: Collection,
     ) {
         super(entity, type, ctx, input);
     }
--- a/packages/core/src/service/services/collection.service.ts
+++ b/packages/core/src/service/services/collection.service.ts
@@ -560,7 +560,16 @@ export class CollectionService implements OnModuleInit {
 
     async update(ctx: RequestContext, input: UpdateCollectionInput): Promise<Translated<Collection>> {
         await this.slugValidator.validateSlugs(ctx, input, CollectionTranslation);
-        const collection = await this.translatableSaver.update({
+        const existingCollection = await this.connection.getEntityOrThrow<Collection>(
+            ctx,
+            Collection,
+            input.id,
+            {
+                channelId: ctx.channelId,
+            },
+        );
+
+        const collection = await this.translatableSaver.update<Collection>({
             ctx,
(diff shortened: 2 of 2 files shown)


grolmus (commented):
Thanks for the PR! The idea of exposing previous state on update events is useful. A couple of concerns before this can be merged:

## Must address

### 1. `existingCollection` is loaded without relations

The `getEntityOrThrow` call fetches the entity with no `relations` option, so the resulting object has no translations, filters, assets, or parent/children loaded. Any event handler trying to diff "what changed" will get an incomplete/empty previous entity — that's a footgun for consumers.

At minimum, the fetch should load the same relations that make the entity useful (translations, filters, assets). Consider reusing the same approach as `findOne()` or explicitly specifying the needed relations.

### 2. Ad-hoc pattern with no precedent in the codebase

No other entity event (`ProductEvent`, `OrderEvent`, `CustomerEvent`, etc.) carries a `previousEntity`. This introduces a one-off pattern on a single event subclass.

If this is a broadly useful feature (and I think it could be), it would be better suited as an optional property on the `VendureEntityEvent` base class, so any entity event can carry previous state consistently. Otherwise we'll end up with each event subclass adding its own ad-hoc version over time.

This is ultimately a design decision for the core maintainers — but worth raising before it goes in.

## Minor

- Two optional positional constructor params in a row (`input?`, `previousEntity?`) can be error-prone. A named options bag or moving this to the base class would help.
- The extra DB query runs on every collection update but returns a mostly empty entity — not a great cost/benefit ratio if the relations aren't loaded.

Happy to re-review once these are addressed!

BibiSebi:
Hey @flobacher I know exactly where this request is coming from :D I think it would be best to just open an issue and describe it first so Michael or someone from the team can have a think about the solution as a whole. If this is something that needs to be added to the other events as well or not.

Ryrahul:
https://github.com/vendurehq/vendure/discussions/4247

Its related, you can upvote here ig

flobacher:
> #4247
> 
> Its related, you can upvote here ig

Just upvoted!

grolmus:
Hi @flobacher — thank you for this contribution, and apologies for the long delay on our side. To get it across the line for the 3.8 release we've taken it forward in #5067, which builds directly on your commit (preserved, so your authorship is credited).

On top of your work it: loads the previous collection **with its relations** so `previousEntity` is actually usable for diffing (the original bare load left `translations` and the other relations empty), consolidates the duplicate active-channel guard load, documents the new field with `@since 3.8.0`, and adds e2e coverage.

Closing this in favour of #5067. Thanks again for kicking it off! 🙏

PR #4404, opened 2026-02-24
[FIX] Save and serve files with correct Content-Type

### LINKED ISSUE #4403: Content-Type not set correctly for .pdf assets (and assets uploaded to S3)
**Describe the bug**
With the current setup of the AssetServer plugin, if there are PDF's linked to a product, e.g. for a download or product sheet, when fetched from the server, it always returns them as `application/octet-stream`. In firefox-based browsers this works fine and PDF's are just opened in a new tab, but on Chromium-based browsers the file instantly downloads instead of opening in a new tab.

Also when using S3, the content type was never properly set when uploading the files.

**To Reproduce**

1. Upload a file with a pdf extension
2. Link to a product (/-variant)
3. Query the product (/-variant) with the PDF asset
4. Query the PDF asset from the server via a Chromium-based browser

**Expected behavior**
Open PDF assets in a new tab so the user has to decide whether to download the file or to just open the file in a new tab for temporarily viewing it.

**Actual behavior**
PDF assets directly download from the server.

**Screenshots/Videos**

https://github.com/user-attachments/assets/1f5bf5e0-84ba-4f4c-b38d-e28e37862045

**Error logs**
- N/A

- Nodejs version: 24.11.1 (LTS)
- Browser (if applicable): Anything Chromium-based

**Configuration**
- N/A

**Minimal reproduction**
A fresh Vendure instance with the AssetServer plugin instance and a local storage strategy will do.

**Workaround**
- N/A

**Additional context**
Happens consistently, ~~I am going to submit a PR to address this~~ PR has been submitted #4404.

kwerie:
# Description
PDFs were served with application/octet-stream, causing Chromium-based browsers to download them instead of opening them in a new tab. Add application/pdf to the MIME type mapping so PDFs are served with the correct Content-Type header.

Also when saving files to an S3 bucket, the content type wasn't being set.

# Breaking changes

No

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

Fixes #4403


biggamesmallworld (approved):
Great fix! The S3 ContentType addition is a valuable improvement beyond just the serving-side fix. Thanks for the contribution.

kwerie:
@biggamesmallworld Hey! Any update on this? :)

packages/asset-server-plugin/package.json
```diff
@@ -40,6 +40,7 @@
     "dependencies": {
         "file-type": "^19.0.0",
         "fs-extra": "^11.2.0",
+        "mime-types": "^3.0.2",
```
biggamesmallworld, line 42: `mime-types@3` has no bundled types — this only compiles because core’s `@types/mime-types` hoists to root. Declare it directly in `devDependencies`:

```jsonc
"@types/mime-types": "^3.0.1",
```

packages/asset-server-plugin/src/asset-server.ts
```diff
@@ -297,6 +297,8 @@ export class AssetServer {
             case '.jpg':
             case '.jpeg':
                 return 'image/jpeg';
+            case '.pdf':
```
biggamesmallworld: Reuse `mime-types` instead of growing the switch (same as core’s `asset.service.ts:526`). `mime.lookup` returns `false` for unknown names, so `|| undefined` keeps the existing `file-type` sniff fallback.

Import:
```ts
import mime from 'mime-types';
```
Replace `getMimeType`:
```ts
private getMimeType(fileName: string): string | undefined {
    return mime.lookup(fileName) || undefined;
}
```

biggamesmallworld (commented):
S3 `ContentType` fix is the right root cause. Happy to approve after, but would ask that you implement the following before merge:

- **Reuse `mime-types` in `asset-server.ts`** instead of extending the hardcoded switch — core already does this (`asset.service.ts:526`). Details inline.
- **Add `@types/mime-types`** (`mime-types@3` has no bundled types; only compiles via core hoisting). Inline.
- **Test the S3 path** — the new e2e only covers local storage, which likely already returned `application/pdf` via `file-type` sniffing. The actual fix (S3) is untested.
- **Security:** `mime.lookup` is extension-driven, so `.svg`/`.html` in a directly-served bucket now go inline. Safe under default upload validation — worth a one-line comment at the S3 `ContentType`.
- **Drop** the `fulfillment`/`payment`/`stock-movement`/`order-process` changes — unrelated `eslint --fix` churn in state-transition code; separate PR.

biggamesmallworld (approved):

kwerie:
@biggamesmallworld thanks for the review — all five points are in now:

- `getMimeType` in `asset-server.ts` now uses `mime.lookup()` instead of the hardcoded switch.
- Added `@types/mime-types` to `devDependencies`.
- Added a unit test for the S3 path (`s3-asset-storage-strategy.spec.ts`) asserting `ContentType` is set from the extension (`application/pdf`, falling back to `application/octet-stream`). It mocks the SDK `Upload` rather than a real bucket, since the repo has no S3/MinIO test setup today.
- Added the security note at the S3 `ContentType`.
- Reverted the unrelated `fulfillment`/`payment`/`stock-movement`/`order-process` formatting churn.

PR #4405, opened 2026-02-24
fix(core): Enforce usage limits for auto-applied promotions

### LINKED ISSUE #4353: Promotion usageLimit and perCustomerUsageLimit ignored for auto-applied promotions without couponCode
**Describe the bug**  
`usageLimit` and `perCustomerUsageLimit` are ignored for promotions that do not have a `couponCode`.

Auto-applied promotions (without a coupon code) can be used an unlimited number of times, even when usage limits are configured in the Admin UI. Usage limits are only enforced when a coupon code is explicitly applied via `applyCouponCode()`.

This results in inconsistent behavior between coupon-based promotions and auto-applied promotions.

---

**To Reproduce**  

1. Go to **Admin UI → Promotions**
2. Create a new promotion **without a coupon code**
3. Set:
   - `usageLimit = 1`
   - `perCustomerUsageLimit = 1`
4. Add a condition so the promotion auto-applies (e.g., minimum order value)
5. Place multiple eligible orders
6. Observe that the promotion continues to apply beyond the configured limits

---

**Expected behavior**  
Promotions should respect `usageLimit` and `perCustomerUsageLimit` regardless of whether a `couponCode` is configured.

Alternatively, if usage limits are intentionally meant to apply only to coupon-based promotions, this should either:

- Be clearly documented, or  
- Be prevented at the Admin UI level for non-coupon promotions.

---

**Actual behavior**  
Usage limits are only enforced inside `PromotionService.validateCouponCode()`, which is called exclusively when applying a coupon via `OrderService.applyCouponCode()`.

Auto-applied promotions follow this flow:

`OrderService.applyPriceAdjustments()`  
→ `OrderCalculator.applyPromotions()`  
→ `Promotion.test()`  
→ `Promotion.apply()`

`Promotion.test()` checks:
- start/end dates  
- coupon presence (if defined)  
- promotion conditions  

However, it does **not** check `usageLimit` or `perCustomerUsageLimit`.

As a result, auto-applied promotions never enforce usage limits.

---

---

- Database: PostgreSQL  
- Operating System: macOS  
- Package manager: yarn  

---

HouseinIsProgramming:
# Description

Fixes #4353

`usageLimit` and `perCustomerUsageLimit` were ignored for auto-applied promotions (those without a coupon code). Usage limits were only enforced in `PromotionService.validateCouponCode()`, which is never called for the auto-apply code path (`OrderCalculator.applyPromotions()` → `Promotion.test()` → `Promotion.apply()`).

This PR adds a new `getExhaustedPromotionIds()` method to `PromotionService` that uses batch queries to identify promotions exceeding their limits, and filters them out in `OrderService.applyPriceAdjustments()` before passing to `OrderCalculator`.

**Performance:**
- Only queries the DB when promotions with usage limits exist (zero overhead in the common case)
- Uses batch `GROUP BY` queries (1-2 max) instead of N individual COUNT queries
- Cached per-request via `RequestContextCacheService`

# Breaking changes

No breaking changes.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/order-promotion.e2e-spec.ts
+++ b/packages/core/e2e/order-promotion.e2e-spec.ts
@@ -1660,7 +1660,6 @@ describe('Promotions applied to Orders', () => {
     });
 
     // https://github.com/vendurehq/vendure/issues/4353
-    // https://github.com/vendurehq/vendure/issues/4353
     describe('usage limit for auto-applied promotions (no coupon code)', () => {
         async function createNewActiveOrder() {
             const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
--- a/packages/core/src/common/constants.ts
+++ b/packages/core/src/common/constants.ts
@@ -85,4 +85,6 @@ export const CacheKey = {
     ActiveTaxZone: (channelId: ID) => `ActiveTaxZone:${channelId}`,
     ActiveTaxZone_PPA: (channelId: ID) => `ActiveTaxZone_PPA:${channelId}`,
     CollectionVariantCounts: 'CollectionService.getProductVariantCounts',
+    ExhaustedPromotions: (channelId: ID, customerId: ID | undefined) =>
+        `ExhaustedPromotions:${channelId}:${customerId ?? 'guest'}`,
 };
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -2129,8 +2129,10 @@ export class OrderService {
 
         // Filter out auto-applied promotions that have exceeded their usage limits.
         // Coupon-based promotions are already validated in validateCouponCode().
+        // Safe to cache per-request: the active promotion list (allPromotions) is stable
+        // within a single request, so the exhausted set won't change mid-request.
         const customerId = order.customerId;
-        const cacheKey = `exhausted-promotions-${ctx.channelId}-${customerId ?? 'guest'}`;
+        const cacheKey = CacheKey.ExhaustedPromotions(ctx.channelId, customerId);
         const exhaustedIds = await this.requestCache.get(ctx, cacheKey, () =>
             this.promotionService.getExhaustedPromotionIds(ctx, allPromotions, customerId),
         );
--- a/packages/core/src/service/services/promotion.service.ts
+++ b/packages/core/src/service/services/promotion.service.ts
@@ -352,7 +352,6 @@ export class PromotionService {
         const exhaustedIds = new Set<string>();
 
         const withUsageLimit = promotions.filter(p => !p.couponCode && p.usageLimit != null);
-        const withPerCustomerLimit = promotions.filter(p => !p.couponCode && p.perCustomerUsageLimit != null);
 
         if (withUsageLimit.length) {
             const usageCounts = await this.getUsageCountsBatch(
@@ -370,6 +369,10 @@ export class PromotionService {
         // perCustomerUsageLimit can only be checked if we know the customer.
         // For guest checkouts without a customer, we skip this check
         // (matching the existing behavior in validateCouponCode).
+        // Skip promotions already exhausted by the global usageLimit check above.
+        const withPerCustomerLimit = promotions.filter(
+            p => !p.couponCode && p.perCustomerUsageLimit != null && !exhaustedIds.has(p.id.toString()),
+        );
         if (customerId && withPerCustomerLimit.length) {
             const perCustomerCounts = await this.getUsageCountsBatch(
                 ctx,


michaelbromley:
## ~Performance concern: PostgreSQL join table index~ Not an issue

~~Nice work on this PR overall — the batch queries and per-request caching are well thought out.~~

~~One thing to flag: the `getUsageCountsBatch` query filters the order-promotions junction table by `promotionId`...~~

**Update:** I checked the actual PostgreSQL schema and the concern is not valid. TypeORM auto-creates individual btree indexes on both FK columns of `@JoinTable()` junction tables, regardless of the database driver:

```
Table "public.order_promotions_promotion"
Indexes:
    "PK_..." PRIMARY KEY, btree ("orderId", "promotionId")
    "IDX_2c26b..." btree ("promotionId")   <-- already exists
    "IDX_67be0..." btree ("orderId")
```

So even though PostgreSQL itself doesn't auto-create FK indexes, TypeORM handles it at the ORM layer. The `WHERE promotion.id IN (:...promotionIds) GROUP BY promotion.id` query will use the existing index on `promotionId` just fine. No migration needed.

biggamesmallworld (commented):
[This review was posted in error — please disregard]

Ryrahul:
> ## Review
> Nice work on this one, @HouseinIsProgramming! This is a real, high-priority bug and the fix is well-structured. The batch query approach is the right call — much better than N+1 individual counts — and the shared `placedOrdersWithPromotionQb()` is a clean refactor. The e2e tests cover the core scenarios well.
> 
> A few things to address before this is ready to merge:
> 
> ### Must fix
> **1. `order.customer` may not always be loaded**
> 
> In `applyPriceAdjustments()`, you're reading `order.customer?.id` to get the customer. This works for most call sites since `getOrderOrThrow()` includes `'customer'` in its default relations, but `applyPriceAdjustments` is called from ~13 places and some callers pass custom relation sets that may omit `customer` (e.g., the `modifyOrder` path).
> 
> When the relation isn't loaded, `order.customer?.id` silently returns `undefined`, and the per-customer limit check gets skipped entirely — no error, just wrong behavior.
> 
> **Suggestion:** You could hydrate the relation if missing:
> 
> ```ts
> const customerId = order.customer?.id
>     ?? (await this.connection.getRepository(ctx, Order)
>         .findOne({ where: { id: order.id }, relations: ['customer'], select: ['id'] }))?.customer?.id;
> ```
> 
> Or, if that feels heavy, at minimum add a comment documenting the limitation.
> 
> **2. Add an empty-array guard in `getUsageCountsBatch`**
> 
> TypeORM's `IN (:...promotionIds)` will produce invalid SQL if `promotionIds` is empty. Your callers guard this today with `if (withUsageLimit.length)`, but the private method should be self-defensive in case it's called from elsewhere later:
> 
> ```ts
> private async getUsageCountsBatch(...): Promise<Map<string, number>> {
>     if (!promotionIds.length) return new Map();
>     // ... existing code
> }
> ```
> 
> ### Recommended
> **3. Cache key could go stale when customer is set mid-request**
> 
> The cache key is `exhausted-promotions-${ctx.channelId}-${customerId ?? 'guest'}`. If a guest gets assigned a customer during checkout (via `setCustomerForOrder`), and then `applyPriceAdjustments` is called again, the cached result was computed without the customer ID.
> 
> Consider including `order.id` in the cache key, or honestly — given the batch query is already efficient — you could skip the cache entirely. Up to you!
> 
> **4. A few more test cases would strengthen confidence:**
> 
> * A promotion with **both** `usageLimit` and `perCustomerUsageLimit` set
> * A guest checkout (no customerId) with `perCustomerUsageLimit` — should still get the discount
> * Verifying coupon-code promotions are **not** affected by this new filtering (regression guard)
> 
> ### Minor / optional
> * The new cache key string could follow the `CacheKey` constants pattern from `packages/core/src/common/constants.ts` for consistency with the rest of the codebase.
> * The `leftJoin` → `innerJoin` change in the base QB is 👍 — semantically more honest since the WHERE clause already makes it an inner join.
> 
> Overall this is a solid approach to a real problem. The batch design and the separation between coupon-based and auto-applied validation is clean. Looking forward to the next iteration! 🙌

Thanks a lot for working on this @HouseinIsProgramming  really great job on the fix and the overall approach 🙌

This was something we had to handle with custom workarounds on our side, so having it properly handled in core will simplify things a lot for us. Appreciate the effort that went into this

Regarding the point about order.customer possibly not being loaded would it make sense to rely on order.customerId directly instead of hydrating the relation? Since the ID is already available on the entity, that might avoid the extra query while still allowing the per-customer limit check to work correctly.

PR #4407, opened 2026-02-24
 Fix relational custom field not saving on draft order

### LINKED ISSUE #4380: Custom fields not saving correctly on draft orders
**Describe the bug**
The PR https://github.com/vendurehq/vendure/pull/4206 fixed the issue where relations would create two input fields (`fieldName` and `fieldNameId`) for update mutations.

This still fails for custom fields on draft orders.

The problem is this check:

```
    // Skip processing if there are no custom fields
    if (!entity.customFields || !processedEntity.customFields) {
        return processedEntity;
    }
```

in [transformRelationFields](https://github.com/vendurehq/vendure/blob/master/packages/dashboard/src/lib/framework/form-engine/utils.ts#L40).

The customFields are stored in `entity.input.customFields` so the check will fail.

When editing a Order not in Draft, everything works just fine.

**To Reproduce**

1. Add custom relation field to `Order`
2. Create Draft order
3. Try to edit the custom field and click "Set custom fields"
4. See error in Network tab

**Expected behavior**
Custom fields should only be added as `fieldNameId` to the mutations input values.

**Actual behavior**
Custom fields are added as `fieldNameId` and `fieldName` to the mutations input values.

**Error logs**

```
{
    "errors": [
        {
            "message": "Variable \"$input\" got invalid value { fieldName: null, fieldNameId: \"3\" } at \"input.customFields\"; Field \"fieldName\" is not defined by type \"UpdateOrderCustomFieldsInput\". Did you mean \"fieldNameId\"?",
            "locations": [
                {
                    "line": 1,
                    "column": 51
                }
            ],
            "extensions": {
                "code": "BAD_USER_INPUT"
            }
        }
    ]
}
```

- Nodejs version: 24.11.0

**Configuration**

```typescript
      name: "fieldName",
      type: "relation",
      entity: SomeEntity,
      nullable: true,
      eager: true,
      // public: false,
    });
```

Ryrahul:
## Description

Fixes #4380 

Custom relation fields on draft orders were sending both `fieldName` and `fieldNameId` in the mutation input, causing a GraphQL validation error:

    Field "fieldName" is not defined by type "UpdateOrderCustomFieldsInput".
    Did you mean "fieldNameId"?

## Root cause

The `transformRelationFields` function in `packages/dashboard/src/lib/framework/form-engine/utils.ts` only looked for `customFields` at the top level of the entity object. This worked for normal orders because `useDetailPage` calls `useGeneratedForm` with `varName: 'input'`, which unwraps the mutation variable so `customFields` appears at the top level.

Draft orders cannot use `varName: 'input'` because the `SetDraftOrderCustomFields` mutation takes two top-level variables (`$orderId: ID!` and `$input: UpdateOrderInput!`). The draft order page calls `useGeneratedForm` directly with `varName: undefined`, meaning the entity has the shape `{ orderId, input: { customFields } }`. The old code's early-return check (`if (!entity.customFields)`) would bail out since `customFields` is nested inside `input`.

## Fix

Refactored `transformRelationFields` to recursively walk the fields tree. When it encounters a non-scalar nested field (like `input`), it recurses into it. When it finds `customFields`, it processes the relation fields at that level. This matches the pattern already used by `removeEmptyIdFields`, `convertEmptyStringsToNull`, and `stripNullNullableFields` in the same file.

## Breaking changes

None.

## Checklist


grolmus (changes requested):
## Review Summary

Good fix for #4380. The recursive approach in `transformRelationFields` is clean and consistent with the patterns already established by `removeEmptyIdFields`, `convertEmptyStringsToNull`, and `stripNullNullableFields` in the same file. Tests are focused and well-structured.

### Must Fix

**Add array handling in the recursion**

For consistency with the other recursive utilities in this file (e.g., `removeEmptyIdFields` handles arrays for translations), the recursion should also handle array fields:

```typescript
} else if (field.typeInfo && !field.isScalar && entity[field.name] != null) {
    if (Array.isArray(entity[field.name])) {
        processedEntity[field.name as keyof E] = entity[field.name].map(
            (item: any) => transformRelationFields(field.typeInfo!, item)
        ) as any;
    } else {
        processedEntity[field.name as keyof E] = transformRelationFields(
            field.typeInfo,
            entity[field.name],
        );
    }
}
```

There may not be an immediate use case for arrays containing `customFields`, but the other utilities in this file handle it and this function should be consistent.

### Looks Good

- Recursion approach matches existing file patterns
- Shallow copies at each level (avoids unnecessary deep cloning)
- Immutability test is a nice addition
- Surgical scope — only touches what's needed

grolmus (approved):
Looks good now. The recursive approach is clean, consistent with the other utilities in the same file, and the array handling was added as requested.

Tests cover all code paths: nested single/list relations, array fields, and immutability.

One suggestion (non-blocking): consider adding a dashboard e2e test for draft order relation custom fields — there's currently no e2e coverage for this specific flow.

michaelbromley:
Hey @Ryrahul,

Thanks for the PR. I just merged a related issue (#4495) which now caused a conflict. Would you be able to take a look at that and see if you can resolve?

LucidityDesign:
@Ryrahul Great work!

I'm having the same issue, when adding custom relational fields to the new Api Keys. Can you double check, if your solution fixes the problems with Api Keys as well, please?

Ryrahul:
> @Ryrahul Great work!
> 
> I'm having the same issue, when adding custom relational fields to the new Api Keys. Can you double check, if your solution fixes the problems with Api Keys as well, please?

image
Yeah , its working

Ryrahul:
@LucidityDesign  they api key page already uses useDetailPage, which always passes varName: 'input' to useGeneratedForm. So it should already work without this fix [check here](https://github.com/vendurehq/vendure/blob/master/packages/dashboard/src/app/routes/_authenticated/_api-keys/api-keys_.%24id.tsx#L85)

LucidityDesign:
Sorry, running `codegen` fixed this issue

michaelbromley (approved):

PR #4414, opened 2026-02-25
fix(dashboard): Ensure nullable numeric fields default to null

### LINKED ISSUE #4328: Dashboard: When creating a promotion, you cannot set the Usage limit and per customer usage limit fields to null. In the admin-ui, the fields are empty by default.
1. Go to  https://demo.vendure.io/admin/promotions/new

Image

grolmus:
## Summary
- Nullable Int, Float, Money, and DateTime fields now correctly default to null instead of 0
- Fixes promotion usageLimit and perCustomerUsageLimit fields defaulting to 0 when creating new promotions

## Test plan
- [x] Unit tests pass
- [x] Manual test: Create new promotion, verify usageLimit fields are empty (null) not 0

Fixes #4328

### DIFF
--- a/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts
+++ b/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts
@@ -366,6 +366,14 @@ export function getDefaultValueFromField(field: FieldInfo, defaultLanguageCode?:
                 return defaultLanguageCode || 'en';
             case 'Boolean':
                 return false;
+            case 'Int':
+            case 'Float':
+            case 'Money':
+            case 'DateTime':
+                // Nullable numeric/date fields should default to null, not 0.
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hey @grolmus, thanks for looking into this! However this is already fixed on master via #4339 (commit 80cdd078, merged Feb 21).

The `getDefaultValueFromField` function's `default` branch in the `nullable` switch already returns `null` for all unhandled scalar types (including `Int`, `Float`, `Money`, `DateTime`), so the explicit cases added here are redundant.

On top of that, the promotion form's number inputs use `value={field.value ?? ''}` which ensures null values render as empty fields rather than 0.

We verified in the browser on current master that both "Usage limit" and "Per customer usage limit" correctly show as empty when creating a new promotion. Closing as already fixed.

PR #4419, opened 2026-02-25
feat: Make cuopon code validation case insensitive

### LINKED ISSUE #4364: Coupon codes are case-sensitive in backend but Admin UI doesn’t enforce/indicate casing
**Describe the bug**

Coupon codes are validated case-sensitively in the backend, but the Admin Dashboard UI doesn't enforce or indicate any casing requirements. This creates an inconsistent UX where admins can create codes like `AbC10`, but customers entering `abc10` get an "invalid coupon" error.

**To Reproduce**

1. Go to Dashboard → Marketing → Promotions
2. Create a new promotion with coupon code: `Summer20` (enabled, valid dates, assigned to channel)
3. In Shop API, run `applyCouponCode(couponCode: "summer20")`
4. See error: `CouponCodeInvalidError`
5. Run `applyCouponCode(couponCode: "Summer20")` (exact case)
6. Coupon applies successfully

**Expected behavior**

Either:
- Coupon codes should be case-insensitive (industry standard - WooCommerce, Shopify, etc. all do this), **or**
- Dashboard should auto-uppercase codes on save and display a hint that codes are case-sensitive

**Actual behavior**

Backend performs exact string comparison (`promotion.couponCode !== couponCode` in `validateCouponCode()`), but Dashboard accepts any casing without warning.

**Screenshots/Videos**

N/A - easily reproducible

**Error logs**

```graphql
{
  "data": {
    "applyCouponCode": {
      "__typename": "CouponCodeInvalidError",
      "errorCode": "COUPON_CODE_INVALID_ERROR",
      "message": "Coupon code \"summer20\" is not valid",
      "couponCode": "summer20"
    }
  }
}
```
**Environment:**

- Nodejs version: 20.x
- Database: PostgreSQL
- Operating System: Linux (Ubuntu)
- Browser: N/A (API issue)
- Package manager: npm

**Configuration**

N/A - default Vendure promotion configuration

**Minimal reproduction**

```typescript
// 1. Create promotion via Admin API with couponCode: "Test123"
// 2. Apply via Shop API:
const result = await shopClient.query(gql`
  mutation {
    applyCouponCode(couponCode: "test123") {
      ... on Order { id }
      ... on CouponCodeInvalidError { errorCode message }
    }
  }
`);
// Result: CouponCodeInvalidError
```
**Workaround**

Create a plugin that overrides `applyCouponCode` to do case-insensitive lookup:

```typescript
const promotion = await promotionRepo
  .createQueryBuilder('promotion')
  .where('LOWER(promotion.couponCode) = LOWER(:code)', { code: couponCode })
  .getOne();
```
**Additional context**

- This is a UX inconsistency between Dashboard and Shop API
- Most e-commerce platforms treat coupon codes as case-insensitive
- The fix could be either in `validateCouponCode()` (use `LOWER()`) or in Dashboard (force uppercase on save)

---

Ready to paste at: https://github.com/vendure-ecommerce/vendure/issues/new

Ryrahul:
## Description

Coupon codes are now validated **case-insensitively**, matching industry-standard behavior (Shopify, WooCommerce, Stripe, etc.).

Previously, a promotion with the coupon code `Summer20` would reject `summer20` and throw a `CouponCodeInvalidError`.  
With this update, different casing variations of the same coupon code are treated as equivalent.

---

## Changes

- `validateCouponCode()` in `PromotionService` now uses `LOWER()` in the database query via TypeORM’s `Raw()` operator, and the redundant case-sensitive JS check (`promotion.couponCode !== couponCode`) has been removed.
- `applyCouponCode()` now stores the canonical coupon code from the `Promotion` entity instead of the user-typed casing.
- `removeCouponCode()`, `Promotion.test()`, and `OrderModifier` coupon comparisons have been updated to use case-insensitive matching.
- Updated JSDoc on `validateCouponCode()` to document case-insensitive behavior.
- Updated e2e test from **"coupon code application is case-sensitive"** to **"coupon code application is case-insensitive"**.

---

## Breaking Changes

None.

Existing coupon codes continue to work as before. Codes that previously failed due to casing differences will now succeed.

---

### 📌 Always
- [x] I have set a clear title.
- [x] My PR is small and contains a single feature.
- [x] I have checked my own PR.

### 👍 Most of the time
- [x] I have added or updated test cases.

---

Fixes: #4364


grolmus (changes requested):
## Review Summary

The canonical code approach is the right architecture — storing the promotion's actual coupon code rather than the user-typed casing avoids data inconsistency. This is clearly the better approach compared to #4438. However, there are a few things that need addressing before merge.

### Must Fix

**Extract a `couponCodesMatch()` helper function**

There are 7 separate locations in this diff with the pattern `cc.toLowerCase() === something.toLowerCase()`. This is a maintenance problem — if the normalization logic ever needs to change (e.g., trimming whitespace, using `toLocaleLowerCase()`), every instance needs to be found and updated. Extract a simple helper:

```typescript
function couponCodesMatch(a: string, b: string): boolean {
    return a.toLowerCase() === b.toLowerCase();
}
```

Place it somewhere accessible to the services that need it (e.g., `packages/core/src/service/helpers/utils/`).

### Should Fix

**1. Missing test for case-insensitive removal**

There's no test that applies a coupon with one casing and removes it with a different casing. For example: apply `TESTCOUPON`, then remove with `testcoupon`. The code handles this correctly (via `matchedCode` in `removeCouponCode`), but it needs test coverage.

**2. Missing test for case-insensitive de-duplication**

There's no test that verifies applying the same coupon code with different casings doesn't create duplicates. For example: apply `testcoupon`, then apply `TESTCOUPON` — assert the order still only has one coupon code and no duplicate history entry is created.

### Nitpicks (non-blocking)

- The `Raw()` TypeORM operator is new to this codebase (no existing usages). A brief comment above the usage explaining why it's needed (case-insensitive DB lookup) would help the next developer who encounters it.
- The useful comment `// This is a new coupon code that hadn't been applied before` was removed in `order-modifier.ts` — consider keeping it, as it explains the "why" of the check.

### Notes

- `Raw()` with parameterized `:couponCode` binding is safe — no SQL injection risk.
- `LOWER()` is supported across all three Vendure DB backends (MariaDB, PostgreSQL, SQLite).
- The `canonicalCouponCodes` dedup logic in `order-modifier.ts` correctly handles the edge case where `input.couponCodes` contains the same code with different casings.

Ryrahul:
> ## Review Summary
> The canonical code approach is the right architecture — storing the promotion's actual coupon code rather than the user-typed casing avoids data inconsistency. This is clearly the better approach compared to #4438. However, there are a few things that need addressing before merge.
> 
> ### Must Fix
> **Extract a `couponCodesMatch()` helper function**
> 
> There are 7 separate locations in this diff with the pattern `cc.toLowerCase() === something.toLowerCase()`. This is a maintenance problem — if the normalization logic ever needs to change (e.g., trimming whitespace, using `toLocaleLowerCase()`), every instance needs to be found and updated. Extract a simple helper:
> 
> ```ts
> function couponCodesMatch(a: string, b: string): boolean {
>     return a.toLowerCase() === b.toLowerCase();
> }
> ```
> 
> Place it somewhere accessible to the services that need it (e.g., `packages/core/src/service/helpers/utils/`).
> 
> ### Should Fix
> **1. Missing test for case-insensitive removal**
> 
> There's no test that applies a coupon with one casing and removes it with a different casing. For example: apply `TESTCOUPON`, then remove with `testcoupon`. The code handles this correctly (via `matchedCode` in `removeCouponCode`), but it needs test coverage.
> 
> **2. Missing test for case-insensitive de-duplication**
> 
> There's no test that verifies applying the same coupon code with different casings doesn't create duplicates. For example: apply `testcoupon`, then apply `TESTCOUPON` — assert the order still only has one coupon code and no duplicate history entry is created.
> 
> ### Nitpicks (non-blocking)
> * The `Raw()` TypeORM operator is new to this codebase (no existing usages). A brief comment above the usage explaining why it's needed (case-insensitive DB lookup) would help the next developer who encounters it.
> * The useful comment `// This is a new coupon code that hadn't been applied before` was removed in `order-modifier.ts` — consider keeping it, as it explains the "why" of the check.
> 
> ### Notes
> * `Raw()` with parameterized `:couponCode` binding is safe — no SQL injection risk.
> * `LOWER()` is supported across all three Vendure DB backends (MariaDB, PostgreSQL, SQLite).
> * The `canonicalCouponCodes` dedup logic in `order-modifier.ts` correctly handles the edge case where `input.couponCodes` contains the same code with different casings.

Thanks for the review and feedback, will address the issues 

- The useful comment // This is a new coupon code that hadn't been applied before was removed in order-modifier.ts — consider keeping it, as it explains the "why" of the check.  Ig this comment is still present and has not been removed

grolmus (dismissed):
All requested changes have been addressed:

- `couponCodesMatch()` helper extracted to `service/helpers/utils/` and used at all comparison sites.
- Added e2e coverage for case-insensitive removal and case-insensitive de-duplication.
- Added the explanatory comment above the `Raw()`/`LOWER()` lookup.

Verified locally: rebuilt `@vendure/core` and ran the `order-promotion` e2e suite against the fresh build — 63 passed, 1 skipped. CI is green across all four DB backends.

Thanks @Ryrahul, LGTM 🚀

TheHypnoo:
@michaelbromley Sorry, could you review this PR so you can include it in the next Vendure update? Thanks.

michaelbromley (approved):

michaelbromley:
Thanks all for your work on this. Will be in the minor release.

PR #4420, opened 2026-02-25
fix: Silent error when creating product variant when no active tax zo…

### LINKED ISSUE #4358: Creating variants for a product with options fails + any dashboard errors are being swallowed
**Describe the bug**

1. Creating variants for a product with options fails
2. Whenever anything fails, there is no feedback to the admin.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a product.
2. Create an option group and variants
3. Click save.
4. Option groups are created, but variant creation fails with `The active tax zone could not be determined. Ensure a default tax zone is set for the current channel.`
5. Try recreating the variants again, which only shows 1 variant input field, which will fail with a different error, because a variant for each option should be given.

**Expected behavior**
1. I would expect to see an error if variant creation fails.
2. I would expect that clicking `Create variants` again would prompt the 3 variant fields again.

**Actual behavior**
Both errors are swallowed; admin doesn't see any error but does see that no variants are created.

**Screenshots/Videos**

Option groups are created

Image

But creating variants doesnt prompt variant fields for all three options:

Image

- Nodejs version: 24
- Operating System (Windows/macOS/Linux): Ubuntu
- Browser (if applicable): Brave

Ryrahul:
## Description

The **"Create Variants"** dialog silently swallowed errors during variant creation.  
If the backend returned an error (e.g., *"The active tax zone could not be determined"*), the admin saw no feedback , the `catch` block only had `console.error` with an unimplemented TODO comment.

Added a `toast.error()` notification so the admin now sees the error message when variant creation fails.

---

**Note:** There is a second related issue where retrying after a failure only shows **1 variant input field** instead of all expected fields. This happens because `handleCreateVariants` runs three sequential steps via `Promise.all` / `mutateAsync`:

1. Create option groups (persisted to server)  
2. Assign option groups to product (persisted to server)  
3. Create variants (fails)

Steps 1–2 succeed and persist before step 3 fails, leaving the product with option groups but no variants. When the dialog reopens, it mounts fresh with empty local state, so `generateVariants([])` returns a single default variant.

Fixing this properly requires either:
- Splitting the flow into separate phases
- Detecting already-assigned option groups on dialog open
- Adding rollback logic  

This needs further discussion on the preferred approach.

---

## Breaking changes

None.

---

## Checklist

📌 **Always:**

Partially fixes #4358

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_products/components/create-product-variants-dialog.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_products/components/create-product-variants-dialog.tsx
@@ -10,11 +10,12 @@ import {
 } from '@/vdb/components/ui/dialog.js';
 import { api } from '@/vdb/graphql/api.js';
 import { useChannel } from '@/vdb/hooks/use-channel.js';
-import { Trans } from '@lingui/react/macro';
 import { normalizeString } from '@/vdb/lib/utils.js';
+import { Trans, useLingui } from '@lingui/react/macro';
 import { useMutation } from '@tanstack/react-query';
 import { Plus } from 'lucide-react';
 import { useCallback, useState } from 'react';
+import { toast } from 'sonner';
 import {
     addOptionGroupToProductDocument,
     createProductOptionGroupDocument,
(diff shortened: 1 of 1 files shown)


Ryrahul:
Closing , As its fixed in #4460

PR #4427, opened 2026-02-26
fix(dashboard): Handle nested customFields in transformRelationFields

### LINKED ISSUE #4380: Custom fields not saving correctly on draft orders

grolmus:
## Summary

- Make `transformRelationFields` recursive to process customFields at any nesting level
- Fix draft orders where customFields are nested inside `input` object (`input.customFields`) rather than at root level
- Fix relation field clearing by using `null` instead of `undefined` (so the value is sent to the server)

## Changes

- `utils.ts`: Refactored `transformRelationFields` to use a separate `processRelationCustomFields` helper that recursively processes nested objects
- `relation-selector.tsx`: Changed `onChange(undefined)` to `onChange(null)` when clearing a single relation field
- Added test cases for nested customFields scenarios

Fixes #4380

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_orders/orders_.draft.$id.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_orders/orders_.draft.$id.tsx
@@ -7,6 +7,7 @@ import { Button } from '@/vdb/components/ui/button.js';
 import { Form } from '@/vdb/components/ui/form.js';
 import { addCustomFields } from '@/vdb/framework/document-introspection/add-custom-fields.js';
 import { useGeneratedForm } from '@/vdb/framework/form-engine/use-generated-form.js';
+import { useCustomFieldConfig } from '@/vdb/hooks/use-custom-field-config.js';
 import {
     Page,
     PageActionBar,
@@ -69,6 +70,30 @@ function DraftOrderPage() {
         params: { id: params.id },
     });
 
+    const orderCustomFields = useCustomFieldConfig('Order');
+
+    const transformCustomFieldsForForm = (customFields: Record<string, any> | undefined) => {
+        if (!customFields) return {};
+        const result: Record<string, any> = {};
+        for (const [key, value] of Object.entries(customFields)) {
+            const fieldConfig = orderCustomFields.find(f => f.name === key);
+            if (fieldConfig?.type === 'relation') {
+                // Convert relation objects to IDs for the form
+                const idKey = fieldConfig.list ? `${key}Ids` : `${key}Id`;
+                if (value === null) {
+                    result[idKey] = null;
+                } else if (Array.isArray(value)) {
+                    result[idKey] = value.map((v: { id: string }) => v.id);
+                } else if (value?.id) {
+                    result[idKey] = value.id;
+                }
+            } else {
+                result[key] = value;
+            }
+        }
+        return result;
+    };
+
     const { form: orderCustomFieldsForm } = useGeneratedForm({
         document: setDraftOrderCustomFieldsDocument,
         varName: undefined,
@@ -78,7 +103,7 @@ function DraftOrderPage() {
                 orderId: entity.id,
                 input: {
                     id: entity.id,
-                    customFields: entity.customFields,
+                    customFields: transformCustomFieldsForForm(entity.customFields),
                 },
             };
         },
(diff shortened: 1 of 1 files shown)


biggamesmallworld:
Nice fix! One thing I noticed: the null handling for list relations doesn't quite match what you've done for single relations.

For single relations, you correctly preserve `null` so it gets sent to the server:

```typescript
customFieldsObj[relationField] = relationValue === null ? null : relationValue?.id;
```

But for list relations, when `relationValue` is `null`, the `if (Array.isArray(relationValue))` check is skipped, so `relationField` never gets set — the `null` is silently dropped. Could you add the same treatment here? Something like:

```typescript
if (customField.list) {
    const propertyAccessorKey = customField.name.replace(/Ids$/, '');
    const relationValue = customFieldsObj[propertyAccessorKey];

    if (relationValue === null) {
        customFieldsObj[relationField] = null;
    } else if (Array.isArray(relationValue)) {
        customFieldsObj[relationField] = relationValue.map((v: { id: string }) => v.id);
    }
    delete customFieldsObj[propertyAccessorKey];
}
```

That way clearing a list relation works the same as clearing a single relation.

biggamesmallworld:
## Additional fix: relation custom fields display & save on draft orders

During testing we found that the original `transformRelationFields` approach doesn't fully work for draft orders because `getOperationVariablesFields` sees `customFields` as an opaque `JSON` scalar (not a typed input like `UpdateOrderCustomFieldsInput`). This means the recursive processing has no field definitions to work with and passes values through unchanged.

This caused two issues reported in #4380:

### 1. Saving failed with `BAD_USER_INPUT`
The form values contained both the relation object (`relatedProduct: null`) and the ID field (`relatedProductId: "28"`). Since `transformRelationFields` couldn't process them, both were sent in the mutation.

### 2. Existing values not displayed (the "second issue" mentioned in #4380)
The `RelationSelector` is bound to `relatedProductId`, but `setValues` passed the raw entity `customFields` which only has `relatedProduct: { id, name, ... }`. Without the transform, `relatedProductId` was never set, so the selector showed nothing.

### Fix
Instead of relying on `transformRelationFields` (which needs typed schema info), the draft order page now uses `useCustomFieldConfig('Order')` to:
- **On init**: Convert relation objects to IDs in `setValues` (e.g. `relatedProduct: { id: "40" }` → `relatedProductId: "40"`)
- **On submit**: Strip relation object keys before sending the mutation (keeping only `relatedProductId`)

This fully covers both issues described in #4380.

biggamesmallworld:
## Manual testing

Verified both fixes locally in the browser with the following dev-config setup:

```typescript
customFields: {
    Order: [
        {
            name: 'relatedProduct',
            type: 'relation',
            entity: Product,
            nullable: true,
            eager: true,
        },
    ],
},
```

After `npm run populate` and starting the dev server + dashboard:

1. Created a draft order, added a customer, product, and shipping method
2. Opened the "Custom fields" section — the `relatedProduct` RelationSelector rendered correctly
3. Selected a product — clicked "Set custom fields" — **mutation succeeded** (no `BAD_USER_INPUT` error)
4. Refreshed the page — **the selected product was displayed** in the RelationSelector (previously it showed nothing)
5. Changed the selection to a different product — saved again — value updated correctly

grolmus:
Closing in favor of #4407, which provides a cleaner recursive fix for `transformRelationFields` that's more consistent with the existing utility patterns in the same file.

This PR did identify a separate real bug: clearing a single relation field uses `onChange(undefined)` in `relation-selector.tsx`, which gets stripped from JSON serialization — so the server never receives the "clear this field" instruction. That fix (changing to `onChange(null)` + handling null explicitly in the transform) will be addressed in a separate focused PR.

PR #4428, opened 2026-02-26
fix(dashboard): Use index-based identification for collection filters

### LINKED ISSUE #4327: Dashboard: Collection filters with same type share state incorrectly
When adding multiple collection filters of the same type (e.g., two "Filter by product variant name" filters), the input fields are incorrectly synchronized. Changing values in one filter causes all filters with the same code to update with identical values. Additionally, the AND/OR combinator toggle reverts to AND regardless of the selected value.

## Steps to Reproduce

1. Open Vendure Admin Dashboard
2. Navigate to **Catalog → Collections**
3. Open any collection or create a new one
4. Click "Add collection filter" and select "Filter by product variant name"
   - Set operator dropdown to "contains"
   - Set term input field to "shirt"
5. Click "Add collection filter" again and select "Filter by product variant name" (same type)
   - Set operator dropdown to "contains"  
   - Set term input field to "pants"
6. Try to change the operator or term in the first filter

**Expected Behavior:**
- Each filter should maintain its own independent state
- Changing values in one filter should not affect other filters
- The AND/OR toggle should persist the selected value

**Actual Behavior:**
- Both filters update with the same values when either is changed
- Input fields are synchronized across all filters with the same code
- The AND/OR toggle reverts to AND after selection

## Environment

- **Vendure Version:** 3.5.2
- **Dashboard Version:** 3.5.2
- **Browser:** Chrome
- **OS:** macOS (likely affects all OS)

## Root Cause

The bug is located in:
```
packages/admin-ui/src/lib/components/shared/configurable-operation-multi-selector.tsx
```

Specifically in the `onOperationValueChange` function (line 143-148):

```typescript
const onOperationValueChange = (
    operation: ConfigurableOperationInputType,
    newVal: ConfigurableOperationInputType,
) => {
    onChange(value.map(op => (op.code === operation.code ? newVal : op)));
};
```

The function identifies filters by their `code` property, which means all filters of the same type share the same code. When updating, it replaces **all** filters with matching codes instead of only the specific filter being edited.

## Proposed Fix

Change the function to use **index-based identification** instead of code-based:

```typescript
const onOperationValueChange = (
    index: number,
    newVal: ConfigurableOperationInputType,
) => {
    onChange(value.map((op, i) => (i === index ? newVal : op)));
};
```

And update the call site (line 209):

```typescript
onChange={value => onOperationValueChange(index, value)}
```

## Impact

This bug affects:
- ✅ Collection filters (primary impact - confirmed)
- ⚠️ Potentially promotion conditions/actions (same component is used)
- ⚠️ Any configurable operation that allows multiple instances of the same type

## Additional Notes

- The `onCombinationModeChange` function already uses index-based updates correctly (line 154-167), so the fix aligns with the existing pattern in the same component
- This is a UI state management bug, not a backend/API issue
- The bug makes it impossible to use multiple filters of the same type with different values

grolmus:
## Summary

- Fixed collection filters with the same type sharing state incorrectly
- Changed `onOperationValueChange` to use index-based identification instead of code-based
- Enabled the existing regression test for this issue

The bug was in `configurable-operation-multi-selector.tsx` where filters were identified by `code`, causing all filters of the same type to update when any one changed.

## Test plan

- [x] Manually verified fix in dashboard (added 2 "Filter by product variant name" filters with different values)
- [ ] CI runs `issue-4327-collection-filter-state.spec.ts` regression test

Fixes #4327


biggamesmallworld:
Hey @grolmus, nice fix! The logic change is correct and clean.

The dashboard e2e failure is a small casing issue in the test. The `variant-name-filter`'s `term` argument has no explicit label, so the UI falls back to the argument name `"term"` (lowercase). The test is looking for `'Term'` (capitalized).

In `issue-4327-collection-filter-state.spec.ts`, lines 33 and 43:

```diff
- has: page.getByText('Term', { exact: true }),
+ has: page.getByText('term', { exact: true }),
```

That should get CI green. 🙂

biggamesmallworld:
The CI has two failures. Here's how to fix them:

### 1. Dashboard e2e test failure (`issue-4327-collection-filter-state.spec.ts`)

The test times out at line 35 looking for a textbox inside `[data-slot="field"]` containing the text `"Term"` (capital T).

The `variant-name-filter` defines its arg as `term: { type: 'string' }` with no explicit `label`. In `configurable-operation-input.tsx:113`, the label renders as `{arg.label || arg.name}` — falling back to `arg.name` = `"term"` (lowercase).

**Fix:** Change `'Term'` to `'term'` on lines 33 and 44 of the test file:

```diff
-            has: page.getByText('Term', { exact: true }),
+            has: page.getByText('term', { exact: true }),
```

### 2. Dashboard i18n sync

The `.po` catalog files are out of sync with the source code.

**Fix:** Run `npm run i18n:extract --workspace=@vendure/dashboard` and commit the updated `.po` files.

---

Also, please enable "Allow edits from maintainers" on this PR so we can push fixes directly if needed.

Issue #4430, opened 2026-02-26
Custom field permission bug on entity updates


Ryrahul:
## Describe the bug

Custom fields with `requiresPermission` on any entity  prevent non-superadmin users from updating any other fields on that entity via the Dashboard UI.

The Dashboard sends all custom field values in the mutation input  including ones the user doesn't have permission to write  causing the server to reject the entire update.

---

## To Reproduce

1. Add a custom field on Collection with requiresPermission: [Permission.SuperAdmin]

   config.customFields.Collection.push({
     name: 'isBrandCollection',
     type: 'boolean',
     defaultValue: false,
     public: true,
     requiresPermission: [Permission.SuperAdmin],
   });

2. On Dashboard UI, login as a non-superadmin (e.g. a seller admin)
3. Go to any Collection detail page
4. Edit any field (e.g. the collection name) and click update
5. See error

---

## Expected behavior

Since the user is not modifying the isBrandCollection custom field, but only updating other fields they have permission to edit, the update should succeed.

---

## Actual behavior

The mutation fails with a permission error because the Dashboard includes the isBrandCollection field in the mutation payload even though the user never touched it.

---

## Error logs

Failed to update collection:  
GraphQL Request Error: You do not have the required permissions to update the "isBrandCollection" field.  
[Vendure Server] USER_INPUT_ERROR: error.field-invalid-no-permission

---

## Environment

- @vendure/core version: 3.5.4  
- Dashboard: React (new)

---

## Minimal reproduction

config.customFields.Collection.push({
  name: 'isBrandCollection',
  type: 'boolean',
  defaultValue: false,
  public: true,
  requiresPermission: [Permission.SuperAdmin],
});

---

## Workaround

Use the GraphQL Playground or log in as SuperAdmin to perform updates.

---

## Additional context

- This affects all entities with custom fields, not just Collection — any entity detail page in the Dashboard copies raw query data (including non-permitted custom fields) into the mutation input.
- The root cause is in removeReadonlyAndLocalizedCustomFields in packages/dashboard/src/lib/utils.ts — it strips readonly and locale fields but does not strip fields the user lacks permission for.

Relates to #4426

mehringer68:
[...]

PR #4431, opened 2026-02-26
fix: Remove non permitted customfield from mutation before submit

### LINKED ISSUE #4426: Dashboard UI: Custom fields on product variant with superadmin permissions interfering with editing other fields on the variant by non-superuser admin.
**Describe the bug**
Custom field on product variant with superadmin permissions required is interfering with other admins with being able to update any of the other fields on the product variant form the Dashboard UI

**To Reproduce**
1.Add a custom field on the product variant with `requiresPermission: [Permission.SuperAdmin]`
2. On Dashboard UI login as non-superadmin admin (like a seller admin)
3. Go to any product variant detail page
4. Edit any field and click on update.
5. You will see an error saying that the user doesn't have permission to update the custom field

**Expected behavior**
Since I am not actually updating the custom field, but am updating other fields that I have permissions to view and edit, I should be able to do that. This works fine on the Admin UI

**Error logs**
- Failed to update product variant:
  GraphQL Request Error: You do not have the required permissions to update the "xxxxxxx" field.

```
[Vendure Server] USER_INPUT_ERROR: error.field-invalid-no-permission
```

**Minimal reproduction**

```typescript
 config.customFields.ProductVariant.push({
      name: 'some-var',
      type: 'boolean',
      defaultValue: false,
      public: true,
      requiresPermission: [Permission.SuperAdmin],
    });
```

**Workaround**
using the superadmin to make the updates for now.

Ryrahul:
## Description

Fixes the issue where custom fields with `requiresPermission` (e.g. `Permission.SuperAdmin`) were being included in update mutations even for users who lack those permissions, causing the server to reject the entire update with error `field-invalid-no-permission`.

### Root cause

The Dashboard's `setValuesForUpdate` copies all custom field values from the raw GraphQL query response (including non-permitted fields) into the mutation input. The existing `removeReadonlyAndLocalizedCustomFields` function only stripped `readonly` and `localeString/localeText` fields it had no concept of permission-based filtering.

### Fix

Updated `removeReadonlyAndLocalizedCustomFields` in `packages/dashboard/src/lib/utils.ts` to also strip any custom field key not present in the permitted config list (which is already filtered by `useCustomFieldConfig()` based on user permissions). This covers both root `customFields` and `translation.customFields`.

Also handles `relation` type custom fields which use `Id/Ids` suffixed names in the mutation input after being transformed by `transformRelationFields()`.

This affects all entities with custom fields 
---

## Breaking changes

No breaking changes.

---

## Checklist

Fixes: #4426  & #4430


grolmus (approved):
Logic is correct. Key points verified:

1. `permittedFieldNames` Set derived from the already-permission-filtered config is the right approach
2. Relation field handling (`nameId`/`nameIds` variants) correctly accounts for `transformRelationFields()` running before this
3. Removing the `!customFieldConfigs?.length` early return is correct — when config is empty (no permitted fields), ALL custom fields should be stripped
4. Both root and translation custom fields are cleaned

Minor nit: no tests added. The `utils.ts` likely has existing test coverage — would be nice to add a case for permission-filtered custom fields, but not blocking.

Ryrahul:
@grolmus  can this be included in next release ? Will add the test cases if required before merge

michaelbromley:
Hi!

My review flagged this potential issue, can you look into it?

---

Readonly relation fields still leak through. readonlyFieldNames is built from config.name only (e.g. featuredProduct), but by the time this runs the key has been renamed to featuredProductId / featuredProductsIds. So a readonly: true relation field:
- isn't matched by the readonly loop (wrong name)
- is in the permitted set (under its Id/Ids form)
- ends up in the mutation input anyway

CodeRabbit flagged this; the follow-up commits didn't address it. Five-line fix:

```ts
const readonlyFieldNames = customFieldConfigs.flatMap(config => {
    if (config.readonly !== true) return [];
    if (config.type === 'relation') {
        return [config.name, `${config.name}Id`, `${config.name}Ids`];
    }
    return [config.name];
});
```

—

Also, since this is covering quite a complex array of different cases, some unit tests would probably make sense at this point - should be very doable with a pure function like this.

michaelbromley (approved):

PR #4434, opened 2026-02-27
fix(core): use channel defaultLanguageCode as translation fallback in FacetService and FacetValueService

mr-mpage:
## Problem

`FacetService.findByCode()` and `FacetValueService.findAll()` both had incomplete implementations, each marked with an explicit `// TODO: Implement usage of channelLanguageCode` comment. Both methods called `translateDeep()` with only a single `LanguageCode`, ignoring `ctx.channel.defaultLanguageCode` as a fallback.

The `translateDeep()` function already supports a fallback chain via `[LanguageCode, ...LanguageCode[]]`, and the modern `TranslatorService.translate()` already uses this correctly:

```typescript
// translator.service.ts — existing correct pattern
translateDeep(translatable, [ctx.languageCode, ctx.channel.defaultLanguageCode, this.configService.defaultLanguageCode], ...)
```

Without this fallback chain, when a facet has no translation in the requested language, the fallback is always the hardcoded system `DEFAULT_LANGUAGE_CODE` (`en`) rather than the channel's configured default language.

## Impact

Any multilingual storefront where a channel's `defaultLanguageCode` is not `en` receives incorrect translation fallbacks from these methods. Both are called internally during data import:

- `importer.ts` calls `findByCode()` — affects all product CSV imports
- `populator.ts` calls `findAll()` — affects all data seeding

## Example

- Channel configured with `defaultLanguageCode: de`
- Facet has `en` and `de` translations, but not `fr`
- Request arrives with `languageCode: fr`
- **Before:** Returns the `en` translation (wrong — ignores channel default)
- **After:** Returns the `de` translation (correct — falls back to channel default)

## Changes

**`packages/core/src/service/services/facet-value.service.ts`**

```diff
-        const [repository, languageCode] = ctxOrLang instanceof RequestContext
-            ? [this.connection.getRepository(ctxOrLang, FacetValue), lang\!]
-            : [this.connection.rawConnection.getRepository(FacetValue), ctxOrLang];
-        // TODO: Implement usage of channelLanguageCode
+        const [repository, languageCode, channelLanguageCode] = ctxOrLang instanceof RequestContext
+            ? [this.connection.getRepository(ctxOrLang, FacetValue), lang\!, ctxOrLang.channel.defaultLanguageCode]
+            : [this.connection.rawConnection.getRepository(FacetValue), ctxOrLang, undefined];
         return repository
             .find({ relations: ['facet'] })
             .then(facetValues => facetValues.map(facetValue =>
-                translateDeep(facetValue, languageCode, ['facet'])));
+                translateDeep(facetValue, channelLanguageCode ? [languageCode, channelLanguageCode] : languageCode, ['facet'])));
```

**`packages/core/src/service/services/facet.service.ts`**

```diff
-        const [repository, facetCode, languageCode] = ctxOrFacetCode instanceof RequestContext
-            ? [this.connection.getRepository(ctxOrFacetCode, Facet), facetCodeOrLang, lang\!]
-            : [this.connection.rawConnection.getRepository(Facet), ctxOrFacetCode, facetCodeOrLang as LanguageCode];
-        // TODO: Implement usage of channelLanguageCode
+        const [repository, facetCode, languageCode, channelLanguageCode] = ctxOrFacetCode instanceof RequestContext
+            ? [this.connection.getRepository(ctxOrFacetCode, Facet), facetCodeOrLang, lang\!, ctxOrFacetCode.channel.defaultLanguageCode]
+            : [this.connection.rawConnection.getRepository(Facet), ctxOrFacetCode, facetCodeOrLang as LanguageCode, undefined];
         return repository
             .findOne({ where: { code: facetCode }, relations })
             .then(facet =>
-                (facet && translateDeep(facet, languageCode, ['values', ['values', 'facet']])) ?? undefined);
+                (facet && translateDeep(facet, channelLanguageCode ? [languageCode, channelLanguageCode] : languageCode, ['values', ['values', 'facet']])) ?? undefined);
```

## Tests

Added `packages/core/e2e/facet-translation-fallback.e2e-spec.ts`.

Setup: German (`de`) added to global available languages; second channel created with `defaultLanguageCode: de`; a facet named "Marke" created with EN + DE translations only (no FR), containing a value named "Acme".

| Test | Method | Entity type | Language requested | Expected result |
|------|---------|-------------|-------------------|-----------------|
| exact match | `findByCode()` | Facet | `de` | `Marke (DE)` |
| fallback | `findByCode()` | Facet | `fr` (no translation) | `Marke (DE)` — channel default |
| exact match | `findAll()` | FacetValue | `de` | `Acme (DE)` |
| fallback | `findAll()` | FacetValue | `fr` (no translation) | `Acme (DE)` — channel default |

The fallback tests are the regression case: before the fix, requesting `fr` with no French translation would fall through to `en` (the hardcoded system default), returning `Brand (EN)` / `Acme (EN)` instead of the channel's configured German default.


mr-mpage:

PR #4436, opened 2026-02-27
feat(core): Allow async `OrderMergeStrategy`

twlite:
# Description

This PR adds support for asynchronous `OrderMergeStrategy` support. Since it already is `InjectableStrategy`, it offers a way to access db connections and other services but almost all of them are asynchronous, making it impossible to use without patching the package. This update makes it possible to create asynchronous order merge strategies enabling us to add complex logic.

# Breaking changes

**Does this PR include any breaking changes we should be aware of?**

This update will not break existing strategies but **direct consumption** of the `OrderMerger.merge()` will break as it returns a Promise. However, the direct usage of this api is unlikely and internal use within `core` has been updated to properly await `OrderMerger.merge()` method.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/list-query-builder/list-query-builder.ts
+++ b/packages/core/src/service/helpers/list-query-builder/list-query-builder.ts
@@ -456,7 +456,7 @@ export class ListQueryBuilder implements OnApplicationBootstrap {
 
         // Get the related entity's table and column info
         const inverseEntityMeta = relation.inverseEntityMetadata;
-        const inverseTableName = inverseEntityMeta.tableName;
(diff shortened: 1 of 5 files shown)


michaelbromley:
Thanks for the PR. To be safe please target `minor` since there is at least a small chance of a break if people use this directly.

twlite:
😂 minor branch increased some work haha

michaelbromley:
Hey @twlite can you just take a look at that merge conflict? Then we should be good to merge this. The failing tests are unrelated and are being sorted out separately.

PR #4439, opened 2026-02-27
fix(dashboard): Isolate PageContext in asset picker dialog

grolmus:
## Summary
- Wraps `AssetPickerDialog` content in isolated `PageContext.Provider` to prevent parent page context leakage
- Fixes issue where custom action bar items registered for detail pages appeared inside the asset picker dialog

## Root Cause
React Context propagates through Radix Dialog portals. When `AssetPickerDialog` opened on a detail page, it inherited the parent's `PageContext` (including `pageId`). Extension action bar items registered for that `pageId` were then incorrectly rendered inside the dialog.

## Test plan
- [ ] Open a detail page with custom action bar buttons (e.g., via dashboard extension)
- [ ] Click an asset field to open the asset picker dialog
- [ ] Verify only the "Upload" button appears, not the detail page's custom actions

Fixes OSS-406


biggamesmallworld (commented):
## Review — one suggestion

Clean fix for the Radix portal context propagation issue. The approach is correct and minimal.

### 💡 Suggestion: Add a comment explaining the `PageContext.Provider` isolation

The wrapper at line 82 exists purely to prevent parent page extension action bar items from leaking into the dialog via Radix portal context propagation. Without a comment, a future developer might remove it while "cleaning up" and reintroduce the bug. Something like:

```tsx
{/* Isolate from parent PageContext to prevent extension action bar items
    registered for the parent page from rendering inside this dialog.
    See OSS-406. */}
<PageContext.Provider value={{}}>
```

Optionally, hoisting to `const EMPTY_PAGE_CONTEXT = {};` at module scope would avoid a new object reference per render (minor, since this is a short-lived modal).

LGTM otherwise 👍

biggamesmallworld (approved):

michaelbromley:
The solution looks good, but it raises a question for me - since the dialog is capable of displaying extended ActionBar items, we could actually make _use_ of that as a feature to allow extension authors to target specific asset picker dialogs. If we were to expose an optional pageId prop on the dialog component, we could put that in the new wrapped context provider, to allow that dialog to be specifically targeted. 

Would you be able to check if that's feasible?

grolmus:
> The solution looks good, but it raises a question for me - since the dialog is capable of displaying extended ActionBar items, we could actually make _use_ of that as a feature to allow extension authors to target specific asset picker dialogs. If we were to expose an optional pageId prop on the dialog component, we could put that in the new wrapped context provider, to allow that dialog to be specifically targeted.
>
> Would you be able to check if that's feasible?

Yes — done in d75909169.

**What changed in `AssetPickerDialog`:**
- New optional `pageId?: string` prop. When provided, it is published through the dialog's isolated `PageContext.Provider`, so extensions can target this specific dialog via the standard `actionBarItems` API (e.g. `{ pageId: 'my-asset-picker', component, ... }`).
- Any `DashboardActionBarItem`s registered for that `pageId` are rendered in the dialog footer (filtered to `type !== 'dropdown'`, wrapped in `PermissionGuard` — same handling as `PageActionBar`).
- When `pageId` is omitted, behaviour is identical to before this change (empty context, no leakage from the parent page).

**Verification:**

1. Regression — `Test Button` registered for the `product-detail` page no longer leaks into the picker:
image

2. New feature — registering `{ pageId: 'asset-picker-test', component: PickerAction }` and passing `pageId="asset-picker-test"` to the dialog renders the item in the footer:
image

michaelbromley (approved):

PR #4442, opened 2026-02-27
refactor(core,dashboard): Deprecate built-in health check features

### LINKED ISSUE #4441: refactor(core,dashboard): Deprecate and remove built-in health check features
## Summary

Deprecate and remove the built-in health check functionality from Vendure core and the health check view from the admin dashboard.

## Motivation

Having the application determine its own health is an **anti-pattern**. Health checking is an infrastructure concern and should be handled by the orchestration layer (Kubernetes liveness/readiness probes, load balancer health checks, Docker healthchecks, etc.), not by the application itself.

The current built-in health check has historically caused issues for users, including:
- Vendure instances restarting unexpectedly due to self-reported unhealthy status
- Confusion about whether the app or the infrastructure is responsible for health management
- False positives/negatives leading to unnecessary downtime

Delegating health checks to the infrastructure layer gives operators full control over probe configuration, thresholds, and restart policies without coupling those decisions to application code.

### v3.6.0 (Deprecation)
- Mark all health check related APIs, services, and configuration options as `@deprecated` with migration guidance
- Add deprecation warnings at startup when health check features are actively configured
- Remove the health check view/page from the admin dashboard
- Update documentation to recommend infrastructure-level health checks instead

### v4.0.0 (Removal)
- Fully remove deprecated health check code from `@vendure/core`

## Migration guidance

Users currently relying on Vendure's built-in health checks should migrate to infrastructure-level solutions:

- **Kubernetes**: Use `livenessProbe` and `readinessProbe` on the pod spec
- **Docker Compose**: Use the `healthcheck` directive
- **Load Balancers**: Configure health check endpoints at the LB level
- **Process Managers (PM2, systemd)**: Use native health monitoring

A simple HTTP check against the Vendure server's GraphQL endpoint (or a lightweight custom endpoint) is sufficient to verify the application is running and responsive.

dlhck:
## Summary

Deprecates all built-in health check features in preparation for full removal in v4.0.0. Application-level health checks are an anti-pattern — health monitoring should be handled by infrastructure (Kubernetes probes, Docker healthchecks, load balancer checks).

- Mark all health check interfaces, strategies, services, and config options with `@deprecated` JSDoc tags
- Add startup deprecation warning when `systemOptions.healthChecks` is non-empty
- Change default `healthChecks` config from `[new TypeORMHealthCheckStrategy()]` to `[]`
- Remove health check auto-registration from BullMQ and Elasticsearch plugins
- Remove health check page and nav item from dashboard
- Update `bootstrapWorker` docs to remove health check promotion

The `/health` endpoint remains functional during the deprecation period. Full removal in v4.0.0.

Fixes #4441

### DIFF
--- a/docs/docs/reference/core-plugins/payments-plugin/braintree-plugin.mdx
+++ b/docs/docs/reference/core-plugins/payments-plugin/braintree-plugin.mdx
@@ -253,7 +253,7 @@ class BraintreePlugin {
 
 
 </div>
-<GenerationInfo sourceFile="packages/payments-plugin/src/braintree/types.ts" sourceLine="25" packageName="@vendure/payments-plugin" />
+<GenerationInfo sourceFile="packages/payments-plugin/src/braintree/types.ts" sourceLine="26" packageName="@vendure/payments-plugin" />
 
 Options for the Braintree plugin.
 
@@ -262,6 +262,7 @@ interface BraintreePluginOptions {
     environment?: Environment;
     storeCustomersInBraintree?: boolean;
     extractMetadata?: (transaction: Transaction) => PaymentMetadata;
+    merchantAccountIds?: BraintreeMerchantAccountIds;
 }
 ```
 
@@ -336,6 +337,13 @@ const metadata = {
   }
 }
 ```
+### merchantAccountIds
+
+<MemberInfo kind="property" type={`BraintreeMerchantAccountIds`}   />
+
+An optional mapping of currency codes to Braintree Merchant Account IDs.
+When set, the corresponding Merchant Account ID will be used for transactions
+in the given currency.
 
 
 </div>
--- a/docs/docs/reference/typescript-api/configuration/default-config.mdx
+++ b/docs/docs/reference/typescript-api/configuration/default-config.mdx
@@ -2,7 +2,7 @@
 title: "DefaultConfig"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/core/src/config/default-config.ts" sourceLine="72" packageName="@vendure/core" />
+<GenerationInfo sourceFile="packages/core/src/config/default-config.ts" sourceLine="71" packageName="@vendure/core" />
 
 The default configuration settings which are used if not explicitly overridden in the bootstrap() call.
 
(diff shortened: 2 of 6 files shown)


michaelbromley:
I thought we'd just deprecate the health check code related to registering health checks for _other_ services like DB, elastic etc. If we remove the server's `/health` endpoint altogether, what would then be the recommended way to monitor health for the server process itself?

dlhck:
This would be done on the API endpoints of shop-api or admin-api - they return status 400 if no query is sent. When the shop-api and admin-api responds with that code, you know the server is up and the server process can serve requests.

dlhck:
but we can leave the health endpoint and just return `{"status: "ok"}` without any system checks.

dlhck:
Addressed the feedback — the `/health` endpoint is kept but now just returns `{ status: "ok" }` with a 200.

- Removed the Terminus `HealthCheckService` dependency from the controller
- No health check strategies are registered or executed anymore
- The deprecation warning still fires if users have `systemOptions.healthChecks` configured
- `HealthCheckRegistryService` remains exported from `PluginCommonModule` so third-party plugins don't break

michaelbromley:
How about the worker? Currently we have that service that registers a very minimal express route to respond to health checks (`WorkerHealthService`) - I think that is also worth keeping and not deprecating because it is self-contained and basically works exactly like the server healthcheck. Unless there is some better way we would recommend to monitor health for the worker?

dlhck:
Good call — removed the deprecation from `WorkerHealthService`, `WorkerHealthCheckConfig`, and `VendureWorker.startHealthCheckServer()`. It's self-contained and there's no better alternative to recommend for worker health monitoring. Pushed in a8b781d87.

packages/core/src/health-check/constants.ts
```diff
@@ -1 +1,5 @@
+/**
+ * @deprecated The built-in health check endpoint is deprecated and will be removed in v4.0.0.
+ * Use infrastructure-level health checks instead.
+ */
```
michaelbromley: I think this shouldn't be deprecated since we are keeping the simple health checks.

PR #4444, opened 2026-02-27
fix: Apply addcustomfields to profile page route loader

### LINKED ISSUE #4443: Dashboard: Adding custom fields to the Administrator entity causes the Dashboard UI profile page (/dashboard/profile) to break.
**To Reproduce**  
1. Add custom fields to the `Administrator` entity via `CustomFields['Administrator']` in your Vendure config.  
2. Restart the server and run any necessary migrations.  
3. Log in to the Admin UI.  
4. Navigate to `http://localhost:3000/dashboard/profile`.  
5. Observe that the page fails to load or throws an error.

Example custom field configuration:  
```ts
export const administratorCustomFields: CustomFields['Administrator'] = [
    {
        name: 'adminBonusAvailable',
        type: 'int',
        defaultValue: 0,
        public: false,
        readonly: true,
    },
];
```
**Screenshots/Videos**  

Image

- @vendure/core version: *3.5.4*   

**Configuration**  
Relevant snippet from `vendure-config.ts`:  
```ts
import { administratorCustomFields } from './custom-fields/administrator';

  // ...
  customFields: {
    Administrator: administratorCustomFields,
  },
  // ...
};
```

Ryrahul:
## Description

The profile page (`/dashboard/profile`) crashes when custom fields are defined on the `Administrator` entity. The route loader was querying `activeAdministratorDocument` without passing it through `addCustomFields()`, causing a GraphQL error — the server expects `customFields` to have sub-field selections, but the raw query sends it as a bare scalar field.

The administrator detail page (`/administrators/:id`) works correctly because it uses `detailPageRouteLoader()`, which already applies `addCustomFields()` before executing the query.

The fix wraps the query document with `addCustomFields()` in the profile page's route loader, matching the behavior of all other detail pages.

## Breaking changes

None.

## Checklist

📌 **Always:**

Fixes : #4443


biggamesmallworld (commented):
## Review

Thanks for the fix! The `addCustomFields` wrapping is the right approach — this definitely addresses the crash when custom fields are defined on `Administrator`. A couple of suggestions to bring this in line with the established patterns in the codebase:

### Suggestions

**1. Use `getDetailQueryOptions()` in the loader to fix query key mismatch**

The loader currently uses a hardcoded `queryKey: ['DetailPage', 'profile']`, but `useDetailPage` constructs its key as `['DetailPage', 'activeAdministrator', { id: 'undefined' }]`. Since these don't match, the loader prefetches into a cache bucket the component never reads — so the data gets fetched twice.

The `global-settings` page (`global-settings.tsx`) is the closest analog and handles this correctly. The profile loader should follow the same pattern:

```ts
const pageId = 'profile';

// in loader:
const { extendedQuery } = extendDetailFormQuery(
    addCustomFields(activeAdministratorDocument),
    pageId,
);
await context.queryClient.ensureQueryData(
    getDetailQueryOptions(extendedQuery, { id: 'undefined' }),
    {},
);
```

This also picks up `extendDetailFormQuery()`, which is needed since the component has `pageId="profile"` on its `<Page>` — without it, any extensions registered against the profile detail query won't be included in the loader's prefetch.

**2. Remove the unrelated `PageBlock` formatting change**

The multi-line → single-line reformatting of the `PageBlock` props is unrelated to the bug fix. Keeping the diff focused makes it easier to review and keeps `git blame` clean.

### Notes
- The `addCustomFields` call is correct — it uses `WeakMap` memoization so calling it in the loader has no performance concern.
- A regression test (e.g., a dashboard e2e test that configures an `Administrator` custom field and navigates to `/profile`) would be a nice addition to prevent this from coming back, but not a blocker.

Ryrahul:
> ## Review
> Thanks for the fix! The `addCustomFields` wrapping is the right approach — this definitely addresses the crash when custom fields are defined on `Administrator`. A couple of suggestions to bring this in line with the established patterns in the codebase:
> 
> ### Suggestions
> **1. Use `getDetailQueryOptions()` in the loader to fix query key mismatch**
> 
> The loader currently uses a hardcoded `queryKey: ['DetailPage', 'profile']`, but `useDetailPage` constructs its key as `['DetailPage', 'activeAdministrator', { id: 'undefined' }]`. Since these don't match, the loader prefetches into a cache bucket the component never reads — so the data gets fetched twice.
> 
> The `global-settings` page (`global-settings.tsx`) is the closest analog and handles this correctly. The profile loader should follow the same pattern:
> 
> ```ts
> const pageId = 'profile';
> 
> // in loader:
> const { extendedQuery } = extendDetailFormQuery(
>     addCustomFields(activeAdministratorDocument),
>     pageId,
> );
> await context.queryClient.ensureQueryData(
>     getDetailQueryOptions(extendedQuery, { id: 'undefined' }),
>     {},
> );
> ```
> 
> This also picks up `extendDetailFormQuery()`, which is needed since the component has `pageId="profile"` on its `<Page>` — without it, any extensions registered against the profile detail query won't be included in the loader's prefetch.
> 
> **2. Remove the unrelated `PageBlock` formatting change**
> 
> The multi-line → single-line reformatting of the `PageBlock` props is unrelated to the bug fix. Keeping the diff focused makes it easier to review and keeps `git blame` clean.
> 
> ### Notes
> * The `addCustomFields` call is correct — it uses `WeakMap` memoization so calling it in the loader has no performance concern.
> * A regression test (e.g., a dashboard e2e test that configures an `Administrator` custom field and navigates to `/profile`) would be a nice addition to prevent this from coming back, but not a blocker.

Thank you for the detailed review and helpful suggestions.

Regarding the `extendDetailFormQuery` + `getDetailQueryOptions` pattern and the query key mismatch it appears these were already part of the original implementation prior to this PR. My change mainly introduced the `addCustomFields` wrapper to address the crash. That said, your suggestion about aligning the loader with the pattern used in `global-settings.tsx` is absolutely valid, and I’ve updated the implementation so the loader now properly uses `extendDetailFormQuery()` and `getDetailQueryOptions()`.

The `PageBlock` formatting change was unintentional and most likely caused by Prettier normalizing the file during save.

Have also added the declaration off `pageId` 

Thanks again for the careful review and for pointing these out I really appreciate it.

biggamesmallworld:
Tested this locally and this is looking good to me! 

Screenshot 2026-03-31 at 12 13 09
Screenshot 2026-03-31 at 12 12 28

Approving 👍

biggamesmallworld (approved):
Looking good to me 👍

Ryrahul:
@michaelbromley @biggamesmallworld 
Thanks again for the review! If everything is good on your end, would it be possible to include this in the next release?

grolmus:
Closing in favor of #4734, which applies the same fix but additionally incorporates the review feedback from @biggamesmallworld and coderabbit-ai that wasn't addressed here:

- Uses the `pageId` constant in `<Page pageId={pageId}>` instead of hardcoded `"profile"` (so loader and component identifier stay locked together).
- Drops the unrelated `PageBlock` formatting change.

Thanks for the original work @Ryrahul — your approach was correct and the new PR follows the same pattern.

PR #4448, opened 2026-02-28
feat(core): Extend custom relational field config

LucidityDesign:
# Description

Adds properties like cascade, onDelete and onUpdate to custom fields of type `relation`

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


grolmus (commented):
Thanks for the PR! Exposing TypeORM relation options on custom fields is a useful feature. A few things to address before this can go in:

## Must address

### 1. `nullable` silently injected into relation decorators

The diff passes `nullable` (from the outer destructure on line 42) into both `ManyToMany` and `ManyToOne` options, but this wasn't part of the type change or PR description.

- **ManyToMany**: `nullable` is semantically meaningless here — ManyToMany uses a junction table, there's no nullable FK column on the entity itself. TypeORM accepts it silently but it does nothing. Should be removed from the ManyToMany call.
- **ManyToOne**: `nullable` _does_ make sense, but this is an undocumented behavioral change. Previously, `nullable: false` on a relation custom field was silently ignored. Now it gets enforced. This should either be called out explicitly or handled in a separate commit.

### 2. Auto-generated reference docs edited directly

`docs/docs/reference/typescript-api/custom-fields/typed-custom-single-field-config.mdx` is auto-generated and will be overwritten on the next `npm run docs:build`. This file shouldn't be edited manually — the line number change will happen automatically when the source types change.

### 3. Documentation for new properties

The new sections for `cascade`, `onDelete`, and `onUpdate` in the custom fields guide look good. Consider adding a note about the implications of cascade delete — e.g., that using `onDelete: 'CASCADE'` on a relation to a core entity could cascade-delete core data.

## Should address

### 4. No guardrails on cascade delete of core entities

With this change, a plugin author could write:
```ts
{
    name: 'myThing',
    type: 'relation',
    entity: Product,
    onDelete: 'CASCADE',
}
```
This would cascade-delete `Product` rows through a custom field relation. At minimum, a runtime `Logger.warn()` when cascade delete targets a built-in Vendure entity would help prevent accidents.

### 5. No tests

Adding options that affect database schema and cascade behavior should ideally have at least a basic test — e.g., verifying that `onDelete`/`cascade` are passed through to the TypeORM metadata correctly.

## Looks good

The `Pick<RelationOptions, 'cascade' | 'onDelete' | 'onUpdate' | 'eager'>` approach is clean and keeps the types in sync with TypeORM. The `eager` type didn't change, so this is backwards-compatible.

LucidityDesign:
Thanks for the comprehensive feedback!
1. `nullable`
   > This should either be called out explicitly or handled in a separate commit.

   Do you mean separate commit or PR?
2. Auto-generated reference docs edited directly
   The changes were generated by running `npm run docs:build`. Is this fine to leave in this PR?

I will implement the rest of the feedback asap.

packages/core/src/entity/register-custom-entity-fields.spec.ts
LucidityDesign, line 1: The tests are vibe coded. They seem to test the right behaviour and look good to me. But please double check and I would need some help if anything is missing in the test file.

LucidityDesign:
@grolmus can you have another look please?

grolmus (changes requested):
Thanks for the PR @LucidityDesign — exposing `cascade`/`onDelete`/`onUpdate` on relation custom fields is a genuinely useful addition, and modelling it via `Pick<RelationOptions, ...>` is the right call (keeps us in sync with TypeORM, and importing TypeORM types into config has precedent here). A few things I'd like to resolve before we merge:

### 🔴 Must fix

**1. `nullable` is now passed into `ManyToOne` — an out-of-scope, migration-generating behavior change**
In `register-custom-entity-fields.ts`, the `ManyToOne` branch previously received only `{ eager }`, so the FK was always nullable and a relation field's `nullable: false` was effectively ignored. It now forwards `nullable`, so any existing project with `nullable: false` on a single relation field will, on the next `migration:generate`, get an `ALTER ... SET NOT NULL` — which can fail on a DB that already contains NULLs. This is outside the stated scope (cascade/onDelete/onUpdate).

Since you already dropped `nullable` from the `ManyToMany` branch as "meaningless", I assume keeping it on `ManyToOne` is deliberate — if so that's defensible, but it still changes the generated schema for existing projects, so it needs an explicit changeset note + a test (`nullable: false → NOT NULL`) rather than riding along silently. Otherwise, please drop it to keep the PR scoped.

**2. `onDelete`/`onUpdate` docs/examples use `list: true` (many-to-many), where the semantics differ from the prose**
All three new property examples use a `list: true` relation. On a many-to-many, `onDelete`/`onUpdate` apply to the *junction table* FKs — not to the relationship described in the note ("deleting that `ProductVariant` can cascade to rows that reference it"), and TypeORM's emission of these on the join table is inconsistent across drivers. Could you switch the `onDelete`/`onUpdate` examples to a single (`list: false`) relation, and add a short note that for list relations these only affect the join-table FKs? Worth deciding whether they should be plumbed into the ManyToMany branch at all.

### 🟡 Should fix

**3. "Defaults to `RESTRICT`" in the docs isn't accurate**
When the option is omitted, `undefined` is passed and TypeORM emits no referential action, so the database default applies (`NO ACTION` on Postgres/MySQL/SQLite) — not `RESTRICT`. Could you reword to "if not set, the database default (`NO ACTION`) applies", and add a note that `onUpdate` is a no-op on SQLite?

**4. The CASCADE-on-core-entity warning has no test coverage**
This is the safety rail that justifies allowing `CASCADE` on core entities, but nothing tests it. A small `vi.spyOn(Logger, 'warn')` test asserting it fires once for a core entity + `CASCADE`, and *not* for a non-core entity or for `SET NULL`, would lock it in.

**5. Target branch + changeset**
This extends the public `RelationCustomFieldConfig` API, so it's a feature — it should target `minor` rather than `master`, and needs a changeset.

**6. The spec mutates global `getMetadataArgsStorage()`**
It decorates the real `Asset` entity and cleans up best-effort via `@ts-ignore` in `beforeEach` only. Could you add an `afterEach` (so the last test cleans up too), and ideally use a throwaway entity instead of `Asset`? Also `toEqual` over `toMatchObject` where the option shape is fully known, plus a "no options set" case.

### 🟢 Nits (non-blocking)
- The core-entity check is name-based (`relatedEntityName in coreEntitiesMap`), so a user entity named e.g. `Order` would false-positive — a one-line comment noting it's best-effort would help.
- The warning covers DB-level `onDelete: 'CASCADE'` but not ORM-level `cascade: true`, which is arguably the more common foot-gun — worth a doc line or a follow-up.

Happy to help with any of these — thanks again for the contribution! 🙏

LucidityDesign:
🔴 Must fix

1. I have removed passing `nullable` to avoid breaking changes. I will create a new PR for that (edit: see https://github.com/vendurehq/vendure/pull/4895)

2. onDelete/onUpdate docs/examples fixed

🟡 Should fix

3. "Defaults to RESTRICT" fixed

4. The CASCADE-on-core-entity warning has test coverage

5. Target branch has been changed. Changelog will be automatically generated, if I'm not mistaken

6. The spec mutates global getMetadataArgsStorage() => Fixed

🟢 Nits (non-blocking)
The core-entity check is name-based (relatedEntityName in coreEntitiesMap), so a user entity named e.g. Order would false-positive — a one-line comment noting it's best-effort would help.
This is already in the documentation:
> Setting `onDelete` to `CASCADE` or `cascade: ['remove']` means that if the referenced entity is deleted, rows owning this relation may also be deleted.
This can have unintended consequences, especially when referencing core Vendure entities.
For example, if you have a relation to `ProductVariant` and set `onDelete: 'CASCADE'`, deleting that `ProductVariant` can cascade to rows that reference it.
Use with caution.

The warning covers DB-level onDelete: 'CASCADE' but not ORM-level cascade: true, which is arguably the more common foot-gun — worth a doc line or a follow-up.
=> Done

grolmus (changes requested):
Thanks for the thorough turnaround @LucidityDesign — `nullable` revert, the `list: false` examples, the doc wording, the target branch, and the spec cleanup (afterEach + throwaway entity + `toEqual`) all look good. One real bug left, then I think we're there.

### 🔴 Cascade warning fires for non-core entities (operator precedence)

In `register-custom-entity-fields.ts`:

```ts
if (
    cascade === true ||
    (Array.isArray(cascade) && (cascade.includes('remove') || cascade.includes('soft-remove'))) &&
    relatedEntityName in coreEntitiesMap
) {
```

`&&` binds tighter than `||`, so this parses as `cascade === true || (<array check> && <core check>)`. The `relatedEntityName in coreEntitiesMap` guard only applies to the array branch — so `cascade: true` warns for **any** entity, including a user's own non-core entity:

```
cascade: true,      non-core  => warns  (should not)
cascade: ['remove'], non-core => silent (correct)
cascade: true,      core      => warns  (correct)
```

The current test doesn't catch this because the `cascade: true` case targets `Asset` (a core entity), where the result is the same either way. Fix by parenthesizing the detection:

```ts
if (
    (cascade === true ||
        (Array.isArray(cascade) && (cascade.includes('remove') || cascade.includes('soft-remove')))) &&
    relatedEntityName in coreEntitiesMap
) {
```

Please also add a negative test (`cascade: true` on a non-core entity → `Logger.warn` not called) so it can't regress.

### 🟡 Non-blocking
Both warnings also fire for `list: true` (many-to-many) relations, but there `onDelete`/`cascade` apply to the junction-table FKs rather than the related entity, so the "deleting a `X` row can delete owning rows" wording doesn't quite hold. Consider gating the warnings to the `ManyToOne` branch.

Everything else is addressed — just the precedence fix + test and I'm happy to approve. Thanks again! 🙏

LucidityDesign:
Good catch!! I fixed the error and added test cases

michaelbromley:
Hi @LucidityDesign — thanks for this, it's a great addition. I'm pushing a couple of small follow-up commits directly to your branch (you have "allow edits from maintainers" enabled) as I'm tight on time to get the final changes merged for v3.7:

1. **`fix(core)`** — tightened the cascade warning so that when `cascade: true` is set, the message now reads `cascade: true (which includes 'remove' and 'soft-remove')` rather than quoting the array form the user didn't write. Updated the corresponding test, which was previously asserting the old (misleading) message, and added negative coverage for `cascade: ['insert']` and `onDelete: 'SET NULL'` on core entities (should not warn) plus a mixed-array case that should.
2. **`docs`** — removed the note claiming `onUpdate` is unsupported/ignored on SQLite. SQLite does support `ON UPDATE` FK actions and TypeORM's SQLite driver emits them, so the implementation passing `onUpdate` through unconditionally is correct — the doc note was the inaccurate part.

michaelbromley:
Hi @LucidityDesign, I've been digging into this while reviewing for the v3.7 release, and before we merge I'd like to understand the concrete use case you're solving, as there's no linked feature request or use-case description on the PR.

The reason I'm asking: I dug into how these options actually behave, and they don't quite do what they appear to. Let me use a concrete example — a Product with a custom field relation pointing to an Asset:

```ts
customFields: {
  Product: [
    { name: 'spec', type: 'relation', entity: Asset, onDelete: 'CASCADE' },
  ],
}
```

A few things I found:

1. "Delete the Product, and its linked Asset goes too" isn't possible with these options.
You might expect cascade: ['remove'] to mean "when I delete the Product, also delete the linked Asset." But TypeORM doesn't actually act on that — it has no effect on deletion, so nothing happens to the Asset.

2. onDelete: 'CASCADE' does the opposite of what it looks like.
Because the Product row is the one that holds the link to the Asset, onDelete: 'CASCADE' means: delete the Asset → the Product gets deleted. So removing an Asset would wipe out every Product that links to it.

3. …and even that doesn't really work on a core entity like Product.
Vendure doesn't truly delete Products — it marks them as deleted (soft-delete), because Orders, ProductVariants etc. still reference them. So when onDelete: 'CASCADE' tries to hard-delete a Product, the database blocks it (the Product still has variants attached) and you get a foreign-key error instead.

So against a core entity, the destructive options (onDelete: 'CASCADE', hard cascade: ['remove']) either do nothing or throw. The ones that are genuinely useful here are eager (auto-load the relation, already exists), onDelete: 'SET NULL' (if the Asset is deleted, just clear the link instead of erroring), and cascade: ['insert', 'update'] (save a linked custom entity along with the Product).

Could you tell me a bit about what you were trying to achieve? For example: which entity has the custom field, what it links to, and what you want to happen when one side is deleted or updated. That'll help us make sure the final feature does what you need, rather than shipping options that quietly don't behave as expected.

LucidityDesign:
Hey @michaelbromley,
my use case is not about core entities. I'm having two custom entities that live in different plugins. Due to separation of concern/functionality I would prefer not to move them into the same plugin.
One entity (EntityA) adds a relation to the other (EntityB) via a custom field and I still want to make sure EntityA is deleted when the related EntityB is deleted.

I can achieve the same with listening for a delete event and then delete the related entities, but a onDelete/cascade functionality would be cleaner IMO.

michaelbromley (approved):

michaelbromley:
@LucidityDesign thanks for your contribution 👍

PR #4452, opened 2026-03-03
fix(asset-server-plugin): Set correct Content-Type for PDF assets

### LINKED ISSUE #4403: Content-Type not set correctly for .pdf assets (and assets uploaded to S3)

biggamesmallworld:
# Description

PDF files served by the asset server were returning `application/octet-stream` instead of `application/pdf`, causing Chromium-based browsers to download them instead of displaying inline.

Added `.pdf` to the `getMimeType()` extension-based lookup in `AssetServer` so PDFs resolve correctly without relying on async `file-type` detection.

Fixes #4403

# Breaking changes

None.

# Screenshots
https://github.com/user-attachments/assets/d0e43311-14f0-4afb-8e2f-32f9a9c4e2b0

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
+++ b/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
@@ -279,7 +279,7 @@ describe('AssetServerPlugin', () => {
         }
 
         beforeAll(async () => {
-            const formats = ['gif', 'jpg', 'png', 'svg', 'tiff', 'webp'];
+            const formats = ['gif', 'jpg', 'png', 'svg', 'tiff', 'webp', 'pdf'];
 
             const filesToUpload = formats.map(ext => path.join(__dirname, `fixtures/assets/test.${ext}`));
             const { createAssets }: CreateAssetsMutation = await adminClient.fileUploadMutation({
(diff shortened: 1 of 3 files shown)


biggamesmallworld:
Closing in favor of #4404 which also fixes the missing ContentType on S3 uploads.

PR #4453, opened 2026-03-03
fix(dashboard): Preserve dirty state when form values update

### LINKED ISSUE #4389: Dashboard: After 3.5.4 some collections are not noticed as changed when editing its information
**Describe the bug**
After upgrading 3.5.3 -> 3.5.4 some collections are not detected as changed when changing for example name or description of a collection so the update button is not getting activated.  See video.

**To Reproduce**
Not sure exactly how to reproduce this to be honest because it's not all collections this happens on. See my video and reach out to me if you need more info.

**Expected behavior**
Changes are detected and update button is activated.

**Actual behavior**
changes not detected when updating information on a collection.

**Screenshots/Videos**
https://youtu.be/5BE899wbZe0

- Nodejs version: 24
- Database (mysql/postgres etc): PG
- Browser (if applicable):
- Package manager (npm/yarn/pnpm): Yarn

**Minimal reproduction**

**Workaround**
Toggle the "Contains any" in the product variant filters, only then the update button becomes active.

**Additional context**

- When did this issue start occurring? After upgrading 3.5.3 -> 3.5.4.
- Does it happen consistently or intermittently? On some collections, not all and haven't found any pattern.
- Any recent changes to your setup? No.

grolmus:
## Summary
- Adds `resetOptions.keepDirtyValues: true` to the generated form hook
- Preserves user edits when the form's external `values` prop updates due to re-renders

## Root Cause
When filter validity state changes (added in 3.5.4), the component re-renders. On each render, the `values` prop passed to react-hook-form is recreated. Without `keepDirtyValues`, this can reset the form's dirty state, causing the Update button to remain disabled even after the user has edited fields like name or description.

## Solution
Use react-hook-form's built-in `resetOptions.keepDirtyValues` to preserve user edits when external values update. This is the intended API for exactly this use case.

## Test plan
- [ ] Open a collection that has filters (e.g., "Filter by product variant name")
- [ ] Edit the collection name or description
- [ ] Verify the Update button becomes enabled
- [ ] Save and verify changes persist

Fixes #4389
Fixes OSS-386

### DIFF
--- a/packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx
+++ b/packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx
@@ -122,6 +122,12 @@ export function useGeneratedForm<
         mode: 'onChange',
         defaultValues: processedDefaultValues,
         values,
+        // Preserve user edits when external values update (e.g., due to re-renders
+        // from filter validity state changes). Without this, the form's dirty state
+        // could be incorrectly reset when the component re-renders.
+        resetOptions: {
+            keepDirtyValues: true,
+        },
     });
     let submitHandler = (event: FormEvent): any => {
         event.preventDefault();


michaelbromley:
@grolmus did you manually test and verify this fix in the browser?

grolmus:
> @grolmus did you manually test and verify this fix in the browser?

yes i am always everything testing in browsers.

michaelbromley:
I am still seeing the issue when testing. There might be some cases where it happens and some not, as was reported in the original issue.

michaelbromley:
## Additional investigation & revised fix

Thanks for the PR! We dug deeper into this and discovered the root cause is actually different from what was initially diagnosed. Here's what we found:

### The `keepDirtyValues` fix is unnecessary

We instrumented `useGeneratedForm` with debug logging to track whether the `values` object changes structurally between renders. Result: **zero structural diffs across 5 re-renders**. react-hook-form v7.70.0 already uses `deepEqual` internally to compare the `values` prop (line 2855 of the RHF source), so it correctly avoids resetting the form when the object reference changes but the data is identical.

The form's dirty state was never being lost — the Update button was disabled by a *different* condition.

### The actual root cause

The collection detail page's Update button is gated by four conditions:

```tsx
disabled={
    !form.formState.isDirty ||
    !form.formState.isValid ||
    isPending ||
    !filtersArgsValid   // <-- this was the problem
}
```

The `filtersArgsValid` flag comes from `ConfigurableOperationInput`'s validity check, which validates that all `required` args have values. The `combineWithAnd` arg (added in 3.5.4) is `required` by default and has `defaultValue: true`, but legacy collections created before 3.5.4 don't have this arg in their stored filter data. The validity check was treating it as permanently invalid, keeping `filtersArgsValid = false` regardless of any form edits.

### What we changed

1. **Removed `keepDirtyValues: true`** from `useGeneratedForm` — it has global blast radius (affects every detail page) and isn't needed
2. **Added a `defaultValue` check** in `ConfigurableOperationInput`'s validity `useEffect` — args with a `defaultValue` are now considered valid even when absent from stored data
3. **Added 3 e2e regression tests** that create a collection with a deliberately incomplete filter (omitting `combineWithAnd`) and verify name editing, description editing, and save persistence
4. **Extracted `VendureAdminClient`** to a shared e2e utility to avoid duplication with orders tests

We verified:
- All 204 dashboard e2e tests pass with only the targeted fix (no `keepDirtyValues`)
- Tests correctly fail when the fix is reverted (sanity check)
- The values are structurally identical between re-renders (confirming RHF's `deepEqual` handles it)

michaelbromley:
I am closing this in favour of 

- #4462 

After a thorough investigation it turned out that the true fix was a bit different and more targeted than the one presented here.

PR #4468, opened 2026-03-05
fix(core): Resolve customer user via relation instead of email lookup

### LINKED ISSUE #4026: Verified customer wrong
**Describe the bug**
Consider the following two scenarios:

1. User is making a Guest-Order and becomes a customer with email _a@b.com_
  - No User entity is created for guest orders

2. Another user is now registering with _a@b.com_ and gets verified by an external authority
  - A second customer is correctly created
  - A User entity is created, associated with the newly created customer and email verified

It's correct that two Customer entities exist in parallel as the first could have been a typo in the guest order, but the first customer should not be marked as verified in the Admin Dashboard. 
The verification check is based on email rather than the direct user-customer relationship

@michaelbromley The implementation appears intentional, but the reasoning behind it is unclear to me in `customer-entity.resolver.ts`:

```js
@ResolveField()
user(@Ctx() ctx: RequestContext, @Parent() customer: Customer) {
    if (customer.user) {
        return customer.user;
    }

    return this.userService.getUserByEmailAddress(ctx, customer.emailAddress, 'customer');
}
```

Why not just `return customer.user;` and if undefined the customer is not verified?

grolmus:
## Summary

- Fix the `user` resolver on `CustomerEntityResolver` which incorrectly fell back to `getUserByEmailAddress()` — this caused guest customers to appear "verified" when a different registered customer shared the same email address
- Replace the email-based lookup with `customerService.findOne()` which loads the actual user relation via the foreign key
- Remove unused `UserService` dependency from the resolver

Fixes #4026

## Test plan

- [x] Added e2e regression test: guest customer with same email as registered customer should not show as verified
- [x] All 47 customer e2e tests pass
- [x] All 48 shop-auth e2e tests pass


gabriellbui (commented):
## Review

The fix is conceptually correct — replacing email-based user lookup with relation-based lookup properly isolates the guest customer's `user` field from a registered customer sharing the same email. However there's a blocking issue in the implementation.

### 🔴 MUST CHANGE

**`packages/core/src/api/resolvers/entity/customer-entity.resolver.ts:52` — Missing `['user']` relation**

```typescript
// Current
const loaded = await this.customerService.findOne(ctx, customer.id);
return loaded?.user ?? null;

// Fix
const loaded = await this.customerService.findOne(ctx, customer.id, ['user']);
return loaded?.user ?? null;
```

`findOne` signature is `findOne(ctx, id, relations: RelationPaths<Customer> = [])`. Without explicitly passing `['user']`, the relation is never loaded and `loaded?.user` is always `undefined`. The fallback path would return `null` for **all** customers whose `user` isn't eagerly pre-loaded on the parent entity — including registered customers. This could be a silent regression worse than the original bug. (CodeRabbit flagged this same issue.)

---

### 🟡 SHOULD CHANGE

**`packages/core/e2e/customer.e2e-spec.ts:718` — Test conditional logic allows the bug scenario to go untested**

```typescript
const guestCustomerId = (setCustomerForOrder as any).customer?.id;
if (guestCustomerId && guestCustomerId !== firstCustomer.id) {
    // actual assertion
} else {
    // fallback that doesn't test the bug
}
```

If `guestCustomerId` is `undefined` (e.g. mutation returned an error type) or equals `firstCustomer.id`, the test silently passes without exercising the regression scenario. The `as any` cast should use the typed `SetCustomerForOrderMutation` shape, and the test should be structured to guarantee the guest customer is created and assert unconditionally.

---

### 🔵 CONSIDER

**`packages/core/e2e/customer.e2e-spec.ts:706` — Validate mutation result before use**

Add a guard like `expect(setCustomerForOrder.__typename).toBe('Order')` before accessing `.customer` so an unexpected error type doesn't silently short-circuit the test through the else branch.

---

Once the `['user']` relation arg is added, this is a clean and well-scoped fix for a real bug. Good find on the root cause.

michaelbromley (changes requested):
Gabirel and CodeRabbit are spot on. This PR is dangerous because it introduces a new bug and has a test that does not test what it should be.

grolmus:
## Addressed review feedback

Thanks @gabriellbui and @michaelbromley for the thorough review. Both issues are now fixed:

### 1. Missing `['user']` relation (blocking)

`customerService.findOne(ctx, customer.id)` was called without requesting the `user` relation, so `loaded?.user` was always `undefined`. Fixed by passing `['user']` as the third argument:

```typescript
const loaded = await this.customerService.findOne(ctx, customer.id, ['user']);
return loaded?.user ?? null;
```

### 2. Regression test rewrite

The original test was placed in `customer.e2e-spec.ts`, but the default `GuestCheckoutStrategy` rejects guest checkout when a registered customer shares the same email (`EMAIL_ADDRESS_CONFLICT_ERROR`). This meant the test could never actually exercise the bug scenario.

**Moved the test to `guest-checkout-strategy.e2e-spec.ts`** where `TestGuestCheckoutStrategy` with `createNewCustomerOnEmailAddressConflict = true` can create the exact conflict scenario: a separate guest Customer entity sharing an email with a registered Customer.

The rewritten test:
- Uses `orderResultGuard.assertSuccess()` instead of `as any` casts
- Asserts unconditionally (no if/else branches that silently skip)
- Queries `activeOrder.customer.user` via the shop API to verify the resolver returns `null` for the guest customer

Also renamed the misleading `describe('Order taxes')` block to `describe('GuestCheckoutStrategy')`.

### Test results
- 46/46 customer e2e tests pass
- 7/7 guest-checkout-strategy e2e tests pass
- 48/48 shop-auth e2e tests pass

packages/core/e2e/guest-checkout-strategy.e2e-spec.ts
```diff
@@ -226,6 +226,47 @@ describe('Order taxes', () => {
...
+                            id
+                            verified
+                        }
+                    }
+                }
+            }
+        `);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        expect(activeOrder?.customer?.id).toBe(setCustomerForOrder.customer!.id);
+        expect(activeOrder?.customer?.user).toBeNull();
```
gabriellbui, line 267: I think this test passes because the customer can't be found at all (missing channel), not because it was found and has no user relation. I think the actual fix logic is not tested. 

The issue is at `packages/core/e2e/guest-checkout-strategy.e2e-spec.ts:70-74`:

```
const newCustomer = await this.connection
    .getRepository(ctx, Customer)
    .save(new Customer(input));
await this.channelService.assignToCurrentChannel(newCustomer, ctx);
return newCustomer;  // <-- channels never persisted
```

grolmus:
@gabriellbui Good catch — `assignToCurrentChannel` only modifies the entity in memory, it doesn't persist. The `save()` was called *before* `assignToCurrentChannel`, so the channel assignment was never written to the DB.

Fixed in 0ae06cc: moved `assignToCurrentChannel` before `save()`, matching the pattern used in `CustomerService.create()` (line 273-274 of customer.service.ts). The guest customer is now properly persisted with its channel, so `findOne` actually locates it and the test validates the resolver logic correctly.

packages/core/e2e/guest-checkout-strategy.e2e-spec.ts
grolmus: done

gabriellbui (approved):

michaelbromley (approved):

PR #4487, opened 2026-03-10
fix(core): Remove stale shipping lines when shipping method no longer exists

### LINKED ISSUE #4486: Stale shipping line causing error during order recalculation
**Describe the bug**

`OrderCalculator.applyShipping()` returns early when a `ShippingLine` references a `ShippingMethod` that no longer exists (soft-deleted or removed from channel), leaving the stale shipping line on the order. This causes a GraphQL error when resolving the order:

```
Cannot return null for non-nullable field ShippingLine.shippingMethod
```

**To Reproduce**

1. Create an order and set a shipping method
2. Delete that shipping method (or remove it from the channel) via the Admin UI
3. As the customer with the active order, perform any action that triggers order recalculation (e.g. add an item to cart)
4. GraphQL returns the error above

**Expected behavior**

The stale shipping line should be removed from the order during recalculation (the same way a shipping line is removed when its method is no longer eligible and no replacement is found).

**Actual behavior**

The `applyShipping()` method returns early when `shippingMethodService.findOne()` returns `undefined`, skipping the rest of shipping recalculation entirely. The orphaned shipping line remains on the order, and the `ShippingLineEntityResolver` cannot resolve the non-nullable `shippingMethod` field.

The bug is in `packages/core/src/service/helpers/order-calculator/order-calculator.ts`:

```typescript
if (!currentShippingMethod) {
    return; // <-- exits entire method instead of removing the shipping line
}
```

Compare with the handling ~20 lines later when a method exists but is no longer eligible and no replacement is found:

```typescript
} else {
    order.shippingLines = order.shippingLines.filter(sl => sl !== shippingLine);
}
```

--

I'll be submitting a PR to fix this.

- @vendure/core version: 3.5.5
- Nodejs version: 22
- Database: PostgreSQL

**Workaround**

Directly delete the orphaned shipping lines from the database.

colinpieper:
# Description

In OrderCalculator.applyShipping(), when a ShippingLine references a ShippingMethod that has been soft-deleted or removed from the channel, the code was returning early from the entire method instead of removing the stale shipping line. This left an orphaned ShippingLine on the order, causing "Cannot return null for non-nullable field ShippingLine.shippingMethod" errors on subsequent GraphQL queries.

Fixes #4486

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
+++ b/packages/core/src/service/helpers/order-calculator/order-calculator.spec.ts
@@ -188,6 +188,74 @@ describe('OrderCalculator', () => {
             expect(order.totalWithTax).toBe(order.subTotalWithTax + 1200);
             assertOrderTotalsAddUp(order);
         });
+
+        it('removes shipping line when shipping method no longer exists', async () => {
+            const deletedShippingMethodId = 'T_deleted';
+            const module = await Test.createTestingModule({
+                providers: [
+                    OrderCalculator,
+                    RequestContextCacheService,
+                    { provide: TaxRateService, useClass: MockTaxRateService },
+                    { provide: ShippingCalculator, useValue: { getEligibleShippingMethods: () => [] } },
+                    {
+                        provide: ShippingMethodService,
+                        useValue: {
+                            findOne: (requestCtx: RequestContext, id: string) => {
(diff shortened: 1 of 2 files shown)


michaelbromley:
Thanks for the fix!

Issue #716, opened 2021-02-16
Admin error: can't load order with deleted shipping method


Clever-Up:
**Describe the bug**
I had a few orders already placed with a test shipping method, which I then deleted. After deleting it, an error message appears when accessing the admin panel (top right corner) saying: `cannot return null for non-nullable field ShippingLine.shippingMethod.`
And it doesn't show orders anymore (empty table). That might be because it cannot find their shipping method.

**To Reproduce**
1. Create a shipping method
2. Place one or more orders choosing it as a shipping method
3. Delete the shipping method
4. Try to access the admin console again and see the error

**Expected behavior**
It should be still possible to load order list even if using a shipping method that's no more available.

**Environment (please complete the following information):**
 - @vendure/core version: 0.18.3
 - Database (mysql/postgres etc): postgres

PR #4494, opened 2026-03-11
fix(core): Unassigning shipping lines from a channel breaks active orders

### LINKED ISSUE #4492: Deleting or unassigning shipping methods from a channel will break active orders
**Describe the bug**
Unassigning a shipping method from a channel will break currently active orders and requires manual intervention by removing shipping lines from the database.

When deleting a shipping method, the shipping method will remain on the active order even though it was just deleted.

**To Reproduce**

Deletion:

1. Create a shipping method
2. Create an order
3. Set the shipping method on the order
4. Delete shipping method
5. View the order detail page of just created order
6. See the shipping method is still there

Unassigning from channel:

1. Create a shipping method
2. Optionally create a channel if no other channels exist besides the default channel
3. Create an order on the channel
4. Set the shipping method on the order
5. Unassign the order from the channel
6. Navigate to orders
7. Click the order you've created with the deleted shipping method
8. See error

**Expected behavior**
Shipping line(s) using deleted shipping method id is removed from all active orders.

**Actual behavior**
Active orders with unassigned shipping methods break and cannot be mutated without manual intervention, and deleted shipping methods will remain on active orders

**Screenshots/Videos**

Unassigning from a channel:

https://github.com/user-attachments/assets/5b6e2c10-ad89-4b37-abb6-44fa8d804cd0

Deletion:

https://github.com/user-attachments/assets/1a8c65fe-5b98-45b3-a078-e379a0fc834a

- Nodejs version: 24.11.1

I am planning on submitting a PR to address this bug.

kwerie:
# Description

When unassigning a shipping method from a channel it breaks currently active orders because a reference on the shipping method still remains on the order, but the resolver cannot find the shipping method because it is removed from the channel and the schema has ShippingMethod declared as non-nullable.

# Breaking changes

No

# Screenshots

See #4492

# Checklist

📌 Always:

👍 Most of the time:

Fixes #4492, relates to #4486


kwerie:
@coderabbitai review

michaelbromley (dismissed):
Thanks for tackling this — the bug is real and the fix direction makes sense. Cleaning up shipping lines proactively on unassign is the right approach over trying to handle nulls in the resolver.

A couple of things to address though, one of which is a blocker:

**1. Order totals go stale after removing the shipping line (blocker)**

When you remove the `ShippingLine` rows, the parent Order's `shipping` and `shippingWithTax` columns still hold the old values — they're persisted `@Money()` columns, not derived on-the-fly. So after this runs, the customer sees no shipping lines but the order total still includes the shipping cost. It won't self-correct until the next `applyPriceAdjustments()` (e.g. adding another item to the cart).

You'll need to either zero out `shipping`/`shippingWithTax` on affected orders and save them, or trigger a proper order recalculation after the removal. Given that pulling in `OrderService` here would create a circular dependency, the simplest option might be to reset those columns directly and save the order.

**2. Test should assert on order totals**

The test checks that `shippingLines` is empty, which is good — but it should also verify that `shipping`, `total`, and `totalWithTax` are correct after the operation. That would have caught the stale totals issue.

See inline comments for the specifics.

michaelbromley:
Had a closer look at this and there's a broader issue to consider beyond active orders.

## The non-active order problem

The `ShippingLineEntityResolver` resolves `shippingMethod` by calling `findOneInChannel()` scoped to `ctx.channelId`. This applies to **all** orders — active, placed, completed, fulfilled, etc. So if you unassign a shipping method from a channel, viewing any historical order that used that method will also blow up, because `shippingMethod` is declared as `ShippingMethod!` (non-nullable) in the schema.

This is arguably worse than the active order case, since historical orders are read-only and there's nothing the customer or admin can do to "fix" them.

## Two separate fixes needed

### 1. Active orders — use `OrderCalculator.applyPriceAdjustments()`

The current approach of removing shipping lines is on the right track, but the order totals need to be recalculated afterwards. Rather than manually resetting `shipping`/`shippingWithTax` columns, the `OrderCalculator.applyPriceAdjustments()` method should be used — its internal `applyShipping()` logic already handles removing invalid/ineligible shipping lines and recalculating totals properly. Rolling your own recalculation here would duplicate that logic and bypass any custom strategies people might have configured.

### 2. Non-active orders — fix at the resolver/query level

For placed/completed orders, we should NOT be modifying shipping lines — that's historical data. The fix here is that the `ShippingLineEntityResolver` (or the underlying `findOne` call) should be able to resolve shipping methods that have been unassigned from the channel when the context is a historical order. Same idea as passing `includeDeleted: true` for soft-deleted methods — channel assignment shouldn't prevent resolving a shipping method that's referenced by an existing order.

kwerie:
@michaelbromley Thanks for the detailed feedback! Good call on the historical orders angle, I hadn't considered that. I'll rework the approach along those lines. 😄

kwerie:
@coderabbitai review

kwerie:
@coderabbitai review

kwerie:
I've added an optional channel param to `RequestContext.copy()`. This way `applyPriceAdjustments` calculates for the right channel while keeping the transaction from `event.ctx`. Happy to change the approach if you'd rather not touch `copy()`.

kwerie:
@coderabbitai review

kwerie:
@coderabbitai review

kwerie:
Hi @michaelbromley, gentle ping on this one. Could you take a look when you have a moment? :)

michaelbromley:
@kwerie Hi, thanks for making the changes. I’ve got this on the radar for the team to review at the next opportunity. Thanks for your patience.

grolmus:
Took another pass now that this has been reworked (leaving @michaelbromley's review as the formal gate). Big improvement — both of Michael's original blockers are genuinely addressed: totals are now recalculated and persisted via `applyPriceAdjustments`, the tests assert `shipping`/`total`/`totalWithTax`, and routing the cleanup through a `ChangeChannelEvent` blocking handler neatly avoids the OrderService↔ShippingMethodService circular dependency. I verified the handler runs inside the removal transaction, so a throw rolls the unassignment back. Nice.

A few things still to address before it's mergeable:

## Blockers

**1. Register the handler in `onApplicationBootstrap`, not the constructor.**
There's no precedent in core for calling `registerBlockingEventHandler` from a service constructor — every lifecycle registration uses `onApplicationBootstrap`/`onModuleInit`. Constructors are for DI wiring, not side effects on shared infrastructure, and `registerBlockingEventHandler` throws on a duplicate `id`, which makes test harnesses that re-instantiate `OrderService` fragile. Implement `OnApplicationBootstrap` on `OrderService` and register there (keep the explicit id).

**2. `filterOnChannel=false` is applied to *all* shipping-line → shippingMethod resolutions — scope it.**
In `shipping-line-entity.resolver.ts` the change drops channel scoping for every order, not just historical ones. In a multi-channel setup this lets an operator on channel A resolve a method that was never assigned to their channel via any order referencing it — a cross-channel visibility broadening beyond the goal. You already have the `shippingLine.shippingMethodId`, so resolve the method that's on *this order* specifically (include unassigned/soft-deleted), or at minimum gate the unscoped path behind the order being inactive/completed. Please don't ship the blanket version.

**3. Outright deletion isn't covered — the bug is only half-fixed.**
`ShippingMethodService.softDelete` publishes `ShippingMethodEvent(..., 'deleted')`, not `ChangeChannelEvent`, so the handler never fires and active orders depending on a *deleted* method stay in the broken state. Either also handle the `'deleted'` event through the same recalc path, or explicitly scope this PR/issue to channel-unassign and note that delete is tracked separately.

## Should-fix

**S1. `ctx.copy(channel)` swaps `_channel` but not `_currencyCode`/`_languageCode`.**
Those are separate backing fields set at construction — after `copy(channel)` they still hold the *original* ctx's currency/language. `applyPriceAdjustments` branches on `ctx.currencyCode`, so if the removal is driven from a different-currency channel than the one being processed, line prices can be recalculated under the wrong currency. Build the ctx for the target channel properly (or also set currency/language), and add a test with a non-default-currency channel.

**S2. Serial recalc inside the transaction.** All affected orders across all channels are recalculated serially while holding write locks, inside the admin's request transaction. Fine for correctness/atomicity, but for a bulk unassign touching many active orders it'll hold locks and trip the blocking-handler slow warning. Worth a comment noting the limitation (a post-commit job queue would scale better but trades away atomicity).

## Tests to add
- Multiple active orders in one channel — assert *all* are recalculated, not just the first.
- Multi-channel: method on A+B, unassigned from A only → A recalculated, B untouched (also covers S1 if B has a different currency).
- An order with an active promotion/surcharge → totals correct after removal.
- Delete-vs-unassign behavior (documents the B3 decision).
- Rollback: force a mid-handler failure and assert the unassignment itself rolls back.

Note: I reviewed statically + architecturally and didn't run the e2e suite locally; the points above stand regardless. Really solid progress though — fix the registration placement, scope the channel-filter change, and settle the delete case and this is close. Thanks @kwerie!

grolmus:
Re-ran CI — the unit-test blip was flake (green now), but there's one real e2e regression, and it's my fault for pointing you at the delete case in B3.

`order.e2e-spec.ts > "get an Order with a deleted ShippingMethod"` (#716) creates an **active** order, deletes the shipping method, and expects the order to **keep** its shipping line (deleted methods stay resolvable):

```
AssertionError: expected undefined to deeply equal { priceWithTax: 500, shippingMethod: { code: "royal-mail", id: "T_4", name: "Royal Mail" } }
```

Your new `ShippingMethodEvent 'deleted'` handler finds that active order and removes the line, so the test now gets `undefined`.

The key distinction:
- **Unassigning** a method from a channel genuinely invalidates it for that order — your `ChangeChannelEvent 'removed'` handler + recalc is exactly right, and that's the actual OSS-445 bug.
- A **soft-delete** doesn't invalidate it — the method still resolves (your scoped-fallback in `shipping-line-entity.resolver.ts` already handles that), the line stays valid, and totals are unaffected. #716 is the intended contract there.

Could you **drop the `'deleted'` handler and its accompanying test**? That removes the regression, and the unassign path stays fully fixed. Sorry for the detour on the delete case — that was my B3 note being over-cautious. Everything else (moving registration to `onApplicationBootstrap`, the scoped resolver fallback, setting currency/language on `copy(channel)`, and the added multi-order / multi-channel / rollback tests) looks great.

kwerie:
Hey @grolmus, thanks for the thorough review — and for the quick CI re-run. Agreed on the delete-vs-unassign distinction: unassigning genuinely invalidates the method, but a soft-delete doesn't, so the line should stay (the scoped resolver fallback already keeps it resolvable, per #716).

The commit I just pushed drops the `'deleted'` handler and rewrites that test to cover the unassign path instead, so nothing asserts line-removal-on-delete anymore and #716 is green again — while the unassign fix stays fully intact. No worries at all on the B3 detour; the distinction was worth spelling out.

grolmus (approved):
Re-reviewed after the latest round — this is in great shape now. All the blockers from both your and @michaelbromley's reviews are resolved, and CI is fully green (e2e across all four DBs, unit tests, CLA).

Rundown of what's addressed:
- **Stale totals (Michael's original blocker #1):** removing the shipping line now recalculates + persists via `applyPriceAdjustments`; tests assert `shipping`/`total`/`totalWithTax`.
- **Test coverage (Michael's #2):** added multi-order, multi-channel (unassign from one channel only, preserving the other's currency), promotion-adjusted totals, and a rollback test.
- **Lifecycle:** handler moved from the constructor to `onApplicationBootstrap`.
- **Channel-filter scoping:** `shipping-line-entity.resolver.ts` now tries the scoped lookup first and only falls back to unscoped when the method has left the channel — no blanket visibility broadening.
- **`copy(channel)`:** now sets language + currency from the target channel, so recalc runs under the correct channel.
- **Delete vs unassign:** the delete handler was correctly dropped — a soft-deleted method stays resolvable on existing orders (#716), so the fix is scoped to channel unassignment, which is the actual bug.

Approving from my side. @michaelbromley — your two blockers are fully addressed and CI is green; over to you to dismiss the earlier change request / merge when you're happy. Nice work @kwerie!

PR #4495, opened 2026-03-11
fix(dashboard): Use null instead of undefined when clearing relation custom fields

### LINKED ISSUE #4485: fix(dashboard): Clearing a relation custom field sends undefined instead of null
## Description

When clearing a single relation custom field via the `RelationSelector` component, `onChange(undefined)` is called. Since `undefined` is stripped from JSON serialization, the server never receives the instruction to clear the field — the relation remains set.

## Root cause

Two locations need fixing:

1. **`relation-selector.tsx`**: The `handleRemove` function calls `onChange(undefined)` when clearing a single-select relation. Should be `onChange(null)` so the value is included in the JSON payload.

2. **`form-engine/utils.ts`** (`transformRelationFields`): When a relation value is `null` (explicitly cleared), the current code does `relationValue?.id` which returns `undefined` — again stripped from JSON. Should pass `null` through explicitly for both single and list relations.

## Expected behavior

Clearing a relation custom field should send `null` to the server, which correctly unsets the relation.

## Steps to reproduce

1. Create a relation custom field on any entity (e.g., Order)
2. Set a value for the relation field
3. Clear the relation field using the "x" button
4. Save — the field remains set (the clear was not persisted)

Originally identified in #4427.

grolmus:
## Summary

- When clearing a single relation custom field via the `RelationSelector`, `onChange(undefined)` was called. Since `undefined` is stripped from JSON serialization, the server never received the clear instruction — the relation remained set.
- Changed `onChange(undefined)` to `onChange(null)` in `relation-selector.tsx` so the clear is included in the payload.
- Updated `transformRelationFields` in `form-engine/utils.ts` to pass `null` through explicitly for both single and list relations (previously `relationValue?.id` returned `undefined` for null values).
- Added dedicated test cases for null clearing behavior.

## Test plan

- [x] Existing `transformRelationFields` tests pass (22/22)
- [x] TypeScript compiles without errors in changed files
- [ ] Manual: Set a relation custom field, clear it, save — verify the field is actually cleared on the server

Fixes #4485

### DIFF
--- a/.github/workflows/back_merge.yml
+++ b/.github/workflows/back_merge.yml
@@ -4,7 +4,20 @@ on:
     push:
         branches:
             - master
-            - minor
+            # NOTE: The minor → major cascade is disabled because GitHub's
+            # "Automatically delete head branches" setting deletes the `minor`
+            # branch when a conflict-resolution PR (minor → major) is merged.
+            # There is no exclusion list for that setting, and rulesets' "Restrict
+            # deletions" rule does not reliably prevent it.
+            #
+            # Merge minor → major manually instead:
+            #   git checkout major && git pull && git merge origin/minor && git push
+            #
+            # See:
+            # - https://github.com/orgs/community/discussions/135012
+            # - https://github.com/orgs/community/discussions/152926
+            #
+            # - minor
     workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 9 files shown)


grolmus:
Addressed review feedback in cf9ffaed:

- Dropped `as string` cast in `SingleRelationInput` and `as string[]` in `MultiRelationInput` — both were unnecessary since `DashboardFormComponentProps.onChange` accepts `any`
- Added comments to `as any` casts in null-clearing tests explaining why they're needed

Re: CodeRabbit's concern about `edit-order-table.tsx` receiving `null` — that's a false positive. The `SingleRelationInput` there is used as a coupon code selector (add-only flow, no remove button), so `handleRemove` is never triggered in that context.

Ready for re-review.

PR #4497, opened 2026-03-11
fix(dashboard): Invalidate child collection query cache on remount

tbouliere-datasolution:
# Description

When navigating away from the Collections list page and returning via client-side (SPA) navigation, the expand/collapse toggle buttons on collection rows stop working. Clicking the folder icon to expand a collection's children has no effect. A full page reload is required to restore the functionality.

# To Reproduce

  1. Go to the Collections list page
  2. Expand one or more collections by clicking the folder icon (verify it works)
  3. Navigate to any other page via the sidebar (e.g., Products)
  4. Navigate back to the Collections list page
  5. Click the folder icon on any collection to expand it

# Fix

Invalidate childCollections queries on component mount so that cached data is refetched and the queryFn correctly populates the local state.
  
# Breaking changes

Maybe check the pagination system ? the useQuery seems to be there for this feature

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_collections/collections.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_collections/collections.tsx
@@ -67,7 +67,32 @@ function CollectionListPage() {
     const [nextPageToFetch, setNextPageToFetch] = useState<Record<string, number>>({});
 
     useEffect(() => {
-        queryClient.invalidateQueries({ queryKey: ['childCollections'] });
+        // Hydrate accumulatedChildren from the query cache so that previously
+        // expanded folders show cached data immediately while refetching.
+        const cachedQueries = queryClient.getQueriesData<
+            ResultOf<typeof collectionListDocument>
+        >({ queryKey: ['childCollections'] });
+        const hydrated: Record<string, { items: Collection[]; totalItems: number }> = {};
+        for (const [queryKey, data] of cachedQueries) {
+            if (!data) continue;
+            const collectionId = queryKey[1] as string;
+            if (!collectionId) continue;
+            const page = queryKey[3] as number;
+            if (page === 0) {
+                hydrated[collectionId] = {
+                    items: data.collections.items,
+                    totalItems: data.collections.totalItems,
+                };
+            } else if (hydrated[collectionId]) {
+                hydrated[collectionId] = {
+                    items: [...hydrated[collectionId].items, ...data.collections.items],
+                    totalItems: data.collections.totalItems,
+                };
+            }
+        }
+        if (Object.keys(hydrated).length > 0) {
+            setAccumulatedChildren(hydrated);
+        }
     }, []);
 
     useQueries({
--- a/packages/dashboard/src/app/routes/_authenticated/_collections/collections.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_collections/collections.tsx
@@ -69,9 +69,14 @@ function CollectionListPage() {
     useEffect(() => {
         // Hydrate accumulatedChildren from the query cache so that previously
         // expanded folders show cached data immediately while refetching.
-        const cachedQueries = queryClient.getQueriesData<
-            ResultOf<typeof collectionListDocument>
-        >({ queryKey: ['childCollections'] });
+        const cachedQueries = queryClient
+            .getQueriesData<ResultOf<typeof collectionListDocument>>({ queryKey: ['childCollections'] })
+            .sort((a, b) => {
+                // Sort by page number to ensure page 0 is processed first
+                const pageA = (a[0][3] as number) ?? 0;
+                const pageB = (b[0][3] as number) ?? 0;
+                return pageA - pageB;
+            });
         const hydrated: Record<string, { items: Collection[]; totalItems: number }> = {};
         for (const [queryKey, data] of cachedQueries) {
             if (!data) continue;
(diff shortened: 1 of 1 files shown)


michaelbromley:
Hey, nice catch on the bug! The root cause analysis is spot on — the `.filter()` was preventing React Query from re-running the `queryFn` after remount, so `accumulatedChildren` never got repopulated.

I tested the fix and it works, but I think we can get the same result with a simpler approach. The `useEffect` cache hydration block (lines 69-101) manually reaches into React Query's cache, sorts entries by page number, and reconstructs `accumulatedChildren` from raw cache data. This is fragile because it depends on the query key structure (`queryKey[1]` = collectionId, `queryKey[3]` = page), and if that ever changes, it silently breaks.

The simpler alternative: just remove the `.filter()` (as you've done) and add `refetchOnMount: 'always'` to the query options:

```ts
{
    queryKey: ['childCollections', collectionId, 'page', 0],
    queryFn: async () => {
        // ...same as current
    },
    staleTime: 1000 * 60 * 5,
    refetchOnMount: 'always',
}
```

This tells React Query to always re-execute the `queryFn` when the component mounts, even if the cache is fresh. Since the `queryFn` already calls `setAccumulatedChildren`, the local state gets repopulated automatically — no manual cache hydration needed.

I tested this locally and it behaves identically in the browser (no loading flash or anything), since React Query returns the cached data optimistically while the refetch happens in the background.

This would let us drop the entire `useEffect` block (~35 lines) and the `useEffect` import, replacing it with a single config option.

What do you think?

biggamesmallworld:
Hey @tbouliere-datasolution, nice find on the root cause! 👍

I agree with @michaelbromley's suggestion — the `useEffect` cache hydration block can be replaced with a much simpler approach. Here's exactly what the updated first `useQueries` block should look like:

```tsx
useQueries({
    queries:
        expanded === true
            ? []
            : Object.entries(expanded).map(([collectionId]) => {
                  return {
                      queryKey: ['childCollections', collectionId, 'page', 0],
                      queryFn: async () => {
                          const result = await api.query(collectionListDocument, {
                              options: {
                                  filter: {
                                      parentId: { eq: collectionId },
                                  },
                                  take: CHILDREN_PAGE_SIZE,
                                  skip: 0,
                              },
                          });
                          setAccumulatedChildren(prev => ({
                              ...prev,
                              [collectionId]: {
                                  items: result.collections.items,
                                  totalItems: result.collections.totalItems,
                              },
                          }));
                          return result;
                      },
                      staleTime: 1000 * 60 * 5,
                      refetchOnMount: 'always',
                  } satisfies FetchQueryOptions;
              }),
});
```

**What changed vs the original code:**
1. Removed the `.filter(([collectionId]) => !accumulatedChildren[collectionId])` — this was the root cause preventing React Query from re-running the `queryFn` after remount
2. Added `refetchOnMount: 'always'` — ensures the `queryFn` always re-executes on mount, which repopulates `accumulatedChildren` via `setAccumulatedChildren`

**What to remove:**
- The entire `useEffect` block (lines 69-101 in your current diff) — it's no longer needed since React Query handles the hydration automatically
- The `useEffect` import

This keeps your key insight (removing the `.filter()`) while avoiding the fragile manual cache hydration that depends on query key structure. The `refetchOnMount: 'always'` option makes React Query do the heavy lifting — it returns cached data optimistically while silently refetching in the background.

Also, consider reverting the unrelated formatting changes (line wrapping, import reordering, etc.) to keep the diff focused on the bug fix — it makes review easier.

tbouliere-datasolution:
Yeah it is a much better solution.
I'm not so used to the tanstack framework, I've should have check the doc and look for the `refetchOnMount `parameter instead for coding a workaround ad-hoc.

michaelbromley:
I also manually tested the pagination ("Load more") feature with 27 child collections under a single parent (well past the page size of 20). Expanding the folder, clicking "Load more", navigating away and back, and loading more pages again all work correctly. The pagination concern in the breaking changes section is a non-issue.

tbouliere-datasolution:
«Also, consider reverting the unrelated formatting changes (line wrapping, import reordering, etc.) to keep the diff focused on the bug fix — it makes review easier.»

I agree.

I have a prettier plugin pre-installed on my vscode that automatically formats the code when saving. It seems to follow the https://github.com/vendurehq/vendure/blob/master/.prettierrc

grolmus (commented):
The root cause analysis is correct — `accumulatedChildren` state resets on unmount, but React Query cache retains data (staleTime: 5min). Since `queryFn` doesn't re-run for fresh cached data, `setAccumulatedChildren` never fires and expand breaks.

The hydration `useEffect` fixes this. A few notes:

### 1. Query key structure is fragile
```ts
const collectionId = queryKey[1] as string;
const page = queryKey[3] as number;
```
This assumes `['childCollections', collectionId, 'page', pageNumber]` — if the key structure changes elsewhere in this file, this silently breaks. Consider extracting a helper or constant for the key structure.

### 2. Filter removal creates redundant queries
The removed `.filter(([collectionId]) => !accumulatedChildren[collectionId])` was an optimization — without it, `useQueries` creates queries for ALL expanded collections on every render (even those already hydrated). They'll be served from cache so no network cost, but it's unnecessary work. Consider keeping the filter.

### 3. Lots of formatting noise
Most of this diff is prettier reformatting. Consider splitting into a separate commit or squashing so the logic change is easier to review.

The core fix (useEffect hydration) is sound though.

HouseinIsProgramming:
Thanks for this, and sorry for the long silence.

Your diagnosis was correct: `setAccumulatedChildren` inside `queryFn` never ran on a cache hit, so a re-mounted page had expanded folders with no children. That is now fixed on master by #4509 (commit 8f133ca), which makes `queryFn` pure and syncs `accumulatedChildren` from the query results in a `useEffect`, so it fires for cache hits and fresh fetches alike. The file has moved on enough that this branch no longer applies.

Closing as superseded, and thank you for the report and the fix.

PR #4501, opened 2026-03-12
fix(core): use schema-qualified table paths in EXISTS subqueries

yasserlens:
# Description

Fixes a bug where `ListQueryBuilder.buildExistsSubquery()` generated raw SQL using unqualified table names (`inverseEntityMeta.tableName` and `junctionMeta.tableName`). 

When a custom PostgreSQL schema is configured (e.g. `schema: 'vendure'`), the generated EXISTS subquery would fail with a `QueryFailedError: relation "payment" does not exist` error. This happens because the query references the table without the schema qualification, causing Postgres to default to searching the `public` schema.

This PR fixes the issue by updating the code to use TypeORM's `.tablePath` instead of `.tableName` for both One-to-Many relations and Many-to-Many junction tables. The `tablePath` correctly includes the schema prefix (e.g. `"vendure"."payment"`), allowing queries to execute properly on non-default schemas.

# Breaking changes

No breaking changes. This simply ensures that TypeORM queries correctly respect the schema configuration provided in the database connection options.

# Screenshots

N/A

# Checklist

📌 Always:
- [x] I have [checked my own PR](<## "Fix typo's and remove unused or commented out code">)

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/lib/components/data-input/relation-input.tsx
+++ b/packages/dashboard/src/lib/components/data-input/relation-input.tsx
@@ -35,7 +35,7 @@ export function SingleRelationInput<T>({
             config={singleConfig}
             value={value}
             selectorLabel={selectorLabel}
-            onChange={newValue => onChange(newValue as string)}
+            onChange={newValue => onChange(newValue)}
             disabled={disabled}
             className={className}
         />
@@ -69,7 +69,7 @@ export function MultiRelationInput<T>({
         <RelationSelector
             config={multiConfig}
             value={value}
-            onChange={newValue => onChange(newValue as string[])}
+            onChange={newValue => onChange(newValue)}
             disabled={disabled}
             className={className}
             selectorLabel={selectorLabel}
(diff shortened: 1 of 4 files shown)


michaelbromley:
Hi @yasserlens — thanks for the PR! The underlying issue is definitely real.

Quick question: did you get a chance to test this fix against a Postgres instance with a custom schema that **doesn't** match the database username? 

We tested it out locally and found that while `tablePath` is the right direction (it includes the schema prefix), there's an escaping issue with the current implementation. The `escapeId()` helper wraps the entire string in quotes as a single identifier:

```
escapeId("myapp.product_facet_values_facet_value")
→ "myapp.product_facet_values_facet_value"   // single identifier with a literal dot
```

What's actually needed is each segment escaped individually:

```
"myapp"."product_facet_values_facet_value"   // schema-qualified reference
```

A fix would be to add a helper that splits on `.` before escaping:

```typescript
const escapeTablePath = (path: string) =>
    path.split('.').map(segment => mainQb.connection.driver.escape(segment)).join('.');
```

Then use `escapeTablePath` for the table names (lines 510, 511, 547) while keeping `escapeId` for aliases.

We verified this locally with `DB_SCHEMA=myapp` (a schema not in Postgres's `search_path`) — the current `tableName` fails, the PR's `tablePath` swap still fails due to the escaping, and the split-and-escape approach works correctly.

yasserlens:
Hi @michaelbromley thanks for reviewing. I just pushed a new commit with the suggested changes.

Issue #4506, opened 2026-03-15
DataTable footer shows "tffg61" instead of row selection text in production builds


Ryrahul:
## Describe the bug
The DataTable footer displays the raw Lingui hash `tffg61` instead of the row selection text (e.g. "0 of 25 row(s) selected.") in production builds.

## To Reproduce
1. Build the dashboard for production
   ```bash
   npx vite build
   ```
2. Serve it via server.
3. Navigate to any page with a DataTable that supports row selection (e.g. Products list).
4. See `tffg61` in the bottom-left footer instead of the selection count.

## Expected behavior
The footer should display "0 of N row(s) selected." (or the localized equivalent).

## Actual behavior
The footer displays the raw message hash `tffg61`.

## Screenshots/Videos

<img width="484" height="688" alt="Image" src="https://github.com/user-attachments/assets/de69347c-7deb-438c-aed8-d83b75c3a89e" />

## Error Logs
No errors in the console. The hash is silently rendered as text.

## Environment
- **@vendure/core version:** 3.5.5  
- **Nodejs version:** >= 20  
- **Database:** Any  
- **Operating System:** Any  
- **Browser:** Any  
- **Package manager:** npm  

## Configuration
Default configuration. No special setup required to reproduce.

## Minimal reproduction
Build the dashboard for production and visit any list view with row selection.

## Additional context
The `<Trans>` component in  
`packages/dashboard/src/lib/components/data-table/data-table-pagination.tsx`  had no corresponding entry in any `.po` locale file.

In development, Lingui uses the inline JSX as a fallback so it works fine.  
In production, `stripMessageField=true` removes the fallback, and the catalog lookup for the auto-hashed ID `tffg61` fails, rendering the raw hash.

Ryrahul:
Also in order notes 

<img width="949" height="541" alt="Image" src="https://github.com/user-attachments/assets/9956740b-b5ba-41e0-b1f9-56518c7b1193" />

PR #4507, opened 2026-03-15
fix: Add missing i8n catalog entry

Ryrahul:
## Description
Fixes missing i18n catalog entry for the DataTable row selection text.

The `<Trans>` component in `data-table-pagination.tsx` had no corresponding entry in any `.po` locale file, causing the raw Lingui hash `tffg61` to be displayed instead of `{selectedRowCount} of {N} row(s) selected.` in production builds.

Production builds use `stripMessageField=true`, which removes inline fallbacks. Without a catalog entry, the hash is rendered as-is.

**Fix:** Ran `lingui extract` to add the missing entry to all 26 locale `.po` files.

Fixes: #4506  

---

## Breaking changes
None.

---

## Screenshots
N/A

---

## Checklist


Ryrahul:
fixed in #4578

PR #4508, opened 2026-03-16
feat(dashboard): Custom form labels

LucidityDesign:
# Description

This PR extends the `defineDashboardExtension` function so users can provide custom form labels.

This is especially handy for custom entities that are rendered using the `DetailPage` component. Currently `DetailPage` renders the column name as form label. E.G. `oderId` for a relational field.

I couldn't find any tests for `defineDashboardExtension` so I did not write new ones.

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/docs/docs/reference/dashboard/detail-views/detail-page.mdx
+++ b/docs/docs/reference/dashboard/detail-views/detail-page.mdx
@@ -2,7 +2,7 @@
 title: "DetailPage"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/page/detail-page.tsx" sourceLine="134" packageName="@vendure/dashboard" since="3.3.0" />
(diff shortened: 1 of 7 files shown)


michaelbromley:
Thanks for the suggestion. Rather than adding a new top-level property, I think it would be a better, more cohesive design to consider whether it's possible to extend the existing `DashboardDetailFormInputComponent` interface with an optional label component.

I also want to point out that `DetailPage` is intentionally limited - it's mainly there to allow you to rapidly get started interacting with your API, as explained in [the docs](https://docs.vendure.io/current/core/extending-the-dashboard/creating-pages/detail-pages#customizing-the-detail-page)

> The auto-generated [DetailPage](https://docs.vendure.io/current/core/reference/dashboard/detail-views/detail-page) is a great way to get started and quickly be able to interact with your entities.

So another alternative when you want more control is to just move to the fully custom detail page explained in the next section of that doc.

LucidityDesign:
Thanks, I understand. I will extend my custom detail page then.

PR #4509, opened 2026-03-16
fix(dashboard): Persist collection tree expanded state in URL search params

### LINKED ISSUE #2: Design: Facets
Facets are properties which can be assigned to a product / product variant and are used to encode properties of that product along arbitrary dimensions. Reference: https://www.sitepoint.com/cms-content-organization-structures-trees-vs-facets-vs-tags/

## Examples of facets

* Brand: HP, Dell, Apple, Microsoft
* Form Factor: Laptop, Desktop, Tablet, Hybrid
* Memory: 4GB, 8GB, 16GB, 32GB

## Purpose of facets

Facets are used to add structured metatdata to a product for the purpose of categorization and filtering.

A typical use-case is to have a filter-based product-discovery interface e.g. like [Asos](http://www.asos.com/men/sale/cat/?cid=8409&ctaref=shop|salehub|viewall) which allows filtering by various facets. Amazon also use this approach.

If a shop would prefer a tree-like category-based navigation, this can be built up via pre-defined combinations of facets which are grouped into a category. Further design on facet-based categories to follow.

## Questions to resolve

1. Should facets be associated with a Product, a ProductVariant, or both?
2. Should the range of facet values be fixed to a pre-defined list of values, or should new values be generated on an ad-hoc basis (like how tags work in WordPress)?
3. How should Facets and FacetValues be stored in the database? A Facet entity with a one-to-many collection of FacetValues? Or is there some simpler way to store them?

### LINKED ISSUE #4388: Dashboard: When clicking on a collection and going back to the collection list it's not possible to expand the collection you previously visited
**Describe the bug**
When visiting a collection in the new dashboard and going back to the list its not possible to expand the collection you visited to show its subcollections without reloading the page.

**To Reproduce**
* Visit a collection with subcollections
* Go back to collections list
* Attempt to expand the collection to view the subcollections
* See that it does not expand and show the children

**Expected behavior**
That collections are expandable after visiting them without reloading the page.

**Actual behavior**
Cant show subcollections when expanding after visiting them.

**Screenshots/Videos**

- Nodejs version: 24
- Database (mysql/postgres etc): PG
- Browser (if applicable):

**Workaround**
Refresh the page

**Additional context**

- When did this issue start occurring? After 3.5.4 upgrade
- Does it happen consistently or intermittently? Always.
- Any recent changes to your setup? NO.

biggamesmallworld:
Persists the expanded/collapsed state of collection tree rows in URL search params (\`?expanded=1,2,5\`). Also fixes several root-cause bugs related to expansion, drag-and-drop positioning, and collection movement.

Fixes #4388

### 1. URL persistence (original approach)
Stores expanded row IDs in \`?expanded=\` search params so the tree state survives navigation and page reload.

### 2. Root cause fix — queryFn cache miss
\`useQueries\` was calling \`setAccumulatedChildren\` as a side effect inside \`queryFn\`. TanStack Query skips \`queryFn\` entirely for cache hits (\`staleTime: 5min\`), so on component re-mount after SPA navigation the children were never populated from cache — clicking the expand button appeared to do nothing. Fixed by returning data from \`queryFn\` and syncing to state via \`useEffect\`, which fires for both cache hits and fresh fetches.

### 3. Root cause fix — URL params wiped on mount
\`list-page.tsx\`'s \`persistListStateToUrl\` was calling \`navigate\` with a fresh search object, dropping all existing params. \`DataTable\` fires \`onSortChange\`/\`onPageChange\` effects on mount, so ~1s after loading a URL like \`?expanded=123\` the param was silently removed. Fixed by spreading the previous search params.

### 4. Fix cross-parent drag positioning
When moving a collection to a different parent via drag-and-drop, it would land at the second-to-last position instead of the intended last position. Two issues contributed:

- **Server: `moveToIndex` skipped position reassignment when no splice was needed.** When a cross-parent move appended the target at the correct array index, `currentIndex === normalizedIndex` caused the position reassignment loop to be skipped. The target kept its stale `position` value from the old parent, so after re-fetching and sorting by position it appeared in the wrong spot. Fix: always reassign sequential positions to all siblings.

- **Dashboard: cross-parent index was derived from the visible siblings array, which could be stale or incomplete.** When dropping after the last visible sibling, the dashboard now uses the parent collection's `children.length` (a non-paginated relation field reflecting the true child count) to calculate the target index.

## Tests
Three new Playwright tests in \`catalog/collections.spec.ts\`:
- URL param is set when expanding a collection
- Expanding works after navigating **directly** to a detail page and back (no prior expansion — specifically tests root cause fix #2)
- Expanded state is restored from URL after navigating to detail and back

No breaking changes.

## Screen recording

https://github.com/user-attachments/assets/c4430863-e101-422a-b68a-0b85ad8e1c0c

### DIFF
--- a/docs/docs/reference/dashboard/vite-plugin/vendure-dashboard-plugin.mdx
+++ b/docs/docs/reference/dashboard/vite-plugin/vendure-dashboard-plugin.mdx
@@ -181,6 +181,7 @@ vendureDashboardPlugin({
 interface PathAdapter {
     getCompiledConfigPath?: GetCompiledConfigPathFn;
     transformTsConfigPathMappings?: TransformTsConfigPathMappingsFn;
+    sourceRoot?: string;
 }
 ```
 
@@ -196,6 +197,21 @@ A function to determine the path to the compiled Vendure config file.
 <MemberInfo kind="property" type={`TransformTsConfigPathMappingsFn`}   />
 
 
+### sourceRoot
+
+<MemberInfo kind="property" type={`string`}  since="3.6.0"  />
+
+The root directory used to compute relative output paths when compiling
+TypeScript files. Compiled files preserve their directory structure
+relative to this root.
+
+In monorepos this should typically be set to the workspace root (where
+the base `tsconfig.json` lives), so that a config file at
+`apps/server/src/vendure-config.ts` outputs to
+`{outputPath}/apps/server/src/vendure-config.js`.
+
+Defaults to the directory containing the `vendureConfigPath` file,
+which places the compiled config at the output root.
 
 
 </div>
--- a/packages/dashboard/src/app/routes/_authenticated/_orders/components/fulfillment-details.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_orders/components/fulfillment-details.tsx
@@ -171,7 +171,8 @@ export function FulfillmentDetails({ order, fulfillment, onSuccess }: Readonly<F
 
             <div className="mt-3 pt-3 border-t">
                 <StateTransitionControl
-                    currentState={getTranslatedFulfillmentState(fulfillment.state)}
+                    currentState={fulfillment.state}
+                    statesTranslationFunction={getTranslatedFulfillmentState}
                     actions={getFulfillmentActions()}
                     isLoading={transitionFulfillmentMutation.isPending}
                 />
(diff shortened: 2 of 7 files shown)


gabriellbui:
I'm not sure if the 2 things below are related, but I just wanted to mention them so we have a record of them.

1. Manually moving a collection also seem to be kinda broken. You have to refresh the page to see the change.
2. Small: when clicking on the browser back/forward buttons, the URL query params and UI is not in sync unlike e.g. when sorting. 


https://github.com/user-attachments/assets/b413e75d-c696-4c1f-9d87-3555e0f4aee9

biggamesmallworld:
@michaelbromley @gabriellbui thanks for keeping me honest here. I believe my refactors after receiving the sonarqube review were not properly tested. 

Here is a screen recording of this all working properly: 


https://github.com/user-attachments/assets/65fa38fc-2e44-484c-b894-5f2db78d5539

michaelbromley:
Nice work Will - that all looks good now 👍

PR #4536, opened 2026-03-23
feat(cli): Add codemod command for Radix to Base UI dashboard migration

dlhck:
## Summary

Adds `vendure codemod` CLI command for running automated code transformations, starting with a dashboard Base UI migration codemod. Also adds a Claude Code skill for agent-assisted migration and updates the existing Angular-to-React migration skill.

### CLI: `vendure codemod <transform> [path]`

```bash
vendure codemod dashboard-base-ui .          # run on a specific path
vendure codemod dashboard-base-ui            # run on current directory
vendure codemod                              # interactive: pick from available codemods
```

New codemods are registered in a simple map — no CLI flags or command declaration changes needed. The `[path]` argument is validated (must exist, must be a directory) and passed explicitly to each codemod. Prefers `tsconfig.dashboard.json` when available, falls back to `tsconfig.json`, and auto-scans for `**/*.tsx` if the tsconfig yields no files.

### `dashboard-base-ui` codemod transforms (all using ts-morph AST)

- **`asChild` to `render` prop** — Converts Radix slot pattern to Base UI render prop for Button, DialogTrigger, PopoverTrigger, SheetTrigger, etc. Skips unconvertible patterns (JSX expression children, multiple children) with a warning and continues processing the rest of the file.
- **`FormField` to `FormFieldWrapper`** — Extracts label, description, and form control content from the old FormField/FormItem/FormControl/FormLabel/FormMessage pattern. Falls back to a TODO comment for complex cases (JSX labels, missing FormControl). Cleans up unused form imports.
- **Import consolidation** — Rewrites `@radix-ui/*`, `@vendure-io/ui/*`, `@base-ui/react/*` imports to `@vendure/dashboard`. Converts namespace imports (`import * as Dialog`) to flat named imports and rewrites all usage sites (`Dialog.Root` to `Dialog`, `Dialog.Trigger` to `DialogTrigger`). Also rewrites third-party imports that are re-exported from `@vendure/dashboard` (`react-hook-form`, `@tanstack/react-query`, `@tanstack/react-router`, `sonner`, `@lingui/react`, etc.) while correctly preserving imports that must stay direct (`@lingui/react/macro`, actual icon components from `lucide-react`).
- **Accordion props** — Removes deprecated `type="single"` and `collapsible` props.
- **Select items prop** — Auto-generates the `items` array from static `<SelectItem>` children (`[{ label: 'Draft', value: 'draft' }, ...]`). Warns for dynamic patterns that can't be inferred.

### Claude Code skill

New skill at `.claude/skills/radix-to-base-ui-migration/` with docs covering all transformation patterns, import consolidation rules (including the Lingui macro exception), and component API changes.

### Existing skill update

Updated `.claude/skills/vendure-dashboard-migration/` to use Base UI patterns instead of stale Radix/shadcn references (`asChild` to `render`, "shadcn/ui" to "Vendure Design System").

## Test plan
- [ ] Build CLI: `cd packages/cli && npm run build`
- [ ] Run codemod against a sample extension: `vendure codemod dashboard-base-ui /path/to/extension`
- [ ] Run `npm test` in `packages/cli` — 46 tests covering all transforms including edge cases
- [ ] Verify `vendure codemod --help` shows positional args syntax
- [ ] Verify `vendure codemod` interactive mode lists available codemods
- [ ] Verify `vendure codemod unknown-name` gives clear error with available codemods list

Relates to #4531

### DIFF
--- a/packages/cli/src/commands/codemod/dashboard-ui/dashboard-ui-migration.spec.ts
+++ b/packages/cli/src/commands/codemod/dashboard-ui/dashboard-ui-migration.spec.ts
@@ -159,6 +159,25 @@ const el = (
         expect(changes).toBe(0);
     });
 
+    it('should preserve inline whitespace between JSX expression children', () => {
+        const sf = createSourceFile(`
+const el = (
+    <Button asChild variant="ghost">
+        <Link to={\`/customers/\${value.id}\`}>
+            {value.firstName} {value.lastName}
+        </Link>
+    </Button>
+);
+`);
+        const changes = transformAsChildToRender(sf);
+        expect(changes).toBe(1);
+        const text = sf.getFullText();
+        expect(text).toContain('render={<Link to={`/customers/${value.id}`} />}');
+        // The space between the two expressions must be preserved
+        expect(text).toContain('{value.firstName} {value.lastName}');
+        expect(text).not.toContain('{value.firstName}{value.lastName}');
+    });
+
     it('should skip asChild with multiple children and warn', () => {
         const sf = createSourceFile(`
 const el = (
--- a/packages/cli/src/commands/codemod/dashboard-ui/transforms/as-child-to-render.ts
+++ b/packages/cli/src/commands/codemod/dashboard-ui/transforms/as-child-to-render.ts
@@ -103,10 +103,15 @@ export function transformAsChildToRender(sourceFile: SourceFile): number {
                     .getAttributes()
                     .map(a => a.getText())
                     .join(' ');
-                grandchildrenText = childEl
-                    .getJsxChildren()
-                    .map(c => c.getText())
-                    .join('');
+                // Use raw source text to preserve inline whitespace between JSX children
+                // (e.g. `{value.firstName} {value.lastName}` — the space would be lost
+                // if we relied on getJsxChildren().map(c => c.getText()).join(''))
+                const innerStart = childOpening.getEnd();
+                const closingElement = childEl.getClosingElement();
+                const innerEnd = closingElement.getStart();
+                const rawInner = sourceFile.getFullText().substring(innerStart, innerEnd);
+                const childIndent = getLineIndent(sourceFile, childOpening.getStartLineNumber());
+                grandchildrenText = childIndent > 0 ? dedentText(rawInner, childIndent) : rawInner;
             } else {
                 const childEl = childNode as JsxSelfClosingElement;
                 childTagName = childEl.getTagNameNode().getText();
@@ -153,4 +158,15 @@ export function transformAsChildToRender(sourceFile: SourceFile): number {
     }
 
     return changeCount;
+}
+
+function getLineIndent(sourceFile: SourceFile, lineNumber: number): number {
+    const lineText = sourceFile.getFullText().split('\n')[lineNumber - 1] ?? '';
+    const match = lineText.match(/^(\s*)/);
+    return match ? match[1].length : 0;
+}
+
+function dedentText(text: string, amount: number): string {
+    const regex = new RegExp(`^ {1,${amount}}`, 'gm');
+    return text.replace(regex, '');
 }


michaelbromley:
## Codemod Review — Tested against a real monorepo

Ran `vendure codemod dashboard-base-ui` against a production monorepo with 43 TSX files (26 files changed, 66 transformations). The `asChild` → `render` pattern is structurally correct across all files, and the import source detection is accurate. Nice work on the manual warnings for dynamic `<Select>` items too.

Found a few issues worth addressing:

---

### Bug: `asChild` → `render` transform loses inline whitespace between JSX children

The example was 

```tsx
<Button asChild variant="ghost">
    <Link to={`/customers/${value.id}`}>
        {value.firstName} {value.lastName}
    </Link>
</Button>
```


**File:** `packages/cli/src/commands/codemod/dashboard-ui/transforms/as-child-to-render.ts`, lines 106–109

`getJsxChildren().map(c => c.getText()).join('')` strips whitespace from `JsxText` nodes. A single space between two expressions (e.g. `{value.firstName} {value.lastName}`) gets lost, producing `{value.firstName}{value.lastName}`.

**Fix:** Replace the `getJsxChildren()` approach with raw source text extraction between the child element's opening and closing tags:

```typescript
const innerStart = childOpening.getEnd();
const innerEnd = childEl.getClosingElement().getStart();
const rawInner = sourceFile.getFullText().substring(innerStart, innerEnd);
```

This also requires dedenting the raw text so that `replaceWithText()` (which auto-indents subsequent lines by the replaced node's column offset) produces correct indentation. Dedent by the **child element's full indentation**, not the parent–child delta:

```typescript
const childIndent = getLineIndent(sourceFile, childOpening.getStartLineNumber());
grandchildrenText = childIndent > 0 ? dedentText(rawInner, childIndent) : rawInner;
```

With two helpers:

```typescript
function getLineIndent(sourceFile: SourceFile, lineNumber: number): number {
    const lineText = sourceFile.getFullText().split('\n')[lineNumber - 1] ?? '';
    const match = lineText.match(/^(\s*)/);
    return match ? match[1].length : 0;
}

function dedentText(text: string, amount: number): string {
    const regex = new RegExp(`^ {1,${amount}}`, 'gm');
    return text.replace(regex, '');
}
```

**Note:** This fixes most cases but indentation is still slightly off at deeper nesting levels (e.g. a `<Button asChild>` inside a ternary inside `<CardContent>` ended up with content 8 spaces too deep instead of 4). The `replaceWithText()` auto-indent behaviour doesn't perfectly match a simple "node column offset" model at all depths. Worth investigating further or documenting as a known limitation that a formatter pass would clean up.

---

### Bug: `<Select>` items prop duplicates children

Note from human: pretty sure this is a false positive based on what you showed me.

**File:** `packages/cli/src/commands/codemod/dashboard-ui/transforms/select-items-prop.ts`

When the codemod can statically extract `<SelectItem>` values, it adds an `items` prop to `<Select>` — but the `<SelectItem>` children remain. The options end up defined in two places:

```tsx
<Select items={[{ label: 'All reviews', value: 'all' }, ...]}>
    <SelectContent>
        <SelectItem value="all">All reviews</SelectItem>   // ← still here
        ...
    </SelectContent>
</Select>
```

Questions:
- Should the codemod also remove the `<SelectContent>` block when it successfully adds `items`?
- Or is the `items` prop meant to coexist with children? If so, the duplication is confusing.
- The `items` array also makes for a very long single line (~250+ chars).

---

### Cosmetic: Import consolidation appends to last line instead of maintaining multi-line format

**File:** `packages/cli/src/commands/codemod/dashboard-ui/transforms/import-consolidation.ts`

All 26 files end up with new specifiers crammed onto the last line of the existing import block:

```tsx
// Before
import {
    PageTitle,
    PermissionGuard,
    useDetailPage,
} from '@vendure/dashboard';
import { AnyRoute, useNavigate } from '@tanstack/react-router';
import { toast } from 'sonner';

// After
import {
    PageTitle,
    PermissionGuard,
    useDetailPage, AnyRoute, useNavigate, toast } from '@vendure/dashboard';
```

Ideally each specifier should stay on its own line with a trailing comma. This could be done by manipulating named imports via the ts-morph API (`addNamedImport`) rather than text manipulation, or by adding each specifier on a new line before the closing `}`.

---

### Cosmetic: Blank line between imports and first code statement removed

About 7 files lose the blank line that separated the import block from the first code statement (e.g. `const redirectDetailDocument = graphql(...)`). This happens when a removed import was the last statement before the gap. The blank line gets consumed with the deletion.

Could be fixed by checking if removing an import declaration also removes a trailing blank line, and re-inserting it.

---

### What looked good

- `asChild` → `render` pattern is structurally correct everywhere
- Import source detection (`@tanstack/react-router`, `@tanstack/react-query`, `sonner`, `react-hook-form` → `@vendure/dashboard`) is accurate
- Partial imports correctly preserved (e.g. `useParams` stays at `@tanstack/react-router`)
- `DropdownMenuTrigger`, `DialogTrigger`, `DialogClose` transforms all correct
- Manual warnings for dynamic `<Select>` items are appropriate and actionable

dlhck:
Thanks for the thorough review Michael. Fixed the inline whitespace bug — you nailed the root cause. Using raw source text extraction now instead of `getJsxChildren().map().join()`, which preserves spaces between JSX expressions like `{value.firstName} {value.lastName}`.

On the other points:

- **Select items duplication**: This is working as intended. The `items` prop is needed so the Select can display the correct label for a selected value without the dropdown being open. The `<SelectContent>` with `<SelectItem>` children is still required for rendering the actual dropdown. They serve different purposes, so both need to be present.
- **Import format (multi-line → single-line)**: A codemod's job is structural correctness, not formatting. Run your formatter (Prettier/ESLint) after the codemod and it'll sort itself out.
- **Blank line between imports and code**: Same deal — formatter territory, not codemod territory.

PR #4538, opened 2026-03-23
fix(docs): correct typo in ProductEvent JSDoc

latifniz:
Description

While working with the EventBus, I noticed a small grammatical issue in the JSDoc in the file:

packages/core/src/event-bus/events/product-event.ts

I corrected the line:

Return an product field to become compatible with the deprecated old version of ProductEvent

to:

Return a product field to become compatible with the deprecated old version of ProductEvent

This is a minor fix, but the typo was slightly confusing on first read, so I thought it would be helpful to correct it.

Breaking changes

No breaking changes


latifniz:
Ops, I made this PR from the wrong branch, closing it. making new one

Issue #4365, opened 2026-02-17
Build fails with Lingui/Vite error when using yarn instead of npm


Ryrahul:
**Describe the bug**
When using `yarn` as the package manager, building the project fails with a Lingui/Vite compilation error. The build process works correctly when using `npm` instead of `yarn`.

**To Reproduce**

1. Clone the repository
2. Run `yarn install`
3. Run `yarn build`
4. See error during the Vite build step

**Expected behavior**
The build should complete successfully, compiling both the Ui's.

**Actual behavior**
The build fails during the Vite compilation step with a Lingui macro compilation error.

**Error logs**
```
[commonjs--resolver] The macro you imported from "@lingui/react/macro" is being executed outside the context of compilation.
error Command failed with exit code 1.
```

**Environment (please complete the following information):**

- @vendure/core version: 3.5.4
- Nodejs version: v20.19.6
- Database (mysql/postgres etc): SQLite (better-sqlite3)
- Operating System (Windows/macOS/Linux): macOS
- Browser (if applicable): N/A
- Package manager (npm/yarn/pnpm): yarn

**Configuration**
The project uses `@vendure/admin-ui-plugin` and `@vendure/ui-devkit` version 3.5.4.

**Minimal reproduction**
Repository: https://github.com/Ryrahul/Vendure-lingui-bug-reproduction

**Workaround**
Use `npm` instead of `yarn` for installing dependencies and building.

**Additional context**
- The Angular UI devkit also uses Vite, which causes a Vite version mismatch when using yarn
- This issue does not occur with npm
- Removing the Angular UI component resolves the issue

Ryrahul:
Did look into it once more 
The real culprit i belive is  @vendure/ui-devkit, which pulls in:
- @angular-devkit/build-angular - pins @babel/core to exactly 7.26.10
- @angular/compiler-cli - pins @babel/core to exactly 7.26.9


1. These exact-pinned versions cause @babel/core@7.26.10 to occupy the top-level
node_modules/
2. @vendure/dashboard needs @babel/core@^7.26.0, which resolves to 7.29.0 - a different
version

3. Since 7.29.0 can't share the top-level slot (7.26.10 is there), Yarn vi nests the
dashboard's entire @babel/* subtree inside node_modules/@vendure/dashboard/node_modules/

4. That includes @babel/preset-typescript, which Vite tries to resolve from the project
root - and fails

Issue #3929, opened 2025-10-31
Dashboard Build Error: failed to invoke plugin on 'Some(...)'


michaelbromley:
### Summary

The Dashboard uses the [lingui](https://lingui.dev/) library to handle UI translations. As part of this, there is a compilation step that internally uses the `@swc/core` package.

Recent updates to SWC and the corresponding `@lingui/swc-plugin` packages can cause errors when attempting to build the Vendure Dashboard which look like this:

```
9:41:26 PM [vite] Pre-transform error: plugin

  × failed to invoke plugin on 'Some("/path/to/my-shop/node_modules/@vendure/dashboard/src/app/main.tsx")'

  Plugin: vite:react-swc
  File: /path/to/my-shop/node_modules/@vendure/dashboard/src/app/main.tsx
```

### Solution

From Vendure v3.5.2 this issue should be fixed. However, if you have other dependencies which also use SWC, you may still see this error.

The fix for now is to use your package manager's **override** feature to make sure these specific versions of the packages are installed:

In your package.json, add this after the "dependencies" and then reinstall. 

```json
{
  "overrides": {
    "@swc/core": "1.13.5"
    "@lingui/swc-plugin": "5.6.1"
  }
}
```

If the problem persists, you can try also deleting your node_modules directory, and also your lockfile

```
rm -rf node_modules && rm package-lock.json && npm install
```

You can find which versions you current have installed using:

```
npm ls @swc/core
npm ls @lingui/swc-plugin
```

### Update: 9th Jan 2026

This issue on the lingui/swc-plugin repo addresses this topic:

- https://github.com/lingui/swc-plugin/issues/179

This explains the rather fragile nature of these dependencies - especially considering that Vendure is often house in a monorepo where it is quite likely that some other dependency may have it's own `@swc/core` dependency, which can then cause conflicts depending on how the packages are hoisted.

I'm looking into a more robust solution for this - perhaps bundling into a single binary so that the Vendure Dashboard compiler is isolated from any other dependencies.

Relecto:
if you're using pnpm, add the following to `package.json`:
```json
  "pnpm": {
    "overrides": {
      "@swc/core": "1.13.5"
    }
  }
```

michaelbromley:
Update: 9th December 2025

It looks like this issue popped up again after yesterday's release of new `@lingui/swc-plugin` versions.

Try setting this version in your package.json:

```
  "overrides": {
    "@lingui/swc-plugin": "5.6.1"
[...]

PR #4571, opened 2026-03-26
fix: Build fails with Lingui/Vite error when using yarn instead of npm

harshit078:
# Description

-  this PR addresses issue #4365
- added createRequire to resolve the absolute paths to @babel/preset-typescript, @babel/preset-react, and @lingui/babel-plugin-lingui-macro from the dashboard location. 
- By passing absolute paths, we bypass Babel's resolution logic entirely. This works no matter if we use yarn or npm

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


harshit078:

grolmus (changes requested):
Several issues with this PR:

1. **Duplicate import** — `import { createRequire } from 'node:module'` appears twice (lines 1 and 3)
2. **Duplicate `createRequire` call** — both `_require` and existing `require` are created from `createRequire(import.meta.url)`, only one is needed
3. **Duplicate presets** — the babel config now has BOTH the new resolved presets AND the old `require.resolve()` ones:
   - `presetTypescript` + `require.resolve('@babel/preset-typescript')` → typescript preset runs twice
   - `presetReact` appears twice in the new code
4. **Incomplete** — `@lingui/babel-plugin-lingui-macro` is still using the old `require.resolve()` approach

The intent is correct (resolve absolute paths for yarn PnP compatibility), but the implementation needs cleanup: remove the duplicate import, remove the duplicate `_require`/`require`, replace the old preset entries instead of adding alongside them, and also resolve the lingui plugin.

michaelbromley:
Hi @harshit078 
I’ve looked at this PR and I can’t understand how it works to fix the reported yarn issue. Can you please explain what’s going on and how the changes address the issue?

harshit078:
Hey @michaelbromley , sure I'll give a brief detailed answer. 

## Original issue

- Build failed with Lingui/Vite error when using yarn instead of npm. When using yarn as the package manager, building the project fails with a Lingui/Vite compilation error. The build process works correctly when using npm instead of yarn.
- basically on research I found out that` @vendure/ui-devkit` depends on `@angular-devkit/build-angular` and `@angular/compiler-cli`, which depends on `@babel/core to 7.26.10 and 7.26.9` respectively. Whereas `@vendure/dashboard` requires us to have `@babel/core@^7.26.0`, which resolves to 7.29.0.

## Why Npm solved it but not yarn

- `npm` simply hoists one copy and then rest of the tree still finds Babel where it should expect it to be.
- `yarn` on the end hand can only put one version at the top level. Since 7.26.10 wins, the dashboard's entire @babel/* subtree gets nested under node_modules/@vendure/.

- What I understood so fat is that babel resolves plugin string names using Node module resolution relative to the file being transformed. With yarn, those packages don't exist on the path from the project root so babel never loads `@lingui/babel-plugin-lingui-macro`, the macro stays in the source, and Vite's thows the error we see.

## Fix

- In `packages/dashboard/vite/vite-plugin-lingui-babel.ts`, we resolve three Babel packages once from line 6-10. We then pass these in line 137-138. When babel receives an absolute path instead of a package name, it just simply skips its own logic entirely and loads the file directly.

## Why this works for both npm and yarn

- For npm the @babel/* lives at the top level root and require.resolve from the dashboard finds it way normally without any logic.
- For yarn the `@babel/*` is nested under the dashboard and require.resolve from the dashboard this time finds it immediately in its local node_modules.

harshit078:
Hey @grolmus , I have pushed a fix for comments you mentioned. Can you please take a look again ? Thanks !

grolmus (approved):
Re-reviewed — all four points from the previous round are addressed, verified against the current code:

1. ✅ Single `import { createRequire } from 'node:module'`.
2. ✅ One `const require = createRequire(import.meta.url)` (the duplicate `_require` is gone).
3. ✅ The TypeScript and React presets are resolved once each via `presetTypescript` / `presetReact` — no more duplicate `require.resolve(...)` entries alongside them.
4. ✅ The Lingui macro plugin now uses the same resolved approach (`const linguiPlugin = require.resolve('@lingui/babel-plugin-lingui-macro')`).

The `require.resolve()` approach correctly resolves absolute paths for yarn PnP compatibility, which is the point of the fix. Thanks for the cleanup, @harshit078 — approving.

grolmus:
Thanks a lot for this @harshit078, and sorry for the slow round-trip. Having re-reviewed, the cleanup you pushed is correct — but in the meantime the exact same fix has already landed on `master`: #4531 (the Radix → Base UI migration) resolves the Babel presets and the `@lingui/babel-plugin-lingui-macro` plugin via `require.resolve(...)`, which is precisely the yarn-PnP absolute-path resolution this PR set out to add.

You can confirm on current `master` in `packages/dashboard/vite/vite-plugin-lingui-babel.ts`:

```ts
const require = createRequire(import.meta.url);
// …
presets: [[require.resolve('@babel/preset-typescript'), …], [require.resolve('@babel/preset-react'), …]],
plugins: [require.resolve('@lingui/babel-plugin-lingui-macro')],
```

So the underlying issue (#3929 / the yarn build failure) is resolved on `master` already, and this PR now only conflicts with it. I'm going to close it as superseded — no reflection on your work, the approach was right. Really appreciate the contribution and hope to see more. 🙏

PR #4576, opened 2026-03-27
fix(core): Load order line variant translations in OrderService

yasserlens:
## What this fixes

Fixes #4566

This PR fixes two related translation-loading issues that can surface during order mutations.

### 1. `EntityHydrator` now detects missing nested relations across all array items

The intermittent failure in `addItemToOrder()` was caused by `EntityHydrator.getMissingRelations()` only checking the first element of array relations.

That broke cases like this:
- an order already has existing lines whose `productVariant.translations` are loaded
- a new line is appended in memory during order modification
- the new line's `productVariant` does not yet have `translations`
- a later `entityHydrator.hydrate(ctx, order, { relations: [...] })` call checks only `order.lines[0]`
- hydration incorrectly concludes that `lines.productVariant.translations` is already loaded
- the translate pass then iterates all lines and hits the new line without translations

This PR updates `EntityHydrator` so missing nested relations are detected across all members of array relations rather than only the first item.

The new traversal preserves existing semantics:
- `undefined` means the relation is missing and should be fetched
- `null` means the relation was loaded and is null in the database
- empty arrays are treated as already loaded, not missing

### 2. `OrderService` continues to guarantee variant translations for split line loading

This PR also keeps the `OrderService` hardening already added:
- normalize `lines.productVariant...` relation requests before the split line query runs
- ensure `lines.productVariant.translations` is loaded whenever line variant relations are requested
- ensure `lines.productVariant.product.translations` is loaded when nested product relations are requested
- add `lines.productVariant.translations` to `getOrderOrThrow()` defaults

Together, these changes fix the intermittent hydrator failure and close the separate relation gap in `OrderService`.

## Regression coverage

Added:
- unit coverage for `EntityHydrator.getMissingRelations()` with mixed loaded/missing array members
- unit coverage confirming empty loaded arrays are not treated as missing
- a targeted shop-order e2e regression using a shipping eligibility checker that hydrates:
  - `lines.productVariant.translations`
  - `lines.productVariant.product`

This reproduces the “existing line + newly added line” case and verifies the mutation now succeeds.


packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
```diff
@@ -204,36 +204,65 @@ export class EntityHydrator {
...
+        }
+        if (Array.isArray(entity)) {
+            if (entity.length === 0) {
+                return;
             }
+            for (const item of entity) {
+                this.collectMissingRelationsForPath(item, parts, partIndex, traversedPath, missingRelations);
+            }
+            return;
+        }
```
michaelbromley, line 238: ### Performance concern: missing short-circuit on array traversal

The old code only checked `entity[part][0]` — that was O(1) but **incorrect** (the whole reason for this fix). The new code correctly checks all elements, but it never short-circuits once a missing relation is found.

Consider that `EntityHydrator` is public API and plugin developers can use it on arbitrary entities. For a path like `productVariants.options.group.translations` on a `Collection` with thousands of variants, each with several options, the traversal becomes **multiplicative across nested arrays**: O(|variants| × |options|). And once we find the first element with a missing relation, there's no point checking the rest — hydration will re-fetch from DB for the entire entity anyway.

Suggested fix — bail as soon as a new missing relation is detected:

```ts
if (Array.isArray(entity)) {
    if (entity.length === 0) {
        return;
    }
    for (const item of entity) {
        const sizeBefore = missingRelations.size;
        this.collectMissingRelationsForPath(item, parts, partIndex, traversedPath, missingRelations);
        if (missingRelations.size > sizeBefore) {
            return;
        }
    }
    return;
}
```

This gives us the correctness of checking all elements (we don't silently miss the Nth element) with the performance of early exit (we stop as soon as we find the first gap). For nested arrays it turns worst-case O(N×M) into O(N+M).

michaelbromley:
@yasserlens thanks for the update - looks like one of the e2e tests is failing

grolmus:
Hi @yasserlens — thank you for this PR and the detailed analysis, and apologies that we missed it.

Unfortunately we didn't notice this PR in time and shipped a separate fix for #4566 in #4738 (now merged), which addresses the `OrderService` symptom by auto-injecting `lines.productVariant.translations` when callers pass explicit `relations`. Because of that, the `OrderService` portion of your PR overlaps with what's already on `master`.

The `EntityHydrator.getMissingRelations()` change is a genuine improvement — the "only first array element is checked" behaviour is a real edge case that #4738 doesn't address. However, that's a non-trivial refactor of a hot path that we'd rather review on its own merits rather than bundled with the (now-redundant) `OrderService` changes.

I'm going to close this PR for now. If you'd like to open a follow-up focused solely on the `EntityHydrator` fix (plus the dedicated unit tests, which are great), we'd happily review it. Again, sorry we didn't get to this sooner, and thanks for the careful work.

PR #4594, opened 2026-03-31
feat(core, dashboard): Support pluggable refund destinations

oliverstreissi:
# Description

Add `RefundDestinationStrategy` interface allowing plugins to define alternative refund destinations (store credit, gift cards, etc.) instead of only refunding to the original payment method.

**Backend:**

* New `RefundDestinationStrategy` interface with `isAvailable`/`createRefund` methods
* `refundDestinations` query to discover available destinations per order
* `destination` field on `RefundOrderInput` and `Refund` entity
* Routes `PaymentService.createRefund` through destination strategy when specified
* Startup validation rejecting strategies using reserved `default` code
* Uses refundable payment capacity (not just Settled state) for availability checks

**Dashboard:**

* Refund dialog split into "Payment methods" and "Other refund destinations" sections
* Independent allocation with payments prioritized before destinations
* Fix multi-target refund sending duplicate RefundLine records
* Remove orphaned `refundFragment` and dead utility functions

Fixes vendurehq/vendure#4563

# Breaking changes

None. Omitting `destination` in `RefundOrderInput` preserves existing behavior.

# Screenshots

https://github.com/user-attachments/assets/679711fc-2a5a-4508-bd94-3daaabe73559

# Checklist

📌 Always:

- [X] I have checked my own PR

👍 Most of the time:


packages/core/src/config/payment/refund-destination-strategy.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * @description
+     * Execute the refund to this destination. Called instead of
+     * `PaymentMethodHandler.createRefund()` when this destination is selected.
+     *
+     * The returned {@link CreateRefundResult} determines the refund state
+     * and any associated transaction ID or metadata.
+     */
+    createRefund(
```
michaelbromley, line 80: Will this not lead to duplication of logic between eg the StoreCreditPaymentMethodHandler and the StoreCreditRefundDestination?

Why do we not delegate this logic to the existing payment method handler?

michaelbromley (commented):
Dumping some notes from Claude here for when we revisit. 

## Architectural Review

Thanks for the solid work on this — the feature is clearly needed, and the strategy interface, e2e tests, and dashboard UI refactor are all well done. However, there are some architectural concerns worth discussing before merge.

### 1. Payment coupling for non-payment destinations

This is the main one. Every `Refund` is forced to reference a `Payment`, even when the destination (store credit, gift card) has nothing to do with the original payment method. The dashboard grabs `firstRefundablePaymentId` just to satisfy the FK — making `Refund.payment` semantically meaningless for destination refunds.

Both `createRefund()` and `isAvailable()` take a `Payment` parameter that's irrelevant for non-payment destinations. This will cause confusion for anyone querying refund data downstream.

**Suggestion:** Make `Refund.payment` nullable for destination-based refunds, and remove the `Payment` parameter from the strategy interface (pass `Order` only). If payment context is genuinely needed for some destinations, it could be optional.

### 2. No database migration

The PR adds `@Column({ nullable: true, type: 'varchar' }) destination` to the `Refund` entity but includes no migration file. This column won't exist until someone generates and runs a migration.

### 3. `isAvailable` inconsistency between query and mutation

`getRefundDestinations()` checks availability against ALL refundable payments (any match = available):

```ts
for (const payment of refundablePayments) {
    if (await strategy.isAvailable(ctx, order, payment)) {
        available = true;
        break;
    }
}
```

But `createRefund()` checks against the specific `paymentId` the dashboard sent — which for destinations is just `firstRefundablePaymentId`. This means the query could report a destination as available (because payment B passes), but the mutation rejects it (because it checks payment A).

### 4. Non-atomic split refunds

Multi-target refunds send sequential mutations from the dashboard. If the first succeeds and the second fails, you get a partial refund with no rollback. Only the first target gets `refundLines`, so subsequent Refund entities have no line references — which will cause issues in reporting and reconciliation.

**Suggestion:** Consider a single-mutation API that accepts multiple targets, so the server can handle this atomically.

### 5. Hardcoded constant duplication

```ts
// Must match DEFAULT_REFUND_DESTINATION_CODE from @vendure/common
const DEFAULT_REFUND_DESTINATION_CODE = 'default';
```

The CJS/ESM issue is understood, but this will go stale silently. Could the `refundDestinations` query mark which destination is default, avoiding the need for the dashboard to know the constant?

### Minor items

- No validation for duplicate strategy codes at startup
- `getRefundDestinations` lives on `OrderService` but deals with payment/refund concerns — `PaymentService` might be a more natural home
- ~60 lines of unrelated formatting changes (import reordering, ternary indentation) make the diff noisier than needed

Issue #4499, opened 2026-03-12
feat(core): Add CustomerChannelAssignmentStrategy to control automatic customer-channel assignment in AuthGuard


Draykee:
## Summary

The `AuthGuard` automatically assigns any authenticated customer to the channel indicated by the request's channel token header (`setActiveChannel()` at `packages/core/src/api/middleware/auth-guard.ts`). There is currently no mechanism to prevent or control this behavior.

In multi-channel setups where certain channels should be gated (e.g., community-specific channels, B2B channels with membership requirements, or partner channels requiring approval), this creates an access control gap: customers are silently assigned to channels — and gain visibility into channel-specific products, pricing, and promotions — simply by making a request with that channel's token.

## Current behavior

When an authenticated customer's request carries a channel token for a channel they are not yet assigned to, the AuthGuard unconditionally calls `channelService.assignToChannels()`:

```typescript
// auth-guard.ts, lines ~111-115
if (customer) {
    await this.channelService.assignToChannels(requestContext, Customer, customer.id, [
        requestContext.channelId,
    ]);
}
```

There is no hook, event, or strategy to intercept or prevent this assignment.

## Proposed solution: `CustomerChannelAssignmentStrategy`

Add a configurable strategy to `AuthOptions` that the AuthGuard consults before auto-assigning a customer to a channel. This follows the existing strategy pattern used throughout Vendure (`InjectableStrategy` with `init(injector)`).

### Strategy interface

```typescript
import { RequestContext } from '../../api/common/request-context';
import { Customer } from '../../entity/customer/customer.entity';
import { ID } from '../../common/types/common-types';
import { InjectableStrategy } from '../../common/types/injectable-strategy';

/**
 * Strategy that determines whether a Customer should be automatically
 * assigned to a channel when the AuthGuard detects they are not yet
 * a member of the active channel.
 */
export interface CustomerChannelAssignmentStrategy extends InjectableStrategy {
    canAssignCustomerToChannel(
        ctx: RequestContext,
        customer: Customer,
        channelId: ID,
    ): boolean | Promise<boolean>;
}
```

### Default implementation (preserves current behavior)

```typescript
export class DefaultCustomerChannelAssignmentStrategy implements CustomerChannelAssignmentStrategy {
    canAssignCustomerToChannel(): boolean {
        return true; // Always allow — same as current behavior
    }
}
```

### AuthGuard integration

```typescript
if (customer) {
    const shouldAssign = await this.configService.authOptions
[...]

PR #4601, opened 2026-04-01
feat(core):  Add CustomerChannelAssignmentStrategy for automatic channel assigment

harshit078:
# Description

- This PR addresses #4499

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:


gabriellbui:
Hey @harshit078! Thanks for the PR, really good starting point!

One thing I noticed in the below `if` condition is that if we are returning false here, I think it's semantically incorrect. 
`setActiveChannel` returns true if the active channel was set and false if not, but in this case we also say "we did not set the active channel" if our customer channel assignment strategy declined the auto assignment, but those should be two separate things. 

```
if (!shouldAssign) {
    return false;
 }
```

I'm planning on opening a new PR that would address it and will add tests and docs as well.

PR #4628, opened 2026-04-04
[RFC] feat(core): Add partitionKey to StockLevel for sub-location stock tracking

niko91i:
## Context

  This PR is a working proof-of-concept related to a discussion on Discord:
  https://discord.com/channels/1100672177260478564/1488690570296692909

  The current `StockLevel` entity enforces a unique constraint on `[productVariantId, stockLocationId]`,
  meaning only one stock record can exist per variant per location. This forces use cases like batch/lot
  tracking, serial number management, or expiration-based stock rotation to create one `StockLocation`
  per batch — polluting a table designed for physical locations.

  At scale (e.g. 10,000 vendors × 30 products × 6 batches/year = 1.8M rows/year), the `stock_location`
  table becomes unmanageable for its original purpose.

  ## Proposed solution

  Add an optional `partitionKey` column to `StockLevel`, allowing multiple stock records for the same
  variant+location pair, each identified by a unique key.

  - `partitionKey` defaults to `''` (empty string) — existing behavior is 100% unchanged
  - The unique index becomes `[productVariantId, stockLocationId, partitionKey]`
  - `LocationWithQuantity` gains optional `partitionKey` and `stockLevelId` fields
  - `StockLevelService` methods accept an optional `partitionKey` parameter
  - `StockMovementService` propagates `partitionKey` through all allocation/sale/release/cancellation flows

  ## Changes

  - `stock-level.entity.ts` — New `partitionKey` column (default `''`) + updated unique index
  - `stock-location-strategy.ts` — Optional `stockLevelId` and `partitionKey` on `LocationWithQuantity`
  - `stock-level.service.ts` — Optional `partitionKey` parameter on `getStockLevel`, `updateStockOnHand`, `updateStockAllocated`
  - `stock-movement.service.ts` — Propagate `partitionKey` through 5 call sites
  - GraphQL schema — `partitionKey` on `StockLevel` type and `StockLevelInput`

  ## What does NOT change

  - `StockLocation` entity — untouched
  - `StockLocationStrategy` interface — method signatures unchanged
  - `DefaultStockLocationStrategy` / `MultiChannelStockLocationStrategy` — no modifications
  - All existing applications without `partitionKey` work identically

  ## Test plan

  - [x] 822 existing unit tests pass (zero regression)
  - [x] 11 new unit tests for `StockLevelService` with/without `partitionKey`
  - [x] 5 new E2E tests verifying partition creation, coexistence, and targeted updates
  - [x] 59 existing stock E2E tests pass (stock-control, stock-location, multi-location)


niko91i:
Tested this PR against a production marketplace codebase via patch-package. Batch/lot tracking with multiple StockLevels per variant+location works as expected. Also added partitionKey propagation to StockMovement entities for full traceability.

packages/core/src/service/services/stock-movement.service.ts
```diff
@@ -115,13 +116,15 @@ export class StockMovementService {
                     quantity: delta,
                     stockLocation: { id: input.stockLocationId },
                     productVariant: { id: productVariantId },
+                    partitionKey: input.partitionKey ?? '',
```
DanielBiegler: this is a bit nitpicky _(sorry)_ because i do understand that once the default partitionkey exists it generally wont change, but i'd really like to see some form of constant like `DEFAULT_STOCK_LOCATION_PARTITION_KEY` (or something shorter) because the current default of `""` is being reused throughout many different files, comments, services, tests, .. and having a named constant makes it easier to understand unfamiliar parts of code intuitively

other than that really cool PR! 🙏

niko91i:
Thanks for the review @DanielBiegler ! I'm already using this pattern in production via patch-package for a marketplace with batch/lot tracking. Before I build further on it, is this approach likely to be accepted (with potential refinements), or would you recommend a different direction?

DanielBiegler:
> is this approach likely to be accepted

Can't say, gotta wait for a comment from the maintainers. I've not looked into this topic deeply, so I don't have informed opinions specifically there, but the pr looks generally well made from my short glance!

I think a research section comparing how other ecom frameworks deal with this task, can help.

Generally I'd say keeping track of batches/lots sounds reasonable to support because Vendure aims to be enterprisy/B2B, where this kind of stuff happens throughout different types of businesses

Ryrahul:
Nice work on this PR — the problem is well-defined, the entity/service plumbing is clean, and the test coverage is nice. Just wanted to flag one issue 


`BaseStockLocationStrategy.getLocationsBasedOnAllocations` reads `Allocation` records (which now correctly store `partitionKey`) but:

- Does NOT include `partitionKey` in the returned `LocationWithQuantity`
- Groups only by `stockLocationId`

This causes allocations from different partitions at the same location to get merged.

Since `forSale`, `forRelease`, and `forCancellation` all delegate to this method, the partition context is lost across the order lifecycle.

##  Impact

In a normal order lifecycle:

- Checkout allocates from the correct partition (e.g. `BATCH-001`)
- Fulfillment / cancellation / release default to the **default partition**

Result:

- Allocated partition gets stuck
- Default partition receives incorrect (phantom) adjustments
- Inventory becomes inconsistent

##  Extensibility Issue

- `getLocationsBasedOnAllocations` is private
- Custom strategies cannot override it

To fix behavior, users would need to reimplement:

- `forSale`
- `forRelease`
- `forCancellation`

This is unnecessarily complex and error-prone.

## Suggested Fix

Update `getLocationsBasedOnAllocations` to:

- Group by `stockLocationId` and `partitionKey`
- Include `partitionKey` in `LocationWithQuantity`

This ensures partition context is preserved across the full order lifecycle.

## Minor Issue

In `StockLocationService.deleteStockLocation`:

- Stock transfers are matched without `partitionKey`

**Risk:**
Different batches/partitions may get merged incorrectly during deletion.

***I’m not a maintainer, so probably its best to wait for a maintainer’s input on this***

Ryrahul:
Worth noting the dashboard doesn't query or send partitionkey yet - admins would see
duplicate rows per location with no way to tell them apart. Probably fine as a follow-up
just flagging it.

michaelbromley:
Hi @niko91i, thanks for taking the time to share your solution here.

I've not got time for a deep-dive review on this right now, as it is quite a complex topic touching many parts of the inventory & order flow.

I just wanted to let you know it's on my radar and might be something we are interested in if it aligns with our overall strategy in terms of what features we want to work on in the coming months.

Sorry I can't be more specific than that right now.

michaelbromley:
Thanks for the continued work on this and for responding so quickly to feedback. I’ve now done a deep review. Summary of my thoughts:

**Review**

It appears there's an issue with the implementation in this PR: `DefaultStockLocationStrategy` is partition-blind. `getAvailableStock` sums stock across all partitions, but `forAllocation` only ever targets the default ('') partition. Since `StockLevelInput.partitionKey` is exposed to every admin API user, any merchant on the default strategy can get into this state without a custom strategy.

Concrete failure: put 50 units in BATCH-001 only. Saleable stock reports 50. An order for 50 allocates against the '' partition, which doesn’t exist, so updateStockAllocatedForLocation silently no-ops. On fulfillment, the sale creates a default-partition row with negative stockOnHand while BATCH-001 still shows 50. No error is raised anywhere.

The existing e2e tests create partitions but never place an order, so this path is untested. Any version of this feature would need full order-lifecycle coverage (allocate → sell → cancel/release) against partitioned stock.

The change from Math.min(quantity, quantityAvailable) to Math.min(quantity - totalAllocated, quantityAvailable) in MultiChannelStockLocationStrategy **fixes a real pre-existing over-allocation bug on master**, independent of this feature. Please extract that into its own PR so it can ship regardless of what happens here — it’s a valuable fix.

**Direction**

Having thought this through, I'm not certain of the partitionKey design going into core in this form. A bare string can’t carry the metadata real batch tracking needs (expiry dates for FEFO, lot references), so users would end up encoding data into the key. My suspicion is that I would need a bit more domain knowledge to judge a better solution.

I wonder if the same use case could be achievable today as a plugin, with no core changes: a `StockBatch` entity owned by the plugin, a custom StockLocationStrategy whose getAvailableStock reports only non-expired batch stock, and batch selection done FEFO at fulfillment time inside a custom FulfillmentHandler. Core StockLevel stays as the aggregate.

PR #4636, opened 2026-04-09
perf(core): Add batch loading for collection product variants

RubenDarioGuerreroNeira:
## Description

### DIFF
--- a/e2e-common/benchmarks/benchmark-list-query.e2e-spec.ts
+++ b/e2e-common/benchmarks/benchmark-list-query.e2e-spec.ts
@@ -1,0 +1,606 @@
+/* eslint-disable no-console */
+import { FacetValue, VendureConfig } from '@vendure/core';
+import { createTestEnvironment } from '@vendure/testing';
+import { gql } from 'graphql-tag';
+import path from 'path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import { initialData } from '../e2e-initial-data';
+import { testConfig } from '../test-config';
+
+describe('ListQueryBuilder Optimization Benchmark', () => {
+    let capturedQueries: string[] = [];
+    const baseConfig = testConfig();
+
+    const benchmarkConfig: VendureConfig = {
+        ...baseConfig,
+        customFields: {
+            Product: [
+                {
+                    name: 'testManyToMany',
+                    type: 'relation',
+                    entity: FacetValue,
+                    graphQLType: 'FacetValue',
+                    list: true,
+                },
+                {
+                    name: 'testManyToOne',
+                    type: 'relation',
+                    entity: FacetValue,
+                    graphQLType: 'FacetValue',
+                    list: false,
+                },
+            ],
+        },
+        dbConnectionOptions: {
+            ...baseConfig.dbConnectionOptions,
+            logging: ['query'],
+            logger: {
+                logQuery(query: string) {
+                    if (
+                        query.includes('SELECT') &&
+                        (query.includes('"product"') || query.includes('`product`'))
+                    ) {
+                        capturedQueries.push(query);
+                    }
+                },
+                logQueryError: (error: string) => console.error(error),
+                logQuerySlow: (time: number, query: string) => console.warn(query, time),
+                logSchemaBuild: () => {
+                    /* no-op */
+                },
+                logMigration: () => {
+                    /* no-op */
+                },
+                log: () => {
+                    /* no-op */
+                },
+            } as any,
+        },
+    };
+
+    const { server, adminClient } = createTestEnvironment(benchmarkConfig);
+
+    // Store IDs of created entities for concrete assertions
+    let plantsCollectionId: string;
+    let electronicsCollectionId: string;
+    let laptopVariantId: string;
+
+    beforeAll(async () => {
+        await server.init({
+            initialData,
+            productsCsvPath: path.join(__dirname, '../../packages/core/e2e/fixtures/e2e-products-minimal.csv'),
+            customerCount: 1,
+        });
+        await adminClient.asSuperAdmin();
+
+        // ── Fetch existing data IDs ──────────────────────────────────
+        const GET_INITIAL_DATA = gql`
+            query GetInitialData {
+                collections(options: { take: 10 }) {
+                    items { id name }
+                }
+                productVariants(options: { take: 10 }) {
+                    items { id name }
+                }
+            }
+        `;
+        const initialResult: any = await adminClient.query(GET_INITIAL_DATA);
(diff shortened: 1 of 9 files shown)


michaelbromley:
Hi, before I can properly review this, it needs to be cleaned up to include _only_ the changes related to the fix.

RubenDarioGuerreroNeira:
Hi @michaelbromley, I have cleaned up the PR as requested. I removed all unrelated changes, including minor
  formatting tweaks, accidental config files, and internal tool settings. The PR now only includes the core logic
  for the N+1 optimization of collection variants and the catalog filter persistence, along with the corresponding
  benchmark test. Ready for your review

grolmus (changes requested):
Thanks for the work here, @RubenDarioGuerreroNeira — the collection-variants N+1 is a real problem and the batching approach is sound: storing the bulk promise in `RequestContextCache` and awaiting it from the field resolvers is proper per-request batching, not a hidden N+1, and the raw-alias/`getRawAndEntities()` mapping and channel/soft-delete scoping all look correct.

That said, I can't take this as-is — it's effectively two changes in one, and one of them changes behaviour well beyond collections. A few things need addressing:

### 1. Please split the shared filter-builder change out
The one-liner in `parse-filter-params.ts` (`|| relationMetadata.isManyToOne`) plus the new ManyToOne EXISTS branch and JOIN-skip in `list-query-builder.ts` reroute **every ManyToOne custom-property filter in the system** through the new machinery — not just collections. Live consumers include the admin `orders` list (`customerLastName` → `Order.customer`, a nullable ManyToOne) and `taxRates` (`zone.id`, `category.id`). That's a high-blast-radius change to shared infra and should be its own PR ("route ManyToOne custom-property filters through EXISTS"), reviewed and tested on its own merits — otherwise it rides in on the strength of the collection fix, which isn't where most of the risk is.

### 2. Pagination in `getProductVariantsForCollections` is silently broken
`{ take: undefined, skip: undefined }` does **not** mean "unlimited" — `parseTakeSkipParams` maps `take: undefined` to `adminListQueryLimit`. So the bulk query caps the **combined** variant rows across **all** collections at one global limit, with **no `ORDER BY`**. In the default admin collection-list grid this means later collections get truncated/empty variant lists, and the resolver's `totalItems: variants.length` becomes a per-collection undercount that diverges from the accurate count on the non-cached path. Please paginate per-collection properly, or explicitly bound this as a capped preview and make `totalItems` honest (fall back to the count query).

### 3. ManyToOne `isNull` semantics change
The old path LEFT-JOINed, so `customerLastName: { isNull: true }` matched orders with a null lastName *or no customer at all*. The EXISTS form only matches orders that have a customer whose lastName is null. Please decide the intended semantics and pin it with a test.

### 4. Multi-hop ManyToOne paths
`getToManyRelationCustomProperties` marks a property as EXISTS whenever `pathParts[0]` is ManyToOne for **any** path length, but `buildExistsSubquery` hardcodes `pathParts[1]` as the column. For a 3-part path (`a.b.c`, `a` M2O) it emits SQL referencing the intermediate relation as a column. No core usage today, but this is public infra plugins extend — either handle depth > 2 or only route 2-part M2O paths through EXISTS.

### 5. Tests
Only a benchmark is added, which asserts nothing about correctness. For a change of this reach please add: M2O filter equivalence on `orders`/`taxRates` (incl. the `isNull` case from #3), the batch method's grouping (a variant in multiple collections; the >limit truncation case; empty input), and the new unbuildable-EXISTS `throw` path.

### Minor
- `getProductVariantCounts` returns `Map<ID, number>` while the new method returns `Map<string, ProductVariant[]>` — reconcile the key type.
- A few comments restate the code (`// High-performance mapping … O(1)`, `// Using an explicit alias … to avoid fragility`) — the code already says it; drop them.
- The two JOIN-skip guards use different length checks (`relationPath.length === 1` vs `parts.length === 2`) for the same idea.
- Unrelated, spotted in passing: `tax-rate.service.ts` sets `customPropertyMap.zoneId` in both the zone and category branches — the category one looks like it should be `categoryId`. Not for this PR, just flagging.

The batching core is genuinely good — splitting the infra change out and tightening the pagination/tests would get this moving.

RubenDarioGuerreroNeira:
Thanks for reviewing this PR. I aimed to reduce the collection variants N+1 query pattern while keeping the behavior consistent with the existing list-query loading flow.

RubenDarioGuerreroNeira (commented):
The changes for Issues 1–4 were applied, and test coverage was added to the existing benchmark.

Final commit: `c11b58f85` (amend). The branch was pushed to `RubenDarioGuerreroNeira:fix/collection-variants-n-plus-1-v3` using `--force-with-lease`.

Note: I was unable to run the tests in this environment due to configuration issues; please validate locally using the repo's E2E runner and let me know if you need help reproducing the environment or the execution commands.

grolmus:
Thanks for the update, @RubenDarioGuerreroNeira. I had a look at the reworked commit, but the main blockers from my review are still open, so I can't move this forward yet:

- **Split the shared filter-builder change out (#1).** `parse-filter-params.ts` and `list-query-builder.ts` are still in the diff, so this still reroutes **every** ManyToOne custom-property filter in the system (admin `orders` → `customerLastName`, `taxRates` → `zone.id`/`category.id`, plus anything plugins add) through the new EXISTS machinery. That's the highest-risk part and it needs to be its own PR ("route ManyToOne custom-property filters through EXISTS"), reviewed on its own merits. The collection PR should depend only on the batch-loading.

- **Pagination (#2).** `getProductVariantsForCollections` still passes `{ take: undefined, skip: undefined }`, which `parseTakeSkipParams` maps to `adminListQueryLimit` — a single global cap across all collections' variants, with no `ORDER BY`. Later collections still get truncated and `totalItems: variants.length` is still a per-collection undercount. The added test comments describe this behaviour rather than fixing it — please paginate per-collection, or bound it explicitly as a capped preview and make `totalItems` honest via the count query.

- **Tests (#3, #5).** Still only the benchmark, which asserts nothing about correctness. For a change with this reach I need real assertions: M2O filter equivalence on `orders`/`taxRates` (including the `isNull` semantics change from #3), the batch method's grouping (a variant in multiple collections, the >limit truncation case, empty input), and the unbuildable-EXISTS `throw` path.

The batching core is still good — if you split the infra change into its own PR and tighten the pagination + tests on this one, it'll move much faster. Happy to review the split-out PR in parallel.

RubenDarioGuerreroNeira:
Closing this PR to reopen with the cleaned branch. The ManyToOne EXISTS infrastructure has been extracted into PR #5034. The new PR will contain only the collection batch loading changes.

PR #4652, opened 2026-04-14
fix(dashboard): Handle empty customFields selection when all fields h…

### LINKED ISSUE #4650: Dashboard Order detail page crashes when all Order custom fields have ui: { dashboard: false }
**Describe the bug**

When all custom fields added to the `Order` entity have `ui: { dashboard: false }` (i.e., hidden from the Dashboard), the Admin Dashboard's Order detail page throws a GraphQL validation error and fails to load.

**To Reproduce**

1. Create a Vendure plugin that adds one or more custom fields to the `Order` entity, all with `ui: { dashboard: false }`
2. Start the Vendure server and open the Dashboard
3. Navigate to **Orders** and open any order
4. The page fails to render

**Expected behavior**

The Order detail page should load normally. Custom fields flagged as `ui: { dashboard: false }` should be excluded from the UI without causing a broken query.

**Actual behavior**

The page throws an error and does not render:

```
GraphQL Request Error: Field "customFields" of type "OrderCustomFields" must have a selection of subfields. Did you mean "customFields { ... }"
```

**Screenshots/Videos**

**Error logs**

```
GraphQL Request Error: Field "customFields" of type "OrderCustomFields" must have a selection of subfields. Did you mean "customFields { ... }"
```

**Root cause analysis**

Adding any custom field to `Order` converts `customFields` in the schema from a JSON scalar to the `OrderCustomFields` object type, requiring explicit subfield selection. The `OrderDetail` fragment in the Dashboard hardcodes a bare `customFields` field, which is then enriched by the `addCustomFields()` utility at runtime using the list of custom fields returned by `entityCustomFields` from the server config.

However, fields configured with `ui: { dashboard: false }` are excluded from `entityCustomFields`. When **all** Order custom fields are hidden this way, `addCustomFields()` receives an empty list for the `Order` entity and skips rewriting the `customFields` selection — leaving the bare `customFields` in place. Since `customFields` is now an object type (not a scalar), this produces an invalid GraphQL query.

The bug is in `src/lib/framework/document-introspection/add-custom-fields.ts`: when an existing bare `customFields` field is found in a fragment but there are no visible custom fields to add (empty `selectionNodes`), the field should be **removed** from the selection rather than left as a bare reference.

- `@vendure/core` version: `^3.6.2`
- `@vendure/dashboard` version: `^3.6.2`
- Node.js version: `v22.16.0`
- Database: PostgreSQL
- Operating System: Windows
- Package manager: npm

**Configuration**

```typescript
// Plugin configuration — all Order custom fields have ui: { dashboard: false }
    name: 'offerType',
    type: 'string',
    nullable: true,
    public: true,
    ui: { dashboard: false },  // <-- hidden from dashboard
});

    name: 'sourceOfferId',
    type: 'int',
    nullable: true,
    public: true,
    ui: { dashboard: false },  // <-- hidden from dashboard
});
```

**Minimal reproduction**

1. Add at least one custom field to `Order` via a plugin
2. Set `ui: { dashboard: false }` on **every** Order custom field
3. Start the server and navigate to any Order in the Dashboard

**Workaround**

Remove `ui: { dashboard: false }` from at least one Order custom field, or patch `orders.graphql.ts` in `node_modules/@vendure/dashboard` to remove the bare `customFields` from the `OrderDetail` fragment and let `addCustomFields()` add it only when needed.

**Additional context**

The underlying `addCustomFields()` function (in `document-introspection/add-custom-fields.ts`) correctly handles the case when fields exist — it rewrites `customFields` to `customFields { field1 field2 }`. The missing case is when the entity *does* have custom fields in the schema but none are visible in the Dashboard: the bare `customFields` selection must be removed rather than left unrewritten.

Ryrahul:
# Description

Fixes https://github.com/vendurehq/vendure/issues/4650

When all custom fields of an entity have `ui: { dashboard: false }`, the dashboard's GraphQL queries include a bare `customFields` field with no subfield selections, causing a GraphQL error:

> Field "customFields" of type "OrderCustomFields" must have a selection of subfields. Did you mean "customFields { ... }"?

This happens because many GraphQL fragments (e.g. `orderDetailFragment`) contain a bare `customFields` placeholder that `addCustomFields()` is supposed to expand with the specific subfields. When all fields are filtered out, the bare `customFields` was left as-is, producing invalid GraphQL.

**Root cause:** Two gaps in the custom fields handling:

1. **Dashboard (`add-custom-fields.ts`):** `applyCustomFieldsToSelection()` did not remove pre-existing bare `customFields` fields when there were no visible custom fields to select. When `customFieldsForType` is empty or missing from the map, the main `if` block is skipped but the bare `customFields` in the fragment remains, causing the GraphQL error.

2. **Server (`global-settings.resolver.ts`):** `generateEntityCustomFieldConfig()` returned entities with empty `customFields` arrays when all fields had `dashboard: false`, instead of omitting them entirely.

**Fix:**

- In `applyCustomFieldsToSelection()`: added an `else if` branch that detects and removes bare `customFields` fields when the entity has no visible custom fields in the map (via `splice` on the selections array, consistent with existing in-place mutation pattern used throughout the function).
- In `generateEntityCustomFieldConfig()`: skip entities where all custom fields were filtered out after applying `internal` and `dashboard: false` filters.

> **_Note:_** _The same issue can also occur inside the main `if` block when the `includeCustomFields` option filters out all fields (resulting in an empty `selectionNodes` array that gets pushed as `customFields { }`). However, `includeCustomFields` is currently not used anywhere in the dashboard app — it only exists as an API for potential dashboard extensions. Happy to add a guard for that path as well if needed._

# Breaking changes

None. This is a bug fix. Custom fields with `ui: { dashboard: false }` now work correctly when applied to all fields of an entity.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/api/resolvers/admin/global-settings.resolver.ts
+++ b/packages/core/src/api/resolvers/admin/global-settings.resolver.ts
@@ -170,6 +170,11 @@ export class GlobalSettingsResolver {
                         }
                         return customFieldConfig;
                     });
+                if (!customFieldsConfig.length) {
+                    // All custom fields were filtered out (e.g. all marked as internal
+                    // or hidden from the dashboard), so skip this entity entirely.
+                    return;
+                }
                 return { entityName: entityType, customFields: customFieldsConfig };
             })
             .filter(notNullOrUndefined);
--- a/packages/dashboard/src/lib/framework/document-introspection/add-custom-fields.spec.ts
+++ b/packages/dashboard/src/lib/framework/document-introspection/add-custom-fields.spec.ts
@@ -756,6 +756,90 @@ describe('addCustomFields()', () => {
             expect(printed).toContain('vatNumber');
         });
 
+        // https://github.com/vendurehq/vendure/issues/4650
+        // When all custom fields have ui.dashboard: false, the server returns an empty array
+        // for that entity. The dashboard should remove the bare `customFields` field from the
+        // query to avoid a GraphQL error ("must have a selection of subfields").
+        it('Should remove bare customFields field when entity has no custom fields in the map', () => {
+            const documentNode = graphql(`
+                query GetOrder($id: ID!) {
+                    order(id: $id) {
+                        id
+                        code
+                        customFields
+                    }
+                }
+            `);
+
+            // Simulate: Order entity exists in the map but with no custom fields
+            // (all were filtered out due to ui.dashboard: false)
+            const customFieldsConfig = new Map<string, CustomFieldConfig[]>();
+            customFieldsConfig.set('Order', []);
+
+            const result = addCustomFields(documentNode, {
+                customFieldsMap: customFieldsConfig,
+            });
+            const printed = print(result);
+
+            // customFields should be removed entirely to avoid the empty selection set error
+            expect(printed).not.toContain('customFields');
+            expect(printed).toContain('id');
+            expect(printed).toContain('code');
+        });
+
+        // https://github.com/vendurehq/vendure/issues/4650
+        it('Should remove bare customFields field when entity has no custom fields in the map at all', () => {
+            const documentNode = graphql(`
+                query GetOrder($id: ID!) {
+                    order(id: $id) {
+                        id
+                        code
+                        customFields
+                    }
+                }
+            `);
+
+            // Simulate: Order entity is not in the map at all
+            const customFieldsConfig = new Map<string, CustomFieldConfig[]>();
+
+            const result = addCustomFields(documentNode, {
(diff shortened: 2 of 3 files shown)


grolmus (approved):
Clean fix. Two complementary changes:
1. Dashboard removes bare `customFields` field when no visible custom fields exist → prevents GraphQL error
2. Backend skips entities with 0 (visible) custom fields from schema info response

Good test coverage with 4 cases covering both `addCustomFields()` and `addCustomFieldsToFragment()`.

michaelbromley (approved):

Issue #1508, opened 2022-04-06
Found a bug when product is deleted with existing records of orders


bbtronald:
**Customer tries to AddToCart proceed upto Checkout.. Check the GraphQL**

```graphql 
activeCustomer {
    orders(options: $options) {
      items {
        id
        code
        updatedAt
        orderPlacedAt
        lines {
            id
            quantity
            featuredAsset{
                source
            }
            items {
                cancelled
            }
            productVariant {
                id
               price
                product {
                    id
                    name
                }
            }
        }
        code
        state
        currencyCode
        totalWithTax
        totalQuantity
        shippingAddress {
            streetLine1
            streetLine2
            city
        }
        customer {
            firstName
            emailAddress
        }
      }
      totalItems
      __typename
    }
    __typename
}
```
__

ProductVariant > price  and priceWithTax gives error below
`"message": "No ProductVariant with the id '5f6f57b8-a88c-4c14-9893-dfb6663d9f42' could be found",`

But when it remove on the graphql everything is fine.

PR #4653, opened 2026-04-15
fix(services): findByCustomerId productVariant relation filter breaks

harshit078:
# Description

- This PR address issue #4560 
- the logic implemented here is to revert to passing relations through directly and drop the unused filter.

# Breaking changes

No

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/services/order.service.ts
+++ b/packages/core/src/service/services/order.service.ts
@@ -327,15 +327,9 @@ export class OrderService {
         options?: ListQueryOptions<Order>,
         relations?: RelationPaths<Order>,
     ): Promise<PaginatedList<Order>> {
-        const effectiveRelations = (relations ?? ['lines', 'customer', 'channels', 'shippingLines']).filter(
-            r =>
-                // Don't join productVariant because it messes with the
-                // price calculation in certain edge-case field resolver scenarios
-                !r.includes('productVariant'),
-        );
         return this.listQueryBuilder
             .build(Order, options, {
-                relations: effectiveRelations,
+                relations: relations ?? ['lines', 'customer', 'channels', 'shippingLines'],
                 channelId: ctx.channelId,
                 ctx,
             })
--- a/packages/core/e2e/customer.e2e-spec.ts
+++ b/packages/core/e2e/customer.e2e-spec.ts
@@ -1,7 +1,13 @@
 import { DeletionResult, ErrorCode, HistoryEntryType } from '@vendure/common/lib/generated-types';
 import { omit } from '@vendure/common/lib/omit';
 import { pick } from '@vendure/common/lib/pick';
-import { AccountRegistrationEvent, EventBus } from '@vendure/core';
+import {
+    AccountRegistrationEvent,
+    ConfigService,
+    EventBus,
+    OrderService,
+    RequestContextService,
+} from '@vendure/core';
 import { createErrorResultGuard, createTestEnvironment, ErrorResultGuard } from '@vendure/testing';
 import path from 'path';
 import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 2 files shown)


packages/core/src/service/services/order.service.ts
```diff
@@ -327,15 +327,9 @@ export class OrderService {
         options?: ListQueryOptions<Order>,
         relations?: RelationPaths<Order>,
     ): Promise<PaginatedList<Order>> {
-        const effectiveRelations = (relations ?? ['lines', 'customer', 'channels', 'shippingLines']).filter(
-            r =>
-                // Don't join productVariant because it messes with the
```
Draykee, line 332: This filter got introduced for a reason: 
https://github.com/harshit078/vendure/commit/b98220b6ae191dfc1bdbd9b6b6b418e6be9ed075

But at this point I guess only god knows why, or @michaelbromley

michaelbromley:
Thanks for tracking this down @harshit078! I dug into the history of this filter because the rationale wasn't obvious, and it turns out to be a proper saga worth documenting here for future reference:

## Timeline of the `productVariant` filter

**1. April 2022, commit `5061dd9`** — Fix for #1508 (deleted variants throwing entity-not-found errors). `lines.productVariant` was removed from the default relations and lazy-loading was delegated to `OrderLineEntityResolver.productVariant`.

**2. April 2022, commit `b98220b6a`** — I added the `effectiveRelations` filter with the comment *"Don't join productVariant because it messes with the price calculation in certain edge-case field resolver scenarios"*. **However, the filter was dead code** — `effectiveRelations` was computed but never actually passed to `build()`. It sat dormant for nearly 4 years without anyone noticing.

**3. January 2026, PR #4118** — The dead code was spotted and "activated" by wiring `effectiveRelations` into the `build()` call. This is when the regression actually entered the codebase.

**4. Now (#4560 / this PR)** — Service-layer callers explicitly passing `productVariant` in their relations get silent data loss, plus N+1 queries via `CustomerEntityResolver.orders` when GraphQL queries request `lines { productVariant { ... } }`.

## The likely original rationale

`OrderLineEntityResolver.productVariant` has an early-return path:

```ts
if (orderLine.productVariant) return orderLine.productVariant; // raw entity
return this.productVariantService.getVariantByOrderLineId(ctx, orderLine.id); // applies channel price + translation
```

If `productVariant` is pre-loaded via a TypeORM join, the field resolver returns the raw entity without running `applyChannelPriceAndTax()` or translation. That was almost certainly the "edge-case" the 2022 filter was guarding against — incorrect multi-channel pricing on the joined entity.

The deleted-variant concern from #1508 is no longer relevant here, since `getVariantByOrderLineId` handles it via `includeSoftDeleted: true`.

## Request: regression test

Given this code has now bounced back and forth three times across four years, I'd really like a regression test before merging. Specifically:

- Call `OrderService.findByCustomerId` with `lines.productVariant` (and ideally `lines.productVariant.product`) in the relations
- Assert the variant actually loads (catches the silent data loss)
- Assert correct channel price and translation are applied (catches the original 2022 concern, if it's still real)

It would also be worth verifying manually on a multi-channel setup that prices come through correctly when the variant is loaded via the join, since that's the one risk this change introduces that the test history can't fully settle.

Otherwise the change itself looks sensible — appreciate the work to diagnose this!

grolmus:
Hi @harshit078 — Vendure core maintainer here. Picked up this PR during triage of OSS-465. Since the fix has been blocked on "is the filter actually needed?", I did a deeper investigation and pushed two follow-up commits directly to your branch (you have `maintainerCanModify` enabled). Hope that's OK — your fix commit is preserved with your authorship.

## What I added

1. **Rebased your fix onto current master** — no content change, just brings the branch up to date.
2. **`test(core): Add regression test for findByCustomerId relations`** — new e2e test in `customer.e2e-spec.ts` covering the exact bug described in #4560: calls `OrderService.findByCustomerId` with `['lines', 'lines.productVariant', 'lines.productVariant.product']` and asserts `productVariant` is loaded. Verified by re-applying the filter locally — without your fix the test fails on `expect(result.items.length).toBe(1)`.

## Why the filter is safe to remove (addressing @Draykee's concern)

The original "messes with the price calculation in certain edge-case field resolver scenarios" comment is stale. I traced the resolver wiring:

- `OrderLineEntityResolver.productVariant` returns the eagerly-loaded variant as-is when present.
- But `ProductVariantEntityResolver` handles `price` / `priceWithTax` / `currencyCode` / `taxRateApplied` via `productVariantService.hydratePriceFields()` — which lazy-loads `productVariantPrices` + `taxCategory` if missing and applies `applyChannelPriceAndTax` with a per-request cache. Translations go through `LocaleStringHydrator.hydrateLocaleStringField`.
- Net effect: GraphQL queries against eagerly-loaded variants get correct prices/translations regardless of how the variant was loaded. The pricing concern that motivated the filter is solved at the field-resolver layer, so the filter is genuinely dead code.

Soft-deleted variants are also fine — `ProductVariant.deletedAt` is a plain nullable `@Column` (not `@DeleteDateColumn`), so TypeORM doesn't auto-filter, matching the existing `findOne` and `getVariantByOrderLineId` semantics.

## On the CodeRabbit ask

This is the regression test it asked for. It's in `customer.e2e-spec.ts` rather than near other `OrderService` tests because the existing seed/fixture flow there already creates the test order via `addItemToOrder` — placing it elsewhere would duplicate setup.

## Note on commit message

I left your `fix(services):` commit untouched. A core maintainer code review (@nigel review with @michaelbromley as reference) flagged that the scope should be `core` per repo convention, but I didn't want to rewrite your authored commit. Up to you / Michael whether to amend at merge time.

Thanks again for the fix — sorry it sat for a month. From my side this is ready for merge. cc @michaelbromley

harshit078:
Hey @michaelbromley and @grolmus , I really appreciate the such highly detailed analyses and remarks. I did see the comment regarding regression test but I forgot to push the fix for it hence it was left unattended from my side. @grolmus  I completely agree with your approach and really appreciate you taking over. Thanks !

grolmus (approved):
Coming back to formally approve — the fix, the regression test, and the saga analysis are all in place from the earlier round.

Recap for any future reader of this PR:

- The filter being removed was dead code in master for ~4 years before #4118 wired it up. The "messes with price calculation" comment is stale — `OrderLineEntityResolver.productVariant` early-returns the eager variant, and `ProductVariantEntityResolver` lazy-loads `productVariantPrices` + `taxCategory` via `hydratePriceFields` regardless, so dropping the filter does not change price-calc behaviour.
- The regression test in `customer.e2e-spec.ts` (`findByCustomerId honors explicitly requested productVariant relations`) calls `OrderService.findByCustomerId` with `['lines', 'lines.productVariant', 'lines.productVariant.product']` and asserts the relations are loaded. I re-applied the filter locally during the earlier round and confirmed the test fails without the fix.

### CI

The `e2e (postgres) (22.x)` failure on the current head is unrelated to this PR — it's an assertion mismatch in `e2e/order-modification.e2e-spec.ts:1515` ("reduce quantity to create second refund", `total: 17249` vs expected `16649`). mariadb / mysql / sqljs all pass on the same commit, and the failing test exercises payment refund logic that doesn't touch `findByCustomerId`. Looks like a known postgres-only flake. Worth a re-run rather than blocking on it.

Approving. Thanks again @harshit078, and @michaelbromley for the historical dig.

michaelbromley (approved):

PR #4370, opened 2026-02-17
fix(core): Use tax-inclusive amount when cancelling shipping with pricesIncludeTax

BibiSebi:
## Summary

- When cancelling an order with `cancelShipping=true` and `pricesIncludeTax=true`, the shipping cancellation adjustment used the tax-exclusive `discountedPrice` instead of the tax-inclusive `discountedPriceWithTax`
- This left a remainder equal to the tax amount on the shipping cost
- Now uses `discountedPriceWithTax` when `listPriceIncludesTax=true` so the adjustment fully zeroes out shipping

## Test plan

- [x] Added e2e test: cancels order with `pricesIncludeTax=true` and verifies `shipping`, `shippingWithTax`, and `totalWithTax` are all 0
- [x] Existing `cancellation by orderId` tests still pass

## Manual verification

Tested locally by enabling `pricesIncludeTax` on the default channel and adding tax to the Express Shipping method, then cancelling an order with shipping.

Screenshot 2026-05-21 at 17 14 02

### Before
Screenshot 2026-05-21 at 17 10 53

### After

Screenshot 2026-05-21 at 17 13 29

Fixes #4348

PR #4659, opened 2026-04-16
fix(core): Fix shipping tax calculation in cancelShipping

dev-ish:
# Description
Calling `cancelOrder` with `cancelShipping: true` on a channel where `pricesIncludeTax: true` leaves tax residue on the shipping line instead of fully zeroing it out. The order's `totalWithTax` is non-zero after "cancellation".

The bug manifests whenever the shipping line is tax-inclusive — either because the `ShippingMethod`'s `includesTax` is `include`, or because it's `inherit` and the channel's `pricesIncludeTax` is `true`. The same code path is used by `refundOrder` via order modifications, so that flow is affected too.

**Example:** 
shipping = `10 SEK incl. / 8 SEK excl.` (25% tax). 
After cancellation: discount `-8 SEK incl. / -6.40 SEK excl.` and shipping line `2 SEK incl. / 1.60 SEK excl.` remain — should be all zero.

# Breaking changes
No breaking changes.  

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/order-modifier/order-modifier.ts
+++ b/packages/core/src/service/helpers/order-modifier/order-modifier.ts
@@ -349,7 +349,9 @@ export class OrderModifier {
                     adjustmentSource: 'CANCEL_ORDER',
                     type: AdjustmentType.OTHER,
                     description: 'shipping cancellation',
-                    amount: -shippingLine.discountedPrice,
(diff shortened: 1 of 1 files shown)


grolmus:
Closing as duplicate of #4370 which has the identical fix plus an e2e test.

PR #4660, opened 2026-04-17
fix(core): Prevent coupon usage limit bypass via concurrent checkout race condition

grolmus:
## Summary

Fixes a TOCTOU (Time-of-Check-Time-of-Use) race condition in coupon code usage limit enforcement (CWE-367). Concurrent `addPaymentToOrder` / `addManualPaymentToOrder` calls could bypass `usageLimit` and `perCustomerUsageLimit` restrictions, allowing a single-use coupon to be redeemed unlimited times.

Reported by Mert Akdag via security disclosure.

## Root Cause

The `validateCouponCode()` method reads the current usage count from completed orders (`active = false`), but between validation and order settlement, no serialization exists. Multiple concurrent checkouts all read `usageCount = 0` and all pass the check.

The investigation was more involved than expected:

1. **Initial approach (pessimistic lock in `validateCouponCode`)** — didn't work because the lock released before the order settled (separate transaction scope)
2. **Lock in `applyCouponCode`** — didn't work because usage count only considers completed orders, not active ones
3. **Counting active orders** — didn't work because the current order counted against itself
4. **Lock in the resolver's `@Transaction()`** — didn't work because `@vendure/core` loads from `dist/` (compiled JS), not TS source — changes weren't picked up until we rebuilt
5. **`Object.assign` for order refresh** — crashed on `Order.discounts` getter-only property

The final working fix required all three components together:
- Pessimistic lock on the Promotion row (via TypeORM `setLock('pessimistic_write')`)
- Counting orders in `ArrangingPayment` state alongside completed orders  
- Excluding the current order from the count (`excludeOrderId`)

## Changes

**`packages/core/src/service/services/order.service.ts`**
- New `revalidateCouponCodesForOrder()` method that acquires a pessimistic write lock on the Promotion row, re-validates all coupon codes, and removes invalid ones
- Called from both `addPaymentToOrder` and `addManualPaymentToOrder` before payment processing
- Re-fetches order after coupon removal so totals reflect recalculated prices

**`packages/core/src/service/services/promotion.service.ts`**
- `validateCouponCode()` accepts optional `excludeOrderId` to avoid the order blocking itself
- `countPromotionUsages()` now counts both completed orders AND orders in `ArrangingPayment` state (excluding the current order)

**`packages/core/e2e/order-promotion.e2e-spec.ts`**
- Concurrent checkout test: creates multiple independent clients, each with an order in `ArrangingPayment`, fires concurrent `addPaymentToOrder` calls
- Skipped on sql.js (no pessimistic lock support)

## Verification

Tested on a live Postgres dev server with 5 concurrent `addManualPaymentToOrder` mutations against a coupon with `usageLimit: 1`:

| Before fix | After fix |
|-----------|----------|
| 5/5 orders settled with discount | 1/5 settled with discount, 4/5 settled at full price |

Over 100 test orders were created during the investigation process to identify and verify the correct fix approach.

## Limitations

- `perCustomerUsageLimit` does not yet count `ArrangingPayment` orders — this could be added as a follow-up
- Orders abandoned in `ArrangingPayment` state will count against the usage limit until they are cleaned up or cancelled

## Test plan

- [x] Existing e2e tests pass (55 passed, sqljs)
- [x] Live Postgres test: concurrent checkout correctly enforces `usageLimit`
- [x] CI: Postgres e2e run should exercise the concurrent test
- [ ] Manual: verify single-order checkout still works correctly with usage-limited coupons


michaelbromley:
### Cross-DB lock granularity concern

The lock query filters by `couponCode`, which has no index on the `Promotion` entity. Consequences:

- **Postgres**: full table scan + lock on matching rows. Fine for small tables.
- **MySQL/MariaDB**: under default `REPEATABLE READ` isolation, this takes **next-key (gap) locks** on every scanned row, not just matches. A coupon-locked checkout would block concurrent writes to *other* promotions for the duration of the payment gateway call. Bounded by table size, but a real throughput concern under contention.
- **SQLite**: lock skipped via the catch — engine-level write serialisation provides a coarse fallback.

Adding `@Index()` to `couponCode` would resolve this but requires a migration, which makes it ineligible for `master`/patch.

### Options

- **A** — Ship as-is, follow-up index on `minor`. Correctness fine; perf suboptimal on MySQL/MariaDB under contention.
- **B** ⭐ — **Lock by PK instead of `couponCode`.** Channel-scoped `findOne` → `setLock('pessimistic_write').where('id = :id', ...)`. Always hits the PK index → single-row lock on every DB, no gap-lock fan-out. As a bonus, this incidentally fixes the channel-scope concern in the current lock query (it filters only by `couponCode + enabled`, not `ctx.channelId` — so the lock can land on the wrong channel's row in multi-channel setups). No schema change.
- **C** — Drop the lock from the patch; ship only the `ArrangingPayment` count + `revalidateCouponCodesForOrder` re-validation. Narrows the race window significantly via "last-wins" semantics but doesn't fully close it.

### Recommendation: Option B

Code-only, patch-eligible, and addresses both the gap-lock perf concern AND the channel-scope bug in one go. Lock semantics are identical from the caller's perspective — only the predicate column changes. A follow-up `@Index('couponCode')` on `minor` then becomes a pure perf optimisation for `validateCouponCode`, not a correctness fix.

michaelbromley:
## Update: e2e tests + a finding worth flagging

Pushed three further commits while chasing the failing `e2e (mysql) (22.x) / (mariadb) / (postgres)` jobs:

- `8ba75bc9a` — switch the post-payment assertions to shipping-agnostic relative comparisons. The `freeOrderAction` (`orderPercentageDiscount` at 100%) discounts the line subtotal but not shipping, so a winning order's `totalWithTax` is the shipping cost, not 0. My earlier "Tighten" commit had baked in `toBe(0)` / `toBe(6000)` which was the actual cause of the CI failure (`expected 414 to be +0`). Pre-existing behaviour, not something this PR changed — that was just my over-tight assertion.
- `d48cdfac6` — restructure the setup to actually exercise the lock. The previous loop transitioned each order to `ArrangingPayment` before the next client called `applyCouponCode`, so with the new count semantics (which now include `ArrangingPayment` orders) every applyCouponCode after the first silently returned a `CouponCodeLimitError` and only one order ever reached payment with the coupon. The lock was never under contention. Setup is now phased: addItem + applyCouponCode + setCustomer for every client (all in `AddingItems`, `count = 0` each time, every coupon application succeeds), then transition all to `ArrangingPayment`, then concurrent payment. Loud `assertSuccess` on each phase-1 result so a future regression of the count semantics fails the setup loudly instead of producing an under-set fixture.
- `06f85951c` — exact per-DB winner counts in the concurrent assertion (replaced `<= 1` / `>= N-1` with branched `toBe(expectedWinners)`).

### The finding

With the restructured test (orders genuinely co-holding the coupon at payment time), the lock + count fix produces a **DB-isolation-dependent** outcome that is worth knowing about:

- **Postgres (READ COMMITTED default)**: exactly **1 winner**. Each non-locking SELECT sees the latest committed data, so the last lock holder observes `count = 0` and keeps the coupon. This is the "last-wins" semantics the PR's docstring originally described, and it works as intended.
- **MySQL/MariaDB (REPEATABLE READ default)**: exactly **0 winners**. The transaction's consistent-read snapshot is established by the very first non-locking SELECT in `addPaymentToOrder` (the `getOrderOrThrow` at the top of the method) — *before* the pessimistic lock is acquired. Every subsequent non-locking SELECT in the same transaction (including the count query inside `validateCouponCode`) reads from that fixed snapshot. So even though the lock perfectly serialises the three concurrent transactions, when each finally runs its count query it still sees the *other* orders as holding the coupon, gets `count >= 1`, fails validation, and strips. Every transaction strips. No one wins.

Both outcomes uphold the safety invariant — the bug class this PR guards against (N winners) is impossible under either — so this is **safe to ship as-is to master**. But MySQL/MariaDB users get worse UX under contention than Postgres users: instead of one customer getting the coupon, none do. The JSDoc on `revalidateCouponCodesForOrder` now documents both branches explicitly, and the test asserts each exactly so any future drift forces the assertion to be updated.

### Suggested follow-up (minor PR)

Closing the MySQL gap properly likely needs one of:

1. Bypass the snapshot for the count query — `setLock('pessimistic_write')` or `'pessimistic_read'` on the count `QueryBuilder`. Cleanest at the call site, but locks every counted row (and likely brings gap locks back via the JOIN through `order_promotions_promotion`). May be unacceptable on busy promotion tables.
2. Restructure so the lock query is the very first read in `addPaymentToOrder`'s transaction. This means either pre-locking by `couponCode` (re-introduces the gap-lock cost on the unindexed column we explicitly avoided in Option B) or using an advisory lock keyed on `couponCode` (`GET_LOCK` on MySQL, `pg_advisory_xact_lock` on Postgres, no-op on SQLite). The advisory-lock route is probably the most surgical.
3. Add a unique index on `Promotion.couponCode` so a `couponCode`-based lock takes a record lock instead of a next-key lock. This is the migration path I mentioned in the original triage comment, scheduled for `minor`.

Happy to follow up with whichever direction the team prefers — or to drop a separate Linear issue/PR for it. None of these block the current race-condition fix from shipping; they'd close the residual MySQL contention gap.

michaelbromley:
## Update: silent over-charge mitigation (CodeRabbit follow-up)

CodeRabbit flagged that when `revalidateCouponCodesForOrder` strips a now-invalid coupon, `addPaymentToOrder` silently recomputed the total and charged the customer the new (higher) amount. Concrete example: customer applies a 100%-off `usageLimit: 1` coupon, expecting to pay $0; a concurrent checkout claims the slot; the customer is debited $60 with no user-visible signal.

Two-step plan to address this:

### Now (this PR — `master` patch — commit 62dd52843)

`addPaymentToOrder` now returns `PaymentFailedError` with a coupon-related `paymentErrorMessage` when:

1. A coupon was removed during revalidation, **AND**
2. That removal increased the order's `totalWithTax` (i.e. there's an actual surprise charge to guard against).

Strips that don't change the total — a coupon whose promotion was already deleted, a coupon stacked on top of another that already maxed out the discount — fall through silently. There's no over-charge in those cases and rejecting would just add friction.

`PaymentFailedError` is reused here deliberately: it's already in the `AddPaymentToOrderResult` union, so this is a **zero-schema-change** fix that ships safely on `master`. Storefronts that already handle `PaymentFailedError` (which they must, for actual gateway failures) get the correct default UX — show the error, let the user retry. Storefronts can opt to detect this specific case via the `paymentErrorMessage` content if they want a specialised UI.

`addManualPaymentToOrder` is admin-facing and intentionally unchanged — admins are aware of the manual-payment flow and the silent-strip behaviour there is acceptable.

The e2e race-condition test is updated to assert the new typed-error return shape.

### Next (immediate follow-up `minor` PR)

Add `CouponRemovedDuringCheckoutError` to the `AddPaymentToOrderResult` union and switch the return from `PaymentFailedError` to the new type. Storefronts that adopt the message-based detection in this PR continue to work after the new type lands; storefronts with exhaustive `__typename` switches add one case for the dedicated type. Cleaner semantics, no behaviour change beyond the error type.

### Rationale for the split

- The bug class this PR was originally written to close (over-redemption) is a **merchant** financial loss. The CodeRabbit-flagged regression is a **customer** financial loss. Customer protection should be addressed in the same release that closes the merchant-side bug — shipping master with the customer over-charge in place would be a step backwards from a consumer-protection standpoint.
- Adding to a GraphQL union is, by Vendure convention, a `minor`-branch change. Reusing an existing union member to deliver the safety property today, and upgrading the type semantics in the next minor, is the route that gets the safety property out the soonest without the schema-breaking change.

PR #4662, opened 2026-04-20
fix(core): use default Channel context when creating an Asset from a file stream

### LINKED ISSUE #4651: AssetService.createFromFileStream() fails without RequestContext because of missing languageCode
**Describe the bug**

Since v3.6, `AssetService.createFromFileStream()` fails with a NOT NULL constraint violation on `asset_translation.languageCode` when called without an explicit `RequestContext`. The method internally calls `createAssetInternal()` without passing a `translations` parameter. The fallback in `createAssetInternal` creates a default translation using `ctx.languageCode`, but when no `RequestContext` is provided, `RequestContext.empty()` is used, which has `languageCode: undefined` because it creates `new Channel()` with no `defaultLanguageCode`.

**To Reproduce**

```typescript
// In a plugin service (e.g., generating a sitemap file as an asset)
const stream = new Readable({ encoding: 'utf-8' });
stream.push('sitemap content');
stream.push(null);

// This worked pre-v3.6, fails in v3.6+ with NOT NULL constraint violation
const asset = await this.assetService.createFromFileStream(stream, 'sitemap.txt');
```

**Error output**

```
[TypeORM] Query error: INSERT INTO "asset_translation"("createdAt", "updatedAt", "languageCode", "name", "baseId", "customFields__fix_relational_custom_fields__")
VALUES (DEFAULT, DEFAULT, DEFAULT, $1, $2, DEFAULT) RETURNING "createdAt", "updatedAt", "id"
-- PARAMETERS: ["sitemap.txt",34130]

{
  "severity": "ERROR",
  "code": "23502",
  "detail": "Failing row contains (2026-04-14 16:57:08.266205, 2026-04-14 16:57:08.266205, null, sitemap.txt, 27762, 34130, null).",
  "schema": "public",
  "table": "asset_translation",
  "column": "languageCode"
}
```

Note that `languageCode` is `null` in the failing row and `DEFAULT` in the query — the value is never set.

**Expected behavior**

`createFromFileStream` should work without requiring an explicit `RequestContext`, as it did before v3.6. The method should either:
1. Accept an optional `translations` parameter and forward it to `createAssetInternal`, or
2. Use a sensible default `languageCode` (e.g., `LanguageCode.en`) when `ctx.languageCode` is undefined, or
3. Have `RequestContext.empty()` provide a fallback `languageCode`

**Root cause analysis**

In `asset.service.ts`:
https://github.com/vendurehq/vendure/blob/88467a28a474d725acaeb320b9d2cabc173fb7c4/packages/core/src/service/services/asset.service.ts#L488
All overloads of `createFromFileStream` have an optional `RequestContext` parameter.

```javascript
// Line ~380 - createFromFileStream calls createAssetInternal WITHOUT translations
const result = await this.createAssetInternal(ctx, stream, filename, mimetype);
//                                                                    ^ can be undefined -> used empty without languageCode

```

When `translations` is undefined, `createAssetInternal` falls back to:
```javascript
// Lines ~483-491
assetTranslations = [
    new AssetTranslation({
        languageCode: ctx.languageCode, // undefined when ctx = RequestContext.empty()
        name: defaultName,
        base: savedAsset,
    }),
];
```

And `RequestContext.empty()` creates a context with `new Channel()` which has no `defaultLanguageCode`, so `ctx.languageCode` is `undefined`.
https://github.com/vendurehq/vendure/blob/88467a28a474d725acaeb320b9d2cabc173fb7c4/packages/core/src/api/common/request-context.ts#L225-L232

**Environment**
- Vendure version: 3.6.2
- Database: PostgreSQL

**Workaround**

Pass an explicit `RequestContext` with a valid `languageCode`:
```typescript
const ctx = await this.requestContextService.create({
    apiType: 'admin',
});
const asset = await this.assetService.createFromFileStream(stream, 'sitemap.txt', ctx);
```

... or just use any other valid `RequestContext`

kevmtt:
# Description

  Fixes #4651.

  Since v3.6, `AssetService.createFromFileStream()` fails with a NOT NULL constraint violation on `asset_translation.languageCode` when called without an explicit `RequestContext`. The root cause: the fallback used `RequestContext.empty()`, which constructs `new Channel()` with no `defaultLanguageCode`, leaving `ctx.languageCode` as `undefined` when `createAssetInternal` builds the default translation.

  This PR introduces a new `RequestContextService.createDefaultContext()` method that builds a `RequestContext` based on the default `Channel`, guaranteeing a valid `languageCode` (falling back to `ConfigService.defaultLanguageCode` if the channel has none) and  `currencyCode`. `AssetService.createFromFileStream()` now uses this helper instead of  RequestContext.empty()` when no ctx is supplied, restoring the pre-v3.6 behavior.

  Changes:
  - `packages/core/src/service/helpers/request-context/request-context.service.ts` — add `createDefaultContext()` (admin API type,  authorized, derived from the default channel).
  - `packages/core/src/service/services/asset.service.ts` — inject `RequestContextService` and use `createDefaultContext()` as the fallback in `createFromFileStream`.
  - `packages/core/e2e/asset-create-from-file-stream.e2e-spec.ts` — new e2e test reproducing the original issue by calling
  `createFromFileStream` without a `RequestContext`.
  - `docs/.../request-context-service.mdx` — documents the new method.

  # Breaking changes

  None. `createDefaultContext()` is additive, and the asset service change only affects the previously-broken path where no
  `RequestContext` is passed (which threw a DB constraint error before).

  # Screenshots

  N/A — backend-only change.

  # Checklist

  📌 Always:

  👍 Most of the time:


Draykee:
I created the e2e test first, confirmed it failed and then added the fix.
image

Draykee:

packages/core/e2e/asset-create-from-file-stream.e2e-spec.ts
```diff
@@ -0,0 +1,47 @@
...
+            customerCount: 1,
+        });
+    }, TEST_SETUP_TIMEOUT_MS);
+
+    afterAll(async () => {
+        await server.destroy();
+    });
+
+    // https://github.com/vendurehq/vendure/issues/4651
+    it('should create an asset from a file stream without a RequestContext', async () => {
```
michaelbromley: Please put this test in the existing asset e2e suite

packages/core/src/service/helpers/request-context/request-context.service.ts
```diff
@@ -30,6 +30,26 @@ export class RequestContextService {
...
 
+    /**
+     * @description
+     * Creates a RequestContext based on the default Channel. This is useful when a service method
+     * must be called outside the normal request-response cycle and needs a context that has a valid
+     * `languageCode` derived from the default Channel's configuration.
+     *
+     * @since 3.6.3
+     */
+    async createDefaultContext(): Promise<RequestContext> {
```
michaelbromley: I think there's an opportunity here to allow one to pass in a `Partial<request context constructor args>` so they can easily eg create a `ctx` on the default channel but for the shop API, for instance.

michaelbromley (dismissed):

packages/core/src/service/helpers/request-context/request-context.service.ts
```diff
@@ -30,6 +30,30 @@ export class RequestContextService {
...
+     * @description
+     * Creates a RequestContext based on the default Channel. This is useful when a service method
+     * must be called outside the normal request-response cycle and needs a context that has a valid
+     * `languageCode` derived from the default Channel's configuration.
+     * @param config optional manual configuration overwriting the default values
+     *
+     * @since 3.6.3
+     */
+    async createDefaultContext(
+        config: Partial<ConstructorParameters<typeof RequestContext>[0]>,
```
Draykee: @michaelbromley Super awesome typescript magic ~ looks bad in the generated docs though.

michaelbromley:
Hi Kevin, just picking this one up again - not sure why all the CI workflows were failing but I just merged latest into the branch, let’s see what happens. Also the `@since` tags need to be updated now :)

kevmtt:
@michaelbromley I updated it to next patch version 3.6.5

packages/core/e2e/asset.e2e-spec.ts
```diff
@@ -25,7 +26,7 @@ describe('Asset resolver', () => {
     const { server, adminClient } = createTestEnvironment(
         mergeConfig(testConfig(), {
             assetOptions: {
-                permittedFileTypes: ['image/*', '.pdf', '.zip'],
+                permittedFileTypes: ['image/*', '.pdf', '.zip', '.md'],
```
biggamesmallworld: Widening `permittedFileTypes` for all 23 tests in this file — including `not permitted type` — to accommodate one new test. `.pdf` is already permitted; the new test can use it.

```suggestion
                permittedFileTypes: ['image/*', '.pdf', '.zip'],
```

packages/core/e2e/asset.e2e-spec.ts
```diff
@@ -358,6 +359,23 @@ describe('Asset resolver', () => {
...
+            stream.push(null);
+
+            const result = await assetService.createFromFileStream(stream, 'test-file.md');
+
+            expect(result).toBeDefined();
+            expect('id' in result).toBe(true);
+            if ('id' in result) {
+                expect(result.name).toBe('test-file.md');
+            }
+        });
```
biggamesmallworld: This describe's siblings assert absolute asset counts, so a persisted asset here breaks `filter by tags > empty array` (line 411). Moving the test to a trailing describe means it runs last and the counts stay put.

Delete it here via the suggestion below, then paste this just before the file's final `});`:

```ts
    // https://github.com/vendurehq/vendure/issues/4651
    describe('createFromFileStream without a RequestContext', () => {
        it('creates the default translation with the default Channel languageCode', async () => {
            const assetService = server.app.get(AssetService);
            const defaultChannel = await server.app.get(ChannelService).getDefaultChannel();

            const result = await assetService.createFromFileStream(
                Readable.from(['test file content']),
                'test-file.pdf',
            );

            assetGuard.assertSuccess(result);
            expect(result.name).toBe('test-file.pdf');
            expect(result.translations.map(t => t.languageCode)).toEqual([
                defaultChannel.defaultLanguageCode,
            ]);
            expect(result.channels.map(c => c.id)).toEqual([defaultChannel.id]);
        });
    });
```

It also needs a guard declared after the `createTestEnvironment(...)` call:

```ts
    const assetGuard: ErrorResultGuard<Asset> = createErrorResultGuard(input => input.id != null);
```

and the import on line 4 becomes:

```ts
import { Asset, AssetService, ChannelService, mergeConfig } from '@vendure/core';
```

Three changes from the original test beyond the move:

- `if ('id' in result) { ... }` is a pass-regardless branch — if the guard is ever false the assertion silently evaporates and the test still goes green. This file already imports `ErrorResultGuard`/`createErrorResultGuard`; use those so a failure throws.
- `expect(result).toBeDefined()` can't fail on a normally-returning async function.
- The original asserted nothing about `languageCode`, which is the entire bug — it only proved "no exception was thrown". The new version pins the translation language to the default Channel's, and pins channel assignment: the old `RequestContext.empty()` path also fed a channel-less context into `assignToCurrentChannel`, so your fix repairs that too and it's worth locking in.

```suggestion
```

packages/core/src/service/helpers/request-context/request-context.service.ts
```diff
@@ -16,11 +16,44 @@ import { User } from '../../../entity/user/user.entity';
...
+ * @description
+ * Optional overrides passed to {@link RequestContextService.createDefaultContext}. This is
+ * effectively an all-optional version of the {@link RequestContext} constructor parameters.
+ *
+ * @docsCategory request
+ * @docsPage RequestContextService
+ * @since 3.6.5
+ */
+export type CreateDefaultRequestContextConfig = Partial<ConstructorParameters<typeof RequestContext>[0]>;
+
```
biggamesmallworld: `Partial<ConstructorParameters<typeof RequestContext>[0]>` makes the `RequestContext` constructor's anonymous inline options object part of the published API by reflection. It isn't a named, exported, documented type, so any future tweak to it — adding a field, renaming `session` — becomes a silent breaking change. The generated docs also render it verbatim as `Partial<ConstructorParameters<typeof RequestContext>[0]>`, which tells a plugin author nothing about what they can pass.

Given the method below goes away, this goes with it.

```suggestion
```

packages/core/src/service/helpers/request-context/request-context.service.ts
```diff
@@ -30,6 +63,28 @@ export class RequestContextService {
...
+            apiType: 'admin',
+            channel,
+            isAuthorized: true,
+            authorizedAsOwnerOnly: false,
+            languageCode: channel.defaultLanguageCode ?? this.configService.defaultLanguageCode,
+            currencyCode: channel.defaultCurrencyCode,
+            ...config,
+        });
+    }
+
```
biggamesmallworld: `create({ apiType: 'admin' })` already does all of this: with no `channelOrToken` it calls `getDefaultChannel()`, then sets `isAuthorized: true, authorizedAsOwnerOnly: false`, and the `RequestContext` constructor derives `languageCode`/`currencyCode` from the channel. The only behavioural delta is the `?? this.configService.defaultLanguageCode` fallback, which can't fire — `Channel.defaultLanguageCode` is `@Column('varchar')`, non-nullable.

Separately, if this were kept, the `...config` spread lands *after* the derived values. So `createDefaultContext({ channel: someOtherChannel })` returns a context on channel B carrying channel A's `languageCode` and `currencyCode` — the exact invariant the method exists to guarantee, broken by the override parameter it advertises.

```suggestion
```

packages/core/src/service/services/asset.service.ts
```diff
@@ -505,7 +507,7 @@ export class AssetService {
                     ? maybeFilePathOrCtx
                     : maybeCtx instanceof RequestContext
                       ? maybeCtx
-                      : RequestContext.empty();
+                      : await this.requestContextService.createDefaultContext();
```
biggamesmallworld:
```suggestion
                      : await this.requestContextService.create({ apiType: 'admin' });
```

Same behaviour, no new public API, no new docs page. The `requestContextService` injection you added stays — it's used right here.

---

Separately, and outside this diff: `createAssetInternal` (line 633) still does an unguarded `languageCode: ctx.languageCode`. `RequestContext.empty()` is *documented* as the thing to use when a service method must be called outside the request-response cycle (`request-context.ts:217`), and `FastImporterService.initialize()` with no channel argument produces the same context — so any other caller still hits the NOT NULL violation. One line covers every entry point rather than one overload:

```ts
languageCode: ctx.languageCode ?? this.configService.defaultLanguageCode,
```

`configService` is already injected here. Worth doing regardless of what happens to the rest of this PR. Optionally also update the `RequestContext.empty()` docblock to warn that it yields a context with no `languageCode` and no real Channel — that's the trap that produced #4651.

docs/docs/reference/typescript-api/request/request-context-service.mdx
```diff
@@ -39,4 +41,74 @@ in the API layer by the AuthGuard, and creates the RequestContext which is then
 to all resolvers & controllers.
 
 
+</div>
```
biggamesmallworld: This file is marked `generated: true` and is produced by the "Generate & commit docs" CI job. The missing blank line before `### create` suggests it was hand-massaged rather than regenerated.

`git checkout master -- docs/docs/reference/typescript-api/request/request-context-service.mdx` and let the generator own it — it'll pick up the `CreateRequestContextConfig` extraction on its own.

biggamesmallworld (changes requested):
Thanks for tracking this down — the diagnosis in the description is exactly right, and I confirmed the fix works locally.

Two things before merge.

**The e2e suite is red.** The new test creates a persisted asset inside a describe whose sibling assertions count assets, so `filter by tags > empty array` at `asset.e2e-spec.ts:411` expects 11 and gets 12. All four DB legs fail on it.

**`createDefaultContext()` duplicates `create({ apiType: 'admin' })`.** That method already falls back to the default Channel when no `channelOrToken` is given, and the `RequestContext` constructor already does `languageCode || channel.defaultLanguageCode`. The `?? configService.defaultLanguageCode` guard can't fire, because `Channel.defaultLanguageCode` is a non-nullable column — a default Channel loaded from the DB always has one. `FastImporterService.initialize()` already uses this idiom. So the PR adds permanent public API surface for what is a two-word fix.

The suggestions below take this down to a one-line change in `asset.service.ts` plus a hardening line in `createAssetInternal`. I applied all of them locally: `tsc` build clean, `asset.e2e-spec.ts` 23/23 green.

The `CreateRequestContextConfig` extraction stays — genuine readability win, behaviourally identical.

Three items land outside the diff so they can't be committable suggestions; they're in the inline comments where they're relevant. Also: the CLA check is failing, which will block merge regardless.

biggamesmallworld:
Checking in on this one, @kevmtt. Every in-diff change in my review is a committable suggestion, so applying them is mostly clicking through. The three out-of-diff items are described in the inline comments where they are relevant.

Are you still able to pick this up? Either answer is fine, I just want to know whether to keep it in the queue. If you would rather hand it off, say so and I will take it over with you credited as the author.

biggamesmallworld (changes requested):
Thanks for the rework. The source side is now exactly right, and it is a smaller diff than where this started:

- `createDefaultContext()` and the reflected `Partial<ConstructorParameters<...>>` type are gone, so no new public API and no new docs page.
- `createFromFileStream` uses `this.requestContextService.create({ apiType: 'admin' })`.
- `createAssetInternal` guards `languageCode`, which covers `FastImporterService.initialize()` and every other `RequestContext.empty()` caller, not just this one overload.
- The generated docs file is back to master.

I retitled the PR so the `conventional commit format` check passes.

Requesting changes for the test only. It currently breaks the suite it was added to, which is why `e2e (mariadb)`, `(mysql)`, `(postgres)` and `(sqljs)` are all red. I ran `asset.e2e-spec.ts` on your branch:

```
✓ createAssets > create an asset from a file stream without RequestContext
✗ filter by tags > empty array — expected 12 to be 11
1 failed | 22 passed
```

Three things, all in `packages/core/e2e/asset.e2e-spec.ts`:

**1. The test persists an asset inside `describe('createAssets')`, before siblings that assert absolute counts.** `filter by tags > empty array` at line 411 expects 11. Moving the test to a trailing `describe` at the end of the file means it runs last and the counts stay put.

**2. `permittedFileTypes` is widened with `.md` for all 23 tests in the file, including `not permitted type`.** `.pdf` is already permitted, so the new test can use it and the config can go back to what it was.

**3. The assertions cannot fail on the thing being fixed.** `if ('id' in result) { ... }` is a pass-regardless branch: if the guard is ever false the assertion silently evaporates and the test still goes green. `expect(result).toBeDefined()` cannot fail on a normally-returning async function. And nothing asserts `languageCode`, which is the whole bug, so the test only proves no exception was thrown.

This file already imports `ErrorResultGuard` / `createErrorResultGuard`. Using those makes a failure throw, and pinning the translation language and the channel assignment locks in both halves of the fix (the old `RequestContext.empty()` path also fed a channel-less context into `assignToCurrentChannel`).

Here is the whole change. Revert the `permittedFileTypes` line to `['image/*', '.pdf', '.zip']`, delete the test from `describe('createAssets')`, change the import on line 4 to:

```ts
import { Asset, AssetService, ChannelService, mergeConfig } from '@vendure/core';
```

and paste this just before the file's final `});`:

```ts
    // https://github.com/vendurehq/vendure/issues/4651
    describe('createFromFileStream without a RequestContext', () => {
        const assetGuard: ErrorResultGuard<Asset> = createErrorResultGuard(input => input.id != null);

        it('creates the default translation with the default Channel languageCode', async () => {
            const assetService = server.app.get(AssetService);
            const defaultChannel = await server.app.get(ChannelService).getDefaultChannel();

            const result = await assetService.createFromFileStream(
                Readable.from(['test file content']),
                'test-file.pdf',
            );

            assetGuard.assertSuccess(result);
            expect(result.name).toBe('test-file.pdf');
            expect(result.translations.map(t => t.languageCode)).toEqual([
                defaultChannel.defaultLanguageCode,
            ]);
            expect(result.channels.map(c => c.id)).toEqual([defaultChannel.id]);
        });
    });
```

I ran this locally both ways before suggesting it:

- With your fix: `23 passed`.
- With `asset.service.ts` reverted to master: only the new test fails, with `NOT NULL constraint failed: asset_translation.languageCode`. So it reproduces #4651 and does not disturb anything else.

Run it with:

```
bun run e2e asset.e2e-spec.ts
```

from `packages/core`.

Once that is in, this is good to merge.

biggamesmallworld (approved):
The source side is right, and it is now two lines.

`createFromFileStream` uses `requestContextService.create({ apiType: 'admin' })`, which resolves the default Channel and gives the context a real `languageCode` and `currencyCode`. `createAssetInternal` guards `languageCode` against `configService.defaultLanguageCode`, so every `RequestContext.empty()` caller is covered, not just this overload. `createDefaultContext` and the generated docs page are gone.

I checked out `c4413bb04` locally. The only remaining reference to `createFromFileStream` in the repo is `AssetImporter` at `asset-importer.ts:45`, which always passes a `ctx`, so the no-ctx overload has no in-repo caller and no test. That is what let #4651 sit unnoticed for a release cycle.

The regression test was deleted during review and the replacement was never pasted, so `asset.e2e-spec.ts` now only carries the `AssetService` and `Readable` imports. I have moved the test to #5204 rather than hold the fix for it. The issue has the full block and the commands to verify it fails on the pre-fix code. The two imports are left in place because that test uses them.

One thing before merge: `conventional commit format` is red on the two `Apply suggestion from @biggamesmallworld` commits. A squash merge takes the PR title, so the merge commit will be fine.

biggamesmallworld (approved):

PR #4663, opened 2026-04-20
fix(admin-ui): Handle tokenMethod array form when generating ui-config (#4656)

### LINKED ISSUE #4656: bug(admin-ui): tokenMethod array not handled — breaks auth when configured as ['bearer']
## Description

When `authOptions.tokenMethod` is configured as an array (e.g. `['bearer']` or `['cookie', 'bearer']`), the Admin UI and Dashboard fail to recognize the bearer token method. Auth breaks because the UI performs a strict `=== 'bearer'` comparison instead of checking whether the array includes `'bearer'`.

The server-side core code already handles this correctly using `Array.isArray()` checks (see `bootstrap.ts`, `set-session-token.ts`, `extract-session-token.ts`), but the UI layer does not.

### 1. Admin UI — `packages/admin-ui/src/lib/core/src/data/data.module.ts` (line ~70)

```typescript
if (tokenMethod === 'bearer') {
```

Strict equality fails when `tokenMethod` is `['bearer']`.

### 2. Admin UI — `packages/admin-ui/src/lib/core/src/data/providers/interceptor.ts` (line ~149)

```typescript
if (this.tokenMethod === 'bearer') {
```

Same issue — `checkForAuthToken()` never extracts/stores the auth token from responses.

### 3. Type mismatch — `packages/common/src/shared-types.ts` (line ~255)

The `AdminUiConfig` type defines:
```typescript
tokenMethod: 'cookie' | 'bearer';
```

But the server-side `AuthOptions['tokenMethod']` allows:
```typescript
tokenMethod?: 'cookie' | 'bearer' | ReadonlyArray<'cookie' | 'bearer' | 'api-key'>;
```

The `AdminUiConfig` type does not account for the array form at all.

## Expected Behavior

Configuring `authOptions.tokenMethod` as `['bearer']` or `['cookie', 'bearer']` should work the same as the string form. The admin UI should check whether bearer is **included** in the configured method(s), not whether the value **equals** `'bearer'`.

## Steps to Reproduce

1. Set `authOptions.tokenMethod` to `['bearer']` in the Vendure config
2. Start the server and open the admin UI
3. Try to log in
4. Login fails — no `Authorization` header is sent, and auth tokens from responses are not stored

## Suggested Fix

Apply the same pattern the core already uses:

```typescript
const usingBearer =
    tokenMethod === 'bearer' || (Array.isArray(tokenMethod) && tokenMethod.includes('bearer'));
```

This needs to be applied in:
- `data.module.ts` — Apollo link header logic
- `interceptor.ts` — `checkForAuthToken()`
- `ui-config.ts` — dashboard vite config builder
- `shared-types.ts` — update `AdminUiConfig['tokenMethod']` type to accept arrays

## References

The server-side code already handles this correctly:
- `packages/core/src/bootstrap.ts:216` — `Array.isArray(tokenMethod) && tokenMethod.includes('cookie')`
- `packages/core/src/api/common/set-session-token.ts:19-23` — same pattern for both cookie and bearer
- `packages/core/src/api/common/extract-session-token.ts:27-32` — same pattern

kevmtt:
# Description

Fixes #4656.

When `authOptions.tokenMethod` is configured as an array (e.g. `['bearer']` or `['cookie', 'bearer']`), the Admin UI and Dashboard failed to recognize the bearer token method. The root cause was in the config-generation layer: both `AdminUiPlugin` (for the legacy Angular Admin UI) and the dashboard's Vite `getUiConfig()` helper used a strict `authOptions.tokenMethod === 'bearer'` comparison, which silently falls back to `'cookie'` for any array value. The resulting `ui-config.json` / virtual ui-config module therefore advertised `tokenMethod: 'cookie'` even when bearer was configured, so the UI never sent the `Authorization` header nor stored auth tokens from responses.

This PR normalizes the array form at the point where the server config is translated into the UI-facing config, mirroring the pattern already used elsewhere in core (`bootstrap.ts`, `set-session-token.ts`, `extract-session-token.ts`):

```ts
const serverUsesBearer =
    serverTokenMethod === 'bearer' ||
    (Array.isArray(serverTokenMethod) && serverTokenMethod.includes('bearer'));
```

Changes:
- `packages/admin-ui-plugin/src/plugin.ts` — normalize `authOptions.tokenMethod` before writing it into the generated `vendure-ui-config.json`.
- `packages/dashboard/vite/utils/ui-config.ts` — same normalization in the Dashboard's Vite ui-config builder.

Because the normalization happens before the value reaches the UI, no changes are needed in the Admin UI runtime (`data.module.ts`, `interceptor.ts`) or in the `AdminUiConfig` type — they continue to see a concrete `'cookie' | 'bearer'` value.

# Breaking changes

None. The change only affects the previously-broken path where `authOptions.tokenMethod` was supplied as an array — that path now behaves the same as the equivalent string form.

# Screenshots

N/A — server/build-time configuration change, no visual impact.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/admin-ui-plugin/src/plugin.ts
+++ b/packages/admin-ui-plugin/src/plugin.ts
@@ -305,14 +305,15 @@ export class AdminUiPlugin implements NestModule {
                 return partialConfig ? (partialConfig as AdminUiConfig)[prop] || defaultVal : defaultVal;
             }
         };
+        const serverTokenMethod = authOptions.tokenMethod;
+        const serverUsesBearer =
+            serverTokenMethod === 'bearer' ||
+            (Array.isArray(serverTokenMethod) && serverTokenMethod.includes('bearer'));
         return {
             adminApiPath: propOrDefault('adminApiPath', apiOptions.adminApiPath),
             apiHost: propOrDefault('apiHost', 'auto'),
             apiPort: propOrDefault('apiPort', 'auto'),
-            tokenMethod: propOrDefault(
-                'tokenMethod',
-                authOptions.tokenMethod === 'bearer' ? 'bearer' : 'cookie',
-            ),
+            tokenMethod: propOrDefault('tokenMethod', serverUsesBearer ? 'bearer' : 'cookie'),
             authTokenHeaderKey: propOrDefault(
                 'authTokenHeaderKey',
                 authOptions.authTokenHeaderKey || DEFAULT_AUTH_TOKEN_HEADER_KEY,
--- a/packages/dashboard/vite/utils/ui-config.ts
+++ b/packages/dashboard/vite/utils/ui-config.ts
@@ -19,13 +19,17 @@ export function getUiConfig(
 ): Omit<ResolvedUiConfig, 'version'> {
     const { authOptions, apiOptions } = config;
 
+    const serverTokenMethod = authOptions.tokenMethod;
+    const serverUsesBearer =
+        serverTokenMethod === 'bearer' ||
+        (Array.isArray(serverTokenMethod) && serverTokenMethod.includes('bearer'));
+
     // Merge API configuration with defaults
     const api = {
         adminApiPath: pluginOptions.api?.adminApiPath ?? apiOptions.adminApiPath ?? ADMIN_API_PATH,
         host: pluginOptions.api?.host ?? 'auto',
         port: pluginOptions.api?.port ?? 'auto',
-        tokenMethod:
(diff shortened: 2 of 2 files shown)


Draykee:

williamrijksen:
Thanks for the work on this PR!

@michaelbromley Do you have an estimate of when this might be released? We're currently looking to introduce API keys, and this change would play an important role in that.

Also, would it be possible to include this in an upcoming patch (hotfix) release if feasible?

Thanks in advance!

michaelbromley (approved):

PR #4669, opened 2026-04-23
fix(core): Avoid OOM in product-to-channel assign/remove via query relation strategy

arthur-nesterenko:
## Description

`ProductService.assignProductsToChannel` and `removeProductsFromChannel` load products with relations `['variants', 'assets', 'optionGroups', 'optionGroups.options']` using TypeORM's default join strategy. This produces a cartesian product (variant × asset × optionGroup × option) — TypeORM hydrates every duplicated row into entities before deduping, so heap usage scales with join-row count, not distinct entities. Even single-product calls can spike memory severely; batched channel assignment OOMs.

Switching to `relationLoadStrategy: 'query'` with `loadEagerRelations: false` makes TypeORM issue one query per relation, eliminating the cartesian blow-up. Downstream code in both methods only consumes the IDs of the explicitly-listed relations, so disabling eager relations on these reads is safe.

### Benchmark (1 product, 20 variants, 2 option groups, `@vendure/core` 3.6.2)

| Strategy | Time | Heap delta |
|---|---|---|
| default join | 1956 ms | +274.7 MB |
| query + no eager | 44 ms | ~0 MB |

**Testing:** ran locally — `packages/core` unit suite (811/811), plus `product-channel`, `asset-channel`, and `channel` e2e suites (62/62) all pass against the fix.

## Breaking changes

None. Public API and behavior unchanged; only the internal relation-load strategy changes.

## Screenshots

N/A (backend perf fix).

## Checklist

### DIFF
--- a/packages/core/src/service/services/product.service.ts
+++ b/packages/core/src/service/services/product.service.ts
@@ -331,6 +331,8 @@ export class ProductService {
         const productsWithVariants = await this.connection.getRepository(ctx, Product).find({
             where: { id: In(input.productIds) },
             relations: ['variants', 'assets', 'optionGroups', 'optionGroups.options'],
+            relationLoadStrategy: 'query',
+            loadEagerRelations: false,
         });
         const productIds = unique(productsWithVariants.map(p => p.id));
         await Promise.all(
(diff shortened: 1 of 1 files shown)


arthur-nesterenko:

michaelbromley (approved):

PR #4672, opened 2026-04-26
fix(core): Handle undefined relation in entity hydrator

### LINKED ISSUE #4661: EntityHydrator: visit() crashes on undefined relation while handling null correctly
**Describe the bug**

`EntityHydrator.getRelationEntityAtPath` throws `TypeError: Cannot read properties of undefined (reading '<relation>')` when a relation along the path it is traversing is `undefined`. The helper's internal `visit()` function handles `null` (e.g. a soft-deleted relation target), but not `undefined`, so it recurses into `visit(undefined, [...])` and then dereferences `undefined[part]`.

The bug is in `node_modules/@vendure/core/dist/service/helpers/entity-hydrator/entity-hydrator.service.js` (v3.6.2, around line 245):

```js
else if (target === null) {     // ← only null is guarded
    result.push(target);
}
else {
    if (parts.length === 0) {
        result.push(target);
    }
    else {
        visit(target, parts.slice());   // ← when target is undefined,
                                        //   the next call crashes on parent[part]
    }
}
```

**To Reproduce**

Steps to reproduce the behavior (observed path):

1. Have an `Order` with at least one `OrderLine` whose `productVariant` relation resolves to `undefined` rather than `null` when loaded (can happen with certain soft-delete + `relationLoadStrategy: 'query'` combinations).
2. Call `entityHydrator.hydrate(ctx, order, { relations: ['lines.productVariant.facetValues', 'lines.productVariant.translations', 'surcharges'] })` from a plugin — e.g. a custom `ShippingCalculator.calculate`.
3. The hydrator computes `missingRelations`, runs the query, merges results, then calls `getRelationEntityAtPath(target, ['lines', 'productVariant', 'translations'])` for each missing relation.
4. When `visit()` reaches a line whose `productVariant` is `undefined`, it recurses into `visit(undefined, ['translations'])`, which then throws `TypeError: Cannot read properties of undefined (reading 'translations')`.

**Expected behavior**

`getRelationEntityAtPath` should treat `undefined` the same as `null`: push it into the result array and stop traversing that branch. Downstream code (`isTranslatable`, `applyChannelPriceAndTax`) already handles `undefined` entries safely via optional chaining / `instanceof` checks.

**Actual behavior**

The hydrator throws a `TypeError`, which bubbles up out of `hydrate()` and aborts whatever called it. In our case this is a custom `ShippingCalculator`, so the calculator fails and the customer would see no shipping price — we only avoid that because we wrap the call in our own `try/catch` and return a fallback price.

**Screenshots/Videos**

N/A

**Error logs**

```
Error calculating express shipping. Order #<redcted-order-code>. args: {"taxRate":0}, error: TypeError: Cannot read properties of undefined (reading 'translations')
Error calculating express shipping. Order #<redcted-order-code>. args: {"taxRate":0}, error: TypeError: Cannot read properties of undefined (reading 'translations')
```

(These were caught at our call site; without the try/catch the stack would terminate in `visit` → `getRelationEntityAtPath` → `EntityHydrator.hydrate`.)

- Nodejs version: 20.19.0 (local dev); Node 20.18.0 in production Docker image
- Database (mysql/postgres etc): PostgreSQL (via `pg ^8.11.5`)
- Operating System (Windows/macOS/Linux): Linux (Google Cloud Run); macOS 14 for local dev
- Browser (if applicable): N/A — happens server-side during a GraphQL mutation

**Configuration**

Happens when using a `EntityHydrator` instance, e.g. in a shipping calculator:

```typescript
import { EntityHydrator, ShippingCalculator, ... } from '@vendure/core';

let hydrator: EntityHydrator;

export const myShippingCalculator = new ShippingCalculator({
  code: 'my-shipping-calculator',
  // ...
  init(injector) {
    hydrator = injector.get(EntityHydrator);
  },
  calculate: async (ctx, order, args) => {
    await hydrator.hydrate(ctx, order, {
      relations: [
        'lines.productVariant.facetValues',
        'lines.productVariant.translations',
        'surcharges',
      ],
    });
    // ... compute shipping from order.lines[i].productVariant ...
  },
});
```

**Minimal reproduction**

Not yet isolated into a standalone repo, but the bug is reproducible by calling `getRelationEntityAtPath` directly with an entity graph where any intermediate relation along the path is `undefined`:

```typescript
// Synthetic repro
const order = {
  lines: [
    { productVariant: { translations: [{ languageCode: 'en' }] } },
    { productVariant: undefined },   // ← triggers the bug
  ],
};
// Internally, the hydrator would call something equivalent to:
// getRelationEntityAtPath(order, ['lines', 'productVariant', 'translations'])
// which throws TypeError instead of returning [translations[], undefined]
```

**Workaround**

Patch `entity-hydrator.service.js` to replace strict null check with loose equality, so `undefined` is handled alongside `null`:

```diff
-            else if (target === null) {
+            else if (target == null) {
                 result.push(target);
             }
```

Applied via `patch-package`. This is a 1-character change that matches the existing "missing relation" handling pattern.

**Additional context**

- When did this issue start occurring? After upgrading from `@vendure/core@3.5.7` to `3.6.2`. It's possible the bug existed earlier but was masked by a different error (`InternalServerError: error.entity-has-no-translation-in-language`) we had already fixed. Post-upgrade, the translation error went away and exposed this `TypeError`.
- Does it happen consistently or intermittently? Intermittently — only for specific orders.
- Any recent changes to our setup? Upgrade from Vendure 3.5.7 → 3.6.2, no plugin or schema changes at the same time. The shipping calculator code has been stable for weeks.
- The sibling `null` branch in `visit()` was added specifically to distinguish "relation has been fetched but was null in the database" from "relation has not been fetched" (comment in `getMissingRelations` says as much). Extending that to also handle `undefined` — which can arise for the same semantic reason in practice — is consistent with that intent.

izumi0uu:
# Description

Fixes #4661.

This PR updates `EntityHydrator.getRelationEntityAtPath()` so that an `undefined` intermediate relation is treated the same way as `null`.

Previously, the helper stopped traversing when a relation was `null`, but continued traversing when it was `undefined`. For a path such as `lines.productVariant.translations`, this could cause a `TypeError` when one `OrderLine` had `productVariant: undefined`.

A focused regression test has been added for this case.

# Breaking changes

No breaking changes.

# Screenshots

N/A. This is a server-side bug fix.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from 'vitest';
+
+import { EntityHydrator } from './entity-hydrator.service';
+
+describe('EntityHydrator', () => {
+    describe('getRelationEntityAtPath()', () => {
+        // https://github.com/vendurehq/vendure/issues/4661
+        it('treats undefined intermediate relations as terminal values', () => {
+            const hydrator = new EntityHydrator(undefined as any, undefined as any, undefined as any);
+            const translation = { languageCode: 'en', name: 'Laptop' };
+            const order = {
+                lines: [
+                    {
+                        productVariant: {
+                            translations: [translation],
+                        },
+                    },
+                    {
+                        productVariant: undefined,
+                    },
+                ],
+            };
+
+            const result = (hydrator as any).getRelationEntityAtPath(order, [
+                'lines',
+                'productVariant',
+                'translations',
+            ]);
+
+            expect(result).toEqual([translation, undefined]);
+        });
+    });
+});
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
@@ -277,7 +277,7 @@ export class EntityHydrator {
                         visit(item, parts.slice());
                     }
                 }
-            } else if (target === null) {
+            } else if (target == null) {
                 result.push(target);
             } else {
                 if (parts.length === 0) {


izumi0uu:

izumi0uu:
recheck

michaelbromley (approved):
👍

PR #4676, opened 2026-04-27
fix(dashboard): Support action bar positioning relative to extensions

### LINKED ISSUE #4667: DashboardActionBarItem position id targeting extensions doesn't work
**Describe the bug**
When defining a new DashboardActionBarItem, the position argument doesn't work when the itemId is another extension. The item simply isn't rendered anymore.

**To Reproduce**

1. Define two DashboardActionBarItem, A and B, and give them IDs
2. Have A refer to the other with the position
3. Button B is not rendered at all

**Expected behavior**
B is rendered and appears before or after A. It's also stated in [the documentation](https://docs.vendure.io/current/core/reference/dashboard/extensions-api/action-bar#position) that the itemId will respect items defined by another extension.

**Actual behavior**
B does not exist at all. In contrast, the position option does it's job when referring to items defined by the page (e.g. the Fulfillment button). 

- Nodejs version: 22

**Minimal reproduction**
```typescript
defineDashboardExtension({
    actionBarItems: [
        {
            pageId: 'order-detail',
            component: ({ context }) => <Button>A</Button>,
            id: 'a'
        },
        {
            pageId: 'order-detail',
            component: ({ context }) => <Button>B</Button>,
            id: 'b',
            position: {
                itemId: 'a',
                order: 'before'
            }
        },
    ],
});
```
Note: In this example, both extensions are defined by the same (pseudo) plugin, but the outcome doesn't change by whether or not the referenced item is from another extension. 

**Additional context**
I also think if A doesn't exist (or at least not in this context/state), B should still be rendered with the default behaviour of just appearing first. This is regardless of whether or not A is a custom extension or an item provided by the page. Although it might be convenient to have one item disappear if the other one does, in practice, every component should decide visibility for itself. Especially when chained to the action bar items provided by Vendure, behaviour might change, and this would cause unintentional side-effects. It's also not very transparent to the user/developer.

izumi0uu:
# Description

Fixes action bar extension positioning when one extension item targets another extension item via `position.itemId`.

Previously, `mergeAndSortActionBarItems()` only used inline `ActionBarItem.itemId` values as positioning targets. This worked when an extension item was positioned relative to a built-in or inline action bar item, but it failed when `position.itemId` pointed to another extension item's `id`. In that case, the positioned extension item was never matched during rendering and disappeared from the action bar.

This PR replaces the inline-only matching flow with a unified sortable item model:

- inline `ActionBarItem` children and dashboard extension action bar items are both converted into `MergedActionBarItem` nodes
- unpositioned extension items and inline items are used as root render items
- positioned extension items are grouped by their target id
- each root item renders its `before`, `replace`, and `after` extension items recursively

With this approach, `position.itemId` can target either an inline `ActionBarItem.itemId` or another `DashboardActionBarItem.id`, while preserving the previous behavior for inline targets and unpositioned extension items.

Test coverage has been added for extension-to-extension `before`, `after`, and `replace` positioning, as well as the existing inline target behavior.

Closes #4667

# Breaking changes

No.

# Screenshots

Not applicable.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/lib/framework/layout-engine/page-layout.spec.tsx
+++ b/packages/dashboard/src/lib/framework/layout-engine/page-layout.spec.tsx
@@ -266,4 +266,23 @@ describe('PageLayout', () => {
 
         expect(getRenderedActionBarIds(markup)).toEqual(['b', 'a', 'save-button']);
     });
+
+    it('renders a positioned extension action bar item when its target is missing', () => {
+        registerActionBarItem('orphan', { itemId: 'missing', order: 'before' });
+
+        const markup = renderActionBar();
+
+        expect(getRenderedActionBarIds(markup)).toEqual(['save-button', 'orphan']);
+    });
+
+    it('renders cyclic positioned extension action bar items without dropping them', () => {
+        registerActionBarItem('a', { itemId: 'b', order: 'before' });
+        registerActionBarItem('b', { itemId: 'a', order: 'before' });
+
+        const markup = renderActionBar();
+        const renderedIds = getRenderedActionBarIds(markup);
+
+        expect(renderedIds).toHaveLength(3);
+        expect(renderedIds).toEqual(expect.arrayContaining(['a', 'b', 'save-button']));
+    });
 });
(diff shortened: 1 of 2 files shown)


michaelbromley (approved):
Thanks for the fix!

PR #4679, opened 2026-04-28
fix(dashboard): Check required permissions when rendering custom page…

### LINKED ISSUE #4678: requiresPermission is ignored when defining pageBlocks
**Describe the bug**
When adding a page block to a detail page, `requiresPermission` is ignored.

The permissions are never checked in  https://github.com/vendurehq/vendure/blob/master/packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx#L220

**To Reproduce**

```
defineDashboardExtension({
	pageBlocks: [
		{
			id: 'some-id',
			title: <Trans>Test</Trans>,
			location: {
				pageId: 'customer-detail',
				position: {
					blockId: 'custom-fields',
					order: 'after',
				},
				column: 'main',
			},
			// shouldRender works:
			// shouldRender(context) {
			//	const { hasPermissions } = usePermissions();
			//	return hasPermissions(['UpdateAdministrator']);
			// },
			component: CustomerRoleSelector,
			// requiresPermission is ignored
			requiresPermission: 'UpdateAdministrator',
		},
	],
}
```

**Expected behavior**
The block shouldn't be rendered for users without the `UpdateAdministrator` permission

**Actual behavior**
Block is always rendered

- Nodejs version: v24.13.0

**Workaround**
`shouldRender` with the `usePermissions` hook is a valid workaround.

LucidityDesign:
# Description

Considers `requiresPermission` when rendering page blocks.

Fixes https://github.com/vendurehq/vendure/issues/4678

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
+++ b/packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
@@ -284,7 +284,8 @@ export function PageLayout({ children, className }: Readonly<PageLayoutProps>) {
                     const ExtensionBlock =
                         extensionBlock.component &&
                         extensionBlockShouldRender &&
-                        hasPermissions(permissions ?? []) ? (
+                        // using `hasPermissions` over `PermissionGuard` as this would defeat the `isPageBlock` typeguard
+                        hasPermissions(permissions) ? (
                             <BlockComponent
                                 key={extensionBlock.id}
                                 column={extensionBlock.location.column}
--- a/packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
+++ b/packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
@@ -277,14 +277,18 @@ export function PageLayout({ children, className }: Readonly<PageLayoutProps>) {
 
                     const ExtensionBlock =
                         extensionBlock.component && extensionBlockShouldRender ? (
-                            <BlockComponent
+                            <PermissionGuard
                                 key={extensionBlock.id}
-                                column={extensionBlock.location.column}
-                                blockId={extensionBlock.id}
-                                title={extensionBlock.title}
+                                requires={extensionBlock.requiresPermission ?? []}
                             >
-                                {<extensionBlock.component context={page} />}
-                            </BlockComponent>
+                                <BlockComponent
+                                    column={extensionBlock.location.column}
+                                    blockId={extensionBlock.id}
+                                    title={extensionBlock.title}
+                                >
+                                    {<extensionBlock.component context={page} />}
+                                </BlockComponent>
+                            </PermissionGuard>
                         ) : undefined;
 
                     if (extensionBlockShouldRender && ExtensionBlock) {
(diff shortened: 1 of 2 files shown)


michaelbromley:
Hey, thanks for the fix - can you look into the test failures when you get a chance?

LucidityDesign:
@michaelbromley this adds a implicit dependency to `AuthProvider` via the usePermissions hook. Do you think this will be a problem for the PageLayout component?

packages/dashboard/src/lib/framework/layout-engine/page-layout.tsx
```diff
@@ -274,9 +276,15 @@ export function PageLayout({ children, className }: Readonly<PageLayoutProps>) {
...
+                    const requiresPermission = extensionBlock.requiresPermission ?? [];
+                    const permissions = Array.isArray(requiresPermission)
+                        ? requiresPermission
+                        : [requiresPermission];
 
                     const ExtensionBlock =
-                        extensionBlock.component && extensionBlockShouldRender ? (
+                        extensionBlock.component &&
+                        extensionBlockShouldRender &&
+                        hasPermissions(permissions ?? []) ? (
```
LucidityDesign: I'm using the hook instead of the `PermissionGuard` component, because `isPageBlock` would filter that out

michaelbromley: thanks for the clarification, that's useful

michaelbromley (changes requested):
Thanks for the fix — this addresses a real bug (the `requiresPermission` field on `DashboardPageBlockDefinition` was being silently ignored). Two changes needed before this can merge:

### 1. Retarget to `master`

This is a bug fix to existing behaviour — the `requiresPermission` field already exists on `DashboardPageBlockDefinition`, the layout engine just wasn't honouring it. There's no new API surface here, so per the branching convention this should target `master` rather than `minor`. That also gets the fix out to users on the patch line sooner, which matters for anything permission-adjacent.

### 2. Drop the dead `?? []` on the `hasPermissions` call

```ts
const requiresPermission = extensionBlock.requiresPermission ?? [];
const permissions = Array.isArray(requiresPermission) ? requiresPermission : [requiresPermission];
// ...
hasPermissions(permissions ?? [])
//                        ^^^^^ never fires
```

`permissions` is already normalised to `string[]` by the time it reaches the call — both branches of the `Array.isArray` ternary return a string array, and `hasPermissions` is strictly typed `(permissions: string[]) => boolean`. The trailing `?? []` can't ever fire and slightly muddies the reader's mental model by implying `permissions` might be nullable.

```ts
hasPermissions(permissions)
```

### Re: using the hook directly vs `PermissionGuard`

Your reasoning in the inline comment is correct — wrapping with `PermissionGuard` would defeat the `isPageBlock` typeguard and the column-based filtering at lines 305-309, since the wrapper element doesn't carry `column` / `blockId` props. The hook approach is the right call here.

Worth dropping a one-liner comment in the code explaining that, so the next person to touch this doesn't try to "tidy it up" by switching to `PermissionGuard`.

### Nice-to-have (non-blocking)

A negative test exercising `hasPermissions: () => false` would be valuable — right now the mock unconditionally returns `true`, so the suppress-when-denied path isn't actually covered by the test you added. Given the hook-based gating is now the only safety net, a regression test feels worth the small extra cost.

Also: PR title typo — "dashbaord" → "dashboard" — worth fixing so changelog tooling picks it up cleanly.

michaelbromley:
Thank you!

PR #4681, opened 2026-04-28
fix(core): Prevent duplicate execution of fast scheduled tasks

BibiSebi:
# Description

When multiple workers are configured, scheduled tasks with very short execution times (sub-50ms) can be executed more than once per scheduled tick. The Postgres-based locking introduced in #4069 is correct in isolation, but it does not account for clock skew between worker instances: a faster worker acquires the lock, completes the task, and releases the lock before a worker with a lagging clock reaches the same tick. The lagging worker then sees `lockedAt IS NULL`, acquires the lock, and runs the task a second time.

This PR adds a short post-completion **hold window** to `tryAcquireLock`. The lock itself is still released immediately on completion (so `lockedAt` semantics are unchanged); the acquire path additionally rejects if `lastExecutedAt` is more recent than `now - holdMs`. The hold is computed per task as `min(intervalMs * 0.8, 5_000ms)` — large enough to absorb realistic inter-worker clock skew, always strictly less than the schedule interval, and capped at 5s for long-interval tasks.

Manual triggers (`runScheduledTask` / "Run" button) bypass the hold via `skipHoldCheck: true`. Manual triggers are already deduplicated by `manuallyTriggeredAt` (atomically cleared by exactly one worker), so they don't have the inter-worker race that the hold guards against, and silently dropping a user click would be poor UX. The cron-driven path (`executeTask(task)(job)`) and the manual-trigger path (`runManually(task)`) now go through a shared `runTask` helper with an explicit flag, so the "is this a manual trigger?" decision is made at the call site instead of being inferred from the `job` argument.

### Drive-by fix

Moved the timeout timer cleanup in `runTask` into a `finally` block so the timer is cleared on both success and error paths. Previously a task that failed before the timeout fired would leak the timer and emit a misleading `Scheduled task <id> timed out after <ms>ms` warning for an already-failed task.

Relates to OSS-511.

# Breaking changes

None. No new public options, no migration, no schema change.

# Screenshots

N/A.

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have added or updated test cases — tests for the hold-window behavior exist locally and will follow in a separate commit
- [x] I have updated the README if needed — not applicable


BibiSebi:
TODO:


- add sufficient tests

BibiSebi:
Tests without the implementation changes: 
image

michaelbromley (approved):

michaelbromley:
Great work!

PR #4684, opened 2026-04-29
fix(dashboard): correct mistranslated Swedish strings in sv.po

comega-johan:
## Summary

103 entries in `packages/dashboard/src/i18n/locales/sv.po` had Polish text in their `msgstr` rather than Swedish translations. This appears to be a copy/paste accident at some point in the translation pipeline — the same Polish strings can be found verbatim in `pl.po`.

The result is that Swedish-speaking users see things like:

- *Product variants* → "Warianty produktu" (Polish for "product variants")
- *Simple product* → "Produkt prosty" (Polish for "simple product")
- *Single variant, no options* → "Pojedynczy wariant, bez opcji"
- *Add payment* → "Dodaj płatność"
- *No* → "Nie", *Yes* → "Tak", *Close* → "Zamknij"

…and so on. Some examples (the full list is in the diff):

| msgid                       | before                          | after                              |
| --------------------------- | ------------------------------- | ---------------------------------- |
| Product variants            | Warianty produktu               | Produktvarianter                   |
| Simple product              | Produkt prosty                  | Enkel produkt                      |
| Single variant, no options  | Pojedynczy wariant, bez opcji   | En variant, inga alternativ        |
| Add payment                 | Dodaj płatność                  | Lägg till betalning                |
| No                          | Nie                             | Nej                                |
| Yes                         | Tak                             | Ja                                 |
| Close                       | Zamknij                         | Stäng                              |
| Items per page              | Pozycji na stronę               | Antal per sida                     |
| Drop files here to upload   | Upuść pliki tutaj, aby przesłać | Släpp filer här för att ladda upp  |

## Detection method

Each `msgstr` in `sv.po` was compared against every other locale file. Strings that exactly matched a non-English locale and contained language-specific characters (ą, ę, ł, ć, ż, ź for Polish, etc.) were flagged as candidates and manually re-translated.

Norwegian (nb) was excluded from the comparison because Swedish and Norwegian share many short phrases ("Synlighet", "Ny skattesats", …) which would otherwise produce false positives. Likewise, Swedish words that happen to coincide with another language's translation (e.g. "Stock"→"Lager" matches German, "Catalog"→"Katalog" matches Polish) are intentionally left untouched — they're correct Swedish.

## Test plan

- [x] Diff shows exactly 103 inserted lines and 102 deleted lines (one entry was already correct)
- [x] No structural changes to the `.po` format (only `msgstr` values changed)
- [x] Verified by re-running the diagnostic comparison after the fix — only false-positive matches remain

This is a docs-only fix that doesn't touch any code paths — should be safe to merge.


comega-johan:

michaelbromley:
Thanks for this fix! It also brought my attention to a number of issues with other translations which are now also fixed and will be in the next patch next week.

PR #4687, opened 2026-04-30
feat(dashboard): support assigning multiple channels in bulk actions

casperiv0:
# Description

# Breaking changes
Nope!

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:


packages/dashboard/src/lib/components/shared/multi-select.tsx
```diff
@@ -127,33 +132,40 @@ export function MultiSelect<T extends boolean>(props: MultiSelectProps<T>) {
...
-                {(showSearch === true || items.length > 10) && (
-                    <div className="p-2">
-                        <Input
-                            type="text"
-                            placeholder={searchPlaceholder}
-                            value={search}
-                            onChange={e => setSearch(e.target.value)}
-                            className="w-full px-2 py-1 text-sm border rounded"
-                        />
+            <PopoverPrimitive.Portal>
```
casperiv0: Had to wrap the `PopoverContent` in a Portal because the Dialog is also wrapped in its own Portal and would therefor not show the `PopoverContent`. Tested on the `MultiSelect` places and continued working as expected!

michaelbromley (commented):
## Review

The core idea here is sound — turn the single-channel select into a multi-select and loop the existing per-channel mutation. The `Promise.allSettled` fan-out is the right shape too: since `buildInput` already bundles all selected entity IDs into each call, one mutation per channel is genuinely what the GraphQL mutation shape requires. But there are a couple of things that should be addressed before merge.

### Must change

**1. Dropped `ChannelCodeLabel` — the default channel will show its raw internal code.**

The previous code rendered each channel option through `<ChannelCodeLabel code={channel.code} />`, which swaps the default channel's internal system `code` string for a translated "Default channel" label. This PR replaces that with a plain string label:

```js
const selectItems = useMemo(
    () => availableChannels.map(ch => ({ value: ch.id, label: ch.code })),
    [availableChannels],
);
```

`availableChannels` only filters out the *currently active* channel, not the default one. So an admin working in a non-default channel will see the default channel in the list showing its raw internal code instead of "Default channel". `MultiSelect` already has a `display` field designed for exactly this case, so the fix is:

```js
{ value: ch.id, label: ch.code, display: <ChannelCodeLabel code={ch.code} /> }
```

`ChannelCodeLabel` is used everywhere else channels are listed (`channel-switcher`, `channel-selector`, `roles`, `channels`, `manage-languages-dialog`), so it's worth keeping consistent here.

**2. The `PopoverPrimitive.Portal` wrapper in `multi-select.tsx` appears to be a no-op.**

Tracing this through `@base-ui/react@1.3.0` and `@vendure-io/ui`: `PopoverContent` (from `@vendure-io/ui/components/atoms/popover.tsx`) already wraps its `Positioner`/`Popup` in `PopoverPrimitive.Portal` internally — the same portal component this diff imports and wraps around it a second time. Nesting two of the same portal here doesn't change where anything renders; it just adds an extra empty wrapper `<div>`, and the pre-existing inner portal still does the actual work.

On top of doing nothing, it:
- Imports `@base-ui/react/popover` directly in application code, bypassing the design-system's own re-export (`@/vdb/components/ui/popover.js`).
- Pulls in `@base-ui/react`, which isn't a declared dependency of `packages/dashboard/package.json` — it only resolves because it happens to be hoisted today.

If this was chasing a genuine z-index/stacking bug when the multi-select is used inside a Dialog, this change doesn't fix it — it would need root-causing (likely the Dialog's own portal/stacking context). Otherwise it should be reverted.

### Should change

On partial failure the dialog stays open and the already-succeeded channels remain selected in the multi-select, so retrying will re-send the mutation for channels that already succeeded. Whether that's harmless depends on whether the underlying assign mutations are idempotent for an "already assigned" channel. If they aren't, it'd be worth pruning `selectedChannelIds` down to the failed ones after a partial failure.

### Nitpicks

- `description ? { description } : undefined` in the error toast is unnecessary — `{ description: undefined }` behaves identically for `toast.error`.
- The `useMemo` wrapping on `availableChannels` and `selectItems` is ceremony for a trivial `.filter()`/`.map()` over a short channel list.

casperiv0:
@michaelbromley - thanks for the review! Updated the PR 🙏!

PR #4693, opened 2026-05-02
fix(dashboard): Apply addCustomFields() to detail page mutations

### LINKED ISSUE #4692: Updating custom fields on asset in React Admin Dashboard fails
**Describe the bug**
Trying to update some custom fields on an Asset in the React Dashboard. When i try to save i get a graphql validation error:
```
{
    "errors": [
        {
            "message": "Field \"customFields\" of type \"AssetCustomFields\" must have a selection of subfields. Did you mean \"customFields { ... }\"?",
            "locations": [
                {
                    "line": 8,
                    "column": 5
                }
            ],
            "extensions": {
                "http": {
                    "status": 400,
                    "headers": {}
                },
                "code": "GRAPHQL_VALIDATION_FAILED"
            }
        }
    ]
}
```

**To Reproduce**
1. Add a custom field to Asset entity
2. Go to assets in the dashboard
3. Upload an asset and try to set this custom field to some value
4. Hit update
5. Go to browser console and view the response from the server

**Expected behavior**
It should be possible to save assets and update their custom fields in the dashboard

**Actual behavior**
Cannot save assets, getting validation error from API

**Error logs**
```
{
    "errors": [
        {
            "message": "Field \"customFields\" of type \"AssetCustomFields\" must have a selection of subfields. Did you mean \"customFields { ... }\"?",
            "locations": [
                {
                    "line": 8,
                    "column": 5
                }
            ],
            "extensions": {
                "http": {
                    "status": 400,
                    "headers": {}
                },
                "code": "GRAPHQL_VALIDATION_FAILED"
            }
        }
    ]
}
```
```
Paste error logs here
```

- Nodejs version: 24
- Database (mysql/postgres etc): pg
- Operating System (Windows/macOS/Linux): linux
- Browser (if applicable):

**Configuration**

```typescript
// Your configuration here
```

**Minimal reproduction**

**Workaround**
I have currently no workaround. Blocked.

**Additional context**

Carnewal:
The useDetailPage hook wraps the query document with addCustomFields() so that bare `customFields` selections are replaced with proper sub-selections at runtime. 

The same wrapping was missing from the create and update mutation documents, so any detail page whose mutation response selection includes `customFields` (e.g. the asset detail page) fails with a GraphQL validation error once the entity has at least one custom field registered:

  Field "customFields" of type "AssetCustomFields" must have a
  selection of subfields.

Closes #4692

# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4693"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/framework/page/use-detail-page.ts
+++ b/packages/dashboard/src/lib/framework/page/use-detail-page.ts
@@ -283,7 +283,7 @@ export function useDetailPage<
     };
 
     const createMutation = useMutation({
-        mutationFn: createDocument ? api.mutate(createDocument) : undefined,
+        mutationFn: createDocument ? api.mutate(addCustomFields(createDocument)) : undefined,
         onSuccess: data => {
             if (createDocument) {
                 const createMutationName = getMutationName(createDocument);
@@ -294,7 +294,7 @@ export function useDetailPage<
     });
 
     const updateMutation = useMutation({
-        mutationFn: updateDocument ? api.mutate(updateDocument) : undefined,
+        mutationFn: updateDocument ? api.mutate(addCustomFields(updateDocument)) : undefined,
         onSuccess: async data => {
             if (updateDocument) {
                 const updateMutationName = getMutationName(updateDocument);


Carnewal:

michaelbromley (changes requested):
Hi, thanks for the PR. This is a valid way to solve the issue but I'd rather use a different approach:

## why is this only an issue for the asset detail page?

The reason is that all the other detail pages on select `{ id }` in their mutation documents like this:

```ts
export const updateProductDocument = graphql(`
    mutation UpdateProduct($input: UpdateProductInput!) {
        updateProduct(input: $input) {
            id
        }
    }
`);
```

whereas on the asset detail page we have:

```ts
export const assetUpdateDocument = graphql(
    `
        mutation AssetUpdate($input: UpdateAssetInput!) {
            updateAsset(input: $input) {
                ...Asset
                tags {
                    id
                    value
                }
                customFields
            }
        }
    `,
    [assetFragment],
);
```

which triggers the error. I believe the better solution is to bring the asset detail page in line with the other detail pages and just have:

```ts
export const assetUpdateDocument = graphql(
    `
        mutation AssetUpdate($input: UpdateAssetInput!) {
            updateAsset(input: $input) {
                id
            }
        }
    `
);
```

### why not this solution?

This solution will add a custom field selection to the mutation documents for _all_ entities. Most of the time this will not be a big deal, but where there are a lot of custom fields and especially relation custom fields, this will cause 2x the load on the server on every save, as the relations will need to be joined and returned on the mutation, and then immediately again when the query document gets re-fetched (we invalidate the tanstack query cache after mutations).

michaelbromley:
Hey @Carnewal - I had to manually fix this in order to get a planned patch out on schedule, so I’ll close this. Thanks again for taking the time to work on that 👍

PR #4699, opened 2026-05-05
fix(core): Assign new variants to all product channels

### LINKED ISSUE #4532: Additional product variant not visible in assigned channel when created after channel assignment
**Describe the bug**
When a new product variant is created after a channel has already been assigned to the product, the new variant does not appear in that channel.

**To Reproduce**
1. Create a new channel.
2. Open a product detail page, for example “Modern Cafe Chair”.
3. Assign the newly created channel to the product.
4. Click Manage variants.
5. Add a new option, for example to the color option group.
6. Reload the page, since the new option is not immediately selectable in the Add variant popup.
7. Click Add variant and fill in the fields.
8. Return to the product detail page and confirm that the new product variant has been created.
9. Switch to the assigned channel.

**Expected behavior**
The newly created product variant should also be visible in the already assigned channel.

**Actual behavior**
The newly created product variant is not visible in the channel.

**Environment**
- @vendure/core version: 3.5.5

**Workaround**
The channel can be removed from the product and added again, but this does not seem like the intended or optimal behavior, especially because the channel-specific prices are lost when the channel is removed.

Ryrahul:
Fixes #4532

# Description

When a new product variant is created after a channel has already been assigned to the product, the new variant does not appear in that channel. This is because `createSingle` in `ProductVariantService` only assigns the variant to the current channel + default channel via `assignToCurrentChannel()`, without checking what other channels the parent product belongs to.

## Root Cause

There is an asymmetry in channel assignment:

- **`assignProductsToChannel`** assigns the product AND all its existing variants, option groups, and options to the target channel.
- **`createSingle` (variant creation)** only assigns the new variant to `ctx.channelId` + default channel — ignoring any other channels the parent product is already assigned to.

## Fix

After creating the variant and its prices for the current/default channel, we now:

1. Load the parent product's channels (using `getRepository().findOne()` with `relations: ['channels']` and `relationLoadStrategy: 'query'` — same pattern as `assignProductsToChannel`).
2. Filter out channels already handled (current + default).
3. For each additional channel:
   - Assign the variant via `channelService.assignToChannels()`.
   - Create a `ProductVariantPrice` via `createOrUpdateProductVariantPrice()`.
   - Assign the variant's option groups and options via `channelService.assignToChannels()` — matching what `assignProductsToChannel` does.

All methods used are existing public APIs already called elsewhere in the same file. No new patterns introduced.

### Price handling

The new variant's price in additional channels uses `input.price` directly (priceFactor = 1). The original `priceFactor` from the initial `assignProductsToChannel` call is a one-time conversion factor that is not stored anywhere, so it cannot be retroactively applied to new variants. The admin can adjust channel-specific prices after creation. This is strictly better than the current behavior where the variant is completely invisible in the other channel.

### Option/OptionGroup handling

When a new option is created after the product was assigned to a channel (e.g. adding a new color), `ProductOptionService.create()` only assigns it to the current + default channel via `assignToCurrentChannel()`. The fix also assigns the new variant's options and their option groups to the additional channels, preventing `Cannot return null for non-nullable field ProductOption.group` errors.

## Scope

This fix is specific to `ProductVariant` creation. The same gap exists for other parent-child relationships:

| Parent -> Child | Same gap? | Impact |
|---|---|---|
| **Product -> ProductVariant** | **Fixed in this PR** | High |
| Facet -> FacetValue | Yes | New facet values created after facet is assigned to a channel won't appear there |
| ProductOptionGroup -> ProductOption | Yes | New options won't appear in assigned channels |
| Collection -> Child Collection | Yes | Child collections created after parent is assigned won't appear there |

All of these use the same `assignToCurrentChannel()` pattern on creation without checking the parent's channels.

**Question for maintainers:** Would you prefer a common helper/pattern that handles this for all parent-child relationships, or is scoping to `ProductVariant` sufficient for now?

# Breaking changes

None. This is purely additive — new variants now get assigned to more channels than before. Existing behavior for variants created in products that are only in the default channel is unchanged (the additional channels list is empty, so no extra work is done).

# Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4699"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/product-channel.e2e-spec.ts
+++ b/packages/core/e2e/product-channel.e2e-spec.ts
@@ -11,12 +11,16 @@ import { initialData } from '../../../e2e-common/e2e-initial-data';
 import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-config';
 
 import { channelFragment, productVariantFragment } from './graphql/fragments-admin';
+import { graphql } from './graphql/graphql-admin';
 import {
+    addOptionGroupToProductDocument,
     assignProductToChannelDocument,
     assignProductVariantToChannelDocument,
     createAdministratorDocument,
+    createAssetsDocument,
     createChannelDocument,
     createProductDocument,
+    createProductOptionGroupDocument,
     createProductVariantsDocument,
     createRoleDocument,
     getChannelsDocument,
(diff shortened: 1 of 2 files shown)


grolmus (changes requested):
Thanks for tackling this, @Ryrahul — the bug in #4532 is real and your e2e nicely captures the symptom. Before this can go into core I'd like to see it reworked to build on the existing machinery rather than adding a parallel path.

**Reuse `assignProductVariantsToChannel`.** `ProductVariantService.assignProductVariantsToChannel` already handles everything a new variant needs when it enters another channel: the permission check (`UpdateCatalog` on the target channel), correct price derivation from the channel's `pricesIncludeTax` plus `priceFactor`, `defaultCurrencyCode`, stock-level seeding, asset assignment, and the `ProductVariantChannelEvent`. `ProductService.assignProductsToChannel` is the model for also propagating option groups/options. The current inline implementation in `createSingle` reproduces a subset of this and misses several pieces:

- **Pricing:** it writes the raw `input.price` as the net price into every channel, ignoring tax-inclusive channels, per-channel currency, and `priceFactor`. Please derive the price the same way `assignProductVariantsToChannel` does.
- **Permissions:** there's no check that the user may write to the other channels — please don't bypass `userHasPermissionOnChannel`. As written, a user with catalog rights only on the current channel could cause writes into channels they can't administer.
- **Events:** no `ProductVariantChannelEvent` is published, so search indexing / plugins won't learn the variant now exists in the other channels — meaning the variant may not actually become searchable there (i.e. the fix may not fully deliver on its goal).
- **Stock & assets:** no stock-level seeding (see #4864) and no asset assignment for the new channels, so `stockLevels` resolves to `[]` and a featured asset is invisible in the newly-covered channel.

Concretely, after the variant is created you should be able to compute the product's "other" channel IDs and route them through the existing assign method (or a shared helper extracted from it) rather than hand-rolling the loop.

A couple of practical notes:
- The branch currently **conflicts with `master`** and needs a rebase — `createSingle` has changed around the price-seeding block.
- Once pricing goes through the canonical path, the `priceWithTax === 2000` assertion will need recomputing, and please double-check the query fragment actually selects `currencyCode`.
- Could you extend the e2e to cover stock levels + assets in the assigned channel, a `priceFactor` / tax-inclusive pricing case, and a negative-permission case?

Happy to help map out the refactor if useful — thanks again for the contribution.

Ryrahul:
Thanks @grolmus for the review !. Will address the mentioned changes 👐

grolmus (commented):
Excellent rework, @Ryrahul — this now builds on `assignProductVariantsToChannel` exactly as hoped, so pricing (incl. tax-inclusive channels), permissions, stock seeding, assets and `ProductVariantChannelEvent` all come for free and stay consistent with `ProductService.assignProductsToChannel`. The option-group/option propagation is the right call, and the `@Transaction()` wrapper means a `ForbiddenError` on a target channel cleanly rolls back the create — nicely covered by the new negative-permission test. The extended e2e (tax-inclusive price, stock, assets, options, permissions) is exactly what I was hoping for.

A few notes — one worth tightening before I approve, the rest optional:

- **Stock test**: it asserts `stockOnHand === 0`, which is what you'd get whether or not a `StockLevel` row was actually seeded for the channel. Could you assert on the channel-filtered `stockLevels` array (its length / entry) instead? That's what actually proves `ensureStockLevelsForChannel` did its job — the behaviour this test is meant to guard.
- Optional: the comment around the price test says "priceWithTax (2000)" — 2000 is the net price; the gross is 2400 (which the assertion correctly checks). Just the wording.
- Optional: after the permission failure, assert that no orphaned `CHAN-VAR-GREEN` variant exists. The `@Transaction()` already guarantees the rollback, but it makes the intent explicit.

Once the stock assertion is tightened I'm happy to approve — the implementation itself looks great.

Ryrahul:
> Excellent rework, @Ryrahul — this now builds on `assignProductVariantsToChannel` exactly as hoped, so pricing (incl. tax-inclusive channels), permissions, stock seeding, assets and `ProductVariantChannelEvent` all come for free and stay consistent with `ProductService.assignProductsToChannel`. The option-group/option propagation is the right call, and the `@Transaction()` wrapper means a `ForbiddenError` on a target channel cleanly rolls back the create — nicely covered by the new negative-permission test. The extended e2e (tax-inclusive price, stock, assets, options, permissions) is exactly what I was hoping for.
> 
> A few notes — one worth tightening before I approve, the rest optional:
> 
> * **Stock test**: it asserts `stockOnHand === 0`, which is what you'd get whether or not a `StockLevel` row was actually seeded for the channel. Could you assert on the channel-filtered `stockLevels` array (its length / entry) instead? That's what actually proves `ensureStockLevelsForChannel` did its job — the behaviour this test is meant to guard.
> * Optional: the comment around the price test says "priceWithTax (2000)" — 2000 is the net price; the gross is 2400 (which the assertion correctly checks). Just the wording.
> * Optional: after the permission failure, assert that no orphaned `CHAN-VAR-GREEN` variant exists. The `@Transaction()` already guarantees the rollback, but it makes the intent explicit.
> 
> Once the stock assertion is tightened I'm happy to approve — the implementation itself looks great.

Hi @grolmus, I've made the requested changes. Could you please review it again when you have a moment? Thanks for your time!

grolmus (approved):
Thanks @Ryrahul — the stock test now asserts on the channel-filtered `stockLevels` (proving the row was actually seeded, not just that `stockOnHand === 0`), and you added the orphaned-variant check after the permission-denied create. Both exactly what I was after.

Heads-up: the red `e2e (sqljs)` is an unrelated harness flake — `EEXIST: mkdir __data__` race in `SqljsInitializer` while setting up `order.e2e-spec.ts`; it passes on mariadb/mysql/postgres. A re-run should clear it.

Approving — nice work. 🙌

PR #4704, opened 2026-05-06
fix(core): Fix race condition in manual task trigger

martijnvdbrug:
## Problem

`runScheduledTask` mutation returns `success: true` but the task only executes ~50% of the time. The `manuallyTriggeredAt` flag is cleared in `checkForManuallyTriggeredTasks` **before** `executeTask` attempts to acquire the lock. If a concurrent cron execution holds the lock at that moment, `tryAcquireLock` fails silently and the trigger is permanently lost with no retry.

## Fix

Move the `manuallyTriggeredAt = null` clearing into `tryAcquireLock` itself, so the flag is only cleared atomically when the lock is successfully acquired. If the lock fails (cron holds it), the flag remains set and the next polling cycle retries.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4704"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts
+++ b/packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts
@@ -200,10 +200,6 @@ export class DefaultSchedulerStrategy implements SchedulerStrategy {
         Logger.debug(`Checking for manually triggered tasks: ${taskEntities.length}`);
 
         for (const taskEntity of taskEntities) {
-            await this.connection.rawConnection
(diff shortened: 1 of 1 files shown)


martijnvdbrug:
Too soon, this is not the fix

PR #4712, opened 2026-05-09
fix: Prevent facet value chip remove button from being hidden on long…

### LINKED ISSUE #4711: React dashboard Facet Values close button be hidden
Image

Ryrahul:
# Description

Fixes #4711

When a facet value name is very long (e.g. "Sports & Outdoor & Swimming & Gym & Fashion..."), the × remove button gets clipped and becomes invisible. This is because the Badge base component uses `shrink-0` and `overflow-hidden`, so the text pushes the button outside the visible area.

This PR constrains the chip to its parent width and truncates long names with an ellipsis, ensuring the remove button is always visible.

**Changes:**
- `facet-value-chip.tsx`: Added `max-w-full shrink` on Badge, `min-w-0` on text container, `truncate` on name span, `shrink-0` on remove button
- `assigned-facet-values.tsx`: Added `max-w-full` on chip wrapper div

# Breaking changes

None.

# Screenshots

Before:
image

After:
image

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4712"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/shared/assigned-facet-values.tsx
+++ b/packages/dashboard/src/lib/components/shared/assigned-facet-values.tsx
@@ -45,7 +45,7 @@ export function AssignedFacetValues({
                     const facetValue = knownFacetValues.find(fv => fv.id === id);
                     if (!facetValue) return null;
                     return (
-                        <div className="mb-2 mr-1" key={facetValue.id}>
+                        <div className="mb-2 mr-1 max-w-full" key={facetValue.id}>
                             <FacetValueChip
                                 facetValue={facetValue}
                                 removable={canUpdate}
--- a/packages/dashboard/src/lib/components/shared/facet-value-chip.tsx
+++ b/packages/dashboard/src/lib/components/shared/facet-value-chip.tsx
@@ -36,10 +36,10 @@ export function FacetValueChip({
     return (
         <Badge
             variant="secondary"
-            className="flex items-center gap-2 py-0.5 pl-2 pr-1 h-6 hover:bg-secondary/80"
+            className="flex items-center gap-2 py-0.5 pl-2 pr-1 h-6 max-w-full shrink hover:bg-secondary/80"
         >
-            <div className="flex items-center gap-1.5">
-                <span className="font-medium">{facetValue.name}</span>
+            <div className="flex items-center gap-1.5 min-w-0">
+                <span className="font-medium truncate">{facetValue.name}</span>
                 {displayFacetName && (
                     <span className="text-muted-foreground text-xs">in {facetValue.facet.name}</span>
                 )}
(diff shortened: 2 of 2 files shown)


Ryrahul:
Before the migration from Radix to Base UI, long facet names would scroll. After the migration, overflow-hidden clips the × button instead.

grolmus (approved):
Clean focused fix, thanks @Ryrahul. The flex `min-w-0` + `truncate` + `shrink-0` on the button is exactly the right Tailwind pattern for "text shrinks, button stays". Screenshots confirm the visual result.

I confirmed both call sites — `assigned-facet-values.tsx` and `data-input/facet-value-input.tsx` — render inside `flex flex-wrap` parents, so the `max-w-full` on the chip itself handles the second usage even though only the first got a wrapper update. Good.

CodeRabbit's earlier nit (add `truncate` to the facet-name span on line 44) is already addressed in the current diff. Aria-label still uses the full untruncated name, so accessibility is preserved.

### Optional, non-blocking

Consider adding `title={facetValue.name}` (or a tooltip) on the truncated name span, so hovering a clipped chip surfaces the full name. Right now once truncated, a user can't read the value without removing it and re-selecting. Strictly optional — could be a follow-up.

Approving.

Also closes Linear OSS-525 when this merges.

Ryrahul:
> Clean focused fix, thanks @Ryrahul. The flex `min-w-0` + `truncate` + `shrink-0` on the button is exactly the right Tailwind pattern for "text shrinks, button stays". Screenshots confirm the visual result.
> 
> I confirmed both call sites — `assigned-facet-values.tsx` and `data-input/facet-value-input.tsx` — render inside `flex flex-wrap` parents, so the `max-w-full` on the chip itself handles the second usage even though only the first got a wrapper update. Good.
> 
> CodeRabbit's earlier nit (add `truncate` to the facet-name span on line 44) is already addressed in the current diff. Aria-label still uses the full untruncated name, so accessibility is preserved.
> 
> ### Optional, non-blocking
> Consider adding `title={facetValue.name}` (or a tooltip) on the truncated name span, so hovering a clipped chip surfaces the full name. Right now once truncated, a user can't read the value without removing it and re-selecting. Strictly optional — could be a follow-up.
> 
> Approving.
> 
> Also closes Linear OSS-525 when this merges.

Thanks @grolmus  for the detailed review. I have added  ``` title={facetValue.name}``` as suggested

michaelbromley (approved):

PR #4718, opened 2026-05-11
fix(core): Refuse default superadmin credentials in production

grolmus:
## Summary

Vendure ships with the well-known defaults `superadmin` / `superadmin` for the initial admin user and has no built-in guard against running with them in production. This is a hard-coded-credentials issue (CWE-798): anyone with network access to a vanilla 3.6.x deployment can log in as superadmin if the operator did not override `authOptions.superadminCredentials`.

This PR adds a bootstrap-time check that:

- **Refuses to start** the server when `NODE_ENV === 'production'` and the password matches the default constant from `@vendure/common`. The default `superadmin` identifier is intentionally allowed — only a default password is treated as insecure.
- **Logs a prominent `Logger.warn`** with remediation hints in development / staging / any other non-test environment.
- **Stays silent in `NODE_ENV === 'test'`** so the existing e2e suite (which uses the defaults by design) keeps a clean output.

The remediation hint points operators at wiring `superadminCredentials` from environment variables, which is the same pattern most starters already use.

## Scope

- No schema migration, no public API change.
- The well-known constants in `@vendure/common` are kept as-is — they are still exported because plugins and the e2e/testing layer rely on them.
- A more invasive follow-up (auto-generated random password on first init, forced password change on first login) is intentionally left for a future `major` change. This PR is the minimum-blast-radius patch that closes the critical concern.
- **Targets `minor`**: refusing to start in production on a default password is a behavioural change, so it ships in a minor release rather than a patch.

## Changes

- New helper `packages/core/src/service/helpers/utils/check-superadmin-credentials.ts` containing the pure check (testable, no DI).
- `AdministratorService.ensureSuperAdminExists` now calls the helper before creating / reconciling the superadmin user.
- 7 new unit tests in the paired spec file covering production / development / staging / test / safe-credentials paths.

## Test plan

- [x] `vitest run service/helpers/utils/check-superadmin-credentials.spec.ts` — 7/7 pass
- [x] Full `@vendure/core` unit suite (`vitest run`) — 819/819 pass; no existing tests regressed
- [ ] Manual smoke: boot dev server with default config (NODE_ENV unset) → expect a single `Logger.warn` line about the default password, server starts normally
- [ ] Manual smoke: boot with `NODE_ENV=production` and default password → expect bootstrap to throw `[Vendure] Refusing to start: Default superadmin password is configured…`
- [ ] Manual smoke: boot with `NODE_ENV=production` and explicit override → silent, server starts normally

## Notes

- Reported privately via responsible disclosure (internal Linear OSS-490 — Pratik Karan, 2026-04-03). No public GitHub issue exists.
- Only the default **password** is treated as insecure. Using the well-known `superadmin` identifier with a strong, non-default password is fully supported (per maintainer review). The guard cannot detect weak-but-custom passwords — enforcing "non-default" is the enforceable boundary; "strong" remains the operator's responsibility.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4718"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/src/service/helpers/utils/check-superadmin-credentials.spec.ts
+++ b/packages/core/src/service/helpers/utils/check-superadmin-credentials.spec.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { checkSuperadminCredentials } from './check-superadmin-credentials';
+
+const DEFAULTS = { identifier: 'superadmin', password: 'superadmin' };
+const SAFE = { identifier: 'admin@example.com', password: 'a-very-long-random-passphrase' };
+
+describe('checkSuperadminCredentials', () => {
+    it('throws in production when both identifier and password are default', () => {
+        expect(() => checkSuperadminCredentials(DEFAULTS, { nodeEnv: 'production' })).toThrow(
+            /Refusing to start/,
+        );
+    });
+
+    it('throws in production when only the password is default', () => {
+        expect(() =>
+            checkSuperadminCredentials(
+                { identifier: 'admin@example.com', password: 'superadmin' },
+                { nodeEnv: 'production' },
+            ),
+        ).toThrow(/Refusing to start/);
+    });
+
+    it('throws in production when only the identifier is default', () => {
+        expect(() =>
+            checkSuperadminCredentials(
+                { identifier: 'superadmin', password: 'something-strong-1234' },
+                { nodeEnv: 'production' },
+            ),
+        ).toThrow(/Refusing to start/);
+    });
+
+    it('warns in development when defaults are used', () => {
+        const logger = { warn: vi.fn() };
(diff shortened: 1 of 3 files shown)


michaelbromley (changes requested):
Hi Martin, this is generally a good idea but a couple of changes needed:

1. We should not disallow the `superadmin` identifier. It’s totally fine to use that, as long as there’s a non-default, strong password.
2. This must target the `minor` branch. It is a breaking change in behaviour.

grolmus:
all done

michaelbromley (approved):

Issue #15, opened 2018-09-04
Remove Auth REST endpoint, use GraphQL


michaelbromley:
Currently, auth is still handled via a REST call. The AuthController should be replaced by a GraphQL resolver.

Issue #11, opened 2018-08-27
API tests


michaelbromley:
Currently we have Jest unit tests for the server, but no tests for the API itself (i.e. tests which make GraphQL API calls and assert the response)

Now that the general API parts already implemented are (hopefully) somewhat stable, it makes sense to set up API testing infrastructure.

This would involve:

* A method of setting up a pre-defined test database. Look into using sql.js (which is supported by TypeORM) to allow all DB operations to be performed in-memory, which should ensure the tests run reasonably quickly.
* Setting up the test runner to make real HTTP calls and asserting on results. Jest snapshots would probably make sense for this. Look into [supertest](https://www.npmjs.com/package/supertest) for making the HTTP calls.
* Setting up Jenkins to run the tests on push.
* Writing tests to cover all existing queries and mutations.

Issue #10, opened 2018-08-24
Upgrade to Apollo Server 2


michaelbromley:
We are currently on 1.3.6. v2 is broadly backwards-compatible but since we are using the [Nest GraphQLModule](https://github.com/nestjs/graphql), we need to wait until v2 is properly supported.

There is an thread in which [some people get it to work](https://github.com/nestjs/graphql/issues/32#issuecomment-412048317), but there currently seems to be an issue with the Passport.js integration.

Best to wait for a bit to see if these issues get resolved.

michaelbromley:
This was done in 51ddd7c84ab393a357833201de2ec4588e98baf6

PR #4721, opened 2026-05-12
fix(core): Block SSRF in DefaultAssetImportStrategy

grolmus:
## Summary

`DefaultAssetImportStrategy.getStreamFromUrl` previously fetched any URL supplied by an admin without checking what host it actually resolves to. Combined with Node's permissive `http.get`, this gave an authenticated admin a free SSRF primitive: pointing the importer at e.g. `http://169.254.169.254/latest/meta-data/` extracts cloud IAM tokens, and any hostname pointing at a loopback / RFC 1918 / link-local address could probe internal services that trust same-host traffic (CWE-918, CVSS 8.6 per the disclosure).

## Fix

A new pre-flight `assertPublicUrl(...)` runs before each fetch:

1. Reject non-`http(s)` schemes (defensive — `getStreamFromPath` already filters via regex, but this catches future callers and explicit `file://` / `gopher://` attempts).
2. Resolve the hostname (or use the literal if it's already an IP) and reject when ANY resolved address lands in a private, loopback, link-local, CGNAT or unspecified range. Uses `net.BlockList` so both IPv4 and IPv6 ranges are covered (`0.0.0.0/8`, `10.0.0.0/8`, `100.64.0.0/10`, `127.0.0.0/8`, `169.254.0.0/16`, `172.16.0.0/12`, `192.0.0.0/24`, `192.168.0.0/16`, `198.18.0.0/15`, `::`, `::1`, `fc00::/7`, `fe80::/10`).
3. Resolving multiple addresses → all must be public; ANY private address rejects (defends against split-horizon DNS).

An opt-in `allowPrivateNetworks?: boolean` constructor option lets operators bypass the block when importing from a trusted internal asset server or running locally. The protocol check still applies even with the bypass.

## Why all-private (not just metadata IP)

The disclosure focuses on `169.254.169.254`, but the actual threat surface is wider: SSRF against any internal address can probe Redis, ES, Postgres admin endpoints, sidecar containers, the Vendure admin port itself, etc. Blocking the whole private-network bucket is the standard mitigation (matches AWS's IMDSv2 hardening guidance, GitHub's webhook validator, Stripe's webhook block-list).

## Test plan

- [x] 14 new unit tests in `assert-public-url.spec.ts` — every range, scheme, DNS path covered, including the all-private bypass and DNS-failure paths
- [x] Full `@vendure/core` unit suite — 826 tests pass
- [ ] Manual: configure a Vendure server, attempt to import an asset from `http://169.254.169.254/latest/meta-data/` → expect a clear refusal error and no outbound request
- [ ] Manual: same with `allowPrivateNetworks: true` → request proceeds (errors come from the actual fetch behaviour, not from the validator)
- [ ] Manual: legitimate public CDN URL works exactly as before

## Notes

- Reported privately via responsible disclosure (internal Linear OSS-487 — Pratik Karan, 2026-04-03). No public GitHub issue exists.
- The strategy options were previously typed with required `retryDelayMs` / `retryCount`; the PR loosens them to optional (with the same defaults at the use site) to keep the new option additive without forcing existing callers to pass retry settings they didn't pass before.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4721"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/import.e2e-spec.ts
+++ b/packages/core/e2e/import.e2e-spec.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
 import { omit } from '@vendure/common/lib/omit';
-import { User } from '@vendure/core';
+import { DefaultAssetImportStrategy, User } from '@vendure/core';
 import { createTestEnvironment } from '@vendure/testing';
 import * as fs from 'node:fs';
 import http from 'node:http';
@@ -14,8 +14,17 @@ import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-conf
 import { graphql } from './graphql/graphql-admin';
 
 describe('Import resolver', () => {
+    const baseConfig = testConfig();
     const { server, adminClient } = createTestEnvironment({
-        ...testConfig(),
+        ...baseConfig,
+        importExportOptions: {
+            ...baseConfig.importExportOptions,
+            // The "asset urls" suite below spins up a local static server on
+            // localhost:3456 and imports from it. The default strategy blocks
+            // loopback/private IPs to mitigate SSRF (see assert-public-url.ts),
+            // so the test needs the documented escape hatch enabled.
+            assetImportStrategy: new DefaultAssetImportStrategy({ allowPrivateNetworks: true }),
+        },
         customFields: {
             Product: [
                 { type: 'string', name: 'pageType' },
--- a/packages/core/src/config/asset-import-strategy/assert-public-url.spec.ts
+++ b/packages/core/src/config/asset-import-strategy/assert-public-url.spec.ts
@@ -1,0 +1,267 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { InternalServerError } from '../../common/error/errors';
+import { Logger } from '../logger/vendure-logger';
+
+import { assertPublicUrl, DnsResolverFn } from './assert-public-url';
+
+const makeResolver = (records: Array<{ address: string; family: number }>): DnsResolverFn =>
+    vi.fn().mockResolvedValue(records);
+
+describe('assertPublicUrl', () => {
+    let warnSpy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(() => {
+        warnSpy = vi.spyOn(Logger, 'warn').mockImplementation(() => undefined);
+    });
+
+    afterEach(() => {
+        warnSpy.mockRestore();
+    });
+
+    describe('accepts', () => {
+        it('a public IPv4 literal and returns the pinned address', async () => {
+            const result = await assertPublicUrl('https://1.1.1.1/img.jpg');
+            expect(result.url.host).toBe('1.1.1.1');
+            expect(result.pinned).toEqual({ address: '1.1.1.1', family: 4 });
+            expect(warnSpy).not.toHaveBeenCalled();
+        });
+
+        it('a hostname that resolves to a public IPv4 and pins to the resolved IP', async () => {
+            const result = await assertPublicUrl('https://cdn.example.com/img.jpg', {
+                resolver: makeResolver([{ address: '93.184.216.34', family: 4 }]),
+            });
+            expect(result.pinned).toEqual({ address: '93.184.216.34', family: 4 });
+        });
+
+        it('a public IPv6 literal', async () => {
+            const result = await assertPublicUrl('https://[2606:4700::1111]/x');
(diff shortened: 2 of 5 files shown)


grolmus:
Pushed a follow-up commit (40a7394) addressing all of the deep-review findings.

## Security correctness

- **IPv4-mapped IPv6 bypass (CodeRabbit + hex-security HIGH).** `[::ffff:127.0.0.1]`, `[::ffff:169.254.169.254]` etc. now blocked. The naive `bl.addSubnet('::ffff:0:0', 96, 'ipv6')` approach causes Node's `BlockList` to false-positive every plain IPv4 lookup (it projects mapped addresses into the v4 space), so the v6→v4 projection is done explicitly in a new `isBlocked()` helper. Hex-form mapped addresses (`::ffff:7f00:1`) are also handled.
- **Extra IPv6 SSRF vectors (Nigel).** Added `2002::/16` (6to4), `2001::/32` (Teredo), `64:ff9b::/96` (NAT64) and `100::/64` (discard) — each is a known tunnelling vector into the v4 private space.
- **DNS rebinding TOCTOU (hex-security MEDIUM, Nigel).** `assertPublicUrl` now returns the validated `PinnedAddress` and `fetchUrl` connects directly to that IP, preserving the original hostname in the `Host` header and (for HTTPS) the TLS `servername`. Eliminates the second DNS lookup that `http.get` would otherwise perform.
- **IPv6 zone-id stripping + bracket handling (Nigel).** `fe80::1%eth0` is now stripped to `fe80::1` before validation. The WHATWG `URL` keeps `[ ]` around IPv6 literals in `.hostname`; we now strip them properly.
- **`family` normalisation bug (Nigel).** Was using `family: isIP(hostname)` raw; now explicit `4 | 6` with a defensive reject of `isIP === 0`.

## Error / log pipeline

- **`InternalServerError` instead of bare `Error` (Vendure-expert, Nigel).** Goes through the i18n/GraphQL pipeline like every other throw in the asset-strategy folder.
- **Generic thrown message + detailed `Logger.warn` (Nigel).** Threw error no longer leaks the URL, resolved IP, or remediation hint (which used to advertise the `allowPrivateNetworks` bypass to the client). Details now go to `Logger.warn` with a `'DefaultAssetImportStrategy'` context so operators can alert on them.
- **Log-forgery defence (Nigel).** Truncate + strip control characters from `urlString` before interpolating into log messages. Closes CRLF-in-URL log-line injection.

## API quality

- Extracted `DefaultAssetImportStrategyOptions` as an exported interface (Nigel #3).
- Renamed `Resolver` → `DnsResolverFn` (Nigel #11) — no collision with `dns.Resolver`.
- Dropped the `await import('dns/promises')` lazy import (Nigel #10) — theatre, no test benefit.
- Documented in `fetchUrl` that adding redirect-following would re-open SSRF.
- `@since 3.6.4` JSDoc on `allowPrivateNetworks`.

## Tests

34 unit tests pass (up from 14), full `@vendure/core` suite (846 tests) still green. Coverage now includes:

- IPv4-mapped IPv6 (literal + hex form, loopback / metadata / RFC 1918)
- 6to4, Teredo, NAT64, discard, unique-local, link-local
- Zone-id stripping
- Uppercase IPv6 hex
- Public IPv6 acceptance + `PinnedAddress` return value
- Generic-error contents (does not leak URL / `allowPrivateNetworks`)
- Log-forgery defence (CRLF stripping, length truncation)
- `allowPrivateNetworks` bypass paths

## Reviewer notes

- I confirmed there is no existing URL-validation helper in `packages/core/src/common/utils.ts` (Nigel #15) — nothing to reuse.
- `assertPublicUrl` and `DefaultAssetImportStrategyOptions` are intentionally **not** exported from any barrel (Vendure-expert #5) — keep the SSRF policy internal to the default strategy.
- The CodeRabbit `Docstring Coverage 50%` pre-merge check should now flip green given the added JSDoc on `DefaultAssetImportStrategyOptions`, `PinnedAddress`, `DnsResolverFn` and `fetchUrl`.

grolmus:
Squashed the original commit and the review follow-up into a single commit (`caf2063`) for a cleaner history. The combined diff is identical to what was here before; nothing further changed since the last force-push. All 34 unit tests + the full `@vendure/core` suite (846) still pass, and I re-ran an end-to-end smoke against real network (`httpbin.org/headers`) — 10/10: every blocked range rejects with the generic message, and the pinned-IP fetch correctly preserves the `Host` header for both HTTP and HTTPS public targets.

PR #4171, opened 2026-01-28
feat(core)!: Make Asset entity translatable

biggamesmallworld:
# Description

Makes the Asset entity translatable, allowing asset names to be localized per language.

### Core (`packages/core`)

**New Entity:**
- `AssetTranslation` - Translation entity with `name` and `customFields` columns

**Entity Updates:**
- `Asset` - Now implements `Translatable` interface with `translations` relation
- `custom-entity-fields.ts` - Added `CustomAssetFieldsTranslation` class

**Service Updates:**
- `AssetService` - Updated to use `TranslatableSaver` for handling translation persistence
- `TranslationDiffer` - Fixed to set `baseId` directly on new translations for proper foreign key persistence

**GraphQL Schema:**
- Added `languageCode` and `translations` fields to `Asset` type
- Added `AssetTranslation` type
- Added `AssetTranslationInput` input type
- Added `translations` field to `CreateAssetInput` and `UpdateAssetInput`

### Dashboard (`packages/dashboard`)

- `assets.graphql.ts` - Updated mutation to return full Asset fragment including translations
- `translatable-form-field.tsx` - Fixed to use `useFormContext().watch()` for reactive form values and added key prop for proper Controller remounting on language change

# Breaking changes
This PR makes the `Asset` entity translatable by adding a new `asset_translation` table.

  ### Migration steps

  After updating to this version:

  1. **Generate the migration:**
     ```bash
     npx vendure migrate --generate add-asset-translations

  2. Review the generated migration - It will:
    - Create the asset_translation table
    - Migrate existing name values from the asset table to asset_translation
    - Remove the name column from the asset table
  3. Run the migration:
  npx vendure migrate --run

  API changes

  - The Asset type now includes a translations field
  - The name field on UpdateAssetInput is deprecated - use the translations array instead:
  ```
  mutation {
    updateAsset(input: {
      id: "1"
      translations: [
        { languageCode: EN, name: "My Asset" }
        { languageCode: DE, name: "Mein Asset" }
      ]
    }) {
      id
      name
    }
  }
```
  - Using the deprecated name field will log a warning but continue to work for backward
  compatibility

# Screenshots


https://github.com/user-attachments/assets/c0da06da-5a3e-443d-88ce-d1192ef59da2


# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have checked my own PR

👍 Most of the time:
- [x] I have added or updated test cases
- [x] I have updated the README if needed

PR #4723, opened 2026-05-12
fix(core): Default languageCode in RequestContext.empty()

### LINKED ISSUE #4651: AssetService.createFromFileStream() fails without RequestContext because of missing languageCode

grolmus:
## Summary

- `RequestContext.empty()` now defaults `languageCode` to `LanguageCode.en` so it produces a runtime-valid context. The `languageCode` getter is typed as `LanguageCode` (non-optional) but returned `undefined` at runtime when callers used `empty()`.
- Fixes `AssetService.createFromFileStream()` failing with `NOT NULL constraint failed: asset_translation.languageCode` when called without an explicit `RequestContext` (regression introduced in v3.6 by #4171 making `Asset` translatable).

## Root cause

`RequestContext.empty()` builds a context with `new Channel()` (no `defaultLanguageCode`). The constructor falls back to `channel.defaultLanguageCode` → `undefined`. Any downstream code that relies on `ctx.languageCode` for a NOT NULL column then crashes.

In `AssetService.createAssetInternal()` the default-translation path uses `ctx.languageCode` directly, so `createFromFileStream(stream, filename)` (no `ctx` arg) inserts `NULL` into `asset_translation.languageCode`.

## Test plan

- [x] Unit test in `request-context.spec.ts` pinning `RequestContext.empty().languageCode === LanguageCode.en`
- [x] E2E test in `asset.e2e-spec.ts` that reproduces the exact bug from the issue: calls `AssetService.createFromFileStream(stream, 'oss-499.pdf')` without a `RequestContext` and verifies the resulting `AssetTranslation` row has `languageCode: 'en'`. Confirmed to fail on `master` and pass with this change.
- [x] All 23 tests in `asset.e2e-spec.ts` pass (sqljs)
- [x] All 8 spec files that use `RequestContext.empty()` continue to pass (62 tests)

Fixes #4651

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4723"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/asset.e2e-spec.ts
+++ b/packages/core/e2e/asset.e2e-spec.ts
@@ -1,10 +1,16 @@
-import { DeletionResult, LogicalOperator, SortOrder } from '@vendure/common/lib/generated-types';
+import {
+    DeletionResult,
+    LanguageCode,
+    LogicalOperator,
+    SortOrder,
+} from '@vendure/common/lib/generated-types';
 import { omit } from '@vendure/common/lib/omit';
 import { pick } from '@vendure/common/lib/pick';
-import { mergeConfig } from '@vendure/core';
+import { Asset, AssetService, AssetTranslation, mergeConfig, TransactionalConnection } from '@vendure/core';
 import { createErrorResultGuard, createTestEnvironment, ErrorResultGuard } from '@vendure/testing';
 import fs from 'fs-extra';
 import path from 'node:path';
+import { Readable } from 'node:stream';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import { initialData } from '../../../e2e-common/e2e-initial-data';
@@ -542,6 +548,33 @@ describe('Asset resolver', () => {
             productGuard.assertSuccess(product);
             expect(product.featuredAsset).toBeNull();
             expect(product.assets.length).toEqual(0);
+        });
+    });
+
+    // https://github.com/vendurehq/vendure/issues/4651
+    describe('createFromFileStream without RequestContext', () => {
+        it('uses default languageCode for the asset translation', async () => {
+            const assetService = server.app.get(AssetService);
+            const connection = server.app.get(TransactionalConnection);
+
+            const stream = new Readable();
+            stream.push('oss-499 content');
+            stream.push(null);
+
+            const result = await assetService.createFromFileStream(stream, 'oss-499.pdf');
+
+            if (!(result instanceof Asset)) {
+                throw new Error(`Expected Asset, got ${JSON.stringify(result)}`);
+            }
+            expect(result.id).toBeDefined();
+            expect(result.name).toBe('oss-499.pdf');
+
+            const translations = await connection.rawConnection
+                .getRepository(AssetTranslation)
+                .find({ where: { base: { id: result.id } } });
+            expect(translations.length).toBe(1);
+            expect(translations[0].languageCode).toBe(LanguageCode.en);
+            expect(translations[0].name).toBe('oss-499.pdf');
         });
     });
 });
--- a/packages/core/src/api/common/request-context.spec.ts
+++ b/packages/core/src/api/common/request-context.spec.ts
@@ -130,6 +130,13 @@ describe('RequestContext', () => {
         });
     });
 
+    describe('empty', () => {
+        // https://github.com/vendurehq/vendure/issues/4651
+        it('languageCode defaults to LanguageCode.en', () => {
+            expect(RequestContext.empty().languageCode).toBe(LanguageCode.en);
+        });
+    });
+
     describe('userHasPermissions', () => {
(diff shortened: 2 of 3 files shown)


kevmtt:
Duplicate with: https://github.com/vendurehq/vendure/pull/4662

- I used an approach to introduce a new method though that is using the language code from the default channel

playbeatarena-rgb:
Teri ma ki kus stop mailing me

grolmus:
Closing in favor of #4662, which I missed when starting this work — apologies for the duplicate.

After comparing, #4662 is the better approach:
- Correctly derives `languageCode` from the default channel (`channel.defaultLanguageCode ?? configService.defaultLanguageCode`), whereas this PR hardcodes `LanguageCode.en` — wrong for installations whose default language isn't English.
- Adds a proper `RequestContextService.createDefaultContext(config?)` factory that handles Admin/Shop API contexts and custom overrides, per @michaelbromley's review suggestion.
- Already endorsed by the maintainer and the contributor has incorporated review feedback.

Thanks @kevmtn for the original work.

PR #4732, opened 2026-05-13
fix(dashboard): Filter dashboard extensions to plugins active in runtime config

### LINKED ISSUE #4706: Dashboard discovers and bundles extensions for plugins not present in runtime config
## Summary

The dashboard's plugin extension discovery (Vite-side) is based on **static import-graph analysis** and is not cross-referenced against the runtime `VendureConfig.plugins` array. Any plugin reachable through static imports — directly from the user's config, or transitively via the `dependencies` field of an imported package — has its dashboard extensions bundled and registered in the dashboard runtime, even when that plugin is **not** present in the actual `plugins` array passed to `bootstrap()`.

This breaks any pattern where plugins are conditionally added to the runtime `plugins` array based on options, environment, or feature flags.

## Reproduction (minimal generalised form)

```ts
// my-plugin-bundle.ts
import { FooPlugin } from 'foo-plugin';
import { BarPlugin } from 'bar-plugin';

export function getPlugins(opts: { includeBar: boolean }) {
    return opts.includeBar ? [FooPlugin.init({}), BarPlugin.init({})] : [FooPlugin.init({})];
}
```

```ts
// vendure-config.ts
import { getPlugins } from './my-plugin-bundle';

    plugins: [
        ...getPlugins({ includeBar: false }),
        DashboardPlugin.init({...}),
    ],
};
```

Expected: dashboard does not load `BarPlugin`'s dashboard extension (nav items, routes, page blocks, etc.).

Actual: `BarPlugin`'s dashboard extension is bundled and runs on dashboard startup. Its UI surfaces appear as if it were active.

The same applies to a wrapper package (`my-plugin-bundle`) imported from `node_modules` whose `package.json` lists `bar-plugin` in `dependencies`: even if `bar-plugin` is never imported into source code reachable from the config, the dashboard discovers it via package.json expansion and bundles it.

## Root cause (in current dashboard code)

In `@vendure/dashboard`'s Vite pipeline:

- `vite/utils/plugin-discovery.ts` builds `pluginInfo` via two channels:
  1. `analyzeSourceFiles()` walks `ImportDeclaration` / `ExportDeclaration` nodes from `vendure-config.ts`, recording every npm package name encountered.
  2. `expandPackageImports()` reads each recorded package's `package.json` `dependencies` and adds any dep whose own deps include `@vendure/core`.
- `findVendurePluginFiles()` then globs `node_modules/<pkg>/**/*.js` for every package in that combined set and identifies plugins via the `dashboard:` field on `__decorate` metadata.
- `vite/vite-plugin-dashboard-metadata.ts` emits a `runDashboardExtensions()` virtual module that imports **all** discovered extension entry paths — without ever looking at `vendureConfig.plugins`.

The same unfiltered `pluginInfo` is consumed by `vite-plugin-translations.ts`, `vite-plugin-tailwind-source.ts`, and `vite-plugin-lingui-babel.ts`, so plugin translations and Tailwind sources are also pulled in for plugins that are not active at runtime.

The `vendureConfig` object itself **is** loaded by `vite-plugin-config-loader.ts` and is available alongside `pluginInfo` in `getVendureConfig()` results — it is just not used to filter `pluginInfo`.

### Filter `pluginInfo` against active plugins from `vendureConfig.plugins`

Add a single filter step before consuming `pluginInfo` in the four Vite plugins listed above. Sketch:

```ts
// shared helper
export function filterActivePluginInfo(
    pluginInfo: PluginInfo[],
    vendureConfig: VendureConfig,
): PluginInfo[] {
    const activeNames = new Set<string>();
    for (const entry of vendureConfig.plugins ?? []) {
        // Each entry is a class with @VendurePlugin or the result of Plugin.init(opts).
        // Both cases: the constructor's `name` is the class name (init() returns the class).
        const ctor = typeof entry === 'function' ? entry : entry?.constructor;
        if (ctor?.name) activeNames.add(ctor.name);
    }
    return pluginInfo.filter(p => activeNames.has(p.name));
}
```

Then in `vite-plugin-dashboard-metadata.ts`:

```ts
const { pluginInfo, vendureConfig } = await configLoaderApi.getVendureConfig();
const activePluginInfo = filterActivePluginInfo(pluginInfo, vendureConfig);
const pluginsWithExtensions = activePluginInfo.map(...);
```

Same change in `vite-plugin-translations.ts` (`getPluginTranslations`), `vite-plugin-tailwind-source.ts` (`getDashboardPaths`), and `vite-plugin-lingui-babel.ts`.

### Edge cases worth considering in the implementation

- **Plugin classes that wrap their constructor**: if `Plugin.init(opts)` ever returns something other than the class itself (e.g. a configured module object), the `name` lookup needs a fallback. A `static pluginName` convention or a metadata reflection lookup would be more robust than relying on `constructor.name`.
- **Anonymous / minified class names**: production builds of plugin packages sometimes mangle class names. The `pluginInfo[].name` is read from the `__decorate` call's second argument identifier in the compiled output, which is also subject to the same source. As long as both sides are read from the same compiled artefact, names should match — but this should be sanity-checked.
- **Backwards compat for users relying on the current behaviour**: unlikely there are any (the current behaviour is the bug), but a flag like `DashboardPlugin.init({ extensionFilter: 'active' | 'all' })` could be offered with default `'active'` and an escape hatch.

### Alternative / additional ideas

1. **Expose a discovery hook**: let plugin authors / wrapper packages provide a `pluginPackageScanner.filter(pluginInfo, vendureConfig)` callback so that custom orchestrator packages can apply their own filtering logic. Useful even with the default filter in place — for cases where the active set is computed in ways the default filter cannot reconstruct.
2. **Make the contract two-way**: in the runtime config loader, also emit a `dashboardEntryPaths` array derived from the active `plugins` array directly (e.g. by reading a static `dashboardEntryPath` on each plugin class). The Vite side then consumes this rather than rebuilding it via static analysis. More invasive but eliminates a whole class of import-graph drift bugs.
3. **Disambiguate by full path, not just class name**: store `(pluginPath, name)` tuples and match accordingly, in case two installed plugins share a class name across packages.

## Why this matters

This is not a niche concern. Any of these patterns hits it:

- Conditional plugin loading based on `process.env` (e.g. dev-only / test-only plugins).
- Feature-flag-driven plugin sets.
- Meta-packages / orchestrators that enable plugins based on a unified options object (the pattern `VendurePlatform.init({...})` or any equivalent in user codebases).
- Multi-tenant setups where different deployments enable different plugin combinations from a shared bundle.

In all of these, the runtime config is correct but the dashboard ships extensions for plugins that aren't active, leading to broken nav items, missing services on the API side, or extension code crashing because its plugin's resolvers / GraphQL types aren't registered.

## Repro environment

Reproducible with `@vendure/core` >= 3.6 and `@vendure/dashboard` at any current version. No specific OS / Node version requirement.

I'm happy to put together a PR for the filtering change if there's appetite — would value a steer on the preferred approach for `(2)` (pushing dashboard entry paths out from the runtime side) before opening one, since that intersects with the plugin metadata reflection story.

grolmus:
## Summary

The dashboard's Vite plugin discovery walks the import graph from `vendure-config.ts` and additionally expands `package.json` `dependencies` of every imported package. Any plugin reachable through these channels has its dashboard extension bundled and registered in the dashboard runtime — even when the plugin is **not** present in the actual `plugins` array passed to `bootstrap()`.

This breaks any pattern where plugins are conditionally added to the runtime `plugins` array (env-driven, feature-flagged, wrapper/meta-package orchestrators, multi-tenant setups with a shared bundle). The dashboard ends up loading nav items, routes and page blocks for plugins whose server-side resolvers, GraphQL types and services are absent at runtime — causing broken nav items and extension code that crashes on load.

## Solution

Add a shared `filterActivePluginInfo(pluginInfo, vendureConfig)` helper in `packages/dashboard/vite/utils/get-active-plugin-info.ts` that intersects the statically discovered `PluginInfo[]` against `VendureConfig.plugins[]`. Apply it before every consumer that emits dashboard artefacts:

| Consumer | What it emits | Now filtered? |
|---|---|---|
| `vite-plugin-dashboard-metadata.ts` | `runDashboardExtensions()` virtual module | ✅ |
| `vite-plugin-translations.ts` (load + generateBundle) | merged plugin `.po` catalogs | ✅ |
| `vite-plugin-tailwind-source.ts` | `@source` directives for Tailwind v4 scanning | ✅ |
| `vite-plugin-lingui-babel.ts` | allowed node_modules packages for Lingui macro transform | ✅ |

The helper handles both forms a `VendureConfig.plugins[]` entry can take:
- a class with `@VendurePlugin` (including the return value of `SomePlugin.init(opts)`, which by convention returns the class itself)
- a NestJS `DynamicModule` of shape `{ module: SomePluginClass, ... }`

Matching is done by class name because the runtime config and static discovery load plugin modules through different import paths and therefore see distinct class objects. The discovery step already keys on `name`, so this filter introduces no new ambiguity — the limitation is documented in the helper's doc comment together with a note about how to upgrade to `(pluginPath, name)` matching if needed.

The discovery log in `vite-plugin-config-loader.ts` is intentionally left untouched — it still reports what discovery found, which is the right thing for users debugging why their plugin is or isn't picked up.

Fixes #4706

## Test plan

- [x] **Unit tests** for `filterActivePluginInfo` in `packages/dashboard/vite/tests/plugin-hooks.spec.ts` — 5 new tests covering: class-name match, empty `plugins` array, missing `plugins` key, NestJS `DynamicModule` unwrap, ignoring entries without resolvable class name.
- [x] **Consumer behaviour tests** — new filter-specific tests added to the `dashboardMetadataPlugin` and `dashboardTailwindSourcePlugin` describe blocks, asserting that only the active plugin's extension path appears in the generated `runDashboardExtensions()` and `@source '...'` lines respectively. The existing tests in those blocks were preserved unchanged by updating the fake config loader factory to auto-build a matching `VendureConfig.plugins` array from the supplied `pluginInfo`.
- [x] **End-to-end fixture test** in `packages/dashboard/vite/tests/meta-package.spec.ts` — exercises the real `compile()` pipeline against the existing meta-package fixture (`MetaPlugin.init()` returns `[ChildPluginA, ChildPluginB]`, while transitive `package.json` dependency expansion also discovers `ChildPluginC`). Asserts that discovery still finds all 3 (no regression) but `filterActivePluginInfo` drops `ChildPluginC` — i.e. the exact bug scenario from #4706 is now caught.
- [x] `npm test` from `packages/dashboard` — **357 tests pass** (53 in plugin-hooks.spec.ts, 3 in meta-package.spec.ts).
- [x] `npm run check-types` — clean.
- [x] `npm run build:vite` — clean.
- [x] Pre-commit lint (`lint-staged` on the changed files) — clean.

## Notes for reviewers

- This is a behavioural change visible to anyone whose dashboard was inadvertently displaying extensions for non-active plugins; for the documented bug pattern (#4706), the new behaviour matches user expectations. There is no documented case where users would want the previous behaviour, and the helper docstring outlines an escape hatch if one ever materialises.
- The filter is called once per consumer plugin per build; it's an O(n) operation over a small set, so memoisation isn't worth the indirection at this stage.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4732"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


michaelbromley:
Hey @grolmus — the fix looks solid (filter is applied consistently, tests are thorough, the class-name-matching trade-off is well documented). One thing worth thinking about before/after merge: the current shape of `getActivePluginNames` works perfectly for the documented plugin patterns, but a handful of less common `plugins` array entries will be **silently dropped** from the active set, causing the dashboard extension to be absent without any warning.

Not blocking — just worth either documenting or guarding against. Cases I can think of:

### Silent false-negatives (plugin active at runtime, extension stripped from bundle)

1. **Anonymous classes**
   ```ts
   plugins: [class extends VendurePlugin { /* ... */ }]
   ```
   `.name === ''`, so the entry never makes it into the `Set`.

2. **Factory-generated plugins that don't set `.name`**
   ```ts
   function createPluginFor(kind: string) {
       return class { static kind = kind; }; // anonymous
   }
   plugins: [createPluginFor('Stripe')];
   ```
   The returned class has `.name === ''`. Factories like this need an explicit `Object.defineProperty(Plugin, 'name', { value: 'StripePlugin' })` to be recognised.

3. **Promise- or instance-wrapped entries**
   ```ts
   plugins: [Promise.resolve(StripePlugin)]   // typeof === 'object', no .module
   plugins: [new EmailPlugin()]               // instance, not a constructor
   ```
   Neither matches `typeof === 'function'` nor `{ module: ... }`, so they're skipped.

4. **Class decorators returning a wrapping class**
   If `@VendurePlugin` (or any meta-decorator layered on top) is ever changed to `return class extends target { ... }` instead of mutating the original class, the wrapper is anonymous and **every** plugin gets filtered out across every project. Modern TS decorators usually mutate in place so this is latent, but it's worth a regression test that locks the contract.

### Silent false-positives (rare)

5. **Same-class-name collisions across packages** — two packages both exporting `class BillingPlugin` are indistinguishable. The active one keeps its extension correctly, but the inactive one also slips through. Already acknowledged in the docstring; just noting it stays a known limitation.

### Environmental

6. **Build-time vs runtime env mismatch** — if the config reads `process.env` to decide which plugins to include, the dashboard build sees the build-time env, not the eventual server runtime env. Not this PR's bug, but a new failure mode the filter makes more visible (extensions and backend plugins can now diverge if envs differ).

### Suggestion (one-liner, non-blocking)

The cheapest mitigation for cases 1–4 is a single warning log next to the silent drop:

```ts
const name = pluginClass?.name;
if (name) {
    names.add(name);
} else {
    // Don't have access to Vite's logger here, but even a console.warn
    // converts \"silent broken UI\" into \"check the build log\".
    console.warn(
        '[dashboard] Could not derive a class name for a plugins[] entry; ' +
        'its dashboard extension (if any) will not be bundled. ' +
        'Ensure plugin entries are named classes or DynamicModules with a named .module.'
    );
}
```

Or even better, plumb a Vite logger through `filterActivePluginInfo` so the warning shows up in the standard build output.

Either way, none of this blocks the merge — it's a real improvement over the status quo. Mostly flagging it so the failure modes are documented somewhere durable if you don't want to add the warning right now. Cheers!

PR #4733, opened 2026-05-13
fix(core): Skip v3.6 asset translation migration on empty DB

### LINKED ISSUE #4647: Unable to run migration from v3.6.0 with empty database
**Describe the bug**
The function `migrateAssetTranslationData` seems to access the default channel. This works when you already have an existing database you want to migrate but in the case where you delete your databse (e.g. in development) to start new it fails with the following error:

```bash
[server] Successfully migrated ProductOptionGroup data to new join tables.
[server] An error occurred when running migrations:
# This is why it fails, the default channel does not exist
[server] Could not find the default channel. The __default_channel__ must exist before running this migration.
[server] Your database schema does not match your current configuration. Generate a new migration for the following changes:
[server]  - ALTER TABLE "product_option_group" DROP CONSTRAINT "FK_a6e91739227bf4d442f23c52c75"
[server]  - DROP INDEX "public"."IDX_a6e91739227bf4d442f23c52c7"
[server]  - CREATE TABLE "asset_translation" ("createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), "languageCode" character varying NOT NULL, "name" character varying NOT NULL, "id" uuid NOT NULL DEFAULT uuid_generate_v4(), "baseId" uuid, CONSTRAINT "PK_2f22e63eefeef14d245bdb956b6" PRIMARY KEY ("id"))
[server]  - CREATE INDEX "IDX_4eed4464adef51f53e1c7d8021" ON "asset_translation" ("baseId") 
[server]  - CREATE TABLE "api_key_translation" ("createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), "languageCode" character varying NOT NULL, "name" character varying NOT NULL, "id" uuid NOT NULL DEFAULT uuid_generate_v4(), "baseId" uuid, CONSTRAINT "PK_b703f4951cec9da71354120bd8a" PRIMARY KEY ("id"))
[server]  - CREATE INDEX "IDX_bf45bd67c7b3278d7e1f2f9517" ON "api_key_translation" ("baseId") 
[server]  - CREATE TABLE "api_key" ("createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), "lookupId" character varying NOT NULL, "apiKeyHash" character varying NOT NULL, "lastUsedAt" TIMESTAMP, "deletedAt" TIMESTAMP, "id" uuid NOT NULL DEFAULT uuid_generate_v4(), "ownerId" uuid NOT NULL, "userId" uuid NOT NULL, CONSTRAINT "UQ_ade30668b991772489bf875be5f" UNIQUE ("lookupId"), CONSTRAINT "UQ_3c254ac4ce1a6d4a26da30c5575" UNIQUE ("apiKeyHash"), CONSTRAINT "PK_b1bd840641b8acbaad89c3d8d11" PRIMARY KEY ("id"))
[server]  - CREATE TABLE "product_option_channels_channel" ("productOptionId" uuid NOT NULL, "channelId" uuid NOT NULL, CONSTRAINT "PK_be681a3bd2f92f17d084fa4375a" PRIMARY KEY ("productOptionId", "channelId"))
[server]  - CREATE INDEX "IDX_8dbe001861ca34ae8b687e6bae" ON "product_option_channels_channel" ("productOptionId") 
[server]  - CREATE INDEX "IDX_717e7792b8f31c319b6c7b8135" ON "product_option_channels_channel" ("channelId") 
[server]  - CREATE TABLE "product_option_group_channels_channel" ("productOptionGroupId" uuid NOT NULL, "channelId" uuid NOT NULL, CONSTRAINT "PK_ddad9696c49ebbc8032caf76fe3" PRIMARY KEY ("productOptionGroupId", "channelId"))
[server]  - CREATE INDEX "IDX_4fbe6303db2827370c0ec2d027" ON "product_option_group_channels_channel" ("productOptionGroupId") 
[server]  - CREATE INDEX "IDX_d689965b8c58ebf316fce60fab" ON "product_option_group_channels_channel" ("channelId") 
[server]  - CREATE TABLE "product_option_groups_product_option_group" ("productId" uuid NOT NULL, "productOptionGroupId" uuid NOT NULL, CONSTRAINT "PK_6a7a0291e226fbb0d4df828a483" PRIMARY KEY ("productId", "productOptionGroupId"))
[server]  - CREATE INDEX "IDX_9148fe2c2fd83f5b59d391088c" ON "product_option_groups_product_option_group" ("productId") 
[server]  - CREATE INDEX "IDX_9b03a92219b0684dbd4403e624" ON "product_option_groups_product_option_group" ("productOptionGroupId") 
[server]  - CREATE TABLE "api_key_channels_channel" ("apiKeyId" uuid NOT NULL, "channelId" uuid NOT NULL, CONSTRAINT "PK_acb0650ccd9b2df593d1b4f1c52" PRIMARY KEY ("apiKeyId", "channelId"))
[server]  - CREATE INDEX "IDX_460b1afc096014ca2dc5a5f5aa" ON "api_key_channels_channel" ("apiKeyId") 
[server]  - CREATE INDEX "IDX_d37be6b22047f56ea87bea795b" ON "api_key_channels_channel" ("channelId") 
[server]  - ALTER TABLE "product_option_group" DROP COLUMN "productId"
[server]  - ALTER TABLE "asset" DROP COLUMN "name"
[server]  - ALTER TABLE "administrator" DROP CONSTRAINT "UQ_154f5c538b1576ccc277b1ed631"
[server]  - ALTER TABLE "asset_translation" ADD CONSTRAINT "FK_4eed4464adef51f53e1c7d80212" FOREIGN KEY ("baseId") REFERENCES "asset"("id") ON DELETE CASCADE ON UPDATE NO ACTION
[server]  - ALTER TABLE "api_key_translation" ADD CONSTRAINT "FK_bf45bd67c7b3278d7e1f2f95170" FOREIGN KEY ("baseId") REFERENCES "api_key"("id") ON DELETE CASCADE ON UPDATE NO ACTION
[server]  - ALTER TABLE "api_key" ADD CONSTRAINT "FK_74d2236b1de818d00bd3fd01602" FOREIGN KEY ("ownerId") REFERENCES "user"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
[server]  - ALTER TABLE "api_key" ADD CONSTRAINT "FK_277972f4944205eb29127f9bb6c" FOREIGN KEY ("userId") REFERENCES "user"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
[server]  - ALTER TABLE "product_option_channels_channel" ADD CONSTRAINT "FK_8dbe001861ca34ae8b687e6baef" FOREIGN KEY ("productOptionId") REFERENCES "product_option"("id") ON DELETE CASCADE ON UPDATE CASCADE
[server]  - ALTER TABLE "product_option_channels_channel" ADD CONSTRAINT "FK_717e7792b8f31c319b6c7b81352" FOREIGN KEY ("channelId") REFERENCES "channel"("id") ON DELETE CASCADE ON UPDATE NO ACTION
[server]  - ALTER TABLE "product_option_group_channels_channel" ADD CONSTRAINT "FK_4fbe6303db2827370c0ec2d0276" FOREIGN KEY ("productOptionGroupId") REFERENCES "product_option_group"("id") ON DELETE CASCADE ON UPDATE CASCADE
[server]  - ALTER TABLE "product_option_group_channels_channel" ADD CONSTRAINT "FK_d689965b8c58ebf316fce60fab2" FOREIGN KEY ("channelId") REFERENCES "channel"("id") ON DELETE CASCADE ON UPDATE NO ACTION
[server]  - ALTER TABLE "product_option_groups_product_option_group" ADD CONSTRAINT "FK_9148fe2c2fd83f5b59d391088c5" FOREIGN KEY ("productId") REFERENCES "product"("id") ON DELETE CASCADE ON UPDATE CASCADE
[server]  - ALTER TABLE "product_option_groups_product_option_group" ADD CONSTRAINT "FK_9b03a92219b0684dbd4403e6246" FOREIGN KEY ("productOptionGroupId") REFERENCES "product_option_group"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
[server]  - ALTER TABLE "api_key_channels_channel" ADD CONSTRAINT "FK_460b1afc096014ca2dc5a5f5aa9" FOREIGN KEY ("apiKeyId") REFERENCES "api_key"("id") ON DELETE CASCADE ON UPDATE CASCADE
[server]  - ALTER TABLE "api_key_channels_channel" ADD CONSTRAINT "FK_d37be6b22047f56ea87bea795b6" FOREIGN KEY ("channelId") REFERENCES "channel"("id") ON DELETE CASCADE ON UPDATE CASCADE
```

**To Reproduce**

1. Use an existing project with the v36 migration including this function `migrateAssetTranslationData`
2. Clear the database and run the migrations
3. There should be a similar error

**Expected behavior**
The migrations should all complete without problems

**Actual behavior**
Our v36 migrations fails

- Nodejs version: 24.2.0

**Additional context**
It seems that sometimes it suceeds running all migrations but in most cases it doesn't.

grolmus:
## Summary

`migrateAssetTranslationData()` unconditionally queried the `__default_channel__` row to obtain a default language code. On a fresh install, migrations run *before* the default channel is seeded, so an empty database threw:

> Could not find the default channel. The `__default_channel__` must exist before running this migration.

…even though there was no asset data to migrate.

This PR counts assets with a non-null `name` first. If there are none, the function logs and returns early without touching the channel table. The error still surfaces when real asset data exists but the default channel is missing — that's a legitimately broken state that should not be silently ignored.

Note: the sibling helper `migrateProductOptionGroupData()` also references `__default_channel__` but is accidentally safe — it joins through `product_option_group`, so an empty table makes the `INSERT … SELECT` a no-op rather than a throw. No parallel fix needed.

Fixes #4647

## Test plan

- [x] New e2e test in `migrate-asset-translations.e2e-spec.ts` reproduces the fresh-install scenario: renames the default channel so any channel lookup would fail, adds an empty `name` column to `asset`, and asserts the migration neither throws nor inserts any `asset_translation` rows
- [x] All 8 tests in `migrate-asset-translations.e2e-spec.ts` pass (sqljs)
- [x] Sibling `migrate-product-option-groups.e2e-spec.ts` still green (5/5)
- [x] `tsc --noEmit` clean

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4733"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/migrate-asset-translations.e2e-spec.ts
+++ b/packages/core/e2e/migrate-asset-translations.e2e-spec.ts
@@ -103,6 +103,40 @@ describe('migrateAssetTranslationData()', () => {
         await migrateAssetTranslationData(queryRunner);
     });
 
+    // Reproduces https://github.com/vendurehq/vendure/issues/4647 (OSS-496):
+    // on a fresh install with no asset data and no default channel yet,
+    // the migration should be a no-op rather than throwing.
+    it('should skip if there are no assets to migrate, even without a default channel', async () => {
+        // Rename the default channel to simulate it not existing. If the
+        // function still queried the channel table for __default_channel__,
+        // it would throw.
+        await queryRunner.query(
+            `UPDATE ${esc('channel')} SET ${esc('code')} = '__tmp_renamed__' WHERE ${esc('code')} = '__default_channel__'`,
+        );
+        await queryRunner.query(`ALTER TABLE ${esc('asset')} ADD ${esc('name')} varchar(255) NULL`);
+        try {
+            const beforeRows: Array<{ c: string | number }> = await queryRunner.query(
+                `SELECT COUNT(*) AS ${esc('c')} FROM ${esc('asset_translation')}`,
+            );
+
+            // All asset.name values are NULL — nothing to migrate. Should not throw.
+            await expect(migrateAssetTranslationData(queryRunner)).resolves.not.toThrow();
+
+            // And no asset_translation rows should have been inserted.
+            const afterRows: Array<{ c: string | number }> = await queryRunner.query(
+                `SELECT COUNT(*) AS ${esc('c')} FROM ${esc('asset_translation')}`,
+            );
+            expect(Number(afterRows[0].c)).toBe(Number(beforeRows[0].c));
+        } finally {
+            if (await queryRunner.hasColumn('asset', 'name')) {
+                await queryRunner.query(`ALTER TABLE ${esc('asset')} DROP COLUMN ${esc('name')}`);
+            }
+            await queryRunner.query(
+                `UPDATE ${esc('channel')} SET ${esc('code')} = '__default_channel__' WHERE ${esc('code')} = '__tmp_renamed__'`,
+            );
+        }
+    });
+
     it('should populate asset_translation from name column', async () => {
         try {
             await simulatePreMigrationState();
--- a/packages/core/e2e/migrate-asset-translations.e2e-spec.ts
+++ b/packages/core/e2e/migrate-asset-translations.e2e-spec.ts
@@ -107,20 +107,21 @@ describe('migrateAssetTranslationData()', () => {
     // on a fresh install with no asset data and no default channel yet,
     // the migration should be a no-op rather than throwing.
     it('should skip if there are no assets to migrate, even without a default channel', async () => {
-        // Rename the default channel to simulate it not existing. If the
-        // function still queried the channel table for __default_channel__,
-        // it would throw.
-        await queryRunner.query(
-            `UPDATE ${esc('channel')} SET ${esc('code')} = '__tmp_renamed__' WHERE ${esc('code')} = '__default_channel__'`,
-        );
-        await queryRunner.query(`ALTER TABLE ${esc('asset')} ADD ${esc('name')} varchar(255) NULL`);
         try {
+            // Rename the default channel to simulate it not existing. If the
+            // function still queried the channel table for __default_channel__,
+            // it would throw.
+            await queryRunner.query(
+                `UPDATE ${esc('channel')} SET ${esc('code')} = '__tmp_renamed__' WHERE ${esc('code')} = '__default_channel__'`,
+            );
+            await queryRunner.query(`ALTER TABLE ${esc('asset')} ADD ${esc('name')} varchar(255) NULL`);
+
             const beforeRows: Array<{ c: string | number }> = await queryRunner.query(
                 `SELECT COUNT(*) AS ${esc('c')} FROM ${esc('asset_translation')}`,
             );
 
             // All asset.name values are NULL — nothing to migrate. Should not throw.
-            await expect(migrateAssetTranslationData(queryRunner)).resolves.not.toThrow();
+            await migrateAssetTranslationData(queryRunner);
 
             // And no asset_translation rows should have been inserted.
             const afterRows: Array<{ c: string | number }> = await queryRunner.query(
--- a/packages/core/src/migration-utils/v3_6_asset_translations.ts
+++ b/packages/core/src/migration-utils/v3_6_asset_translations.ts
@@ -49,7 +49,19 @@ export async function migrateAssetTranslationData(queryRunner: QueryRunner): Pro
 
     const esc = (name: string) => queryRunner.connection.driver.escape(name);
 
-    // 1. Get the default language code from the default channel
+    // 1. If there is no asset data to migrate, skip entirely. This covers the
+    // fresh-install case where migrations run against an empty database before
+    // the default channel has been populated.
+    const assetCounts: Array<{ count: string | number }> = await queryRunner.query(
+        `SELECT COUNT(*) AS ${esc('count')} FROM ${esc('asset')} WHERE ${esc('name')} IS NOT NULL`,
+    );
+    const assetCount = Number(assetCounts?.[0]?.count ?? 0);
+    if (assetCount === 0) {
+        console.log('No asset rows with a name to migrate. Skipping asset translation data migration.');
+        return;
+    }
+
+    // 2. Get the default language code from the default channel
     const rows: Array<{ defaultLanguageCode: string }> = await queryRunner.query(
         `SELECT ${esc('defaultLanguageCode')} FROM ${esc('channel')} WHERE ${esc('code')} = '__default_channel__'`,
     );
@@ -60,7 +72,7 @@ export async function migrateAssetTranslationData(queryRunner: QueryRunner): Pro
     }
     const defaultLanguageCode = rows[0].defaultLanguageCode;
 
-    // 2. Copy asset names into the asset_translation table
+    // 3. Copy asset names into the asset_translation table
     await queryRunner.query(
         `INSERT INTO ${esc('asset_translation')} (${esc('createdAt')}, ${esc('updatedAt')}, ${esc('languageCode')}, ${esc('name')}, ${esc('baseId')})
          SELECT a.${esc('createdAt')}, a.${esc('updatedAt')}, '${defaultLanguageCode}', a.${esc('name')}, a.${esc('id')}
--- a/packages/core/src/migration-utils/v3_6_asset_translations.ts
+++ b/packages/core/src/migration-utils/v3_6_asset_translations.ts
@@ -55,7 +55,7 @@ export async function migrateAssetTranslationData(queryRunner: QueryRunner): Pro
     const assetCounts: Array<{ count: string | number }> = await queryRunner.query(
         `SELECT COUNT(*) AS ${esc('count')} FROM ${esc('asset')} WHERE ${esc('name')} IS NOT NULL`,
     );
-    const assetCount = Number(assetCounts?.[0]?.count ?? 0);
+    const assetCount = Number(assetCounts[0].count);
     if (assetCount === 0) {
         console.log('No asset rows with a name to migrate. Skipping asset translation data migration.');
         return;


packages/core/e2e/migrate-asset-translations.e2e-spec.ts
```diff
@@ -103,6 +103,40 @@ describe('migrateAssetTranslationData()', () => {
...
     });
 
+    // Reproduces https://github.com/vendurehq/vendure/issues/4647 (OSS-496):
+    // on a fresh install with no asset data and no default channel yet,
+    // the migration should be a no-op rather than throwing.
+    it('should skip if there are no assets to migrate, even without a default channel', async () => {
+        // Rename the default channel to simulate it not existing. If the
+        // function still queried the channel table for __default_channel__,
+        // it would throw.
+        await queryRunner.query(
```
michaelbromley: Both setup mutations are outside the `try` block. If either of these throws, the `finally` cleanup never runs — the channel stays renamed to `__tmp_renamed__` and any subsequent tests that rely on `__default_channel__` existing will fail in confusing ways.

Move both queries to be the first statements inside the `try` block:

```ts
try {
    await queryRunner.query(
        `UPDATE ${esc("channel")} SET ${esc("code")} = "__tmp_renamed__" WHERE ${esc("code")} = "__default_channel__"`,
    );
    await queryRunner.query(`ALTER TABLE ${esc("asset")} ADD ${esc("name")} varchar(255) NULL`);

    const beforeRows = ...
    // rest of assertions
} finally {
    // cleanup as-is
}
```

This is consistent with how all other tests in this file wrap their setup (either directly or via `simulatePreMigrationState()`).

packages/core/src/migration-utils/v3_6_asset_translations.ts
```diff
@@ -49,7 +49,19 @@ export async function migrateAssetTranslationData(queryRunner: QueryRunner): Pro
 
     const esc = (name: string) => queryRunner.connection.driver.escape(name);
 
-    // 1. Get the default language code from the default channel
+    // 1. If there is no asset data to migrate, skip entirely. This covers the
+    // fresh-install case where migrations run against an empty database before
+    // the default channel has been populated.
+    const assetCounts: Array<{ count: string | number }> = await queryRunner.query(
+        `SELECT COUNT(*) AS ${esc('count')} FROM ${esc('asset')} WHERE ${esc('name')} IS NOT NULL`,
+    );
```
michaelbromley, line 57: Minor nit: `COUNT(*)` always returns exactly one row, so the optional chaining and `?? 0` fallback are not needed here. `Number(assetCounts[0].count)` is sufficient.

packages/core/e2e/migrate-asset-translations.e2e-spec.ts
```diff
@@ -103,6 +103,40 @@ describe('migrateAssetTranslationData()', () => {
...
+        try {
+            const beforeRows: Array<{ c: string | number }> = await queryRunner.query(
+                `SELECT COUNT(*) AS ${esc('c')} FROM ${esc('asset_translation')}`,
+            );
+
+            // All asset.name values are NULL — nothing to migrate. Should not throw.
+            await expect(migrateAssetTranslationData(queryRunner)).resolves.not.toThrow();
+
+            // And no asset_translation rows should have been inserted.
+            const afterRows: Array<{ c: string | number }> = await queryRunner.query(
```
michaelbromley, line 127: Non-blocking style note: `resolves.not.toThrow()` is a slightly unusual pattern here since the function returns `void`. A cleaner option is either `await migrateAssetTranslationData(queryRunner)` on its own (the test fails if it throws) or `await expect(migrateAssetTranslationData(queryRunner)).resolves.toBeUndefined()`.

michaelbromley (changes requested):
Thanks for this fix, @grolmus — this is a real pain point for anyone doing fresh-install development against a v3.6 schema, and the approach is exactly right. Counting assets before touching the channel table is clean and minimal, and the explanation of why the sibling `migrateProductOptionGroupData()` is accidentally safe is a nice touch.

Two things to address before we merge — one required, one just a suggestion.

**Required: setup queries need to be inside the `try` block**

Both setup mutations in the new test run before the `try`, so if either throws (e.g. the `name` column already exists from a previous dirty run), the `finally` cleanup never runs. That leaves the channel renamed to `__tmp_renamed__` and potentially a stale column on `asset`, which breaks every subsequent test in the suite since they all assume `__default_channel__` exists. CodeRabbit independently flagged this same concern, which is good corroboration.

The fix is straightforward — just move both queries to be the first two lines inside the `try` block, consistent with how every other test in this file wraps its setup.

**Non-blocking suggestions** (see inline comments)

packages/core/e2e/migrate-asset-translations.e2e-spec.ts
grolmus: Fixed in f3f5740. Both setup mutations are now the first statements inside the `try` block, so the `finally` cleanup always runs even if the channel UPDATE or ALTER TABLE throws.

packages/core/src/migration-utils/v3_6_asset_translations.ts
grolmus: Fixed in f3f5740. Simplified to `Number(assetCounts[0].count)`.

packages/core/e2e/migrate-asset-translations.e2e-spec.ts
grolmus: Fixed in f3f5740. Replaced with a plain `await migrateAssetTranslationData(queryRunner)` — the test fails naturally if it throws.

michaelbromley (approved):

PR #4738, opened 2026-05-15
fix(core): Load ProductVariant translations in OrderService.findOne

grolmus:
## Summary

`OrderService.findOne()` translates every `line.productVariant` via `translator.translate()` but did not guarantee `lines.productVariant.translations` was loaded when callers passed an explicit `relations` list. The auto-injection block only added `lines.productVariant.taxCategory`, leaving the variant without translations and causing `error.entity-has-no-translation-in-language` on common flows (notably `addItemToOrder` -> `getOrderOrThrow`).

## Root cause

In v3.3.6 line-loading was split into a separate query using `setFindOptions({ relations: lineRelations })`. TypeORM does not honour transitive `eager: true` relations (like `ProductVariant.translations`) when an explicit `relations` array is passed to `setFindOptions`, so the translations relation that was previously picked up via eager loading is no longer present. Meanwhile `findOne` unconditionally translates every loaded variant.

## Changes

- `packages/core/src/service/services/order.service.ts`
  - Extracted a `wantsLineVariant` predicate that matches both `lines.productVariant` and any `lines.productVariant.*` sub-relation, and used it for both the auto-inject block and the downstream translator block (previously two coupled literal-string checks).
  - When the caller passes `relations` containing the variant, auto-inject both `lines.productVariant.taxCategory` and `lines.productVariant.translations` when missing.
- `packages/core/e2e/order.e2e-spec.ts`
  - Added a regression test referencing #4566 inside the `issues` describe block. It calls `OrderService.findOne` directly with the exact buggy relations list from the report and asserts translations are loaded and `variant.name` is populated. A follow-up `addItemToOrder` step exercises the original user-visible failure path (`addItemToOrder` -> `getOrderOrThrow` -> `findOne` -> translate).

`getOrderOrThrow` defaults were intentionally not changed: they pass an explicit `relations` array to `findOne`, which then triggers the auto-inject — keeping a single source of truth.

## Follow-up

A separate `translator.translate(line.productVariant, ctx)` call exists in `OrderService.assertInsufficientStockOnHand()` (`order.service.ts:1843`). The variant there is loaded with `{ relations: ['productVariant'] }` only — the same root cause class. It is currently masked because that call uses a plain `repository.find()` rather than the split-load path, but worth tracking as a follow-up.

## Test plan

- [x] New test passes locally: `bun e2e -t "loads ProductVariant translations"` (7.5s)
- [x] No regressions in the other 92 order e2e tests
- [x] TypeScript build is clean

Fixes #4566

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4738"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/order.e2e-spec.ts
+++ b/packages/core/e2e/order.e2e-spec.ts
@@ -12,6 +12,8 @@ import {
 import { omit } from '@vendure/common/lib/omit';
 import { pick } from '@vendure/common/lib/pick';
 import {
+    OrderService,
+    RequestContextService,
     defaultShippingCalculator,
     defaultShippingEligibilityChecker,
     manualFulfillmentHandler,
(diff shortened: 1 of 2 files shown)


michaelbromley (approved):

michaelbromley:
Thanks for the fix Martin!

Issue #4725, opened 2026-05-12
Bug: Bulk actions crash with DialogRootContext error in monorepo dev server


Ryrahul:
**Describe the bug**
Clicking any bulk action on list pages (Products, Collections, Facets, etc.) throws a `DialogRootContext` error. The `DialogTitle` component from `@vendure-io/ui` cannot find its parent `Dialog.Root` context.

This only occurs in the monorepo dev server — a fresh `@vendure/create` 3.6.3 project works fine.

**To Reproduce**
1. Clone the vendure monorepo (`master` branch)
2. `bun install && bun run build`
3. Build dashboard: `rm -rf packages/dev-server/dist && bunx vite build --base /dashboard/ --outDir ../dev-server/dist` from `packages/dashboard`
4. Populate and start dev server from `packages/dev-server`
5. Go to Products list, select one or more products
6. Click "Actions" dropdown
7. Click any bulk action (Delete, Edit facet values, Assign to channel, etc.)
8. See error

**Expected behavior**
The bulk action should open its respective dialog/confirmation without errors.

**Actual behavior**
An error boundary catches a crash from `DialogTitle` inside `@vendure-io/ui`, and the page shows an error message.

**Screenshots/Videos**

<img width="1470" height="881" alt="Image" src="https://github.com/user-attachments/assets/838832be-a64f-48ef-a735-c7c49ef5b0a3" />

**Error logs**
```
react-dom_client.js?v=5837f0ea:7001 TypeError: Cannot destructure property 'store' of 'useDialogRootContext(...)' as it is undefined.
    at DialogTitle2 (@vendure-io_ui_lib_base-ui.js?v=5837f0ea:9722:5)
    at Object.react_stack_bottom_frame (react-dom_client.js?v=5837f0ea:18509:20)
    at renderWithHooks (react-dom_client.js?v=5837f0ea:5654:24)
    at updateForwardRef (react-dom_client.js?v=5837f0ea:7198:21)
    at beginWork (react-dom_client.js?v=5837f0ea:8735:20)
    at runWithFiberInDEV (react-dom_client.js?v=5837f0ea:997:72)
    at performUnitOfWork (react-dom_client.js?v=5837f0ea:12561:98)
    at workLoopSync (react-dom_client.js?v=5837f0ea:12424:43)
    at renderRootSync (react-dom_client.js?v=5837f0ea:12408:13)
    at performWorkOnRoot (react-dom_client.js?v=5837f0ea:11827:37)

[...]

PR #4739, opened 2026-05-15
fix(dashboard): Preserve column visibility and fix Dialog title context

grolmus:
## Summary

Two related dashboard fixes uncovered while triaging the column-visibility regression on the Scheduled Tasks page.

### 1. Column visibility reset after a mutation refetch

`DataTable`'s effect that synced `defaultColumnVisibility` into state compared the prop against the user's *current* state and reset whenever they differed. With inline literals like `<DataTable defaultColumnVisibility={{ schedule: false }} />` (used on Scheduled Tasks), every parent re-render produced a fresh reference, so any mutation that invalidated the query — including `disable` / `enable` — silently reverted the user's column visibility.

The effect now tracks the last-applied default content via a ref and only resets when that content actually changes (e.g. the user clicks "Reset" in the column settings menu). User-modified state is preserved across incidental re-renders.

Fix is in `DataTable` rather than at the caller because `paginated-list-data-table.tsx` and `order-table.tsx` have the same shape (`defaultColumnVisibility={computeEveryRender(...)}`) and would be vulnerable to the same regression on any future mutation cycle.

### 2. `DialogTitle` / `AlertDialogTitle` primitive context error

While verifying the column fix on the dev server I hit:

```
TypeError: Cannot destructure property 'store' of 'useDialogRootContext(...)' as it is undefined.
The above error occurred in the <DialogTitle> component.
```

…on every dialog with a title (Scheduled Tasks > View result, and any other consumer of the dashboard's `DialogTitle` / `AlertDialogTitle`).

Root cause: the dashboard's overrides rendered the `@base-ui/react` primitive (`DialogPrimitive.Title` / `AlertDialogPrimitive.Title`) directly while the surrounding `<Dialog>` / `<AlertDialog>` came from the `@vendure-io/ui` wrappers. Vite can resolve those two paths (`@vendure-io/ui/lib/base-ui` vs `@vendure-io/ui/components/ui/dialog` → `@base-ui/react/dialog`) to separate module instances, so the title rendered with no Dialog root context.

Fix: wrap `@vendure-io/ui`'s `DialogTitle` / `AlertDialogTitle` rather than the underlying primitive, and apply `font-heading` via `className`. The merge keeps the original utility classes and there is now a single primitive instance for the whole Dialog subtree.

## Test plan

- [x] Manually verified on dev server (which has real scheduled tasks): hidden Schedule Pattern via column settings shows up, disable a task → column stays visible. Re-enable cleans up.
- [x] Manually verified View result dialog opens without console errors and renders payload.
- [x] Added regression smoke test in `e2e/tests/system/scheduled-tasks.spec.ts` (refresh button instead of disable, since the e2e config registers no scheduled tasks — comment in the test explains).
- [x] All 4 scheduled-tasks playwright tests pass locally.
- [x] TypeScript build clean.

## Notes

- The e2e regression test does not fail without the column-visibility fix in this test environment: with no scheduled tasks in the e2e Vendure config, refetches return the same empty data and TanStack Query's tracked properties skip the parent re-render that triggers the bug. The test still guards against the broader "resets-on-refresh" class of regressions; the underlying fix was verified manually on a dev server with real tasks.
- The `data-testid="dt-refresh-button"` added to the `RefreshButton` is reused for the new test and is generally useful for future e2e coverage.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4739"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

## Release Notes

* **Bug Fixes**
  * Improved column visibility persistence in data tables to prevent resetting when page refreshes
  
* **Tests**
  * Added end-to-end test verifying column visibility settings persist after page refresh

* **Style**
  * Updated dialog and alert dialog components for improved UI consistency

[Review Change Stack](https://app.coderabbit.ai/change-stack/vendurehq/vendure/pull/4739)

### DIFF
--- a/packages/dashboard/e2e/tests/system/scheduled-tasks.spec.ts
+++ b/packages/dashboard/e2e/tests/system/scheduled-tasks.spec.ts
@@ -14,4 +14,37 @@ test.describe('Scheduled Tasks', () => {
         await expect(page.getByTestId('page-heading')).toBeVisible();
         await expect(page.getByRole('table')).toBeVisible();
     });
+
+    // OSS-470 — column visibility must survive a refresh.
+    //
+    // Original bug: the parent's inline `defaultColumnVisibility` got a new
+    // reference each render, and the DataTable effect compared that against
+    // the user-modified state and reset it. The user-visible trigger was the
+    // disable/enable mutation's onSuccess → queryClient.invalidateQueries →
+    // parent re-render. This e2e Vendure config registers no scheduled tasks,
+    // so refetches return the same empty data and TanStack Query's tracked
+    // properties skip the parent re-render — meaning this test does not fail
+    // against the unfixed code. It is kept as a smoke test for the wider class
+    // of "resets-on-refresh" regressions; the actual fix was verified manually
+    // on a dev server with real scheduled tasks.
+    test('should preserve column visibility after a refresh', async ({ page }) => {
+        await page.goto('/scheduled-tasks');
+        await expect(page.getByRole('table')).toBeVisible();
+
+        const dataTable = page.getByRole('table');
+        const descriptionHeader = dataTable.locator('thead th').filter({ hasText: 'Description' });
+
+        // "Description" column is visible by default — hide it via column settings.
+        await expect(descriptionHeader).toBeVisible();
+        await page.getByTestId('dt-column-settings-trigger').click();
+        await page.getByRole('menuitemcheckbox', { name: 'Description', exact: true }).click();
+        await page.keyboard.press('Escape');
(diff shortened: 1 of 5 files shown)


Ryrahul:
I think this PR also resolves issue #4725

michaelbromley (approved):

PR #4740, opened 2026-05-15
fix(core): prevent double-decoding of ID args in updatePromotion

### LINKED ISSUE #4700: Can't save promotion with products_percentage_discount and hashed ids
**Describe the bug**
You can't update promotion with products_percentage_discount as an action, when you have hashed ids. I received an error "GraphQL Request Error: Invalid ID: 29".

This is my EntityIdStrategy:

```typescript

import {EntityIdStrategy} from "@vendure/core";
import Hashids from 'hashids'

const salt = "salt"
const hashids = new Hashids(salt, 10)

export class HashAutoIncrementIdStrategy implements EntityIdStrategy<'increment'> {
    readonly primaryKeyType = "increment";
    decodeId(id: string): number {
        const decoded = hashids.decode(id);
        if (!decoded || decoded.length === 0) {
          throw new Error(`Invalid ID: ${id}`);
        }
        return Number(decoded[0]);
    }
    encodeId(primaryKey: number): string {
        return hashids.encode(primaryKey)
    };
}
```

**To Reproduce**

1. Add entity id strategy
2. Create promotion with products_percentage_discount as an action
3. Change something, for example try to enable/disable it
4. Click update 
5. See error GraphQL Request Error: Invalid ID: {$id}

**Expected behavior**
Promotion updates succesfully

**Actual behavior**
Promotion does not update

**Screenshots/Videos**
Image

**Error logs**

```
[server] error 5/6/26, 10:51 AM - [Vendure Server] Invalid ID: 142
[server] Error: Invalid ID: 142                                                                                                                                                      
[server]     at HashAutoIncrementIdStrategy.decodeId (/src/strategies/entity-id.strategy.ts:12:17)
[server]     at /node_modules/@vendure/core/src/api/common/id-codec.ts:30:44
[server]     at IdCodec.transformRecursive (/node_modules/@vendure/core/src/api/common/id-codec.ts:65:20)
[server]     at IdCodec.decode (/node_modules/@vendure/core/src/api/common/id-codec.ts:28:21)
[server]     at IdCodecService.decode (/node_modules/@vendure/core/src/api/common/id-codec.service.ts:21:29)
[server]     at /node_modules/@vendure/core/src/api/common/configurable-operation-codec.ts:43:78
[server]     at Array.map (<anonymous>)
[server]     at ConfigurableOperationCodec.decodeConfigurableOperationIds (/node_modules/@vendure/core/src/api/common/configurable-operation-codec.ts:43:48)
[server]     at PromotionResolver.updatePromotion (/node_modules/@vendure/core/src/api/resolvers/admin/promotion.resolver.ts:101:41)
[server]     at /node_modules/@nestjs/core/helpers/external-context-creator.js:67:33 
```

- @vendure/core version: 3.6.3 & 3.6.2
- Nodejs version: v24.14.0
- Operating System (Windows/macOS/Linux): macOS & Linux

**Configuration**

```typescript

 entityOptions: {
        entityIdStrategy: new HashAutoIncrementIdStrategy(),
    },
```

**Minimal reproduction**

**Workaround**

**Additional context**

Ryrahul:
Fixes #4700

# Description

`updatePromotion` was calling `decodeConfigurableOperationIds` twice on the same `args.input.actions` array — once with `PromotionOrderAction` and once with `PromotionItemAction`. Since both types resolve to the same action defs pool in `getAvailableDefsOfType()`, the second call attempted to decode already-decoded values, causing `Invalid ID` errors with custom `EntityIdStrategy` implementations (e.g. Hashids).

The fix removes the duplicate decode call. A single call with either type is sufficient since `PromotionItemAction` and `PromotionOrderAction` share the same `case` in the switch statement and return the same list of action definitions.

Also updated `createPromotion` to use `PromotionItemAction` instead of `PromotionOrderAction` for consistency (functionally equivalent).

# Breaking changes

No.

# Screenshots

N/A

# Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4740"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


michaelbromley (changes requested):
Hey, thanks for this fix. The fix looks good, but I find the new spec file a bit weak. It is essentially defining a mock strategy and then testing how that mock strategy behaves.

If we are going to include tests, I would suggest something that actually exercises the promotion resolver. Maybe there’s an opportunity to add a spy to the existing e2e test suite to assert only a single call of the entityIdStrategy?

Ryrahul:
> Hey, thanks for this fix. The fix looks good, but I find the new spec file a bit weak. It is essentially defining a mock strategy and then testing how that mock strategy behaves.
> 
> If we are going to include tests, I would suggest something that actually exercises the promotion resolver. Maybe there’s an opportunity to add a spy to the existing e2e test suite to assert only a single call of the entityIdStrategy?

Thanks for the feedback , that makes sense. You're right that the current spec mostly validates the mock behavior rather than the actual resolver flow. I’ll look into adding a spy around the `entityIdStrategy` in the existing e2e promotion tests and assert it’s only called once instead.

Ryrahul:
Hey, @michaelbromley

Updated the test to be an e2e test that exercises the actual promotion resolver instead of the isolated unit test.

The test:

1. Calls `updatePromotion` with actions containing `facetValueIds: '["T_1"]'` (ID-type list arg)
2. Asserts the round-trip — the returned action args value equals `["T_1"]`, confirming IDs were correctly decoded and re-encoded
3. Adds a `vi.spyOn` on `entityIdStrategy.decodeId` to assert it was never called with a raw numeric string (which would indicate double-decoding)

I went with looping over `decodeIdSpy.mock.calls` rather than a simple `toHaveBeenCalledTimes(n)` check because `decodeId` is a shared function that gets called for every ID in the request — the promotion’s own ID, translation IDs, channel IDs, etc. — not just the `facetValueIds` arg we’re testing.

For example, a single `updatePromotion` call might trigger `decodeId` 5 times today, but if someone later adds another relation to the `Promotion` entity, it becomes 6 and the test breaks for no meaningful reason.

I haven't used spyOn much before so had claude write that part.

michaelbromley (approved):

michaelbromley:
Great, thanks for sorting out the test!

PR #4763, opened 2026-05-21
feat(core): Admin-driven customer verification and password reset

tbouliere-datasolution:
# Description

  This PR adds two administrator-driven actions on the Customer detail to handle the cases where a customer can no longer self-serve their account state — either because they never confirmed their verification email or because they've lost access to their inbox.

  Two new admin-API mutations are exposed on the Customer namespace:

  - verifyCustomerAccountAsAdmin(customerId: ID!) — completes the email-verification step for a registered but unverified Customer with a randomly generated 16-character password, bypassing the verification-token flow. The cleartext password is returned once in the response
  (via the new AdminGeneratedPassword type) so the administrator can communicate it to the customer; it is not persisted anywhere else. A CUSTOMER_VERIFIED history entry is written and an AccountVerifiedEvent is published.
  - resetCustomerPasswordAsAdmin(customerId: ID!) — administrator-driven counterpart to the storefront requestPasswordReset mutation. For a verified Customer, it issues a password-reset token and publishes a PasswordResetEvent so the configured email handler can send the
  customer a reset link. No password is generated or returned — the customer chooses their new password through the standard reset flow.

  A new CustomerAccountStateError (with an accountState: String field — "guest", "registered", or "verified") surfaces the cases where the action does not apply to the current account state. Under the hood, a new UserService.resetPasswordAsAdmin(ctx, userId, newPassword)
  primitive performs the direct password reset without a token (also marking the user as verified if it wasn't already), and a getUserByIdWithPasswordHash helper consolidates the passwordHash-selecting query previously inlined in setPassword.

  The Vendure Dashboard customer detail page gains a corresponding action button to invoke either mutation depending on the customer's current state.

# Breaking changes

  No breaking changes. This PR is purely additive:
  - New GraphQL types (AdminGeneratedPassword, CustomerAccountStateError) and unions (VerifyCustomerAccountAsAdminResult, ResetCustomerPasswordAsAdminResult).
  - New mutations behind the existing Permission.UpdateCustomer permission.
  - New public method on UserService (resetPasswordAsAdmin).
  - The internal refactor in UserService.setPassword (extracting getUserByIdWithPasswordHash) preserves the existing behavior.

# Screenshot

Verify

----

Verify_OK

----

Reset

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4763"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781948658&installation_id=128408429&pr_number=4763&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4763&signature=927c92ee9769b8f5fec919cee2e20c5f3bc7724c01f950d4ef0b853605de16ed"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


latifniz:
I identified this same issue and implemented a fix locally. Happy to share my approach or help review if useful.
but my focus was just on verification of the customer from the admin side, a simple mutation.

Ryrahul:
I think this might already be covered by the existing admin customer creation flow in resolver/service layer. (password is missing in the dashboard ui)

[`CustomerService.create()](https://github.com/vendurehq/vendure/blob/master/packages/core/src/service/services/customer.service.ts#L210) is documented as:

> "This method is intended to be used in admin-created Customer flows."

and it already immediately verifies the customer when a password is provided by calling the existing verification flow internally. 

So from my understanding, admins can already create a verified customer account through the existing API/service behavior, previous angular admin ui already had this behaviour so maybe adding password in the dashboard ui may be the appropriate way to go ahead **( i am not a maintainer so , you should probably wait for maintainer take on it )**,

michaelbromley (commented):
Hey @tbouliere-datasolution thanks for the PR and thanks for the additional input @latifniz and @Ryrahul 

My thoughts:

The generated password should not be returned in the GraphQL response. That cleartext ends up in the browser devtools, in whatever proxy or APM sits in front of the API, and then in an admin's clipboard on its way to the customer over some channel we know nothing about. 

It also leaves a member of staff holding a customer's password. Verification, password reset and email change in Vendure all go through a token that only ever travels to the registered email address and expires by itself, and this should work the same way. The flow also notifies the customer of nothing, so they end up verified with a password they don't know until someone messages it to them.

What I'd like to see:

- **Split the PR.** `resetCustomerPasswordAsAdmin` is useful and close to mergeable — it sends a reset link and returns nothing sensitive. Two fixes needed: the fabricated shop `RequestContext` drops out of the resolver's transaction (the transaction manager is a symbol on the ctx object, not on `req`) and loses the admin as the actor on the history entry. `ctx.copy(ctx.channel)` gives you the channel's default language for the email while keeping both. It also needs a `CUSTOMER_ACCOUNT_STATE_ERROR` entry in `packages/core/src/i18n/messages/en.json`, otherwise the toast shows the raw error code. Note that this would be a new feature and should be made to the `minor` branch.

- **Replace the verify half with a mutation that wraps the existing `refreshVerificationToken()`.** It re-issues the token and republishes `AccountRegistrationEvent`, so the customer gets the verification email again. Covers the scenario from your description with no secret in flight, and it's a much smaller change.

- Regenerate the codegen artifacts against the latest version of the target branch. `generated-shop-types.ts` is churning ~2700 lines on an admin-only change and every locale file has picked up unrelated string edits, which makes the diff hard to read and conflicts inevitable.

The e2e coverage is good. Guest, registered and verified branches for both mutations with event assertions is more than most PRs arrive with. The `getUserByIdWithPasswordHash` extraction is a tidy cleanup that stands on its own too.

Ryrahul:
@michaelbromley  Should admin also be allowed to add password from ui for customer in a separate PR or can be part of this to have parity with the angular admin panel or it was dropped in react dashboard for some reason?

PR #4765, opened 2026-05-21
feat(dashboard): Add `additionalStylesheets` option to vendureDashboa…

tbouliere-datasolution:
# Description

  Adds an `additionalStylesheets` option to `vendureDashboardPlugin` so that projects embedding the dashboard can inject their own CSS file(s) into the dashboard's main stylesheet declaratively, instead of having to hand-roll a custom Vite `transform` plugin that splices an `@import` into `app/styles.css`.

  Each provided path is injected as an `@import` statement right after `@import 'tailwindcss';`. This is the conventional spot for user CSS in a Tailwind v4 setup — it keeps the CSS valid (imports must precede rules) and lets the file participate in Tailwind's build pipeline, so projects can use `@source`, `@theme`, `@apply`, `@utility`, custom variants, etc. inside it.

  The option accepts either a single string or an array. Paths may be absolute or relative to `process.cwd()`; backslashes are normalized to forward slashes so the generated `@import` is valid on Windows. Duplicate imports are deduped, so the transform is idempotent.

```
  plugins: [
    vendureDashboardPlugin({
      // ...
      additionalStylesheets: [resolve(__dirname, 'src/dashboard.css')],
    }),
  ],
```

# Breaking changes

  None. additionalStylesheets is fully optional and defaults to undefined,
  so projects that don't set it behave exactly as before.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4765"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781951122&installation_id=128408429&pr_number=4765&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4765&signature=f59eaf939b8a6700a8934a4bd5ce18a1ae0eefeefb57edd8ef1afd6df6a7a5ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/app/styles.css
+++ b/packages/dashboard/src/app/styles.css
@@ -29,6 +29,14 @@ html[dir="rtl"] * {
  */
 @import 'virtual:admin-theme-inline';
 
+/*
+ * Replaced at build time by the themeVariablesPlugin with the @import
+ * statements for any stylesheets passed via the `theme.additionalStylesheets`
+ * option. Placed here so user styles participate in Tailwind's build pipeline.
+ * When no stylesheets are configured, this placeholder is removed.
+ */
+@import 'virtual:vendure-user-styles';
+
 @layer base {
     * {
         @apply border-border outline-ring/50;
--- a/packages/dashboard/src/app/styles.css
+++ b/packages/dashboard/src/app/styles.css
@@ -35,6 +35,7 @@ html[dir="rtl"] * {
  * option. Placed here so user styles participate in Tailwind's build pipeline.
  * When no stylesheets are configured, this placeholder is removed.
  */
+/* stylelint-disable-next-line no-invalid-position-at-import-rule */
 @import 'virtual:vendure-user-styles';
 
 @layer base {
--- a/packages/dashboard/vite/tests/plugin-hooks.spec.ts
+++ b/packages/dashboard/vite/tests/plugin-hooks.spec.ts
@@ -173,6 +173,73 @@ describe('themeVariablesPlugin', () => {
         const result = callTransform(plugin, css, '/app/styles.css');
         expect(result).toContain('@theme inline');
     });
+
+    it('replaces virtual:vendure-user-styles with additionalStylesheets (single path)', () => {
+        const plugin = themeVariablesPlugin({
+            additionalStylesheets: '/abs/project/src/dashboard.css',
+        });
+        const css = `@import 'tailwindcss';\n@import 'virtual:vendure-user-styles';\n@import 'tw-animate-css';`;
+        const result = callTransform(plugin, css, '/app/styles.css');
+        expect(result).toContain(
+            `@import 'tailwindcss';\n@import '/abs/project/src/dashboard.css';\n@import 'tw-animate-css';`,
+        );
+        expect(result).not.toContain('virtual:vendure-user-styles');
+    });
+
+    it('replaces virtual:vendure-user-styles with multiple stylesheets in order', () => {
+        const plugin = themeVariablesPlugin({
+            additionalStylesheets: ['/abs/project/a.css', '/abs/project/b.css'],
+        });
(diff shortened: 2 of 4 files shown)


michaelbromley (changes requested):
Thanks for this — it's a well-scoped PR with thorough tests and genuinely good docs on the option. The use case is real and worth supporting. I've got two structural requests before we merge, both about *how* it's wired rather than *what* it does.

**1. Nest the option under `theme` rather than adding a new top-level option**

I'd prefer not to grow the top-level surface of `vendureDashboardPlugin`. Custom stylesheets are an appearance concern, same as the existing `theme` token overrides, so they belong in the same namespace:

```ts
vendureDashboardPlugin({
    theme: {
        light: { /* ... */ },
        additionalStylesheets: [resolve(__dirname, 'src/dashboard.css')],
    },
})
```

Since `ThemeVariables` literally means colour-token maps, I'd keep that type pure and add a thin wrapper for the option so the naming stays honest:

```ts
export interface ThemeVariables {
    light?: ThemeColors;
    dark?: ThemeColors;
}

export interface DashboardThemeOptions extends ThemeVariables {
    additionalStylesheets?: string | string[];
}
```

Then `theme?: DashboardThemeOptions` on the dashboard options, and the wiring reads from the nested field:

```ts
plugin: () =>
    themeVariablesPlugin({
        theme: options.theme,
        additionalStylesheets: options.theme?.additionalStylesheets,
    }),
```

`themeVariablesPlugin`'s own options can stay flat internally — this is purely about the public surface, so it shouldn't disturb your transform logic. Worth a small JSDoc update on the `theme` option so it no longer reads as colours-only.

**2. Inject via a `virtual:` placeholder seam instead of regex-matching the `tailwindcss` import**

The current approach anchors to the `@import 'tailwindcss';` line in our own `styles.css` and splices after it. That line exists for Tailwind's sake, not as a deliberate injection point — so it's an incidental anchor. If it's ever reformatted, moved, or absent (the `transform` also runs on `extension-tailwind.css`, which has no tailwindcss import), injection silently no-ops — the PR's own `skips injection when @import tailwindcss is absent` test exercises exactly that path.

We already have an established pattern for this in the same plugin — `styles.css` contains `@import 'virtual:admin-theme';` / `@import 'virtual:admin-theme-inline';`, and `themeVariablesPlugin` swaps those placeholders out in `transform` (see `vite-plugin-theme.ts`).

I'd add a third placeholder — e.g. `@import 'virtual:vendure-user-styles';` — at the intended spot in `styles.css`, and replace *that* with the user's `@import` statements. This matches a marker we put there on purpose, removes the dependency on the position (or presence) of the tailwindcss import, and keeps the injection mechanism consistent with the rest of the file.

(Minor, while you're in there: `vite-plugin-theme.ts` declares `resolvedVirtualModuleId` but never registers `resolveId`/`load` — it's dead code from a since-abandoned approach. Happy for that to be tidied up in this PR or left for a separate cleanup.)

The mechanism, idempotency, path-normalisation, and the test coverage are all solid — these are about fitting it into the existing conventions, not reworking the core. Thanks again!

tbouliere-datasolution:
Hello @michaelbromley.
I've just applied the requested changes.

michaelbromley (approved):

michaelbromley:
Thank you!

Issue #4761, opened 2026-05-21
Dependency audit: reduce supply-chain surface area


michaelbromley:
# Dependency audit: reduce supply-chain surface area

## Motivation

In light of the wave of npm supply-chain attacks over the past few weeks, we
want to systematically reduce Vendure's dependency footprint. Existing
defences (install-cooldown, Bun's refusal to run lifecycle scripts) help, but
attack methods will evolve. Fewer dependencies is simply a better security
posture — every direct dep is a maintainer we trust transitively.

Concrete goals:

- Remove dependencies we don't actually use.
- Swap dependencies for alternatives with smaller transitive trees.
- Vendor or reimplement small, self-contained dependencies where appropriate.
- Replace dependencies whose functionality is now available as a Node built-in.

## Baseline numbers

(Generated by `scripts/dependency-tracker/snapshot.mjs` against `package-lock.json`.
Detailed running ledger lives at `scripts/dependency-tracker/CHANGES.md` on the
`chore/dependency-audit` branch.)

- **Total unique production packages across the repo:** ~1,880
- **Per-package transitive footprint (production only):**

  | Package | Direct deps | Unique transitive (prod) |
  |---------|------------:|-------------------------:|
  | `@vendure/core` | 41 | 453 |
  | `@vendure/email-plugin` | 7 | 290 |
  | `@vendure/telemetry-plugin` | 9 | 162 |
  | `@vendure/harden-plugin` | 2 | 153 |
  | `@vendure/cli` | 11 | 92 |
  | `@vendure/admin-ui-plugin` | 3 | 71 |
  | `@vendure/graphiql-plugin` | 1 | 65 |
  | `@vendure/create` | 11 | 46 |
  | `@vendure/testing` | 8 | 34 |
  | `@vendure/job-queue-plugin` | 2 | 23 |
  | `@vendure/asset-server-plugin` | 3 | 20 |
  | `@vendure/dashboard` | 67 | 838 (mostly build-time) |
  | `@vendure/ui-devkit` | 11 | 810 (Angular) |
  | `@vendure/admin-ui` | 48 | 115 |

### 🔴 Critical: wrong-categorisation / dead deps (zero-risk wins)

| # | Package | Problem | Action | Saves |
|---|---------|---------|--------|------:|
[...]

PR #4771, opened 2026-05-21
refactor(core): Drop @nestjs/terminus dependency

michaelbromley:
## Summary

- Vendors the four symbols we actually used from `@nestjs/terminus` (`HealthIndicatorFunction`, `HealthIndicatorResult`, `HealthIndicator`, `HealthCheckError`) into `packages/core/src/health-check/terminus-compat.ts` (~80 LOC, zero runtime dependencies) and removes `@nestjs/terminus` as a direct dependency of `@vendure/core`.
- Removes six packages from every published install of `@vendure/core`: `@nestjs/terminus`, `boxen`, `check-disk-space`, `ansi-align`, `cli-boxes`, `widest-line`.
- Replaces the `TypeOrmHealthIndicator` usage in `TypeORMHealthCheckStrategy` with a direct `dataSource.query('SELECT 1')` racing against a `setTimeout`. The indicator function is still never invoked at runtime (see context below) but the implementation is now self-contained.

## Context

PR #4442 deprecated the built-in health check feature and gutted `HealthController.check()` so it just returns `{ status: 'ok' }` — no registered indicators are executed. `@nestjs/terminus` was being kept around purely to provide compile-time types for an inert public-API surface. The feature is slated for removal in v4.0.0, so this PR is an interim cleanup that lets us drop the dependency one version cycle earlier, prepares the `health-check/` directory for clean removal in v4, and corrects docs that incorrectly claimed health checks were 'built on NestJS Terminus'.

## Footprint analysis

| Subtree member | Status |
| --- | --- |
| `@nestjs/terminus` | EXCLUSIVE — removed |
| `boxen` | EXCLUSIVE — removed |
| `check-disk-space` | EXCLUSIVE — removed |
| `ansi-align` | only depended on by `boxen` — removed |
| `cli-boxes` | only depended on by `boxen` — removed |
| `widest-line` | only depended on by `boxen` — removed |
| `chalk`, `string-width`, `type-fest`, `wrap-ansi`, `ansi-styles`, `strip-ansi`, `ansi-regex`, `color-convert`, `color-name`, `emoji-regex`, `is-fullwidth-code-point`, `supports-color`, `has-flag`, `camelcase` | anchored by other declarers — unchanged |

## Blast radius for plugin authors

| Tier | Pattern | Impact | Migration |
| --- | --- | --- | --- |
| 0 | No custom health-check work | None | — |
| 1 | Implement `HealthCheckStrategy` with inline indicator function | None — structural typing handles it | — |
| 2 | Import `HealthIndicatorFunction` / `HealthIndicatorResult` from `@nestjs/terminus` for type annotations | Install-time break if relying on transitive install | One-line swap to `from '@vendure/core'` |
| 3 | `extends HealthIndicator` directly from `@nestjs/terminus` | Install-time break | One-line: either swap import to `@vendure/core` or add `@nestjs/terminus` to own deps |
| 4 | `imports: [TerminusModule]` in a plugin | Install-time break | Add `@nestjs/terminus` to own deps |

All affected code was already non-functional at runtime since PR #4442 — the break is purely compile/install-time and the fix is one line in every case. The vendored types are structurally compatible with the originals.

## Test plan

- [ ] `cd packages/core && bun run test src/health-check/terminus-compat.spec.ts` — 8 new unit tests pass (`getStatus` shape, `HealthCheckError` instance/causes/name/message)
- [ ] `cd packages/core && npx tsc -p ./build/tsconfig.build.json --noEmit` — no new errors (the pre-existing `order.service.ts:1461` error is on `minor` already)
- [ ] `cd packages/job-queue-plugin && npx tsc -p tsconfig.build.json --noEmit` — clean
- [ ] `cd packages/dev-server && npx tsc --noEmit` — no new errors touching health-check, bullmq, or terminus paths
- [ ] Bullmq `RedisHealthIndicator` and `RedisHealthCheckStrategy` import from `@vendure/core` instead of `@nestjs/terminus` and compile

Relates to #4761

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4771"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781969342&installation_id=128408429&pr_number=4771&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4771&signature=f32c6dc5e8c0776c8cc5c33fea49f84237ae5bdaaccac996d2c90cae7794605f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


michaelbromley:
Nigel review pass — six findings, all addressed in commit c0b65ff76 (now on top of the doc-bot regen at 5024a490c, pushed as 6549f0850).

**Must-fix (both addressed):**

1. **`HealthCheckError.causes` was narrowed to `HealthIndicatorResult`** — terminus types it `any`. Widened to `any` so plugin authors can pass through arbitrary upstream error payloads (HTTP responses, DB driver objects) without forced casts.
2. **`isHealthCheckError = true` discriminator was missing** — terminus sets this in the constructor and plugin authors may use it instead of `instanceof` to handle cross-realm cases. Added back as a public property defaulting to `true`.

**Should-fix (both addressed):**

3. **Driver-matrix comment added** in `typeorm-health-check-strategy.ts` explaining why `SELECT 1` is unconditional — Vendure's driver matrix (postgres, mysql, sqlite, better-sqlite3, sqljs) all accept ANSI `SELECT 1`; the Oracle/SAP HANA variants present in terminus are not needed.
4. **Three structural choices pinned in unit tests** — the new tests assert (a) `isHealthCheckError` is `true`, (b) `causes` accepts non-conforming payloads, (c) `HealthIndicatorResult<Key, Status, OptionalData>` generic parameters compile and behave as expected.

**Low / Nit (both addressed):**

5. **`HealthIndicatorResult` generic params added** — `HealthIndicatorResult<Key extends string, Status extends HealthIndicatorStatus, OptionalData extends Record<string, any>>` matches terminus. The plain alias still resolves to the original shape so this is a strict superset.
6. **File-header package-count clarified** — was "5-package subtree", now reads "six packages in total" with the terminus itself + its five exclusive sub-deps explicitly listed.

**Skipped:**

7. Splitting the docs-regen commit (Nigel flagged as a nit, not a blocker). Left as-is.

Net effect: the parent commit's "one-line import-path migration" claim is now structurally accurate for every documented plugin pattern. Test count: 8 → 13.

PR #4773, opened 2026-05-22
feat(core,dashboard): add verifyCustomerAccount mutation to the admin API

latifniz:
# Description

Adds a `verifyCustomerAccount(id: ID!, password: String): VerifyCustomerAccountResult!` mutation
to the admin API, allowing an administrator to verify a customer account without an email
verification token.

This fills a gap when `authOptions.requireVerification: true` (the default) but the
customer never received the verification email: no email plugin configured in
dev/staging, the email landed in spam, or B2B onboarding where the admin creates
accounts on behalf of clients.

**Core:**
- Adds `verifyCustomerAccount(id: ID!, password: String)` to the admin GraphQL schema,
  returning a `VerifyCustomerAccountResult` union of `Customer`, `MissingPasswordError`,
  `PasswordAlreadySetError` and `PasswordValidationError`. The first two are new to the admin
  schema; they are declared there the same way `PasswordValidationError` already is.
- Adds `CustomerService.verifyCustomerAccount()`, which sets `user.verified = true`,
  clears the pending verification token, records a `CUSTOMER_VERIFIED` history entry
  and emits `AccountVerifiedEvent`. A Customer with no User (a guest checkout) has no account to
  verify and is rejected with a `UserInputError`.
- Adds `UserService.verifyUserWithoutToken()`, the counterpart to `verifyUserByToken()`, which
  owns the credential side of that. It loads the `select: false` `passwordHash`, since what it
  does depends on whether one is set:
  - No password on the credential: the `password` argument sets it. A native
    credential with an empty `passwordHash` cannot be logged into, so verifying such
    an account without a password would produce an account which is verified and
    still unusable. Omitting it returns a `MissingPasswordError` instead.
  - Password already set: passing a `password` returns a `PasswordAlreadySetError`.
    This is not a route to take over an account.
  - No native credential at all (an SSO-only User): the account is verified only without a
    `password`, since there is nothing to set it on.

  The password rules and the `passwordHash` query are shared with `verifyUserByToken()` and
  `nativeCredentialAwaitsActivation()` rather than duplicated.
- Verifying an account also ends the `refreshCustomerVerification` flow for it, since
  that mutation only issues a new token while one is pending. Requiring a password
  rather than making it optional is what keeps that from being a trapdoor: the
  account is left usable rather than verified, passwordless and unable to be sent
  another verification email.
- The `verified` flag does not short-circuit the mutation. The credential logic runs either way,
  and only the history entry and the event are guarded by the transition, so an account which is
  verified but has no password (the state left by `requireVerification: false`) can still be
  given one.
- `createCustomer` now returns `PasswordValidationError` rather than throwing it, and
  `CreateCustomerResult` includes it. The new password field on the create form is the first
  thing to make that path reachable from the UI.

**Admin UI (legacy):**
- `createCustomer`'s result switch had cases for `Customer` and `EmailAddressConflictError` and no
  default. Now that `PasswordValidationError` is returned rather than thrown, it would have matched
  nothing: no toast, no error, no navigation. It reports the message for any non-`Customer` member,
  preferring `validationErrorMessage` where there is one.

**Dashboard:**
- Adds a "Verify account" action to the customer detail page, shown only when the
  customer is not yet verified. It opens a dialog with an optional password field.

  Whether a password is required depends on whether the Customer already has one, and the
  `Customer` type carries nothing which says. So the server's answer is what settles it: a
  `MissingPasswordError` or `PasswordAlreadySetError` is shown next to the field with the dialog
  still open, rather than as a toast over a closed one. The customer who never received the
  verification email has already set a password, and that is the case the feature exists for.
- The customer create form now has a password field, passed to
  `createCustomer(input:, password:)`. `CustomerService.create()` verifies
  immediately when given a password, so B2B onboarding is a single step. This needed
  a new `transformCreateVariables` option on `useDetailPage`, whose create path
  previously only ever sent `input`.

**Tests:**
- E2E coverage for verification, the history entry, the event payload, the pending token being
  cleared, and idempotency (asserting that the history entry is not repeated and no second event
  fires).
- Every error branch is covered: no password given, password given when one is already set,
  a password which fails validation, and a Customer with no User. A caller without
  `UpdateCustomer` is refused, and a Customer on another Channel cannot be reached by id.
- The tests log in as the customer after verifying rather than only asserting the
  `verified` flag, and cover both an admin-created customer (no password) and a
  self-registered one (password already set).
- Two tests in the GHSA-wr5h-x3x6-4h23 suite: an SSO-only account cannot be given a password, and
  an administrator activating a credential registered against an SSO account discards the
  registering caller's password.
- Dashboard e2e tests for the verify dialog (the inline error, then the happy path), for creating a
  customer with a password, and for a password the server rejects.

Targets `minor`: new schema.

Relates to https://github.com/vendurehq/vendure/discussions/4756

# Breaking changes

`verifyCustomerAccount` is new, so its union return type breaks nothing.
`CreateCustomerResult` gains a `PasswordValidationError` member. A client which exhaustively
matches on that union will need a case for it; one which falls through to `... on ErrorResult`
is unaffected. Previously that condition was thrown rather than returned, so it could not be
handled at all.


Ryrahul:
I think this might already be covered by the existing admin customer creation flow in resolver/service layer. (password is missing in the dashboard ui)

[`CustomerService.create()](https://github.com/vendurehq/vendure/blob/master/packages/core/src/service/services/customer.service.ts#L210) is documented as:

> "This method is intended to be used in admin-created Customer flows."

and it already immediately verifies the customer when a password is provided by calling the existing verification flow internally. 

So from my understanding, admins can already create a verified customer account through the existing API/service behavior, previous angular admin ui already had this behaviour so maybe adding password in the dashboard ui may be the appropriate way to go ahead **( i am not a maintainer so , you should probably wait for maintainer take on it )**,


> This fills a gap when `authOptions.requireVerification: true` (the default) but
> the customer never received the verification email. for example when no email
> plugin is configured in dev/staging, the email landed in spam, or during B2B
> onboarding where the admin creates accounts on behalf of clients.

As for this , dev server has mailbox where you can check all the emails to verify , while for b2b cases it normally requires much more complex verification flow than this , or the above method told should be used

biggamesmallworld (dismissed):
Thanks for the PR, and for writing up the problem in #4756 first.

The feature is one we're open to, but as it stands the mutation doesn't fix the case the discussion leads with, so I'd like some changes before we take it.

## The account is verified but still can't log in

`verifyCustomerAccount` sets `user.verified = true` and clears the `verificationToken`, but never touches `passwordHash`.

For a customer created by an admin via `createCustomer` without a password, `passwordHash` is `''`. `NativeAuthenticationStrategy.verifyUserPassword` rejects an empty stored hash, so after the mutation runs the customer still cannot log in. That is exactly the B2B onboarding scenario in #4756: "sales teams create accounts on behalf of clients directly from the admin UI".

It also closes off the recovery paths. Clearing the token removes the only credential the customer could have used to set a password, and `refreshVerificationToken` is gated on `!user.verified` (customer.service.ts:534), so `refreshCustomerVerification` becomes a silent no-op for that customer from then on. The only route left is `requestPasswordReset`, which needs the working email setup the discussion says is missing.

The tests don't catch this because both fixtures are admin-created customers with no password, and every assertion is `verified === true`. None of them attempts a login.

The case that does work today is a customer who self-registered with a password and whose verification email went astray. That's a real case, just not the one the discussion opens with.

### What I'd like instead

- `verifyCustomerAccount(id: ID!, password: String)`. When the user's native auth method has an empty `passwordHash`, set the password rather than only clearing the token.
- An e2e test that logs in as the customer after verifying, not just asserting the `verified` flag.
- A decision on the `refreshCustomerVerification` trapdoor. Even with a password argument, verifying an account permanently disables that mutation for it. Worth a comment at minimum.

## Missing password in the dashboard create form

Separately, `createCustomerDocument` (packages/dashboard/src/app/routes/_authenticated/_customers/customers.graphql.ts:107) calls `createCustomer(input:)` and never passes the `password` argument the admin API already accepts. `CustomerService.create()` verifies immediately when a password is given, so wiring that through the create form solves B2B onboarding with no API change at all. Worth including here.

## Branch and generated files

- This is a `feat` with new schema, so please rebase onto `minor`. It's currently based on `master`.
- Please drop the hand-edited generated files and let codegen produce them: `packages/common/src/generated-types.ts`, `packages/admin-ui/src/lib/core/src/common/generated-types.ts`, `packages/core/e2e/graphql/graphql-env-admin.d.ts`, `schema-admin.json`.
- The branch is currently conflicting with its base.

biggamesmallworld:
Hi @latifniz! Thanks for the contribution, as we have recently been inundated with a considerable number of PR's recently, we have made the decision to begin following up on the fixes that we have identified from community member PR's. You will continue to receive credit for your contribution. We hope that this will speed up the process of including very necessary fixes into the platform as fast as possible. We believe that this is in the best interest of the platform and the Vendure community.

latifniz:
> Hi @latifniz! Thanks for the contribution, as we have recently been inundated with a considerable number of PR's recently, we have made the decision to begin following up on the fixes that we have identified from community member PR's. You will continue to receive credit for your contribution. We hope that this will speed up the process of including very necessary fixes into the platform as fast as possible. We believe that this is in the best interest of the platform and the Vendure community.

Hi @biggamesmallworld, thanks for the update!

I'd honestly lost track of this PR — I raised it a few months back, and after a community member mentioned it might already be covered by existing behavior, I let it sit.

Looks like you've already pushed the fixes needed, so I don't think there's anything on my end right now. But happy to take another look at the whole PR if it'd help, just let me know.

Thanks again for picking this up and moving it forward!

biggamesmallworld (approved):

PR #4774, opened 2026-05-22
feat(cli): Add dev/build/start lifecycle commands and a distributable agent skill

dlhck:
## Summary

Adds first-class Vendure CLI lifecycle commands, wires generated/dev projects to those commands, hardens related dashboard/dev-server behavior, and packages a distributable `vendure-cli` agent skill.

### CLI lifecycle commands
- Adds `vendure dev` for server, worker, and Dashboard targets, with custom entry files, Vite config support, Node inspector flags, backend file reloads, `.env` reloads, dashboard-extension reload exclusions, and `--no-reload`.
- Adds `vendure build` for server, worker, and Dashboard targets, with tsconfig discovery, separate worker tsconfig support, `--experimental-tsgo`, `--clean`, `--watch`, `--verbose`, and `--no-progress`.
- Adds `vendure start` for compiled server and worker entrypoints, including prefixed child-process output and sibling shutdown handling.
- Adds shared process utilities and test coverage for CLI child-process orchestration.

### Existing CLI commands
- Extends command declarations and README coverage for the new lifecycle commands.
- Hardens `add`, `migrate`, `schema`, and `codemod` behavior in non-interactive environments so agents and CI fail fast instead of hanging on prompts.
- Adds tests for non-interactive prompt handling, migration operation selection, and revert failures.
- Updates CLI package dependencies and build packaging so new command modules are shipped.

### Generated apps and dev server
- Updates `create-vendure-app` scripts to use `vendure dev`, `vendure build`, and direct production server start scripts.
- Removes generated-app dependency on `concurrently` for the standard server lifecycle.
- Updates the generated README to describe server, worker, and Dashboard startup/build behavior.
- Wires `packages/dev-server` scripts to the CLI lifecycle, adds server/worker/dashboard build scripts, adds build tsconfigs, and sets Dashboard build output to `dist/dashboard`.
- Adjusts `dev-config.ts` so Dashboard static assets resolve correctly from source and built dev-server layouts.

### Dashboard Vite integration
- Isolates Dashboard temporary compilation directories per process/run.
- Serializes config compilation for a given output path and writes temp `package.json` files atomically.
- Prevents duplicate config-loader work while a config load is already in flight.
- Ensures gql-tada temp directories exist before writing.
- Updates dashboard Vite tests and docs for the adjusted behavior.

### CI, docs, and lockfile
- Ensures CI installs Playwright Chromium even when the browser cache is restored.
- Updates `bun.lock` for the CLI/dev-server dependency changes.
- Regenerates `docs/manifest.json` and updates Dashboard Vite plugin docs.

### Agent skill
- Adds `skills/vendure-cli/` with `SKILL.md` and per-command references for `dev`, `build`, `start`, `add`, `migrate`, `schema`, and `codemod`.
- Documents package-manager runner selection for agents instead of hardcoding `npx`.
- Documents non-interactive command usage, long-running command cautions, production-start caveats, and current CLI edge cases.
- Updates the skill to match current CLI implementation details for project-root assumptions, `add` requirements, `dev --no-reload`, and codemod directory validation.

## Testing
- `bunx vitest --config vitest.config.mts --run src/commands/build/build.spec.ts src/commands/dev/dev.spec.ts src/commands/start/start.spec.ts`
- `bun run build` in `packages/common`
- `bun run build` in `packages/core`
- `bun run build` in `packages/cli`
- `bun run build` in `packages/create`
- `bun run build -- --clean --no-progress` in `packages/dev-server`
- `bun run build:server -- --watch --no-progress` smoke-tested and stopped
- `npx skills list` confirmed the `vendure-cli` skill is discovered
- `git diff --check -- skills/vendure-cli`

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4774"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782039706&installation_id=128408429&pr_number=4774&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4774&signature=d2937bb0f254d4638ac5f50d4e451d7428d0c9de8ca3277bc4d5663a44e794f7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `vendure dev` command to run Vendure in development mode with auto-reload for server, worker, and dashboard.
  * Added `vendure build` command to compile server, worker, and dashboard for production.
  * Added `vendure start` command to run compiled Vendure projects.
  * Added non-interactive mode support across commands.
  * Dashboard extension discovery and auto-reload during development.

* **Bug Fixes**
  * Fixed CLI executable permissions after local builds.
  * Improved concurrent compilation safety for dashboard artifacts.

* **Documentation**
  * Added comprehensive CLI command documentation for `dev`, `build`, `start`, `add`, `codemod`, `migrate`, and `schema`.
  * Updated project generation templates with new CLI commands.

* **Chores**
  * Added process management utilities for CLI commands.
  * Updated dev-server dependencies and TypeScript configurations.
  * Enhanced error handling and logging across CLI commands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vendurehq/vendure/pull/4774?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/build_and_test.yml
```diff
@@ -142,7 +142,6 @@ jobs:
                   restore-keys: |
                       playwright-${{ runner.os }}-
             - name: Install Playwright browsers
-              if: steps.playwright-cache.outputs.cache-hit != 'true'
```
dlhck, line 148: why has that been removed. If we have cached playwright browsers available we should not install again

packages/cli/src/commands/command-declarations.ts
```diff
@@ -159,6 +159,136 @@ export const cliCommands: CliCommandDefinition[] = [
             process.exit(0);
         },
     },
+    {
+        name: 'dev',
+        description: 'Run Vendure in development mode',
+        arguments: [
+            {
+                name: 'target',
+                description: 'Target to run: all, server, worker or dashboard',
```
dlhck: include that `all` is default

packages/cli/src/commands/command-declarations.ts
```diff
@@ -159,6 +159,136 @@ export const cliCommands: CliCommandDefinition[] = [
...
+                required: false,
+            },
+        ],
+        options: [
+            {
+                long: '--server-entry <path>',
+                description: 'Path to the server entry file',
```
dlhck: include that this needs to point to a TypeScript file that calls bootstrap(

packages/cli/src/commands/command-declarations.ts
```diff
@@ -159,6 +159,136 @@ export const cliCommands: CliCommandDefinition[] = [
...
+                required: false,
+            },
+            {
+                long: '--worker-entry <path>',
+                description: 'Path to the worker entry file',
```
dlhck: include that this needs to point to a TypeScript file that calls bootstrapWorker(

packages/cli/src/commands/command-declarations.ts
dlhck, line 1: in all the optional arguments and options I miss a hint to the default value

packages/cli/build.ts
```diff
@@ -12,6 +12,11 @@ for (const file of templateFiles) {
     fs.copyFileSync(file, distPath);
 }
 
+const cliBinPath = path.join(__dirname, 'dist', 'cli.js');
+if (fs.existsSync(cliBinPath)) {
+    fs.chmodSync(cliBinPath, 0o755); // NOSONAR - published CLI bin must be executable after install.
+}
+
```
dlhck, line 19: Not sure if this is really necessary, because it was never the case before and the package manager should take care of that.

packages/dev-server/package.json
dlhck, line 1: we should use `vendure` as the entrypoint for the CLI here so that we're using the package manager wiring

.github/workflows/build_and_test.yml
dlhck: Leaving this unconditional. The failing dashboard e2e run had a Playwright cache hit, but the cache was incomplete for the installed Playwright revision (`chromium_headless_shell-1208`), so the tests failed before starting. `bunx playwright install chromium` is idempotent with a valid cache and repairs stale or partial caches when the Playwright version changes.

packages/cli/build.ts
dlhck: I checked this while switching the dev-server scripts back to `vendure`. Without this chmod, `bun run build` in `packages/dev-server` fails with `/node_modules/.bin/vendure: Permission denied`, because Bun’s workspace bin is a symlink to `packages/cli/dist/cli.js` and `tsc` recreates that file without executable permissions. Package managers handle installed package shims, but not this local post-build symlink target, so I kept the chmod and clarified the comment.

michaelbromley (changes requested):
Thorough pass over this one. The architecture is sound — the `ManagedDevProcess` abstraction, the build-group sequencing that solves the Vite-before-TypeScript ordering, and the compiler serialisation lock in `compiler.ts` are all well thought out. But a few correctness bugs slipped in alongside, two of which I've verified against the source and consider blocking.

## 🔴 Must fix

**1. `schema.ts` sets the env var to the string `"undefined"` instead of removing it** — `packages/cli/src/commands/schema/schema.ts:38` and `:125`

```ts
process.env.VENDURE_RUNNING_IN_CLI = undefined;
```

`process.env` coerces every value to a string, so this sets the literal string `"undefined"`, which is **truthy**. Any downstream `if (process.env.VENDURE_RUNNING_IN_CLI)` check sees it as still set after the intended cleanup. `migrate.ts` does this correctly with `delete process.env.VENDURE_RUNNING_IN_CLI` (lines 78, 143). Use `delete` in both schema locations.

**2. `waitForDevProcesses` reports the wrong exit code on graceful shutdown** — `packages/cli/src/commands/dev/dev.ts:326-336`

On a SIGINT/SIGTERM-initiated shutdown the promise resolves with the exit code of whichever child closes **last** (`code ?? signalToExitCode(signal) ?? 0`), rather than a canonical shutdown code. `waitForBuildProcesses` in `build.ts` handles this correctly by tracking `firstNonZeroExitCode`/`shutdownExitCode`. Because `devCommand` pipes this straight to `process.exit`, the result is a non-deterministic shell return code (0 vs 130 vs 143) depending on close ordering, which affects automation/CI consumers.

## 🟠 Should fix

**3. Asymmetric `devDependencies` coverage in project resolution** — `dev.ts:143` vs `utilities/monorepo-utils.ts:94`

`hasVendureCoreDependency` (dev.ts) checks both `dependencies` and `devDependencies`, but `hasNamedDependency` (monorepo-utils) only checks `dependencies`. A monorepo where `@vendure/core` lives in `devDependencies` resolves on the direct path but silently falls back to `cwd` on the monorepo path, producing a confusing "could not find ./src/index.ts" error rather than a helpful one.

**4. `withCompileLock` is an unbounded queue** — `packages/dashboard/vite/utils/compiler.ts:55-74`

Each call chains `queued = previous.then(() => current)` with no "supersede" logic. Under rapid HMR bursts every stale recompile still executes in order after the current one finishes. Bounded in practice, but the comment describes this as serialisation when it's also an unbounded queue — a last-one-wins drop of already-queued work would match intent. Worth at least a doc note.

**5. Generic timeout utility hardcodes migrate-specific examples** — `utilities/utils.ts:72-86`

`withInteractiveTimeout` is called from both `migrateCommand` and `schemaCommand`, but the timeout message lists migrate-specific examples (`vendure add`, `vendure migrate -r`). When it fires from `schema`, the guidance shown is wrong. Suggest accepting an `examples` parameter (parallel to `abortIfNonInteractive`) or keeping the message generic.

**6. Dead branch in `resolveInspectAddress`** — `dev.ts:609-611`

The `inspectValue === false` branch is unreachable; `getInspectArgs` (line 589) already returns early when `inspectValue == null || inspectValue === false`. Either delete this branch or remove the upstream guard — currently both exist and neither trusts the other.

**7. Pointless passthrough wrapper** — `dev.ts:419-425`

`shouldIgnoreReloadWatchPath` only delegates to `isAlwaysIgnoredReloadPath` with an identical signature. Pass `isAlwaysIgnoredReloadPath` directly to chokidar's `ignored` option.

## 🟡 Test gaps

**8. `cli-process-utils.spec.ts`** tests only the happy path. For a module handling production process orchestration and signal semantics, it needs: non-zero exit propagation, `error`-event handling and sibling shutdown, SIGINT/SIGTERM forwarding + handler cleanup, and the simultaneous-close race on `settledChildren`.

**9. `build.spec.ts:545`** — the `validateTsConfig` "rejects invalid tsconfig" test uses bare `.toThrow()`, which accepts any error (file/JSON/diagnostic). Assert the specific message.

**10. `migrate.spec.ts`** asserts env cleanup on the happy path but not on the throw path — which is exactly the gap that allowed bug #1 to exist in `schema.ts`. Add a test that the `finally` block deletes the env var after an operation throws.

## ⚪ Nitpicks

- The Node default inspector port `9229` appears three times in `dev.ts` unnamed; extract a `DEFAULT_INSPECT_PORT` constant (the file already extracts `reloadDebounceMs`/`restartShutdownGraceMs` at module level).
- `e: any` catch blocks in the touched `migrate.ts` (line 70) and `schema.ts` (line 39) — prefer `unknown` + `instanceof Error` narrowing, per the repo convention to add explicit types only when necessary.

## Structural note

Bug #2 and the duplication around it share a root cause: there are now three near-identical "wait for N children, forward signals, resolve once" implementations — `waitForChildProcesses` (cli-process-utils), `waitForDevProcesses` (dev), and `waitForBuildProcesses` (build). The dev variant drifted and dropped the canonical-exit-code tracking the other two have. Worth considering whether the build/dev variants can extend the shared one with a hook (e.g. the progress renderer, the `ManagedDevProcess` wrapper) rather than copying the algorithm — that's where bug #2 lives, and where the next one will too.

dlhck:
@michaelbromley Thanks for the thorough review. Addressed in e1da40cba:

- fixed schema CLI env cleanup with delete/finally
- made dev shutdown preserve canonical SIGINT/SIGTERM exit codes
- included devDependencies in monorepo project resolution
- made interactive timeout guidance command-specific/generic
- removed the dead inspector branch and reload ignored wrapper
- documented the compile lock as serialized queue behavior rather than changing it to supersede queued compiles
- added focused regression coverage for schema/migrate env cleanup, process signal/error handling, monorepo devDependency resolution, and the tsconfig diagnostic assertion

Validation:
- bunx vitest --config vitest.config.mts --run src/commands/dev/dev.spec.ts src/commands/schema/schema.spec.ts src/commands/migrate/migrate.spec.ts src/shared/cli-process-utils.spec.ts src/commands/build/build.spec.ts
- bun run build in packages/cli

michaelbromley:
Found three related shutdown / dev-loop issues while testing `vendure dev all` locally. Have patches running locally and the parent now exits in <500ms after Ctrl-C with no zombies. All three changes are in `packages/cli/src/commands/dev/dev.ts`.

### 1. Dashboard child crashes with unhandled `read EIO` on Ctrl-C

```
[dashboard] Error: read EIO
[dashboard]     at TTY.onStreamRead (node:internal/stream_base_commons:216:20)
[dashboard] Emitted 'error' event on Interface instance
```

**Cause:** `spawnDevChild` uses `stdio: ['inherit', 'pipe', 'pipe']` when `prefixOutput` is true — i.e. in `dev all`. All three children inherit the same TTY stdin. Vite's `readline` keypress handler emits an unhandled `error` event when the parent exits and the shell reclaims the TTY. Also semantically wrong: three siblings can't share one interactive stdin.

**Fix:** detach stdin in multi-process mode.

```ts
stdio: options.prefixOutput
    ? ['ignore', 'pipe', 'pipe']  // dev all: no shared stdin
    : 'inherit',                   // single target: full TTY ownership
```

Vite skips its keypress handler when `process.stdin.isTTY === false`, so the EIO disappears.

### 2. Parent hangs forever if any child ignores SIGINT

Once #1 was fixed, Ctrl-C would deliver SIGINT but the parent would wait indefinitely for the dashboard child to close. Required a second Ctrl-C (which hit Node's default behaviour because the handler was registered with `process.once`).

**Cause:** the `stop` callbacks for both plain and supervised processes call `stopChild(child, signal)` — a bare `child.kill(signal)` with no SIGKILL fallback. A child that fails to handle SIGINT pins the parent.

**Fix:** reuse the existing `stopChildWithGrace` helper in both stop paths and drop the no-grace `stopChild`. Industry-standard SIGTERM/SIGINT → SIGKILL after `restartShutdownGraceMs` (5s) is consistent across both lifecycles now.

```ts
// startPlainDevProcess
const runningProcess = new ManagedDevProcess(signal => {
    stopChildWithGrace(child, signal, restartShutdownGraceMs);
});

// startSupervisedDevProcess
const runningProcess = new ManagedDevProcess(signal => {
    stopping = true;
    if (restartTimer) clearTimeout(restartTimer);
    closeWatcher();
    if (child) {
        stopChildWithGrace(child, signal, restartShutdownGraceMs);
    }
});
```

### 3. Generated `.d.ts` files trigger spurious server/worker restarts on startup

```
[server] Change detected in graphql/graphql-env.d.ts. Restarting server...
[worker] Change detected in graphql/graphql-env.d.ts. Restarting worker...
```

**Cause:** `shouldRestartOnFileChange` uses `path.extname(fileName)` which only looks at the last dot. For `graphql-env.d.ts` it returns `.ts` (matches `reloadFileExtensions`). Declaration files have no runtime effect — they should never trigger a runtime restart. gql-tada regenerates `graphql-env.d.ts` on every dashboard startup, so every fresh `dev all` triggers an immediate gratuitous restart of server + worker, which itself causes the dashboard's `esbuild` service to fall over with `The service is no longer running` mid-transform of `main.tsx`.

**Fix:** explicitly skip `.d.ts` / `.d.cts` / `.d.mts`:

```ts
const fileName = path.basename(filePath);
if (reloadIgnoredFileNames.has(fileName)) {
    return false;
}
// Type declaration files have no runtime effect; skip them so generated
// declaration files (e.g. gql-tada's graphql-env.d.ts) don't churn restarts.
if (/\.d\.[cm]?ts$/.test(fileName)) {
    return false;
}
```

**Note:** there's a closely related observation that gql-tada also generates `graphql/graphql.ts` (no `.d.`), which the `.d.ts` filter doesn't catch and which I saw trigger a restart in one test run (though that run was contaminated by zombie Vite processes from earlier attempts, so it isn't a clean repro). Worth investigating whether a generated-file marker check (e.g. `@generated` header sniff) or a configurable ignore-pattern would be more robust than enumerating filenames. Not blocking for the EIO/hang fixes above.

### Repro for all three

```bash
cd packages/dev-server
bun run dev
# wait for server up, then Ctrl-C
```

Pre-fix: dashboard EIO crash → parent hang → forced second Ctrl-C → multiple zombie processes survive (verifiable via `ps aux | grep vendure`).
Post-fix: clean shutdown, exit code 130, no surviving processes.

dlhck:
@michaelbromley fixed in 4e81a81f8.

This covers the three shutdown/dev-loop issues from your comment:

- `dev all` now detaches child stdin while keeping prefixed stdout/stderr, so Vite no longer shares the parent TTY and the Ctrl-C `read EIO` path is avoided.
- Plain and supervised dev processes now both use the existing grace shutdown path, so SIGINT/SIGTERM falls back to SIGKILL after the existing 5s grace period instead of allowing one stuck child to pin the parent.
- Runtime reload filtering now skips `.d.ts`, `.d.cts`, and `.d.mts` files, including `graphql-env.d.ts`; dashboard metadata discovery skips declarations too.

Also added focused coverage for the declaration-file restart filter and verified with:

```bash
bunx vitest --config packages/cli/vitest.config.mts --run packages/cli/src/commands/dev/dev.spec.ts
./node_modules/.bin/tsc -p packages/cli/tsconfig.cli.json --noEmit
git diff --check
```

michaelbromley (approved):

PR #4779, opened 2026-05-23
fix: add administrator and seller to createEntityConfigs

### LINKED ISSUE #4778: Relation-type custom fields in React Dashboard only render entity IDs instead of related entity details on Order Detail page
## Describe the bug

In the Order Detail page, relation-type custom fields behave differently between the legacy Angular Admin UI and the new React Dashboard.

In the old Angular Admin UI, relation custom fields displayed the related entity object with readable fields (e.g. `firstName`, `lastName`, `emailAddress`, etc.), making it easy to understand the assigned relation directly from the UI.

However, in the React Dashboard, the same relation custom field only displays the raw entity ID value instead of rendering the related entity details or selector UI.

Example:
- `assignedRider` relation field in Angular Admin UI shows the full `Administrator` entity details.
- The React Dashboard only renders `65` (the entity ID).

This makes relation custom fields much less usable in the new dashboard, especially for operational/admin workflows.

---

## To Reproduce

1. Create a relation-type custom field on `Order` (e.g. relation to `Administrator`)
2. Assign a related entity value
3. Open the Order Detail page in:
   - Legacy Angular Admin UI
   - New React Dashboard
4. Compare the rendering of the relation custom field

---

## Expected behavior

Relation-type custom fields should render similarly to the legacy Angular Admin UI:
- Display related entity information
- Show meaningful labels instead of only raw IDs

Example:

```text
Administrator: Super Admin
emailAddress: superadmin
```

instead of:

```text
65
```

---

## Actual behavior

The React Dashboard only renders the related entity ID for relation custom fields.

Example:

```text
65
```

No entity details, labels, or relation UI are displayed.

---

### Legacy Angular Admin UI

Image

### React Dashboard

Image

---

## Error logs

No error logs are produced.

---

## Environment

- @vendure/core version:3.6.3
- Nodejs version:v22.22.3
- Database: PostgreSQL
- Operating System: macOS
- Browser: Chrome
- Package manager: pnpm

---

## Configuration

Example custom field:

```ts
customFields: {
  Order: [
    {
      name: 'assignedRider',
      type: 'relation',
      entity: Administrator,
      nullable: true,
    },
  ],
}
```

---

## Minimal reproduction

1. Create a relation custom field on `Order`
2. Open the Order Detail page in the React Dashboard
3. Observe that only the relation ID is displayed

---

## Workaround

See the entity in list view , it actually has the expanded entity view there in the list 

---

## Additional context

This appears to be a parity gap between the legacy Angular Admin UI and the new React Dashboard for relation-type custom fields.

Happy to raise a PR for this .

Ryrahul:
# Description

Adds relation selector configurations for `Administrator` and `Seller` entity types in the Dashboard's `DefaultRelationInput` component.

Previously, relation-type custom fields targeting these entities (e.g. `{ name: 'assignedRider', type: 'relation', entity: Administrator }`) would fall back to a plain text input showing only the raw entity ID. Now they render a proper searchable selector with meaningful labels:

- **Administrator**: Shows `firstName lastName` as title, `emailAddress` as subtitle, searchable by last name
- **Seller**: Shows `name` as title, searchable by name

Fixes #4778

> **Note:** Other entity types with list queries (Country, Zone, StockLocation, TaxCategory, Role, Tag, etc.) could also be added as relation selector configs if there's demand. Happy to add more in a follow-up.

# Breaking changes

None.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4779"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782161137&installation_id=128408429&pr_number=4779&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4779&signature=475420f20dfacb101e48880acf2ef1f7fe964c3df5c4ee28bb5f1f5f1607ee6e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
+++ b/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
@@ -33,6 +33,11 @@ export interface RelationSelectorConfig<T = any> {
     multiple?: boolean;
     /** Custom filter function for search */
     buildSearchFilter?: (searchTerm: string) => any;
+    /**
+     * Logical operator applied to the search filter fields. Use 'OR' when
+     * `buildSearchFilter` returns multiple fields that should be matched independently.
+     */
+    searchFilterOperator?: 'AND' | 'OR';
     /** Custom filter function for fetching by IDs */
     buildIdsFilter?: (ids: string[]) => any;
     /** Custom label renderer function for rich display */
@@ -132,6 +137,9 @@ export function useRelationSelector<T>(config: RelationSelectorConfig<T>) {
             // Add search filter if there's a search term
             if (debouncedSearch.trim().length > 0) {
                 variables.options.filter = buildFilter(debouncedSearch.trim());
+                if (config.searchFilterOperator) {
+                    variables.options.filterOperator = config.searchFilterOperator;
+                }
             }
 
             const response = (await api.query(config.listQuery, variables)) as any;
--- a/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
+++ b/packages/dashboard/src/lib/components/data-input/relation-selector.tsx
@@ -33,11 +33,6 @@ export interface RelationSelectorConfig<T = any> {
     multiple?: boolean;
     /** Custom filter function for search */
     buildSearchFilter?: (searchTerm: string) => any;
-    /**
-     * Logical operator applied to the search filter fields. Use 'OR' when
-     * `buildSearchFilter` returns multiple fields that should be matched independently.
-     */
-    searchFilterOperator?: 'AND' | 'OR';
     /** Custom filter function for fetching by IDs */
     buildIdsFilter?: (ids: string[]) => any;
     /** Custom label renderer function for rich display */
@@ -137,9 +132,6 @@ export function useRelationSelector<T>(config: RelationSelectorConfig<T>) {
             // Add search filter if there's a search term
             if (debouncedSearch.trim().length > 0) {
                 variables.options.filter = buildFilter(debouncedSearch.trim());
-                if (config.searchFilterOperator) {
-                    variables.options.filterOperator = config.searchFilterOperator;
-                }
             }
 
             const response = (await api.query(config.listQuery, variables)) as any;
--- a/packages/dashboard/src/lib/components/data-input/default-relation-input.tsx
+++ b/packages/dashboard/src/lib/components/data-input/default-relation-input.tsx
@@ -548,6 +548,56 @@ const createEntityConfigs = (i18n: any) => ({
             );
         },
     }),
+
+    Administrator: createRelationSelectorConfig({
+        ...createBaseEntityConfig('Administrator', i18n, 'emailAddress', 'lastName'),
+        listQuery: graphql(`
+            query GetAdministratorsForRelationSelector($options: AdministratorListOptions) {
+                administrators(options: $options) {
+                    items {
+                        id
+                        firstName
+                        lastName
+                        emailAddress
+                    }
+                    totalItems
+                }
+            }
+        `),
+        label: (item: any) => (
+            <EntityLabel
+                title={`${item.firstName} ${item.lastName}`}
+                subtitle={item.emailAddress}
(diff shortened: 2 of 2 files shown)


BibiSebi:
Pushed a couple of follow-up fixes in c0888ddf6:

**Administrator relation selector — multi-field search**
Previously the config sorted by `emailAddress` but only filtered search by `lastName`, while displaying the full name — so typing a first name or email returned nothing. Replaced the single-field `createBaseEntityConfig` setup with a custom `buildSearchFilter` that matches the term against `firstName`, `lastName`, and `emailAddress`, combined with `searchFilterOperator: 'OR'`. Now any of those fields will match (same pattern as the existing `CustomerSelector`). Sorting is by `lastName`.

**Seller label — removed duplicate subtitle**
The Seller label rendered the name as both title and subtitle. Since Seller has no meaningful secondary field, the subtitle is now an empty string.

**Plumbing**
Added an optional `searchFilterOperator?: 'AND' | 'OR'` to `RelationSelectorConfig`; when set it's applied to `options.filterOperator` alongside the search filter. Generic, so other entity configs can opt into multi-field search the same way.

BibiSebi (approved):
Hey @Ryrahul thank you for your contribution 🙏 the PR looks great. I added some small improvements that we discussed with @michaelbromley.

BibiSebi:
Tested locally 

https://github.com/user-attachments/assets/c8388815-e911-4d2a-b48d-cbffbec92bd4

packages/dashboard/src/lib/components/data-input/relation-selector.tsx
```diff
@@ -33,6 +33,11 @@ export interface RelationSelectorConfig<T = any> {
     multiple?: boolean;
     /** Custom filter function for search */
     buildSearchFilter?: (searchTerm: string) => any;
+    /**
+     * Logical operator applied to the search filter fields. Use 'OR' when
+     * `buildSearchFilter` returns multiple fields that should be matched independently.
+     */
+    searchFilterOperator?: 'AND' | 'OR';
```
michaelbromley: Rather than adding this new config property, I think it is possible to use the existing `buildSearchFilter` but pass an or expression like:
```
buildSearchFilter: (term: string) => ({
  _or: [
   { firstName: { contains: term } },
   { lastName: { contains: term } },
   { emailAddress: { contains: term } },
  ],
})
```

michaelbromley (changes requested):

packages/dashboard/src/lib/components/data-input/relation-selector.tsx
michaelbromley: docs ref: https://docs.vendure.io/current/core/how-to/paginated-list#advanced-filtering

BibiSebi: Switched to the nested `_or` expression in `buildSearchFilter` and dropped the `searchFilterOperator` config property and its plumbing in `relation-selector.tsx`. Done in 1369427b5.

BibiSebi: https://github.com/user-attachments/assets/7afc1226-7fb4-4940-879b-1c732678d1b0

michaelbromley (approved):

PR #4781, opened 2026-05-24
fix(dashboard): prevent struct custom field overflow in narrow contai…

### LINKED ISSUE #4780: Struct custom field values overflow container in narrow layouts (e.g. order detail side column)
**Describe the bug**
Struct custom fields (e.g. location fields with latitude/longitude sub-fields) overflow their container when rendered in narrow layouts such as the order detail side column. Long numeric values like `85.36084579999999` break out of the card boundary and overlap adjacent UI elements.

**To Reproduce**
1. Define struct custom fields on the `Fulfillment` entity with float sub-fields (e.g. latitude/longitude)
2. Create an order, add payment, and create a fulfillment
3. Set the fulfillment struct custom field values to long decimals (e.g. `85.36084579999999`)
4. Go to the order detail page in the React Dashboard
5. Scroll to the "Fulfillment details" section in the side column
6. See the struct field values overflow outside the card container

**Expected behavior**
Long numeric values in struct custom field display mode should wrap or break within the card boundaries, staying contained in the side column layout.

**Actual behavior**
The `<dd>` elements inside the struct field's `grid grid-cols-2` layout overflow horizontally because CSS grid children default to `min-width: auto` (preventing shrink below content size), and there is no word-break rule applied. This causes long unbroken numbers to push past the grid track and card boundary.

**Screenshots/Videos**

Image

- @vendure/core version: 3.6.3
- Nodejs version: v22
- Database: PostgreSQL
- Operating System: macOS
- Browser: Chrome / Safari
- Package manager: bun

**Additional context**
This affects any struct custom field rendered in a narrow container — not just Fulfillment. The side column in `PageLayout` is `col-span-1` of a 4-column grid (~25% page width), making it particularly susceptible to overflow from long values.

Ryrahul:
# Description

Fixes #4780

Struct custom field values (e.g. coordinates like `85.36084579999999`) overflow their container when rendered in narrow layouts such as the order detail side column. This is caused by CSS grid children defaulting to `min-width: auto` and no word-break rule being applied to the `<dd>` value cells.

**Changes:**
- Added `overflow-hidden` to the struct field display mode container as a safety net
- Added `min-w-0 break-all` to `<dd>` value cells so long unbroken numbers wrap within their grid track

This affects any struct custom field rendered in a narrow container, not just Fulfillment.

# Breaking changes

None.

# Screenshots

image

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4781"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782246292&installation_id=128408429&pr_number=4781&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4781&signature=ba9f5f4d1be5b38f5bbcf4d67c8ffcf190f1d66da9b453801e8a7bec6215f6a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/data-input/struct-form-input.tsx
+++ b/packages/dashboard/src/lib/components/data-input/struct-form-input.tsx
@@ -61,10 +61,10 @@ function DisplayMode({
             <dl className="grid grid-cols-2 divide-y divide-muted -mt-2">
                 {fieldDef.fields.map(structField => (
                     <React.Fragment key={structField.name}>
-                        <dt className="text-sm font-medium text-muted-foreground py-2">
+                        <dt className="text-sm font-medium text-muted-foreground py-2 min-w-0">
                             {getTranslation(structField.label) ?? structField.name}
                         </dt>
-                        <dd className="text-sm text-foreground py-2">
+                        <dd className="text-sm text-foreground py-2 min-w-0 break-words">
                             {formatFieldValue(watchedStructValue[structField.name], structField)}
                         </dd>
                     </React.Fragment>


BibiSebi:
Thanks for tackling #4780 🙏  the fix works for the reported case, but a couple of pushbacks before merging:

**`overflow-hidden` on the outer container (line 47)** is redundant with the `min-w-0` fix on `<dd>` — the latter is what actually resolves the overflow. Layering it as a "safety mechanism" means any future tooltip, popover, dropdown, or focus ring rendered inside a struct value will be silently clipped with no visible symptom. I'd drop it and rely on `min-w-0` alone. If that turns out not to be sufficient in some case, that's a layout bug worth surfacing rather than hiding.

**`break-all` on the `<dd>` (line 67)** is too aggressive as a default. It breaks mid-word/mid-token — fine for the reported numeric case (`85.36084579999999`), but it makes strings, datetimes and joined list values render awkwardly at narrow widths (e.g. `Configurat\nion`, `5/2\n6/2026`). `break-words` (→ `overflow-wrap: break-word`) only breaks unbreakable tokens when there's no other option, which is what you actually want here. Ideally the choice would be driven by `structField.type` inside `formatFieldValue` (`break-all` for `int`/`float`, `break-words` otherwise), but a blanket `break-words` is a strict improvement.

**Also worth applying `min-w-0` to the `<dt>`** on line 64 — long translated labels can hit the same overflow.

Ryrahul:
> Thanks for tackling #4780 🙏 the fix works for the reported case, but a couple of pushbacks before merging:
> 
> **`overflow-hidden` on the outer container (line 47)** is redundant with the `min-w-0` fix on `<dd>` — the latter is what actually resolves the overflow. Layering it as a "safety mechanism" means any future tooltip, popover, dropdown, or focus ring rendered inside a struct value will be silently clipped with no visible symptom. I'd drop it and rely on `min-w-0` alone. If that turns out not to be sufficient in some case, that's a layout bug worth surfacing rather than hiding.
> 
> **`break-all` on the `<dd>` (line 67)** is too aggressive as a default. It breaks mid-word/mid-token — fine for the reported numeric case (`85.36084579999999`), but it makes strings, datetimes and joined list values render awkwardly at narrow widths (e.g. `Configurat\nion`, `5/2\n6/2026`). `break-words` (→ `overflow-wrap: break-word`) only breaks unbreakable tokens when there's no other option, which is what you actually want here. Ideally the choice would be driven by `structField.type` inside `formatFieldValue` (`break-all` for `int`/`float`, `break-words` otherwise), but a blanket `break-words` is a strict improvement.
> 
> **Also worth applying `min-w-0` to the `<dt>`** on line 64 — long translated labels can hit the same overflow.

Thanks  @BibiSebi  for the detailed feedback, have made the changes

michaelbromley (approved):

PR #4793, opened 2026-05-27
feat(dashboard): Add configurable Vite dev server timeout and force proxy options

taxilian:
Adds two new options to `DashboardPluginOptions`:

- **`viteDevServerTimeout`** (default: 1000) — Configurable timeout in milliseconds for detecting a running Vite dev server. Useful for slow or proxied connections (e.g. VPN) where the default 1s timeout is insufficient.
- **`viteDevServerForceProxy`** (default: false) — When true, skips Vite dev server detection entirely and always proxies requests to it. Eliminates per-request latency overhead when you know Vite is running.

Both options are backward-compatible and default to existing behavior.

---

## Context:

My frontend is a static site built using hugo; to avoid complications with CORS and cookies, I have it all on the same hostname. When developing, I use a proxy similar to ngrok (internal kubernetes black magic) to allow me to access everything over https on a dev domain.

The issue with this is that the default 1000ms vite check timeout periodically fails, likely due to the latency of going to the proxy, through the vpn, and to my laptop when I'm on some arbitrary connection somewhere.  Thus, while I'm troubleshooting dashboard updates I keep getting the "vite isn't running" page. I wanted the option to disable that check when in dev mode, and I believe this should provide that. If I'm missing something obvious let me know =]

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4793"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782513217&installation_id=128408429&pr_number=4793&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4793&signature=b9b8270c4499583afbfd23f6e52198809408861b692c16cb004447d02e5e1f33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


michaelbromley:
Hi,

Thanks for the PR. I’m not yet decided on this - the use-case seems quite niche and in general I am averse to adding new config options for very special cases.

What tends to happen is that people over time come up with more and more suggestions for top-level options to handle this and that.

Sometimes they are legit. But sometimes it requires stepping back and thinking, maybe there is another approach that will solve not only this particular problem, but also others too.

That’s where I’m at with this. But I don’t have time to think about that right now, so I’m leaving this open.

Just wanted to give some feedback so you know it’s on the radar at least.

taxilian:
I understand the perspective, and I'd completely agree if you can give me literally any alternative workaround for this hardcoded value which leaves me entirely unable to do meaningful development =]

You could alternately hardcode it to a higher value, I suppose. I honestly don't see much benefit to making it auto switch; I'd much prefer just setting a config parameter to tell it which to use. Auto switching is neat if it works, but when it doesn't work it leaves one completely unable to work with the system.

Just my $0.02

michaelbromley:
Thanks for the feedback. For now I suggest you use patch-package so you can implement your workaround in a safe way.

Issue #3548, opened 2025-05-18
Adding filter to include products in collections by facet values does not work on first save


robinjoh:
**Describe the bug**
As my title describes the issue i have noticed a problem with the vendure Admin UI when you add facet values to include products in a collection and save, the products are not properly included on the first save. But if i refresh the page and trigger an update on the page and save again it works. What i noticed is that on the first save no "apply-collection-filters" job is triggered in the job queue on the first save, but it is on the second save. 

**To Reproduce**
1. Go to a collection in Admin UI
2. Click "Add condition" in Filters section
3. Click "filter by facet values"
4. Add a facet of choice that products should be included by
5. Hit "Save"
6. Refresh the page
7. There are no products included in the contents list
8. Make any modification to the filter (for example check the "contains any" checkbox and uncheck it again).
9. Hit save
10. Now products should be listed in the contents list

**Expected behavior**
I expect the products to appear in the contents list on first save.

**Environment (please complete the following information):**
 - @vendure/core version: 3.3.0
 - Nodejs version 23
 - Database (mysql/postgres etc): PostgreSQL 17.1

**Additional context**
I have uploaded a recording of the reproduction of the issue.
Note that i have no customizations done in the Admin UI, i'm using the vanilla Admin UI with no plugins.

[Screen Recording 2025-05-18 at 11.11.24.zip](https://github.com/user-attachments/files/20272229/Screen.Recording.2025-05-18.at.11.11.24.zip)

PR #4794, opened 2026-05-28
fix(dashboard): initialize boolean configurable args

genm:
## Problem

In the React Dashboard, adding a collection `facet-value-filter` can leave the collection form invalid until the `Contains any` switch is manually toggled.

The switch is visually rendered as unchecked/off, but the underlying configurable operation argument can be initialized as an empty string (`""`). Since `containsAny` is a required scalar argument, the Create/Update button remains disabled even though the UI appears to show a valid `false` value.

Refs #3548

## Root cause

New configurable operation arguments are initialized from their `defaultValue` when present. For args without a default, list args are initialized as `[]`, but scalar args fall back to an empty string.

That means a required boolean arg without a `defaultValue` starts as:

```json
{ "name": "containsAny", "value": "" }
```

`BooleanInput` renders string values as checked only when the value is `"true"`, so `""` is shown as off/false. However, `ConfigurableOperationInput` validation correctly treats a required scalar `""` as invalid.

So the UI and validation disagree:

- UI: unchecked switch, looks like `false`
- internal value: `""`
- validation: invalid required scalar

## Fix

Introduce shared configurable operation arg initialization used by both the single and multi configurable operation selectors.

The initialization now preserves existing behavior for defaults and list args, while making boolean args explicit:

- `defaultValue != null` -> `defaultValue.toString()`
- list args -> `"[]"`
- boolean args without a default -> `"false"`
- other scalar args without a default -> `""`

This keeps the switch UI and required scalar validation aligned without changing `BooleanInput`, validation rules, or legacy stored-data handling.

## Test plan

- [x] `bun run --cwd packages/dashboard check-types`
- [x] `CI=true VITE_TEST_PORT=5176 bunx playwright test --config e2e/playwright.config.ts e2e/tests/catalog/collections.spec.ts --reporter=list`

The new E2E regression creates a collection with `facet-value-filter`, selects a facet value, does not toggle `Contains any`, verifies the Create button becomes enabled, saves the collection, and asserts the saved args include `containsAny: "false"`.

## Breaking changes

None.

## Screenshots

Not applicable. The regression is covered by an E2E test.

## Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4794"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782534887&installation_id=128408429&pr_number=4794&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4794&signature=e40556bf10f0fc67c80f6f5af0ce2c494a8f9e208ce059acb295c0edffe21869"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/catalog/collections.spec.ts
+++ b/packages/dashboard/e2e/tests/catalog/collections.spec.ts
@@ -293,3 +293,72 @@ test.describe('Issue #4389: Collection form dirty state with filters', () => {
         await expect(dp.formItem('Name').getByRole('textbox')).toHaveValue('E2E Filter Test Saved');
     });
 });
+
+test.describe('Issue #3548: Collection facet filter boolean args', () => {
+    let collectionId: string;
+
+    const detailPage = (page: Page) =>
+        new BaseDetailPage(page, {
+            newPath: '/collections/new',
+            pathPrefix: '/collections/',
+            newTitle: 'New collection',
+        });
+
+    test.afterEach(async ({ page }) => {
+        if (!collectionId) return;
+        const client = new VendureAdminClient(page);
+        await client.login();
+        await client.gql(
+            `
+            mutation ($id: ID!) { deleteCollection(id: $id) { result } }
+        `,
+            { id: collectionId },
+        );
+        collectionId = '';
+    });
+
+    // #3548 — A newly-added facet-value-filter should initialize the required
+    // containsAny boolean to the same explicit false value shown by the switch.
+    test('should allow saving a facet-value-filter without toggling Contains any', async ({ page }) => {
+        const client = new VendureAdminClient(page);
+        await client.login();
+        const { facetValues } = await client.gql(`
+            query { facetValues(options: { take: 1 }) { items { name } } }
+        `);
+        const facetValueName = facetValues.items[0].name as string;
+
+        const dp = detailPage(page);
+        await dp.gotoNew();
+        await dp.expectNewPageLoaded();
+
+        await dp.fillInput('Name', 'E2E Boolean Arg Filter');
+        await expect(dp.formItem('Slug').getByRole('textbox')).not.toHaveValue('', { timeout: 5_000 });
+
+        await page.getByRole('button', { name: /Add collection filter/i }).click();
+        await page.getByRole('menuitem', { name: /Filter by facet values/i }).click();
+        await page.getByRole('button', { name: /Add facet values/i }).click();
+        await page.getByPlaceholder('Search facet values...').fill(facetValueName);
+        await page.getByRole('option', { name: facetValueName, exact: true }).click();
+
+        await expect(dp.createButton).toBeEnabled({ timeout: 5_000 });
+        await dp.clickCreate();
+        await dp.expectNavigatedToExisting();
+        collectionId = new URL(page.url()).pathname.split('/').pop() ?? '';
+
+        const { collection } = await client.gql(
+            `
+            query ($id: ID!) {
+                collection(id: $id) {
+                    filters {
+                        code
+                        args { name value }
+                    }
+                }
+            }
+        `,
+            { id: collectionId },
+        );
+        const facetFilter = collection.filters.find((filter: any) => filter.code === 'facet-value-filter');
+        expect(facetFilter?.args).toEqual(expect.arrayContaining([{ name: 'containsAny', value: 'false' }]));
+    });
+});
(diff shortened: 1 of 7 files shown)


genm:

michaelbromley (changes requested):
Thanks for the fix, @genm! The root-cause analysis is spot on and the core idea — initialising boolean args to `"false"` so the validator and `BooleanInput` agree — is the right call.

I'd like a few changes before merging. Most of these come from the fact that once you extract a shared utility, it's worth making sure (a) it handles all the cases correctly and (b) all the existing inline implementations of the same logic actually use it.

### Requested changes

**1. Reorder the checks in `getInitialConfigArgValue` so `list` is handled before `defaultValue`.**

Currently a `list: true` arg with a `defaultValue` returns a scalar string, not a JSON array. For example, `{ type: 'boolean', list: true, defaultValue: true }` returns `'true'` instead of `'[true]'`. Downstream code (e.g. `JSON.parse(arg.value)` in `decodeConfigurableOperationIds` on the core side) expects an array for list args and will throw on a scalar. No built-in operation hits this combination today, but the utility should be correct by construction since it's now the single source of truth.

Suggested order:

```ts
if (arg.list) {
    // If a defaultValue is present, wrap it in an array; otherwise empty list.
    return arg.defaultValue != null ? JSON.stringify([arg.defaultValue]) : '[]';
}
if (arg.defaultValue != null) return arg.defaultValue.toString();
if (arg.type === 'boolean') return 'false';
return '';
```

**2. Migrate the other two consumers to use the new utility.**

The same broken initialisation pattern lives in:

- `packages/dashboard/src/app/common/duplicate-entity-dialog.tsx` (around line 57)
- `packages/dashboard/src/app/routes/_authenticated/_orders/components/fulfill-order-dialog.tsx` (around line 123)

Both do `arg.defaultValue != null ? arg.defaultValue.toString() : ''`, which is the exact bug this PR fixes for collections. If a duplicator or fulfillment handler ships with a required boolean arg, the same UX issue reproduces. Please update both to call `getInitialConfigArgValue()` so the fix actually covers all entry points.

**3. Strengthen the e2e test so it exercises the UI-driven initialisation path.**

The current test creates the collection filter directly via the API with `containsAny: 'false'`, then verifies the Update button enables when the name changes. That's a useful test, but it doesn't actually exercise `getInitialConfigArgValue()` — if the utility regressed to returning `''`, the test would still pass.

What I'd like is a test that reproduces the original user-facing bug:

1. Open the new-collection form
2. Add the `facet-value-filter` via the filter dropdown
3. Pick a facet value
4. Assert that the Create/Save button becomes enabled **without** toggling the `Contains any` switch
5. Save and verify `containsAny: 'false'` is persisted

That path is the one #3548 is about, and it's the path the new utility lives on.

### Out of scope but worth noting

There's a pre-existing concern about legacy persisted records that may have `''` stored for a boolean arg. When loaded for edit they render fine (unchecked), but on save the server's `JSON.parse('')` throws and validation fails for *required* boolean args. Not something I'm asking you to fix here — it's a separate data-shape issue — but worth being aware of.

Thanks again for the contribution! Once these three changes are in I'll be happy to approve.

— Michael

genm:
Thanks for the detailed review, Michael. I pushed a follow-up commit addressing the three requested changes:

- `getInitialConfigArgValue()` now handles `list` args before scalar defaults, and wraps list defaults with `JSON.stringify([arg.defaultValue])`.
- `duplicate-entity-dialog.tsx` and `fulfill-order-dialog.tsx` now use the shared initializer as well.
- The collection regression test name/assertions now make the UI-driven initialization path explicit: it adds the filter through the Dashboard UI, verifies the `Contains any` switch is off without toggling it, confirms Create becomes enabled, saves, and checks `containsAny: "false"` is persisted.

I also added a small unit test for the initializer to cover list defaults, boolean scalar fallback, scalar defaults, and empty scalar fallback.

Validated locally:

- `bun run --cwd packages/dashboard check-types`
- `bun run --cwd packages/dashboard test -- src/lib/components/shared/configurable-operation-utils.spec.ts`
- `CI=true VITE_TEST_PORT=5176 bunx playwright test --config e2e/playwright.config.ts e2e/tests/catalog/collections.spec.ts --reporter=list`

michaelbromley (approved):

PR #3196, opened 2024-11-06
feat(admin-ui): Set default shipping and billing address for draft orders

oliverstreissi:
# Description

Set default shipping and billing address for draft orders

# Breaking changes

No

# Screenshots

You can add screenshots here if applicable.

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [x] I have added or updated test cases
- [x] I have updated the README if needed

Issue #2342, opened 2023-08-11
Select default shipping and billing address for draft orders


martijnvdbrug:
**Is your feature request related to a problem? Please describe.**
It's a bit annoying that you have to select a shipping and billing address when creating a draft order for existing customers, because they do have default addresses.

**Describe the solution you'd like**
It would be nice if a customer's default address is set on a draft order when I select an existing customer.

**Describe alternatives you've considered**
Doing it manually...

PR #4810, opened 2026-06-06
fix(dashboard): Set default addresses when selecting customer on draft order

Lennix:
# Description

## Summary

In the Angular Admin UI, selecting a customer on a draft order automatically populates the order's shipping and billing addresses from the customer's default addresses (added in v3.1.0 via #3196, closing #2342). This behavior was forgotten when the draft order page was reimplemented for the React dashboard: only the *clear-addresses-on-customer-change* half was ported. The `setCustomerForDraftOrder` `onSuccess` handler in `orders_.draft.$id.tsx` unconditionally unsets any previously-set addresses and never looks up the new customer's defaults — so selecting a customer always leaves the draft with no addresses, even when defaults are flagged.

This PR ports the admin-ui logic (`draft-order-detail.component.ts` → `setCustomer()`): after the customer is set, fetch the customer's addresses, then set the flagged default shipping/billing address on the draft order — or clear the previously-set address if the new customer has no default, so changing customers never leaves the old customer's address behind.

One small deviation from the admin-ui version: where admin-ui always fires set-or-unset, this only calls unset when an address was actually set before (preserving the dashboard's existing guard), avoiding redundant mutations and spurious "address unset" toasts on fresh drafts. The end state is identical in all cases.

## Changes

- `orders_.draft.$id.tsx`: on `setCustomerForDraftOrder` success, query the customer's addresses and call `setDraftOrderShippingAddress`/`setDraftOrderBillingAddress` with the flagged defaults, falling back to the existing unset behavior when no default exists. Extracts a `mapToAddressInput()` helper (same field mapping as admin-ui's, incl. the `defaultShippingAddress`/`defaultBillingAddress` flags) which also replaces the two duplicated inline mappings in the manual address-select paths.
- `orders.graphql.ts`: add `customer { id }` to the `SetCustomerForDraftOrder` selection set; move the `GetCustomerAddresses` document here from `customer-address-selector.tsx` so the page and the selector share one definition.
- `e2e/tests/sales/orders.spec.ts`: regression test — selecting a customer with flagged default addresses populates both address blocks; switching to a customer without defaults clears them and offers the manual selectors again.

A nice side effect: since the shipping address is now set right after customer selection, eligible shipping methods appear immediately, removing a manual step from the draft order flow.

# Breaking changes

None. Dashboard-only behavioral fix restoring Admin UI parity.

# Screenshots

Selecting a customer with default addresses on a fresh draft order now populates both address blocks (previously both remained "No address").

# Checklist

📌 Always:

👍 Most of the time:

## Test plan

- New e2e test fails on `master` without the fix, passes with it
- Full `e2e/tests/sales/orders.spec.ts` suite passes (15/15) — the seeded e2e customers have no default-flagged addresses, so the existing manual address-select flow is unaffected
- Manually verified against the dev-server (sqlite): defaults populate on first selection; switching to a customer without defaults clears both addresses; a customer change replaces a previously-set address with the new customer's defaults; manual address selection still works
- `bun run check-types` clean

Relates to #2342 / #3196 (the admin-ui implementation of this behavior).

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4810"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783352746&installation_id=128408429&pr_number=4810&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4810&signature=592bbfe2b4d8684819412b2f710f26dac1fd34e6bdd8d3f33dd76edeea481238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/sales/orders.spec.ts
+++ b/packages/dashboard/e2e/tests/sales/orders.spec.ts
@@ -157,6 +157,80 @@ test.describe('Orders', () => {
         await expect(page.getByTestId('page-heading')).toBeVisible();
     });
 
+    // selecting a customer on a draft order should populate the order's addresses
+    // from the customer's default addresses (parity with the Angular admin-ui, see #3196)
+    test('should populate default addresses when selecting a customer', async ({ page }) => {
+        test.setTimeout(60_000);
+
+        const client = new VendureAdminClient(page);
+        await client.login();
+
+        // Create a customer whose address is flagged as default shipping & billing.
+        // The seeded e2e customers have addresses without default flags.
+        const { createCustomer } = await client.gql(
+            `mutation ($input: CreateCustomerInput!) {
+                createCustomer(input: $input) {
+                    ... on Customer { id }
+                    ... on ErrorResult { errorCode message }
+                }
+            }`,
+            {
+                input: {
+                    firstName: 'Daphne',
+                    lastName: 'Defaults',
+                    emailAddress: `daphne.defaults.${Date.now()}@test.com`,
+                },
+            },
+        );
+        await client.gql(
+            `mutation ($customerId: ID!, $input: CreateAddressInput!) {
+                createCustomerAddress(customerId: $customerId, input: $input) { id }
+            }`,
+            {
+                customerId: createCustomer.id,
+                input: {
+                    fullName: 'Daphne Defaults',
+                    streetLine1: '42 Default Lane',
+                    city: 'Defaultville',
+                    postalCode: 'D1 1AA',
+                    countryCode: 'GB',
+                    defaultShippingAddress: true,
+                    defaultBillingAddress: true,
+                },
+            },
+        );
+
+        // Create a draft order and select the customer
+        const lp = listPage(page);
+        await lp.goto();
+        await lp.expectLoaded();
+        await lp.newButton.click();
+        await expect(page).toHaveURL(/\/orders\/draft\//, { timeout: 10_000 });
+
+        await page.getByRole('button', { name: /Select customer/i }).click();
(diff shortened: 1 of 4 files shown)


Lennix:

Lennix:
recheck

michaelbromley:
Hi! Thanks for this improvement.

Testing this: shipping address gets set but billing fails with `GraphQL Request Error: Record has changed since last read in table 'order'`.

Cause: in the `setCustomerForDraftOrder` `onSuccess`, the shipping and billing mutations are fired with `mutate()` (fire-and-forget), so they race on the same version-tracked `Order`. Shipping commits first → bumps the version → billing was read at the old version → optimistic-lock failure.

Fix: sequence them. Use `mutateAsync` and `await` shipping, then billing, then a single `refreshEntity()` at the end (drop the eager `refreshEntity()` call). That also removes the flicker and the stacked toasts.

michaelbromley:
Thank you!

PR #4827, opened 2026-06-10
fix(dashboard): navigate to list page on channel switch when on detail page

### LINKED ISSUE #4826: fix(dashboard): Stale navigation history retains cross-channel entity links after channel switch
Describe the bug

I'm running a multi-channel Vendure setup (retail + wholesale channels) and noticed that when you switch channels in the admin dashboard, the top navigation history doesn't clear. It still shows links to entities from the channel you just left. If you click one of those stale links, you get a "Not Found" error because that entity doesn't belong to the new channel. The only way to recover is a full page refresh.

Attaching a screen recording to make it clearer.

To Reproduce

Log into the admin dashboard on the Default channel
Open any order (e.g. #ABC-123)
Switch to a different channel (e.g. Wholesale) using the channel switcher in the top-left
The top navigation still shows Order #ABC-123 from the previous channel
Click it — you get a "Not Found" error
Only a full page refresh or clicking a top-level nav item resets the state
Expected behavior

When you switch channels, the navigation history should either clear completely or filter itself to only show entities that belong to the newly selected channel.

Actual behavior

Navigation history persists across channel switches. The stale links are still clickable and crash with "Not Found" because they're being accessed from the wrong channel context.

Environment

@vendure/core version: 3.6.3
@vendure/dashboard version: 3.6.3
Node.js: 22
Database: PostgreSQL 16
OS: Linux
Browser: Chrome
Package manager: npm
Workaround

Do a full page refresh after switching channels, or navigate directly from the Orders list instead of using the nav history.

Additional context

Reproducible consistently in any multi-channel setup. Looks like the navigation history state is global and not tied to the active channel at all, so it never gets invalidated on switch.

Happy to submit a PR if someone can point me to the relevant component. I've contributed a couple of fixes to the storefront starter before, so I'm comfortable with the codebase style.

https://github.com/user-attachments/assets/06345214-85d0-46c6-a05d-20a13ba8f412

latifniz:
## What's the bug

When you're on a detail page (like /orders/ABC-123) and switch channels,
the router re-runs the loader for that same URL with the new channel token.
That order belongs to the old channel so the API returns null, the loader
throws, and you land on a broken error screen. Only a full page refresh
gets you out.

## Why it happens

The router doesn't know the channel changed. it just tries to reload
whatever URL you're currently on. But that entity doesn't exist in the
new channel, so it crashes.

## What the fix does

When you switch to a different channel, we check if you're on a detail
page. If you are, we navigate to the parent list before the loader runs:

- /orders/ABC-123 → /orders
- /products/XYZ → /products
- /customers/123 → /customers

If you're already on a list page, nothing changes, the list just
refetches with the new channel token automatically.

## What else this fixes

The same crash was happening across all detail pages, orders, products,
customers, collections, promotions. This one change covers all of them
since the logic is purely URL-based, no entity-specific code needed.

Fixes #4826

https://github.com/user-attachments/assets/fc9b5685-73e2-4f7f-8a83-d9621ab08fd5

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4827"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783682764&installation_id=128408429&pr_number=4827&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4827&signature=4b3067d8ab5acc430624422947024b6672af6105121e3814c77f9afa1d9f1cf8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
+++ b/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
@@ -24,7 +24,7 @@ import { useSortedLanguages } from '@/vdb/hooks/use-sorted-languages.js';
 import { useUserSettings } from '@/vdb/hooks/use-user-settings.js';
 import { cn } from '@/vdb/lib/utils.js';
 import { Trans } from '@lingui/react/macro';
-import { Link } from '@tanstack/react-router';
+import { Link, useNavigate, useRouterState } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
 import { ManageLanguagesDialog } from './manage-languages-dialog.js';
 
@@ -58,6 +58,8 @@ export function ChannelSwitcher() {
         setContentLanguage,
     } = useUserSettings();
     const [showManageLanguagesDialog, setShowManageLanguagesDialog] = useState(false);
+    const navigate = useNavigate();
+    const currentPath = useRouterState({ select: s => s.location.pathname });
     const displayChannel = activeChannel;
 
     // Get available languages from server config
@@ -82,7 +84,18 @@ export function ChannelSwitcher() {
 
     const renderChannel = (channel: (typeof channels)[number]) => (
         <div key={channel.code}>
-            <DropdownMenuItem onClick={() => setActiveChannel(channel.id)} className="gap-2 p-2">
+            <DropdownMenuItem
+                onClick={() => {
+                    if (channel.id !== displayChannel?.id) {
+                        setActiveChannel(channel.id);
+                        const segments = currentPath.split('/').filter(Boolean);
+                        if (segments.length > 1) {
+                            navigate({ to: `/${segments[0]}` as any });
+                        }
+                    }
+                }}
+                className="gap-2 p-2"
+            >
                 <div
                     className={cn(
                         'flex size-8 items-center justify-center rounded border',
--- a/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
+++ b/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
@@ -24,7 +24,7 @@ import { useSortedLanguages } from '@/vdb/hooks/use-sorted-languages.js';
 import { useUserSettings } from '@/vdb/hooks/use-user-settings.js';
 import { cn } from '@/vdb/lib/utils.js';
 import { Trans } from '@lingui/react/macro';
-import { Link } from '@tanstack/react-router';
+import { Link, useNavigate, useRouter, useRouterState } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
 import { ManageLanguagesDialog } from './manage-languages-dialog.js';
 
@@ -58,6 +58,9 @@ export function ChannelSwitcher() {
         setContentLanguage,
     } = useUserSettings();
     const [showManageLanguagesDialog, setShowManageLanguagesDialog] = useState(false);
+    const navigate = useNavigate();
+    const router = useRouter();
+    const currentPath = useRouterState({ select: s => s.location.pathname });
     const displayChannel = activeChannel;
 
     // Get available languages from server config
@@ -82,7 +85,20 @@ export function ChannelSwitcher() {
 
     const renderChannel = (channel: (typeof channels)[number]) => (
         <div key={channel.code}>
-            <DropdownMenuItem onClick={() => setActiveChannel(channel.id)} className="gap-2 p-2">
+            <DropdownMenuItem
+                onClick={() => {
+                    if (channel.id !== displayChannel?.id) {
+                        setActiveChannel(channel.id);
+                        const segments = currentPath.split('/').filter(Boolean);
+                        if (segments.length > 1 && channel.code !== DEFAULT_CHANNEL_CODE) {
+                            const target = `/${segments[0]}`;
+                            const exists = router.matchRoutes(target).length > 0;
+                            navigate({ to: (exists ? target : '/') as any });
+                        }
+                    }
+                }}
+                className="gap-2 p-2"
+            >
                 <div
                     className={cn(
                         'flex size-8 items-center justify-center rounded border',
--- a/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
+++ b/packages/dashboard/src/lib/components/layout/channel-switcher.tsx
@@ -24,7 +24,7 @@ import { useSortedLanguages } from '@/vdb/hooks/use-sorted-languages.js';
 import { useUserSettings } from '@/vdb/hooks/use-user-settings.js';
 import { cn } from '@/vdb/lib/utils.js';
 import { Trans } from '@lingui/react/macro';
-import { Link, useNavigate, useRouterState } from '@tanstack/react-router';
+import { Link } from '@tanstack/react-router';
 import { useEffect, useState } from 'react';
 import { ManageLanguagesDialog } from './manage-languages-dialog.js';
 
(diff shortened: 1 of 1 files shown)


Ryrahul:
If i switch to default channel , lets say from channel X . I think it should be fine for a user to be on the same page , as default channel has all the records

latifniz:
yes you will be on same page. if you are on a detail page say order X detail page and then you swtich to any other channel that don't have that order X then it should switch you to the main order page, thats what this PR does

Ryrahul:
> yes you will be on same page. if you are on a detail page say order X detail page and then you swtich to any other channel that don't have that order X then it should switch you to the main order page, thats what this PR does

Switching to the default channel should keep you on the same page.
The default channel (__default_channel__) has access to all records across all channels. So if I'm on /orders/ABC-123 and switch to the default channel, that order will still exist and load fine -- no need to redirect.

Example:
1. I'm on Channel X, viewing /orders/ABC-123
2. I switch to the default channel
3. Current behavior: I get redirected to /orders (unnecessary -- the default channel can see that order)
4. Expected behavior: I stay on /orders/ABC-123 since the default channel has all records
The redirect should only happen when switching to a non-default channel, where the entity might not exist.

nice fix!  overall

latifniz:
hmm maybe i don't have deep understanding of channels yet but I don't see all records on the default channels, I have 3 channels, 
1. defualt
2. retail
3. wholesell
and when i am on retail channel viewing order ABC and then I switch to channel wholesell I should go to main orders page because that ABC order does not exit in that channel, Before this fix, you got stale nav bar ID for order and also 404 Errors, 
Oh wait yeah now i get it, so the default channel is like a global channel it has access to everything. 
so yes we should check that condition, like if we are switching from any X channel to default channel do nothing, but if we are switching from any X channel to any Y channel then we should go to the parent page.
like that?

Ryrahul:
> hmm maybe i don't have deep understanding of channels yet but I don't see all records on the default channels, I have 3 channels,
> 
> 1. defualt
> 2. retail
> 3. wholesell
>    and when i am on retail channel viewing order ABC and then I switch to channel wholesell I should go to main orders page because that ABC order does not exit in that channel, Before this fix, you got stale nav bar ID for order and also 404 Errors,
>    Oh wait yeah now i get it, so the default channel is like a global channel it has access to everything.
>    so yes we should check that condition, like if we are switching from any X channel to default channel do nothing, but if we are switching from any X channel to any Y channel then we should go to the parent page.
>    like that?

Yupp

latifniz:
got it happy to add that condition. should I?

michaelbromley:
@latifniz thanks for this fix, and yes, I agree with Rahul’s assessment regarding the default channel. Go ahead and make that fix.

latifniz:
yes sure

latifniz:
done

latifniz:
re check

michaelbromley:
Logic is safe for all **built-in** detail routes — list (`entity.tsx`) and detail (`entity_.$id.tsx`) share a first segment by file-naming convention, so `/${segments[0]}` always resolves. Verified across all of them incl. deep nests (`/orders/$id/modify`, `/option-groups/$groupId/options/$id`, `/facets/$facetId/values/$id`). Skip-on-default is correct too (default channel is a superset, loader can't 404).

Two issues remain:

1. **Extension routes can break.** Extension routes take an arbitrary `path` and are added as flat children under `_authenticated`, with no enforced list↔detail naming. A detail route at `/reviews/$id` whose list lives elsewhere will navigate to a non-existent `/reviews` → TanStack's bare not-found (no `defaultNotFoundComponent` is set), a dead-end. The `as any` on `to` hides this at build time.
2. **Overly broad.** Global (non-channel-aware) entities — countries, zones, tax rates, sellers, administrators, channels — never 404 on channel switch, yet still get bounced to the list, losing the user's place.

Suggested fix (smallest): guard the target before navigating and fall back to `/` if it doesn't resolve, e.g.

```ts
const target = `/${segments[0]}`;
const exists = navigate.router?.matchRoutes(target, {}).length > 0; // or inspect the route tree
navigate({ to: exists ? target : '/' });
```

More robust: derive the parent from `router.state.matches` instead of splitting the URL string — correct for both built-ins and extensions. Best long-term: handle the root cause in the loader (`throw redirect()` to the route's own parent list when the entity isn't found), so it covers any channel change, not just this dropdown.

latifniz:
ok let me implement this

latifniz:
Hey, I dug deeper into this and wanted to share what I found before we move forward.

**What the fix does**

When you switch channels while on a detail page, instead of crashing with "Not Found", it now navigates you to the parent list page:

- `/orders/ABC-123` on channel switch goes to `/orders`
- Deep nested routes like `/orders/ABC-123/modify` or `/facets/1/values/2` both land on the correct list page because we take `segments[0]` from the URL
- For extension routes, we check if the target route actually exists in the router before navigating. If it doesn't, we fall back to `/` so there's no dead-end or crash

**What it doesn't cover**

- If a plugin registers a detail page at `/rev/123` but its list lives at `/product-reviews`, we can't know that relationship from the channel switcher. We fall back to `/` safely so no crash, but the user doesn't land on the right list. There's no way to solve this without the extension explicitly declaring its parent.
- Non-channel-aware entities like administrators, countries and zones don't belong to a specific channel, so switching channels while on `/administrators/123` will unnecessarily bounce the user to `/administrators`. Minor inconvenience. We don't have a clear list of such entities on the frontend to filter against, so I left it as a known tradeoff for now.

**Alternatives I explored**

I also looked into using `router.state.matches`, walking the router's matched route tree backwards to find the first ancestor with no dynamic params. More correct conceptually since it uses the actual route tree instead of URL string parsing. But in practice it gives the same result for the unrelated extension route case, because if you're on `/rev/123`, `/product-reviews` simply isn't in the matched routes. The router has no record of that relationship. Both approaches hit the same wall.

**Overall**

This is a meaningful improvement over the original crash but not a perfect long-term solution. The extension route case might need a more structured approach, like extensions explicitly declaring their parent route. If anyone has a better idea I'd love to hear it before we settle on this.

Also worth noting, I don't have any extension routes in my own project and haven't used them, so I couldn't reproduce or test exactly how that scenario plays out. Would appreciate if someone with extensions set up could verify that side of things.

michaelbromley (approved):

michaelbromley (commented):
I have thought about this some more and I'm not fully happy with the behaviour of kicking you out to the list page when eg a product _does_ exist in the other channel.

I'm exploring a different approach that could handle this in a better way, and will update with results.

latifniz:
That would be great. would love to see a better approach for this

michaelbromley:
Superceded by https://github.com/vendurehq/vendure/pull/4874

Thanks @latifniz for the time you put into this nonetheless.

Issue #490, opened 2020-10-10
Option to make custom field channel-specific


chladog:
**Is your feature request related to a problem? Please describe.**
Make option for channel-specific custom fields. Currently change I make on channel1 applies to channel2 and other way around, however I need it to behave same as price on ProductVariant. (channel1 price and currency is independent of channel2 price and currency).
I need this independency only for **some** CFs, most of them are supposed to be in-sync between channels

**Describe the solution you'd like**
Add new configuration property to custom field **global?**: boolean (defaults to true). If set to **false** then CF would be saved in separate table holding its values for each channel (analogous to current price implementation).
When switching channels in admin UI these values would get switched as well.

**Describe alternatives you've considered**
As I see in DB prices are saved in separate table with channel mapping. Actually I need this for additional price fields as well (used to define price for bulks). Therefore what I would basically need is "customPriceField". However I believe option for channel-specific custom fields would be handy for any other data type as well.

incorvia:
I wonder if an alternative implementation might look like storing the properties in a jsonb column:

1. Letting the admin configure the properties in the admin interface and what channels the properties were allowed 
2. Using those configured values to validate what values could be set within the jsonb column and 
3. Possibly using these values as a filter when presenting them as well.

michaelbromley:
@incorvia I think the idea is worth exploring. Here are my initial thoughts:

### Config-based custom fields (status quo)

* Pro: custom fields are part of the repo, under source control
* Pro: being in config also makes it easy for plugins to define their custom field requirements
* Con: changing custom fields requires DB migrations. This is good in that it makes you think about how the change will affect any existing data. Bad in the fact that if you don't plan carefully, you can lose existing data. It is also extra work.
* Con: Non-dev admins cannot set custom fields

### Dynamic custom fields (JSON-based/EAV)

* Pro: Admins can create & modify
* Pro: No migrations needed
* Con: not under source control, so harder to track changes
* Con: Would need to expose new APIs for plugins to define required custom fields, and somehow make it so certain fields _cannot_ be dynamically modified (otherwise the plugins will fail)

chladog:
> ### Dynamic custom fields (JSON-based/EAV)
> * Pro: Admins can create & modify
> * Pro: No migrations needed
> * Con: not under source control, so harder to track changes
> * Con: Would need to expose new APIs for plugins to define required custom fields, and somehow make it so certain fields _cannot_ be dynamically modified (otherwise the plugins will fail)

[...]

PR #4836, opened 2026-06-16
feat(core): Add custom field value resolver

sauraww:
## Description 

This PR lets custom fields return a computed value when they are read through GraphQL. 
The value stored in database is not changed. Vendure still stores and validates custom fields the same way as before. 
The new hook only changes what Admin API or shop API returns for that custom field. 

Relates to #490 

## What Changed 

- Added a read-time resolver for non-relation custom fields
- Passed vendure Injector , RequestContext and Vendure Entity to the resolver. 
- Kept the stored database value changed 
- Added to e2e coverage showing graphql can return computed value without overwriting the stored value. 

This is useful for plugins that need to keep Vendure as the source of stored product data, while resolving the value shown to API clients from another system or from request context. For example, a plugin can store a fallback custom field value in Vendure, but return a channel/currency/entity-specific value from an external configuration or pricing service when the field is queried.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4836"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784188025&installation_id=128408429&pr_number=4836&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4836&signature=47c4798bd7e721599359e21c4a56f6d061554ccf281bc15e218b246b7c7fb346"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/custom-fields.e2e-spec.ts
+++ b/packages/core/e2e/custom-fields.e2e-spec.ts
@@ -25,6 +25,7 @@ import { fixPostgresTimezone } from './utils/fix-pg-timezone';
 fixPostgresTimezone();
 
 const validateInjectorSpy = vi.fn();
+const resolveInjectorSpy = vi.fn();
 
 const customConfig = mergeConfig(testConfig(), {
     dbConnectionOptions: {
@@ -161,6 +162,16 @@ const customConfig = mergeConfig(testConfig(), {
                 name: 'stringList',
                 type: 'string',
                 list: true,
+                resolve: (value, injector, ctx, entity) => {
+                    const connection = injector.get(TransactionalConnection);
+                    resolveInjectorSpy({
+                        value,
+                        apiType: ctx.apiType,
+                        entityId: String(entity.id),
+                        connection,
+                    });
+                    return [`resolved:${String(entity.id)}`];
+                },
             },
             {
                 name: 'localeStringList',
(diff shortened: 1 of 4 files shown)


michaelbromley:
Hi,
Can you provide some documentation on this design and examples of how it would be used?

sauraww:
Thanks @michaelbromley yes, absolutely. 

The design intent is to add a read time hook for custom fields. Vendure would continue to own persistence, validation, writes, filtering and sorting exactly as it does today. The new `resolve` function only affects the value returned by GraphQL when the custom field is read.

A simple example would be a plugin-owned custom field whose stored value acts as a fallback:

```ts
customFields: {
  ProductVariant: [
    {
      name: 'bulkPrice',
      type: 'int',
      nullable: true,
      readonly: true,
      resolve: async (value, injector, ctx, variant) => {
        const service = injector.get(BulkPriceService);

        return service.resolve({
          fallback: value,
          channelCode: ctx.channel.code,
          currencyCode: ctx.currencyCode,
          variantId: variant.id,
        });
      },
    },
  ],
}
```

So a GraphQL query like:

```
query {
  productVariants {
    items {
      id
      customFields {
        bulkPrice
      }
    }
  }
}
```
would return the resolved value, while the database value remains unchanged.

The use case I am working on is a plugin which resolves custom field values from an external configuration service ( 
https://github.com/juspay/superposition)  based on request context, for example channel, currency, entity type and entity id. That lets Vendure keep the default/stored value, while plugins can provide context-specific values to API clients.

For example, a merchant might have a `bulkPrice` custom field on `ProductVariant`. The stored Vendure value could be the default bulk price, but a plugin could return a different value for:

- `channelCode = b2b`
- `currencyCode = USD`
- `entity = product_variant`
- `entityId = 42`

So the same `customFields.bulkPrice` query can return the normal stored value in the default context, but a B2B/USD-specific value for a particular product variant.

A plugin could also add customer-specific dimensions, such as customer group, by looking up the active customer from `RequestContext` and passing that into its own resolver.

michaelbromley:
Hi @sauraww 

Thanks for the clear explanation and example.

I think I’ll pass on this API extension right now, for the following reasons:

- The same functionality can be achieved by defining a custom field resolver directly on the entity in question, which can read the custom field value just like in your example
- Custom fields are already quite a complex feature to maintain and evolve, and I’m therefore extra reluctant to extend the API surface unless absolutely necessary.

If you think I missed something and this API is truly not replicable using the existing field resolvers mechanism, I’m happy to reconsider with more input.

PR #4839, opened 2026-06-17
fix(dashboard): Guard order list state cells against undefined state

### LINKED ISSUE #4830: Dashboard: enabling the State column on an order list table crashes the page (undefined.toLowerCase in OrderStateCell)
**Describe the bug**
In the React Dashboard, enabling the **State** column via the column settings dropdown on an order list table crashes the whole page with `TypeError: Cannot read properties of undefined (reading 'toLowerCase')`.

This affects any order list table whose initially visible columns do not include `state` — e.g. the **Latest Orders** widget on the Insights page (default visibility is `code`, `total`, `orderPlacedAt` only).

**To Reproduce**

1. Open the Dashboard → Insights page.
2. In the **Latest Orders** widget, open the column settings dropdown (Settings2 icon).
3. Enable the **State** column.
4. The page crashes with the error boundary showing `An error occurred: Cannot read properties of undefined (reading 'toLowerCase')`.

**Expected behavior**
The State column appears, showing the order state badge once the refetched data arrives.

**Actual behavior**
Page crashes. Root cause analysis:

1. `PaginatedListDataTable` trims the list query to only the visible columns (`includeOnlySelectedListFields`), so the initial fetch does not include `state`.
2. Enabling the column changes the query key (`visibleColumns` is part of it) and triggers a refetch — but the query uses `placeholderData: keepPreviousData`, so while the refetch is in flight, the table renders the **previous rows, which have no `state` field**, into the newly added column.
3. `OrderStateCell` passes the undefined value straight into `getTypeForState`:

```tsx
// packages/dashboard/src/lib/components/shared/table-cell/order-table-cell-components.tsx
export const OrderStateCell: DataTableCellComponent<{ state: string }> = ({ row }) => {
    const { getTranslatedOrderState } = useDynamicTranslations();
    const value = row.original.state;
    return <Badge variant={stateTypeToBadgeVariant(getTypeForState(value))}>{getTranslatedOrderState(value)}</Badge>;
};
```

```ts
// packages/dashboard/src/lib/utils/state-type.ts
export function getTypeForState(state: string): StateType {
    const stateLower = state.toLowerCase(); // ← throws when state is undefined
```

Note that `CustomerCell` in the same file already guards against this exact situation (`if (!value) return null;`) — `OrderStateCell` just misses the guard. Any other cell component that assumes its field is always present is susceptible to the same stale-data race.

**Error logs**

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at getTypeForState (state-type.ts)
    at OrderStateCell (order-table-cell-components.tsx)
    ... React render stack
```

- @vendure/core version: 3.6.3 (`OrderStateCell` on current `master` is unchanged)
- Nodejs version: 20
- Operating System (Windows/macOS/Linux): macOS / Linux

**Workaround**
`pnpm patch @vendure/dashboard` adding the same guard `CustomerCell` uses:

```tsx
export const OrderStateCell: DataTableCellComponent<{ state: string }> = ({ row }) => {
    const { getTranslatedOrderState } = useDynamicTranslations();
    const value = row.original.state;
    if (!value) {
        return null;
    }
    return <Badge variant={stateTypeToBadgeVariant(getTypeForState(value))}>{getTranslatedOrderState(value)}</Badge>;
};
```

Happy to send a PR if that helps.

grolmus:
## Summary
Enabling the **State** column on an order list crashed the whole page with `TypeError: Cannot read properties of undefined (reading 'toLowerCase')`. This guards the affected list cells so the column can be toggled safely.

Fixes #4830

## Root cause
`PaginatedListDataTable` trims the list query to only the visible columns (`includeOnlySelectedListFields`). Enabling **State** changes the query key and triggers a refetch, but with `placeholderData: keepPreviousData` the previous rows — which were fetched **without** `state` — render into the newly-added State column before the refetch returns. `OrderStateCell` passed that `undefined` straight into `getTypeForState(state)`, which calls `state.toLowerCase()` → crash, tripping the router error boundary.

`CustomerCell` in the same file already guards against exactly this transient-empty case (`if (!value) return null`); `OrderStateCell` just missed it.

## Change
- `order-table-cell-components.tsx` — `OrderStateCell` returns `null` when `state` is not yet present (matches the existing `CustomerCell` pattern).
- `customer-order-table.tsx` — the orders table on the customer detail page has the same `getTypeForState(value)` pattern (with an `as string` cast that hid the risk). Guarded it the same way and narrowed the cast to `string | undefined`.
- `getTypeForState` is left strict on purpose — an `undefined` state on a *detail* page is a real bug worth surfacing; the transient-empty contract belongs at the list-cell layer.

## Test plan
**Automated (e2e):** `packages/dashboard/e2e/tests/sales/orders.spec.ts` — *should not crash when toggling the State column off and on*. Hides then re-enables the State column and asserts the table keeps rendering (a crashed boundary unmounts the table). Verified it **fails on master** (table/State header gone after the crash) and **passes on this branch**.

**Manual:** before/after screenshots below (toggle State on the Orders list).

**Note:** the customer-detail orders table received the same guard for consistency, but I could not get a reliable fail-on-master e2e for it — on re-enable React Query appears to serve the cached initial (state-bearing) data rather than the state-less placeholder, so the crash doesn't reproduce there in the tested scenario. Shipped as defensive consistency; the shared mechanism is covered by the orders-list regression.

## Screenshots
| Before (master) | After (this PR) |
| --- | --- |
| _error boundary: "Cannot read properties of undefined (reading 'toLowerCase')"_ | _State column renders the order-state badge_ |

_(images attached below)_

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4839"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784297550&installation_id=128408429&pr_number=4839&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4839&signature=c0d7f397b8d0827ce53881e1715e8d5b8627124d7da3d5f479b77f7b946b3b6f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/sales/orders.spec.ts
+++ b/packages/dashboard/e2e/tests/sales/orders.spec.ts
@@ -100,6 +100,46 @@ test.describe('Orders', () => {
         await lp.expectRowCountGreaterThan(0);
     });
 
+    // #4830 — toggling the State column must not crash the order list.
+    // The list query only fetches visible columns; re-enabling State refetches,
+    // but `placeholderData: keepPreviousData` first renders the previous rows
+    // (which no longer carry `state`) into the State column. Pre-fix this threw
+    // `Cannot read properties of undefined (reading 'toLowerCase')` in
+    // OrderStateCell and tripped the router error boundary.
+    test('should not crash when toggling the State column off and on', async ({ page }) => {
+        const client = new VendureAdminClient(page);
+        await client.login();
+        await createPaidOrder(client);
+
+        const lp = listPage(page);
+        await lp.goto();
+        await lp.expectLoaded();
+        await lp.expectRowCountGreaterThan(0);
+
+        const stateHeader = lp.dataTable.locator('thead th').filter({ hasText: 'State' });
+        const stateToggle = page.getByRole('menuitemcheckbox', { name: /^state$/i });
+        await expect(stateHeader).toBeVisible();
+
+        // Hide State → the list refetches without the `state` field.
+        await lp.openColumnSettings();
+        await stateToggle.click();
+        await page.keyboard.press('Escape');
+        await expect(stateHeader).toBeHidden();
+
+        // Re-enable State → triggers the stale-placeholder render that used to crash.
+        await lp.openColumnSettings();
+        await stateToggle.click();
+        await page.keyboard.press('Escape');
+
+        // No crash: a crashed boundary unmounts the table, so the header + rows
+        // staying visible after the refetch is the real signal. The error-text
+        // assertion is belt-and-suspenders.
+        await expect(stateHeader).toBeVisible();
+        await lp.expectRowCountGreaterThan(0);
+        await expect(lp.getRows().first()).toBeVisible();
+        await expect(page.getByText(/An error occurred:/i)).toHaveCount(0);
+    });
+
     test('should create and delete a draft order', async ({ page }) => {
         // Create a new draft
         const lp = listPage(page);
--- a/packages/dashboard/src/app/routes/_authenticated/_customers/components/customer-order-table.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_customers/components/customer-order-table.tsx
@@ -55,7 +55,10 @@ export function CustomerOrderTable({ customerId }: Readonly<CustomerOrderTablePr
                 state: {
                     header: 'State',
                     cell: ({ cell }) => {
-                        const value = cell.getValue() as string;
+                        const value = cell.getValue() as string | undefined;
+                        if (!value) {
+                            return null;
+                        }
                         return <Badge variant={stateTypeToBadgeVariant(getTypeForState(value))}>{value}</Badge>;
                     },
                 },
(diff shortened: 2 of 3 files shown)


grolmus:
before
after

also added tests

dlhck (approved):

PR #4843, opened 2026-06-18
fix(dashboard): Support enum fields in list filters and form defaults

### LINKED ISSUE #4748: Dashboard: Enum fields on custom entities are not filterable in ListPage and wrong default value in detail forms
**Describe the bug**
When adding a GraphQL enum field to a custom entity (e.g. `status: StatusType!`),
two separate bugs occur in the Vendure Dashboard (v3.6.2):

1. **The field does not appear in the "Add filter" dropdown** of a `<ListPage>`.
2. **The default value for the field is `{}` (empty object) or `""` (empty string)**
   instead of the first enum value when opening a new detail form.

**To Reproduce**
1. Define a custom entity with an enum column:

   ```graphql
   enum StatusType { active inactive }

   type Employer implements Node {
     id: ID!
     status: StatusType!
     ...
   }
   ```

2. Create a ListPage querying this entity — status is not available in the
   "Add filter" dropdown.
3. Open the detail page for a new entity — the status field defaults to {}
   instead of "active".

**Expected behavior**
1. Enum fields should appear in the "Add filter" dropdown of a ListPage, , ideally rendered as a dropdown/select with the known enum values.
2. getDefaultValueFromField should return the first enum value (e.g. "active") as the default for enum fields, not {} or "".

**Actual behavior**
1. Enum columns are silently excluded from the filter menu. 
2. When opening a "new entity" form, an enum field defaults to {} (nullable) or "" (non-nullable), causing Zod validation errors or a visually empty select with no value selected.

**Screenshots/Videos**
Bug 1
Image

Bug 2
Image

@vendure/core version: 3.6.2
@vendure/dashboard version: 3.6.2
Node.js version: v24.13.0
Database: PostgreSQL
Operating System: macOS
Browser: Chrome
Package manager: npm

**Configuration**

```typescript
// custom entity with enum field
@Column(enumColumn({ enum: StatusType, nullable: false, default: StatusType.active }))
status: StatusType;

# admin-api-extensions.ts
enum StatusType {
    active
    inactive
}

type Employer implements Node {
    id: ID!
    status: StatusType!
    ...
}
```

**Workaround**
Bug 1: Use a facetedFilter instead of the built-in column filter — this completely bypasses enableColumnFilter and allows custom filter UI.

Bug 2: -

**Root Cause**
**Bug 1** - packages/dashboard/src/lib/components/data-table/use-generated-columns.tsx:
```ts
const enableColumnFilter = fieldInfo.isScalar && !facetedFilters?.[fieldInfo.name];
```

fieldInfo.isScalar is computed in get-document-structure.ts as:

```ts
const isScalar = isScalarType(type); // = schemaInfo.scalars.includes(type)
```

schemaInfo.scalars contains only true GraphQL scalars (String, Int, Boolean, ID) Enum types like StatusType live in enums, so isScalar is always false for enums -> enableColumnFilter is false -> the column never appears in the filter menu.

Suggested Fix:
```ts
import { isEnumType } from '@/vdb/framework/document-introspection/get-document-structure.js';

const enableColumnFilter =
    (fieldInfo.isScalar || isEnumType(fieldInfo.type)) &&
    !facetedFilters?.[fieldInfo.name];
```

**Bug 2** - packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts -> getDefaultValuesFromField:
isEnumtype is already imported in this file but not called inside getDefaultValuesFromField. For a nullable enumerator field the function falls into the default branch:
```ts
default:
    if (!field.isScalar || field.type === 'JSON') {
        return {};  // enum hits this because !isScalar === true
    }
    return null;
```

For a non-nullable enum field it falls to:
```ts
default:
    return ''; // empty string
```

Suggested fix:
```ts
// nullable path — add before the object check:
if (isEnumType(field.type)) {
    return schemaInfo.enums[field.type]?.[0] ?? null;
}

// non-nullable path — add before the fallthrough return:
if (isEnumType(field.type)) {
    return schemaInfo.enums[field.type]?.[0] ?? '';
}
```

grolmus:
## Summary
GraphQL enum fields on custom entities were unusable in the dashboard in two ways. This makes them filterable in list pages and gives them a valid default in detail forms.

Fixes #4748

### Bug 1 — enum columns not filterable
`use-generated-columns.tsx` gated `enableColumnFilter` on `fieldInfo.isScalar`; enums aren't scalars, so enum columns never appeared in the **Add filter** menu. Forcing it on wasn't enough either — `data-table-filter-dialog.tsx` only had inputs for String/Int/Float/DateTime/Boolean/ID/Money, so an enum rendered an **empty dialog**.
- Enable the filter for enum columns (`isScalar || isEnumType(type)`).
- Add `filters/data-table-enum-filter.tsx` (operator select `eq`/`notEq`/`isNull` + a value select of the enum members) and wire it into the dialog via a new `getEnumValues(type)` helper.
- Core maps **every** enum column to `StringOperators` (`generate-list-options.ts:175`), so `eq` is valid server-side for both built-in and custom-entity enum fields.

### Bug 2 — wrong default value for enum fields
`getDefaultValueFromField` returned `{}` (nullable) or `''` (non-nullable) for enums, leaving the select unset / invalid.
- Non-nullable enum → first enum member.
- Nullable enum → `null` (valid; empty select).

## Test plan
- **Unit** (`form-schema-tools.spec.ts`): non-nullable enum defaults to its first value, nullable enum defaults to `null`. Fails on master ({}/''), passes here.
- **e2e** (`sales/orders.spec.ts`): opens Add filter on the orders list, asserts the enum `type` (`OrderType`) column is offered, then **applies** `type eq Regular` and asserts the list still returns the order (verifies the enum filter works server-side, not just renders). Fails on master (column absent), passes here.
- Dashboard typecheck passes.
- Manual before/after screenshots below.

## Out of scope / follow-ups
- Enum **sorting** stays disabled (`enableSorting` is `isScalar`-gated) — not changed.
- `in`/`notIn` (multi-value) enum filtering: `StringOperators` supports it; could be added later with a multi-select. v1 ships `eq`/`notEq`/`isNull`.

## Screenshots
| Before (master) | After (this PR) |
| --- | --- |
| Add filter menu has no "Type" entry | "Filter by type" renders a working enum filter (is equal to / Regular) |

_(images attached below)_

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4843"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784363417&installation_id=128408429&pr_number=4843&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4843&signature=3c3ed18543a78b70280e75d30e1a7921d066a35a740bea613104139959090370"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
after-dialog
after-menu
before

michaelbromley (approved):

PR #4848, opened 2026-06-18
fix(dashboard): Stabilise useUiLanguageLoader callback to prevent render loop

### LINKED ISSUE #4847: useUiLanguageLoader returns an unstable loadAndActivateLocale, causing an infinite render loop when used in an effect
### Describe the bug

`useUiLanguageLoader()` returns a new `loadAndActivateLocale` function instance on every render. That function calls `i18n.activate()`, which re-renders every `useLingui` consumer. If a Dashboard extension calls `loadAndActivateLocale` from a `useEffect` — a natural way to enforce or restore a display locale — and lists the function in the dependency array (as the `react-hooks/exhaustive-deps` rule requires), the effect re-runs on every render and the component never settles:

`render → effect runs → loadAndActivateLocale() → i18n.activate() → re-render → new function reference → effect runs again → …`

Because activation re-renders Base UI consumers, the loop also continuously resets transient UI state. In practice this can make **row checkboxes and the per-row "⋮" action menus unusable across the whole dashboard**, since their open/selection state is reset on every frame.

### Root cause

`packages/dashboard/src/lib/hooks/use-ui-language-loader.ts` (on `master`):

```ts
export function useUiLanguageLoader() {
    const { i18n } = useLingui();

    async function loadAndActivateLocale(locale: string) {
        if (currentlyLoading === locale) {
            return;
        }
        currentlyLoading = locale;
        const messages = await loadI18nMessages(locale);
        i18n.load(locale, messages);
        i18n.activate(locale);
        currentlyLoading = null;
    }

    return { loadAndActivateLocale };
}
```

Two properties combine:

1. `loadAndActivateLocale` is declared inline, so a **new reference** is returned on every render. Any consumer that depends on its identity (effect deps, `useMemo`/`useCallback` deps, `React.memo` children) re-runs on every render.
2. The module-level `currentlyLoading` guard only blocks **concurrent** loads of the same locale; it is reset to `null` once the load completes, so it does not prevent **re-activation** on the next render. And `@lingui/core`'s `i18n.activate()` emits a `change` event unconditionally (even for the already-active locale), which `I18nProvider` turns into a `setContext(...)` — re-rendering the subtree. Each completed activation therefore triggers another render, which triggers the effect again.

The built-in caller (`language-dialog.tsx`) invokes it from a click handler, so the core UI never hits the loop — but the exported hook is part of the dashboard extension API surface, and calling it from an effect is a reasonable pattern that should be safe.

### To reproduce

In a Dashboard extension component:

```tsx
import { useUiLanguageLoader, useUserSettings } from '@vendure/dashboard';
import { useEffect } from 'react';

export function LocaleEnforcer() {
    const { loadAndActivateLocale } = useUiLanguageLoader();
    const { settings } = useUserSettings();

    useEffect(() => {
        void loadAndActivateLocale(settings.displayLanguage || 'en');
    }, [loadAndActivateLocale, settings.displayLanguage]); // exhaustive-deps requires the function

    return null;
}
```

Mount it (e.g. as an `afterForm`/toolbar component). The component re-renders without settling; React DevTools shows a continuous render loop, and across the dashboard the row-selection checkboxes and "⋮" menus cannot be opened because their state is reset each frame.

### Expected behaviour

Calling `loadAndActivateLocale` from an effect should activate the locale once and then settle. The returned function should have a stable identity across renders unless its real dependencies change.

### Suggested fix

Memoise the returned callback with `useCallback` (keyed on `[i18n]`) so its identity is stable, and skip re-activation when the requested locale is already active. This is low-risk: the existing click-handler usage is unchanged.

I've confirmed the mechanism and the fix with a minimal reproduction on the real Lingui runtime (faithful copy of the hook + `@lingui/core`/`@lingui/react` + React 19): the original returns a new callback each render and loops when activated from an effect, while the memoised version keeps a stable reference and settles after one activation.

### Environment

- `@vendure/dashboard`: 3.6.4 (current); source unchanged on `master`
- React 19, `@lingui/react` v5

I'd like to fix this — PR incoming.

TheStreamCode:
# Description

`useUiLanguageLoader` returned a new `loadAndActivateLocale` function on every render. Because that function calls `i18n.activate()` — which `@lingui`'s `I18nProvider` turns into a re-render of every `useLingui` consumer — using it from a `useEffect` (as `react-hooks/exhaustive-deps` requires) makes the effect re-run on every render and compounds into an infinite render loop. The loop continuously resets Base UI popover / data-table selection state, which can make row checkboxes and the per-row "⋮" action menus unusable across the dashboard.

This memoises the callback with `useCallback` (keyed on `i18n`) so its identity is stable across renders, and skips re-activation when the requested locale is already active. The existing click-handler caller (`language-dialog`) is unaffected.

Fixes #4847.

### Verification

I confirmed the mechanism and the fix with a minimal reproduction on the real Lingui runtime (a faithful copy of the hook + `@lingui/core`/`@lingui/react` + React 19):

- the original hook returns a new `loadAndActivateLocale` reference on every render; an effect depending on it re-runs on every unrelated re-render; activating a locale from such an effect loops indefinitely (`i18n.activate()` emits a `change` event unconditionally, so `I18nProvider` re-renders even for the already-active locale);
- the memoised version returns a stable reference, the effect does not re-run on unrelated re-renders, and activating a locale from an effect settles after a single activation.

Note: the `@vendure/dashboard` unit tests don't currently include a React renderer (`@testing-library/react`), so I've kept this PR to the fix itself rather than pulling in a new test dependency. Happy to add a regression test in your preferred style (e.g. `@vitest/browser`) if you'd like.

# Breaking changes

None.

# Screenshots

n/a

# Checklist

📌 Always:

👍 Most of the time:
- [ ] I have added or updated test cases (see note above — no React test harness in the package yet)
- [x] I have updated the README if needed (no change needed)

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4848"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784400489&installation_id=128408429&pr_number=4848&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4848&signature=10222140f8f68122755533fa02b14b7e366e3818e6f77674b209af7af8a2dcf8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


TheStreamCode:

TheStreamCode:
recheck

PR #4849, opened 2026-06-19
fix(dashboard): Leave entity detail page on channel switch

### LINKED ISSUE #4826: fix(dashboard): Stale navigation history retains cross-channel entity links after channel switch

grolmus:
## Summary
In a multi-channel setup, switching the active channel while viewing a channel-scoped entity detail (order, product, …) left the previous channel's entity on screen, and its breadcrumb/nav links 404'd when clicked. This navigates back to the dashboard home when you switch channel away from an entity-detail page.

Fixes #4826

## Root cause
On channel switch, `channel-provider.tsx`'s `setSelectedChannel` calls `queryClient.invalidateQueries()` (refetches React Query data — so the detail form empties), but the **TanStack Router loaders are not re-run**. The breadcrumb is derived from route `loaderData.breadcrumb`, so the stale entity name (and its clickable, now cross-channel link) persists.

## Change
In `InnerApp` (`main.tsx` — the one place that holds both the router instance and a `useChannel()` consumer, since `ChannelProvider` sits above `RouterProvider`), an effect watches the active-channel token and, on an actual change, navigates to home **if the current route is an entity-detail/edit route** (any matched route carrying a dynamic path param). List/settings routes are channel-agnostic and left in place (they just refetch via the provider's invalidation). A ref guards the initial mount.

Why navigate rather than `router.invalidate()`: I tried `router.invalidate()` first and verified it doesn't fix the visible state (re-running the detail loader for an entity missing in the new channel just hits its "not found" throw → error page).

## Test plan
`e2e/tests/regression/issue-4826-channel-switch-nav.spec.ts` (2-channel setup):
- detail → switching channel navigates to home (asserted via URL) and the stale breadcrumb is gone. **Fails on master, passes on this branch.**
- list page → switching channel does **not** redirect (stays on `/products`), locking the scoping.

Dashboard typecheck passes. Before/after screenshots attached.

## ⚠️ Open question for @michaelbromley — which behaviour do you prefer?
We deliberately took the **simple** route: on channel switch from any entity-detail page → redirect to **Insights** (home). It's robust and needs no extra fetch.

The trade-offs / alternative:
- **Over-redirect (accepted):** it redirects from a detail page even when that entity *does* exist in the new channel (e.g. a product assigned to both channels), and also from **global, non-channel entities** (Administrators, Roles, Countries, Tax Rates, etc.) where staying would be fine. All harmless (you land on home), just more aggressive.
- **Alternative — membership-aware:** only redirect when the entity isn't in the new channel (stay otherwise). More correct UX, but needs to fetch/verify the entity in the new channel and work around the detail loader's not-found throw.

If you'd prefer the membership-aware approach (or scoping it differently), let us know and we'll rework it. Otherwise this simple version closes the reported bug.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4849"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784452026&installation_id=128408429&pr_number=4849&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4849&signature=e9dbe6ba0f39717811b82b2e7f89b0e2bee185f0d3366f4c21667adf68f98fbf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/regression/issue-4826-channel-switch-nav.spec.ts
+++ b/packages/dashboard/e2e/tests/regression/issue-4826-channel-switch-nav.spec.ts
@@ -59,7 +59,7 @@ async function switchToSecondChannel(page: Page) {
     await page.getByRole('menuitem').filter({ hasText: 'second-channel' }).click();
 }
 
-test('returns to dashboard home when switching channel from an entity detail', async ({ page }) => {
+test('returns to the entity list view when switching channel from an entity detail', async ({ page }) => {
     test.setTimeout(120_000);
     const client = new VendureAdminClient(page);
     await client.login();
@@ -74,8 +74,8 @@ test('returns to dashboard home when switching channel from an entity detail', a
 
     await switchToSecondChannel(page);
 
-    // Primary signal: navigated to the (channel-neutral) dashboard home.
-    await page.waitForURL(url => new URL(url).pathname === '/', { timeout: 10_000 });
+    // Primary signal: navigated to the product list view (not all the way to home).
+    await page.waitForURL(url => new URL(url).pathname === '/products', { timeout: 10_000 });
     // The stale product breadcrumb is gone.
     await expect(breadcrumb.getByText(product.name, { exact: false })).toHaveCount(0);
 });
--- a/packages/dashboard/e2e/tests/regression/issue-4826-channel-switch-nav.spec.ts
+++ b/packages/dashboard/e2e/tests/regression/issue-4826-channel-switch-nav.spec.ts
@@ -1,0 +1,82 @@
+import { expect, type Page, test } from '@playwright/test';
+
+import { VendureAdminClient } from '../../utils/vendure-admin-client.js';
+
+// #4826 — switching channel while viewing a channel-scoped entity must not leave
+// the stale entity (and its breadcrumb/nav links, which 404 when clicked) on
+// screen. The dashboard returns to home when leaving an entity detail page, but
+// leaves channel-agnostic list pages in place.
+
+async function ensureSecondChannel(client: VendureAdminClient) {
+    const { channels } = await client.gql(`query { channels { items { id code } } }`);
+    if (channels.items.some((c: any) => c.code === 'second-channel')) {
+        return;
+    }
+    const { zones } = await client.gql(`query { zones { items { id } } }`);
+    const zoneId = zones.items[0].id;
+    try {
+        await client.gql(
+            `mutation ($input: CreateChannelInput!) {
+                createChannel(input: $input) {
+                    ... on Channel { id }
+                    ... on ErrorResult { errorCode message }
+                }
+            }`,
+            {
+                input: {
+                    code: 'second-channel',
+                    token: 'second-channel-token',
+                    defaultLanguageCode: 'en',
+                    currencyCode: 'USD',
+                    pricesIncludeTax: false,
+                    defaultTaxZoneId: zoneId,
+                    defaultShippingZoneId: zoneId,
+                },
+            },
+        );
+    } catch {
+        // Already created by a prior run/test — fine, it exists either way.
+    }
+}
+
+async function switchToSecondChannel(page: Page) {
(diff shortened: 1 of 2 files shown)


grolmus:
@michaelbromley heads-up on a design decision in this one 👇

We fixed #4826 the **simple** way: on channel switch, if you're on an entity-detail page, redirect to **Insights** (home). This reliably clears the stale, now cross-channel entity + its 404-ing breadcrumb links.

Trade-off: it also redirects when the entity *does* exist in the new channel, and from global (non-channel) entity details (Administrators, Roles, Countries, …) — all harmless (you just land on home), but more aggressive than strictly necessary.

The alternative is **membership-aware** (only redirect when the entity isn't in the new channel, stay otherwise) — nicer UX but needs to fetch/verify the entity in the new channel and work around the detail loader's not-found throw.

If you'd rather we go membership-aware (or scope it differently), say the word and we'll rework it. Details + trade-offs in the PR description.

grolmus:
before on product + change channel
before

after change channel redirect
after

michaelbromley:
I think a better alternative would be to navigate to the list view, rather than going all the way back up to the insights page. Can you try that?

grolmus:
Good call — done in 2cfbf89. On channel switch from an entity detail page it now navigates to that entity's **list view** (the static path prefix before the first dynamic segment, e.g. `/products/$id` → `/products`, `/facets/$facetId/values/$id` → `/facets`) instead of going to the dashboard home. Updated the regression to assert it lands on `/products`; both channel-switch e2e tests pass and typecheck is green.

michaelbromley:
I just realised that this is yet another duplicate of an existing earlier PR:

- https://github.com/vendurehq/vendure/pull/4827

Therefore I will close this one.

PR #4850, opened 2026-06-19
feat(create): Add AI-assisted project guidance

dlhck:
## Summary

- Add `@vendure/eslint-plugin` with Vendure-specific plugin rules and tests.
- Update `@vendure/create` to generate `AGENTS.md`, ESLint or Biome setup, lint scripts, and selected lint dependencies.
- Add Biome GritQL rule templates for the supported syntactic rule subset.
- Add an AI-assisted development guide and cross-link it from CLI and plugin docs.

## Validation

- `npm run test` in `packages/eslint-plugin`
- `npm run build` in `packages/eslint-plugin`
- ESLint 9 flat-config smoke test
- `npm run build` in `packages/create`
- `npm run test` in `packages/create`
- Template render smoke test
- `npm run test:mdx` in `docs`

Note: `packages/create` tests and docs MDX check needed to be rerun outside the sandbox because local port binding / TSX IPC were blocked with `EPERM`.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4850"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784456172&installation_id=128408429&pr_number=4850&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4850&signature=65d8ae7e0658058fa8f5681c54f1996062709588739870de25cf850729c24129"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/create/templates/agents.hbs
```diff
@@ -0,0 +1,53 @@
...
+{{#if isMonorepo}}
+## Workspace Layout
+
+- Vendure backend: `apps/server`
+- Next.js storefront: `apps/storefront`
+- Start both apps: `npm run dev`
+- Start only the server: `npm run dev:server`
+- Start only the storefront: `npm run dev:storefront`
+- Backend custom code belongs in `apps/server/src/plugins`
+- Backend runtime configuration is in `apps/server/src/vendure-config.ts`
```
michaelbromley, line 14: the `static` dir explanation is missing from the monorepo mode

michaelbromley (changes requested):
Hi, I would prefer to narrow the scope of this PR because I think it is bunching together 2 distinct features:

- agents.md with new projects
- linting plugin / configs

Including an agents.md is pretty straightforward I think, so that's quite an easy one to review and approve.

The linting one is more involved - we are talking about adding 88 new dependencies for the eslint version for instance plus a new package maintain. So I need to apply a different level of scrutiny. I do like the concept but I think the current implementation needs some work, based on some testing I did on a few real-world Vendure projects locally. 

The eslint rules have a number of issues with both false positives and false negatives, to the degree that in their current form they might prove a net-negative. But I think these can all be fixed with tighter definitions and ideally a bunch of tests for the many different cases that might come up. I'll add full details with examples in a follow-up comment.

The Biome rules also need significant tightening to bring them even to the standard of the current ESLint rules. For example, the rule about raw ctx in a queue. The current form will fail to catch the following:

```
reviewQueue.add({ ctx })
```

and will false-positive on:
```
mySet.add(ctx)
this.subscriptions.add(ctx)
```

I made a new version to correctly handle them and bring it more into line with the ESLint version:

```
pattern raw_ctx_member() {
    or {
        `ctx`,
        `$k: $v` where { $v <: raw_ctx_value() }   // recurse into the VALUE slot only
    }
}

pattern raw_ctx_value() {
    or {
        `ctx`,
        `{ $props }` where { $props <: some(raw_ctx_member()) }
    }
}

`$queue.add($arg)` as $call where {
    $queue <: r".*[Qq]ueue.*",
    $arg <: raw_ctx_value(),
    register_diagnostic(span = $call, message = "raw ctx in job data")
}
```

so I think with some effort, the GritQL queries can perform similarly to the ESLint rules, but I also suspect that the lack of real meta-programming (e.g. storing references in a map like some of the ESLint rules) will mean they will always be a bit more limited.

michaelbromley:
Some examples of issues with current rules based on testing on real code bases plus analysis. This can form the basis of a test suite so that we can make the linting really solid.

---

### B2 — `no-synchronize-true` fires on any object property · Medium · ✅
The rule flags any `Property` named `synchronize` with value `true`, anywhere in matched files, with
no check that it sits in a TypeORM/`VendureConfig` object.

```ts
const wsOpts = { synchronize: true, debounce: 200 }; // unrelated → still flagged ✅ reproduced
```

---

### B4 — `require-plugin-compatibility` silent FN on non-literal args + FP on spread · High · ✅
`require-plugin-compatibility.ts` bails if `metadata.type !== 'ObjectExpression'`, and `getObjectProperty`
only matches a literal `Property` named `compatibility`.

```ts
const OPTS = { imports: [] };
@VendurePlugin(OPTS)            export class P1 {}   // ❌ no error — silent false negative ✅ reproduced
@VendurePlugin({ ...base })     export class P2 {}   // ⚠️ ALWAYS errors even if base has compatibility ✅ reproduced
```

Silently passing is worse than a false positive — the rule's entire purpose is to guarantee the
field is present. It should report when it *cannot verify* the argument.

---

### B6 — `no-static-plugin-options-outside-plugin` blind to cross-file usage · High · 🔍 (architecturally certain)
Plugin classes are tracked in a per-file `Map` populated by `ClassDeclaration` visits. When the
plugin is imported into another file and `MyPlugin.options` is read there, the declaration visitor
never fires in that file, the `.has()` guard fails, and the violation is missed — i.e. the **primary
real-world form** of the anti-pattern is invisible. A non-type-aware rule cannot see cross-file
references; needs `parserServices` (type info) or a name-based heuristic (flag `*Plugin.options`).

---

### B11 — `require-job-queue-lifecycle-registration` FP on delegated helpers · High · ✅
`getEnclosingMethodName` returns the **immediate** enclosing method only. Queues created in a helper
method called from `onModuleInit()` are flagged:

```ts
async onModuleInit()  { await this.initQueues(); }
private async initQueues() {
  this.q = await this.jobQueueService.createQueue({ … }); // ❌ false positive ✅ reproduced (fredhopper-plugin)
}
```

---

### B13 — `no-raw-request-context-in-job-data` hardcodes the name `ctx` · Medium · ✅
`containsRawCtx` checks `node.name === 'ctx'` only — both the bare arg and object *values*.

```ts
function f(requestCtx: RequestContext) {
  this.myQueue.add(requestCtx);          // ❌ missed (arg named requestCtx) ✅ reproduced
  this.myQueue.add({ ctx: requestCtx }); // ❌ missed (value is requestCtx)  ✅ reproduced
}
```

---

### C2 — Biome `no-process-env-in-plugin` misses bracket access + destructuring · High · ✅
Pattern: `` `process.env.$name` ``.

```ts
process.env.API_KEY;            // ✅ fires
process['env'].SECRET;          // ❌ missed (bracket)        ✅ reproduced
const { TOKEN } = process.env;  // ❌ missed (destructuring)  ✅ reproduced
```

The ESLint version catches all three (its `isMemberExpression`/`isPropertyNamed` handle `Literal`
keys, and `process.env` is still a `MemberExpression` in destructuring). Destructuring env in config
code is a common real-world form.

---

### C3 — Biome `no-synchronize-true` misses quoted key · Medium · ✅
Pattern: `` `synchronize: true` ``.

```ts
{ synchronize: true }      // ✅ fires
{ "synchronize": true }    // ❌ missed (string-literal key)  ✅ reproduced
```

The ESLint version handles `Literal` keys via `isPropertyNamed`.

---

michaelbromley (approved):

PR #4857, opened 2026-06-22
fix(core): Export and register Province entity and ProvinceService

### LINKED ISSUE #4792: Province entity and ProvinceService not found
**Describe the bug**
In Vendure document, there are [Province Entity](https://docs.vendure.io/current/core/reference/typescript-api/entities/province) and [ProvinceService](https://docs.vendure.io/current/core/reference/typescript-api/services/province-service) but not available in @vendure/core

**To Reproduce**
In a vendure project If you do
```typescript
import { ProvinceService, Province } from '@vendure/core'
```
You will get error 
```terminal
"@vendure/core"' has no exported member named 'ProvinceService'.
"@vendure/core"' has no exported member named 'Province'.
```

**Expected behavior**
Both ProvinceService and Province are exported from @vendure/core

**Actual behavior**
You will get error 
```terminal
"@vendure/core"' has no exported member named 'ProvinceService'.
"@vendure/core"' has no exported member named 'Province'.
```

- @vendure/core version: 3.6.3
- Nodejs version: v24.10.0

**Workaround**
If you would like to create a zone with province member: You can
- create Region with type `province` and point a parent in Country entity, and use script to add the province as a member into a Zone.

**Additional context**
- As I was trying to setup Tax Zone for US, which state (province) tax play important role on the setup. I stumpped into this issue.
- Currently, Zone (tax zone) can only be configured (via dashboard) with members from Country. However, in original Vendure setup, it should be able to configure with any thing in Region entity, or say, Province (Both Province and Country extends the Region entity)
- To clear the issue, it might be as easy as to add the following to the package
```typescript
// in `packages/core/src/entity/index.ts`
export * from './region/province.entity';
```
and
```typescript
// in packages/core/src/service/index.ts
export * from './services/province.service.ts'
```
- However, it would also be nice to have:
  - In dashboard -> Settings
     - Additional `Province` setting that allow user to create province through dashboard, and
     - In `Zone` setting, one can define member from Province (instead of just from Country)
  - In @vendure/create: user can populate Province with json file, similar to populate Country for initial dataset.

grolmus:
## Summary

`Province` (entity) and `ProvinceService` are documented as public API (docs.vendure.io has pages for both) but importing them fails:

```ts
import { Province, ProvinceService } from '@vendure/core';
// '@vendure/core' has no exported member named 'ProvinceService' / 'Province'
```

Fixes #4792

## Root cause

`Province` (a `@ChildEntity` of `Region`, single-table inheritance) and `ProvinceService` (a complete service: `findAll`/`findOne`/`create`/`update`/`delete`, already publishing `ProvinceEvent`) both exist, but:

- `entity/index.ts` exported `region/country.entity` but **not** `region/province.entity`.
- `service/index.ts` exported `country.service` but **not** `province.service`.
- `event-bus/index.ts` exported `country-event` but **not** `province-event`.
- `ProvinceService` was **not** in `service.module.ts`'s `services` array, so it was never registered as a provider — even if imported, it could not be injected.

In a built `@vendure/core`, `Country`/`CountryService` resolve to functions while `Province`/`ProvinceService` were `undefined`. This is an oversight from the original "Regions & Provinces" work (the Country counterparts were all wired up; the Province ones were missed), not an intentional private API.

## The change (mirrors the existing `Country` wiring)

- `entity/index.ts` — export `region/province.entity`
- `service/index.ts` — export `services/province.service`
- `event-bus/index.ts` — export `events/province-event` (so consumers injecting `ProvinceService` can subscribe to the events it emits)
- `service.module.ts` — import `ProvinceService` and add it to the `services` array (which feeds both `providers` and `exports`), making it injectable

`Province` uses single-table inheritance, so no `dbConnectionOptions.entities` registration is needed.

## Test plan

New `packages/core/e2e/province.e2e-spec.ts`:

- Asserts `Province`, `ProvinceService` and `ProvinceEvent` are exported from `@vendure/core`.
- Asserts `server.app.get(ProvinceService)` is injectable (DI registration).
- Exercises the full CRUD round-trip through the service (`create` → `findOne` → `findAll` → `delete`) and asserts the emitted `ProvinceEvent`s are `['created', 'deleted']`.

Verified: **fails on `master`** (`Nest could not find given element` — `ProvinceService` isn't a provider; the imports resolve to `undefined`), **passes on this branch**. Sibling `country.e2e` (6 tests) still green; core `tsc` build + ESLint clean.

## Out of scope (follow-up)

The Admin API schema (`province.api.graphql`) declares `provinces` / `province` / `createProvince` / `updateProvince` / `deleteProvince`, but there is **no `ProvinceResolver`** registered (only `CountryResolver`) — those GraphQL operations have never had a backing resolver. That is a pre-existing gap from the same original commit and is independent of this programmatic-export fix; it deserves its own issue. This PR restores the documented **programmatic** (`@vendure/core`) surface only.

Note: docs generation walks source `@docsCategory` annotations, not the index barrels, which is why `ProvinceService` had a published doc page despite being unexported — so no docs change is required.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4857"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784725373&installation_id=128408429&pr_number=4857&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4857&signature=a8ac27bdf231bb0bc9d7a58dde7f4740350b6fad2962231756804ffeae453ba7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


michaelbromley (dismissed):
Hi Martin, 
Thanks for this fix - adding the exports is probably all that is needed here because the e2e tests are not "true" e2e tests in the sense that all the other e2e tests are making Graphql API calls and asserting against the response, in general.

We _could_ have an e2e suite for Province, but it should work via API not direct service instance injection. I'm not even sure if we have any exposed API surface for this however, as Province as a concept was created with future work in mind (potentially supporting nested regions) but we thus far never really did anything meaningful with it.

So for now I'd say you can remove those e2e tests (not something I often say! 😅)

Unless of course we _are_ exposing a graphql interface for Province, in which case adapt the tests to go via the graphql API.

grolmus:
Thanks Michael! Makes sense — there's no GraphQL surface for Province (no `ProvinceResolver`), so I've removed the service-injection e2e tests rather than adapt them. The PR now just adds the missing exports and the `ProvinceService` provider registration, which is the actual fix. Re-requesting review.

biggamesmallworld (approved):
🏄‍♂️

Issue #4048, opened 2025-12-11
Allow configuration of `tanstackRouter.tmpDir` via `vendureDashboardPlugin` config.


bummzack:
**Is your feature request related to a problem? Please describe.**

The deployment platform I use has the `node_modules` folder on a different device than the checked out code. This causes a problem when trying to build the new Dashboard, resulting in an error like: 

> Error: EXDEV: cross-device link not permitted, rename …

I tracked it down to the `tanstackRouter` vite plugin, that uses `fs.rename` instead of actually moving a file. They also have a related issue: https://github.com/TanStack/router/issues/4360

The problem can be fixed quite easily, by explicitly setting the `tmpDir` option for the `tanstackRouter` plugin.
Adding the following to https://github.com/vendure-ecommerce/vendure/blob/master/packages/dashboard/vite/vite-plugin-vendure-dashboard.ts#L181C17-L186C20 solves the problem:

```
tmpDir: path.join(packageRoot, '.tmp'),
```

**Describe the solution you'd like**

It would be nice if there was a way to configure the `tanstackRouter` plugin `tmpDir` option, maybe via `vendureDashboardPlugin` options?

**Describe alternatives you've considered**

Currently I'm using `patch-package` to inject this change… but that's not really a good solution for long-term maintainability.

michaelbromley:
proper solution coming in v3.7 this week

PR #4862, opened 2026-06-23
feat(dashboard): Allow configuring TanStack Router plugin options

grolmus:
## Summary

Adds an opt-in `tanstackRouterPluginOptions` option to `vendureDashboardPlugin`. These options are merged on top of the Dashboard's own TanStack Router defaults, so any aspect of the underlying `tanstackRouter()` Vite plugin can be overridden — including `tmpDir`.

Relates to #4048

> Targets `minor` — adds a new public config option (feature), not a behaviour change.

## Problem

TanStack Router's generator writes its output to a temp file and then `fs.rename`s it onto the generated route tree. Its default temp dir resolves to `<cwd>/.tanstack/tmp`, while the dashboard plugin pins the route tree to `<packageRoot>/src/app/routeTree.gen.ts`. When those two paths are on different devices (e.g. `node_modules` on a separate volume — a common deployment-platform layout), the rename fails:

```
Error: EXDEV: cross-device link not permitted, rename ...
```

…and the dashboard build crashes. The only workaround so far was `patch-package`.

## Change

Rather than expose a single-use `tmpDir` option (which would tend to grow as new override needs emerge), the whole options object passed to `tanstackRouter()` is now configurable and merged into the defaults:

- `vite/utils/tanstack-router-config.ts` → `buildTanstackRouterPluginConfig(packageRoot, pluginOptions?)` spreads `...pluginOptions` over the defaults.
- `vite-plugin-vendure-dashboard.ts`: new optional `tanstackRouterPluginOptions?: TanstackRouterPluginOptions` on `VitePluginVendureDashboardOptions` (`@since 3.7.0`).
- The plugin exposes no named options type (it's an inline, inferred `Partial<{...}>`), so the type is derived straight from the plugin's own signature to stay in sync with the installed version:
  ```ts
  export type TanstackRouterPluginOptions = NonNullable<Parameters<typeof tanstackRouter>[0]>;
  ```

```ts
vendureDashboardPlugin({
  vendureConfigPath: './vendure-config.ts',
  tanstackRouterPluginOptions: {
    tmpDir: path.join(packageRoot, '.tanstack-tmp'), // same device as the route tree
  },
});
```

## Test plan

`build:vite` (tsc) clean. The cross-device EXDEV scenario isn't unit-testable (needs two real devices); manual repro: build the dashboard with the temp dir on a separate mount from the checkout. No dashboard UI surface, so no screenshots apply.


michaelbromley (changes requested):
Rather than exposing new single-use options on the vite plugin config type (which will have a tendency to grow over time as new cases emerge where the user needs control over some further aspect of the TanStack router config), I would rather just expose the entire config object that we then pass to the `tanstackRouter()` plugin call. 

This allows the user to override _any_ of the defaults, without us needing to then change our top-level options shape every time.

The challenge is how to represent the TypeScript type of that tanstack options object - I think the router plugin makes heavy use of type inference so that’s the main challenge here.

But the top-level option should be `tanstackRouterPluginOptions` and we merge those into the existing defaults.

By the way, please drop the tests - they are trivial and are basically testing built-in JS functionality and are therefore not worth the extra code and abstraction.

grolmus:
Reworked per your feedback (002f9437b):

- Dropped the single-use `tanstackRouterTmpDir` option. Replaced it with `tanstackRouterPluginOptions`, which is merged on top of the Dashboard's defaults — so any aspect of the router plugin config can be overridden, not just `tmpDir`.
- On the typing question: the plugin exposes no named options type (it's an inline inferred `Partial<{...}>`), so rather than restate it I derive it straight from the plugin's own signature, which keeps it in sync with whatever version is installed:
  ```ts
  export type TanstackRouterPluginOptions = NonNullable<Parameters<typeof tanstackRouter>[0]>;
  ```
- Dropped the unit tests.

`build:vite` (tsc) is clean. Usage now:
```ts
vendureDashboardPlugin({
  vendureConfigPath: './vendure-config.ts',
  tanstackRouterPluginOptions: { tmpDir: path.join(packageRoot, '.tanstack-tmp') },
});
```

PR #4863, opened 2026-06-23
feat(core): Add CustomerChannelAssignmentStrategy to control channel auto-assign

### LINKED ISSUE #4499: feat(core): Add CustomerChannelAssignmentStrategy to control automatic customer-channel assignment in AuthGuard

gabriellbui:
# Description

Adds `CustomerChannelAssignmentStrategy` that controls whether the `AuthGuard` auto-assign an authenticated Customer to the non-default Channel their request token points at.

The strategy has a single hook, `canAssignCustomerToChannel(ctx, customer, channelId)`, that controls **channel assignment only, not access**:

- Returning `true` makes the Customer a member of the active Channel, preserving Vendure's existing always-assign behaviour.
- Returning `false` declines the channel assignment but does **not** block the request; the Customer can still operate on the Channel for the current session.

It is consulted only by the `AuthGuard`, and only for a non-member on a non-default Channel. It is never consulted for the default Channel, under `disableAuth`, or during account-creation flows (registration, verification, external auth, guest checkout). 

The shipped `DefaultCustomerChannelAssignmentStrategy` returns `true`, so existing projects behave exactly as before.

Fixes #4499

# Breaking changes

None. The default strategy preserves the current behaviour: every authenticated Customer is auto-assigned to the Channel they authenticate against.

# Screenshots

N/A — server-side behaviour only.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784820926&installation_id=128408429&pr_number=4863&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4863&signature=40c926eb46a386b8e21e2b0f0666cc2042004b20e95971969c840cab0bf08183"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
+++ b/packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
@@ -10,11 +10,9 @@ import { CustomerService } from '../../services/customer.service';
 
 /**
  * @description
- * Decides whether a signed-in Customer is silently added to the Channel their request points at, so
- * the {@link AuthGuard} doesn't have to carry that logic itself. Someone who already belongs is left
- * untouched; otherwise the configured {@link CustomerChannelAssignmentStrategy} says whether to
- * quietly add them. The default Channel and disableAuth dev mode skip the strategy and always
- * auto-join.
+ * Handles the assignment of a signed-in Customer to the active Channel.
+ *
+ * @docsCategory services
  */
 @Injectable()
 export class CustomerChannelAssignmentService {
@@ -26,22 +24,21 @@ export class CustomerChannelAssignmentService {
 
     /**
      * @description
-     * Auto-joins the active Customer to the active Channel where appropriate. Does not block the
+     * Assigns the active Customer to the active Channel where appropriate. Does not block the
      * request: a Customer the strategy declines to assign may still operate on the Channel for the
      * current session, just without a persisted membership.
      */
-    async resolve(ctx: RequestContext): Promise<void> {
+    async tryAssignToActiveChannel(ctx: RequestContext): Promise<void> {
         const userId = ctx.activeUserId;
         if (!userId) {
             return;
         }
         const { disableAuth, customerChannelAssignmentStrategy } = this.configService.authOptions;
 
-        // The default Channel and disableAuth dev mode always auto-join and never consult the strategy.
-        const isUngated = disableAuth || ctx.channel.code === DEFAULT_CHANNEL_CODE;
+        // The default Channel and disableAuth dev mode always assign and never consult the strategy.
+        const isGated = !disableAuth && ctx.channel.code !== DEFAULT_CHANNEL_CODE;
 
-        // On a non-default Channel an existing member is already where they belong, so there's nothing to do.
-        if (!isUngated) {
+        if (isGated) {
             const member = await this.customerService.findOneByUserId(ctx, userId, true);
             if (member) {
                 return;
@@ -55,7 +52,7 @@ export class CustomerChannelAssignmentService {
         }
 
         const canAssign =
-            isUngated ||
+            !isGated ||
             (await customerChannelAssignmentStrategy.canAssignCustomerToChannel(
                 ctx,
                 customer,
--- a/packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
+++ b/packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
@@ -1,0 +1,84 @@
+import { Injectable } from '@nestjs/common';
+import { DEFAULT_CHANNEL_CODE } from '@vendure/common/lib/shared-constants';
+import { ID } from '@vendure/common/lib/shared-types';
+
+import { RequestContext } from '../../../api/common/request-context';
+import { ConfigService } from '../../../config/config.service';
+import { Customer } from '../../../entity/customer/customer.entity';
+import { ChannelService } from '../../services/channel.service';
+import { CustomerService } from '../../services/customer.service';
+
+/**
+ * @description
+ * Decides whether a signed-in Customer is silently added to the Channel their request points at, so
+ * the {@link AuthGuard} doesn't have to carry that logic itself. Someone who already belongs is left
+ * untouched; otherwise the configured {@link CustomerChannelAssignmentStrategy} says whether to
+ * quietly add them. The default Channel and disableAuth dev mode skip the strategy and always
+ * auto-join.
+ */
+@Injectable()
+export class CustomerChannelAssignmentService {
+    constructor(
+        private configService: ConfigService,
+        private customerService: CustomerService,
+        private channelService: ChannelService,
+    ) {}
+
+    /**
+     * @description
+     * Auto-joins the active Customer to the active Channel where appropriate. Does not block the
+     * request: a Customer the strategy declines to assign may still operate on the Channel for the
+     * current session, just without a persisted membership.
+     */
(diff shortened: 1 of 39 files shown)


packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { RequestContext } from '../../../api/common/request-context';
+import { ConfigService } from '../../../config/config.service';
+import { Customer } from '../../../entity/customer/customer.entity';
+import { ChannelService } from '../../services/channel.service';
+import { CustomerService } from '../../services/customer.service';
+
+/**
+ * @description
+ * Decides whether a signed-in Customer is silently added to the Channel their request points at, so
```
michaelbromley: I don’t like the word “silently” - I don’t get what it is supposed to add to our understanding. 

Also:

> so the {@link AuthGuard} doesn't have to carry that logic itself

This is just describing the general concept of abstraction. You might as well say “this method implements logic so that the call-sites don’t need to implement the logic themselves”. We can get rid of that.

packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+        if (!userId) {
+            return;
+        }
+        const { disableAuth, customerChannelAssignmentStrategy } = this.configService.authOptions;
+
+        // The default Channel and disableAuth dev mode always auto-join and never consult the strategy.
+        const isUngated = disableAuth || ctx.channel.code === DEFAULT_CHANNEL_CODE;
+
+        // On a non-default Channel an existing member is already where they belong, so there's nothing to do.
+        if (!isUngated) {
```
michaelbromley: “not is un-gated” requires a PhD in advanced Boolean logic.

Please refactor to avoid negating a negated verb, i.e. `if (isGated)` or maybe pick another verb.

packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+export class CustomerChannelAssignmentService {
+    constructor(
+        private configService: ConfigService,
+        private customerService: CustomerService,
+        private channelService: ChannelService,
+    ) {}
+
+    /**
+     * @description
+     * Auto-joins the active Customer to the active Channel where appropriate. Does not block the
```
michaelbromley: Can we use the word “assign” rather than “join” to keep the concepts consistent?

packages/core/src/service/helpers/customer-channel-assignment/customer-channel-assignment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+     * request: a Customer the strategy declines to assign may still operate on the Channel for the
+     * current session, just without a persisted membership.
+     */
+    async resolve(ctx: RequestContext): Promise<void> {
```
michaelbromley: I’m not a big fan of the method name `resolve()` - it tells me zero about what this method might do.

Think of a more descriptive name, such as `tryAssignToActiveChannel()`. Yes that’s verbose but it is much less ambiguous.

michaelbromley (changes requested):
Basically good but I have some thorough nit-picks on comments and naming to address 😄

michaelbromley (approved):

PR #4864, opened 2026-06-23
fix(core): seed StockLevel on channel assignment (multi-channel)

### LINKED ISSUE #4860: assignProductVariantsToChannel creates no StockLevel for the target channel's location (MultiChannelStockLocationStrategy) → Dashboard variant Update button stays disabled
## Describe the bug

With `MultiChannelStockLocationStrategy` (the default since 3.1.0), sharing a `Product` across two channels that each have their **own** `StockLocation` leaves variants **without a `StockLevel` for the secondary channel's location**. In the new React Dashboard this makes the **ProductVariant detail "Update" button stay permanently disabled** (the form is treated as invalid), while the browser's unsaved‑changes prompt still fires — so the variant cannot be edited at all from the secondary channel.

Verified against `@vendure/core` **3.6.4**.

## Root cause

Two layers combine:

**1. Core never seeds a `StockLevel` for a secondary channel's location.**

- On variant **create**, a numeric `stockOnHand` is routed through `StockMovementService.adjustProductVariantStock()`, which defaults to a single location only:
  ```ts
  // stock-movement.service.ts
  if (typeof stockOnHandNumberOrInput === 'number') {
    const defaultStockLocation = await this.stockLocationService.defaultStockLocation(ctx);
    stockOnHandInputs = [{ stockLocationId: defaultStockLocation.id, stockOnHand: stockOnHandNumberOrInput }];
  }
  ```
  → only the **default** location gets a `StockLevel`.

- On **assignment** to another channel, `ProductVariantService.assignProductVariantsToChannel()` assigns the catalog + creates a `ProductVariantPrice`, but performs **no stock work at all** — no `StockLevel` is created for the target channel's `StockLocation`:
  ```ts
  // product-variant.service.ts (assignProductVariantsToChannel)
  for (const variant of variants) {
    await this.channelService.assignToChannels(ctx, ProductVariant, variant.id, [input.channelId]);
    await this.createOrUpdateProductVariantPrice(ctx, variant.id, price, input.channelId, targetChannel.defaultCurrencyCode);
    // …assets… ; ProductVariantChannelEvent('assigned') is published — but no stockLevelService/stockMovementService call
  }
  ```

**2. The channel-filtered read then returns empty**, because `StockLevelService.getStockLevelsForVariant()` only returns levels whose location is in the active channel:
```ts
// stock-level.service.ts
.leftJoin('stockLocation.channels', 'channel')
.where('stockLevel.productVariantId = :productVariantId', { productVariantId })
.andWhere('channel.id = :channelId', { channelId: ctx.channelId })
```
So in the secondary channel the `ProductVariant.stockLevels` field resolves to `[]`.

**3. Dashboard symptom:** for an empty `stockLevels`, the generated variant form ends up with a phantom array item `[{}]` whose `stockLocationId` / `stockOnHand` fail the generated zod schema, so `formState.isValid === false` forever → the "Update" button is disabled, even when editing core fields (name/sku). The unsaved‑changes guard only checks `isDirty`, hence the contradictory UX.

## To Reproduce

1. Vendure 3.6.4, default `MultiChannelStockLocationStrategy`.
2. Channel A (default) with "Default Stock Location"; Channel B with its own "Stock B" location.
3. Create a product/variant (it gets a `StockLevel` for the default location only).
4. Assign the product to Channel B.
5. Open the variant in **Channel B** in the Dashboard and edit the name or SKU → the **Update button stays disabled**; on reload the browser warns about unsaved changes.

`productVariant(id).stockLevels` queried with Channel B's token returns `[]`.

## Expected behavior

Either (ideally both):

- **Core:** assigning a variant to a channel (and creating a variant within a channel context) should ensure a `StockLevel` exists for that channel's `StockLocation(s)` (seeded at `stockOnHand: 0`), so per‑channel inventory is usable out of the box.
- **Dashboard:** an empty `stockLevels` should yield `[]`, not an invalid `[{}]` row that silently invalidates the whole form.

## Suggested fix (core)

In `ProductVariantService.assignProductVariantsToChannel()`, after assigning each variant to the target channel, ensure a `StockLevel` exists for that channel's stock location(s). `StockLevelService.getStockLevel(ctx, variantId, locationId)` is already idempotent (it creates a `0/0` row when missing), so the addition is minimal and side‑effect‑free for existing rows, e.g.:

```ts
// inside the per-variant loop, after assignToChannels + price:
const targetCtx = await this.requestContextService.create({ apiType: 'admin', channelOrToken: targetChannel });
const locations = await this.stockLocationService.findAll(targetCtx); // channel-scoped
for (const location of locations.items) {
  await this.stockLevelService.getStockLevel(ctx, variant.id, location.id); // no-op if it already exists
}
```

The same guarantee is needed on the create-in-channel path. Happy to open a PR with a regression test if the maintainers agree on the approach (seed at 0, all of the target channel's locations).

## Environment

- `@vendure/core` / `@vendure/dashboard`: **3.6.4**
- `StockLocationStrategy`: `MultiChannelStockLocationStrategy` (default)
- DB: PostgreSQL

## Workaround (for others hitting this)

Backfill the missing levels, then keep them in sync. Via Admin API, in the target channel's token context:
```graphql
mutation {
  updateProductVariants(input: [{ id: "<variantId>", stockLevels: [{ stockLocationId: "<channelLocationId>", stockOnHand: 0 }] }]) { id }
}
```
A small plugin subscribing to `ProductVariantChannelEvent('assigned')` + `ProductVariantEvent('created')` that idempotently calls `StockLevelService.getStockLevel()` for each of the variant's non-default channel locations closes the gap until core handles it.

michael-attal:
## Summary

Fixes #4860.

With `MultiChannelStockLocationStrategy` (the default since 3.1.0), assigning a `ProductVariant` to a channel created a `ProductVariantPrice` but **no `StockLevel`** for that channel's `StockLocation(s)`. Since `StockLevelService.getStockLevelsForVariant` is channel-filtered (`stockLocation.channels.id = ctx.channelId`), `ProductVariant.stockLevels` then resolves to `[]` in the newly-assigned channel.

In the new React Dashboard this makes the variant detail form invalid (the empty `stockLevels` becomes a phantom `[{}]` row that fails the generated zod schema), so the **"Update" button stays permanently disabled** while the unsaved-changes guard still fires — the variant can't be edited at all from the secondary channel.

## Change

`ProductVariantService.assignProductVariantsToChannel()` now ensures a `StockLevel` exists for each of the target channel's stock locations, right after assigning the variant + price + assets:

```ts
const targetChannelStockLocations = await this.connection
    .getRepository(ctx, StockLocation)
    .createQueryBuilder('stockLocation')
    .innerJoin('stockLocation.channels', 'channel')
    .where('channel.id = :channelId', { channelId: input.channelId })
    .getMany();
// …per variant, after assignToChannel(assets):
for (const stockLocation of targetChannelStockLocations) {
    await this.stockLevelService.getStockLevel(ctx, variant.id, stockLocation.id);
}
```

- `StockLevelService.getStockLevel` is **idempotent**: it creates a `stockOnHand: 0` row when missing and is a no-op when one already exists, so existing levels are untouched.
- Implemented by querying `StockLocation` directly through the already-injected `TransactionalConnection` (no new service dependencies / no `RequestContextService` round-trip), which keeps the change minimal and avoids any DI cycle.

## Notes / open questions

- **Scope:** this covers the channel-**assignment** path (the case in the issue). The issue also mentions the *create-in-a-channel* path (`createSingle` only seeds the default location via `adjustProductVariantStock`). Happy to extend this PR to cover that too if you'd prefer it in one go.
- **Approach:** seeding at `stockOnHand: 0` for *all* of the target channel's stock locations — flagging in case you'd rather restrict to a single/default location for the channel.
- **Tests:** I'd add a regression e2e test to `packages/core/e2e/product-channel.e2e-spec.ts` (assign a product to a channel that has its own `StockLocation`, then assert `productVariant.stockLevels` is non-empty in that channel's context). I couldn't fully run the e2e suite locally — glad to add it once the approach looks right to you.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4864"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784820978&installation_id=128408429&pr_number=4864&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4864&signature=bc7e8ba51d6602a6aeaf4d217239571187ae3150664a003bd869a0e216efd543"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -811,8 +811,9 @@ export class ProductVariantService {
      */
     /**
      * Ensures a `stockOnHand: 0` StockLevel exists for each given variant at every StockLocation of
-     * the given channel. Idempotent and batched: it pre-fetches existing levels and only saves the
-     * missing ones (one SELECT + one bulk insert), so existing stock is never touched.
+     * the given channel. Idempotent, batched and concurrency-safe: it issues a single bulk insert
+     * with `orIgnore()`, so rows that already exist (unique productVariantId + stockLocationId) are
+     * skipped at the DB level and never overwritten — even under concurrent assignment/create requests.
      */
     private async ensureStockLevelsForChannel(
         ctx: RequestContext,
@@ -831,24 +832,21 @@ export class ProductVariantService {
         if (stockLocations.length === 0) {
             return;
         }
-        const locationIds = stockLocations.map(location => location.id);
-        const existing = await this.connection.getRepository(ctx, StockLevel).find({
-            where: { productVariantId: In(variantIds), stockLocationId: In(locationIds) },
-        });
-        const existingKeys = new Set(existing.map(sl => `${sl.productVariantId}:${sl.stockLocationId}`));
-        const toCreate: StockLevel[] = [];
-        for (const productVariantId of variantIds) {
-            for (const stockLocationId of locationIds) {
-                if (!existingKeys.has(`${productVariantId}:${stockLocationId}`)) {
-                    toCreate.push(
-                        new StockLevel({ productVariantId, stockLocationId, stockOnHand: 0, stockAllocated: 0 }),
-                    );
-                }
-            }
-        }
-        if (toCreate.length) {
-            await this.connection.getRepository(ctx, StockLevel).save(toCreate);
-        }
+        const newStockLevels = variantIds.flatMap(productVariantId =>
+            stockLocations.map(stockLocation => ({
+                productVariantId,
+                stockLocationId: stockLocation.id,
+                stockOnHand: 0,
+                stockAllocated: 0,
+            })),
+        );
+        await this.connection
+            .getRepository(ctx, StockLevel)
+            .createQueryBuilder()
+            .insert()
+            .values(newStockLevels)
+            .orIgnore()
+            .execute();
     }
 
     async assignProductVariantsToChannel(
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -33,6 +33,7 @@ import {
     OrderLine,
     ProductOptionGroup,
     ProductVariantPrice,
+    StockLocation,
     TaxCategory,
 } from '../../entity';
 import { FacetValue } from '../../entity/facet-value/facet-value.entity';
@@ -824,6 +825,16 @@ export class ProductVariantService {
         });
         const priceFactor = input.priceFactor != null ? input.priceFactor : 1;
         const targetChannel = await this.connection.getEntityOrThrow(ctx, Channel, input.channelId);
+        // The stock locations that belong to the target channel. A StockLevel is seeded for each
+        // of them (per variant) below, so that per-channel inventory works immediately: otherwise
+        // the channel-filtered `stockLevels` field resolves to `[]` in the newly-assigned channel,
+        // which in the Dashboard produces an invalid variant form / permanently-disabled Update.
+        const targetChannelStockLocations = await this.connection
+            .getRepository(ctx, StockLocation)
+            .createQueryBuilder('stockLocation')
+            .innerJoin('stockLocation.channels', 'channel')
+            .where('channel.id = :channelId', { channelId: input.channelId })
+            .getMany();
         for (const variant of variants) {
             if (variant.deletedAt) {
                 continue;
@@ -841,6 +852,11 @@ export class ProductVariantService {
             );
             const assetIds = variant.assets?.map(a => a.assetId) || [];
             await this.assetService.assignToChannel(ctx, { channelId: input.channelId, assetIds });
+            // `getStockLevel` is idempotent: it creates a `stockOnHand: 0` StockLevel when missing
+            // and is a no-op when one already exists.
+            for (const stockLocation of targetChannelStockLocations) {
+                await this.stockLevelService.getStockLevel(ctx, variant.id, stockLocation.id);
+            }
         }
         const result = await this.findByIds(
             ctx,
--- a/packages/core/src/service/services/product-variant.service.ts
+++ b/packages/core/src/service/services/product-variant.service.ts
@@ -33,6 +33,7 @@ import {
     OrderLine,
     ProductOptionGroup,
     ProductVariantPrice,
+    StockLevel,
     StockLocation,
     TaxCategory,
 } from '../../entity';
@@ -473,6 +474,9 @@ export class ProductVariantService {
                 defaultChannel.defaultCurrencyCode,
             );
         }
+        // Seed a StockLevel for the current channel's stock locations (idempotent), so a variant
+        // created directly within a non-default channel is immediately editable in the Dashboard.
+        await this.ensureStockLevelsForChannel(ctx, [createdVariant.id], ctx.channelId);
         return createdVariant.id;
     }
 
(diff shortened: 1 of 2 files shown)


michael-attal:

michael-attal:
Thanks for the thorough reviews 🙏 — I've pushed `0c8e0a6` addressing all three points:

1. **Create-in-channel path** (raised by CodeRabbit, also noted in #4860): `createSingle` now seeds a `StockLevel` for the current channel's stock locations too, so a variant created directly within a non-default channel is immediately editable in the Dashboard — same root cause as the assignment path.

2. **Performance / N×M queries** (CodeRabbit + Copilot): the per-`(variant, location)` `getStockLevel` loop is replaced by a shared, idempotent `ensureStockLevelsForChannel` helper that **batches** the work — one query to fetch the existing levels for all `(variant, location)` pairs, then a single bulk insert of only the missing ones (relying on the unique `(productVariantId, stockLocationId)` index for safety). Both the assignment and create paths now use it, so bulk `assignProductsToChannel` no longer multiplies queries.

3. **Regression test** (Copilot): added an e2e test in `stock-location.e2e-spec.ts` — assigning a product to a channel that owns a `StockLocation` now asserts a `stockOnHand: 0` `StockLevel` is seeded for that location and is visible via the channel-filtered `stockLevels` field.

Existing levels are never overwritten (the helper only inserts missing pairs).

Note: I wasn't able to run the full e2e suite locally for this change, so I'd appreciate a CI run to confirm — happy to iterate on anything else.

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -804,6 +809,48 @@ export class ProductVariantService {
...
+                if (!existingKeys.has(`${productVariantId}:${stockLocationId}`)) {
+                    toCreate.push(
+                        new StockLevel({ productVariantId, stockLocationId, stockOnHand: 0, stockAllocated: 0 }),
+                    );
+                }
+            }
+        }
+        if (toCreate.length) {
+            await this.connection.getRepository(ctx, StockLevel).save(toCreate);
+        }
```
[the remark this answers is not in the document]
michael-attal: Good catch — fixed in 5a4282c. The check-then-insert flow is replaced by a single bulk insert with `.orIgnore()` (`ON CONFLICT DO NOTHING`), so already-existing `(productVariantId, stockLocationId)` rows are skipped at the DB level. This makes the seeding atomic and idempotent even under concurrent assignment/create requests (and drops it to a single statement).

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -804,6 +809,46 @@ export class ProductVariantService {
...
+                stockAllocated: 0,
+            })),
+        );
+        await this.connection
+            .getRepository(ctx, StockLevel)
+            .createQueryBuilder()
+            .insert()
+            .values(newStockLevels)
+            .orIgnore()
+            .execute();
```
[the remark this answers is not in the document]
michael-attal: Fair point — `.insert()` doesn't chunk, so a single very large assignment could approach the driver parameter limit (≈999 on SQLite < 3.32, 65535 on PostgreSQL). In practice it's bounded by how many variants are assigned in one call, and the surrounding loop already performs per-variant sequential work (price, channel + asset assignment), so this path isn't tuned for unbounded bulk regardless. I'd rather not add chunking just here for a scenario the method isn't otherwise optimized for — but happy to wrap the insert with a `chunkArray` (matching `collection.service.ts`) if you'd prefer it. Deferring to a maintainer on whether it's worth it here.

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -841,7 +887,13 @@ export class ProductVariantService {
             );
             const assetIds = variant.assets?.map(a => a.assetId) || [];
             await this.assetService.assignToChannel(ctx, { channelId: input.channelId, assetIds });
+            assignedVariantIds.push(variant.id);
         }
+        // Seed a StockLevel for each of the target channel's stock locations so that per-channel
+        // inventory is usable immediately (otherwise the channel-filtered `stockLevels` field
+        // resolves to `[]` in the newly-assigned channel — which the Dashboard renders as an
```
michaelbromley: If possible, I’d like to justify or explain the code without reference to the Dashboard, as these are de-coupled systems. Can this comment be re-phrased to explain the need for this call without reference to the Dashboard in particular?

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -804,6 +809,46 @@ export class ProductVariantService {
      * Assigns the specified ProductVariants to the specified Channel. In doing so, it will create a new
      * {@link ProductVariantPrice} and also assign the associated Product and any Assets to the Channel too.
      */
+    /**
```
michaelbromley: This code has been inserted between the doc block of `assignProductVariantsToChannel` and the actual method it is documenting. Please fix this.

michaelbromley (changes requested):
Thanks for this fix! All good, but just a couple of small issues to fix before merge.

packages/core/src/service/services/product-variant.service.ts
michael-attal: sure, reworded in 822662e. Both seeding comments (here and in createSingle) now justify the call purely via the channel-filtered stockLevels field resolving to an empty array, with no reference to the Dashboard.

packages/core/src/service/services/product-variant.service.ts
michael-attal: fixed in 822662e.

michael-attal:
@michaelbromley 822662e handles both points (helper moved out of the doc block + comments reworded without the Dashboard reference). 
Also, I fixed in e593039 the cause of the red MySQL e2e, the insert().orIgnore() needed .updateEntity(false), mirroring DefaultSchedulerStrategy.

michaelbromley (approved):

PR #4868, opened 2026-06-24
fix(dashboard): Surface validation errors when the submit button is disabled

### LINKED ISSUE #4741: [React dashboard] Operation buttons are frequently disabled with no on‑page prompts.
Image

Operation buttons are frequently disabled with no on‑page prompts.

This issue occurs frequently across all pages, including edit, update and create functions. The cause is unknown, and there are no on‑page prompts whatsoever.

product page, variant page, collection page. so on .....

Strongly recommend removing button state control entirely and keeping them always highlighted and clickable. It is unclear whether this is a side effect caused by custom fields.

grolmus:
## Summary

Detail-page forms in the dashboard disable the submit button whenever the form is invalid (`disabled={!isDirty || !isValid || isPending}`), but there was **no on-page indication of *why***. When a field — very often a custom field whose stored value or generated default fails validation — is invalid, the button silently stayed disabled and the only trace was a `console.log`. This is the behaviour reported in #4741: "operation buttons are frequently disabled with no on-page prompts."

This PR keeps the disabled-button logic and the backend untouched and instead **surfaces the reason**.

## Root cause

`useGeneratedForm` builds a zod schema from the entity's input type + custom fields and runs react-hook-form in `mode: 'onChange'`. An invalid value (e.g. a custom field whose value violates its `pattern`/`min`/`max`/enum) makes `formState.isValid` false, disabling submit. Field-level errors only render once a field is *touched*, so on a freshly-loaded edit page nothing was shown — the button just looked broken.

## Change

- **New `FormErrorSummary`** (`lib/components/shared/form-error-summary.tsx`): a destructive `Alert` listing the fields that block saving, rendered once in the shared `PageContentWithOptionalForm`, so **every** detail page (generated and bespoke, e.g. product variants) gets it for free. Flattens nested RHF errors (`customFields.*`, array paths like `stockLevels.0.stockOnHand`) and never renders empty while the form is invalid (errors without a message fall back to a translated generic; the reserved `root` bucket is skipped).
- **Validate on load for existing entities** (`use-generated-form.tsx`): `form.trigger()` runs once per loaded entity (keyed on `entity.id`) so pre-existing invalid values surface immediately instead of only after the user pokes a field.
- Simplified the resolver back to `zodResolver(schema)` and removed two debug `console.log`s.
- Exported `FormErrorSummary` from the public API (`lib/index.ts`).

No change to the `disabled` condition, the mutations, or the backend. A valid form (nothing to save) shows no banner — the button is still correctly disabled with no false alarm.

## Test plan

**Automated** — `e2e/tests/regression/issue-4741-form-error-summary.spec.ts`:
- Adds a nullable `oss540NumericCode` product custom field (pattern `^[0-9]*$`, empty allowed so other tests are unaffected).
- On the product create page: a valid name enables Create; entering `abc` disables it **and** shows the summary banner with the reason; correcting to `12345` clears the banner and re-enables Create.
- Passes on this branch; fails on master (the summary banner does not exist there).

**Manual** (chrome-devtools, before/after screenshots attached on the Linear issue OSS-540): edit page of an entity whose custom field value is invalid — before: Update disabled, no prompt; after: banner + field-level error explain exactly which field and why. Verified a valid page (Zone) shows no banner.

## Out of scope / follow-ups

- The summary uses a humanized field path as a fallback label; resolving the configured (translated) custom-field label in the summary would need the entity type threaded into the shared component. The precise, configured label is already shown next to each offending field.
- The zod validation messages themselves (`Value must match pattern: …`, etc., from `form-schema-tools.ts`) are not yet wrapped for i18n — a pre-existing gap now made more visible. Worth a dedicated follow-up.

Fixes #4741

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4868"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784891532&installation_id=128408429&pr_number=4868&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4868&signature=31950cf983e1492179e904e646671b23800f301691d53f042eceb9cecda8c375"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
Thanks @coderabbitai — all three addressed in 2a0aaf18b:

1. **Revalidate after the form shape settles** (Major): the on-load `trigger()` now also depends on `schema` and `availableLanguages`, so it re-runs if those settle after the entity loads — hidden invalid custom fields/translations no longer stay hidden on first paint.
2. **Disambiguate repeated/nested fields**: `humanizeFieldName` now preserves path context, e.g. `translations.0.name` → `Translations #1 › Name` (array indices shown 1-based; the noise `customFields` segment dropped).
3. **E2E comment format**: switched to the mandated `// \`#4741\` — …` form.

Verified: e2e regression still passes on the branch (and fails without the fix), `tsc` and check-lib-imports green.

michaelbromley:
@grolmus can you include screenshots and/or screen recording of how this looks?

grolmus:
> @grolmus can you include screenshots and/or screen recording of how this looks?

01-valid-no-banner
02-invalid-banner
02b-banner-closeup
03-fixed-no-banner

grolmus:
@michaelbromley Ran the `vendurehq/design` skills as suggested — `npx skills add vendurehq/design`, then `/vendure-ui` and `/vendure-tokens` against this branch.

**TL;DR:** both pass, no blockers. `/vendure-tokens` is clean; `/vendure-ui` flagged only enhancements, and I've applied the one that's actually worth it.

### `/vendure-tokens` — clean
- Host correctly classified as Vendure-owned → component code must consume published token slots, no raw values.
- No literal colors, no generic palette ramps. All colour is delegated to the `destructive` semantic slot via `Alert variant="destructive"`. `list-disc pl-4` / `font-medium` are layout/weight only (typography carries no state colour). Nothing for `@vendure-io/design-lint` to catch.

### `/vendure-ui` — pass, minor findings
1. **(a11y, medium) — fixed.** Summary items were plain text, not linked to the offending field, so they missed the standard WCAG error-summary "skip to field" affordance. Each item is now a `button` that calls `form.setFocus(path)`, which also scrolls a field that's out of view back into focus — the exact scenario this PR targets.
2. **(low) — left as-is.** The banner shifts content when it toggles between header and body. Reserving space would leave a permanent empty gap, which is worse.
3. **(low) — left as-is.** Fallback humanized labels (`Translations #1 › Name`) differ from the configured/translated field labels. Already called out as an out-of-scope follow-up in the PR description; the precise label is still shown next to each field.
4. **(graduation)** `FormErrorSummary` is generic enough to graduate into `@vendure-io/ui` as a molecule if a second consumer wants it. Noting for later, not doing it here.

Typecheck passes with the change.

grolmus:
01-summary-full-page
02-summary-banner
03-click-focuses-field

grolmus:
## CI fixes after updating from `minor`: infinite validation loop + i18n drift

After bringing this branch up to date with the latest `minor`, two jobs were red: the whole `dashboard e2e` matrix (all 4 shards) and `dashboard i18n sync`. Both are now fixed.

### 1. `dashboard e2e` — infinite validate → re-render loop on detail pages

**Symptom.** Dozens of unrelated tests across the suite (product-variants, products/customers CRUD, breadcrumb, translation-placeholders, customer-group…) timed out at 40s, with Playwright reporting the target element *"resolved … waiting for element to be visible, enabled and stable"* — the signature of a page stuck re-rendering. Importantly, `minor` alone is green on these specs, so the regression is introduced by this branch's change, not the base.

**Root cause.** This PR adds an on-load validation effect in `useGeneratedForm`:

```ts
useEffect(() => { if (entity) void form.trigger(); }, [entity?.id, schema, availableLanguages]);
```

`schema` is `useMemo(() => createFormSchemaFromFields(updateFields, customFieldConfig), [updateFields, customFieldConfig])`. The `customFieldConfig` comes from `useCustomFieldConfig()`, which returned a **new array on every render** (`.filter(...) ?? []`, unmemoised). So `schema` got a fresh identity every render → the new effect fired every render → `form.trigger()` updated form state → re-render → new `customFieldConfig` → new `schema` → trigger → … an infinite loop on every entity detail page. The page never settled, so every interaction timed out.

The unstable identity was actually a pre-existing latent issue — `schema`, `defaultValues` and `values` were all recomputing every render, and the form-engine memos explicitly document that they *assume* `customFieldConfig` is identity-stable. The on-load effect was simply the first consumer to turn that instability into a hang.

**Fix.** Memoise the return value of `useCustomFieldConfig` (`useMemo(..., [serverConfig, entityType, hasPermissions])`; `hasPermissions` is already `useCallback`-stable). This restores the identity contract the form engine assumes and repairs every downstream consumer at once — no change to the on-load effect itself. Also added a note on `GeneratedFormOptions.customFieldConfig` documenting that it must be referentially stable, since that's the remaining entry point to the same loop.

**Verification.** Local dashboard e2e, same environment, before/after:

- `products.spec.ts` — 15/15 pass; `product-variants.spec.ts` — 13/13 pass. The exact tests that timed out at 40s in CI (e.g. *should navigate to manage variants page*, *should add an option group via the sidebar dialog*, *force-remove an in-use option group*) now pass in 1–3s each.
- Without the fix, the same spec hangs past 2 minutes.

### 2. `dashboard i18n sync`

The two new `FormErrorSummary` strings (*"This cannot be saved until the following are fixed:"*, *"This field is invalid"*) were only added to `en.po`; the other locale catalogs never received the msgids. Ran `lingui extract` and committed the updated `.po` files (new entries land untranslated, as expected). `scripts/check-i18n-sync.sh` is clean.

michaelbromley:
The code changes look good now. I'd like @dlhck to review the visuals as he's been doing a lot of work in this area for the minor.

PR #4876, opened 2026-06-26
fix(dashboard): Allow removing an option group from the product detail page

### LINKED ISSUE #4703: Should allow us to delete wrong filled `product option group`
Image

Image

for above shown, i wrong filled `size` (existed option), for now i want to delete this product option group relation?

how to do this?, for now i have not create variants..

:(. the new dashboard is very difficult use to upload a new product with many variants..

vendure@3.6.3

grolmus:
## Summary

In the new React dashboard, a product's option groups are shown as badges in the **Product Options** block on the product detail page. Previously each badge only had an *edit* link — there was no way to remove a single option group that had been added by mistake. A user who added the wrong group (e.g. `Size`) before creating any variants could only "remove all" via the variant-setup back button, or use the Manage Variants page — which is only linked once variants already exist. This is the scenario reported in #4703.

## Root cause

`ProductOptionGroupBadge` rendered only the group name + an edit link. The backend mutation `removeOptionGroupFromProduct` (and its `force` argument + `ProductOptionInUseError` result) already existed and is used on the Manage Variants page, but the product detail page exposed no per-group remove affordance.

## Change

- **`ProductOptionGroupBadge`**: new optional `onRemoved` callback. When provided, the badge renders a trash control wrapped in a `ConfirmationDialog`, and — mirroring the Manage Variants page — shows a force-remove `AlertDialog` when the API returns `ProductOptionInUseError` (the group's options are in use by existing variants). The control is opt-in, so the badge's behaviour is unchanged anywhere `onRemoved` is not passed.
- **`products_.$id.tsx`**: passes `onRemoved={() => refreshEntity()}` so removal refreshes the page.
- Adds one i18n string (`Failed to remove option group`) to `en.po`.

No backend changes; reuses the existing `removeOptionGroupFromProductDocument`.

## Test plan

**Automated** — `e2e/tests/catalog/product-variants.spec.ts` (`#4703`):
- Creates a product, adds a `Size` option group, then removes it via the new badge control: asserts the confirmation dialog, the `Option group removed` toast, that the remove control is gone, and that the variant-setup choice (only shown at zero option groups) reappears.
- Passes on this branch; **fails on master** (the remove control does not exist there).
- Full `product-variants.spec.ts` suite passes (12/12); `tsc` (check-types) and `check-lib-imports` are green.

**Manual** (before/after screenshots attached on the Linear issue OSS-521): badge previously showed only the edit pencil; now shows edit + a destructive remove control with a confirmation dialog. Removing the only group returns the page to the variant-setup choice.

## Out of scope / follow-ups

- The `ProductOptionInUseError` → force-remove flow now exists in three places (this badge, the page-level `removeAllOptionGroups`, and the Manage Variants page). Extracting a shared `useRemoveOptionGroupFromProduct` hook + `ForceRemoveOptionGroupDialog` would de-duplicate all three — left as a dedicated follow-up to keep this PR focused on the reported bug.

Fixes #4703

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4876"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785062015&installation_id=128408429&pr_number=4876&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4876&signature=1d56bbc81334aca660384d30ae2fe8bf3eab990ae133e0cf2877ea8c4985e478"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
Before: 
oss521-before-01-overview
oss521-before-02-badge

After:
oss521-after-01-overview
oss521-after-02-badge
oss521-after-03-confirm-dialog
oss521-after-04-removed

biggamesmallworld:
Useful fix for #4703, and the opt-in design is clean — verified the force dialog is only reachable via the trash control, which only renders under `{onRemoved && ...}`, so existing call sites are unaffected.

Blockers before merge:

**1. Run `lingui extract`.** The new `Failed to remove option group` string was hand-added to `en.po`, but the reused strings weren't re-extracted — e.g. `Remove option group` (now used in the badge) still only references `products_.$id_.variants.tsx:315`, and the new string is missing from the non-English `.po` files. Running extract and committing the result fixes the source refs, propagates the key to all locales, and keeps CI's extract-check green.

**2. Cover the force-remove path.** The E2E only exercises the happy path (removing a group with no variants). The `ProductOptionInUseError` → force `AlertDialog` flow — the riskier half of the new code — isn't triggered. Add a case that creates a variant first, then asserts the "Force remove option group" dialog and clicks through.

**3. Extract the shared hook here, not as a follow-up.** This is the third copy of the flow (badge, page-level `removeAllOptionGroups`, Manage Variants). A `useRemoveOptionGroup` hook would collapse all three plus the two near-identical mutations in this file. Consolidating while it's fresh is lower-risk than a later sweep.

Minor / non-blocking (all inherited from the variants page — fix once in the shared hook):
- Trash control isn't permission-gated, whereas the page save uses `requiresPermission={['UpdateProduct', 'UpdateCatalog']}`.
- Force `AlertDialogAction` isn't disabled while pending → fast double-click fires two mutations.
- `result: any` discards the generated discriminated union; the typed result removes the need for the `'__typename' in` guard.

grolmus:
@biggamesmallworld thanks for the thorough review — all three blockers addressed, plus the minors.

1. **`lingui extract`** run and committed — source refs now point at the new hook/dialog files and the strings are propagated to all locale catalogs.
2. **Force-remove E2E** added (`force-remove an in-use option group`): sets up a product + option group + variant via the Admin API so the group is genuinely in use, then drives remove → asserts the *Force remove option group* dialog → clicks through → asserts the removed toast and that the control is gone.
3. **Shared `useRemoveOptionGroup` hook** extracted; the badge, the page-level batch remove, and Manage Variants all go through it now (the batch uses the typed `removeOptionGroupAsync` primitive). The duplicated force dialog is now a single `ForceRemoveOptionGroupDialog`.

Minors, all fixed in the shared code:
- trash control is now behind `PermissionGuard` (`UpdateProduct`/`UpdateCatalog`);
- the force action is `disabled` while pending (with `preventDefault` so a fast double-click can't fire the mutation twice);
- the `result: any` + `'__typename' in` guard are gone in favour of the generated discriminated union.

13/13 `product-variants` e2e pass locally.

biggamesmallworld (approved):
tests pass ✅  looks good to me

PR #4877, opened 2026-06-26
fix(create): Detect package manager for dependency install

### LINKED ISSUE #4390: bunx @vendure/create my-shop not working
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.

**Screenshots/Videos**

**Error logs**

```
Created package.json
│
▲  Something went wrong
 7 | function spawn(command, args, options) {
 8 |     // Parse the arguments
 9 |     const parsed = parse(command, args, options);
10 | 
11 |     // Spawn the child process
12 |     const spawned = cp.spawn(parsed.command, parsed.args, parsed.options);
                            ^
error: Executable not found in $PATH: "npm"
      path: "npm",
     errno: -2,
   syscall: "spawn npm",
 spawnargs: [
  "install", "--save", "--save-exact", "--loglevel", "error", "@vendure/core@3.5.4", "@vendure/email-plugin@3.5.4",
  "@vendure/asset-server-plugin@3.5.4", "@vendure/graphiql-plugin@3.5.4", "@vendure/dashboard@3.5.4",
  "dotenv", "pg"
],
      code: "ENOENT"

      at spawn (node:child_process:667:35)
      at spawn (node:child_process:14:39)
      at spawn (/private/tmp/bunx-501-@vendure/create@latest/node_modules/cross-spawn/index.js:12:24)
      at <anonymous> (/private/tmp/bunx-501-@vendure/create@latest/node_modules/@vendure/create/lib/helpers.js:223:41)
      at new Promise (1:11)
      at installPackages (/private/tmp/bunx-501-@vendure/create@latest/node_modules/@vendure/create/lib/helpers.js:214:12)
      at installDependenciesWithSpinner (/private/tmp/bunx-501-@vendure/create@latest/node_modules/@vendure/create/lib/create-vendure-app.js:492:29)
      at installDependenciesWithSpinner (/private/tmp/bunx-501-@vendure/create@latest/node_modules/@vendure/create/lib/create-vendure-app.js:487:47)
      at createVendureApp (/private/tmp/bunx-501-@vendure/create@latest/node_modules/@vendure/create/lib/create-vendure-app.js:237:11)

Bun v1.3.5 (macOS arm64)
```

- @vendure/core version:
- Nodejs version:
- Database: postgres
- Operating System: macOS
- Browser (if applicable):
- Package manager: bun

**Configuration**

```typescript
// Your configuration here
```

**Minimal reproduction**

**Workaround**
If you found a temporary workaround, please describe it here.

**Additional context**

grolmus:
## Summary

`bunx @vendure/create my-shop` (and the pnpm/yarn equivalents) crashed during dependency installation with:

```
error: Executable not found in $PATH: "npm"
  syscall: "spawn npm"
```

because the installer hard-coded `npm`. When the CLI is run via `bunx` on a machine where `npm` isn't on the PATH, the spawn fails and scaffolding aborts.

## Root cause

`installPackages()` in `packages/create/src/helpers.ts` always spawned `npm install …`. The `PackageManager` type was likewise pinned to `'npm'`.

## Change

- New `detectPackageManager()` — reads the standard `npm_config_user_agent` env var that npm/yarn/pnpm/bun all set when running a package's bin, and returns the detected manager (`'npm' | 'yarn' | 'pnpm' | 'bun'`), falling back to `npm`. This is the same approach used by `create-vite` / `create-next-app`.
- New `getInstallCommand(packageManager, …)` — builds the manager-specific install command (`npm install --save --save-exact …`, `yarn add --exact …`, `pnpm add --save-exact …`, `bun add --exact …`, with the dev-dependency variant).
- `installPackages()` now installs with the detected manager instead of hard-coding `npm`.
- Widened the `PackageManager` type accordingly.

**Scope: install step only.** The generated project is intentionally npm-shaped (its scripts use `concurrently npm:dev:*` shorthands, and the monorepo workspace scripts call `npm`), and the maintainers had deliberately simplified the CLI to npm. This PR therefore changes only the install step that actually crashes; the generated project and the printed `npm run dev` instructions are left as-is. Running the scaffolded project still uses npm.

## Test plan

**Automated** — `packages/create/src/helpers.spec.ts` (new; the package had no tests before, so this also adds a minimal vitest setup):
- `detectPackageManager` resolves npm/yarn/pnpm/bun from representative `npm_config_user_agent` strings, a bare name without a version, and falls back to npm for empty/unknown/unset agents.
- `getInstallCommand` produces the exact, idiomatic command + args for each manager (prod and dev). 8 tests, all passing.
- `tsc` (build + base) and `eslint` are green.

**Manual** — verified end-to-end outside the repo by driving the built `installPackages()` against a throwaway project under different `npm_config_user_agent` values:

| `npm_config_user_agent` | command run | result |
| --- | --- | --- |
| `bun/1.3.5 …` | `bun add` | `bun.lock` + dependency installed |
| `pnpm/8.15 …` | `pnpm add` | `pnpm-lock.yaml` + dependency installed |
| `npm/10.2.4 …` | `npm install` | `package-lock.json` + dependency installed (regression intact) |

Because the bun case invokes `bun add` rather than `npm`, the original `spawn npm ENOENT` can no longer occur. No UI, so no screenshots.

## Out of scope / follow-ups

- **Run-time package manager.** A bun/pnpm/yarn user is still told to run `npm run dev`, and the `--include-storefront` monorepo template hard-codes `npm` in its workspace scripts. Making the generated project fully package-manager-agnostic (per-manager workspace syntax) is a larger, separate change.
- **`--use-npm` override.** The deprecated `--use-npm` flag is still ignored; honouring it as an explicit override of detection could be a small follow-up.

Fixes #4390


michaelbromley:
Pushed two commits onto this branch (coordinated with @grolmus) that extend the fix from "install step only" to **full end-to-end package-manager support** for bun and pnpm. Wanted to lay out the scope so it's not a surprise to reverse-engineer from the diff.

### Why the scope grew

The original fix made `installPackages()` use the detected manager, but several other npm-isms remained, so a bun-only user (the exact case in #4390) still hit npm afterwards:

- **A second `spawn('npm', …)`** in the quick-start auto-launch — identical failure mode to the one this PR fixed, just later in the flow.
- **Empty-deps install bug**: the `--include-storefront` path calls `installPackages({ dependencies: [] })`. `npm install` with no args installs from the manifest, but `bun add` / `pnpm add` / `yarn add` with no args **error**. So storefront install failed for non-npm users (swallowed by `warnOnFailure`, with the recovery message ironically telling them to run `npm install`).
- **Generated project was npm-shaped**: `concurrently npm:dev:*`, `npm run X -w server` workspace scripts, `npm run dev` in READMEs/outro, and a Dockerfile hard-coding `npm install`.
- **Inverted guard**: `checkThatNpmCanReadCwd()` (which spawns npm) was gated `packageManager !== 'npm'` — i.e. it would only ever run for *non-npm* users, where npm may not exist. Flipped to `=== 'npm'`.

### Approach

Single source of truth — `getPackageManagerInfo(pm)` in `helpers.ts` — that every consumer reads from (`runScript`, `install`, `ciInstall`, `lockfile`, `workspaceScript`, `usesPackageJsonWorkspaces`). Generated scripts, CLI text, Dockerfile and READMEs all derive their commands from it.

Notable specifics:
- `concurrently@9.2.1` natively supports `pnpm:`/`bun:`/`yarn:` script-runner prefixes, so the `dev`/`start` scripts just swap the prefix.
- **pnpm** doesn't read the package.json `workspaces` field, so the monorepo layout generates a `pnpm-workspace.yaml` and omits the field for pnpm.
- **Dockerfile** is now pm-aware: `oven/bun:1` base for bun, `corepack enable` for pnpm/yarn, lockfile-aware frozen install (`npm ci` / `pnpm install --frozen-lockfile` / `bun install --frozen-lockfile`). I also dropped `--production`, which was silently breaking the image build — it strips devDependencies, so the subsequent `tsc` had no TypeScript.
- `--use-npm` is now honoured as an explicit override of detection (was parsed and ignored).

### CI

New `test_package_managers` job in `publish_and_install.yml`: matrixed over bun + pnpm, scaffolds via `bunx` / `pnpm dlx` (no `--use-npm`, so detection drives it), asserts the right lockfile was produced, then runs `<pm> run dev` through the existing smoke-test. Reuses the verdaccio artifact.

### Caveats / out of scope

- **Yarn is best-effort, not CI-covered.** The descriptor assumes modern (berry) semantics (`yarn install --immutable`, `yarn workspace <name> <script>`); classic yarn differs. Only bun + pnpm are validated in CI.
- **Monorepo Dockerfile** remains the single-server sample — a Dockerfile inside `apps/server` can't `COPY` the root-level lockfile. Pre-existing; not expanded here.
- The Dockerfile now ships devDependencies into the image (needed for the build step); a proper multi-stage prune is a separate enhancement.

Tests: added empty-deps coverage + a full `getPackageManagerInfo` suite (12 passing). Build + lint green locally; rendered the bun/pnpm Dockerfile and root `package.json` to confirm real output.

PR #4886, opened 2026-06-27
fix(core): Stop double-encoding non-list ID config args

### LINKED ISSUE #4856: Dashboard: disabling coupon visibility resets the assigned customer group
When a promotion's visibility (enabled/disabled toggle) is changed, the previously selected customer group is cleared and must be reassigned manually.

https://github.com/user-attachments/assets/492ee9a4-eda0-450d-8f9d-1782ec85ed66

Expected behavior:
Changing the visibility toggle should not affect any other fields. The customer group assignment should persist.

Actual behavior:
The customer group field is silently cleared after toggling visibility and saving.

Environment: https://demo.vendure.io/admin/promotions/3

ryandiginomad:
# Description

Fixes #4856.

The non-list branch of `encodeConfigurableOperationIds` wrapped the encoded id in `JSON.stringify`:

```ts
} else {
    const encodedId = this.idCodecService.encode(arg.value);
    arg.value = JSON.stringify(encodedId);   // ← re-quotes the id
}
```

Since #2483 (`47f606cc`), the matching **decode** branch reads non-list ids *raw* (no `JSON.parse`), and `coerceValueToType` also reads `ID` back raw (`case 'ID': return value`). So the two sides were asymmetric: encode emitted `'"T_1"'` while decode expected `'T_1'`.

A **read → save round-trip** — which the dashboard performs on any edit — therefore takes a raw id out of the DB, re-quotes it on the way to the client, and mis-decodes the quoted value on the way back in. With `AutoIncrementIdStrategy`, `+'"1"'` → `NaN` → `-1`, so the referenced entity id is silently replaced with an invalid one; with opaque (uuid/base64) ids the quotes simply accumulate. This is why a promotion's assigned customer group disappears after editing: the first save already stores the corrupted value, so on the next load the group shows as empty.

The fix makes **encode** symmetric with the raw decode by dropping the `JSON.stringify`:

```ts
} else {
    arg.value = this.idCodecService.encode(arg.value);
}
```

## Why the encode side (not decode)

This started as a decode-side change (re-adding `JSON.parse`/`JSON.stringify` to the non-list decode branch). That was the wrong layer: it reverts #2483 (reintroducing #2019, where quoted uuids break lookups) and throws on the bare values the admin-ui and e2e actually send (`JSON.parse('T_1')` → `Unexpected token 'T'`). Thanks to @grolmus for catching this and pointing at the real culprit — the asymmetry lives on the encode side. Fixing it there resolves #4856 without touching #2483, without breaking bare input, and consistently with the raw read-back in `coerceValueToType`.

## Backward compatibility

No migration needed. The encode output is only ever produced on the way *out* to a client — it is never persisted; the DB stores the decoded bare primary key. So there is no stored data in the old `'"..."'` shape to migrate, and the decode branch is deliberately left untouched (adding a legacy-quoted tolerance there would re-break #2019). The list branch is also unchanged.

Incidentally this also fixes a latent bug the original issue didn't mention: before this change, a non-list uuid id was double-quoted on encode too, so the fix is slightly broader than the customer-group case.

# Breaking changes

None.

# Checklist

- [x] I have added test cases:
  - `configurable-operation-codec.spec.ts` (the codec had no prior unit tests): raw non-list encode for `AutoIncrementIdStrategy` and `UuidIdStrategy`, the encode → decode round-trip, and the unaffected list branch.
  - `order-promotion.e2e-spec.ts`: a `customerGroup` read-back → re-save round-trip guarding #4856 end-to-end.

### 🤖 AI assistance

Investigated and implemented with the help of Claude Code. I reproduced the bug against a local dev server, traced the root cause to the id-codec layer, and verified the fix end-to-end locally (Bun + core build + the e2e suite on sqljs, plus the new unit spec — including confirming the unit tests fail on the un-fixed code and the full `order-promotion` e2e suite stays green). I reviewed every change myself.

### DIFF
--- a/packages/core/e2e/order-promotion.e2e-spec.ts
+++ b/packages/core/e2e/order-promotion.e2e-spec.ts
@@ -52,7 +52,9 @@ import {
     deletePromotionDocument,
     getFacetListDocument,
     getProductsWithVariantPricesDocument,
+    getPromotionDocument,
     removeCustomersFromGroupDocument,
+    updatePromotionDocument,
 } from './graphql/shared-definitions';
 import {
     addItemToOrderDocument,
@@ -691,6 +693,64 @@ describe('Promotions applied to Orders', () => {
             expect(adjustOrderLine.totalWithTax).toBe(12000);
             expect(adjustOrderLine.discounts.length).toBe(0);
 
+            await deletePromotion(promotion.id);
+        });
+
+        // https://github.com/vendurehq/vendure/issues/4856
+        // A read → save round-trip (what the dashboard does on any edit) must not
+        // corrupt the customerGroupId. Before the fix, encode re-quoted the id
+        // (`'T_1'` → `'"T_1"'`), which then mis-decoded on the way back in and
+        // silently cleared the customer group. The single-create `customerGroup`
+        // test above never exercises this read → re-encode → save path where the
+        // corruption occurs, so this guards the round-trip explicitly.
+        it('customerGroup id survives a read-back → re-save round-trip', async () => {
+            const { createCustomerGroup } = await adminClient.query(createCustomerGroupDocument, {
+                input: { name: 'Round-trip Group', customerIds: ['T_1'] },
+            });
+
+            const promotion = await createPromotion({
+                enabled: true,
+                name: 'Round-trip promotion',
+                conditions: [
+                    {
+                        code: customerGroup.code,
+                        arguments: [{ name: 'customerGroupId', value: createCustomerGroup.id }],
+                    },
+                ],
+                actions: [freeOrderAction],
+            });
+
+            // Read the promotion back — this encodes the condition args for the client.
+            const { promotion: readBack } = await adminClient.query(getPromotionDocument, {
+                id: promotion.id,
+            });
+            const encodedGroupId = readBack!.conditions[0].args.find(
+                a => a.name === 'customerGroupId',
+            )?.value;
+            // Encode must emit the raw id, not a re-quoted `'"T_1"'`.
+            expect(encodedGroupId).toBe(createCustomerGroup.id);
+
+            // Save it again unchanged, exactly as the dashboard does on edit.
+            const { updatePromotion } = await adminClient.query(updatePromotionDocument, {
+                input: {
+                    id: promotion.id,
+                    conditions: readBack!.conditions.map(c => ({
+                        code: c.code,
+                        arguments: c.args.map(a => ({ name: a.name, value: a.value })),
+                    })),
+                },
+            });
+            const savedGroupId = (updatePromotion as PromotionFragment).conditions[0].args.find(
+                a => a.name === 'customerGroupId',
+            )?.value;
+            // After the round-trip the id is still the original group, not a corrupted
+            // `-1` (AutoIncrement) or a quote-accumulated value.
+            expect(savedGroupId).toBe(createCustomerGroup.id);
+
+            await adminClient.query(removeCustomersFromGroupDocument, {
+                groupId: createCustomerGroup.id,
+                customerIds: ['T_1'],
+            });
             await deletePromotion(promotion.id);
         });
     });
--- a/packages/core/src/api/common/configurable-operation-codec.spec.ts
+++ b/packages/core/src/api/common/configurable-operation-codec.spec.ts
@@ -1,0 +1,114 @@
+import { LanguageCode } from '@vendure/common/lib/generated-types';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import { ConfigService } from '../../config/config.service';
+import { MockConfigService } from '../../config/config.service.mock';
+import { AutoIncrementIdStrategy } from '../../config/entity/auto-increment-id-strategy';
+import { EntityIdStrategy } from '../../config/entity/entity-id-strategy';
+import { UuidIdStrategy } from '../../config/entity/uuid-id-strategy';
+import { PromotionCondition } from '../../config/promotion/promotion-condition';
+
+import { ConfigurableOperationCodec } from './configurable-operation-codec';
+import { IdCodecService } from './id-codec.service';
+
+/**
+ * A minimal PromotionCondition def exercising all three arg shapes the codec cares
+ * about: a non-list `ID`, a list `ID`, and a non-ID arg (which must be left alone).
+ */
+const testCondition = new PromotionCondition({
+    code: 'test_condition',
+    description: [{ languageCode: LanguageCode.en, value: 'test' }],
+    args: {
+        singleId: { type: 'ID' },
+        manyIds: { type: 'ID', list: true },
+        label: { type: 'string' },
+    },
+    check: () => true,
+});
+
+function createCodec(idStrategy: EntityIdStrategy<any>): ConfigurableOperationCodec {
+    const configService = new MockConfigService() as unknown as ConfigService;
+    (configService as any).entityIdStrategy = idStrategy;
+    (configService as any).entityOptions = {};
+    configService.promotionOptions.promotionConditions = [testCondition];
+    const idCodecService = new IdCodecService(configService);
+    return new ConfigurableOperationCodec(configService, idCodecService);
+}
+
+describe('ConfigurableOperationCodec', () => {
+    describe('encodeConfigurableOperationIds()', () => {
+        // #4886: a non-list ID must be encoded *raw* (no JSON.stringify), so that a
+        // read → save round-trip does not re-quote the id. Decode reads raw since
+        // #2483, so encode must be symmetric.
+        it('encodes a non-list ID as a raw string with AutoIncrementIdStrategy', () => {
+            const codec = createCodec(new AutoIncrementIdStrategy());
+
+            const [result] = codec.encodeConfigurableOperationIds(PromotionCondition, [
+                { code: 'test_condition', args: [{ name: 'singleId', value: '1' }] },
+            ] as any);
+
+            // bare '1', NOT '"1"'
+            expect(result.args[0].value).toBe('1');
+        });
+
+        it('encodes a non-list ID as a raw string with UuidIdStrategy', () => {
+            const codec = createCodec(new UuidIdStrategy());
+            const uuid = 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee';
+
+            const [result] = codec.encodeConfigurableOperationIds(PromotionCondition, [
+                { code: 'test_condition', args: [{ name: 'singleId', value: uuid }] },
+            ] as any);
+
+            // bare uuid, NOT '"<uuid>"' — the pre-fix double-quoting also corrupted uuids
+            expect(result.args[0].value).toBe(uuid);
+        });
+
+        it('still JSON-stringifies a list of IDs', () => {
+            const codec = createCodec(new AutoIncrementIdStrategy());
+
+            const [result] = codec.encodeConfigurableOperationIds(PromotionCondition, [
+                { code: 'test_condition', args: [{ name: 'manyIds', value: JSON.stringify(['1', '2']) }] },
+            ] as any);
+
+            expect(result.args[0].value).toBe(JSON.stringify(['1', '2']));
+        });
+
+        it('leaves non-ID args untouched', () => {
+            const codec = createCodec(new AutoIncrementIdStrategy());
+
+            const [result] = codec.encodeConfigurableOperationIds(PromotionCondition, [
+                { code: 'test_condition', args: [{ name: 'label', value: 'hello' }] },
+            ] as any);
+
+            expect(result.args[0].value).toBe('hello');
+        });
+    });
+
+    describe('encode → decode round-trip', () => {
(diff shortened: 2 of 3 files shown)


ryandiginomad:

ryandiginomad:
recheck

michaelbromley:
I think this reintroduces #2019. The non-list branch here was changed by #2483 back in Oct 2023, which deliberately removed the JSON.stringify(decodedId) because it was storing quoted ids ('"abc-uuid"') that broke UUID lookups — it only ever worked for integers because JSON.stringify(1) === '1'. This PR adds that stringify back, plus a JSON.parse(arg.value) that'll throw on a bare UUID.

The reason it looks fine is the tests use AutoIncrementIdStrategy, where both problems are masked. The actual contract is in coerceValueToType (configurable-operation.ts:459) — non-list IDs are read back raw with no parse, so the DB needs to hold the bare id, not a stringified one.

Worth noting the bug itself is dashboard-specific (the Angular admin never had it), so this might be better fixed client-side, or at least kept bare on the server, e.g. parse-if-JSON then decode without re-stringifying.

There's another PR for the same issue that is tackling it from the dashboard side only: #4887, which I'm looking at too.

ryandiginomad:
@michaelbromley You're right on all counts — thank you for the #2483 context, that's the piece I missed. I dug into the history and you've correctly identified that this PR reintroduces #2019, plus adds a new failure mode. Walking through it:

**#2483 deliberately removed the `JSON.stringify`.** Commit `47f606cc` ("Fix configurable operation id decoding when using uuids") changed the non-list branch from `JSON.stringify(decode(arg.value))` to bare `decode(arg.value)` precisely so UUIDs aren't stored quoted. My PR adds that `stringify` back, so it's a straight regression of that fix.

**My added `JSON.parse(arg.value)` is actually worse than the pre-#2483 state** — it *throws* on a bare UUID rather than silently mis-decoding. Running the real `UuidIdStrategy`/`AutoIncrementIdStrategy` against the three implementations:

| input | master | this PR | parse-if-JSON, store bare |
|---|---|---|---|
| bare integer (Angular/e2e) | `1` | `"1"` | `"1"` |
| quoted integer (the #4856 case) | `-1` ❌ | `"1"` ✅ | `"1"` ✅ |
| bare uuid (Angular/e2e) | `"abc-uuid"` | 💥 throws | `"abc-uuid"` ✅ |
| quoted uuid | `"\"abc-uuid\""` ❌ | `"\"abc-uuid\""` ❌ | `"abc-uuid"` ✅ |

So this PR would 500 promotion create/update on any UUID-strategy deployment (the e2e suite itself sends bare ids at `order-promotion.e2e-spec.ts:485`), and still corrupt quoted UUIDs. And you're right that the contract is `coerceValueToType` (`configurable-operation.ts:459`, `case 'ID': return value as any` — raw, no parse), so the DB must hold the bare id. My tests only passed because they hard-code `AutoIncrementIdStrategy` *and* feed `JSON.stringify('1')`, which masks both problems.

One thing I noticed while verifying: master *also* corrupts a quoted UUID (last row) — the real root cause is the asymmetry where non-list IDs are written JSON-stringified (encode path, lines 76-78, untouched by #2483) but read back raw. #4887 fixes the dashboard symptom cleanly and is the right near-term fix; the deeper read/write asymmetry is a separate, larger thing.

Happy to defer to #4887 here. If it's useful, I can repoint my regression test to use a bare value + a `UuidIdStrategy` case (the input shape the server actually receives) so whichever fix lands is guarded against both the NaN→-1 wipe and the quoted-UUID regression — either as a small PR against #4887's branch or a standalone test PR. Otherwise I'll close this. Let me know what you'd prefer.

grolmus:
Thanks for the thorough investigation and writeup, @ryandiginomad — the root-cause analysis is spot on, and this is the correct layer to fix. The non-list decode branch was asymmetric with the encode branch (`encodeConfigurableOperationIds` already does `JSON.stringify(encodedId)`), so `decodeId` was receiving `'"1"'` with the quotes intact and `AutoIncrementIdStrategy` mapped the resulting `NaN` to `-1`. I reproduced the same corruption independently against the default strategy and confirmed your fix round-trips correctly for both auto-increment and opaque (uuid/base64) id strategies. The regression tests are a nice touch.

We compared this against the alternative dashboard-side PR (#4887) and confirmed the fix belongs here in core — a client-side change leaves the server decode broken for non-dashboard clients, so we're going with this one.

Two small, non-blocking notes:
1. Consider a one-line comment on the new branch pointing at the symmetry with the encode branch, so the `JSON.parse` / `JSON.stringify` pairing doesn't look accidental to a future reader.
2. `JSON.parse(arg.value) as string` assumes the value is always JSON-stringified — which is guaranteed by the encode path, so this is fine; just flagging the assumption.

Before we can merge, two CI items need sorting:
- **CLA** — the CLAAssistant check is failing; please sign the CLA so we're able to merge.
- **e2e** — the e2e jobs are currently red across all four databases (unit tests pass). Given how targeted this change is, it may be a flaky/infra failure or something already fixed on `master`. Could you rebase onto latest `master` and re-run? If it persists, take a look in case an e2e exercising configurable-operation id args is affected.

Approving in principle — happy to hit the button once CLA + CI are green. Nice work tracing this to the codec layer.

ryandiginomad:
Thanks for taking the time to reproduce this, @grolmus, and for confirming the codec is the right layer. One bit of history worth surfacing before you merge, in case it didn't surface in the thread: @michaelbromley looked at this on June 29 and flagged that it reintroduces #2019. I dug in and think he's right — so I'd hold off on merging `8faaa574c` as-is. The details:

- #2483 (commit `47f606cc`, "Fix configurable operation id decoding when using uuids") deliberately changed the non-list branch to decode the id **raw**, so ids aren't stored quoted. This PR re-adds the `JSON.parse` / `JSON.stringify` that #2483 removed — effectively reverting it.
- The read-back contract is `coerceValueToType` (`configurable-operation.ts:459`, `case 'ID': return value as any` — raw, no parse), so the DB is meant to hold the bare id.

I assumed the red e2e was flaky/infra too at first, but it turns out to be caused by the fix. `JSON.parse(arg.value)` throws when the value arrives bare, which is what the existing customerGroup test sends:

```
✕ Promotions applied to Orders > default PromotionConditions > customerGroup
  → Unexpected token 'T', "T_1" is not valid JSON
    at createPromotion (order-promotion.e2e-spec.ts)
    variables: { ... value: 'T_1' }
```

The test sends `value: createCustomerGroup.id`, and both the e2e strategy and the real Angular admin-ui (`encodeConfigArgValue` → `(value ?? '').toString()` for non-list args) send the id bare, never JSON-stringified — so a rebase won't clear it. I think that also reconciles your note #2: the value is stringified for the round-tripped-through-encode case (which is what round-trips cleanly for you), but the direct client input arrives bare. That's why it works for the numeric default id (`JSON.parse("1")` is valid) and throws for `T_1` / uuid / base64.

Since #4887 was closed, whichever you'd prefer:

1. I rework this to store the bare id (parse only when the value is genuinely JSON-stringified, matching the `coerceValueToType` read-back), plus a `UuidIdStrategy` + bare-value regression test so the e2e suite guards both shapes. Keeps the fix in core.
2. Or reopen #4887 for the dashboard symptom and I contribute the bare-value / UuidIdStrategy regression test standalone.

Either way I'd want it green against the full e2e suite before merge, not just the unit spec. Happy to turn around whichever direction you and @michaelbromley prefer.

grolmus (changes requested):
You're right, and thanks for the rigor — my earlier "approve in principle" was premature. I've now traced both client paths end-to-end to land on the strongest fix, and I want to lay out the reasoning since you and @michaelbromley leaned toward the dashboard-only route (#4887).

## Root cause is server-side, not the dashboard

The tell is *why* the Angular admin doesn't corrupt while the React dashboard does:

- **Angular admin** unwraps the value on read (`getConfigArgValue` → `JSON.parse('"1"')` → `'1'`) and re-serializes **every** arg on save (`encodeConfigArgValue`), so it absorbs the server's quoting and always sends bare `'1'`.
- **React dashboard** maps `arg.value` in **raw** (`promotions_.$id.tsx`, `setValuesForUpdate`) and only re-serializes the args you actually edit. So an **untouched** ID arg is submitted still-quoted `'"1"'` → `decodeId('"1"')` → `NaN` → `-1`. That's #4856: save a promotion without touching `customerGroupId` and it resets. (Edit the arg and it round-trips fine — which is why it looks intermittent.)

Both symptoms trace to one server defect: non-list **encode** still `JSON.stringify`s (`configurable-operation-codec.ts:75-76`) while **decode** (post-#2483) reads raw (`:46`), and `coerceValueToType` reads `ID` back raw (`configurable-operation.ts:459`). Encode quoting is a wire artifact the client is expected to unwrap — Angular does, React doesn't.

## The fix: make encode symmetric (drop the stringify), keep decode raw

```ts
// encode, non-list branch
} else {
    arg.value = this.idCodecService.encode(arg.value);
}
```

Why this over the decode-side change in this PR, and over a dashboard-only fix:

- **Reverting #2483 (what this PR currently does) is wrong** — re-adding `JSON.parse`/`JSON.stringify` to decode reintroduces #2019 (quoted UUIDs) and *throws* on the bare values the admin-ui and e2e actually send (`JSON.parse('T_1')` → `Unexpected token 'T'` — that's the red e2e, not flake).
- **Encode-side (A) keeps Angular working** — I verified its read path tolerates bare (`JSON.parse('1')` is a scalar; UUIDs fall back to raw via the catch), so no regression there.
- **It fixes the React round-trip** (bare in → bare out → raw decode).
- **It also fixes any non-dashboard/REST client** that faithfully echoes the server value — which a dashboard-only fix (#4887) structurally cannot. That's the deciding factor for keeping this in core.
- It respects #2483 (UUIDs stay raw) and matches the raw `coerceValueToType` read-back, and it's a one-line change with the smallest blast radius.

## What I'd like to see

- Rework to the **encode** side as above (decode untouched).
- Your proposed **`UuidIdStrategy` + bare-value** regression tests, plus a **round-trip (encode→decode)** case, so the suite guards both the numeric and opaque/uuid shapes.
- Green against the **full e2e suite** (the configurable-operation-heavy specs — promotions/shipping/payment/collection — since those admin-ui detail components also read these args), not just the unit spec.

Want to push that here on this branch, or shall I take it from your branch and credit you? Either way #4887 stays closed — the core fix covers the dashboard symptom and everything else. Genuinely great debugging on this; sorry for the detour my first pass caused.

ryandiginomad:
Done — pushed to this branch, reworked exactly to the encode side. And your React-vs-Angular breakdown nails it: the tell really is that the React dashboard submits an untouched ID arg still-quoted while Angular re-serializes every arg, so both symptoms collapse to the one server-side asymmetry. That framing is clearer than where I'd left it.

What's in the push:

- **Encode-side fix, decode untouched.** `encodeConfigurableOperationIds` now drops the `JSON.stringify` for the non-list branch (`arg.value = this.idCodecService.encode(arg.value)`); the decode branch and #2483 are left exactly as they are. Added a one-line comment there pointing at the symmetry with decode so it doesn't read as accidental.

- **Regression tests, both shapes.** New `configurable-operation-codec.spec.ts` (the codec had no unit tests): raw non-list encode for `AutoIncrementIdStrategy` *and* `UuidIdStrategy`, an encode→decode round-trip for both, and the unaffected list branch. I confirmed all four ID-path assertions fail on the un-fixed encode (AutoIncrement round-trips to `-1`, uuid accumulates quotes) and pass with it. Plus an e2e `customerGroup` read-back → re-save round-trip in `order-promotion.e2e-spec.ts` reproducing #4856 directly.

- **Green against the full e2e suite.** Ran the configurable-operation-heavy specs you named, locally on sqljs: `order-promotion` (66), `collection` (77), `promotion` + `promotion-side-effects` (22), `shipping-method` + `shipping-method-eligibility` (23), `payment-method` (16) — all green. (The collection filters happen to use `list` ID args, which this change doesn't touch, but I ran them anyway to be sure.) CI is re-running now that CLA is signed.

PR title/description are updated to describe the encode-side fix rather than the old decode approach. Thanks again for the catch and the detailed reasoning — over to you and @michaelbromley.

grolmus (approved):
This is exactly right now — thanks for the persistence and for catching my premature approval earlier.

The fix is on the correct side: `encodeConfigurableOperationIds` non-list branch now emits the id raw (`this.idCodecService.encode(arg.value)`, no `JSON.stringify`), symmetric with the decode branch that's read raw since #2483 — so it fixes the #4856 read→save round-trip without reverting the UUID fix or breaking bare input. The comment pointing at #2483/#4856 will save the next reader the archaeology.

Test coverage is spot-on:
- Unit spec covers non-list encode raw for **both** AutoIncrement and Uuid strategies, list still JSON-stringified, non-ID args untouched, and the **encode→decode round-trip** for both strategies.
- The e2e guards the actual reported scenario — customerGroup id surviving a read-back → re-save round-trip (the path the single-create test never exercised).

I verified CI locally-equivalent: the full suite is green now, including **e2e across all four databases** (mariadb/mysql/postgres/sqljs) that the earlier revision failed — confirming the `JSON.parse('T_1')` throw is gone and the round-trip holds.

LGTM 🚀 Approving. Really nice debugging on this one.

PR #4887, opened 2026-06-27
fix(dashboard): preserve configurable ID args

### LINKED ISSUE #4856: Dashboard: disabling coupon visibility resets the assigned customer group

realicon23:
## What

Fixes #4856.

This updates the dashboard form value transformer so scalar `ID` configurable-operation args are treated like opaque string values instead of being parsed through `JSON.parse`.

## Why

Promotion conditions store their configured arguments as strings. The customer-group condition uses a `customerGroupId` argument whose GraphQL type is `ID`. Since GraphQL IDs are opaque identifiers, a value such as `"3"` should stay a string all the way into custom form inputs.

Before this change, the `jsonStringValueTransformer` only bypassed JSON parsing for scalar `string` fields. Numeric-looking `ID` values could therefore become numbers, which is a poor fit for relation-style inputs and can cause unrelated promotion updates, such as toggling `enabled`, to lose the selected customer group.

The narrowest fix is to preserve scalar `ID` args at the shared transformer boundary, rather than special-casing the promotion page or the customer-group input.

## Verification

- `git diff --check`
- `npx --yes vitest@3.2.4 run packages/dashboard/src/lib/framework/form-engine/value-transformers.spec.ts --environment node`
- `npx --yes prettier@3.2.5 --no-config --single-quote --tab-width 4 --print-width 110 --check packages/dashboard/src/lib/framework/form-engine/value-transformers.ts packages/dashboard/src/lib/framework/form-engine/value-transformers.spec.ts`
- Checked the touched files for common secret/token patterns before pushing.

I did not run the full dashboard test suite locally because this environment does not have the full Bun-installed Vendure dependency tree available.

### DIFF
--- a/packages/dashboard/src/lib/framework/form-engine/value-transformers.spec.ts
+++ b/packages/dashboard/src/lib/framework/form-engine/value-transformers.spec.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from 'vitest';
+
+import type { ConfigurableFieldDef } from './form-engine-types.js';
+import { jsonStringValueTransformer } from './value-transformers.js';
+
+const fieldDef = (overrides: Partial<ConfigurableFieldDef>): ConfigurableFieldDef =>
+    ({
+        name: 'test',
+        type: 'string',
+        list: false,
+        ...overrides,
+    }) as ConfigurableFieldDef;
+
+describe('jsonStringValueTransformer', () => {
+    it('does not JSON-parse scalar ID values', () => {
+        const def = fieldDef({ type: 'ID' });
+
+        expect(jsonStringValueTransformer.parse('3', def)).toBe('3');
+        expect(jsonStringValueTransformer.parse('001', def)).toBe('001');
+    });
+
+    it('serializes scalar ID values as opaque strings', () => {
+        const def = fieldDef({ type: 'ID' });
+
+        expect(jsonStringValueTransformer.serialize('3', def)).toBe('3');
+        expect(jsonStringValueTransformer.serialize(3, def)).toBe('3');
+    });
+});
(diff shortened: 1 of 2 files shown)


grolmus:
Thank you for digging into this and for the clean, well-tested PR — the list-serialization fix in your follow-up commit was also the right call. After tracing the full round-trip, though, we're going to fix this one core-side (#4886) rather than in the dashboard.

The corruption actually happens on the server in `ConfigurableOperationCodec.decodeConfigurableOperationIds`: the non-list `ID` branch decoded the raw JSON-stringified value (`'"3"'`) without `JSON.parse`-ing it first, so `AutoIncrementIdStrategy.decodeId` coerced it to `NaN` → `-1` and overwrote the reference. Because the value the server hands back to any client is the stringified `'"3"'`, adjusting only the dashboard transformer doesn't remove the `'"3"'` that trips the server decode — the same corruption still occurs on save, and would still affect non-dashboard clients (REST/custom).

Once the core decode is made symmetric with the encode path (#4886), the existing dashboard parse/serialize round-trips correctly with no dashboard change needed.

Really appreciate the effort and the quality of the PR — I hope you'll pick up another issue. Closing in favour of #4886.

PR #4913, opened 2026-07-01
fix(dashboard): Compile config outside the type:module package

### LINKED ISSUE #4979: Dashboard build fails on Windows/pnpm when config is compiled inside type:module package
**Describe the bug**

On Windows in a pnpm workspace, the React Dashboard build can fail while loading the Vendure config because the default temporary compilation directory is inside the published `@vendure/dashboard` package, whose package scope is `"type": "module"`.

The config compiler emits CommonJS JavaScript. During the Vite build, Node can still classify the generated `.js` file under the parent ESM package scope and reject the CommonJS `exports` assignment.

**To Reproduce**

1. Use Vendure 3.7.0 in a pnpm workspace on Windows.
2. Configure the React Dashboard with the default config-loader options (do not set `tempCompilationDir`).
3. Run the dashboard production build.
4. The Vendure config compilation is written below `@vendure/dashboard/dist/vite/.vendure-dashboard-temp` and config loading fails.

**Expected behavior**

The compiled Vendure config loads as CommonJS and the dashboard build completes.

**Actual behavior**

The build fails while loading the generated config:

```
[vendure:config-loader] Error loading Vendure config: Error loading config:
exports is not defined in ES module scope
```

Node identifies the parent `@vendure/dashboard/package.json` with `"type": "module"` as the relevant package scope.

**Error logs**

```
ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension
and '@vendure/dashboard/package.json' contains "type": "module".
```

**Environment**

- @vendure/dashboard version: 3.7.0
- Nodejs version: 24.15.0
- Operating System: Windows 11
- Package manager: pnpm 10.33.0

**Minimal reproduction**

A dashboard build using the default `vendureDashboard()` config-loader path is sufficient in the affected Windows/pnpm workspace. The failure is deterministic in the original project. A direct isolated compiler invocation succeeds, so the observed failure is specific to the Vite build flow rather than TypeScript compilation itself.

**Workaround**

Set `configLoaderOptions.tempCompilationDir` to a package-neutral location under the consuming project:

```ts
vendureDashboard({
  configLoaderOptions: {
    tempCompilationDir: path.join(
      process.cwd(),
      'node_modules/.cache/vendure-dashboard-temp',
    ),
  },
})
```

With that override, the same dashboard build completes successfully.

TheStreamCode:
Fixes #4979.

# Description

The React Dashboard build can fail on Windows/pnpm while loading a compiled Vendure config:

```
[vendure:config-loader] Error loading Vendure config: Error loading config:
exports is not defined in ES module scope
```

The config compiler emits CommonJS and previously defaulted its temporary output to a directory inside the published `@vendure/dashboard` package. In the affected Vite build, Node resolves the generated `.js` file against the parent package's `"type": "module"` scope despite the compiler-written child package manifest.

This changes the default to the consuming project's:

```
node_modules/.cache/vendure-dashboard-temp
```

That location remains close enough for normal package resolution while avoiding the published ESM package scope. The existing `tempCompilationDir` option still overrides the default and is now documented.

# Verification

- Reproduced before the fix on Windows 11, Node 24.15.0, pnpm 10.33.0 and Vendure 3.7.0: the dashboard production build failed while loading the generated CommonJS config from `@vendure/dashboard/dist/vite/.vendure-dashboard-temp`.
- Verified the same downstream build succeeds when the temp directory is placed under the consuming project's `node_modules/.cache` (9,320 modules transformed; build completed).
- Built the required `@vendure/common` and `@vendure/core` workspaces from this checkout.
- `vitest run vite/tests/temp-compilation-dir.spec.ts`: 2 tests passed.
- ESLint on the three affected files: 0 warnings, 0 errors.
- Upstream CI on the original revision passed build, unit, dashboard E2E and package-manager checks. The separate deploy failure was caused by unavailable fork credentials, not this code path.

# Breaking changes

None. Only the default temporary location changes; explicit `tempCompilationDir` values retain their current behavior.

# Checklist


packages/dashboard/vite/utils/temp-compilation-dir.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Resolves the default directory into which the VendureConfig is transpiled (to
+ * CommonJS) and dynamically imported during the dashboard build.
+ *
+ * This deliberately does NOT live inside the `@vendure/dashboard` package. That
+ * package is published as `"type": "module"`, so a CommonJS file emitted into it
+ * is treated by Node as ESM and loading it throws
+ * `ReferenceError: exports is not defined in ES module scope`. This surfaces
+ * during `vite build` (where several build environments compile into the shared
+ * temp dir) on Windows/pnpm — see https://github.com/vendurehq/vendure/issues.
```
biggamesmallworld: @TheStreamCode can you add the link to the original issue reported in the JSDoc here please?

packages/dashboard/vite/tests/temp-compilation-dir.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getDefaultTempCompilationDir, TEMP_COMPILATION_DIR_NAME } from '../utils/temp-compilation-dir.js';
+
+describe('getDefaultTempCompilationDir', () => {
+    it('resolves under the project node_modules/.cache', () => {
+        const cwd = path.join(path.sep, 'tmp', 'my-project');
+
+        const dir = getDefaultTempCompilationDir(cwd);
+
+        expect(dir).toBe(path.join(cwd, 'node_modules', '.cache', TEMP_COMPILATION_DIR_NAME));
+        expect(dir.startsWith(cwd)).toBe(true);
```
biggamesmallworld: `toBe()` on line 12 already fully pins the value; `startsWith(cwd)` is logically implied by it and adds nothing. Drop it.

```suggestion
```

packages/dashboard/vite/tests/temp-compilation-dir.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+
+    it('is not located inside the @vendure/dashboard package', () => {
+        // The config is compiled to CommonJS; if the output lived inside the
+        // `type: module` @vendure/dashboard package, Node would load it as ESM and
+        // fail with "exports is not defined in ES module scope".
+        const dir = getDefaultTempCompilationDir(path.join(path.sep, 'tmp', 'my-project'));
+
+        expect(dir).not.toContain(path.join('@vendure', 'dashboard'));
+    });
```
biggamesmallworld: This test is tautological: it feeds a synthetic `/tmp/my-project` and then asserts the result doesn't contain `@vendure/dashboard` — true by construction, regardless of what the function does with the path. It does not verify the property its name claims ("not located inside the @vendure/dashboard package"), which is only meaningful against the real install location that a unit test never exercises. Concretely: mutate the util to return a package-internal path and this test stays green, so it can't catch the regression it's named after. Tests 1 and 3 (the exact `toBe()` assertions) fully cover the helper — recommend removing this one.

```suggestion
```

packages/dashboard/vite/utils/temp-compilation-dir.ts
TheStreamCode: Addressed in b9d598401. The JSDoc now links to the upstream bug report (#4979) and describes the observed Windows/pnpm package-scope behavior.

packages/dashboard/vite/tests/temp-compilation-dir.spec.ts
TheStreamCode: Addressed in b9d598401. Removed the redundant `startsWith(cwd)` assertion; the exact path assertion fully covers it.

packages/dashboard/vite/tests/temp-compilation-dir.spec.ts
TheStreamCode: Addressed in b9d598401. Removed the tautological package-location test; the two remaining exact path assertions cover the helper contract.

TheStreamCode:
Thanks @biggamesmallworld — all three points are addressed in `b9d598401`:

- The JSDoc now links the upstream report (#4979) and describes the observed Windows/pnpm package-scope behaviour.
- Dropped the redundant `startsWith(cwd)` assertion.
- Removed the tautological package-location test; the two exact-path assertions cover the helper contract.

Application CI is green. The only remaining red check is the preview deployment, which cannot run from a fork without repository secrets — nothing in the diff affects it.

Happy to rebase or adjust anything else if that helps.

TheStreamCode:
Pushed 8551812: reworded the JSDoc on `getDefaultTempCompilationDir` to describe the actual trigger (upward local import escaping `outputPath` past the commonjs guard) rather than the Windows/pnpm framing your repro disproved. Also aligned on `node:path` for consistency with the test file.

The `tempCompilationDir` option doc was already stated in terms of `type: module` scope so it's untouched.

Happy to open the follow-up issue for the `sourceRoot` fix (common ancestor of collected files, compiled config path derived from actual emit to keep `getCompiledConfigPath` working for monorepo overrides) once this lands.

biggamesmallworld:
Verified locally on macOS by reproducing the failure **and** the fix against the real compiler under node-native module loading (the same mechanism `vite build` uses).

**Reproduction**

Drove the real `compile()` (`vite/utils/compiler.ts`) with a config that has an upward local import (`import { marker } from '../shared/util.js'`), so the compiler emits the resolved file one level above `outputPath` — escaping the compiler-written `{"type":"commonjs"}` guard, exactly the scenario this PR targets. Then loaded the emitted entry with node-native `import()`:

| Temp dir location | Produced by | node-native `import()` |
|---|---|---|
| inside `@vendure/dashboard` (type:module) | pre-fix default | ❌ `exports is not defined in ES module scope` |
| `<project>/node_modules/.cache/vendure-dashboard-temp` | this PR's `getDefaultTempCompilationDir()` | ✅ loads successfully |

The error on the old path matches the issue title verbatim (#4979). The NEW path is generated by the PR's actual default resolver, not hardcoded, and the escaped file lands under `node_modules/.cache` whose nearest `package.json` is the (CommonJS) project root — so it loads correctly.

**Note on runtime faithfulness:** running `compile()` under `bun` or `tsx` makes *both* paths pass, because those runtimes register loader hooks that bypass Node's native `type:module` enforcement. Only node-native loading surfaces the bug — worth knowing for anyone trying to reproduce, and the reason a plain unit test of the path helper isn't sufficient evidence on its own.

**Not run locally:** the full end-to-end `vite build` (blocked here by an unrelated missing dev dep and the later build stages requiring a running API). Instead I exercised the decisive subsystem in isolation — `compile()` → node-native `import()` — which is where the failure occurs and what this PR changes.

**Also confirmed**
- Clean diff against current upstream `master` (3 files, +67/-1).
- Unit tests pass (2/2). Mutation check: reverting the util to a package-internal path makes both tests fail — the earlier tautological third test (which stayed green under that mutation) has been removed, so every remaining test is load-bearing.
- JSDoc now describes the real mechanism (upward `../` import escaping the guard) and links #4979.

biggamesmallworld (approved):

PR #4074, opened 2025-12-21
fix(dashboard): Add several missing translation strings

lucatk:
# Description

I have found several occurrences of single strings not being (properly) translated yet. I have wrapped them in \<Trans> or t`` and added translations with the help of Claude Code.

German translations have been verified, the other languages I'm of no use unfortunately.

# Breaking changes

No.

# Screenshots

-

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

PR #4914, opened 2026-07-02
fix(dashboard): Localize configurable operation selector buttons

### LINKED ISSUE #4410: Dashboard i18n regression: Collections “Add collection filter” UI remains in English after switching language (new collection page)
**Describe the bug**
When switching the Dashboard language from English to another language (Spanish), several UI strings on the New Collection page remain in English. Specifically, the “Add collection filter” button, its dropdown options, and multiple labels inside the inserted filter cards do not translate, even though the rest of the dashboard reflects the selected language.

**To Reproduce**

1. Set the Dashboard language to English.
2. Navigate to the Collections page at `/dashboard/collections`.
3. Click the “New Collection” button.
4. Change the language to a non-English language (for example Spanish).
5. On the New Collection page at `/dashboard/collections/new`, observe that the “Add collection filter” button in the Filters section is still shown in English.
6. Click the “Add collection filter” button and observe that the dropdown options are also in English (not translated).
7. Click any dropdown option and observe that the newly inserted filter card contains multiple English labels (not translated).

**Expected behavior**
After changing the Dashboard language to a non-English language, all UI strings on `/dashboard/collections/new` should update to the selected language, including the “Add collection filter” button, its dropdown options, and any labels inside newly added filter cards.

**Actual behavior**
The “Add collection filter” button remains in English after switching language. The dropdown menu options are also in English, and selecting an option adds a filter card with multiple English-only labels.

**Screenshots/Videos**

Image

Image

**Error logs**
No errors observed in the UI.

* @vendure/core version: 3.5.4
* Nodejs version: v22.13.0
* Database (mysql/postgres etc): postgres
* Operating System (Windows/macOS/Linux): Windows 11
* Browser (if applicable): Brave
* Package manager (npm/yarn/pnpm): npm

**Configuration**
Not relevant

**Minimal reproduction**
Occurs on a standard Dashboard served via `DashboardPlugin`.

**Workaround**
None found yet. Reloading the page does not consistently resolve the untranslated strings on the New Collection page.

**Additional context**

* Happens consistently when following the steps above.
* Looks like certain strings used in the collection filter UI are not connected to the i18n system or are not reacting to locale changes.
* The issue is noticeable after navigating from `/dashboard/collections` to `/dashboard/collections/new` and then switching language.

grolmus:
## Summary

Fixes a Dashboard i18n regression where the **"Add collection filter"** button — and the equivalent buttons on promotion conditions/actions and the payment/shipping method selectors — stayed in English after switching the UI language.

## Root cause

The two shared selector components rendered the button/empty labels as:

```tsx
<Trans>{buttonText}</Trans>
```

`buttonText` is a runtime string prop. Lingui's `<Trans>` macro only extracts **static literals** at compile time, so wrapping a variable produces a placeholder message that is never translated — the raw English prop was rendered regardless of locale.

## Change

- `configurable-operation-multi-selector.tsx` / `configurable-operation-selector.tsx`: `buttonText` / `dropdownTitle` / `emptyText` prop types → `ReactNode`; render `{buttonText}` directly; default `emptyText` → `<Trans>No options found</Trans>`.
- The 7 call sites (collection filters, promotion conditions/actions, payment handler + eligibility checker, shipping calculator + eligibility checker) now pass static `<Trans>…</Trans>` elements, which extract correctly.
- Ran `lingui extract` and translated the 11 new message ids across all 26 locale catalogs.

## Test plan

**Automated**
- `tsc` (tsconfig.check.json) passes.
- `lingui extract` is idempotent → the `check-i18n-sync` CI gate stays green.

**Manual** (screenshots on the Linear issue / attached to this PR)
- Drove the live dashboard (`/collections/new`) via the e2e harness and switched UI language:
  - `es` → button renders **"Añadir filtro de colección"**
  - `de` → button renders **"Sammlungsfilter hinzufügen"**
  - asserted the old English label is absent under `es`.

## Notes / out of scope

- The dropdown **option names** and the labels **inside** an added filter card are server-sourced (`ConfigurableOperationDef.description` / `arg.label`) and are not addressed here — that is the separate configurable-operation server-translation gap.
- `product-multi-selector-input.tsx` has the same `<Trans>{buttonText}>` anti-pattern but in a template-literal-with-interpolation form; left as a follow-up.

Fixes #4410

Manual-verification screenshots (en/es/de) are attached on the Linear issue OSS-405.

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_promotions/components/promotion-actions-selector.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_promotions/components/promotion-actions-selector.tsx
@@ -1,6 +1,7 @@
 import { ConfigurableOperationMultiSelector } from '@/vdb/components/shared/configurable-operation-multi-selector.js';
 import { configurableOperationDefFragment } from '@/vdb/graphql/fragments.js';
 import { graphql } from '@/vdb/graphql/graphql.js';
+import { useLingui } from '@lingui/react/macro';
 import { ConfigurableOperationInput as ConfigurableOperationInputType } from '@vendure/common/lib/generated-types';
 
 export const promotionActionsDocument = graphql(
@@ -25,6 +26,7 @@ export function PromotionActionsSelector({
     onChange,
     onValidityChange,
 }: Readonly<PromotionActionsSelectorProps>) {
+    const { t } = useLingui();
     return (
         <ConfigurableOperationMultiSelector
             value={value}
@@ -32,8 +34,8 @@ export function PromotionActionsSelector({
             queryDocument={promotionActionsDocument}
             queryKey="promotionActions"
             dataPath="promotionActions"
-            buttonText="Add action"
-            dropdownTitle="Available Actions"
+            buttonText={t`Add action`}
+            dropdownTitle={t`Available Actions`}
             showEnhancedDropdown={true}
             onValidityChange={onValidityChange}
         />
--- a/packages/dashboard/src/app/routes/_authenticated/_promotions/components/promotion-conditions-selector.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_promotions/components/promotion-conditions-selector.tsx
@@ -1,6 +1,7 @@
 import { ConfigurableOperationMultiSelector } from '@/vdb/components/shared/configurable-operation-multi-selector.js';
 import { configurableOperationDefFragment } from '@/vdb/graphql/fragments.js';
 import { graphql } from '@/vdb/graphql/graphql.js';
+import { useLingui } from '@lingui/react/macro';
 import { ConfigurableOperationInput as ConfigurableOperationInputType } from '@vendure/common/lib/generated-types';
 
 export const promotionConditionsDocument = graphql(
@@ -25,6 +26,7 @@ export function PromotionConditionsSelector({
     onChange,
     onValidityChange,
 }: Readonly<PromotionConditionsSelectorProps>) {
+    const { t } = useLingui();
     return (
         <ConfigurableOperationMultiSelector
             value={value}
@@ -32,8 +34,8 @@ export function PromotionConditionsSelector({
             queryDocument={promotionConditionsDocument}
             queryKey="promotionConditions"
             dataPath="promotionConditions"
-            buttonText="Add condition"
-            dropdownTitle="Available Conditions"
+            buttonText={t`Add condition`}
+            dropdownTitle={t`Available Conditions`}
             showEnhancedDropdown={true}
             onValidityChange={onValidityChange}
         />
--- a/packages/dashboard/src/i18n/locales/ar.po
+++ b/packages/dashboard/src/i18n/locales/ar.po
@@ -1061,6 +1061,10 @@ msgstr "طلب مسودة"
 msgid "Scope"
 msgstr "النطاق"
 
+#: src/app/routes/_authenticated/_promotions/components/promotion-conditions-selector.tsx:37
+msgid "Add condition"
+msgstr "إضافة شرط"
+
 #: src/app/routes/_authenticated/_shipping-methods/components/test-single-method-result.tsx:58
 msgid "Shipping method is eligible for this order"
 msgstr "طريقة الشحن مؤهلة لهذا الطلب"
@@ -1289,8 +1293,6 @@ msgid "{cancellableCount, plural, one {Cancel {cancellableCount} job} other {Can
 msgstr ""
 
 #: src/lib/components/data-input/product-multi-selector-input.tsx:499
-#: src/lib/components/shared/configurable-operation-multi-selector.tsx:265
-#: src/lib/components/shared/configurable-operation-selector.tsx:140
 msgid "{buttonText}"
 msgstr "{buttonText}"
 
@@ -3475,6 +3477,10 @@ msgstr "إدارة العلامات"
 msgid "Items refunded"
 msgstr "العناصر المستردة"
 
+#: src/app/routes/_authenticated/_promotions/components/promotion-actions-selector.tsx:37
+msgid "Add action"
+msgstr "إضافة إجراء"
+
 #: src/app/routes/_authenticated/_products/components/create-product-options-dialog.tsx:325
 #~ msgid "Assign options to existing variant"
 #~ msgstr "تعيين خيارات إلى الشكل الموجود"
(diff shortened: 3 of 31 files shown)


grolmus:
01-en-add-collection-filter
02-es-add-collection-filter
03-de-add-collection-filter

biggamesmallworld (commented):
Reviewed before I spotted that `master` had already fixed part of this — see my follow-up comment for the full picture. Superseding this review body, and I've deleted the inline suggestions since they assumed the `ReactNode` shape.

Two points survive regardless of which shape you land:

- **The `@example` blocks show raw strings** (`configurable-operation-selector.tsx:69`, `configurable-operation-multi-selector.tsx:83-84,95`). Whatever the final API is, the examples should demonstrate the localized form — they're the copy-paste source for extension authors, and a raw string type-checks while silently never translating.
- **The prop doc comments say "Text to display"** (`configurable-operation-selector.tsx:32,34`, multi-selector `:34,36,38`). Worth a pass once the prop types settle.

Otherwise: root cause is correctly diagnosed, call sites are all covered, and the catalogs match `lingui extract`.

biggamesmallworld:
*(Edited: my original wording claimed master's fix "works by accident" via a failed catalog lookup. That's wrong — corrected below.)*

Heads up — [`dbc31ffb5`](https://github.com/vendurehq/vendure/commit/dbc31ffb5) (#4074) landed after you branched and fixed the same bug via the `t` macro, on 5 of your 7 call sites. That's the conflict source. Neither PR referenced the other.

**#4410 is genuinely fixed on master.** `<Trans>{buttonText}</Trans>` isn't broken — it extracts to an identity passthrough:

```po
#: src/lib/components/shared/configurable-operation-multi-selector.tsx:265
#: src/lib/components/shared/configurable-operation-selector.tsx:140
msgid "{buttonText}"
msgstr "{buttonText}"
```

It renders the prop verbatim in every locale. Where the caller passes ``t`Add collection filter` `` the string arrives pre-translated, so it works. The wrapper is a no-op, not a bug.

Still worth landing this, for three reasons:

1. **Promotions are actually broken.** `promotion-conditions-selector.tsx:35-36` and `promotion-actions-selector.tsx:35-36` pass raw strings into that no-op and render English forever. None of the 4 msgids exist in `en.po`.
2. **The wrapper is a decoy.** It looks like localization, does nothing, and mints junk msgids. There are four: `{buttonText}`, `{label}` (`data-table-bulk-action-item`), `{title}` (`base-widget`), `{operator}` (`human-readable-operator`). Both selectors are exported from `lib/index.ts`, so extension authors will read `<Trans>` and reasonably assume a plain string gets translated.
3. `emptyText` still defaults to an untranslated `'No options found'`.

**Suggested reshape** — rebase, then do only what master didn't:

- Delete the `<Trans>` wrapper in both shared components; render `{buttonText}` directly.
- Convert the 2 promotions selectors to `t`, matching master.
- Translate the `emptyText` default via `useLingui()`.

That lets you drop the `string` → `ReactNode` widening. Both macros only extract what's statically visible at the call site, so a prop like `buttonText` has to be translated *by the caller* — which makes `t` the idiomatic choice here, not a workaround. `<Trans>` earns its keep for static JSX content and rich text, not for a value arriving through a prop. The `ReactNode` element form would also work, but it widens a public API to buy nothing.

Diff shrinks to ~15 lines, and most `.po` conflicts evaporate: master already carries the 5 msgids, so you'd only add the 4 promotions ids + `No options found`.

I've deleted my inline suggestions above — they assumed the `ReactNode` shape.

biggamesmallworld:
Took you up on the offer — force-pushed a reshaped commit over `574055eea`. Your original is preserved as the co-author trailer, and in this PR's force-push history if you want it back.

**What changed vs. your version:**

- Dropped the `string` → `ReactNode` widening. Both selectors are exported from `lib/index.ts`, and with every caller passing a `t` string there was nothing to widen for.
- Kept master's `t` macro at the 5 call sites it already fixed, so no churn there.
- Converted the 2 promotions selectors (`Add condition` / `Add action` + dropdown titles) — the ones master missed, and the only user-visible bug left.
- Removed the no-op `<Trans>{buttonText}</Trans>` from both shared components.
- `emptyText` now falls back to ``t`No options found` `` at render instead of an untranslated default.
- Fixed the three `@example` blocks to show the `t` form, and the prop docs now say the value must arrive pre-translated.

**Result:** 4 source files, ~30 lines, +5 msgids across the catalogs. Conflicts resolved. `tsc` clean, `lingui extract` idempotent (so `check-i18n-sync` stays green), `lingui compile` passes. Verified no existing `msgstr` was clobbered — the only catalog deletions are the two stale source-refs on `{buttonText}`.

Retitled in the commit since it's no longer about collection filters — master's [`dbc31ffb5`](https://github.com/vendurehq/vendure/commit/dbc31ffb5) already fixed #4410, so this is `Relates to` rather than `Fixes`.

Still open as follow-ups, not touched here: `product-multi-selector-input.tsx:499` (same anti-pattern, template-literal form — it's why `msgid "{buttonText}"` survives in the catalog), the `{label}` / `{title}` / `{operator}` equivalents, and the orphaned `ko.po`.

Thanks for the original diagnosis — the root cause writeup is what made this quick.

biggamesmallworld (approved):

grolmus (commented):
Thanks @biggamesmallworld — this is a clean fix. `<Trans>{buttonText}</Trans>` was the classic lingui anti-pattern (the macro can't extract a runtime variable, so those button/dropdown strings were never actually translated). Moving the translation to the static call sites with the `t` macro, rendering the passed string plainly, and translating the internal `emptyText` fallback via `useLingui` is exactly the right approach. Good call documenting "must be pre-translated" on the `buttonText` prop too.

I checked it out and verified:
- ✅ Ran the exact CI check (`scripts/check-i18n-sync.sh`) → **i18n catalogs are in sync**, so `lingui extract` produces no drift.
- ✅ No leftover `<Trans>` references after the import swap in either shared selector — clean.
- ✅ All configurable-operation selector call sites are localized: the two promotion selectors here, plus the payment/shipping/collection selectors which were already migrated on `master` — so coverage is complete, not just promotions.
- ✅ `dropdownTitle` is rendered plainly, consistent with it now being passed pre-translated.

No regression for any caller (the old `<Trans>{var}</Trans>` didn't translate either), and the anti-pattern is gone repo-wide for these components. LGTM 🚀

michaelbromley (approved):

PR #4915, opened 2026-07-02
fix(core): Localize default collection filter names

grolmus:
## Summary

Localizes the default collection-filter names shown in the Dashboard. The collection-filter `ConfigurableOperationDef` labels/descriptions in `@vendure/core` were English-only, so the "Add collection filter" dropdown options and the arg labels inside a filter card stayed in English even when the Dashboard UI language was switched.

This is the **server-side complement to OSS-405** (#4410): that PR fixed the dashboard-owned strings, but the filter names/labels are served by core via the `collectionFilters` query and resolved server-side by `languageCode`, so they can only be translated here.

## Root cause

In `packages/core/src/config/catalog/default-collection-filters.ts`, the user-facing `label`/`description` fields are `LocalizedString[]` that only defined `LanguageCode.en`. When the dashboard requests `collectionFilters` with e.g. `languageCode: de`, the resolver falls back to the English value.

## Change

Expanded the **9 short** user-facing strings to include the **26 locales the Dashboard ships** (en + de, es, fr, it, nl, pt_BR, pt_PT, cs, pl, ro, hr, hu, nb, sv, tr, uz, ru, bg, uk, ar, fa, he, ja, ne, zh_Hans, zh_Hant):

- Filter descriptions: "Filter by facet values", "Filter by product variant name", "Manually select product variants", "Manually select products"
- Arg labels: "Combination mode", "Facet values", "Contains any", "Product variants", "Products"

The `apply()` query-builder logic is untouched.

## Scope notes

- **Two long arg *tooltip* descriptions** (`combineWithAnd`, `containsAny`) are intentionally left English-only: translated, they exceed the eslint `max-len` (170) in verbose languages after prettier reformatting, which would require a `// eslint-disable` per line. Tracked as a follow-up. (Result: a `de` user sees a German label with an English hover tooltip on those two args.)
- Collection filters are intentionally the **only** configurable operation localized inline here. Promotions/shipping/payment configurable-ops remain English-only; a broader extraction of all such strings into lingui catalogs is proposed separately in **OSS-560** rather than inlining everywhere. This PR is the targeted fix for the collection-filter UI.

## Test plan

- `tsc` on `default-collection-filters.ts`: no errors; all 26 codes are valid `LanguageCode` members.
- eslint clean; prettier reports the file already conforms.
- No behavioural/query logic changed — the diff is data (`LocalizedString` entries) only.

Relates to OSS-449. Complements #4410 (OSS-405).

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4915"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785592046&installation_id=128408429&pr_number=4915&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4915&signature=78ed330f73e102b684fc775a15a36aa72ece04866da08b66924066fdd318786f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/core/src/config/catalog/default-collection-filters.ts
```diff
@@ -51,11 +79,67 @@ export const facetValueCollectionFilter = new CollectionFilter({
...
+                { languageCode: LanguageCode.nb, value: 'Fasettverdier' },
+                { languageCode: LanguageCode.sv, value: 'Facettvärden' },
+                { languageCode: LanguageCode.tr, value: 'Faset değerleri' },
+                { languageCode: LanguageCode.uz, value: 'Faset qiymatlari' },
+                { languageCode: LanguageCode.ru, value: 'Значения фасетов' },
+                { languageCode: LanguageCode.bg, value: 'Стойности на фасети' },
+                { languageCode: LanguageCode.uk, value: 'Значення фасетів' },
+                { languageCode: LanguageCode.ar, value: 'قيم الأوجه' },
+                { languageCode: LanguageCode.fa, value: 'مقادیر وجه' },
+                { languageCode: LanguageCode.he, value: 'ערכי פאסטה' },
```
biggamesmallworld, line 105: ~~Wrong Hebrew word for facet (`פאסטה` = pasta).~~

**Superseded** — see the implementation-plan comment below. These translations should move to the dashboard `.po` files rather than being fixed inline, so this becomes moot.

packages/core/src/config/catalog/default-collection-filters.ts
```diff
@@ -68,7 +152,35 @@ export const facetValueCollectionFilter = new CollectionFilter({
...
+        { languageCode: LanguageCode.nb, value: 'Filtrer etter fasettverdier' },
+        { languageCode: LanguageCode.sv, value: 'Filtrera efter facettvärden' },
+        { languageCode: LanguageCode.tr, value: 'Faset değerlerine göre filtrele' },
+        { languageCode: LanguageCode.uz, value: 'Faset qiymatlari boʻyicha filtrlash' },
+        { languageCode: LanguageCode.ru, value: 'Фильтр по значениям фасетов' },
+        { languageCode: LanguageCode.bg, value: 'Филтриране по стойности на фасети' },
+        { languageCode: LanguageCode.uk, value: 'Фільтрувати за значеннями фасетів' },
+        { languageCode: LanguageCode.ar, value: 'التصفية حسب قيم الأوجه' },
+        { languageCode: LanguageCode.fa, value: 'فیلتر بر اساس مقادیر وجه' },
+        { languageCode: LanguageCode.he, value: 'סינון לפי ערכי פאסטה' },
```
biggamesmallworld, line 178: ~~Wrong Hebrew word for facet (`פאסטה` = pasta).~~

**Superseded** — see the implementation-plan comment below. These translations should move to the dashboard `.po` files rather than being fixed inline, so this becomes moot.

biggamesmallworld (commented):
Mostly looks good to me. Left 4 inline notes: 2 committable Hebrew fixes, 2 untranslated tooltips.

Two non-inline points:

1. **Locale-parity test.** This list is hand-maintained to mirror `lingui.config.js` with no automated linkage — a future locale added there silently drifts to the English fallback here. A small test asserting each filter's `LocalizedStringArray` covers the configured locale set would catch it.
2. **Scope (optional).** Collection filters are now localized, but payment/shipping/promotion `ConfigArgDef`s aren't — the "Add filter" dropdown will be translated while "Add promotion condition" next to it isn't. Fine as scoping, worth a line in the PR body.

packages/core/src/config/catalog/default-collection-filters.ts
```diff
@@ -26,7 +26,35 @@ export function randomSuffix(prefix: string) {
...
+        { languageCode: LanguageCode.uk, value: 'Режим комбінування' },
+        { languageCode: LanguageCode.ar, value: 'وضع الدمج' },
+        { languageCode: LanguageCode.fa, value: 'حالت ترکیب' },
+        { languageCode: LanguageCode.he, value: 'מצב שילוב' },
+        { languageCode: LanguageCode.ja, value: '組み合わせモード' },
+        { languageCode: LanguageCode.ne, value: 'संयोजन मोड' },
+        { languageCode: LanguageCode.zh_Hans, value: '组合模式' },
+        { languageCode: LanguageCode.zh_Hant, value: '組合模式' },
+    ],
     description: [
```
biggamesmallworld, line 58: ~~Description left English-only while its label is localized.~~

**Superseded** — see the implementation-plan comment below. All these strings should be translated in the dashboard `.po` files, not inline in core config.

packages/core/src/config/catalog/default-collection-filters.ts
```diff
@@ -51,11 +79,67 @@ export const facetValueCollectionFilter = new CollectionFilter({
...
+                { languageCode: LanguageCode.uk, value: 'Містить будь-яке' },
+                { languageCode: LanguageCode.ar, value: 'يحتوي على أي' },
+                { languageCode: LanguageCode.fa, value: 'شامل هر یک' },
+                { languageCode: LanguageCode.he, value: 'מכיל כל אחד' },
+                { languageCode: LanguageCode.ja, value: 'いずれかを含む' },
+                { languageCode: LanguageCode.ne, value: 'कुनै पनि समावेश छ' },
+                { languageCode: LanguageCode.zh_Hans, value: '包含任一' },
+                { languageCode: LanguageCode.zh_Hant, value: '包含任一' },
+            ],
             description: [
```
biggamesmallworld, line 143: ~~Description left English-only while its label is localized.~~

**Superseded** — see the implementation-plan comment below. All these strings should be translated in the dashboard `.po` files, not inline in core config.

michaelbromley:
Hi, while this does fix the specific reported issue, it is not the solution I want.

The `LocalizedStringArray` data type that we use for labels for all ConfigurableOperations (collection filters, promotion actions/conditions, shipping calculators, payment method handlers etc) follow this same pattern.

If we start putting _all_ language translations into TS code, this is going to be not only a _lot_ of TS code in core which contains just translation data, and it will be bad for maintainability.

At the same time, it _is_ something that needs to be solved, the original issue is totally correct.

So I see 2 potential solutions:

1. we solve this at the Dashboard level by somehow extracting the english language `LocalizedStringArray` strings and incorporating that into the Dashboard i18n pipeline
2. we introduce a way where we can, at runtime, dynamically create the missing `LocalizedStringArray` entries on the server side based on a set of i18n files that contain _only_ translations.

The second is my preferred method. Specifics of how this would work need to be designed and proven.

biggamesmallworld:
## Suggested approach: translate on the dashboard side, not inline in core

After digging into how these strings flow, I think this PR is solving the problem in the wrong layer. Adding a `LocalizedStringArray` per `ConfigArgDef` means every future language has to be hand-maintained in core TS, and — as noted in review — if we do it for one label we owe it for **every** `LocalizedString` in core. That doesn't scale.

The dashboard already has the right mechanism for exactly this: `useDynamicTranslations` (`packages/dashboard/src/lib/hooks/use-dynamic-translations.ts`) translates runtime server strings (order states, field names, refund reasons) via `.po` catalogs with a safe fallback. Config arg labels/descriptions are the same class of string. This keeps a single English source of truth in core and moves translations into the existing `.po` / Crowdin workflow — adding a language becomes one `.po` file, zero core edits.

### Step-by-step

**1. Core — revert to English-only (`packages/core/src/config/catalog/default-collection-filters.ts`)**
Strip the added arrays back to `label: [{ languageCode: LanguageCode.en, value: '...' }]`. The server keeps sending a single English string via `localizeString`.

**2. Dashboard — add a translator (`src/lib/hooks/use-dynamic-translations.ts`)**
Add one method reusing the existing `translateOrFallback`:
```ts
const getTranslatedConfigArgString = (value: string) =>
    translateOrFallback(value, () => value);
```
English string = message id; unknown strings (e.g. plugin-provided) pass through unchanged.

**3. Dashboard — translate at the render points**
- `src/lib/components/shared/configurable-operation-input.tsx`
  - Arg label (~line 113): `{arg.label || arg.name}` → `{arg.label ? getTranslatedConfigArgString(arg.label) : arg.name}`
  - `interpolateDescription` (~line 151): translate `operation.description` before interpolation.
- `src/lib/components/shared/configurable-operation-selector.tsx` (~line 150): `{operation.description}` → wrap with `getTranslatedConfigArgString(...)`.

**4. Translations — reuse the work already in this PR (`src/i18n/locales/*.po`)**
The 26-language strings already written here port directly in as the `msgstr` values (English source = `msgid`). Register the source strings for extraction (a small `msg` registry file, or hand-add keys as the existing `orderState.*` entries were), run `lingui extract`, then fill from the translations already in this PR.

### Why this is safe
- Follows the dashboard **display language** (UI locale) — correct for admin chrome; the inline/server approach followed the *content* language, which is arguably wrong for these.
- The `translateOrFallback` guard means plugin/custom config args with their own server-side `LocalizedStringArray` still work — server localizes them, the dashboard passes the unknown string through. **Non-breaking.**
- Legacy Angular admin-ui is unaffected (separate i18n stack, deprecated in favour of the dashboard).

Happy to pair on this or open a follow-up if it's easier. None of the translation effort here is wasted — it just moves into the `.po` files.

PR #4916, opened 2026-07-03
feat(dashboard): Send only changed fields on detail page updates

grolmus:
## Summary

A detail page can now opt in to sending **only the fields the user actually changed** on update, via `sendOnlyChangedFields: true` on `useDetailPage`. The default is unchanged: the full form payload is still sent.

This closes a silent data-loss vector on the pages that opt in. Previously a field left untouched by admin A was resubmitted at its stale page-load value and could overwrite a concurrent change made by admin B or an API integration, with no error and a success toast.

Opted in here: **Product, Collection and ProductVariant**. Those three update paths were confirmed patch-style by review (`patchEntity`, `TranslatableSaver`, `TranslationDiffer` and the relevant `beforeSave` callbacks), and each is covered by an e2e payload assertion.

Design was signed off by @michaelbromley on OSS-567 (targeting `minor`). This is Tier 1 of his framing (don't send untouched fields). Tier 2 (touched replace-set element conflicts, beyond the merged stock fix) and Tier 3 (optimistic concurrency) remain separate.

## Root cause

`useGeneratedForm` mapped the whole entity into the form and called `onSubmit` with the entire payload on every save; `useDetailPage` forwarded it to the update mutation. For replace-semantics fields (`facetValueIds`, `assetIds`, scalar stock config, `channelIds`, prices, promotion conditions/actions, and so on) the stale value replaced whatever was there.

## Change

- **`form-engine/utils.ts`**: `deepEqual` (primitives incl. NaN, Date, arrays, plain objects), `getChangedTopLevelFields(submitted, baseline, fields)` returning the top-level keys that differ from the baseline plus any input field that is non-nullable (`FieldInfo.nullable === false`, e.g. `id`, `UpdateShippingMethodInput.translations`) which is always sent, and `pruneToChangedFields(values, changedFields, sendAll)`.
- **`useGeneratedForm`**: computes `changedFields` by deep-comparing the raw submitted values against the form baseline (the `values` memo) and passes it via a new, backward-compatible 2nd `meta` arg to `onSubmit`. RHF `dirtyFields` is deliberately **not** used: array-item removal reads as clean and would drop the user's own deletion, and `assetIds`/`featuredAssetId` are written via `setValue` with no Controller.
- **`useDetailPage`**: adds `sendOnlyChangedFields?: boolean`, default `false`. When set, the update path prunes the payload to `changedFields` before `transformUpdateInput`. Create path is unchanged. Composes with the existing `getChangedStockLevels` transform (#4834).
- **Opt-ins**: `products_.$id.tsx`, `collections_.$id.tsx`, `product-variants_.$id.tsx`.
- No change to `transformUpdateInput`'s signature (avoids the generic-inference hazard from the OSS-554 fix).

Whole-key granularity: a changed top-level key is sent in full (the entire `translations` array, `facetValueIds`, `customFields` object), preserving the "full translation objects per language" invariant.

### Invariant guard

`changedFields` is computed on the raw values but applied by the consumer after `convertEmptyStringsToNull`, `removeEmptyIdFields`, `stripUntouchedTranslations` and `stripNullNullableFields` have run. That is only correct while none of those transforms renames or drops a changed top-level key. `useGeneratedForm` now throws in dev and test builds if a changed key is missing from the processed payload (empty-string IDs excluded, since `removeEmptyIdFields` legitimately removes those). Production is unaffected.

## Why opt-in rather than the default

The first version of this PR flipped the default for every detail page. Review found that "core update inputs are patch-style" does not hold universally: `updatePromotion` recomputes `priorityScore` from `input.conditions`/`input.actions` unconditionally, so a pruned `{ id, name }` payload would write `priorityScore: 0` and silently reorder promotion application.

Opt-in collapses the audit surface from "every update mutation the dashboard can reach" to the three confirmed pages. The precondition is documented on `sendOnlyChangedFields`: the mutation must be patch-style, and no persisted value may be *derived* from the now-partial input. `updatePromotion` is named there as the counter-example.

Not fixed here: the `updatePromotion` `priorityScore` derivation itself. It is a pre-existing resolver issue, unreachable from the dashboard while Promotion keeps the full-payload default. It should be fixed separately before Promotion is opted in.

## Behaviour change / extensions

None by default. Third-party dashboard extensions built on `useDetailPage` keep the existing full-payload behaviour unless they set `sendOnlyChangedFields: true`.

## Test plan

**Unit** (`form-engine/utils.spec.ts`, 69 pass): `deepEqual` (NaN, Date, nested arrays/objects); `getChangedTopLevelFields` (changed scalar kept, untouched pruned, array add and **remove**, key deletion, nested translation/customFields, non-nullable always kept, undefined-baseline to value transitions, missing non-nullable field doesn't throw); `pruneToChangedFields` (prunes, honours the send-all path, no-ops without a change set).

**E2E** (payload assertions, the strongest proof). Each test creates its own entity in `beforeAll` and deletes it in `afterAll`, so the shared seed data is not mutated:
- `catalog/product-variants.spec.ts`: edit only the SKU, update input is exactly `{ id, sku }` (facetValueIds/assetIds/trackInventory/translations/price omitted).
- `catalog/products.spec.ts`: edit only the name, update input is exactly `{ id, translations }`; plus an anti-clobber test that changes `facetValueIds` out of band via the API, edits an unrelated field through the UI, and asserts the out-of-band change survives.
- `catalog/collections.spec.ts`: structurally different page (configurable-operation `filters` array, no update transform), same assertion.

Typecheck clean; existing catalog suites green.

Relates to OSS-567.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4916"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785666069&installation_id=128408429&pr_number=4916&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4916&signature=0be20c32084e833164b42ecbe9f69392f362ba0e3ec03f16d5438b7a57b0fb6d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end tests for collections, products, and product variants to verify that update requests include only the fields modified by the user.

* **Chores**
  * Enhanced the form engine to optimize update payloads by sending only changed fields (with an option to revert to sending all fields if needed).


grolmus:
## CI fix: `dashboard e2e` — test pollution of shared seed data

**Symptom.** `dashboard e2e` (shard 2) failed on `translation-placeholders.spec.ts`:

```
expect(locator).toHaveValue(expected)
Expected: "Laptop"
Received: "Laptop OSS567 1784111122452"
```

**Root cause.** The new products test added in this PR (`product update sends only changed fields (OSS-567)`) navigated to the shared seed **Laptop** product, renamed it to `Laptop OSS567 <timestamp>` and saved. Dashboard e2e tests run in parallel against a single shared server DB within a shard, so this permanently mutated the Laptop product — and `translation-placeholders.spec.ts` asserts the Laptop name stays `"Laptop"`. Classic cross-spec pollution.

Notably, this PR's **collections** OSS-567 test already does it the right way: it creates a dedicated collection via `VendureAdminClient` in `beforeAll` and deletes it in `afterAll`. The products test just didn't follow that pattern.

**Fix.** Bring the products test in line with the collections one:
- Create a **dedicated** product via `VendureAdminClient` in `beforeAll` (seeded with a non-empty `facetValueIds`), edit *that* product by id, and delete it in `afterAll`.
- Seeding a non-empty `facetValueIds` also strengthens the assertion: it now proves the engine omits an *unchanged non-empty* replace-array from the update payload (the actual clobber-prevention behaviour), not merely that an empty one is absent.

No change to product/framework code — this is purely test isolation.

**Verification.** Ran `products.spec.ts` + `translation-placeholders.spec.ts` together locally (same shared-DB scenario, parallel workers): **21/21 pass**, including the OSS-567 test and the previously-failing translation-placeholders test. Before the fix, translation-placeholders failed with the polluted name.

### Note — unrelated flake

The `e2e (mariadb)` failure is `default-scheduler-plugin.e2e-spec.ts:155` (`expected "spy" to be called 2 times, but got 1 times`) — a timing-sensitive scheduler test in core, unrelated to this PR's changes (1968/1972 passed). Should clear on re-run.

biggamesmallworld (commented):
Approach is right and matches the OSS-567 sign-off. The non-nullable carve-out is narrower than it looks: `UpdateShippingMethodInput.translations` is the only non-`id` required top-level field in the schema.

Before merge:

1. **Rebase.** Conflicts with `minor` in `utils.ts`, `utils.spec.ts`, `use-detail-page.ts`, `collections.spec.ts`, all from the #4962 overlap.
2. **`product-variants.spec.ts` mutates seed data with no cleanup.** It renames a Laptop variant's SKU permanently. Same bug the products test in this PR already hit; fix it the same way (dedicated product in `beforeAll`, delete in `afterAll`).
3. **`sendAllFieldsOnUpdate` has no test and no caller.** It's the escape hatch extensions will reach for when this breaks them. Needs one case.
4. **No test proves the bug is fixed.** All three e2e tests assert payload shape. None does the real thing: change `facetValueIds` out of band, save an unrelated field, assert it survived. ~15 lines.

Minor: extract the 5-line prune into a shared util so the next `useGeneratedForm` consumer reuses it; three copies of a form-field locator when `formItem`/`fieldByLabel` already exist; three copies of the mutation matcher; `deepEqual` is generically named but throws on circular input.

Scope: OSS-567 was Tier 1 + Tier 2, this is Tier 1 only. @michaelbromley second PR or rescope? The `T3: Systemic` label also looks stale now the design is signed off.

Inline suggestions below are committable.

packages/dashboard/e2e/tests/catalog/products.spec.ts
```diff
@@ -581,3 +581,92 @@ test.describe('Manage variants inline editing', () => {
...
+            { timeout: 15_000 },
+        );
+        await page.getByRole('button', { name: 'Update' }).click();
+        const input = (await updateRequest).postDataJSON()?.variables?.input;
+
+        expect(input).toBeTruthy();
+        expect(Object.keys(input).sort()).toEqual(['id', 'translations']);
+        expect(input.facetValueIds).toBeUndefined();
+        expect(input.assetIds).toBeUndefined();
+        expect(input.enabled).toBeUndefined();
```
biggamesmallworld: The exhaustive `Object.keys(input).sort()` assertion above already proves all three of these. Extra assertions after an exhaustive one read as thoroughness but can't fail independently.

```suggestion
```

packages/dashboard/e2e/tests/catalog/product-variants.spec.ts
```diff
@@ -982,3 +982,65 @@ test.describe('variant with no option groups', () => {
...
+        // replace-semantics field (facetValueIds, assetIds, trackInventory, translations,
+        // price, …) must be omitted.
+        expect(input).toBeTruthy();
+        expect(Object.keys(input).sort()).toEqual(['id', 'sku']);
+        expect(input.sku).toBe(newSku);
+        expect(input.facetValueIds).toBeUndefined();
+        expect(input.assetIds).toBeUndefined();
+        expect(input.trackInventory).toBeUndefined();
+        expect(input.translations).toBeUndefined();
+        expect(input.price).toBeUndefined();
```
biggamesmallworld: Same here: `Object.keys(input).sort()` on line 1030 already covers all five. `expect(input.sku).toBe(newSku)` earns its place, these don't.

```suggestion
```

packages/dashboard/e2e/tests/catalog/collections.spec.ts
```diff
@@ -608,3 +608,92 @@ test.describe('Collection tree toggles, search subtitles & detail breadcrumb', (
...
+            req => req.method() === 'POST' && (req.postData() ?? '').includes('mutation UpdateCollection('),
+            { timeout: 15_000 },
+        );
+        await dp.updateButton.click();
+        const input = (await updateRequest).postDataJSON()?.variables?.input;
+
+        expect(input).toBeTruthy();
+        expect(Object.keys(input).sort()).toEqual(['id', 'translations']);
+        expect(input.filters).toBeUndefined();
+        expect(input.inheritFilters).toBeUndefined();
```
biggamesmallworld: Covered by the `Object.keys(input).sort()` assertion above.

```suggestion
```

packages/dashboard/src/lib/framework/form-engine/utils.ts
```diff
@@ -203,6 +203,88 @@ export function stripNullNullableFields<T extends Record<string, any>>(values: T
...
+ *
+ * This lets the form engine submit only the fields the user actually changed,
+ * so that a stale page-load value for an untouched field cannot silently
+ * overwrite a concurrent change made by another admin or via the API. The
+ * union of keys is iterated so that a field removed entirely from the submitted
+ * values (e.g. a cleared array) is still detected as changed.
+ *
+ * Comparison must be done on the raw (pre-processed) form values so that
+ * transformations applied before submission (id-stripping, empty-string→null)
+ * do not register as spurious changes.
```
biggamesmallworld: This is the load-bearing part and it's worth spelling out the consequence. The diff runs on `rawValues` but the prune is applied to `filteredValues`, after `removeEmptyIdFields`, `convertEmptyStringsToNull` and `removeReadonlyAndLocalizedCustomFields`. That's correct today because none of them renames a top-level key, but if one ever did, the field would be silently dropped from the update rather than erroring. Silent drop is the worst failure mode here, so the invariant should be written down next to the function that depends on it.

```suggestion
 * Comparison must be done on the raw (pre-processed) form values so that
 * transformations applied before submission (id-stripping, empty-string→null)
 * do not register as spurious changes. This makes it an invariant that those
 * pre-submission transforms must never rename or add a top-level key: the
 * returned set is matched against the processed payload, so a renamed key
 * would be silently dropped from the update instead of raising an error.
```

packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx
```diff
@@ -13,11 +13,29 @@ import {
...
     transformRelationFields,
 } from './utils.js';
 
+/**
+ * @description
+ * Metadata passed as the second argument to a {@link GeneratedFormOptions} `onSubmit`
+ * handler, describing the outcome of the form submission.
+ *
+ * @since 3.8.0
+ */
```
biggamesmallworld, line 31: "the outcome of the form submission" doesn't say anything; it describes what changed. Also missing `@docsCategory`/`@docsPage`, so this renders as a dead type name in the generated `useGeneratedForm` docs.

```suggestion
/**
 * @description
 * Metadata passed as the second argument to a {@link GeneratedFormOptions} `onSubmit`
 * handler, describing which fields the user changed relative to the loaded entity.
 *
 * @docsCategory detail-views
 * @docsPage useGeneratedForm
 * @since 3.8.0
 */
```

packages/dashboard/e2e/tests/catalog/collections.spec.ts
```diff
@@ -608,3 +608,92 @@ test.describe('Collection tree toggles, search subtitles & detail breadcrumb', (
         await expect(page.getByRole('textbox', { name: /Search variants/i })).toBeVisible();
     });
 });
+
+
```
biggamesmallworld, line 697: Stray double blank line, prettier will flag this in CI.

michaelbromley (dismissed):
Reviewed the payload-pruning change. The mechanism itself is well built, but there is one blocker: the core claim that "core update inputs are patch-style by design, so this is safe" does not hold for `Promotion`, and this will silently corrupt data on merge.

## Must fix: `updatePromotion` will zero out `priorityScore` on most promotion edits

In `packages/core/src/service/services/promotion.service.ts`, the `beforeSave` callback guards the field assignments but not the derived score:

```ts
beforeSave: async p => {
    p.priorityScore = this.calculatePriorityScore(input);   // unconditional
    if (input.conditions) { p.conditions = ...; }           // guarded
    if (input.actions)    { p.actions    = ...; }           // guarded
},
```

`calculatePriorityScore` treats absent `conditions`/`actions` as empty arrays and returns `0`. Both fields are nullable on `UpdatePromotionInput`, and the Promotion detail page (`packages/dashboard/src/app/routes/_authenticated/_promotions/promotions_.$id.tsx`) uses plain `useDetailPage` with no `sendAllFieldsOnUpdate` override.

So with this change, editing only the promotion's name sends `{ id, name }`, the conditions and actions are correctly left alone on the entity, but `priorityScore` is recomputed from nothing and written as `0`. That field drives `orderBy('promotion.priorityScore', 'ASC')` in `promotion.service.ts`, which determines the order promotions are applied to an order. The user sees a success toast and no error.

This is the same class of bug the PR is trying to eliminate. Guarding the assignments is not enough — any value *derived* from the input needs the same guard, and derived values are easy to miss because they do not look like field writes.

To resolve:

1. Only recompute `priorityScore` when `input.conditions` or `input.actions` are actually present.
2. Add an e2e test asserting `priorityScore` survives an edit that does not touch conditions or actions. There is currently no coverage at all for the Promotion detail page, which is why this went unnoticed — the tests cover collections, products and variants only.
3. Audit the remaining update resolvers reachable from a `useDetailPage`, not just the four services listed in the test plan. This change flips the default for every detail page in the dashboard at once, so the audit surface is every core update mutation the dashboard can reach.

On point 3, the safer rollout would be to make the pruning opt-in per page first and flip the default once the resolvers are confirmed clean. The `sendAllFieldsOnUpdate` escape hatch already exists for exactly this.

## Should fix: the changed-fields invariant is enforced only by a comment

`getChangedTopLevelFields` is computed in `use-generated-form.tsx` against the raw values, deliberately before `convertEmptyStringsToNull`, `removeEmptyIdFields`, `stripUntouchedTranslations` and `stripNullNullableFields` run. The resulting set is then applied three transforms later, in `use-detail-page.ts`, against the processed payload.

That only works as long as none of those transforms renames or adds a top-level key. I traced all five and none of them currently does, so the code is correct today. But the thing preventing a stale field from being silently dropped is a docblock spanning two files, with no assertion or test that would catch a future violation. Worth making structural — for example, diff the changed-field set against the final payload's key set in dev and test builds and throw if a changed key has vanished.

## Nitpick

`product-variants.spec.ts` (around lines 99-134) and `products.spec.ts` (around lines 247-319) contain pure whitespace and line-wrap changes on code this PR does not otherwise touch. Harmless, but it makes the behavioural changes harder to spot.

## What is good here

- `deepEqual` is a tight, correct implementation (NaN, Date, array and object aware), sensibly hand-rolled rather than pulling in a dependency for one narrow use, and properly unit tested.
- The e2e tests are the real thing: they intercept the actual network request and assert `Object.keys(input).sort()` equals exactly `['id', 'sku']`. The "does not clobber a concurrent change to an untouched field" test in `products.spec.ts` mutates out of band via the API, edits an unrelated field through the UI, and confirms the out-of-band change survives. That is a genuinely good regression test.
- `sendAllFieldsOnUpdate` is a real, discoverable escape hatch for third-party extensions, and every new symbol is exported through `packages/dashboard/src/lib/index.ts`.
- I independently verified the patch-style claim for the Product, Collection, ProductVariant, Asset and ShippingMethod update paths by reading `patchEntity`, `TranslatableSaver`, `TranslationDiffer` and the relevant `beforeSave` callbacks. That part of the design is sound.

grolmus:
Thanks for the thorough review, Michael — the `priorityScore` catch is exactly the kind of resolver-side assumption the flip-default couldn't safely make. I've reworked the PR to your suggested rollout rather than trying to audit every resolver:

**Pruning is now opt-in per detail page.**

- The default is back to sending the full form payload, which is safe for *any* update resolver — including `updatePromotion` and any plugin-defined resolver we can't audit. So the `priorityScore` corruption can no longer occur: the Promotion detail page keeps the full-payload default and is untouched.
- A page opts into pruning with `sendOnlyChangedFields: true` (renamed from `sendAllFieldsOnUpdate`, since the polarity flipped). I've opted in only **Product, Collection and ProductVariant** — the three whose update mutations you confirmed are patch-style and which the e2e payload assertions cover. That collapses the audit surface from "every update mutation the dashboard can reach" down to those three, each backed by a test.
- The doc on `sendOnlyChangedFields` spells out the precondition (patch-style, no value *derived* from the partial input) and names `updatePromotion` as the counter-example, so the next person opting a page in has the trap written down.

**Invariant (your Should):** made structural. The form engine now throws in dev/test builds if a pre-submission transform drops a changed top-level key (empty-string IDs excluded, since `removeEmptyIdFields` legitimately removes those), so the changed-field set can no longer silently omit a real change. Production is unaffected.

**Nitpick:** reverted — the two catalog specs now diff against `minor` as only their new OSS-567 `describe` block; the incidental prettier reformatting of untouched lines is gone.

One thing I did **not** do: fix the `updatePromotion` `priorityScore` derivation itself. Since Promotion no longer prunes, the corruption can't be reached from the dashboard, and the underlying "recompute from partial input" quirk is a pre-existing resolver issue that predates this PR — happy to open a separate one for it if you'd like Promotion opted in later.

Verified locally: typecheck clean, unit 69/69, and the OSS-567 e2e (product/collection/variant payload assertions + the anti-clobber test + the translation-omission test) all green with the opt-in flags.

biggamesmallworld (approved):
Approving.

The `priorityScore` blocker is resolved by construction: pruning is opt-in, default off, Promotion keeps the full payload. The three opt-ins were already confirmed patch-style and each has a payload assertion. Seed pollution, the shared prune util and the anti-clobber test are all done.

I rewrote the PR description, it still described the flipped-default version.

Before merge: rebase (73 behind `minor`) and re-run CI. The `unit tests (24.x)` failure is a `better_sqlite3` crash in core, unrelated to a dashboard diff.

Follow-ups, none blocking:

- `updatePromotion` still recomputes `priorityScore` from partial input. Needs its own issue, and fixing before Promotion ever opts in.
- Breaking to change after release, cheap now: `export *` in `lib/index.ts` makes `deepEqual` public API, and `transformUpdateInput` is typed as receiving the complete input when an opted-in page hands it a partial.
- The dev-build guard in `use-generated-form.tsx` cannot fire, and it checks for dropped keys when the invariant in `utils.ts` is about added ones. A unit test over the transforms would enforce it properly.
- `pruneToChangedFields` doc says `sendAll` defaults to `true`. It defaults to `false`, and the call site passes `!sendOnlyChangedFields`.
- Nothing covers the default path.

Issue #2608, opened 2023-12-29
Multiple Cancellations on a same line item not allowed from admin UI


shaheerrs:
** Multiple Cancellations on a same line item not allowed from admin UI  **

**To Reproduce**
1. Go to 'https://demo.vendure.io/'
2. Create and order with 2 items of 3 quantity each
3. Add payment to order
4. Use Refund or cancel 2 quantity from 1st item and 1 quantity from second item.
5. Trigger refund
6. Now try to use Refund or cancel option to cancel the one quantity remaining in the first item. (_you can see the option is disabled_)

**Expected behavior**
The Customer support should be able to cancel multiple times on line item until it reaches 0.

**Environment (please complete the following information):**
 - @vendure/core version: Vendure2.1.0
 - Nodejs version - Node 18
 - Database (mysql/postgres etc): Postgres

**Additional context**
In the above mentioned scenario see the item where I cannot select refund / cancel to proceed with full item cancellation from Admin UI
<img width="1155" alt="Screenshot 2023-12-29 at 7 45 23 AM" src="https://github.com/vendure-ecommerce/vendure/assets/45445534/b200f99f-5a78-4c39-a554-4cf50d64c163">

shaheerrs:
@michaelbromley  I was debugging the code and it seems the following code is the root cause of the issue
`lineCanBeRefundedOrCancelled(line) {
        const refundedCount = this.order.payments
            ?.reduce((all, payment) => [...all, ...payment.refunds], [])
            .filter(refund => refund.state !== 'Failed')
            .reduce((all, refund) => [...all, ...refund.lines], [])
            .filter(refundLine => refundLine.orderLineId === line.id)
            .reduce((sum, refundLine) => sum + refundLine.quantity, 0) ?? 0;
        return refundedCount < **line.quantity**;`

Here the line.quantity will always get reduced with cancellation, so the check is always against the remaning quantity in the order than the order placed quantity.

michaelbromley:
Thanks for the report & the advice on the location of the bug. You are correct! This will be fixed in the next patch.

PR #4917, opened 2026-07-03
fix(dashboard): Reflect modified line quantities in refund dialog

### LINKED ISSUE #4728: Refund UI does not reflect quantity/items changed during order modification
**Describe the bug**
When an order enters the `Modifying` state and the quantity of an existing
item is increased (e.g. from 1 to 2), or a new item is added, the refund
interface does not reflect the updated quantities or items. It continues to
show only the original items and quantities, preventing the admin from
issuing an accurate refund.

This bug is also reproducible on the official Vendure demo at
https://demo.vendure.io/admin/ (v3.6.1).

**To Reproduce**
1. Create an order with one item (quantity 1) and pay (order enters
   `PaymentSettled`).
2. Transition the order to `Modifying`.
3. Increase the quantity of the existing item by 1 (or add a new item).
4. Request and settle the additional payment.
5. Cancel the order.
6. Open the refund dialog for the order.
7. Observe that the item quantity still shows the original value (1 instead
   of 2), and any newly added items are missing.

This can also be verified on the official demo at
https://demo.vendure.io/admin/.

**Expected behavior**
The refund interface should display all items with their current quantities,
including any increases made during order modification, so that the admin
can refund the full amount actually paid by the customer.

**Actual behavior**
Only the original items and their original quantities are shown. Items or
quantities added during `Modifying` are not reflected.

**Screenshots/Videos**
N/A

**Error logs**
N/A

- @vendure/core version: 3.6.3 (also reproducible on 3.6.1 demo)
- Nodejs version: 20 (Alpine)
- Database (mysql/postgres etc): PostgreSQL 16
- Operating System (Windows/macOS/Linux): Docker (Alpine Linux)

**Configuration**
Custom order process plugin adds a `Dispensing` state and modifies the
`Modifying` / `ArrangingAdditionalPayment` transitions. The refund dialog
issue occurs regardless of this custom process.

**Minimal reproduction**
The issue can be reproduced with the default order process on the official
demo by modifying an order's items and then cancelling it. A custom process
is not required to trigger the bug.

**Workaround**
None known.

**Additional context**
This appears to be a limitation of the Vendure Admin UI refund dialog and
not caused by any custom order process plugin. It affects the ability to
perform accurate refunds.

grolmus:
## Summary

The dashboard **Refund & cancel order** dialog did not reflect line quantities that were added or increased while an order was in the `Modifying` state. Units the customer actually paid for (via an additional payment) could not be refunded, and newly-added lines did not appear at all.

## Root cause

`lineCanBeRefunded` / `getMaxRefundableQuantity` capped the refundable quantity at `OrderLine.orderPlacedQuantity`. That value is set once at order placement — it is `0` for lines added after placement and stays at the pre-increase value for increased lines. It was originally chosen (admin-ui #2608) as a *stable* denominator, because cancelling a line reduces `OrderLine.quantity`.

## Change

Use the effective quantity `Math.max(orderPlacedQuantity, quantity)` (extracted as `getRefundableQuantity`), the same pattern Vendure core already uses for `OrderLine` price/discount distribution. This reflects modified quantities while still preserving the #2608 behaviour (cancelling reduces `quantity`, so `orderPlacedQuantity` remains the upper bound). The dialog's displayed "Qty" now uses the same helper.

## Test plan

**Automated**
- New unit tests `order-utils.spec.ts` (new line `orderPlacedQuantity=0`, increased line, #2608 partial cancel+refund regression guard, fully-refunded, `getRefundableQuantity`).
- New e2e test in `sales/orders.spec.ts`: paid order → `Modifying` → increase line 1→2 → settle additional payment → the refund dialog offers `max=2` and accepts `2`. Confirmed this e2e **fails without the fix** (`max="1"`, input clamps to `1`).
- `bunx vitest run` (dashboard) green; `orders.spec.ts` e2e 18/18 green; typecheck clean on changed files.

**Manual**
- Create a paid order, transition to `Modifying`, increase a line's quantity, settle the additional payment, then open **Refund & cancel** — the increased quantity is shown and fully refundable.

Fixes #4728

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4917"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785671328&installation_id=128408429&pr_number=4917&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4917&signature=a1334fdb82df6bf28a9062c2bae70117962111551d89cbc5ed5f77c2f9f647e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/sales/orders.spec.ts
+++ b/packages/dashboard/e2e/tests/sales/orders.spec.ts
@@ -65,7 +65,10 @@ test.describe('Orders', () => {
         await filteredRequest;
         // The Regular order survives the matching filter, and an active filter badge appears.
         await lp.expectRowCountGreaterThan(0);
-        const filterBadge = page.getByRole('button').filter({ hasText: 'type' }).filter({ hasText: 'Regular' });
+        const filterBadge = page
+            .getByRole('button')
+            .filter({ hasText: 'type' })
+            .filter({ hasText: 'Regular' });
         await expect(filterBadge).toBeVisible();
         await expect(page.getByText(/An error occurred:/i)).toHaveCount(0);
 
@@ -495,6 +498,51 @@ test.describe('Orders', () => {
             await dialog.getByRole('button', { name: 'Cancel' }).click();
         });
 
+        // #4728 — refund dialog must reflect line quantities changed during order modification.
+        // Increasing a line from 1→2 while Modifying keeps orderPlacedQuantity=1; the dialog
+        // previously capped the refundable quantity at orderPlacedQuantity, so only 1 of the
+        // 2 paid units could be refunded. It must now offer the full modified quantity.
+        test('should reflect a line quantity increased during modification in the refund dialog', async ({
+            page,
+        }) => {
+            test.setTimeout(60_000);
+
+            const client = new VendureAdminClient(page);
+            await client.login();
+            const orderId = await createOrderWithIncreasedLineQuantity(client);
+
+            await page.goto(`/orders/${orderId}`);
+            await expect(page.getByRole('button', { name: /Fulfill order/i })).toBeVisible({
+                timeout: 10_000,
+            });
+
+            // Open the refund dialog via action bar dropdown
+            const actionBarEllipsis = page.getByTestId('action-bar-dropdown-trigger');
+            await expect(actionBarEllipsis).toBeVisible({ timeout: 10_000 });
+            await actionBarEllipsis.click();
+
+            const menu = page.locator('[data-slot="dropdown-menu-content"]');
+            await expect(menu).toBeVisible();
+            await menu
+                .getByText(/Refund/i)
+                .first()
+                .click();
+
+            const dialog = page.locator('[role="dialog"]');
+            await expect(dialog).toBeVisible({ timeout: 5_000 });
+
+            // The line's refund input must allow the full modified quantity (2).
+            // Pre-fix, `max` was orderPlacedQuantity (1) and entering 2 was clamped to 1.
+            const quantityInput = dialog.getByTestId('refund-quantity').first();
+            await expect(quantityInput).toBeVisible();
+            await expect(quantityInput).toHaveAttribute('max', '2');
+
+            await quantityInput.fill('2');
+            await expect(quantityInput).toHaveValue('2');
+
+            await dialog.getByRole('button', { name: 'Cancel' }).click();
+        });
+
         test('should process a refund', async ({ page }) => {
             test.setTimeout(60_000);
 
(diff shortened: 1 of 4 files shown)


biggamesmallworld (commented):
## Review — refund math checks out ✅

Traced the money path end-to-end (dashboard helpers → refund hook caps → core `order-modifier.ts`/`payment.service.ts`). The core logic is sound and the fix also closes a latent partial-refund bug (old code hid unit 2 of an increased line after refunding unit 1: `opq=1 - refunded=1 = 0`; new code correctly yields `max(1,2)-1 = 1`).

**On the edge case that worried me most — a line *decreased* during `Modifying` (e.g. 3→1): not an over-refund risk.** On decrease, core (`order-modifier.ts`) calls `cancelOrderByOrderLines` (drops `quantity`) *and* pushes `{ orderLineId, quantity: delta }` into the refund inputs, persisted as `RefundLine` rows. `getRefundedQuantity` sums exactly those, so `opq=3, q=1, refundedCount=2` → `getRefundableQuantity=max(3,1)=3` → `maxRefundable = 3-2 = 1`. Correct. And the actual money is hard-capped at the payment-amount layer (`use-refund-order.ts`) independent of quantity, so quantity only drives the *suggested* total.

**CI:** the red ❌ is only the Vercel storybook deploy — the full Build & Test suite (unit + e2e across 4 DBs + dashboard e2e) is green.

Nothing blocking. A few non-blocking suggestions:

### Recommended
1. **Document the invariant on `getRefundableQuantity`.** There's a net-neutral mixed-modification case (decrease line A by 2 *and* add line B, net delta ≥ 0) where core creates no refund but has already dropped A's `quantity` — so `opq=3, q=1, refundedCount=0` → `maxRefundable=3` while only 1 unit of A is live. This is **pre-existing** (old `opq`-based code gave the same 3) and is caught downstream by the payment-amount cap + core rejecting an over-quantity cancel, so it's not a regression or a money leak. But the reasoning currently spans three files — worth a one-line comment: *"effective ordered quantity; over-issuance is prevented at the payment-amount layer, not here."*

2. **Consider renaming `getRefundableQuantity`.** It returns `max(orderPlacedQuantity, quantity)` — the *effective ordered* quantity, before subtracting refunds — whereas the value that's actually refundable is `getMaxRefundableQuantity`. Two helpers whose names both say "refundable" but mean different things is easy to mix up later; something like `getEffectiveOrderedQuantity` would disambiguate.

3. **Add explicit tests for the decrease case + a pending refund.** The decrease-during-modify scenario is currently exercised only *coincidentally* — the `#2608` test (`opq=3, q=1, refundedCount=2 → 1`) is arithmetically identical, but nobody reading the spec would know it covers a modification-decrease. An explicitly-named `decreased during modification` case, plus one with a `state: 'Pending'` modification refund (the mock supports it but no case uses it), would lock the invariant against future refactors.

### Optional
- The e2e asserts `max='2'` and `fill('2')` (clean, exactly distinguishes fixed vs broken 👍) but clicks *Cancel* — it never submits, so it doesn't prove core actually accepts refunding the increased units end-to-end. A follow-up that completes the refund and asserts the resulting `Refund.total`/history would be the stronger regression guard.
- Minor: per line per render, `getRefundedQuantity` runs three times (via `lineCanBeRefunded`, `getMaxRefundableQuantity`, and the display). `n` is tiny so it's cosmetic — could compute `{ canRefund, maxRefundable, effectiveQty }` once per line.

Nice work — the unit-test mock hard-codes literal expected values rather than re-deriving them from the implementation, which is the right way to avoid tautological tests.

biggamesmallworld:
I manually tested this locally and it looks correct, able to refund 2 line items from an order. 

Screenshot 2026-07-06 at 16 20 31

biggamesmallworld (approved):
🏄‍♂️

PR #4918, opened 2026-07-03
fix(dashboard): Fix stock location deletion and add stock transfer on delete

### LINKED ISSUE #4641: Not possible to delete a Stock Location
**Describe the bug**
In the Admin Dashboard it is not possible to delete a Stock Location even when all product variants assigned to that Stock Location have zero availability.

The UI prevents deletion as if stock still exists, even though inventory is empty.

**To Reproduce**

1. Go to Admin Dashboard → Settings → Stock Locations
2. Select an existing Stock Location that has product variants assigned.
3. Ensure all variants in that location have 0 stock on hand and 0 stock allocated.
4. Attempt to delete the Stock Location.
5. Observe that deletion is blocked.

**Expected behavior**
If a Stock Location has no available or allocated stock, it should be possible to delete it without restrictions.

**Actual behavior**
The Dashboard does not allow deleting the Stock Location, behaving as if stock still exists even though availability is zero.

**Screenshots/Videos**

Image

**Error logs**
Show toast error message:

```
Failed to delete 1 stock locations
```

- Nodejs version: 22
- Database: postgres
- Operating System: Windows
- Browser: Chrome
- Package manager: npm and pnpm

**Additional context**

- Three custom fields have been added

biggamesmallworld:
## Summary

Fixes the dashboard bug where **deleting a Stock Location always fails** with a "Failed to delete N stock locations" toast (reported in #4641 / OSS-493), and — since core already supports it — exposes **transferring the location's stock to another location on delete**.

## Root cause

The generic `DeleteBulkAction` sends its mutation variables as `{ ids: [...] }`, matching the `deleteX(ids: [ID!]!)` convention used by every other entity's bulk delete. But `deleteStockLocations` is the one exception — it takes `input: [DeleteStockLocationInput!]!` (so it can carry a per-location `transferToLocationId`):

```graphql
deleteStockLocations(input: [DeleteStockLocationInput!]!): [DeletionResponse!]!
```

So the required `$input` variable was never provided and the request failed GraphQL validation:

```
Variable "$input" of required type "[DeleteStockLocationInput!]!" was not provided.
```

That error hits the `DeleteBulkAction` `onError` branch, producing the hard-coded `Failed to delete N stock locations` toast (no trailing message) — which matches the reporter's screenshot exactly. It fails for **every** stock location, with or without stock, so the "zero stock" detail in the report was a red herring.

Note: this is **not** a missing FK-cascade bug in core (as investigated on the issue) — a well-formed `deleteStockLocation(s)` request via the Admin API already works. The failure is purely the dashboard sending the wrong variable shape.

## Reproduction

After reading the comment from @grolmus on the Linear task, I saw that he couldn't reproduce the error via the API. When I went to reproduce it via the UI I was able to reproduce it, with the API response in the network tab returning this:

```
{
    "errors": [
        {
            "message": "Variable \"$input\" of required type \"[DeleteStockLocationInput!]!\" was not provided.",
            "locations": [
                {
                    "line": 1,
                    "column": 31
                }
            ],
            "extensions": {
                "code": "BAD_USER_INPUT"
            }
        }
    ]
}
```

The true issue at hand, not the FK issue originally thought.

## Fix + feature

`DeleteStockLocationInput.transferToLocationId` already lets core move a location's stock (levels + allocations) to another location on delete — but no admin UI ever surfaced it (neither the dashboard nor the legacy Angular app). So instead of only correcting the variable shape:

- Replace the plain delete confirmation for stock locations with a dedicated **delete dialog** that asks what to do with remaining stock: **transfer to another location** or **discard** — then sends the correct `input` array with `transferToLocationId` per location.
- The shared `DeleteBulkAction` is left untouched (still `{ ids }` for all other entities).

## Steps to verify

1. Admin Dashboard → Settings → Stock Locations
2. Select a location via its checkbox → **Delete**
3. Before: toast "Failed to delete 1 stock locations"; `$input ... was not provided` in the network tab
4. After: a dialog lets you transfer remaining stock to another location (or discard), then deletes successfully

## Screen recording

https://github.com/user-attachments/assets/43c76f2a-4244-4b56-9cd7-d0ebc5f6a03b

## Testing

Added a dashboard e2e test (`packages/dashboard/e2e/tests/settings/stock-locations.spec.ts`) that drives the real delete dialog end-to-end (select → **Delete** → choose "Discard remaining stock" → confirm) and asserts the location is removed — guarding against any regression back to the wrong mutation variable shape. The stock-location CRUD suite's generic bulk-delete (which drove the old confirm dialog) was replaced by this dialog-aware test.

## Notes

- Only delete path in the dashboard is the bulk action; the single-delete document was unused dead code.
- `useDetailPage`/list pages unchanged; no core changes.
- The transfer-target picker caps at 100 locations and invalidates its cached list after a delete.

Fixes #4641

### DIFF
--- a/packages/dashboard/e2e/tests/settings/stock-locations.spec.ts
+++ b/packages/dashboard/e2e/tests/settings/stock-locations.spec.ts
@@ -2,8 +2,11 @@ import { expect, test } from '@playwright/test';
 
 import { BaseListPage } from '../../page-objects/list-page.base.js';
 import { createCrudTestSuite } from '../../utils/crud-test-factory.js';
+import { VendureAdminClient } from '../../utils/vendure-admin-client.js';
 
-const UPDATED_NAME = 'E2E Test Warehouse Updated';
+// Unique per run so repeated local runs (which don't reset the DB) don't leave duplicates that
+// would break the final row-count assertion.
+const DELETE_TARGET_NAME = `E2E Delete Dialog Location ${Date.now()}`;
 
 test.describe('Stock Locations', () => {
     test.describe.configure({ mode: 'serial' });
@@ -20,7 +23,7 @@ test.describe('Stock Locations', () => {
             { label: 'Description', value: 'A test warehouse for e2e testing' },
         ],
         updateFields: [
-            { label: 'Name', value: UPDATED_NAME },
+            { label: 'Name', value: 'E2E Test Warehouse Updated' },
             { label: 'Description', value: 'Updated test warehouse description' },
         ],
         // Stock locations use a bespoke delete dialog (transfer/discard remaining stock) rather
@@ -31,8 +34,18 @@ test.describe('Stock Locations', () => {
     // #4641 — Deleting a stock location previously always failed because the shared bulk-delete
     // action sent `{ ids }` while `deleteStockLocations` requires `input: [DeleteStockLocationInput!]!`.
     // This drives the real dialog end-to-end; if the mutation variables regress, the delete fails
-    // and no success toast appears. Also cleans up the entity created by the CRUD suite above.
+    // and no success toast appears.
     test('should bulk-delete a stock location via the transfer/discard dialog', async ({ page }) => {
+        // Seed a throwaway location via the API so the test is self-contained.
+        const client = new VendureAdminClient(page);
+        await client.login();
+        await client.gql(
+            `mutation ($input: CreateStockLocationInput!) {
+                createStockLocation(input: $input) { id }
+            }`,
+            { input: { name: DELETE_TARGET_NAME } },
+        );
+
         const listPage = new BaseListPage(page, {
             path: '/stock-locations',
             title: 'Stock Locations',
@@ -40,23 +53,23 @@ test.describe('Stock Locations', () => {
         });
         await listPage.goto();
         await listPage.expectLoaded();
-        await listPage.search(UPDATED_NAME);
+        await listPage.search(DELETE_TARGET_NAME);
 
-        const row = listPage.getRows().filter({ hasText: UPDATED_NAME });
+        const row = listPage.getRows().filter({ hasText: DELETE_TARGET_NAME });
         await expect(row.first()).toBeVisible();
         await row.first().getByRole('checkbox').click();
 
-        await page.getByRole('button', { name: /With selected/i }).click();
-        await page.getByRole('menuitem').filter({ hasText: 'Delete' }).click();
+        await page.getByRole('button', { name: /Actions/i }).click();
+        await page.locator('[role="menu"]').getByText('Delete', { exact: true }).click();
 
         // Custom delete dialog: choose what to do with any remaining stock, then confirm.
         const dialog = page.getByRole('dialog');
         await expect(dialog.getByText('Delete stock locations')).toBeVisible();
         await dialog.getByRole('combobox').click();
         await page.getByRole('option', { name: /Discard remaining stock/i }).click();
-        await dialog.getByRole('button', { name: 'Delete' }).click();
+        await dialog.getByRole('button', { name: 'Delete', exact: true }).click();
 
         await listPage.expectSuccessToast();
-        await expect(listPage.getRows().filter({ hasText: UPDATED_NAME })).toHaveCount(0);
+        await expect(listPage.getRows().filter({ hasText: DELETE_TARGET_NAME })).toHaveCount(0);
     });
 });
--- a/packages/dashboard/e2e/tests/settings/stock-locations.spec.ts
+++ b/packages/dashboard/e2e/tests/settings/stock-locations.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from '@playwright/test';
+import { expect, test, type Page } from '@playwright/test';
 
 import { BaseListPage } from '../../page-objects/list-page.base.js';
 import { createCrudTestSuite } from '../../utils/crud-test-factory.js';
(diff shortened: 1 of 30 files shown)


grolmus (dismissed):
Thanks @biggamesmallworld — excellent root-cause writeup, and the fix is right: the `input: [DeleteStockLocationInput!]!` shape now matches the schema, and the partial-success handling (per-result `DELETED` check, dialog stays open when nothing was deleted) is more careful than the generic action it replaces. I ran the dashboard e2e locally — 10/10 green.

A few things before merge:

### Should fix

1. **Test the transfer path — the feature this PR is named after is unverified.** The suite only drives *"Discard remaining stock"* (`stock-locations.spec.ts:69`), which exercises the `transferToLocationId === undefined` branch. The branch where `transferToLocationId` is a real id is never executed, so a regression there (wrong id, accidentally sending the `__discard__` sentinel, an off-by-one in the `selection.map`) would pass green. Please add a test that seeds two locations, selects one, picks *"Transfer to \<other\>"*, and asserts success — ideally seeding stock and asserting it moved. A **multi-select** delete is worth covering too, since the input is built by mapping over the whole selection and the description promises "all selected locations transfer to a single chosen target".

2. **Dead `name?` prop.** `DeleteStockLocationsDialogProps.selection: Array<{ id: string; name?: string }>` — `name` is never read (the labels come from the fetched target list). Either drop it to `Array<{ id: string }>` or use it (e.g. list the locations being deleted in the dialog description).

3. **Unbounded paginate-all loop.** `delete-stock-locations-dialog.tsx:65-71` — `do … while (collected.length < totalItems)` loops forever if a page ever returns zero items while `totalItems` still reports more (permission-filtered page, `take` clamped, a server hiccup). Location counts are tiny so it won't bite in practice, but break on an empty page or cap the iterations.

### Minor

4. The `invalidateQueries` after a successful delete is essentially redundant — the target query is `enabled: open` with the default `staleTime: 0`, so it already refetches every time the dialog opens. Either drop it, or fix the comment, which currently states a rationale the default staleness already covers.
5. **a11y:** the *"Remaining stock"* `<label>` (`:147`) isn't programmatically associated with the `Select` (no `htmlFor`/`id`), so assistive tech won't tie the label to the control.
6. **i18n:** `Deleted ${n} stock locations` and the `stock location(s)` in the description skip Lingui's plural support. The generic bulk action has the same gap so it's at least consistent — worth a `plural()` while you're here.
7. The *"Transfer to {name}"* label is defined twice — the `items` map (`:80`) and the `SelectItem` children (`:172`). The `items` map is required by the Select wrapper, but pull the string through one helper so the two can't drift.

The core fix is solid and the failure-path handling is genuinely good; mainly I want the transfer path under test before this goes in. Thanks!

biggamesmallworld:
Thanks @grolmus — all seven points addressed in ee8e3d39f.

**1. Transfer path under test.** Added two e2e tests to `stock-locations.spec.ts`:
- `should transfer remaining stock to the chosen location on delete` — seeds a source and a target location via the API, sets `stockOnHand: 42` on a variant at the source, then drives the dialog with *"Transfer to \<target\>"* and asserts the target location ends up with exactly 42. I verified the test has teeth by hard-coding `transferToLocationId = undefined` locally: it fails.
- `should bulk-delete multiple stock locations at once` — seeds two locations, selects both rows, deletes. Covers the `selection.map` path.

The three delete tests now share a `bulkDelete()` helper so the dialog driving isn't duplicated. Full suite is green locally (12/12).

**2. Dead `name?` prop.** Dropped — `selection` is now `Array<{ id: string }>`.

**3. Unbounded paginate-all loop.** Breaks on an empty page.

**4. Redundant `invalidateQueries`.** Removed, and the comment now states the actual reason (`enabled: open` + default `staleTime`, so it refetches on each open).

**5. a11y.** The label now has an `id` and the `SelectTrigger` an `aria-labelledby`.

**6. i18n plurals.** Delete toasts use `plural()` from `@lingui/core/macro`, the dialog description uses `<Plural>`. Catalogs re-extracted.

**7. Duplicated label.** `transferLabel(name)` is defined once and used by both the `items` map and the `SelectItem` children — extraction now yields a single `Transfer to {name}` message instead of two.

biggamesmallworld:
Addressed the CodeRabbit i18n findings in 0a2f6aa52.

CodeRabbit flagged empty `msgstr` values for the new deletion-flow messages in `es`, `nb`, `ru`, `uk` and `uz`. Rather than fix only those five, I filled all **11 new messages across all 26 non-source locales**, using the workflow documented in `scripts/translate/README.md` (`i18n:extract` → LLM → `i18n:apply`). The tool's script-validation passed for every batch.

Notes:
- **ICU plural categories follow each locale's CLDR rules** — `few`/`many` for `ru`/`uk`/`pl`/`cs`/`hr`/`ro`, `zero`/`two`/`few`/`many` for `ar`, `one`/`two`/`other` for `he`, and `other`-only for `ja`/`zh_Hans`/`zh_Hant`. All placeholders (`{0}`/`{1}`/`{2}`/`{count}`/`{deleted}`/`{messages}`/`#`) preserved.
- **`lingui compile` now passes for every catalog.** It caught a real bug in my first pass: the Italian `nell'#` wrote an ICU apostrophe escape that swallowed the rest of the message. Reworded to `in {count, plural, ...}`.
- The diff is `msgstr`-only — 11 lines per file, 26 files, `en.po` untouched.
- `check-i18n-sync.sh` → *in sync*.

Left alone: the ~39 pre-existing untranslated entries per locale that already exist on `master` and are unrelated to this PR.

Also rebased onto `master` (was 93 commits behind) — e2e suite re-run green post-rebase, and Build & Test is now running, which it wasn't on the previous push.

grolmus (approved):
Thanks @biggamesmallworld — this is a thorough turnaround. I re-checked all seven points from the last round and they're genuinely addressed:

1. **Transfer path under test** ✅ — the new `should transfer remaining stock to the chosen location on delete` test seeds 42 units at the source and asserts `getStockOnHand(target) === 42` after the dialog-driven transfer. That's a real positive anchor on the exact branch that was previously unverified, and the multi-select test covers the `selection.map` path. Nicely done.
2. Dead `name?` prop dropped ✅
3. Paginate-all loop now breaks on an empty page ✅
4. Redundant `invalidateQueries` removed + comment corrected ✅
5. Label associated via `aria-labelledby` ✅
6. Plurals via `plural()` / `<Plural>` ✅
7. `Transfer to {name}` pulled through a single `transferLabel` helper ✅

i18n is filled across all locales with the plural branches preserved (incl. the Slavic `one/few/many/other` in `ru`/`uk`) — so CodeRabbit's empty-`msgstr` findings are already resolved. CI is green and SonarQube passed.

Approving. A few **non-blocking** notes — the first two are one-liners worth landing if quick, the rest are follow-ups:

**Worth a quick fix**

- **Opt the transfer-targets query out of `keepPreviousData`.** The dashboard sets `placeholderData: keepPreviousData` globally, and this query doesn't opt out. So on reopen it briefly shows the *previous* target list — which can still contain a location you just deleted — until the `staleTime: 0` refetch resolves, leaving a just-deleted location momentarily selectable as a transfer destination. Per the dashboard convention (`packages/dashboard/CLAUDE.md`), add `placeholderData: undefined` to this query since the stale identity is misleading here. (This refines my earlier point 4 — dropping `invalidateQueries` was right; the remaining gap is the `keepPreviousData` frame, not the refetch.)
- **Disable `Cancel` while the delete is pending** (`disabled={isPending}`). Otherwise a user can close mid-delete and reopen with a different selection before the mutation settles, and the old closure's `onSuccess` then closes/refetches the newly-opened instance. (Also flagged by CodeRabbit.)

**Minor / follow-up**

- The three delete tests call `expectSuccessToast()` with no argument, which matches any non-error toast. Since you know the exact message, pass `/Deleted \d+ stock location/` so the assertion can't pass on an unrelated success toast. The row-count check covers you today, but the explicit anchor is stronger.
- The `// The dialog stays mounted…` comment (delete dialog, reset effect) is slightly off: on the happy path `onSuccess` resets the row selection, which unmounts the bulk-actions toolbar and this dialog with it; the reset effect only matters in the all-failed path where the menu stays open. The effect is harmless — just tidy the comment so the lifecycle reasoning doesn't mislead.
- `if (value != null)` in `onValueChange` guards a case the `Select` config can't produce — can be dropped.
- Not for this PR: the singular `deleteStockLocationDocument` in `stock-locations.graphql.ts` is now orphaned (nothing imports it) — worth a follow-up sweep. And with `hasBulkDelete: false`, the CRUD factory no longer cleans up its `E2E Test Warehouse` row, so it accumulates on repeated local runs; a `RUN_ID`-scoped name or an API cleanup would keep local reruns clean.

The core fix is solid and the transfer path is now properly covered — thanks for the careful work on this.

PR #4922, opened 2026-07-05
fix(core): validate minimum threshold in promotion conditions

### LINKED ISSUE #4889: contains_products and at_least_n_with_facets: minimum = 0 makes the condition always true (unconditional discount)
**Describe the bug**

The `minimum` argument of both the `contains_products` and `at_least_n_with_facets` promotion conditions can be set to `0`. When it is, the `check()` function always returns `true` regardless of the cart contents, creating an unconditional discount that fires on every order.

The gate in both conditions is `args.minimum <= matches`. With `minimum = 0` and no matching items (`matches = 0`):

```
0 <= 0  →  true
```

The condition applies even though the order contains none of the specified product variants / none of the products with the required facet values.

**To Reproduce**

1. Create a promotion using `contains_products` (or `at_least_n_with_facets`).
2. Set `minimum = 0` (allowed by the UI; no server-side validation rejects it).
3. Place any order.
4. The promotion condition fires unconditionally, regardless of what is in the cart.

**Expected behavior**

`minimum = 0` should not be a valid threshold — "buy at least 0 items" matches every order and defeats the purpose of the condition. Expected: `check()` returns `false` when no eligible items are present, regardless of `minimum`.

**Actual behavior**

`check()` returns `true` whenever `minimum = 0`, even with an empty or unrelated cart.

**Root cause**

Commit `0f95eccfe` ("fix(core): Add minimums to default promotion actions/conditions") added `ui: { component: '', min: 0 }` to the `minimum` arg of these conditions to prevent negative values. The intended floor is clearly `1` — both conditions have `defaultValue: 1` and their human-readable descriptions say "Buy at least { minimum }…". The `min: 0` applied in the blanket fix is an off-by-one.

There is no server-side range validation; `minimum = 0` passes through the GraphQL API and the Admin UI without any error.

**Suggested fix**

Either (or both):
- Change `ui: { ..., min: 0 }` → `ui: { ..., min: 1 }` on the `minimum` arg in both condition files.
- Add an early guard in `check()`:
  ```ts
  if (args.minimum < 1) return false;
  ```

Apply the same fix symmetrically to both `contains-products-condition.ts` and `has-facet-values-condition.ts`.

**Environment**

- `@vendure/core` — `master`
- Files:
  - `packages/core/src/config/promotion/conditions/contains-products-condition.ts`
  - `packages/core/src/config/promotion/conditions/has-facet-values-condition.ts`

**Additional context**

Found via boundary-value analysis of the `args.minimum <= matches` gate while adding unit tests for these conditions (see #4835 / PRs #4881, #4882). The test suites deliberately exclude assertions on this behaviour to avoid locking in what appears to be a defect.

The root cause is the same commit (`0f95eccfe`) responsible for the related `amountX = 0` issue in `buy_x_get_y_free` (see companion issue #4888).

Sanjuhub:
# Description

This PR fixes a boundary-value bug in the `contains_products` and `at_least_n_with_facets` promotion conditions.

## Changes

- Added server-side validation to reject minimum < 1 in the affected promotion conditions and updated the Admin UI hint to use a minimum value of 1.

Fixes #4889.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:
- [X ] I have added or updated test cases

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4922"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785834622&installation_id=128408429&pr_number=4922&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4922&signature=6026144375d8f6a39e34500938a23a31d96a44143f9910901fd8283c334a6422"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/order-promotion.e2e-spec.ts
+++ b/packages/core/e2e/order-promotion.e2e-spec.ts
@@ -598,6 +598,57 @@ describe('Promotions applied to Orders', () => {
             await deletePromotion(promotion.id);
         });
 
+        // #4889 — minimum of 0 must not create an unconditional discount
+        it('containsProducts does not apply when minimum is 0', async () => {
+            const item5000 = getVariantBySlug('item-5000')!;
+            const promotion = await createPromotion({
+                enabled: true,
+                name: 'Contains products, minimum 0',
+                conditions: [{
+                    code: containsProducts.code,
+                    arguments: [
+                        { name: 'minimum', value: '0' },
+                        { name: 'productVariantIds', value: JSON.stringify([item5000.id]) },
+                    ],
+                }],
+                actions: [freeOrderAction],
+            });
+            // add an item that is NOT in productVariantIds
+            const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
+                productVariantId: getVariantBySlug('item-1000').id,
+                quantity: 1,
+            });
+            orderResultGuard.assertSuccess(addItemToOrder);
+            expect(addItemToOrder.discounts.length).toBe(0);
+            await deletePromotion(promotion.id);
+        });
+
+        // #4889 — same guard for the facet-based condition
+        it('atLeastNWithFacets does not apply when minimum is 0', async () => {
+            const { facets } = await adminClient.query(getFacetListDocument);
+            const saleFacetValue = facets.items[0].values[0];
+            const promotion = await createPromotion({
+                enabled: true,
+                name: 'Facets, minimum 0',
+                conditions: [{
+                    code: hasFacetValues.code,
+                    arguments: [
+                        { name: 'minimum', value: '0' },
+                        { name: 'facets', value: `["${saleFacetValue.id}"]` },
+                    ],
+                }],
+                actions: [freeOrderAction],
+            });
+            // add an item WITHOUT the Sale facet
+            const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
+                productVariantId: getVariantBySlug('item-1000').id,
+                quantity: 1,
+            });
+            orderResultGuard.assertSuccess(addItemToOrder);
+            expect(addItemToOrder.discounts.length).toBe(0);
+            await deletePromotion(promotion.id);
+        });
+
         it('customerGroup', async () => {
             const { createCustomerGroup } = await adminClient.query(createCustomerGroupDocument, {
                 input: { name: 'Test Group', customerIds: ['T_1'] },
--- a/packages/core/src/config/promotion/conditions/buy-x-get-y-free-condition.ts
+++ b/packages/core/src/config/promotion/conditions/buy-x-get-y-free-condition.ts
@@ -16,7 +16,7 @@ export const buyXGetYFreeCondition = new PromotionCondition({
         amountX: {
             type: 'int',
             defaultValue: 2,
-            ui: { component: '', min: 0 },
+            ui: { component: '', min: 1 },
         },
         variantIdsX: {
             type: 'ID',
@@ -39,6 +39,7 @@ export const buyXGetYFreeCondition = new PromotionCondition({
     async check(ctx, order, args) {
         const xIds = createIdentityMap(args.variantIdsX);
         const yIds = createIdentityMap(args.variantIdsY);
+        if (args.amountX < 1) return false;
         let matches = 0;
         const freeItemCandidates: OrderLine[] = [];
         for (const line of order.lines) {
--- a/packages/core/src/config/promotion/conditions/contains-products-condition.ts
+++ b/packages/core/src/config/promotion/conditions/contains-products-condition.ts
@@ -14,7 +14,7 @@ export const containsProducts = new PromotionCondition({
         minimum: {
             type: 'int',
             defaultValue: 1,
-            ui: { component: '', min: 0 },
+            ui: { component: '', min: 1 },
         },
         productVariantIds: {
             type: 'ID',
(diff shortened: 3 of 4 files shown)


Sanjuhub:

GabrielRoc (approved):

biggamesmallworld:
Thanks for this — the runtime guard is the right fix and it fully closes the reported hole. 🙌

Two things before it's ready to merge (details in the two follow-up comments below), both copy-paste ready so there's nothing to figure out:

1. The same bug exists in the `buy_x_get_y_free` condition — worth fixing in the same PR.
2. A regression test for each fixed condition.

**Happy to approve once those two are addressed.**

One minor note (no code change): the description says "Updated the UI validation to require `minimum >= 1`". The `ui.min` is a dashboard rendering hint only — the API still accepts `minimum: 0`, so your runtime `check()` guard is what actually fixes the bug. Just flagging so the framing's accurate for anyone reading later.

biggamesmallworld:
**Same bug in `buy_x_get_y_free` — please fix it here too**

`amountX` carries the same `ui: { min: 0 }`, so the API accepts `0`. In `packages/core/src/config/promotion/conditions/buy-x-get-y-free-condition.ts`, `Math.floor(matches / 0)` evaluates to `Infinity` (truthy), so the `!quantity` guard passes and any cart containing a single X-variant gets the free Y items — the "buy X" threshold is bypassed entirely. Same root cause and same fix shape as your change:

```diff
         amountX: {
             type: 'int',
             defaultValue: 2,
-            ui: { component: '', min: 0 },
+            ui: { component: '', min: 1 },
         },
```

```diff
     async check(ctx, order, args) {
         const xIds = createIdentityMap(args.variantIdsX);
         const yIds = createIdentityMap(args.variantIdsY);
+        if (args.amountX < 1) return false;
         let matches = 0;
```

(`amountY: 0` is harmless — "get 0 free" — so it can stay as-is.)

biggamesmallworld:
**Please add regression tests**

Paste these two cases into `packages/core/e2e/order-promotion.e2e-spec.ts` right after the existing `containsProducts` test (around line 470). They fail on `master` and pass with your fix — each adds an item that is *not* covered by the condition, which pre-fix would still trigger the discount:

```ts
// #4889 — minimum of 0 must not create an unconditional discount
it('containsProducts does not apply when minimum is 0', async () => {
    const item5000 = getVariantBySlug('item-5000')!;
    const promotion = await createPromotion({
        enabled: true,
        name: 'Contains products, minimum 0',
        conditions: [{
            code: containsProducts.code,
            arguments: [
                { name: 'minimum', value: '0' },
                { name: 'productVariantIds', value: JSON.stringify([item5000.id]) },
            ],
        }],
        actions: [freeOrderAction],
    });
    // add an item that is NOT in productVariantIds
    const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
        productVariantId: getVariantBySlug('item-1000').id,
        quantity: 1,
    });
    orderResultGuard.assertSuccess(addItemToOrder);
    expect(addItemToOrder.discounts.length).toBe(0);
    await deletePromotion(promotion.id);
});

// #4889 — same guard for the facet-based condition
it('atLeastNWithFacets does not apply when minimum is 0', async () => {
    const { facets } = await adminClient.query(getFacetListDocument);
    const saleFacetValue = facets.items[0].values[0];
    const promotion = await createPromotion({
        enabled: true,
        name: 'Facets, minimum 0',
        conditions: [{
            code: hasFacetValues.code,
            arguments: [
                { name: 'minimum', value: '0' },
                { name: 'facets', value: `["${saleFacetValue.id}"]` },
            ],
        }],
        actions: [freeOrderAction],
    });
    // add an item WITHOUT the Sale facet
    const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
        productVariantId: getVariantBySlug('item-1000').id,
        quantity: 1,
    });
    orderResultGuard.assertSuccess(addItemToOrder);
    expect(addItemToOrder.discounts.length).toBe(0);
    await deletePromotion(promotion.id);
});
```

Heads-up: these tests share the active order across `it` blocks, so if the suite doesn't already reset the cart between cases you may need to start a fresh order at the top — worth a quick check when running locally. `item-5000` and `item-1000` are already used by the neighbouring tests, so no fixture changes are needed.

Sanjuhub:
Thanks @biggamesmallworld,  for the detailed feedback! I've addressed the requested changes:

- Added the same validation to buy_x_get_y_free.
- Added regression tests covering the minimum = 0 cases.
- Updated the PR description to clarify that the runtime validation is the actual fix.

Please let me know if there's anything else you'd like me to update.

biggamesmallworld (approved):

PR #4923, opened 2026-07-05
perf(core): resolve relation custom fields using request-scoped DataLoader batching

### LINKED ISSUE #4796: Relation custom fields cause N+1 queries in list contexts with no way to optimize
**Describe the bug**

When querying a list of entities that have `relation`-type custom fields, Vendure runs one database query per entity to resolve each relation — the classic N+1 problem. There is no way for plugin authors to optimize this because the auto-generated custom field resolvers silently override any `@ResolveField` a plugin defines.

**To Reproduce**

1. Define a relation custom field on an entity used in list queries:
```typescript
config.customFields.OrderLine.push({
    name: 'subItems',
    type: 'relation',
    entity: SubItem,
    list: true,
});
```
2. Query a list that includes the relation:
```graphql
query {
  activeOrder {
    lines {
      customFields {
        subItems { id name }
      }
    }
  }
}
```
3. Observe database queries — each order line triggers its own query to resolve `subItems`.

**Expected behavior**

Relation custom fields should be resolved efficiently in list contexts, either by batching queries internally or by allowing plugin authors to plug in their own DataLoader.

**Actual behavior**

`CustomFieldRelationResolverService.resolveRelation()` runs a subquery + main query per entity per relation field. For 50 order lines with one relation custom field, that's 51 queries instead of 2-3.

Two expected escape routes don't work:

1. **`eager: true` on the custom field config** — maps to TypeORM's eager decorator option, but `ListQueryBuilder` uses `QueryBuilder` + `setFindOptions()`. While `setFindOptions` does call `joinEagerRelations()`, that method walks `metadata.eagerRelations` on the root entity. Custom field relations live on the embedded `customFields` entity metadata, which `joinEagerRelations` doesn't recurse into. Meanwhile, `joinTreeRelationsDynamically` explicitly skips non-tree, non-self-referencing relations. So `eager: true` has no practical effect on list queries.

2. **Plugging in a DataLoader via `@ResolveField`** — The auto-generated resolvers on `{Entity}CustomFields` silently override plugin resolvers. NestJS GraphQL merges resolvers using a shallow spread (`extend(typesResolvers, options.resolvers)`), and the programmatic resolver map from `generateResolvers()` replaces the entire type object from decorator-based resolvers.

**Workaround**

The only current workaround is to mark the custom field as `internal: true`, then re-add the field as a schema extension on the parent entity with a `@ResolveField` + DataLoader. This works but requires maintaining two representations of the same field.

**Docs note**

The [DataLoader docs](https://docs.vendure.io/guides/developer-guide/dataloaders/) state:

> "CustomField relations are automatically added to the root query for the entity that they are part of."

This is not accurate for list queries. The relation is resolved lazily per entity by the auto-generated resolver.

**Proposed solution**

Add DataLoader-style batching inside `CustomFieldRelationResolverService`. The `resolveRelation()` method would delegate to a request-scoped DataLoader (keyed by request context + entity name + field name) that collects all entity IDs in a single tick and runs one batched query with `WHERE entity.id IN (:...ids)`.

This is a contained change to one file. The auto-generated resolvers in `generate-resolvers.ts` continue calling `resolveRelation()` the same way — they just get batched results instead of individual queries. No plugin code changes needed.

Prior art: commit `7b57d28` added a DataLoader for the Collection → ProductVariants N+1.

**Environment**

- Affects all 26+ entity types that support custom fields

**Relevant files**

- `packages/core/src/api/common/custom-field-relation-resolver.service.ts` — per-entity query (lines 36-57)
- `packages/core/src/api/config/generate-resolvers.ts` — auto-generated resolvers (lines 239-267)
- `packages/core/src/service/helpers/list-query-builder/list-query-builder.ts` — uses QueryBuilder
- `packages/core/src/service/helpers/utils/tree-relations-qb-joiner.ts` — skips non-tree relations (lines 100-108)
- `packages/core/src/entity/register-custom-entity-fields.ts` — eager config registration (lines 45-56)

supermadu7:
# Description

This PR resolves an operational performance bottleneck where relation-type custom fields trigger individual database subqueries for each record in list contexts, generating a classic $N+1$ query loop. 

To eliminate this overhead, a request-scoped `DataLoader` implementation has been introduced inside `CustomFieldRelationResolverService.resolveRelation()`. The loader batches concurrent entity relationship lookups within a single event loop tick, consolidating arbitrary row execution matrices down to a single, efficient relational `IN` batch query.

Related Issue: Fixes #4796

# Breaking changes

None. This optimization completely preserves existing framework structural interfaces, ensuring full backward compatibility across legacy plugin resolver spreads without requiring changes to user-defined custom fields.

# Checklist

📌 Always:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4923"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785873848&installation_id=128408429&pr_number=4923&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4923&signature=a1a152d674d5edeb34759696846d02159b8138df36661b23dfb2a9a34a083d4f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


supermadu7:

grolmus (changes requested):
Thanks for tackling this @supermadu7 — the N+1 in relational custom-field resolution is a real pain in list contexts, and the DataLoader batching approach is the right idea. I verified it does coalesce into a single batched lookup per tick. A few things need addressing before we can merge.

## 🔴 Blocking

**1. `dataloader` is declared in the wrong place.**
It's imported in `packages/core/src/api/common/custom-field-relation-resolver.service.ts` (runtime code that ships in the published package), but it was added to the **root `package.json` `devDependencies`**. It needs to be a runtime dependency of core: add it to `packages/core/package.json` `dependencies` (pinned caret range) and remove it from the root `devDependencies`. Lerna hoisting hides this locally, but consumers of `@vendure/core` would hit `MODULE_NOT_FOUND` at runtime.

**2. Please don't monkeypatch `RequestContext` via `ctx as any`.**
`const ctxWithCache = ctx as any; ctxWithCache._customFieldLoaders = {}` defeats typing and invents an undocumented field. Vendure already has a request-scoped memoization facility for exactly this: `RequestContextCacheService` (`packages/core/src/cache/request-context-cache.service.ts`), which is a DI-injected `WeakMap<RequestContext, Map>` — see `LocaleStringHydrator` for precedent. Inject it and store the loader under a namespaced key, e.g.:

\`\`\`ts
private getLoader(ctx: RequestContext, entityName: string, fieldDef: RelationCustomFieldConfig) {
    const key = \`CustomFieldRelationLoader:\${entityName}:\${fieldDef.name}\`;
    return this.requestCache.get(ctx, key, () =>
        new DataLoader<ID, VendureEntity | VendureEntity[] | null>(ids =>
            this.batchLoad(ctx, entityName, fieldDef, ids),
        ),
    );
}
\`\`\`

Note this adds a constructor dependency — please confirm `RequestContextCacheService` is available in the injector that builds `CustomFieldRelationResolverService` (`api-internal-modules.ts`; it's provided by `CacheModule`).

**3. Tests are required.**
This changes how *every* relational custom field resolves in *every* list query, across three DB drivers, with a caching layer and permission gating — it can't merge without coverage. Minimum:
- **Query-count assertion** — the whole point of the PR. Assert a list of N rows resolves in a *constant* number of queries (`.toBe(2)`), not `2N`.
- **List vs single** relation both return correct, hydrated, translated entities.
- **Order preservation** — row 3's relation must be row 3's, not row 1's. This is the riskiest logic in the diff.
- **Permission gating** still returns `null` for a user lacking the custom-field permission.
- **Empty/null relation** returns `null` (single) / `[]` (list) without poisoning sibling batches.
These should run under the existing sqlite/postgres/mysql e2e matrix.

## 🟡 Should-fix

4. **Raw-alias portability.** `getRawMany()` with hand-written `AS entity_id` / `AS relation_id` is a classic cross-driver footgun (identifier case-folding differs between postgres/mysql/sqlite). Read the aliases back using the exact strings you aliased, and make sure an empty batch can't produce `IN ()`. The cross-DB test above is the guard here.
5. **Cache lifetime / staleness.** DataLoader memoizes per key for the request lifetime, so a read after a same-request mutation would see a stale value. That's acceptable for read-time relation resolution — please add a short comment stating *why* it's safe rather than leaving it implicit.
6. **ctx identity keying.** Batching is keyed on ctx instance identity, so transactional contexts intentionally get isolated (unbatched) loaders. One line of comment saves the next reader the archaeology.

## 🟢 Nice-to-have
- `?? null` instead of `|| null` for the single-relation path (reads clearer).
- Restore the trailing newline at EOF.
- The comments are quite verbose — Vendure leans toward minimal comments; trim to the non-obvious bits.

Overall: solid direction, and the batching mechanics are correct. Fix the dependency placement + swap the monkeypatch for `RequestContextCacheService` + add the tests and this should be in good shape. Thanks again!

supermadu7:
Thanks for the thorough review, @grolmus — addressed everything below.

## 🔴 Blocking

1. **`dataloader` dependency placement** — moved to `packages/core/package.json` `dependencies` (pinned `^2.2.3`), removed from the root `package.json`.
2. **Monkeypatching removed** — dropped the `ctx as any` approach entirely. `CustomFieldRelationResolverService` now injects `RequestContextCacheService` and caches the loader under it, same pattern as `LocaleStringHydrator`. Confirmed `CacheModule` is already in the injector for this service (`api-internal-modules.ts`), so no module wiring changes were needed.
3. **Tests** — added two layers:
   - `custom-field-relation-resolver.service.spec.ts` (unit): asserts a batch of N concurrent loads resolves in exactly 2 repository calls (not 2N), correct list-vs-single shapes, per-row ordering, empty/null handling, and that a raw numeric id vs. a hydrated string id still matches correctly.
   - `custom-field-relations.e2e-spec.ts`: added a real-DB test resolving a list of products with distinct `single` relations, asserting each row's relation maps back to that row and not a neighbor's — the specific "row 3 must be row 3's" case you called out.

## 🟡 Should-fix

- Raw ids from `getRawMany()` are now normalized to strings before being used as `Map` keys, closing the cross-driver mismatch CodeRabbit flagged (this also covers the case where a driver returns numeric ids while hydrated entities/loader keys are strings — added a unit test for it specifically).
- Added a comment on why per-request DataLoader memoization is safe here (relation custom fields are read-only through this path, so there's no same-request write that could go stale).
- Added a comment on why a transactional `RequestContext` naturally gets its own unbatched loader (distinct instance → distinct `WeakMap` entry in `RequestContextCacheService`).

## 🟢 Nice-to-have

- Switched to `??` for the single-relation fallback.
- Trimmed comments down to the non-obvious bits.
- Trailing newline was already fine, double-checked.

## Test results

Full local run: unit spec passes (7/7), and the full e2e suite passes (1940 passed, 25 skipped, 0 failed) including the new `custom-field-relations.e2e-spec.ts` case.

## Open question

On your permission-gating test ask: that check (`userHasPermissionsOnCustomField`) happens upstream in `generate-resolvers.ts` before `resolveRelation` is ever called, so it's outside this service's contract — didn't want to fake coverage for something this class doesn't own. Happy to add e2e coverage for that upstream check if you think it's still missing, just flag it and I'll take care of it separately.

grolmus (approved):
Re-reviewed — everything from the previous round is addressed, and I verified each point against the current code rather than taking it on trust. Thanks for the thorough turnaround, @supermadu7.

**Blocking**
1. ✅ `dataloader` is now a `dependencies` entry in `packages/core/package.json` (`^2.2.3`) and gone from the root.
2. ✅ The `ctx as any` monkeypatch is gone — the service injects `RequestContextCacheService` and caches the loader under a namespaced key, matching the `LocaleStringHydrator` precedent. WeakMap keying on the ctx instance also gives transactional contexts their own unbatched loader for free.
3. ✅ Tests: the unit spec passes locally (7/7) and genuinely asserts the point of the PR — a batch of N concurrent loads resolves in a *constant* 2 repository calls, plus per-row ordering (including deliberately out-of-order mappings), list-vs-single shapes, null/empty handling, and the numeric-vs-string id case. The e2e adds a real-DB "row N maps to row N's relation" regression.

**Should-fix / nice-to-have** — raw id normalization (`String(...)`) with a dedicated type-mismatch test, the staleness/read-only comment, the transactional-ctx comment, and `?? null` are all in.

**On the permission-gating test** — agreed, and I confirmed it: `userHasPermissionsOnCustomField` runs in `generate-resolvers.ts` before `resolveRelation` is ever called, so it's genuinely upstream of this service's contract. No need to add coverage inside this class; the existing gate is unchanged by this PR. Fine to leave as-is.

Approving. Nice work on the batching mechanics.

PR #3023, opened 2024-08-20
Update DefaultMoneyStrategy.round() Logic

dfernandesbsolus:
# Description

This PR modifies the DefaultMoneyStrategy.round() method to implement the updated rounding logic as discussed in https://github.com/vendure-ecommerce/vendure/issues/3003. The changes aim to improve the accuracy and consistency of the rounding behavior.

# Breaking changes

NO

# Screenshots

Before: 
image

Now: 
image

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

PR #4929, opened 2026-07-07
docs(core): Fix description for inner workings of DefaultMoneyStrategy

SiebelsTim:
# Description

The documentation reflected the change in #3023. However, the related description was outdated.

# Checklist

📌 Always:

👍 Most of the time:
- [ ] ~~I have added or updated test cases~~
- [ ] ~~I have updated the README if needed~~

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4929"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786026018&installation_id=128408429&pr_number=4929&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4929&signature=de3bff0354b9798b9a1ed18f8b56ee24e4b118a253b5d118ebfe036fadccb8d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/docs/reference/typescript-api/money/money-strategy.mdx
+++ b/docs/docs/reference/typescript-api/money/money-strategy.mdx
@@ -94,10 +94,19 @@ UI as to how many decimal places to display.
 <MemberInfo kind="method" type={`(value: number, quantity?: number) => number`}   />
 
 Defines the logic used to round monetary values. For instance, the default behavior
-in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to multiply, then round the value.
+in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to multiply the unit value by the
+quantity and then round the result:
 
 ```ts
 return Math.round(value * quantity);
+```
+
+However, it may be desirable to instead round the unit amount _before_
+multiplying (this was the default prior to v3.1). In this case you can
+define a custom strategy with logic like this:
+
+```ts
+return Math.round(value) * quantity;
 ```
 
 
--- a/docs/docs/reference/typescript-api/money/money-strategy.mdx
+++ b/docs/docs/reference/typescript-api/money/money-strategy.mdx
@@ -94,14 +94,7 @@ UI as to how many decimal places to display.
 <MemberInfo kind="method" type={`(value: number, quantity?: number) => number`}   />
 
 Defines the logic used to round monetary values. For instance, the default behavior
-in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to round the value, then multiply.
-
-```ts
-return Math.round(value * quantity);
-```
-
-However, it may be desirable to instead round only _after_ the unit amount has been
-multiplied. In this case you can define a custom strategy with logic like this:
+in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to multiply, then round the value.
 
 ```ts
 return Math.round(value * quantity);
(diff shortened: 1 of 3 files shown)


mehringer68 (approved):
Neat ✅

packages/core/src/config/entity/money-strategy.ts
```diff
@@ -85,14 +85,7 @@ export interface MoneyStrategy extends InjectableStrategy {
...
-     * ```ts
-     * return Math.round(value * quantity);
-     * ```
-     *
-     * However, it may be desirable to instead round only _after_ the unit amount has been
-     * multiplied. In this case you can define a custom strategy with logic like this:
+     * in the {@link DefaultMoneyStrategy} is to multiply, then round the value.
      *
      * ```ts
      * return Math.round(value * quantity);
```
biggamesmallworld: Fix this branch's snippet rather than dropping it:

````suggestion
     * in the {@link DefaultMoneyStrategy} is to multiply the unit value by the
     * quantity and then round the result:
     *
     * ```ts
     * return Math.round(value * quantity);
     * ```
     *
     * However, it may be desirable to instead round the unit amount _before_
     * multiplying (this was the default prior to v3.1). In this case you can
     * define a custom strategy with logic like this:
     *
     * ```ts
     * return Math.round(value) * quantity;
````

**Review by biggamesmallworld — COMMENTED** [review]
Good catch — stale since #3023 updated the code block but not the prose, leaving two identical snippets presented as alternatives to each other.

One request: fix the "however, you may want to..." branch rather than deleting it. It's the only place the docs explain why this extension point exists, and the correct alternative (`Math.round(value) * quantity`) is the pre-v3.1 behaviour — exactly what a v3.0 upgrader needs to restore the old rounding. Suggestions inline for both files.

The `s3asset-storage-strategy.mdx` line bumps are unrelated drift from #4404; verified accurate, fine to leave.

File: docs/docs/reference/typescript-api/money/money-strategy.mdx [hunk 2]
```diff
@@ -94,14 +94,7 @@ UI as to how many decimal places to display.
...
-```ts
-return Math.round(value * quantity);
-```
-
-However, it may be desirable to instead round only _after_ the unit amount has been
-multiplied. In this case you can define a custom strategy with logic like this:
+in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to multiply, then round the value.
 
 ```ts
 return Math.round(value * quantity);
```
biggamesmallworld: Matching regen for the generated doc:

````suggestion
in the [DefaultMoneyStrategy](/reference/typescript-api/money/default-money-strategy#defaultmoneystrategy) is to multiply the unit value by the quantity and then round the result:

```ts
return Math.round(value * quantity);
```

However, it may be desirable to instead round the unit amount _before_ multiplying (this was the default prior to v3.1). In this case you can define a custom strategy with logic like this:

```ts
return Math.round(value) * quantity;
````

packages/core/src/config/entity/money-strategy.ts
SiebelsTim: sure!

docs/docs/reference/typescript-api/money/money-strategy.mdx
SiebelsTim: I've applied and generated it by hand.

biggamesmallworld (approved):

PR #4930, opened 2026-07-07
fix(core): Prevent stock oversell under concurrent checkout

### LINKED ISSUE #1738: Out of stock items can still be bought when inside an active order

grolmus:
Fixes #1738

## Problem

Stock can be oversold under concurrency. The saleable-stock **check** (`AddingItems → ArrangingPayment`) and stock **allocation** (`ArrangingPayment → PaymentSettled/Authorized`) run in **different transitions / transactions** with unlocked reads. Between them, other orders can deplete stock; the later allocation then oversells — or, with the default multi-location strategy, silently under-allocates a paid order. Additionally, `StockLevelService` writes were non-atomic read-modify-write with no clamp, allowing lost updates and negative stock values.

### Atomic, clamped stock writes
`StockLevelService.updateStockOnHandForLocation` / `updateStockAllocatedForLocation` now perform the read-modify-write under a pessimistic row lock and clamp `stockAllocated` at ≥ 0 (and never create a `StockLevel` with a negative value). This removes the lost-update race and prevents negative stock.

### Locked re-check at allocation
`StockMovementService.createAllocationsForOrderLines` acquires a `pessimistic_write` lock on the variant's `StockLevel` rows before computing allocation, so concurrent settlements can no longer consume the same units. The default location strategy already caps allocation to availability; the lock makes that re-check race-free (the locking read is also the freshness re-check, MySQL REPEATABLE-READ-safe).

### `StockShortfallEvent` (new)
When allocation cannot fully satisfy a line because stock was depleted since checkout, allocation is capped to what is available and a new `StockShortfallEvent` (order + per-line `requested`/`allocated`) is published. Because the payment may already be captured (async gateways), allocation is **not** hard-failed — plugins react to the event (refund, backorder, notify).

## Notes
- Allocation timing unchanged; no new config option; backward compatible.
- SQLite is single-writer and not a concurrency-safe production driver — the lock is a documented no-op there (a concurrent write surfaces as `SQLITE_BUSY`, not a silent lost update).
- Known edge (noted in code): a brand-new variant with no `StockLevel` row yet is not covered by the row lock; concurrent inserts would collide on the unique `(productVariantId, stockLocationId)` index — a safe rollback, not an oversell.

## Testing
- `stock-control.e2e-spec.ts`: sequential depletion (order A passes the checkout stock check, order B depletes and settles the stock, order A then settles) → allocation capped, no negative/oversold stock, `StockShortfallEvent` emitted — with exact-value assertions. Plus atomic/clamp coverage.
- `stock-control` and `stock-control-multi-location` suites pass.

Relates to real-time cart recalculation (#3510), handled separately.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4930"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786035489&installation_id=128408429&pr_number=4930&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4930&signature=9d0c78e0b26407a18d57b379ef3eeceba3cb42a9ea1abfe05cb2105f6cbf6311"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


mezzat11:
https://docs.saleor.io/developer/checkout/problems

grolmus:
## Before/after proof (e2e)

Reproduction of the ticket scenario: Order A passes the `ArrangingPayment` stock check (variant `stockOnHand=20`), Order B then depletes all 20 units and settles, then Order A settles. The same e2e test was run with the fix **reverted** vs. **with the fix**, each on a real rebuild.

**BEFORE** — lock + shortfall detection reverted:
```
× Stock control > stock shortfall detection ... > emits StockShortfallEvent and does not oversell ...
  → Timeout has occurred
TimeoutError: Timeout has occurred
```
`StockShortfallEvent` is never published — the shortfall on Order A's **paid** order is silent (the oversell / no-signal bug).

**AFTER** — with the fix: **PASS**
```
✓ Stock control > stock shortfall detection ... > emits StockShortfallEvent and does not oversell ...  ~900ms
```
Allocation is capped under the pessimistic lock (no oversell) and `StockShortfallEvent` is published so the merchant can react. Identical test, red without the fix / green with it — real rebuilds, not mocked.

grolmus:
Thanks @mezzat11 for sharing the [Saleor checkout problems](https://docs.saleor.io/developer/checkout/problems) reference — it's a nice pattern for how a checkout can surface issues to the storefront.

The root cause reported here is the oversell itself: the saleable-stock check (`AddingItems → ArrangingPayment`) and the allocation (`ArrangingPayment → PaymentSettled/Authorized`) ran in separate transactions with unlocked reads, and the stock writes were a non-atomic read-modify-write. This PR closes that:

- **Atomic, clamped stock writes** — no more lost updates / negative stock.
- **Pessimistic-locked re-check at allocation** — concurrent settlements can no longer consume the same units, so the saleable stock can't go negative.
- **`StockShortfallEvent`** — if stock was depleted between checkout and settlement (e.g. an async payment captured elsewhere), allocation is capped and the event is emitted so the merchant can react (refund/backorder), rather than the shortfall being silent.

On top of that, Vendure already refuses the transition to `ArrangingPayment` when there isn't enough saleable stock (the `arrangingPaymentRequiresStock` guard), which is the equivalent of Saleor's blocking `InsufficientStock` problem. The delivery-method side of Saleor's list (`DeliveryMethodInvalid` / `DeliveryMethodStale`) is handled in the related PR #4928 (OSS-274), which refuses checkout when the chosen shipping method is no longer eligible and recalculates stale pricing on read.

So the oversell reported in the issue should be resolved by this change. A fuller Saleor-style *queryable* `problems` field on the order (so the storefront can proactively surface these in the cart view before checkout) would be a nice separate enhancement — happy to open a follow-up issue for that if it's wanted, but it's out of scope for this correctness fix. 🙂

mezzat11:
> Thanks @mezzat11 for sharing the [Saleor checkout problems](https://docs.saleor.io/developer/checkout/problems) reference — it's a nice pattern for how a checkout can surface issues to the storefront.
> 
> The root cause reported here is the oversell itself: the saleable-stock check (`AddingItems → ArrangingPayment`) and the allocation (`ArrangingPayment → PaymentSettled/Authorized`) ran in separate transactions with unlocked reads, and the stock writes were a non-atomic read-modify-write. This PR closes that:
> 
> * **Atomic, clamped stock writes** — no more lost updates / negative stock.
> * **Pessimistic-locked re-check at allocation** — concurrent settlements can no longer consume the same units, so the saleable stock can't go negative.
> * **`StockShortfallEvent`** — if stock was depleted between checkout and settlement (e.g. an async payment captured elsewhere), allocation is capped and the event is emitted so the merchant can react (refund/backorder), rather than the shortfall being silent.
> 
> On top of that, Vendure already refuses the transition to `ArrangingPayment` when there isn't enough saleable stock (the `arrangingPaymentRequiresStock` guard), which is the equivalent of Saleor's blocking `InsufficientStock` problem. The delivery-method side of Saleor's list (`DeliveryMethodInvalid` / `DeliveryMethodStale`) is handled in the related PR #4928 (OSS-274), which refuses checkout when the chosen shipping method is no longer eligible and recalculates stale pricing on read.
> 
> So the oversell reported in the issue should be resolved by this change. A fuller Saleor-style _queryable_ `problems` field on the order (so the storefront can proactively surface these in the cart view before checkout) would be a nice separate enhancement — happy to open a follow-up issue for that if it's wanted, but it's out of scope for this correctness fix. 🙂

Thanks for clarifying — I just realized my comment was posted on the wrong PR.

My point about the Saleor-style `problems` pattern was meant for PR #4928, since it is more related to stale checkout/cart updates and how the storefront should surface those changes to the customer.

For this PR, I agree the focus is the stock correctness / oversell fix, and the transaction + allocation changes make sense.

I’ll move my suggestion to PR #4928. I still think having a queryable cart/order `problems` field would be valuable as a follow-up enhancement, so the storefront can proactively show messages like price updated, item unavailable, quantity reduced due to stock, or delivery method no longer eligible.

michaelbromley (dismissed):
Thorough review with local testing on sqljs, MariaDB and Postgres. The direction of this PR is right: the atomic clamped writes in `StockLevelService` fix a real lost-update bug, and the "cap allocation + emit `StockShortfallEvent`" design is sound, since the payment may already be captured and the base strategy derives Release/Sale/Cancellation quantities from actual `Allocation` records, so a capped allocation propagates consistently downstream. That invariant holds and is what makes the design viable.

However, I found three issues that need fixing before merge. I wrote e2e tests that demonstrate each one against this branch (patch at the bottom — it applies cleanly to `packages/core/e2e/stock-control.e2e-spec.ts` and all existing tests, including the ones added in this PR, still pass alongside it).

## 1. The locked re-check does not work on MySQL/MariaDB (blocking)

The lock in `createAllocationsForOrderLines` executes `getMany()` and discards the result. The actual availability decision happens afterwards in `MultiChannelStockLocationStrategy.forAllocation`, which reads stock levels through `getStockLevelsForVariant` — a plain, non-locking `find()` that is additionally memoized per-request via `RequestContextCacheService`.

On Postgres this happens to work: the default isolation level is READ COMMITTED, so the statement issued after the lock is granted sees the data the lock just waited on. On MySQL and MariaDB the default is REPEATABLE READ, where plain reads use the snapshot established by the transaction's *first* read — and the settlement transaction has already loaded the Order, Payment etc. long before the lock is taken. A locking read returns latest-committed data, but it does not refresh the snapshot used by later plain reads. So when two settlements genuinely overlap, the blocked transaction resumes after the winner commits, re-reads its stale snapshot, and allocates anyway.

The code comment asserts the opposite ("in MySQL REPEATABLE READ the locking read acts as the re-check") — that would only be true if the locked rows were actually used.

**Demonstration.** The `truly concurrent settlement` test in the attached patch: 5 units on hand, three customers each with 5 units in `ArrangingPayment`, all three `addPaymentToOrder` mutations fired via `Promise.all`. A payment handler that sleeps 250ms inside `createPayment` guarantees every settlement transaction has taken its snapshot before the first one reaches the lock.

- MariaDB 10.11: **fails with `stockAllocated: 15`, `stockOnHand: 5`** — a clean 3× oversell. The clean multiple of 5 shows the lock did serialize the writes (no lost update); each transaction simply re-read stale availability.
- Postgres 16: passes.

The existing e2e test in this PR settles Order B to completion before Order A settles, so both transactions get fresh snapshots and it passes on every driver — it exercises the capping and event wiring, but not the lock.

Note there is also no `stockAllocated <= stockOnHand` backstop anywhere — the new clamps only prevent negative values, so nothing catches this after the fact.

**Suggested fix:** make the locked read the read that feeds the strategy. Either perform the locking read inside the stock-level fetch that `forAllocation` uses (bypassing the request-context cache during allocation), or pass the locked rows through to the strategy. The request-context cache also needs care independently of the lock: two order lines for the same variant currently make the second line read pre-allocation cached values on every driver.

## 2. `StockShortfallEvent` can reference the wrong Order (blocking)

`orderForEvent` is captured from the first line and reused for the whole batch. But `Fulfillment.orders` is a many-to-many relation, and `default-fulfillment-process.ts` (line 100) re-allocates a cancelled fulfillment's lines in a single `createAllocationsForOrderLines` call — so the batch can mix lines belonging to different Orders. A shortfall on a later line's Order is then published under the first line's Order, and a plugin reacting to the event (refund, notify) acts on the wrong customer. The fulfillment-cancellation path is exactly the situation where shortfalls occur, because it re-allocates into stock that may have been sold in the meantime.

**Demonstration.** The `shortfall attribution with lines from multiple orders` test calls `createAllocationsForOrderLines` with one fully-allocatable line from Order A and one shortfalling line from Order B, mirroring what the fulfillment process does. It fails deterministically on every driver: the event carries Order A's code while the shortfall belongs to Order B's line.

**Suggested fix:** group the shortfalls by `orderLine.order` and publish one `StockShortfallEvent` per Order. `OrderLine.order` is an ordinary many-to-one, so per-line attribution is available.

## 3. Non-transactional callers now throw (blocking)

`setLock('pessimistic_write')` outside a transaction makes TypeORM throw `PessimisticLockTransactionRequiredError`, and the catch only handles `LockNotSupportedOnGivenDriverError` (the sqlite case). The `StockLevelService` methods are safe because they wrap themselves in `withTransaction`, but the allocation lock runs on the raw `ctx`. Core API paths are transactional via the `@Transaction` decorator, but plugin code calling e.g. `orderService.transitionToState` from a script or job-queue processor without a transaction worked before this PR and now throws — TypeORM's transaction check is driver-independent, so this affects every database including sqlite.

**Demonstration.** The `allocation outside a transaction` test calls the service with a context from `RequestContextService.create()` and fails on every driver with `PessimisticLockTransactionRequiredError`.

**Suggested fix:** wrap the body of `createAllocationsForOrderLines` in `this.connection.withTransaction(ctx, ...)`. This fixes the crash and also closes a correctness hole: for a non-transactional caller the lock would otherwise be acquired and released per-statement, protecting nothing.

## Non-blocking observations

- **Deadlocks / lock ordering:** lines lock variants in order-line order, so two orders containing variants (v1, v2) and (v2, v1) can deadlock under load. Sorting the lines by `productVariantId` before iterating gives a consistent global lock order cheaply. Related: when I ran the concurrent test as part of the full suite on MariaDB, one settlement intermittently failed with `OrderPaymentStateError` rather than overselling, which looks like the `TransactionWrapper` deadlock-retry path misbehaving under contention — worth investigating while you are in here.
- **Merchant visibility:** if no plugin subscribes to `StockShortfallEvent`, an under-allocated paid order is essentially invisible in the admin UI. Consider also writing an order history entry when a shortfall occurs.
- **Legacy strategy:** `DefaultStockLocationStrategy.forAllocation` never caps to availability, so for users of that strategy the shortfall detection never fires and allocation-time oversell continues unchanged. Worth documenting on the event and in the strategy docs.
- **Movement history:** a partially-allocated line's `Sale` record stores the full line quantity while the stock levels move by the allocation-derived amount, so the movement history can diverge from the stock levels. This is pre-existing code, but shortfalls make it reachable.
- Please also consider adopting the concurrent settlement test (or a variant of it) into the suite — it is the only test here that actually exercises the locking behaviour, and it can only do so on the real-database CI jobs, since sql.js serializes everything onto one connection.

## Reproducing

From `packages/core` on this branch with the patch applied:

```bash
# deterministic failures (issues 2 and 3), no database needed:
bun run e2e stock-control.e2e-spec

# issue 1 — fails on MariaDB/MySQL (REPEATABLE READ), passes on Postgres:
DB=mariadb PACKAGE=core bunx vitest --config ../../e2e-common/vitest.config.mts --run stock-control.e2e-spec -t "truly concurrent"
CI=true E2E_POSTGRES_PORT=5433 DB=postgres PACKAGE=core bunx vitest --config ../../e2e-common/vitest.config.mts --run stock-control.e2e-spec -t "truly concurrent"
```

<details>
<summary>Test patch (packages/core/e2e/stock-control.e2e-spec.ts)</summary>

```diff
diff --git a/packages/core/e2e/stock-control.e2e-spec.ts b/packages/core/e2e/stock-control.e2e-spec.ts
index 18389a26a..bbe7602a4 100644
--- a/packages/core/e2e/stock-control.e2e-spec.ts
+++ b/packages/core/e2e/stock-control.e2e-spec.ts
@@ -5,6 +5,7 @@ import {
     type CreateAddressInput,
     ErrorCode,
     GlobalFlag,
+    LanguageCode,
     StockMovementType,
 } from '@vendure/common/lib/generated-types';
 import { pick } from '@vendure/common/lib/pick';
@@ -15,10 +16,19 @@ import {
     mergeConfig,
     type Order,
     type OrderState,
+    PaymentMethodHandler,
     type RequestContext,
+    RequestContextService,
+    StockMovementService,
     StockShortfallEvent,
+    TransactionalConnection,
 } from '@vendure/core';
-import { createErrorResultGuard, createTestEnvironment, type ErrorResultGuard } from '@vendure/testing';
+import {
+    createErrorResultGuard,
+    createTestEnvironment,
+    type ErrorResultGuard,
+    SimpleGraphQLClient,
+} from '@vendure/testing';
 import path from 'path';
 import { firstValueFrom, timeout } from 'rxjs';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
@@ -83,27 +93,53 @@ class TestOrderPlacedStrategy extends DefaultOrderPlacedStrategy {
     }
 }
 
+/**
+ * Settles immediately, but holds the settlement transaction open briefly. When several
+ * orders settle concurrently, this ensures they have all started their settlement
+ * transactions (and taken their DB snapshots) before the first one commits, which is
+ * the timing needed to exercise the allocation re-check under concurrency.
+ */
+const delayedSettlePaymentMethod = new PaymentMethodHandler({
+    code: 'delayed-settle-payment-method',
+    description: [{ languageCode: LanguageCode.en, value: 'Delayed settle payment method' }],
+    args: {},
+    createPayment: async (ctx, order, amount) => {
+        await new Promise(resolve => setTimeout(resolve, 250));
+        return {
+            amount,
+            state: 'Settled' as const,
+            transactionId: 'delayed-12345',
+        };
+    },
+    settlePayment: () => ({
+        success: true,
+    }),
+});
+
 describe('Stock control', () => {
-    const { server, adminClient, shopClient } = createTestEnvironment(
-        mergeConfig(testConfig(), {
-            paymentOptions: {
-                paymentMethodHandlers: [testSuccessfulPaymentMethod, twoStagePaymentMethod],
-            },
-            orderOptions: {
-                orderPlacedStrategy: new TestOrderPlacedStrategy(),
-            },
-            customFields: {
-                Order: [
-                    {
-                        name: 'test1557',
-                        type: 'boolean',
-                        defaultValue: false,
-                    },
-                ],
-                OrderLine: [{ name: 'customization', type: 'string', nullable: true }],
-            },
-        }),
-    );
+    const testEnvConfig = mergeConfig(testConfig(), {
+        paymentOptions: {
+            paymentMethodHandlers: [
+                testSuccessfulPaymentMethod,
+                twoStagePaymentMethod,
+                delayedSettlePaymentMethod,
+            ],
+        },
+        orderOptions: {
+            orderPlacedStrategy: new TestOrderPlacedStrategy(),
+        },
+        customFields: {
+            Order: [
+                {
+                    name: 'test1557',
+                    type: 'boolean',
+                    defaultValue: false,
+                },
+            ],
+            OrderLine: [{ name: 'customization', type: 'string', nullable: true }],
+        },
+    });
+    const { server, adminClient, shopClient } = createTestEnvironment(testEnvConfig);
 
     const orderGuard: ErrorResultGuard<
         UpdatedOrderFragment | TestOrderFragment | TestOrderWithPaymentsFragment
@@ -145,6 +181,10 @@ describe('Stock control', () => {
                         name: twoStagePaymentMethod.code,
                         handler: { code: twoStagePaymentMethod.code, arguments: [] },
                     },
+                    {
+                        name: delayedSettlePaymentMethod.code,
+                        handler: { code: delayedSettlePaymentMethod.code, arguments: [] },
+                    },
                 ],
             },
             productsCsvPath: path.join(__dirname, 'fixtures/e2e-products-stock-control.csv'),
@@ -1540,6 +1580,219 @@ describe('Stock control', () => {
             expect(shortfallEvent.shortfalls[0].allocated).toBe(0);
         });
     });
+
+    // OSS-94: the allocation re-check must hold when settlements genuinely overlap,
+    // not only when one settlement completes before the next begins.
+    describe('truly concurrent settlement', () => {
+        // sql.js executes all queries on a single connection, so transactions cannot
+        // genuinely overlap there and the locking behaviour cannot be exercised.
+        // The CI database matrix (postgres/mysql/mariadb) is where this test has meaning.
+        const isRealDb = (process.env.DB ?? 'sqljs') !== 'sqljs';
+
+        it.runIf(isRealDb)(
+            'concurrent settlements never allocate more than stockOnHand',
+            async () => {
+                const variantId = 'T_4'; // Laptop 15 inch 16GB
+                const quantityPerOrder = 5;
+                const customers = [
+                    'hayden.zieme12@hotmail.com',
+                    'trevor_donnelly96@hotmail.com',
+                    'marques.sawayn@hotmail.com',
+                ];
+
+                // Make exactly one order's worth of stock saleable, accounting for
+                // any allocations left over from earlier tests.
+                const product = await getProductWithStockMovement('T_1');
+                const variant = product!.variants.find(v => v.id === variantId)!;
+                const stockOnHand = variant.stockAllocated + quantityPerOrder;
+                await adminClient.query(updateProductVariantsDocument, {
+                    input: [
+                        {
+                            id: variantId,
+                            stockOnHand,
+                            trackInventory: GlobalFlag.TRUE,
+                            useGlobalOutOfStockThreshold: false,
+                            outOfStockThreshold: 0,
+                        },
+                    ],
+                });
+
+                // Each customer needs their own client, since a client holds a single
+                // session. All of them reach ArrangingPayment: the stock check passes
+                // for each because nothing is allocated until settlement.
+                const clients: SimpleGraphQLClient[] = [];
+                for (const emailAddress of customers) {
+                    const client = new SimpleGraphQLClient(
+                        testEnvConfig,
+                        `http://localhost:${testEnvConfig.apiOptions.port}/${testEnvConfig.apiOptions.shopApiPath}`,
+                    );
+                    await client.asUserWithCredentials(emailAddress, 'test');
+                    const { addItemToOrder } = await client.query(addItemToOrderDocument, {
+                        productVariantId: variantId,
+                        quantity: quantityPerOrder,
+                    });
+                    orderGuard.assertSuccess(addItemToOrder);
+                    const arrangingPaymentOrderId = await proceedToArrangingPayment(client);
+                    expect(
+                        arrangingPaymentOrderId,
+                        `transition to ArrangingPayment failed for ${emailAddress}`,
+                    ).toBeDefined();
+                    clients.push(client);
+                }
+
+                // Settle all orders at once. The delayed payment handler keeps each
+                // settlement transaction open long enough that they all take their DB
+                // snapshots before the first one commits.
+                const results = await Promise.all(
+                    clients.map(client => addPaymentToOrder(client, delayedSettlePaymentMethod)),
+                );
+                for (const result of results) {
+                    orderGuard.assertSuccess(result);
+                }
+
+                // Only quantityPerOrder units were saleable, so at most one order's
+                // quantity may have been allocated on top of the pre-existing
+                // allocations. Anything more is an oversell.
+                const productAfter = await getProductWithStockMovement('T_1');
+                const variantAfter = productAfter!.variants.find(v => v.id === variantId)!;
+                expect(variantAfter.stockOnHand).toBe(stockOnHand);
+                expect(variantAfter.stockAllocated).toBeLessThanOrEqual(variantAfter.stockOnHand);
+            },
+            60_000,
+        );
+    });
+
+    // OSS-94: a StockShortfallEvent must reference the Order which the shortfalling
+    // line belongs to. Fulfillments can span multiple Orders (Fulfillment.orders is
+    // many-to-many), and default-fulfillment-process re-allocates a cancelled
+    // fulfillment's lines in a single createAllocationsForOrderLines() call, so the
+    // batch can mix lines of different Orders.
+    describe('shortfall attribution with lines from multiple orders', () => {
+        const rawId = (id: string) => id.replace(/^T_/, '');
+
+        it('attributes each shortfall to the order that the line belongs to', async () => {
+            const inStockVariantId = 'T_2'; // Laptop 15 inch 8GB
+            const depletedVariantId = 'T_7'; // Curvy Monitor 32 inch
+
+            const laptop = await getProductWithStockMovement('T_1');
+            const inStockVariant = laptop!.variants.find(v => v.id === inStockVariantId)!;
+            const monitor = await getProductWithStockMovement('T_2');
+            const depletedVariant = monitor!.variants.find(v => v.id === depletedVariantId)!;
+            await adminClient.query(updateProductVariantsDocument, {
+                input: [
+                    {
+                        id: inStockVariantId,
+                        stockOnHand: inStockVariant.stockAllocated + 10,
+                        trackInventory: GlobalFlag.TRUE,
+                        useGlobalOutOfStockThreshold: false,
+                        outOfStockThreshold: 0,
+                    },
+                    {
+                        id: depletedVariantId,
+                        stockOnHand: depletedVariant.stockAllocated + 3,
+                        trackInventory: GlobalFlag.TRUE,
+                        useGlobalOutOfStockThreshold: false,
+                        outOfStockThreshold: 0,
+                    },
+                ],
+            });
+
+            // Order A (hayden): a line which can be fully allocated
+            await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+            const { addItemToOrder: orderA } = await shopClient.query(addItemToOrderDocument, {
+                productVariantId: inStockVariantId,
+                quantity: 1,
+            });
+            orderGuard.assertSuccess(orderA);
+
+            // Order B (trevor): a line whose stock will be depleted before allocation
+            await shopClient.asUserWithCredentials('trevor_donnelly96@hotmail.com', 'test');
+            const { addItemToOrder: orderB } = await shopClient.query(addItemToOrderDocument, {
+                productVariantId: depletedVariantId,
+                quantity: 3,
+            });
+            orderGuard.assertSuccess(orderB);
+
+            // Deplete order B's variant: all remaining stock is already allocated elsewhere
+            await adminClient.query(updateProductVariantsDocument, {
+                input: [
+                    {
+                        id: depletedVariantId,
+                        stockOnHand: depletedVariant.stockAllocated,
+                    },
+                ],
+            });
+
+            const stockMovementService = server.app.get(StockMovementService);
+            const requestContextService = server.app.get(RequestContextService);
+            const connection = server.app.get(TransactionalConnection);
+            const eventBus = server.app.get(EventBus);
+            const ctx = await requestContextService.create({ apiType: 'admin' });
+
+            const events: StockShortfallEvent[] = [];
+            const subscription = eventBus.ofType(StockShortfallEvent).subscribe(e => events.push(e));
+            try {
+                // Reproduces what default-fulfillment-process does when a Fulfillment
+                // spanning two Orders is cancelled: one re-allocation call with a
+                // batch of lines from different Orders.
+                await connection.withTransaction(ctx, txCtx =>
+                    stockMovementService.createAllocationsForOrderLines(txCtx, [
+                        { orderLineId: rawId(orderA.lines[0].id), quantity: 1 },
+                        { orderLineId: rawId(orderB.lines[0].id), quantity: 3 },
+                    ]),
+                );
+                // Events are published after the transaction commits
+                await new Promise(resolve => setTimeout(resolve, 500));
+
+                // The only shortfall is on Order B's line, so the event carrying that
+                // shortfall must reference Order B.
+                const shortfallEvent = events.find(e =>
+                    e.shortfalls.some(s => String(s.orderLineId) === rawId(orderB.lines[0].id)),
+                );
+                expect(shortfallEvent).toBeDefined();
+                expect(shortfallEvent!.order.code).toBe(orderB.code);
+            } finally {
+                subscription.unsubscribe();
+            }
+        });
+    });
+
+    // OSS-94: service methods must remain callable with a RequestContext that is not
+    // bound to a transaction (job-queue processors and scripts do this). A pessimistic
+    // lock outside a transaction makes TypeORM throw PessimisticLockTransactionRequiredError.
+    describe('allocation outside a transaction', () => {
+        it('createAllocationsForOrderLines works with a non-transactional context', async () => {
+            const variantId = 'T_2'; // Laptop 15 inch 8GB
+            const product = await getProductWithStockMovement('T_1');
+            const variant = product!.variants.find(v => v.id === variantId)!;
+            await adminClient.query(updateProductVariantsDocument, {
+                input: [
+                    {
+                        id: variantId,
+                        stockOnHand: variant.stockAllocated + 10,
+                        trackInventory: GlobalFlag.TRUE,
+                    },
+                ],
+            });
+
+            await shopClient.asUserWithCredentials('marques.sawayn@hotmail.com', 'test');
+            const { addItemToOrder: order } = await shopClient.query(addItemToOrderDocument, {
+                productVariantId: variantId,
+                quantity: 1,
+            });
+            orderGuard.assertSuccess(order);
+
+            const stockMovementService = server.app.get(StockMovementService);
+            const requestContextService = server.app.get(RequestContextService);
+            const ctx = await requestContextService.create({ apiType: 'admin' });
+
+            await expect(
+                stockMovementService.createAllocationsForOrderLines(ctx, [
+                    { orderLineId: order.lines[0].id.replace(/^T_/, ''), quantity: 1 },
+                ]),
+            ).resolves.toBeDefined();
+        });
+    });
 });
 
 const updateStockOnHandDocument = graphql(
```

</details>

grolmus:
Thanks for the exceptionally thorough review, Michael — and for the test patch, which I've adopted. All three blockers are fixed in `89ea2bff2` / `61eb27e5e`.

## 1. Locked re-check on MySQL/MariaDB

Moved the lock to where the decision actually reads. The built-in `MultiChannelStockLocationStrategy.forAllocation` now reads stock via a new `getLockedStockLevelsForVariant` — a `pessimistic_write`, **uncached** read (with the sqlite `LockNotSupportedOnGivenDriverError` fallback) — so the locked read *is* the read that feeds availability, rather than a discarded pre-read followed by a snapshot/`RequestContextCache` read. The old discarded lock in `createAllocationsForOrderLines` is removed. Bypassing the request cache also closes the second-line-same-variant staleness you noted.

Verified against your `truly concurrent` test on **MariaDB 10.11**: it fails without this change (`expected 15 to be less than or equal to 5` — the exact 3× oversell) and passes with it.

## 2. Shortfall attribution

Shortfalls are grouped by `orderLine.order` and one `StockShortfallEvent` is published per Order (`'order'` relation now loaded for every line). Your `shortfall attribution with lines from multiple orders` test passes.

## 3. Non-transactional callers

`createAllocationsForOrderLines` now wraps its body in `this.connection.withTransaction`, so the locks are valid and the `PessimisticLockTransactionRequiredError` is gone. Nested transactions join the outer one, so the `@Transaction` core paths are unaffected. Your `allocation outside a transaction` test passes on every driver.

## Non-blocking

- **Deadlock lock-ordering** — done. Order lines are sorted by `productVariantId` before the sequential allocation loop, so concurrent batches acquire the per-variant locks in a consistent global order.
- **Legacy `DefaultStockLocationStrategy`** — confirmed its `forAllocation` allocates the full quantity without reading stock, so it never made an availability decision and the removed service-level lock protected nothing for it; behaviour there is unchanged. The locked re-check lives in `MultiChannelStockLocationStrategy`, which is where it matters.
- **Merchant visibility (order-history entry) and the `Sale`-vs-stock movement-history divergence** — left as follow-ups; happy to open OSS tickets if you'd like them tracked.

## Verification

Adopted your test patch verbatim. Full `stock-control` suite green on sqljs (41 passed, the concurrent test correctly gated to a real DB via `it.runIf`); the three new tests green on MariaDB 10.11; core typechecks.

One behavioural note worth a changelog line: `forAllocation` now requires a transaction on pg/mysql — third-party code invoking the strategy / `getAllocationLocations` outside a transaction would hit `PessimisticLockTransactionRequiredError`. Intended (allocation should be transactional), and documented in-code.

mezzat11:
@grolmus if it can be merged soon this critical issue make negative stock

grolmus:
@michaelbromley gentle nudge on this one. All three blockers from your review are addressed (`89ea2bff2` / `61eb27e5e`, and I adopted your test patch), and I've now merged the latest `minor` in so the branch is current — it had drifted ~164 commits behind. CI is green across all four databases (sqljs / mariadb / mysql / postgres); the one mysql red along the way was a flaky `database-transactions` isolation test, green on rerun and unrelated to this change.

Since this is `T3: Systemic` and still holds your `CHANGES_REQUESTED`, it needs your re-review to land — re-requested you. @mezzat11 (and others) are hitting the negative-stock oversell in production, so it'd be good to get it in when you have a moment.

biggamesmallworld:
@michaelbromley I re-verified your three blockers against the current head (`29687aa1`) rather than taking the fix commits at face value. Summary: all three are genuinely addressed, and your MySQL/MariaDB diagnosis in point 1 was exactly right. Details below so you can re-review quickly.

## One gotcha that nearly gave me a false result

The core e2e tests resolve `@vendure/core` through the `node_modules` symlink to `packages/core`, which means they run the built `dist`, not `src`. My first few runs used a stale `dist` left over from another branch, so every "different source state" I tested was actually executing identical, pre-PR code, and the concurrency test passed everywhere. Anyone reproducing this needs `bun run build` in `packages/common` and then `packages/core` between each source state, otherwise the results are meaningless.

## Approach

For each result below I compared two builds:

- **fix**: current PR head, `29687aa1`
- **pre-fix**: same branch with only `stock-movement.service.ts` and `multi-channel-stock-location-strategy.ts` reverted to `f96010ac`, their state at the time of your review

Database was MariaDB 10.11 in Docker, isolation confirmed as `REPEATABLE-READ`, with MariaDB's general query log enabled so I could confirm the `FOR UPDATE` statements were actually reaching the server.

## Your point 1 premise, confirmed at the driver level

Before touching Vendure I checked the InnoDB behaviour directly with a two-connection script, because the whole argument rests on it. Inside a single REPEATABLE READ transaction that has already established a snapshot:

```
A first plain read (establishes snapshot): 0
B commits allocated=5
A locking read        : 5
A plain read AFTER lock: 0   <-- stale
```

So a locking read returns latest-committed data but does not refresh the snapshot used by later plain reads, precisely as you described. That is why the discarded pre-read lock protected nothing.

## Results

| Scenario | pre-fix | fix |
|---|---|---|
| Your test: 3 concurrent settlements, 5 each, `stockOnHand` 5 | `stockAllocated: 15` (3x oversell) | 5 |
| 8 concurrent settlements, 5 each, `stockOnHand` 5 | `stockAllocated: 40` (8x oversell), no errors surfaced | 5, zero failed requests |
| Two order lines for the same variant in one order, 6 requested, 5 available | 6 allocated (oversell by 1) | 5, capped |
| Two orders holding two variants in opposite order, settling concurrently | 1 failure: `Error: SAVEPOINT typeorm_2 does not exist` | 0 failures |

Your reported `stockAllocated: 15` reproduces exactly. Points 2 and 3 are covered by your own tests, adopted verbatim here, and they are driver independent and green on all four CI database jobs.

Two of those rows are scenarios beyond your patch that seemed worth probing:

**Same variant across two order lines.** This is the request-context cache staleness you raised as an aside in point 1, and it oversells pre-fix with no concurrency at all: one order, one request, deterministic on every driver including sql.js. Setup is two lines for the same variant via different `customization` custom field values, stock reduced to 5 after reaching `ArrangingPayment`, then settle. Pre-fix allocates 3 + 3 because line 2 reads line 1's cached pre-allocation values. This is the cheapest possible regression test in the whole set and there is still no coverage for it, so it might be worth asking for.

**Opposite lock ordering.** This looks like the intermittent failure you saw under contention on MariaDB. It reproduces on demand pre-fix with two orders referencing two variants in opposite order, and the `productVariantId` sort clears it. So that loose end appears closed, though the underlying `TransactionWrapper` retry fragility is only being avoided rather than fixed.

## One caveat on what the concurrency test actually guards

On a build with none of this PR present, the test passes at `stockAllocated: 5`. The general log shows why: three separate `UPDATE stock_level SET stockAllocated = 5` statements, one per settlement. All three orders allocated 5 units each and the lost updates overwrote one another, so the stored total stayed at 5 while three customers were sold the same stock.

In other words the assertion only exposes the oversell once the atomic clamped writes from `9af90303` are in place. It is a valid guard against regressing the locking behaviour going forward, but it is not a from-scratch reproduction of the original bug on `minor`, and it supports your observation that there is no `stockAllocated <= stockOnHand` backstop anywhere in the code.

The extra scenarios above are not committed anywhere, they were throwaway local tests.

biggamesmallworld (approved):
Approving. I verified Michael's three blockers against the current head rather than trusting the fix commits, and all three hold up under real concurrent load on MariaDB. Details are in a separate comment below.

Two things I would like to see addressed before this merges. I have raised them as their own review threads so they can be tracked and resolved independently:

1. A shortfall is currently invisible to the merchant, on `stock-movement.service.ts` where `StockShortfallEvent` is published.
2. The `stockOnHand` clamp can silently contradict the ledger, on `stock-level.service.ts`.

Of the two, the second has the wider blast radius, since it can leave `stock_movement` and `StockLevel` permanently disagreeing on a backorder-enabled variant.

## Smaller things, take or leave

- The `stockAllocated stays >= 0` and `stockOnHand stays >= 0` clamp tests never drive a value below zero, so they pass with or without the clamp. Calling `updateStockAllocatedForLocation(ctx, variantId, locationId, -99)` directly and asserting `0` would actually test it.
- The concurrency test ends on `toBeLessThanOrEqual`, which also passes if nothing allocated or everything rolled back. It does genuinely catch the regression, I confirmed that, but asserting the exact expected allocation would make it airtight.
- `requestContextCache` in `MultiChannelStockLocationStrategy` has no readers left now that `getStockLevelsForVariant` is gone. It is `protected`, so removing it is technically breaking for subclasses, which is fine on a `minor` target but worth doing deliberately.
- Docs have not been regenerated for the new public `StockShortfallEvent` and `StockShortfall` types.

Thanks for the very thorough iteration on this one, and for adopting the review tests verbatim. It made re-verification much faster.

packages/core/src/service/services/stock-movement.service.ts
```diff
@@ -154,45 +155,98 @@ export class StockMovementService {
...
-        if (savedAllocations.length) {
-            await this.eventBus.publish(new StockMovementEvent(ctx, savedAllocations));
-        }
-        return savedAllocations;
+            const savedAllocations = await this.connection.getRepository(txCtx, Allocation).save(allocations);
+            if (savedAllocations.length) {
+                await this.eventBus.publish(new StockMovementEvent(txCtx, savedAllocations));
+            }
+            for (const { order, shortfalls } of shortfallsByOrder.values()) {
+                await this.eventBus.publish(new StockShortfallEvent(txCtx, order, shortfalls));
```
biggamesmallworld, line 260: **A shortfall is currently invisible to the merchant.**

`StockShortfallEvent` has no default subscriber, no log line, and no order history entry. If nobody writes a plugin, an under-allocated paid order looks completely normal in the admin UI.

Worth being explicit about the direction of travel. Before this change, an oversell announced itself through stock levels that stopped making sense, and that is usually how someone noticed. After this change the allocation is quietly capped, `orderLine.quantity` still reads the full amount, and every number on the screen looks fine right up until fulfilment cannot be satisfied. We are trading a visible wrong state for an invisible one, which is the right trade only if we add the signal back.

A `Logger.warn` with the order code, variant id, requested and allocated quantities, plus an order history entry on the affected order, would cover it. Small addition, and it makes the feature self-documenting for anyone who has not read this PR.

Michael raised the same point independently under merchant visibility in his review.

Please resolve this thread once addressed, or reply if you would rather track it as a separate issue and I will resolve it myself.

packages/core/src/service/services/stock-level.service.ts
```diff
@@ -82,56 +83,94 @@ export class StockLevelService {
...
+                // making this an edge case; a unique-constraint violation safely rolls back the
+                // transaction rather than producing an oversell.
+                stockLevel = await repo.findOne({ where: { productVariantId, stockLocationId } });
+            }
+            if (!stockLevel) {
+                await repo.save(
+                    new StockLevel({
+                        productVariantId,
+                        stockLocationId,
+                        stockOnHand: Math.max(0, change),
```
biggamesmallworld: **This clamp can silently contradict the ledger.**

Vendure deliberately supports negative `stockOnHand` when a variant has a negative `outOfStockThreshold`, which is how backorders work. If a variant has no `StockLevel` row yet and receives a negative adjustment, we now write `0` here while the `StockAdjustment` row that was just saved still records the full negative delta. The ledger and the running total disagree, permanently, with nothing logged.

It is a narrow path, and the existing negative stock tests do not catch it because they run against variants whose row already exists. But it is a silent divergence between `stock_movement` and `StockLevel`, which is the exact class of problem this PR sets out to fix, so it feels wrong to introduce a new instance of it here.

Two ways out, either is fine by me:

1. Drop the clamp on this branch and let the negative value through, since it is legitimate for backorders.
2. Keep the clamp and `Logger.warn` when it actually fires, including variant id, location id and the attempted delta.

The same applies to the `stockAllocated` clamp further down. Clamping is fine, clamping without a trace is not.

Please resolve this thread once addressed.

biggamesmallworld (commented):
Two blocking follow-ups, split out as their own threads so they can be resolved independently. Both are small. Everything else from my approval stands.

packages/core/src/service/services/stock-movement.service.ts
grolmus: Added a `Logger.warn` per shortfall next to the `StockShortfallEvent` (order code, variant id, requested, allocated) in `createAllocationsForOrderLines` — so a paid-but-under-allocated Order no longer looks normal in the logs.

On the **order history entry**: I'd like to take you up on tracking that separately. A merchant-visible history entry means a new public `HistoryEntryType` enum value (`HistoryEntry.type` is `HistoryEntryType!`, not a free string), plus its `OrderHistoryEntryData` contract and dashboard rendering — an additive-but-public API surface I'd rather not fold into this fix without its own decision. Happy for you to pick it up as a follow-up; the `StockShortfallEvent` + warning cover the immediate visibility gap in the meantime.

packages/core/src/service/services/stock-level.service.ts
grolmus: Fixed. The new-row `stockOnHand` branch now writes the adjustment delta itself instead of `Math.max(0, change)`, so a negative initial value (legit for backorders against a negative `outOfStockThreshold`) is preserved and the row stays consistent with the `StockAdjustment` ledger — matching the existing-row branch, which already let negatives through.

The `stockAllocated` clamp stays (a negative allocation is meaningless — it would mean more was released than allocated), but it now `Logger.warn`s with variant id, location id and the attempted delta when it actually fires, so the clamp is no longer silent.

biggamesmallworld (approved):

biggamesmallworld (approved):

PR #4933, opened 2026-07-08
feat(core): expand telemetry to schema v2 with heartbeat and new signals

dlhck:
## Description

Expands the anonymous telemetry payload to schema v2, closing the largest signal gaps in the current collection: send cadence, order lifecycle, internationalization breadth, runtime environment, security posture, and strategy customization.

### What's new

**Cadence & payload metadata**
- Every payload now carries `schemaVersion: 2`, `sendReason` (`startup` | `heartbeat`) and `uptimeSeconds`.
- In addition to the existing one-off startup send (5s after bootstrap), a repeating 24h heartbeat send is scheduled. The interval is `unref()`ed so it never keeps the process alive, is cleared on shutdown, and all existing guards (worker process, `VENDURE_DISABLE_TELEMETRY`, CI detection) are re-checked on every send. This makes uptime, retention and churn measurable — previously an installation only reported when it restarted.

**`runtime` section** (SystemInfoCollector)
- `runtimeType` (node/bun/deno), `packageManager` (parsed from `npm_config_user_agent`, same approach as `@vendure/create`), `tsNode`, `cpuCount`, `totalMemoryGb`.

**`metrics.orders`** (DatabaseCollector)
- Range-bucketed order lifecycle breakdown: `placed`, `active`, `draft`, `placedLast30d`, and `byType` (Regular/Seller/Aggregate). Total Order count is dominated by abandoned carts; this separates real trading activity and surfaces draft-order and multi-vendor adoption directly.

**`metrics.i18n`** (DatabaseCollector)
- Distinct language and currency code counts across all Channels (union of default + available codes), plus a derived `features.multiCurrency` flag.

**Config posture fields** (ConfigCollector)
- `apiIntrospectionEnabled`, `apiPlaygroundEnabled`, `apiDebugEnabled`, `trustProxyEnabled`, `corsWildcardOrigin`, `tokenMethods`, `requireVerification`, `authDisabled`, `defaultSuperadminCredentials`, `cookieSecure`, `cookieSameSite`, `settingsStoreFieldCount`.
- `customizedStrategies`: dotted config paths (e.g. `orderOptions.guestCheckoutStrategy`) of single-strategy fields whose live class differs from `defaultConfig` — compared dynamically, so defaults stay the single source of truth.

### Privacy

No raw values are collected anywhere: no hostnames, CORS origins, API paths, secrets, credentials, DB details, filesystem paths, or custom strategy class names. `defaultSuperadminCredentials` is a boolean comparison against the well-known defaults; the configured values are never read into the payload. Entity counts remain range-bucketed. Everything sensitive reduces to booleans, short enums, counts, or config paths.

All new payload fields are optional on the receiving side, so this remains backward/forward compatible with the collection endpoint.

### Testing

- 375 unit tests passing across the telemetry suite (+42 new), covering runtime user-agent parsing, posture booleans, `defaultSuperadminCredentials`, `customizedStrategies` (default vs customized), order-metric bucketing and `byType`, i18n counting, `multiCurrency`, and heartbeat scheduling (second send after 24h, interval unref'ed, cleared on shutdown, not scheduled when telemetry is disabled).
- `tsc --noEmit` clean; eslint/prettier clean on all changed files.
- Order/i18n queries use dialect-agnostic TypeORM (`count(where)`, query-builder `groupBy`) — no raw SQL.

## Breaking changes

None. Telemetry remains opt-out via `VENDURE_DISABLE_TELEMETRY` and disabled in CI; the only behavioral change is the added daily heartbeat send.

## Checklist

- [x] Tests added for new behavior
- [x] No breaking changes

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4933"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786095157&installation_id=128408429&pr_number=4933&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4933&signature=df6876eb95b73dfffb49679345f3dc92517966a26a2fd8ca9c2523ae43199d26"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/core/src/telemetry/collectors/database.collector.ts
```diff
@@ -31,12 +47,137 @@ export class DatabaseCollector {
...
+
+            // A handful of small, indexed queries — run them in parallel (unlike
+            // the larger entity-count sweep).
+            const [placed, active, draft, placedLast30d, byType] = await Promise.all([
+                this.safeBucket(() => repo.count({ where: { orderPlacedAt: Not(IsNull()) } })),
+                this.safeBucket(() => repo.count({ where: { active: true } })),
+                this.safeBucket(() => repo.count({ where: { state: 'Draft' as any } })),
+                this.safeBucket(() => repo.count({ where: { orderPlacedAt: MoreThanOrEqual(since) } })),
+                this.collectOrdersByType(repo),
+            ]);
```
michaelbromley: **Must fix:** the comment says "small, indexed queries", but `Order` has no index on `active`, `state`, or `type` — only `code`, `orderPlacedAt`, `aggregateOrder` and `customer` are indexed (see `order.entity.ts`). So `count({ where: { active: true } })`, `count({ where: { state: 'Draft' } })` and the `GROUP BY o.type` in `collectOrdersByType` are full-scan-class queries on what is typically the largest table in a production shop (abandoned carts easily push it into the millions).

And this isn't just a heartbeat cost: `collectOrderMetrics()` runs on the `startup` send too, 5 seconds after bootstrap — exactly when a freshly-deployed server is least able to absorb an unplanned scan of its biggest table.

Options, roughly in order of preference:
- collect order metrics on `heartbeat` sends only, skipping them for `startup`;
- gate collection behind the already-collected total Order count (skip when the bucket is `100k+`);
- add indexes on `active` and `state` via a migration (heavier, affects all installs).

packages/core/src/telemetry/telemetry.service.ts
```diff
@@ -69,31 +90,58 @@ export class TelemetryService implements OnApplicationBootstrap, OnApplicationSh
         // before we check worker mode
         this.delayTimeout = setTimeout(() => {
             this.delayTimeout = undefined;
-            this.sendTelemetry().catch(() => {
+            this.sendTelemetry('startup').catch(() => {
                 // Silently ignore all errors
             });
-        }, 5000);
+            this.scheduleHeartbeat();
+        }, STARTUP_DELAY_MS);
```
michaelbromley, line 104: **Must fix:** `onApplicationBootstrap()` isn't idempotent — this assigns a fresh timeout without clearing any existing one, and `scheduleHeartbeat()` does the same with `this.heartbeatTimeout`. If the hook ever fires twice on the same instance (module double-imported without `@Global()`, hybrid multi-app setups, test harnesses re-triggering lifecycle hooks), you get two startup sends and — worse — two independent self-rescheduling heartbeat chains. Since each chain reassigns `this.heartbeatTimeout` on every reschedule, `onApplicationShutdown()` only holds a reference to whichever timer was scheduled last: it can cancel one chain, and the other keeps sending duplicate heartbeats for the life of the process with no way to stop it.

Cheap fix: a `hasBootstrapped` guard here, or `clearTimeout` any existing ref before reassigning in both places.

Relatedly, the existing spec `'handles double onApplicationBootstrap call without error'` (telemetry.service.spec.ts, ~line 663) is too weak to catch this: `toHaveBeenCalled()` passes whether one send happens or two. Asserting `toHaveBeenCalledTimes(1)` there would fail against the current implementation — and would pin the idempotency fix once it's in.

packages/core/src/telemetry/collectors/config.collector.ts
```diff
@@ -1,9 +1,55 @@
...
+        'orderSellerStrategy',
+        'guestCheckoutStrategy',
+    ],
+    taxOptions: ['taxZoneStrategy', 'taxLineCalculationStrategy', 'orderTaxCalculationStrategy'],
+    entityOptions: ['entityIdStrategy', 'moneyStrategy', 'slugStrategy'],
+    importExportOptions: ['assetImportStrategy'],
+    jobQueueOptions: ['jobQueueStrategy', 'jobBufferStorageStrategy'],
+    schedulerOptions: ['schedulerStrategy'],
+    systemOptions: ['cacheStrategy', 'instrumentationStrategy'],
+};
```
michaelbromley: Nit: this allowlist omits two real single-strategy fields from `default-config.ts`: `authOptions.entityAccessControlStrategy` and `orderOptions.activeOrderStrategy`. Not a correctness bug — just silent under-coverage. `activeOrderStrategy` in particular is a commonly-customized field that seems worth the signal.

packages/core/src/telemetry/telemetry.service.ts
```diff
@@ -69,31 +90,58 @@ export class TelemetryService implements OnApplicationBootstrap, OnApplicationSh
...
         await this.send(payload);
     }
 
     /**
      * Collects all telemetry data from the various collectors.
      */
-    private async collectPayload(): Promise<TelemetryPayload> {
+    private async collectPayload(sendReason: SendReason): Promise<TelemetryPayload> {
         const installationId = await this.installationIdCollector.collect();
         const databaseInfo = await this.databaseCollector.collect();
```
michaelbromley: Nit: these two collects are independent and could be `Promise.all`'d — immaterial at this cadence, just inconsistent with the parallelization pattern this PR uses in the collectors themselves.

michaelbromley (changes requested):
The design is solid and the privacy engineering is careful — I traced the `customizedStrategies` comparison through `mergeConfig`/`simpleDeepClone` and it's genuinely sound (un-overridden strategy instances are kept by reference, and comparing by class name correctly handles plugins re-instantiating a default strategy). No privacy leaks found: `corsWildcardOrigin` only ever reduces to a boolean, `tokenMethod` is a closed union so nothing can leak via stringification, and the superadmin-credentials check never puts the values anywhere. Failure isolation and the guard re-checking claims both hold up.

Two must-fixes though, both cheap:

1. **Order-metrics queries are unindexed full scans on a table that can have millions of rows, and they fire on the startup send too** — see inline comment on `database.collector.ts`.
2. **`onApplicationBootstrap()` isn't idempotent, and a second call creates a heartbeat chain that shutdown can't cancel** — see inline comments on `telemetry.service.ts` and the spec.

Nits (non-blocking):

- `SINGLE_STRATEGY_PATHS` omits `authOptions.entityAccessControlStrategy` and `orderOptions.activeOrderStrategy` — the latter especially is commonly customized and worth adding (inline comment).
- `collectPayload()` awaits `installationIdCollector.collect()` and `databaseCollector.collect()` sequentially even though they're independent — could be `Promise.all`'d for consistency with the parallelization used elsewhere in this PR. Immaterial at a 24h cadence.
- The diff picks up an unrelated line-number drift in `docs/docs/reference/core-plugins/asset-server-plugin/s3asset-storage-strategy.mdx` (`sourceLine="155"` → `"156"`) — harmless doc-generator side effect, but worth a note in the PR description so reviewers aren't puzzled by an S3 doc in a telemetry PR.

Test quality overall is strong — config sections that throw on access, minified class names, rogue OrderType values, jitter timing, unref, shutdown-clears-timer are all pinned. The one gap is the double-bootstrap test (inline comment).

dlhck:
Thanks @michaelbromley — addressed in `857c123fa`.

- Order lifecycle metrics now run on heartbeat only, are skipped for `100k+` Order tables (and when the initial Order count fails), and execute sequentially to minimize peak DB load.
- `onApplicationBootstrap()` is now idempotent, with regression coverage for exactly one startup send, one heartbeat chain, and complete shutdown cleanup.
- Added `entityAccessControlStrategy` and scalar/array-aware `activeOrderStrategy` customization detection.
- Kept installation-ID and database collection deliberately sequential because both can touch the DB; minimizing telemetry load takes precedence over the small latency gain from `Promise.all`.
- The S3 documentation change is generated source-line metadata only and matches the current source; no S3 behavior changed.

Validation: all 390 telemetry tests pass, along with TypeScript and ESLint checks.

Would you mind taking another look?

PR #4934, opened 2026-07-08
feat(core): Add duplicateOptions support to product duplicator

Anujkumar23:
# Description

## Summary

This PR adds a new `duplicateOptions` argument to the default product duplicator.

When enabled, the duplicator creates new `ProductOptionGroup` and `ProductOption` entities for the duplicated product instead of reusing the original option groups and options. Product variants are then recreated using the duplicated options, ensuring the duplicated product is independent of the original.

When the option is disabled, the existing behavior is preserved.

## Related issue

N/A

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

* [x] I have added or updated test cases
* [ ] I have updated the README if needed

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4934"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786103428&installation_id=128408429&pr_number=4934&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4934&signature=c8924b69d0cd06df3cab03f96f0359324e65b3507001caf5a8ed96008f7325d2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/duplicate-entity.e2e-spec.ts
+++ b/packages/core/e2e/duplicate-entity.e2e-spec.ts
@@ -29,6 +29,7 @@ import {
     createAdministratorDocument,
     createChannelDocument,
     createCollectionDocument,
+    createProductDocument,
     createPromotionDocument,
     createRoleDocument,
     duplicateEntityDocument,
@@ -38,6 +39,7 @@ import {
     getFacetWithValuesDocument,
     getProductWithVariantsDocument,
     getPromotionDocument,
+    updateProductOptionGroupDocument,
     updateProductVariantsDocument,
 } from './graphql/shared-definitions';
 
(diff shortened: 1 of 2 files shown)


Anujkumar23:

grolmus (changes requested):
Nice, self-contained addition, @Anujkumar23 — the default (`duplicateOptions: false`) path is byte-for-byte the old shared-groups behaviour, the old→new `optionIdMap` correctly re-points the duplicated variants, and the e2e coverage is genuinely good (asserts translated names, variant re-mapping, original-untouched, twice→independent). Translations are `eager` on both entities so the un-listed relation load is fine.

A few things to sort before this can go in:

- **Target branch:** this is a `feat`, so per [CONTRIBUTING](https://github.com/vendurehq/vendure/blob/master/CONTRIBUTING.md) it should target `minor`, not `master`. Please retarget the PR base to `minor` and rebase.
- **Colliding `-copy` codes:** there's no uniqueness constraint on option / option-group `code`, so duplicating the same product twice produces two groups both coded `foo-copy`, and re-duplicating a duplicate compounds to `foo-copy-copy`. It doesn't error and the duplicate lands disabled for editing, so it's tolerable — but please make it deliberate: either dedupe/suffix the code so repeated duplication stays unambiguous, or add a short comment stating the collision is intentional. Right now the "twice → independent groups" test only compares ids, so it silently green-lights the duplicate codes; asserting on codes there would make the decision explicit.
- **Comment noise:** `// Create a new ProductOptionGroup` and `// Duplicate every ProductOption` just restate the code — please drop them. The `// Share the original…` comment explains the intent of the default path, so keep that one.

Optional: the isolation test only asserts the original is unchanged — asserting that the *duplicate* actually became `'Modified Name'` would also prove the edit landed. And CI's `e2e (mysql)` failure is an unrelated `default-search-plugin` index flake, not your change; a `master` (→ `minor` after retarget) merge + re-run should clear it.

Thanks for the contribution — the core of this is in good shape.

michaelbromley:

michaelbromley:
Thanks for this, and sorry it has sat for so long.

I have gone through the change and the implementation is sound. The `optionIdMap` remap is correct and the e2e coverage is thorough. The reason I am not merging it is not about the code.

Two things need deciding first.

**It reopens a decision made in #4469.** That PR made `ProductOptionGroup` and `ProductOption` shared and channel-aware, and as part of it I removed the option-group duplication from `product-duplicator.ts` on purpose. Duplicated products now share the original's option groups by design. This PR re-adds that removed code behind a flag, which changes the answer from "duplicates share option groups" to "duplicates share option groups unless you tick a box". That is a reasonable position, but it is a different one, and it needs deciding rather than arriving as a side effect. There was no way for you to know this from outside the repository, because the only record of it is the file history.

**It adds to the public API surface.** A new arg on an `EntityDuplicator` is not just TypeScript. It is serialized into the `entityDuplicators` Admin API query, and the dashboard renders every duplicator arg automatically, so `duplicateOptions` becomes a new checkbox for every user and a new argument name that callers can depend on. Once published we have to keep it.

Neither of those is a criticism of your work. Both need a call from me.

Separately, the PR targets `master`, which this project uses for bug fixes. A new capability like this belongs on `minor`.

There is also a real problem underneath this that I do not think anyone has written up: since #4469, editing a duplicated product's option group also edits the original's, because they are the same group. I am going to open an issue for that, because it may change what the right fix here is.

I have labelled this `T3: Systemic` so it is tracked as waiting on a decision rather than waiting on you. I will come back to it.

michaelbromley:
Decision made, and it goes your way: the flag stays.

Both behaviours are legitimate. Sharing option groups between a product and its duplicate is the right default, and wanting independent copies is a reasonable thing to want. So `duplicateOptions` is the right shape for this, and the project will own it as Admin API surface.

Two changes follow from that.

**I have retargeted this PR from `master` to `minor`.** This adds a new capability rather than fixing a bug, and `minor` is where features go in this repository.

**I have relabelled it `T2: Elevated risk` rather than fast-tracking it.** That is not a comment on the code. It is because of one specific thing that needs checking before it merges.

The duplication logic this PR restores was written before #4469. #4469 is the commit that made `ProductOptionGroup` and `ProductOption` channel-aware in the first place, and it removed this code in the same change. So the restored code predates channel awareness, and the question nobody has answered yet is what channels the newly created option groups and options are assigned to when `duplicateOptions` is true. If they land in the wrong channel, the failure is quiet, and channel scoping is the boundary between tenants. That is worth being careful about.

What this needs before merge:

- Confirmation of how the new option groups and options are assigned to channels, and a test covering duplication with `duplicateOptions: true` in a multi-channel setup.
- The existing e2e coverage extended to the case where the source product's option groups are shared with other products.
- A rebase onto `minor`, since this branch is now some way behind.

One more thing worth knowing, which I found while looking into this. There is already a component at `packages/dashboard/src/app/routes/_authenticated/_products/components/shared-option-group-warning.tsx`, added in #4483, which renders "This option group is shared across # products. Changes will affect all of them." It is imported by nothing. Somebody hit this same confusion, built the warning for it, and never wired it up. Whatever happens to this PR, that warning should probably be connected to the product detail page, because the sharing is surprising even when it is what you asked for.

Thanks for your patience on this one. Sixty-four days is too long for a PR that turned out to be answering a real question.

PR #4939, opened 2026-07-09
fix(core): retain v3.6 migration helpers to avoid breaking committed migrations

michaelbromley:
## Summary

Removes the `// TODO: Remove this file once v3.6 has been stable for a while` comment from the two v3.6 migration helpers and documents why they must instead be kept published:

- `packages/core/src/migration-utils/v3_6_shared_option_groups.ts` (`migrateProductOptionGroupData`)
- `packages/core/src/migration-utils/v3_6_asset_translations.ts` (`migrateAssetTranslationData`)

## Why

These helpers are designed to be **imported directly into user migration files** (that's what the docs examples show):

```ts
import { migrateProductOptionGroupData } from '@vendure/core';
```

Migration files are compiled as part of a project's build, and they are immutable historical records. So if a future core release removes the export (as the TODO plans), then **every project that followed the documented v3.6 upgrade path** will have a committed migration whose import no longer resolves — a build-breaking `TS2305` on their next compile, hitting precisely the users who did the right thing.

The two options are mutually exclusive:

- users import the helper → it can never be safely removed, or
- we want to remove it → users must not import it (inline the SQL instead).

Since the helpers already guard against the already-migrated state (`hasColumn(...)`), keeping them exported indefinitely is a cheap no-op and the safe choice. This PR just makes that explicit and drops the plan-to-delete.

No functional change.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4939"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786211049&installation_id=128408429&pr_number=4939&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4939&signature=9be1ac1c0a337574da403427ef4fca3407f5c131f19e60bd6474f9eafef913ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/docs/reference/core-plugins/asset-server-plugin/s3asset-storage-strategy.mdx
+++ b/docs/docs/reference/core-plugins/asset-server-plugin/s3asset-storage-strategy.mdx
@@ -4,7 +4,7 @@ generated: true
 ---
 ## S3AssetStorageStrategy
 
-<GenerationInfo sourceFile="packages/asset-server-plugin/src/config/s3-asset-storage-strategy.ts" sourceLine="155" packageName="@vendure/asset-server-plugin" />
+<GenerationInfo sourceFile="packages/asset-server-plugin/src/config/s3-asset-storage-strategy.ts" sourceLine="156" packageName="@vendure/asset-server-plugin" />
 
 An [AssetStorageStrategy](/reference/typescript-api/assets/asset-storage-strategy#assetstoragestrategy) which uses [Amazon S3](https://aws.amazon.com/s3/) object storage service.
 To us this strategy you must first have access to an AWS account.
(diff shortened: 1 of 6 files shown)


pujux (approved):
🚀

PR #4945, opened 2026-07-10
fix(core): Merge shared entity instances into every hydration target

### LINKED ISSUE #4935: EntityHydrator / mergeDeep silently skips relations when an order has duplicate entity instances
**Describe the bug**
`EntityHydrator.hydrate()` silently fails to populate requested relations on some entities when the target object graph contains duplicate in-memory instances of the same entity (same id, distinct objects).

The cause is the circular-reference guard in `mergeDeep`, it records every source object it has processed in a visited WeakSet and skips any source object it encounters again.

With this guard, that shared instance is merged into the first referencing order line only. When the merge reaches the second line, the variant is already in visited and is skipped entirely 

**To Reproduce**
1. Create an active order that contains two order lines for the same ProductVariant,
2. Call `EntityHydrator.hydrate(ctx, order, { relations: ['lines.productVariant.facetValues.facet'] })`.
3. Inspect order.lines: One line's variant has facetValues; the duplicate line's variant has undefined.

**Expected behavior**
A source object shared by multiple targets is merged into **every** distinct target

**Actual behavior**
The requested relation is populated on the first referencing line only

**Environment:**

**Minimal reproduction**
```js
const { mergeDeep } = require('@vendure/core/dist/service/helpers/entity-hydrator/merge-deep');

const sharedVariant = {
    id: 9572,
    facetValues: [{ id: 1, code: '5kg', facet: { id: 7, code: 'weight' } }],
};

const hydratedLines = [
    { id: 1, productVariant: sharedVariant },
    { id: 2, productVariant: sharedVariant },
];

const targetLines = [
    { id: 1, productVariant: { id: 9572 } },
    { id: 2, productVariant: { id: 9572 } },
];

mergeDeep(targetLines, hydratedLines);

console.log('line 1 facetValues:', JSON.stringify(targetLines[0].productVariant.facetValues));
console.log('line 2 facetValues:', JSON.stringify(targetLines[1].productVariant.facetValues));
```

**Workaround**
We locally patch mergeDeep so the visited set tracks only the current recursion path (add on enter, delete on exit) instead of the whole merge shared instances are then merged into every referencing target, while real circular references are still caught. But to keep performance linear on heavily shared graphs, cleanly-completed (target, source) pairs could be memoized, so re-encountering the same pair skips the redundant re-merge

ryandiginomad:
# Description

Fixes #4935.

`mergeDeep` (used by `EntityHydrator` to merge freshly-loaded relations onto an already-loaded entity) guards against circular references by recording every source object it processes in a persistent `visited` WeakSet and skipping any it encounters again:

```ts
if (isObject(b)) {
    if (visited.has(b)) {
        return a;
    }
    visited.add(b);   // never removed
}
```

The problem is that a `visited` set spanning the **whole** merge can't tell a *circular* reference apart from a *shared* one. When a single source object is referenced by more than one target, it's merged into the first target and then skipped for the rest.

This happens in real orders: two `OrderLine`s can reference the same `ProductVariant` when they carry different `OrderLine` custom fields (e.g. per-line personalisation). Once such an order is loaded with the `'query'` relation strategy, TypeORM assigns the *same* `ProductVariant` instance to both lines. Hydrating a relation (`lines.productVariant…`) then populates it on the first line only; the duplicate line's variant is silently skipped — exactly the symptom in #4935.

## The fix

Track only the **current recursion path** instead of the whole merge: add the source object on the way in, remove it on the way out.

```ts
let addedToPath = false;
if (isObject(b)) {
    if (visited.has(b)) {
        return a;
    }
    visited.add(b);
    addedToPath = true;
}
// …merge…
if (addedToPath) {
    visited.delete(b);
}
```

A genuine self-referential cycle is still caught — it is still on the path when re-encountered — while a shared, non-circular instance is no longer mistaken for a cycle and is merged into every referencing target.

This is the same fix direction the reporter suggested. I kept it to the minimal path-scoped change; the optional `(target, source)` pair-memoisation they mentioned for perf on heavily-shared graphs isn't included, since correctness doesn't need it and it adds state — happy to add it if you'd prefer.

# Breaking changes

None. Single-target merges and genuine circular references behave exactly as before; the only behavioural change is that a shared instance now correctly populates every target instead of just the first.

# Checklist

- [x] I have added test cases:
  - `merge-deep.spec.ts`: a shared source instance is merged into every referencing target (the existing order-sync and circular-reference tests still pass, confirming cycles are still caught).
  - `entity-hydrator.e2e-spec.ts`: two order lines sharing one variant (via different OrderLine custom fields) both get the hydrated relation populated.

### 🤖 AI assistance

Investigated and implemented with the help of Claude Code. I reproduced the bug locally against the reporter's minimal snippet, traced it to the persistent `visited` set, and verified the fix end-to-end (unit spec red→green, the full `entity-hydrator` e2e suite green on sqljs, including confirming genuine circular references are still caught). I reviewed every change myself.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4945"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786272701&installation_id=128408429&pr_number=4945&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4945&signature=f0263d6bf7633798ddeb12b9b6a60461f1c8d501c6ee96594905c5d5d07210d7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


michaelbromley (commented):
Thanks for the thorough investigation on this one.

I verified this locally: reverted the `merge-deep.ts` change with your tests in place and the new e2e test fails exactly as described (`lines[1].productVariant.product` undefined), which confirms TypeORM's `'query'` strategy really does assign the same variant instance to both lines. Re-applied the fix, rebuilt, all green.

One thing I'd like addressed before merging:

**The e2e assertions are looser than they need to be.** `toBeGreaterThan(0)` catches the "nothing hydrated" case but would pass if line 2 got the wrong product's facetValues. T_1 has exactly 2 facetValues (asserted earlier in the same file), so assert `toBe(2)` on both lines, and ideally assert parity between them (e.g. compare `facetValues.map(fv => fv.id)`).

ryandiginomad:
Done — tightened both assertions. Each line now asserts exactly 2 facetValues (T_1's count) and that the two lines' `facetValues.map(fv => fv.id)` sets are identical, so a partial or wrong-product merge would fail rather than slip through. Full `entity-hydrator` e2e suite is green.

Thanks for taking the time to verify it locally against the reverted fix — good to have independent confirmation that the `'query'` strategy really does share the instance.

michaelbromley (approved):

PR #4952, opened 2026-07-10
feat(dashboard): Allow creating customer and address inline on draft …

### LINKED ISSUE #4951: Draft order: allow creating a customer and address inline in the React dashboard (parity with Angular admin-ui)
**Describe the bug**
In the new React dashboard, the draft order creation flow only lets you SELECT an existing customer and SELECT one of the customer's existing addresses. There is no way to create a new customer inline, nor to enter a new/ad-hoc shipping or billing address. The old Angular admin-ui supported both of these inline during draft order creation, so this is a feature-parity regression.

**To Reproduce**

1. Go to the React dashboard → Sales → Orders
2. Click "Draft order" to create a new draft
3. In the "Customer" block, click "Select customer" — you can only search/pick an existing customer; there is no "Create new customer" option
4. In the "Shipping address" / "Billing address" blocks, click "Select address" — you can only pick from the customer's saved addresses; there is no way to type a new address

**Expected behavior**
- The customer selector should let the admin either select an existing customer OR create a new one inline (title, first name, last name, email, phone) — matching the Angular admin-ui .
- The address selector should let the admin either pick an existing address OR enter a new ad-hoc address inline — matching the Angular admin-ui .
- The same inline-address capability should also be available in the order modification flow.

**Actual behavior**
- Customer selector: select-existing only (the `$input: CreateCustomerInput` path of `setCustomerForDraftOrder` is wired in GraphQL but never used by the UI).
- Address selector: select-from-saved-addresses only; disabled when no customer is set; no free-form entry, even though the backend accepts `CreateAddressInput`.

**Screenshots/Videos**
Angular admin-ui: 

Image

React dashboard (current): 

Image

- Nodejs version: 22.x
- Browser (if applicable): Chromium
- Package manager (npm/yarn/pnpm): bun

**Configuration**
N/A — reproducible on a default Vendure install with the React dashboard.

**Minimal reproduction**
Any default Vendure project running the React dashboard; create a draft order and observe the customer/address blocks.

**Workaround**
Create the customer and address separately in the Customers section first, then return to the draft order and select them. This is extra steps compared to the Angular admin-ui inline flow.

**Additional context**
- The backend already fully supports this: `setCustomerForDraftOrder(orderId, customerId, input: CreateCustomerInput)` and `setDraftOrderShippingAddress` / `setDraftOrderBillingAddress` (accepting `CreateAddressInput`). Only the React UI needs to use these paths.

Ryrahul:
Brings the React dashboard draft order flow to parity with the Angular admin-ui by allowing an admin to create a new customer inline and enter a new/ad-hoc shipping or billing address inline, rather than only selecting existing records.

- CustomerSelector gains an opt-in `allowCreateNew` mode rendering a tabbed popover (select existing / create new) that emits a CreateCustomerInput via `onCreateNew`.
- CustomerAddressSelector gains a "New address" tab embedding the shared CustomerAddressForm, emitting a CreateAddressInput via `onSubmitNew`.
- CustomerAddressForm gains a `hideDefaultAddressFlags` prop (the default shipping/billing flags are not applicable in the draft order context).
- Wired into both the draft order page and the order modification page.

No backend changes required: setCustomerForDraftOrder already accepts a CreateCustomerInput, and the draft order address mutations accept a CreateAddressInput.

Fixes #4951 

# Description

The React dashboard draft order flow previously only allowed selecting an existing customer and selecting one of the customer's saved addresses. The old Angular admin-ui also supported creating a new customer inline and entering a new/ad-hoc address inline. This PR restores that parity.

The backend already fully supports this — `setCustomerForDraftOrder` accepts an optional `CreateCustomerInput`, and `setDraftOrderShippingAddress` / `setDraftOrderBillingAddress` accept a `CreateAddressInput`. The React UI simply never used these paths. This PR wires them up:

- **Customer** — `CustomerSelector` gets an opt-in `allowCreateNew` tabbed popover ("Existing customer" / "Create new customer"). The create form has the same fields as Angular (title, first name, last name, email, phone). The existing select-only behaviour is preserved when `allowCreateNew` is not set, so the customer-groups call site is untouched.
- **Address** — `CustomerAddressSelector` gets an "Existing address" / "New address" tabbed popover. The new-address tab reuses the shared `CustomerAddressForm` (default-shipping/billing checkboxes hidden via a new `hideDefaultAddressFlags` prop, matching Angular's draft flow). Works with or without a customer set.
- Both the draft order page and the order modification page are wired to the inline-create paths.

Added e2e coverage in `sales/orders.spec.ts` for creating a customer inline and entering a new shipping address inline. All 20 order e2e tests pass.

# Breaking changes

None. The change is additive and backward-compatible — `CustomerSelector` keeps its original select-only behaviour by default (inline creation is opt-in via `allowCreateNew`), and no backend/GraphQL changes are required.

# Screenshots

image
image

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4952"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786310156&installation_id=128408429&pr_number=4952&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4952&signature=f938e31db58818589e88d7e4691a707f526ebaa11aacb1af33e9bbc52dff7102"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/dashboard/src/app/routes/_authenticated/_orders/orders_.$id_.modify.tsx
```diff
@@ -109,6 +111,16 @@ function ModifyOrderPage() {
...
 
+    function handleSubmitNewShippingAddress(input: CreateAddressInput) {
+        updateShippingAddressRawInInput(input);
+        setEditingShippingAddress(false);
+    }
+
+    function handleSubmitNewBillingAddress(input: CreateAddressInput) {
+        updateBillingAddressRawInInput(input);
+        setEditingBillingAddress(false);
+    }
```
biggamesmallworld: `UpdateOrderAddressInput` has no `customFields`, so passing it through fails the `modifyOrder` dry run. Strip it here rather than in the mapper, so the draft path keeps custom fields (where they do work).

```suggestion
    function handleSubmitNewShippingAddress({ customFields, ...input }: CreateAddressInput) {
        // `customFields` is not part of `UpdateOrderAddressInput`, so modifyOrder rejects it
        updateShippingAddressRawInInput(input);
        setEditingShippingAddress(false);
    }

    function handleSubmitNewBillingAddress({ customFields, ...input }: CreateAddressInput) {
        updateBillingAddressRawInInput(input);
        setEditingBillingAddress(false);
    }
```

packages/dashboard/src/lib/components/shared/customer-selector.tsx
```diff
@@ -66,44 +101,184 @@ export function CustomerSelector(props: CustomerSelectorProps) {
...
-                        className="h-10 flex-1 bg-transparent outline-none placeholder:text-muted-foreground"
+        <Form {...form}>
+            <form
+                onSubmit={e => {
+                    e.stopPropagation();
+                    form.handleSubmit(onSubmit)(e);
+                }}
+                className="space-y-4"
+            >
+                <div className="grid grid-cols-1 md:grid-cols-2 gap-4">
```
biggamesmallworld: Tailwind's `md:` keys off the viewport, not the container, so on any desktop this is 2 columns inside a 420px popover — ~190px per field.

```suggestion
                <div className="grid grid-cols-1 gap-4">
```

packages/dashboard/src/lib/components/shared/customer-selector.tsx
```diff
@@ -6,14 +6,21 @@ import {
...
 import { api } from '@/vdb/graphql/api.js';
 import { graphql } from '@/vdb/graphql/graphql.js';
-import { Trans } from '@lingui/react/macro';
+import { z, zodResolver } from '@/vdb/lib/zod.js';
+import { Trans, useLingui } from '@lingui/react/macro';
 import { useQuery } from '@tanstack/react-query';
 import { useDebounce } from '@uidotdev/usehooks';
-import { Plus } from 'lucide-react';
+import { VariablesOf } from 'gql.tada';
+import { Link, Plus } from 'lucide-react';
```
biggamesmallworld: `Link` is the chain icon — reads as a hyperlink.

```suggestion
import { Plus, Users } from 'lucide-react';
```

packages/dashboard/src/lib/components/shared/customer-selector.tsx
```diff
@@ -66,44 +101,184 @@ export function CustomerSelector(props: CustomerSelectorProps) {
...
+                }
+            }}
+        >
+            {trigger}
+            <PopoverContent className="w-[420px] p-0" align="start">
+                <div className="p-4">
+                    <Tabs value={activeTab} onValueChange={setActiveTab}>
+                        <TabsList className="grid w-full grid-cols-2">
+                            <TabsTrigger value="existing">
+                                <Link className="mr-2 h-4 w-4" />
```
biggamesmallworld:
```suggestion
                                <Users className="mr-2 h-4 w-4" />
```

packages/dashboard/src/app/routes/_authenticated/_orders/components/customer-address-selector.tsx
```diff
@@ -1,32 +1,45 @@
...
+import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/vdb/components/ui/tabs.js';
 import { api } from '@/vdb/graphql/api.js';
 import { ResultOf } from '@/vdb/graphql/graphql.js';
-import { Trans, useLingui } from '@lingui/react/macro';
 import { cn } from '@/vdb/lib/utils.js';
+import { Trans } from '@lingui/react/macro';
 import { useQuery } from '@tanstack/react-query';
-import { Plus } from 'lucide-react';
+import { VariablesOf } from 'gql.tada';
+import { Link, Plus } from 'lucide-react';
```
biggamesmallworld: Same here — chain icon on an address tab.

```suggestion
import { MapPin, Plus } from 'lucide-react';
```

packages/dashboard/src/app/routes/_authenticated/_orders/components/customer-address-selector.tsx
```diff
@@ -35,66 +48,113 @@ export function CustomerAddressSelector({
...
-                            <div className="text-sm text-muted-foreground">
-                                <Trans>Loading addresses...</Trans>
-                            </div>
-                        ) : addresses.length === 0 ? (
-                            <div className="text-sm text-muted-foreground">
-                                <Trans>No addresses found</Trans>
+                    <Tabs value={effectiveTab} onValueChange={setActiveTab}>
+                        <TabsList className="grid w-full grid-cols-2">
+                            <TabsTrigger value="existing" disabled={!canSelectExisting}>
+                                <Link className="mr-2 h-4 w-4" />
```
biggamesmallworld:
```suggestion
                                <MapPin className="mr-2 h-4 w-4" />
```

packages/dashboard/e2e/tests/sales/orders.spec.ts
```diff
@@ -294,6 +294,86 @@ test.describe('Orders', () => {
         await expect(page.getByRole('button', { name: /Select address/i })).toHaveCount(2);
     });
 
+    // #4838 — parity with the Angular admin-ui: a draft order should allow creating a
+    // new customer inline (not just selecting an existing one).
```
biggamesmallworld: PR body says `Fixes #4951`.

```suggestion
    // #4951 — parity with the Angular admin-ui: a draft order should allow creating a
    // new customer inline (not just selecting an existing one).
```

packages/dashboard/e2e/tests/sales/orders.spec.ts
```diff
@@ -294,6 +294,86 @@ test.describe('Orders', () => {
...
+
+        // The mutation runs and the customer becomes set on the order
+        await page.waitForResponse(resp => resp.url().includes('/admin-api') && resp.status() === 200);
+        await expect(page.getByRole('button', { name: /Inline Customer/i })).toBeVisible({
+            timeout: 10_000,
+        });
+    });
+
+    // #4838 — parity with the Angular admin-ui: a draft order should allow entering a
+    // new, ad-hoc address inline (not just selecting from the customer's saved addresses).
```
biggamesmallworld:
```suggestion
    // #4951 — parity with the Angular admin-ui: a draft order should allow entering a
    // new, ad-hoc address inline (not just selecting from the customer's saved addresses).
```

packages/dashboard/e2e/tests/sales/orders.spec.ts
```diff
@@ -294,6 +294,86 @@ test.describe('Orders', () => {
         await expect(page.getByRole('button', { name: /Select address/i })).toHaveCount(2);
     });
 
+    // #4838 — parity with the Angular admin-ui: a draft order should allow creating a
+    // new customer inline (not just selecting an existing one).
+    test('should create a new customer inline on a draft order', async ({ page }) => {
+        test.setTimeout(60_000);
+
```
biggamesmallworld, line 301: This test skips `client.login()`, the next one does it. Pick one.

```suggestion
        test.setTimeout(60_000);

        const client = new VendureAdminClient(page);
        await client.login();

```

biggamesmallworld (changes requested):
Reviewed against the Angular `select-customer-dialog` / `select-address-dialog` / `order-editor`.

One blocker, plus parity gaps that need a follow-up decision.

**Blocker:** `mapFormValuesToInput()` always emits `customFields`. Fine for the draft path (`CreateAddressInput` has it), broken on the modify path — `UpdateOrderAddressInput` (order.api.graphql:173-183) has no `customFields` field, so `modifyOrder` fails with a variable coercion error. TS doesn't catch it because the returned object isn't fresh, so there's no excess-property check. It fails late: fill form → Save Address → address shows in the summary → Preview → GraphQL error. Suggestion inline.

**Parity gaps (no suggestion — need design input):**

1. No `currentAddress` prefill. Angular seeds the dialog from the current address (`select-address-dialog.component.ts:41-49`), passed from both the draft page and the order editor. Here the new-address form always starts blank — on the modify page the button says "Edit" and opens 10 empty fields, so fixing a postcode means retyping the address. Angular's order-editor didn't use a dialog at all: inline forms patched from `order.shippingAddress`, submitted only if dirty.
2. Draft order: an existing address can't be edited. Angular always renders "Set shipping address" (`draft-order-detail.component.html:85`); here only Remove shows once `shippingAddress.streetLine1` is set, so it's remove-then-retype. Natural place to close this.
3. Angular preselects the saved address matching the current one (`streetLine1` + `postalCode`); nothing is highlighted here.

**UX / structure:**

4. Multi-field forms inside a dismiss-on-outside-click popover. Angular used modal dialogs. Half-filled address + stray click = gone, and `onCancel` isn't passed through to `CustomerAddressForm` so there's no Cancel button either. Suggest `Dialog` for both create-customer and new-address, keeping the popover for plain select. Cheaper fix: prevent `onInteractOutside` and wire up `onCancel`.
5. `setOpen(false)` fires synchronously on submit, before the mutation resolves. `createOrUpdate(ctx, input, true)` returns `EmailAddressConflictError` on a duplicate email → toast fires, popover is already gone, all five fields retyped. Angular has the same flaw so it's parity, not a regression, but worth fixing here.
6. Customer custom fields aren't in the inline create form (Address ones are). `createOrUpdate` hard-fails if any required Customer custom field is configured, with no way to supply it.
7. `createCustomerInputDocument` is a dummy mutation used only as a type anchor, and it lands in the operation namespace. `createCustomerDocument` already exists at `_customers/customers.graphql.ts:106`. If this was to avoid `lib/components/shared/*` importing from `app/routes/*`, move the document into `lib/` instead.
8. Same for `CreateAddressInput` — exported from a route-local component and imported by the modify page. Belongs in `lib/graphql` with the other shared input types.
9. "Save Address" is misleading in the draft context (nothing is persisted to the address book; Angular said "Okay"), but the label lives in the shared form where it's correct for the address book. Needs a `submitLabel` prop rather than a rename.

**Minor:** `postalCode` is optional here, required in Angular. Looks deliberate — worth a line in the PR body.

The `waitForResponse` predicate and `Date.now()` emails match what's already on master, so not blocking, but there's no e2e over the order-modification address path — which is exactly where the blocker lives.

Direction is right and the draft-order happy path is wired correctly. Holding on the blocker plus gaps 1 and 2.

biggamesmallworld:
One more thing on the `.po` churn.

`bun run i18n:extract` was committed, and 33 locale files moved. Almost none of it is new strings — in `en.po`, 156 of the 176 changed lines are `#: path:NNN` source-reference renumbering caused by line shifts in the touched components. Eight real new msgids across the set (`Existing customer`, `Create new customer`, `Search customers...`, `Create customer`, `Set customer`, `Existing address`, `New address`, `Select an existing customer or create a new one`).

Two consequences:

1. The diff blew past GitHub's 20k-line cap, so `gh pr diff` and the Files tab both refuse to render it in full. Anyone reviewing this has to pull the branch.
2. The PR is currently `CONFLICTING` — 27 conflicting files, every one of them a locale `.po`. No source file conflicts, so it's purely extract-vs-extract churn against what has landed on master since.

Suggest rebasing on master and re-running `i18n:extract` as the last commit, so the reference renumbering is regenerated rather than merged. Worth keeping it as its own commit too — it makes the real diff reviewable and the conflict trivially resolvable with `--theirs` + re-extract next time this happens.

Also: `en.po` references `customer-selector.tsx:266` for `Select an existing customer or create a new one`, but that string isn't in the component at head — looks like the extract was run against an earlier revision, so it'll be stale even after the rebase.

Ryrahul:
Thanks for the detailed review @biggamesmallworld  ! . Will work on the blocker and push a fix for it

Ryrahul:
Remaining  @biggamesmallworld  need your input on these

1. **Edit existing address on draft:** Should **"Select address"** remain visible alongside **"Remove"** so users can overwrite the existing address, or should **"Remove"** be replaced with **"Edit"**, which reopens the form pre-filled?

2. **Address highlighting:** Should we replicate Angular's behavior by matching **`streetLine1` + `postalCode`** to automatically preselect the currently saved address?

3. **Dismiss on outside click:** Should we switch the form tabs to use **`Dialog`**, or simply prevent **`onInteractOutside`** and wire up **`onCancel`**?

4. **`setOpen` before mutation:** Should this be fixed as part of this PR, or handled in a follow-up? (As noted, this matches the current Angular behavior.)

5. **Customer custom fields:** Should we add **`CustomFieldsForm`** to the inline customer creation form, or leave it for a follow-up?

biggamesmallworld:
Answers below, then where things stand.

**1. Edit existing address on draft** — Keep "Select address" visible alongside "Remove", and pass `initialAddress` so the New address tab opens prefilled. Don't replace Remove with Edit: unsetting an address is still a distinct action (`unsetShippingAddressForDraftOrder`), and Angular renders "Set shipping address" unconditionally at `draft-order-detail.component.html:85`. So: both buttons, prefilled form — same wiring you already added on the modify page.

**2. Address highlighting** — Yes, `streetLine1` + `postalCode`, matching Angular. Cheap, and it also resolves the `cn()` call at `customer-address-selector.tsx:115-117`, which currently wraps a single static string and does nothing — clearly the seam left for this. Add the selected styling plus `aria-selected` so the e2e can assert on it.

**3. Dismiss on outside click** — Not `Dialog`, not in this PR. Prevent `onInteractOutside` (and `onEscapeKeyDown`) *only while the create/new tab is active*, so plain select keeps click-away dismissal, and pass `onCancel` through to `CustomerAddressForm` so the Cancel button appears. `Dialog` is the better long-term shape but it's a bigger structural change than this PR needs.

**4. `setOpen` before mutation** — In this PR too. It touches the same files you're already changing, and it's small: let `onCreateNew` / `onSubmitNew` return `void | Promise<unknown>`, `await` the result in the selector, and only `setOpen(false)` once it resolves. On the draft page that means `mutateAsync` instead of `mutate` at `orders_.draft.$id.tsx:166` (and the two address mutations), with the handler returning the promise.

One detail that makes the naive version not work: `EmailAddressConflictError` comes back as a member of the result union, not as a rejection — so awaiting alone still closes the popover on conflict. The handler needs to reject on the error typename, e.g. `const result = await setCustomerForDraftOrder({...}); if (result.setCustomerForDraftOrder.__typename !== 'Order') throw new Error(...)`. The existing `onSuccess` already toasts for those cases, so the throw only has to signal "keep the form open".

**5. Customer custom fields** — In this PR, please. This is the one I wouldn't defer: `createOrUpdate` hard-fails when a required Customer custom field is configured, and inline create has no way to supply it — so the feature is unusable on those installs. The address form already does `<CustomFieldsForm entityType="Address" ... />`; the symmetric addition to `CreateCustomerForm` plus `customFields` in the zod schema is a few lines.

---

**Blocker (new): the build is red.** `initialAddress` is typed `Order['shippingAddress']`, but on the modify page it receives `modifyOrderInput.updateShippingAddress ?? entity.shippingAddress` — the first branch is `UpdateOrderAddressInput`, whose fields are all optional, so it isn't assignable. Fails at `orders_.$id_.modify.tsx:269` and `:296`. Introduced by `3a5c714`. Widen the prop to the shape both sides satisfy:

```ts
type InitialAddressValues = Partial<
    Record<
        | 'fullName' | 'company' | 'streetLine1' | 'streetLine2' | 'city'
        | 'province' | 'postalCode' | 'countryCode' | 'phoneNumber',
        string | null
    >
>;
```

and take `InitialAddressValues | null | undefined` in `mapOrderAddressToFormValues`.

**Still missing:** no e2e over the modify-page inline address path. The `customFields` fix lives at the call site (`handleSubmitNew*Address`), so nothing prevents a future caller from reintroducing it — and that path fails late, only at Preview. One test: modify page → Edit shipping address → New address → Update address → Preview, assert no GraphQL error.

**Resolved:** the `customFields` blocker, the prefill on modify, `submitLabel`, and both input types moved to `lib/graphql`. All nine inline suggestions applied.

On the i18n concern — that's clean now. `dashboard i18n sync` is green and the diff is pure `#:` reference renumbering: zero `msgstr` values were emptied in any of the 33 catalogs, so the "restore the deleted translations" comments above are false positives.

Last minor from my earlier review, still open: `postalCode` is optional here, required in Angular. Looks deliberate — worth a line in the PR body.

Ryrahul:
Thanks for the quick reply @biggamesmallworld ! I was still working through some of the remaining issues and honestly wasn't expecting such a quick response so those are remaining from previous review . I'll address the remaining CI issues and update the PR once everything is sorted.

biggamesmallworld:
Easy @Ryrahul I will hold off until tomorrow 👍

biggamesmallworld:
Q1, Q2 and Q5 look right, and the modify-page e2e is the coverage I wanted. Four things left.

**1. Build is red — the outside-click fix doesn't compile.** `onInteractOutside` / `onEscapeKeyDown` are Radix props; our `PopoverContent` is Base UI (`@base-ui/react/popover` via `@vendure-io/ui`), which has neither. `check-types` gives three errors at `customer-address-selector.tsx:116` and `:122`. Base UI cancels dismissal from the root instead — this version typechecks clean:

```tsx
<Popover
    open={open}
    onOpenChange={(value, eventDetails) => {
        // While the form tab is open, only our own submit/cancel may close the popover
        if (!value && effectiveTab === 'new') {
            return eventDetails.cancel();
        }
        setOpen(value);
        if (!value) {
            setActiveTab('existing');
            onCancel?.();
        }
    }}
>
```

and drop both handlers from `PopoverContent`.

No reason check needed: `onOpenChange` is only called from `PopoverStore.setOpen`, i.e. from Base UI's own interactions — submit and Cancel set `open` directly through controlled state and never reach it. That also covers `focus-out`, which a check against just `outside-press` / `escape-key` would miss.

**2. Same fix is missing on `CustomerSelector`.** The "Create new customer" tab still dismisses on outside click and has no Cancel button, so five filled fields are one stray click from gone.

**3. Q4 isn't in.** No `mutateAsync` in `orders_.draft.$id.tsx`; both `onCreateNew` and `onSubmitNew` still `setOpen(false)` synchronously.

**4. Required `postalCode` over-applies.** Angular required it only in the order dialog (`select-address-dialog.component.ts:46`). The customer address book form doesn't — `customer-detail.component.ts:176-183` marks only `streetLine1` and `countryCode` required. Making it required in the shared `CustomerAddressForm` means every address in the address book now needs a postcode, including for countries that don't use them. Please revert to `.optional()`, or gate it behind a prop that only the order selector sets.

Ryrahul:
Thanks for the thorough reviews throughout this PR @biggamesmallworld — really appreciate the detailed feedback and helping me out .

All four items from the latest review are addressed:

1. **Build fix** — replaced Radix `onInteractOutside`/`onEscapeKeyDown` with Base UI `eventDetails.cancel()` on the `Popover` root, no reason check (covers `focus-out` too)
2. **CustomerSelector** — same dismiss prevention applied to the "Create new customer" tab, `onCancel` wired through with a Cancel button
3. **Q4 (mutateAsync)** — you'd mentioned opening a separate issue for this earlier so I'd deferred it, but pushed it now. All three draft-order mutations use `mutateAsync`, callbacks are async, selectors await before closing
4. **postalCode** — reverted to `.optional()` in the shared form

biggamesmallworld:
Tested the branch locally against a real order (draft → complete → payment → Modify), plus the draft order flow.

The `customFields` blocker is properly fixed — Preview opens clean on the modify path. Prefill, `submitLabel`, the inline Select address/Remove row, and the inline address create all work.

Two things left.

**1. Duplicate email still loses the form.** `setCustomerForDraftOrder` returns a union:

```graphql
# order.api.graphql:463
union SetCustomerForDraftOrderResult = Order | EmailAddressConflictError
```

An `EmailAddressConflictError` is a *successful* response, so `mutateAsync` resolves, `onError` never runs, and the selector closes the popover — the toast fires and all five fields are gone. The `mutateAsync` change fixed the timing, not this case. Needs the callback to signal failure:

```ts
// orders_.draft.$id.tsx:509
onCreateNew={async input => {
    const result = await setCustomerForDraftOrder({ orderId: entity.id, input });
    if (result.setCustomerForDraftOrder.__typename !== 'Order') {
        // ErrorResult, not a transport error — mutateAsync resolves, so throw to
        // keep the popover open with the entered values.
        throw new Error(result.setCustomerForDraftOrder.message);
    }
}}
```

and swallow the rejection so it isn't unhandled:

```ts
// customer-selector.tsx:141
form.handleSubmit(onSubmit)(e).catch(() => undefined);
```

Verified both directions with that applied — conflict keeps the form, unique email still creates and closes.

**2. The i18n catalogs blanked 15 existing translations** in all 32 locales — unrelated stock-location and customer-group strings from the rebase. Per locale: 1250 → 1234 populated, 40 → 63 empty. Reset the `.po` files to master and re-run `bun run i18n:extract`.

Not blocking, separate issue: the address form never prefills Country (`Select a country` even when the order is GB). It's pre-existing — `customer-address-form.tsx` passes both `defaultValue` and `value` to the Select, and it reproduces on the untouched customer address-book dialog. Worth an issue since it does undercut the new prefill.

Happy to approve once those two are sorted.

biggamesmallworld (approved):
Both blockers from the last round are fixed. Verified on `a422cdc0`:

**1. `EmailAddressConflictError`** — the handler now throws on `__typename !== 'Order'`, and `customer-selector.tsx` swallows the rejection so it isn't unhandled. Conflict keeps the form, unique email creates and closes.

**2. i18n catalogs** — clean. The only removed `msgstr` is `"Select an address"`, correctly moved to obsolete (`#~`) now that the string is gone from the UI; nothing else was blanked in any of the 32 locales. `dashboard i18n sync` is green.

CI is green apart from `deploy`, which fails on an empty `VERCEL_TOKEN` — fork PRs don't get the secret. Not yours.

One thing before merge: please revert `packages/dev-server/graphql/graphql-env.d.ts` to master.

```
git checkout origin/master -- packages/dev-server/graphql/graphql-env.d.ts
```

It got regenerated with `FieldTestPlugin` enabled locally, so the snapshot has `RRP`, `productSpecs` and `seoTitle` — but that plugin is commented out in `dev-config.ts:132` on master, so the committed schema wouldn't match the committed config. Harmless at runtime, but it makes the next regeneration produce a phantom revert diff.

Approving now — push that one revert and I'll merge.

Separately, I'll open an issue for the Country field not prefilling in `CustomerAddressForm`; it's pre-existing and reproduces on the untouched address-book dialog, so it's out of scope here.

Ryrahul:
> Both blockers from the last round are fixed. Verified on `a422cdc0`:
> 
> **1. `EmailAddressConflictError`** — the handler now throws on `__typename !== 'Order'`, and `customer-selector.tsx` swallows the rejection so it isn't unhandled. Conflict keeps the form, unique email creates and closes.
> 
> **2. i18n catalogs** — clean. The only removed `msgstr` is `"Select an address"`, correctly moved to obsolete (`#~`) now that the string is gone from the UI; nothing else was blanked in any of the 32 locales. `dashboard i18n sync` is green.
> 
> CI is green apart from `deploy`, which fails on an empty `VERCEL_TOKEN` — fork PRs don't get the secret. Not yours.
> 
> One thing before merge: please revert `packages/dev-server/graphql/graphql-env.d.ts` to master.
> 
> ```
> git checkout origin/master -- packages/dev-server/graphql/graphql-env.d.ts
> ```
> 
> It got regenerated with `FieldTestPlugin` enabled locally, so the snapshot has `RRP`, `productSpecs` and `seoTitle` — but that plugin is commented out in `dev-config.ts:132` on master, so the committed schema wouldn't match the committed config. Harmless at runtime, but it makes the next regeneration produce a phantom revert diff.
> 
> Approving now — push that one revert and I'll merge.
> 
> Separately, I'll open an issue for the Country field not prefilling in `CustomerAddressForm`; it's pre-existing and reproduces on the untouched address-book dialog, so it's out of scope here.

Sure Thanks @biggamesmallworld  !. I can work on that issue happy to take that up

biggamesmallworld (approved):

biggamesmallworld:
@Ryrahul looks like we need to run the lingui extract once more to get this one ready, all else looks good to go!

biggamesmallworld (approved):

PR #4954, opened 2026-07-12
fix(core, common, dashboard, admin-ui): calculate order summary total value correctly (#3648)

### LINKED ISSUE #3648: Order Summary Widget shows the wrong amount with > 1000 orders
**Describe the bug**
We've discovered a bug in the Vendure Admin UI's Order Summary Widget.

When there are more than 1000 orders, the total amount in the summary is not calculated correctly. This is due to the summary only considering the first 1000 orders.

**To Reproduce**
Include > 1000 orders and check the amount of the orders.

**Expected behavior**
Calculating the amount correct.

 - @vendure/core version: 3.3.5

**Additional context**

Sanjuhub:
# Description

This PR fixes an issue #3648 in the Admin UI Order Summary widget where the total order value was calculated incorrectly when more than 1000 orders existed.

Previously, the widget calculated the total by summing the `items` returned from the `orders` query. Since the `orders` query is paginated with a default page size of 1000, only the first 1000 orders were included in the calculation.

This change introduces a dedicated server-side order summary query that returns:

- Total number of orders
- Total order value
- Channel currency code

The widget now consumes these aggregated values directly instead of calculating the total on the client.

Fixes: #3648

# Breaking changes

# Screenshots
Before fix:
Screenshot 2026-07-12 at 5 57 22 PM

After fix: 
Screenshot 2026-07-12 at 5 51 31 PM

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4954"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786451033&installation_id=128408429&pr_number=4954&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4954&signature=926f69336fb55a77764ba61a985af9d773df97fae98b5e3b2ce4093babb9532e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
Thanks for digging into this, @Sanjuhub — and for the clear write-up of the root cause. The >1000 pagination cap on the client-side sum is a real bug and you diagnosed it well.

We're going to handle this one internally, in a slightly different way. The main reason is that we'd rather not extend the **core Admin API** with a widget-specific `orderSummary` query and `OrderSummary` type: that's a schema surface every deployment then carries, for what is really a dashboard concern. This kind of aggregation is better placed in the **dashboard plugin** — the summary widget can fetch/compute what it needs on the dashboard side without a new core query.

So no changes needed from you here — we really appreciate the contribution and the investigation, it helped confirm the underlying issue. 🙏

Sanjuhub:
@grolmus Thanks for the detailed feedback and for explaining the architectural direction. That makes sense. I'm glad the investigation helped identify the root cause. I'll leave this one to your implementation.

PR #4957, opened 2026-07-13
feat(dashboard): Migrate to Vendure Design System v2

dlhck:
## What this does

Migrates `@vendure/dashboard` to the **Vendure Design System v2** (`@vendure-io/ui@2.0.0-beta` / `@vendure-io/design-tokens@2.0.0-beta`).

**Status: the basic v2 adoption is done.** All wrappers, tokens, molecules, and layout chrome now run on v2. What remains is a **per-view optimization pass** (spacing, hierarchy, and composition tuning on individual list/detail views) to finalize the redesign — that will land in follow-up PRs against `minor`.

### Migration
- Bump to `@vendure-io/ui@^2.0.0-beta.12` and `@vendure-io/design-tokens@^2.0.0-beta.7`; migrate all wrapper imports from the removed `components/ui/*` surface to `components/atoms/*`
- Delete local wrapper patches that were upstreamed into v2 (Select remount workaround, dropdown width fix, wider sheets, heading-styled dialog titles)
- Reconcile the theme Vite plugin with v2's published `brand` slots (no more slot shadowing); emit the new motion and text-style token groups; v2's surface/subtle/info token slots flow through automatically
- Adopt `@vendure-io/design-lint` (`no-raw-colors`) in the ESLint config

### Rebuilt on v2 molecules
- **DataTable** rebuilt as a bridge over the v2 data-table molecule; table controls merged into a settings menu, embedded table titles/actions moved into the header band, redesigned toolbar with unified filter menu and page-aware search placeholders
- **MultiSelect** rebuilt on v2 selection primitives; **date pickers** adopted behind instant-valued bridges; `useLocalFormat` bridged into the v2 `FormatProvider`
- **StatusBadge + per-domain state dictionaries** (orders/payments/fulfillments, job queue, scheduled tasks, customer verification) replace the global state→color map, with tone corrections (cancelled→neutral, queued→neutral, failures→critical, running→progress)
- **LoadingState/EmptyState/ErrorState + scenario illustrations**: first-run vs filtered emptiness with a clear-filters action, error/empty seams in list tables, route error page rebuilt on ErrorState, real async states in the metrics widget
- **password-input, confirm-dialog, copyable-text** replaced by thin adapters over the v2 molecules (exported names/props unchanged, labels stay lingui-translated)

### UI/UX reworks on top of v2
- Page layout chrome rebuilt on v2 primitives; app shell header and channel switcher refined
- Rich text editor UI/UX overhauled
- Configurable operations (promotion/payment/shipping conditions & actions) rendered as editable sentences
- Brand color reintroduced at identity moments; extension action bar buttons no longer render as primary

### Corrections (live bugs found during the audit)
- 15 silently dropped CSS declarations: legacy `hsl(var(--token))` wrappers around what are now oklch tokens
- Font fallback to system-ui: fonts now load via the v2 `css/fonts` export
- Duplicate login error toasts; human-readable field labels in filter badges

### Notable intended visual changes
- v2 repurposes `primary` as a near-black action color; the blue moved to the `brand` slot
- Dialog titles use the `text-style-section-title` typography role
- Badge keeps a deprecated `secondary` alias for extension compatibility; internal call sites migrated to the v2-neutral `default`

## Follow-ups (before this is "finalized")
- **Per-view optimization pass** over the individual views — tracked as follow-up PRs
- Resolve and remove `packages/dashboard/V2_MIGRATION_OPEN_QUESTIONS.md` before this leaves draft
- The DS packages must be promoted from beta to stable `2.0.0` before the September minor release ships

## Verification
- `check-types`, unit tests, codegen, and design-lint (zero violations in `src`) pass in CI
- Dashboard e2e suites are being adapted to the new UI (unified filter menu etc.) — currently red, being brought green before undrafting

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4957"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786535975&installation_id=128408429&pr_number=4957&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4957&signature=dc10795b4bee0b14d3fd19859cd6ce6c55032874f700bd470133375b6eeb325b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/dashboard/V2_MIGRATION_OPEN_QUESTIONS.md
```diff
@@ -0,0 +1,124 @@
...
+   (deliberate fixed contrast over arbitrary user images) behind a justified
+   design-lint disable. If this concept recurs, it should become a named semantic
+   slot (e.g. `overlay-on-media`) in the token package.
+8. **Global visual QA.** v2's `primary` change shifts every primary button, ring and
+   selection accent from blue to near-black — intended by the DS, but the dashboard
+   should get a full visual pass (storybook / dev-server) before this leaves draft.
+
+## API / technical decisions
+
+9. **Badge `secondary` alias.** v2 removed the variant; the dashboard Badge wrapper
```
michaelbromley: This item is stale: the deprecated `secondary` alias described here was removed later in the branch by `d27377666` — `BadgeProps` at HEAD only extends the base variants with `success` and `warning`. The PR description also still claims the alias is kept.

Since this PR targets `minor`, a removed variant with no alias is a real breaking change for existing extensions. Can we either restore the deprecated alias, or make the break explicit in the migration notes/changelog so extension authors aren't surprised on a minor bump?

packages/dashboard/src/lib/index.ts
```diff
@@ -136,11 +144,15 @@ export * from './components/shared/powered-by-vendure.js';
 export * from './components/shared/product-variant-selector.js';
 export * from './components/shared/remove-from-channel-bulk-action.js';
 export * from './components/shared/rich-text-editor/image-dialog.js';
-export * from './components/shared/rich-text-editor/link-dialog.js';
+export * from './components/shared/rich-text-editor/link-bubble-menu.js';
```
michaelbromley, line 147: `LinkDialog` is removed from the public exports here — the commit message for `b92e7ad49` flags it as BREAKING, but it isn't mentioned in either `V2_MIGRATION_OPEN_QUESTIONS.md` or `V2_MIGRATION_NOTES.md`. Any extension importing `LinkDialog` from `@vendure/dashboard` will break at import time. Please add it to the migration notes so it makes it into the changelog.

packages/dashboard/src/lib/utils/state-type.ts
```diff
@@ -1,36 +1,46 @@
...
+ *
+ * Tone rationale (see design-system state dictionary):
+ * - terminal & healthy (Delivered / Settled / Completed) → `success`
+ * - awaiting admin action (ArrangingPayment / ArrangingAdditionalPayment /
+ *   Modifying / Pending) → `warning`
+ * - user/terminal cancellation → `neutral` (a cancelled order is an outcome,
+ *   not a failure)
+ * - hard failure (Declined / Error) → `critical`
+ */
+export const orderStateDictionary = defineStateEntries({
```
michaelbromley, line 18: This dictionary is missing about half of the real state values from core's default processes:

- `OrderState`: `Created`, `Draft`, `AddingItems`, `PaymentAuthorized`, `PaymentSettled`, `PartiallyShipped`, `Shipped`, `PartiallyDelivered`
- `PaymentState`: `Created`, `Authorized`
- `FulfillmentState`: `Created`, `Shipped`

They fall back safely to a neutral tone via `toneFor`, so nothing crashes — but `Shipped` and `PaymentSettled` are two of the most common states an admin sees, and they now render as plain grey badges on the order list and payment/fulfillment detail pages. Can we fill out the dictionary completely? A unit test asserting every state in the default order/payment/fulfillment processes has an entry would prevent this from silently regressing again.

packages/dashboard/eslint.config.js
```diff
@@ -6,12 +6,43 @@ import globals from 'globals'
...
+      ecmaVersion: 'latest',
+      sourceType: 'module',
+      ecmaFeatures: { jsx: true },
+    },
+  },
+  plugins: {
+    '@vendure-io/design': designLint,
+  },
+  rules: {
+    '@vendure-io/design/no-raw-colors': 'error',
```
michaelbromley, line 37: The rule itself looks right (and `src` is clean), but as far as I can see no CI workflow runs eslint for this package — enforcement is currently manual via `bun run lint`. Is that intentional for now, or should a lint step be added to CI so raw colors can't creep back in?

packages/dashboard/vite/vite-plugin-theme.ts
```diff
@@ -121,20 +144,72 @@ function generateThemeInlineBlock(): string {
...
+    const easingLines = Object.entries(easing).map(
+        ([key, value]) => `    --ease-${camelToKebab(key)}: ${value};`,
+    );
+    const durationLines = Object.entries(duration).map(
+        ([key, value]) => `    --transition-duration-${key}: ${value};`,
+    );
+    const animationLines = Object.entries(animation).map(([key, value]) => `    --animate-${key}: ${value};`);
+
+    // Dashboard-specific tokens not present in the base design-tokens.
+    // `brand`/`brand-foreground` are published by @vendure-io/design-tokens v2
```
michaelbromley, line 161: Two doc-drift issues around the brand tokens:

1. `docs/docs/guides/extending-the-dashboard/theming/index.mdx` still documents `brand-lighter`/`brand-darker` as valid theme override keys, but the tokens backing them are gone — the docs should be updated in this PR so they don't describe a non-functional option post-merge.
2. `--color-vendure-brand` is also silently dropped but isn't mentioned in the migration notes — worth adding.

packages/dashboard/src/lib/components/data-table/data-table.tsx
```diff
@@ -107,15 +95,33 @@ function DraggableRow<TData>({ row, isDragDisabled, getRowCanDrag }: Readonly<Dr
...
+// Captures the TanStack table instance the @vendure-io/ui DataTable molecule
+// creates internally, so the DataTableProvider (whose consumers — saved-view
+// buttons etc. — live in the toolbar slot) can expose it. useLayoutEffect so
+// the capture re-render happens before paint. The instance is stable across
+// renders, so the setState bails out after the first call.
+function TableInstanceCapture<TData>({
+    table,
+    onTable,
+}: Readonly<{ table: TableType<TData>; onTable: (table: TableType<TData>) => void }>) {
+    React.useLayoutEffect(() => {
```
michaelbromley: Non-blocking, flagging as tech debt: capturing the table instance by force-rendering the toolbar slot once and grabbing it in a `useLayoutEffect` works today (the instance is referentially stable and the effect fires pre-paint), but it leans on undocumented internals of `UiDataTable` rather than a supported capture API. Fragile against future `@vendure-io/ui` changes — ideally we'd request a proper `onTableReady` callback upstream and simplify this once it exists.

packages/dashboard/src/lib/components/shared/confirmation-dialog.tsx
```diff
@@ -1,57 +1,33 @@
...
-                    >
-                        {confirmText ?? <Trans>Continue</Trans>}
-                    </AlertDialogAction>
-                </AlertDialogFooter>
-            </AlertDialogContent>
-        </AlertDialog>
+        <ConfirmDialog
+            title={title}
+            description={description}
+            onConfirm={onConfirm}
```
michaelbromley, line 25: Minor: the `onConfirm` prop is typed `() => void`, but the underlying v2 `ConfirmDialog` genuinely supports (and behaves differently for) promise-returning handlers. An async handler compiles fine today due to return-type covariance, so nothing is broken — but widening the type to `() => void | Promise<void>` would make the signature match what the component actually does.

michaelbromley (commented):
Non-blocking review of the v2 migration. Overall this is well executed: the bridge/adapter layers are sound, the migration is genuinely complete (no leftover `hsl(var(` wrappers or raw color literals anywhere), the custom-state fallback in `toneFor` handles plugin-defined states safely, the theme plugin's token emission checks out against the real design-tokens v2 source, and the e2e adaptations are faithful (no sleeps, no skipped tests, no loosened assertions — selectors actually got more specific).

That said, a few things worth addressing before this leaves draft, detailed inline:

1. **Stale tracking doc**: `V2_MIGRATION_OPEN_QUESTIONS.md` no longer matches HEAD in at least three places (Badge `secondary` alias claim, item 3 cancel/destructive styling is actually already resolved in code via `isDestructiveTransition()`, and the motion/text-style token emission note is outdated — `41df353a0` does emit them). Worth a doc-vs-HEAD re-audit before deleting it.
2. **Two undocumented public API breaks** for extension authors (Badge `secondary` variant, `LinkDialog` export) — see inline. Since this targets `minor`, these need either restoring or an explicit callout in the changelog/migration notes.
3. **`orderStateDictionary` is missing common core states** (`Shipped`, `PaymentSettled` among others) which now render as neutral grey badges — visible regression, small fix.
4. Smaller points inline on CI lint enforcement, theming docs drift, the DataTable table-instance capture trick, and the `onConfirm` type.

Two test-coverage gaps worth considering, no inline anchor: no test covers a user override of `brand`/`brand-foreground` specifically (existing override tests only cover `background`), and the `vite-plugin-tailwind-source.ts` path-resolution fix in `e8110a0f4` has no regression test, so a future revert to a stale path would only surface in manual visual QA.

dlhck:
Thanks for the thorough review, Michael. Addressed in 5eaa9ee7d:

- Completed the order/payment/fulfillment state dictionary with unit coverage for every default state. Routine and healthy states remain neutral; color is reserved for noteworthy partial/authorized states, actionable waiting states, active progress, and failures.
- Kept the Badge secondary removal as an accepted breaking change, and documented it together with the LinkDialog removal in a permanent Design System v2 migration guide. The temporary migration/open-questions files are gone, and the commit includes a BREAKING CHANGE footer for the generated changelog.
- Updated the theming guide to remove brand-lighter/brand-darker and document the current brand/brand-foreground tokens plus the removed color-vendure-brand alias.
- Added dashboard lint enforcement to CI.
- Widened the local ConfirmationDialog onConfirm type to match the design-system promise-aware contract.
- Updated @vendure-io/ui to 2.0.0-beta.13. The new onTableReady API replaces the toolbar-slot table capture workaround, and the new copy/loading accessibility label props are now wired through localized Dashboard wrappers.
- Added regression coverage for brand overrides and the @vendure-io/ui Tailwind source-root resolution.

Validation: full Dashboard suite passes (503 tests, 4 skipped), build:vite passes, TypeScript passes, lint has zero errors, the i18n catalogs are synchronized, and all 833 MDX documents compile.

Issue #4958, opened 2026-07-13
Scaffolded app ships no initial migration — first deploy against an empty database creates no schema


michaelbromley:
## Problem

A fresh app created with `@vendure/create` ships with `dbConnectionOptions.synchronize: false` and a `migrations` glob, but **no migration files**. As a result, the first deployment against a fresh/empty production database creates **no tables** — the schema is never built, and nothing in the flow explains why.

This is currently masked during local development because `@vendure/create` force-sets `synchronize: true` while populating the seed data, so the local schema is created by synchronize, not by a migration. The gap therefore only surfaces on the first real deployment, where nothing turns synchronize on.

The manual workaround — temporarily enabling `synchronize: true` for the first deploy, or hand-generating an initial migration against a throwaway empty database — is not something a first-time user would know to do, and `synchronize: true` is not acceptable for production.

## Root cause

`generateMigration()` produces a migration by diffing the entity metadata against **whatever database is currently connected** (TypeORM's `SchemaBuilder.log()`):

- against an **empty** database → the diff is the full schema (exactly the "initial migration" we want);
- against a **populated** database → the diff is empty, so no migration can be generated.

So once the local database has been built (via synchronize), there is no straightforward way to bootstrap the initial migration — you'd need to point at an empty database of the same dialect.

## Proposed solution

Two complementary changes:

### 1. Generate an initial migration at scaffold time (`@vendure/create`)

At scaffold time the configured database is genuinely empty, so a plain `generateMigration({ name: 'init' })` yields the complete baseline. The scaffold then runs that migration to build the schema and populates with `synchronize: false`. Result: every new project ships with a schema-creating migration from day one, the first deployment builds its schema via migrations, and local dev matches production (no `synchronize`).

### 2. Add a shadow-database baseline mode to the CLI

`vendure migrate --generate <name> --from-empty` — generate a complete "zero to current schema" baseline **even when the configured database is already populated**, by diffing against a freshly-provisioned, disposable empty database of the same dialect (the approach used by Prisma's shadow DB and Atlas's dev DB). For Postgres/MySQL this creates and drops a temporary database on the same server; for SQLite it uses an in-memory database. This covers CI, regeneration, and scaffold paths where no empty database is available at generate time.

## Prototype status

A working prototype exists for both parts (in `@vendure/core`, the CLI, and `@vendure/create`) and has been verified against Postgres 16 and `better-sqlite3`:

| Scenario | Result |
|---|---|
[...]

PR #4959, opened 2026-07-13
feat(core): bootstrap initial migration from an empty database

michaelbromley:
## Description

A fresh app created with `@vendure/create` ships with `synchronize: false` and a `migrations` glob but **no migration files**, so the first deployment against an empty database creates no schema. This is masked locally because the scaffold force-sets `synchronize: true` while populating seed data, so the gap only surfaces on the first real deploy.

The root cause is that `generateMigration()` diffs the entity metadata against **whatever database is currently connected** — against a populated database the diff is empty, so once the local schema exists there is no straightforward way to bootstrap the initial migration.

This PR addresses that in two complementary ways:

### 1. `@vendure/create` generates an initial migration during scaffold

At scaffold time the configured database is genuinely empty, so a plain `generateMigration({ name: 'init' })` yields the complete baseline. The scaffold now generates that migration, runs it to build the schema, and populates with `synchronize: false`. Every new project ships with a schema-creating migration from day one, the first deployment builds its schema via migrations, and local dev matches production.

### 2. `vendure migrate --generate <name> --from-empty`

A new flag to generate a complete "zero to current schema" baseline **even when the configured database is already populated**, by diffing against a freshly-provisioned, disposable empty database of the same dialect (the shadow-database approach used by Prisma and Atlas). For Postgres/MySQL this creates and drops a temporary database on the same server; for SQLite it uses an in-memory database. Useful for CI, regeneration, and any scaffold path without an empty database at generate time.

## Verification

Verified against real Postgres 16, MySQL 8, and better-sqlite3:

| Scenario | Result |
|---|---|
| `--from-empty` against an empty DB (Postgres / MySQL) | full baseline (87 `CREATE TABLE`) |
| plain generate against a populated DB | "No changes found" — the bug |
| `--from-empty` against a populated DB | full baseline anyway |
| shadow DB lifecycle | created and dropped, no leaks, target DB untouched |
| scaffold path (generate → runMigrations → populate) | 88 tables created and recorded in the `migrations` table |
| SQLite `--from-empty` (in-memory) | full baseline, real db file untouched |

Added an integration test (`packages/core/src/migrate.spec.ts`) covering empty→baseline, populated→nothing, and populated+`--from-empty`→baseline. Existing migrate tests still pass; `@vendure/core`, `@vendure/cli`, and `@vendure/create` typecheck cleanly.

## Notes

- The shadow-database path requires the database user to have `CREATE DATABASE` privilege (as with Prisma's shadow DB). A missing privilege fails with a clear, actionable error. Option 1 (scaffold-time generation) never needs this, since the real database is already empty at that point.
- Follow-up hardening (e2e matrix, bring-your-own-DB / `sql.js` scaffold fallback, docs) is tracked in #4958.

Relates to #4958

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4959"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786540443&installation_id=128408429&pr_number=4959&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4959&signature=739402a4caf24202924789fbd6f59775ca67d4d6e58221d5e825510d48d28fb9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


dlhck (changes requested):
I found two correctness issues that should be addressed before merge:

1. **[P1] The shadow connection can still target the configured real database** (`packages/core/src/migrate.ts:309-313`). The code spreads `baseOptions` and overrides only the top-level `database`. TypeORM parses `url` after explicit options, so a database embedded in `url` wins over `database: shadowName`; replicated Postgres/MySQL drivers likewise source the database from `replication.master` / `replication.slaves`. In either supported configuration, `createSchemaBuilder().log()` runs against the configured database rather than the newly-created shadow database. Against a populated database this silently returns no baseline, which is the exact failure `--from-empty` is intended to fix. Please construct shadow-specific connection options that rewrite or remove `url` and update replication credentials too, and add coverage for at least URL-based configuration (ideally replication as well).

2. **[P2] Generated code imports `typeorm`, but only pnpm projects declare it** (`packages/create/src/helpers.ts:572-575`). Resolving a transitive dependency from application source relies on package-manager hoisting, which is not part of the npm/Bun install contract and can be disabled (for example npm nested installs or isolated linkers). Those scaffolded apps then fail to compile/load the generated initial migration even though their manifest is otherwise valid. Since every generated migration imports `MigrationInterface` and `QueryRunner` from `typeorm`, the scaffold should declare the matching TypeORM range for every package manager, or generate imports from a package that is already direct.

The current head is `46ceced3f04ab17a85d07c2a182e9af2ad084fa6`; all CI checks are green, but the matrix uses top-level host/database options and default hoisting, so it does not exercise either case.

oliverstreissi (changes requested):
Review by Nigel (AI reviewer), posted by Oliver.

The core shadow-database design is sound and the tests exercise the real path instead of mocking it away. But the scaffold integration has a silent-failure hole that reintroduces the exact bug this PR fixes, and the cleanup path has an ordering flaw.

## Blocking

**1. The scaffold silently ships *no* initial migration when the target DB isn't empty — the PR's own bug, reintroduced on the retry path.**

In `packages/create/src/create-vendure-app.ts`:

```ts
await generateMigration(config, { name: 'init', outputDir: migrationsDir });
await runMigrations(config);
```

Both return values are discarded. `generateMigration` returns `undefined` and writes nothing when the diff is empty. Concrete failure: a user runs `@vendure/create`, it dies mid-populate, they re-run into a fresh directory pointing at the **same Postgres DB** — which now has the full schema from attempt one. `checkDbConnection` passes (it only checks existence), `generateMigration` finds no diff, prints a yellow warning into the spinner noise, and the scaffold completes "successfully" with an empty `src/migrations/` and `synchronize: false`. That project's first production deploy against an empty DB fails exactly as described in the PR's problem statement. Either check the return value and fail loudly, or use `fromEmpty: true` here — the tool was built for precisely this scenario.

**2. `runMigrations` swallows errors in the create context.**

Same location. `runMigrations` only *throws* when `VENDURE_RUNNING_IN_CLI` is set; otherwise it logs, sets `process.exitCode = 1`, and returns `[]`. `create-vendure-app` never sets that env var. So if running the init migration fails, the scaffold sails on into `populate()` with `synchronize: false` against an empty schema and dies with an unrelated "relation does not exist" error. Check `migrationsRan.length` or set the env var so the error propagates to the existing catch.

**3. Shadow DB cleanup is skipped if `connection.close()` throws.**

`packages/core/src/migrate.ts`:

```ts
} finally {
    await connection.close();
    if (cleanup) {
        await cleanup();
    }
    resetConfig();
}
```

If `close()` rejects, the shadow DB is orphaned *and* the admin connection stays open (process hang). Nest it: `try { await connection.close(); } finally { await cleanup?.(); resetConfig(); }`. The cleanup callback inside `provisionShadowDatabase` already gets this right (`try/finally` around the DROP) — same discipline needed one level up.

## Should fix

**4. The sqljs "empty" database isn't guaranteed empty.**

`emptySqliteOptions` unsets `location` but spreads `baseOptions`, which for sqljs can carry `database: Uint8Array` — the canonical way to boot sqljs from saved bytes. Pass a populated buffer and the "empty shadow" is a fully populated database, yielding an empty baseline — the exact failure mode `fromEmpty` exists to prevent. Unset `database` (and `autoSaveCallback`) too. The scaffold template happens to use `new Uint8Array([])`, but this is a public API.

**5. The e2e spec copy-pastes connection config from e2e-common.**

`packages/core/e2e/migrate-from-empty.e2e-spec.ts` reimplements `getDbConfig` from `e2e-common/test-config.ts` — host, hardcoded creds, the `CI ? E2E_*_PORT : default` dance. The file already imports `TEST_SETUP_TIMEOUT_MS` from that module. When CI ports or credentials change, this spec breaks separately. Export `getDbConfig` (or the per-dialect options) and reuse it.

**6. Shadow-DB name uniqueness is per-process only.**

The comment says a counter + timestamp "guarantees a unique shadow database name" — only within one process. Two concurrent `vendure migrate --from-empty` runs against the same server (CI matrix, shared staging DB) can collide on the same millisecond, and the "database exists" error then gets wrapped in the misleading "Ensure the configured database user has permission to create databases" message. Append `process.pid` or a random hex chunk — one line — and the comment becomes true.

## Nits

- ~60 lines of the mysql/backtick `forEach` blocks were reindented wholesale inside the new `try`, which hides the actual change. Those four blocks were already copy-paste duplicates of one operation — this was the moment to extract a `formatQuery(query, quoteChar)` helper.
- The unsupported-driver error lists "postgres, mysql, mariadb, better-sqlite3, sqlite, sql.js" but the switch also handles `aurora-postgres`/`aurora-mysql`. Has `CREATE DATABASE` over the Aurora Data API actually been verified, or is that case aspirational?
- The MySQL shadow DB is created with the *server-default* charset/collation, ignoring any `charset` in `baseOptions` — the baseline can differ subtly from what the real target produces.
- `migrate.spec.ts`: the three tests are order-dependent (test 1 populates the DB tests 2/3 rely on). Works under vitest's sequential default; breaks under shuffle.
- Interactive `vendure migrate` mode never offers from-empty — flag-only. Fine for v1, worth a follow-up. Likewise `--from-empty` combined with `--run` is silently ignored.
- `helpers.spec.ts`: the new `import { TYPEORM_VERSION } from './constants'` lands after the `./helpers` import block — check import-order lint.
- The docstring claim that fromEmpty is useful "in CI where no populated database is available" is misleading for Postgres: the admin connection connects to the *configured* database, so that database must still exist. MySQL can connect database-less; Postgres can't.

## What's genuinely good

The `try/finally` around generation fixes a pre-existing leak (previously an error in `createSchemaBuilder().log()` skipped `connection.close()` and `resetConfig()` entirely). `createConnectionOptions` forcing `synchronize: false / migrationsRun: false / dropSchema: false` means the shadow and admin connections inherit safe options. The shadow name is fully internally generated, so the interpolated `pg_terminate_backend`/`CREATE DATABASE` strings are not an injection surface. The `TYPEORM_VERSION` pin-sync test (reading core's actual package.json) is exactly the kind of guard that prevents the "two TypeORM copies, broken decorators" bug nobody would diagnose in under a day. And the unit test triad — baseline works / plain generate reproduces the bug / fromEmpty fixes it — documents the regression in executable form.

## Verdict

Request changes. Items 1–3 must be fixed: 1 and 2 mean the headline feature silently fails on realistic paths, and 3 leaks the resource the whole design promises to clean up. Items 4–6 are small enough to land in the same PR. The nits are the author's call.

michaelbromley:
Thanks for the reviews guys. Regarding the point of including `typeorm` as a default top-level dep for _all_ package managers.

I opted against that in the scope of this PR. The reason is that it touches on a larger issue of how we handle these peer deps in general - in this case it’s `typeorm` but generally when working with Vendure you also run into it with `@nestjs/core` `@nestjs/graphql` and many others that you need to import into your services and resolvers.

I consider this a separate issue and I’m not fully decided on the best solution, so for now I decided to limit the scope to where absolutely needed, and currently that just pnpm.

oliverstreissi (approved):

PR #4962, opened 2026-07-14
fix(dashboard): Don't persist empty translations for unfilled languages

### LINKED ISSUE #4885: React dashboard saves empty translation rows for unfilled languages, breaking language fallback (search shows empty)
**Describe the bug**

When you create or edit a product in the new React dashboard with several languages configured, it writes
an empty translation row for every configured language you didn't fill in (empty `name`, `slug`,
`description`). With `uk`, `en`, `pl` configured, filling only `uk` and `en` still creates an empty `pl`
row.

These empty rows break language fallback. With a real `pl` translation absent, a product should fall back
to the default language. Once an empty `pl` row exists, code that looks up "the `pl` translation" finds the
empty row instead of falling back, so it shows nothing.

It comes from `ensureTranslationsForAllLanguages` in the form engine, which seeds the form with an entry
per available language and submits the unfilled ones. It happens in the React dashboard, not the old
Angular admin, and it also happens when creating in a single language (you get an empty row for each of the
others).

**To Reproduce**

1. Configure 3 languages (e.g. uk, en, pl).
2. In the React dashboard, create a product, fill only uk and en, and save.
3. Check the DB: there's a `product_translation` row for `pl` with empty `name`/`slug`/`description`.
4. Request the product in `pl` via the Shop API. Listing/search results show an empty name instead of the
   English fallback.

Confirmed in our DB on a freshly created product:

```
languageCode | name           | slug         | description
uk           | тестовий товар | test-product |
en           | test-product   |              |
pl           |                |              |     <-- empty row, created automatically
```

**Expected behavior**

Only languages the user actually filled should be saved. A language with no content should have no
translation row, so the normal fallback applies.

**Actual behavior**

An empty translation row is created for every configured language. Fallback then breaks wherever a consumer
looks the translation up by language and finds the empty row.

**Screenshots/Videos**

https://github.com/user-attachments/assets/76ba8cca-b8c4-4c01-b229-adb724abefd0

**Error logs**

```
None — the empty rows are written silently.
```

- @vendure/core version: 3.6.4 (also @vendure/dashboard 3.6.4)
- Nodejs version: 24
- Database (mysql/postgres etc): PostgreSQL 15
- Operating System (Windows/macOS/Linux): Linux (server) / macOS (dev)
- Browser (if applicable): Chrome (React dashboard)

**Configuration**

Default setup with 3+ languages configured on the channel. Nothing custom is needed.

**Minimal reproduction**

Create a product via the dashboard with N languages configured, filling fewer than N. The createProduct
input includes an empty translation for each unfilled language (from `ensureTranslationsForAllLanguages`),
and they are persisted.

**Workaround**

Delete the empty rows (review the condition per entity/fields first):

```sql
DELETE FROM product_translation
WHERE name = '' AND slug = '' AND (description = '' OR description IS NULL);
```

**Additional context**

Where it comes from:

- `ensureTranslationsForAllLanguages` in
  `packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx` adds an empty translation for
  every `availableLanguage` not already present, and it is applied to the create defaults. The submit sends
  them unchanged (no filtering of empty entries). The Angular admin only sends languages that were actually
  edited, which is why it doesn't happen there.

Why fallback still breaks even though core has a field-level fallback:

- `translateEntity` (core) does field-level fallback for empty values, so a direct `product` / `products`
  query masks the empty row and shows the default language.
- The search index does not go through `translateEntity`. The DefaultSearchPlugin indexer stores
  `getTranslation(...).name` raw
  (`packages/core/src/plugin/default-search-plugin/indexer/indexer.controller.ts`), and `getTranslation` is
  `find(requested) || find(default) || translations[0]` — it only falls back when the translation is
  *missing*. An empty row exists, so `find(requested)` returns it and the index stores an empty name.
  Storefront listings (which use the `search` query) then show empty for that language. So on a stock
  install the visible breakage is in search/listings.

Suggested fixes:

1. Primary: the dashboard should not persist empty translations — drop entries the user didn't fill before
   sending the create/update input.
2. Defense-in-depth: the DefaultSearchPlugin indexer (`getTranslation`) could treat an empty translation as
   absent, so a stray empty row doesn't poison the index.

Reproducible on stock Vendure 3.6.4: the dashboard creates the empty rows and the stock search plugin
indexes them empty.

grolmus:
## Summary

Stops the React dashboard from persisting **empty translation rows for languages the user never filled in**. Creating/editing an entity with N configured languages while filling fewer than N wrote an empty `*_translation` row (empty `name`/`slug`/`description`) for each unfilled language. Those empty rows break language fallback — most visibly the search index stores an empty name, so storefront listings show blank for that language. (#4885 / OSS-579)

## Root cause

`ensureTranslationsForAllLanguages` (form engine) seeds a translation row for **every** configured language so any of them can be edited in the form. That's correct for the form UI, but the submit pipeline in `use-generated-form.tsx` sent the untouched (empty) rows unchanged, so they were persisted. The Angular admin only sent edited languages, which is why it didn't happen there.

Core's `translateEntity` masks this on direct `product`/`products` queries (field-level fallback), but the `DefaultSearchPlugin` indexer reads the raw translation via `getTranslation`, which only falls back when a translation is *missing* — an empty row is found instead, so the index stores an empty name.

## Fix

New pure helper `stripEmptyTranslations(values, fields)` in the form-engine `utils.ts`, added to the submit pipeline (for both create and update). It drops a translation entry only when it has **no `id`** (i.e. it was seeded, not an existing persisted row) **and** every field other than `languageCode` is empty. Existing translations (with an `id`) and any language the user actually filled are always kept. It's field-metadata driven (detects a translations array by a `languageCode` sub-field), so it works at any nesting depth and for any translatable sub-entity, not just top-level product translations.

## Tests

`utils.spec.ts` — unit tests using the real `CreateProduct` document: drops a seeded empty translation while keeping the filled one; keeps an existing (`id`) translation even when empty; keeps a translation with any field filled; leaves clean input unchanged. Full form-engine suite green (81/81), `tsc` clean.

## Out of scope / follow-up

- **Defense-in-depth (separate PR):** the issue also suggests making the `DefaultSearchPlugin` indexer's `getTranslation` treat an empty translation as absent, so a stray empty row from *any* source can't poison the index. Left as a `@vendure/core` follow-up.
- A dashboard e2e driving the multi-language create flow would be a good addition but needs a ≥2-language channel in the e2e fixture; noting rather than expanding the fixture here.

Fixes #4885

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4962"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786612301&installation_id=128408429&pr_number=4962&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4962&signature=a24fdd20d9c9f04df5e082ebb4270be9ff64d4ece8c4ccb908b0b9f7e91c6ad5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


packages/dashboard/src/lib/framework/form-engine/utils.ts
```diff
@@ -203,6 +203,75 @@ export function stripNullNullableFields<T extends Record<string, any>>(values: T
...
+        for (const field of fieldDefs) {
+            const value = obj?.[field.name];
+            if (!value || typeof value !== 'object' || !field.typeInfo) {
+                continue;
+            }
+            if (Array.isArray(value)) {
+                const isTranslationsArray = field.typeInfo.some(f => f.name === 'languageCode');
+                if (isTranslationsArray) {
+                    obj[field.name] = value.filter(entry => !isSeededEmptyTranslation(entry));
+                }
```
biggamesmallworld, line 260: The filter has no floor — if every row is seeded-empty, this submits `translations: []`.

Reachable from the UI: `getZodTypeFromField` maps non-nullable `String` to a bare `z.string()` (no `.min(1)`), so a completely blank create form passes validation. Confirmed against this branch. Previously that created a product with an empty name; now it sends zero translations into `ProductService.create`.

```suggestion
                if (isTranslationsArray) {
                    const kept = value.filter(entry => !isSeededEmptyTranslation(entry));
                    // Never strip every row: a fully-empty form would otherwise submit
                    // `translations: []`. Leave the input as-is and let validation surface
                    // the empty required fields.
                    obj[field.name] = kept.length ? kept : value;
                }
```

Verified locally: all-empty preserved, mixed case still strips, 37 existing tests green. Worth a test pinning it.

packages/dashboard/src/lib/framework/form-engine/utils.spec.ts
```diff
@@ -43,6 +44,63 @@ describe('removeEmptyIdFields', () => {
...
+    });
+
+    it('keeps a translation where at least one field is filled', () => {
+        const values: CreateProductInput = {
+            input: {
+                translations: [{ languageCode: 'pl', name: '', slug: 'polski-slug', description: '' }],
+            },
+        };
+        const result = stripEmptyTranslations(values, fields());
+        expect(result.input.translations).toHaveLength(1);
```
biggamesmallworld: Length-only assertion — passes even if the row's contents were mangled. Every sibling suite here uses exact `toEqual`:

```suggestion
        expect(result.input.translations).toEqual([
            { languageCode: 'pl', name: '', slug: 'polski-slug', description: '' },
        ]);
```

Gaps worth covering while you're here:

- **All rows empty** — the `translations: []` case from the other comment. Untested and currently wrong.
- **Update with a persisted row + a seeded row** — the actual #4885 scenario. The id test above casts to `any` to put an `id` on a `CreateProductInput`; creates don't have translation ids, so use an `UpdateProduct` document and the `as any` disappears.
- **`customFields: {}` on an otherwise-empty row** — I checked, the recursive `isEmptyValue` object branch handles it correctly, but it's the most fragile code in the diff with zero coverage.
- **Does not mutate the input** — also verified working; `convertEmptyStringsToNull`, `stripNullNullableFields` and `transformRelationFields` all pin this, so this should too.
- **`null` input** — there's a guard at the top of `stripEmptyTranslations`, both sibling suites test theirs.

biggamesmallworld:
I actually think we can improve the approach by using RHF dirty state instead of inferring intent from values.

`ensureTranslationsForAllLanguages` constructs the seeded rows, and `dirtyFields` already records exactly which ones the user touched — a seeded row never typed into is by definition not dirty. That makes `isEmptyValue` and `isSeededEmptyTranslation` (and the `id` guard, and its unstated dependency on `removeEmptyIdFields` running first) unnecessary.

It's also less code: 49 lines vs 69, and that 49 already includes the floor guard from my other comment.

```ts
/**
 * @description
 * Removes translation rows the form seeded but the user never edited. The form engine
 * seeds a row for every configured language so any of them can be edited; submitting the
 * untouched ones persists empty translation rows that break language fallback (#4885).
 * Driven by react-hook-form's `dirtyFields`, which records exactly which rows the user
 * touched.
 */
export function stripUntouchedTranslations<T extends Record<string, any>>(
    values: T,
    fields: FieldInfo[],
    dirtyFields: any,
): T {
    if (!values) {
        return values;
    }
    const result = structuredClone(values);

    function process(obj: any, dirty: any, fieldDefs: FieldInfo[]) {
        for (const field of fieldDefs) {
            const value = obj[field.name];
            if (!value || typeof value !== 'object' || !field.typeInfo) {
                continue;
            }
            const dirtyValue = dirty?.[field.name];
            if (Array.isArray(value)) {
                if (field.typeInfo.some(f => f.name === 'languageCode')) {
                    const kept = value.filter((_, i) => isDirty(dirtyValue?.[i]));
                    obj[field.name] = kept.length ? kept : value;
                }
                for (const [i, item] of obj[field.name].entries()) {
                    process(item, dirtyValue?.[i], field.typeInfo);
                }
            } else {
                process(value, dirtyValue, field.typeInfo);
            }
        }
    }

    process(result, dirtyFields, fields);
    return result;
}

function isDirty(value: any): boolean {
    if (typeof value === 'object' && value !== null) {
        return Object.values(value).some(isDirty);
    }
    return value === true;
}
```

Wiring is one line — `form.formState.dirtyFields` is already in scope in `onSubmitWrapper`.

The main reason to prefer it: `isEmptyValue` treats `0` and `false` as *filled*, and `getDefaultValueFromField` seeds non-nullable `Boolean` → `false`, `Int`/`Money` → `0`, enums → first member. So as soon as a translation input has a non-string field (plugin-extended translation, custom translatable field type) every seeded row looks filled and the current fix silently stops working. Core translations are all-string today, so nothing would catch it. Dirty state doesn't care about value types. I verified both behaviours locally.

Tests stay pure, `dirtyFields` is just a nested object of booleans:

```ts
const dirty = { input: { translations: [{ name: true }, {}] } };
```

One thing to check if you take this: RHF's `formState` is a Proxy that only computes what you subscribe to during render. If `dirtyFields` isn't read during render it can come back empty in the submit handler — which combined with the floor guard fails *silently* (every row looks untouched, guard keeps them all, fix no-ops). Destructuring `dirtyFields` during render fixes it, but worth confirming against the real form rather than unit tests.

biggamesmallworld (approved):
I think this solution is correct, but leaves 1 gap and there is an opportunity to improve the use of RHF. Approving now but please address the comments first!

grolmus:
Great catch, Will — the missing floor is a real regression, not just a test gap. Fixed in b309bf4741.

**The fix** — took your suggestion verbatim: keep the rows when the filter would remove all of them, and let validation surface the empty required fields rather than silently submitting `translations: []`.

```ts
const kept = value.filter(entry => !isSeededEmptyTranslation(entry));
obj[field.name] = kept.length ? kept : value;
```

**Tests** — addressed all your points:

- Tightened the length-only assertion at `utils.spec.ts:88` to exact `toEqual`.
- **All rows empty** — added, and verified it goes red without the fix (`[]` instead of the two rows) and green with it.
- **Update with a persisted row + a seeded row** — added the real #4885 scenario. As you suggested, it uses an `UpdateProduct` document, so the `as any` is gone — creates have no translation ids, updates do.
- **`customFields: {}` on an otherwise-empty row** — added; confirms the recursive `isEmptyValue` object branch drops it.
- **Does not mutate the input** — added, snapshot-compared.
- **`null` input** — added.

40 tests green. Thanks for the thorough pass — the update/`as any` point in particular made the coverage honest.

biggamesmallworld:
@grolmus the blockers are fixed but can you take a look at the RHF suggestion? If you think it is invalid/unnecessary, feel free to disregard

grolmus:
Reworked to RHF dirty state as you suggested, @biggamesmallworld — 59adf4db6c. You were right that the value-based approach had a real hole.

**The rewrite** — `stripEmptyTranslations` → `stripUntouchedTranslations(values, fields, dirtyFields)`, driven by `dirtyFields` exactly as you proposed. Drops `isEmptyValue`, `isSeededEmptyTranslation`, the id guard, and the ordering dependency on `removeEmptyIdFields`; keeps the floor guard. 49 lines vs 69.

I added a unit test that pins the exact hole you identified — an untouched row whose values are `false`/`0` (a non-string translation field's seeded default). The old value-based check kept it; the dirty-state one correctly drops it.

**On the Proxy caveat** — good flag, and I verified it against the real form rather than trusting unit tests. I wired `const { dirtyFields } = form.formState;` in the hook body (read during render). I then added an e2e in `translation-placeholders.spec.ts` that drives the actual form: with German in the channel's available languages, it creates a product, fills only English, and asserts the request carries a single `en` translation with no empty `de` row.

Interesting result: I also ran that e2e with the subscription deliberately broken (reading `form.formState.dirtyFields` only inside the submit handler) — and it **still passed**. Because the form is `mode: 'onChange'` and re-renders as the user types, `dirtyFields` ends up populated regardless in this path. So the caveat doesn't bite here in practice. I kept the destructure-during-render anyway: it's the correct, documented way to depend on `formState`, and other form contexts (e.g. draft-order custom fields) may not re-render the same way — cheap insurance against a real react-hook-form footgun.

Verification: 40 unit tests + 7 e2e (the full `translation-placeholders` suite, including the new case) green locally; check-types clean.

Thanks again — the non-string-field hole was a genuinely better catch than the original approach deserved.

packages/dashboard/src/lib/framework/form-engine/utils.ts
```diff
@@ -203,6 +203,75 @@ export function stripNullNullableFields<T extends Record<string, any>>(values: T
...
+        for (const field of fieldDefs) {
+            const value = obj?.[field.name];
+            if (!value || typeof value !== 'object' || !field.typeInfo) {
+                continue;
+            }
+            const dirtyValue = dirty?.[field.name];
+            if (Array.isArray(value)) {
+                const isTranslationsArray = field.typeInfo.some(f => f.name === 'languageCode');
+                if (isTranslationsArray) {
+                    const kept = value.filter((_, i) => isDirty(dirtyValue?.[i]));
```
biggamesmallworld: Sorry, my suggestion was incomplete and left a single case open, here it is. If you can fix this we can close the currently open update path and keep the empty translations from being created. Sorry!

**The case I missed.** On an update, react-hook-form's `values` prop resets the form and promotes the entity to `_defaultValues`:

```js
// react-hook-form 7.71.2, index.esm.mjs:2918
control._reset(props.values, { keepFieldsRef: true, ...control._options.resetOptions });
// :2262 — no keepDefaultValues passed
if (!keepStateOptions.keepDefaultValues) { _defaultValues = updatedValues; }
// :2358 — no keepDirtyValues / keepDefaultValues / keepDirty
dirtyFields: ... : {},
```

So nothing is dirty until the user types, and an untouched *seeded* row is indistinguishable from an untouched *persisted* one. Change a product's Enabled toggle, price, stock or facets without touching translations and no row is dirty, the floor guard keeps all of them, the seeded `de` row lands in `TranslationDiffer.toAdd`, and core creates the empty translation. That's #4885 on the most common edit path, and it re-adds the row on every save.

**Reproduced**, same test against a real form on `59adf4d` vs `b309bf4` — product with only an `en` translation, German on the channel, toggle Enabled, save:

```
59adf4d (dirty state)   submitted: [ {id: T_21, en, ...}, {de, name: "", slug: "", ...} ]
                        persisted: en, de          <- empty de created
b309bf4 (value-based)   submitted: [ {id: T_21, en, ...} ]
                        persisted: en              <- correct
```

**The fix: keep a row if it is dirty _or_ persisted.** The two predicates are complementary rather than redundant — `id` cannot carry create (no row has one yet, everything would drop and the floor guard would keep it all), and dirty cannot carry update (nothing is dirty after the reset). Each covers exactly what the other is blind to.

Importantly this keeps the `0`/`false` hole closed, which was the whole point of moving to dirty state: there is still no value inspection anywhere, so an untouched row seeded with `false` / `0` / an enum member has no `id` and is still dropped. Your unit test pinning that case passes unchanged.

It also drops the ordering dependency on `removeEmptyIdFields` (`id: ''` counts as not-persisted either way), and makes the doc-block sentence true as written — an untouched persisted row really is never dropped now.

```suggestion
                    const kept = value.filter((entry, i) => isDirty(dirtyValue?.[i]) || isPersisted(entry));
```

Verified locally with this change: the update case above passes, your create-path e2e passes, all 40 unit tests pass unchanged, and the full `translation-placeholders` suite is 7/7.

One edge I'd accept rather than chase: typing into the German row and then clearing it leaves it dirty with no `id`, so an empty `de` row is submitted. Closing that means reintroducing value inspection and the type hole with it, and the user did touch the field.

Happy to push the update-path e2e as well — nothing in the suite covers editing an existing product today, which is why this got through. Just say the word.

packages/dashboard/src/lib/framework/form-engine/utils.ts
```diff
@@ -203,6 +203,75 @@ export function stripNullNullableFields<T extends Record<string, any>>(values: T
...
+    process(result, dirtyFields, fields);
+    return result;
+}
+
+function isDirty(value: any): boolean {
+    if (value != null && typeof value === 'object') {
+        return Object.values(value).some(isDirty);
+    }
+    return value === true;
+}
```
biggamesmallworld, line 279: The `isPersisted` helper for the suggestion above.

```suggestion
function isDirty(value: any): boolean {
    if (value != null && typeof value === 'object') {
        return Object.values(value).some(isDirty);
    }
    return value === true;
}

/**
 * A row that already exists in the database carries an `id`. Dirty state alone cannot identify
 * these: react-hook-form's `values` prop resets the form and promotes the entity to
 * `defaultValues`, so on an update nothing is dirty until the user types — an untouched persisted
 * row and an untouched seeded row look identical. The `id` is the only thing that separates them.
 */
function isPersisted(entry: any): boolean {
    return !!entry && typeof entry === 'object' && entry.id != null && entry.id !== '';
}
```

grolmus:
Reworked to your `dirty || persisted` hybrid, @biggamesmallworld — thanks, that's the right call and the update-path hole was a real one.

## The fix (52395638e)

Took the suggestion verbatim:

```ts
const kept = value.filter((entry, i) => isDirty(dirtyValue?.[i]) || isPersisted(entry));
```

plus the `isPersisted` helper (`id != null && id !== ''`). `dirty` carries create (no ids yet), `id` carries update (nothing is dirty after RHF resets from `values`), and since there's still no value inspection anywhere the `0`/`false`/enum-default hole stays closed.

## Tests

- Replaced the old update test, which wrongly assumed the persisted row would be *dirty* — exactly the false premise that masked this bug. The new one is the real scenario: on update, edit only a non-translation field (toggle Enabled), nothing in `translations` is dirty, and the seeded `de` row must still drop by its missing `id`. Verified it goes **red** under the old dirty-only filter and green under the hybrid.
- Added a mixed-update unit test (973a1b11b): an untouched persisted row + a newly-typed dirty row + a seeded untouched row in one payload — kept, kept, dropped.
- Added the update-path e2e you offered to write: edits Laptop, toggles Enabled only, and asserts the outgoing `UpdateProduct` carries the persisted `en` translation and no empty `de` row. 41 unit tests + the `translation-placeholders` e2e green locally.

On the edge you flagged — typing into `de` then clearing it leaves it dirty with no `id`, so an empty `de` is submitted — agreed that's the right thing to accept; closing it would mean reintroducing value inspection and the type hole with it, and the user did touch the field.

biggamesmallworld (approved):

packages/dashboard/e2e/tests/catalog/translation-placeholders.spec.ts
```diff
@@ -224,6 +224,90 @@ test.describe('Translation fallback placeholders', () => {
...
+    test('updating a non-translation field submits only the existing translation, not a seeded empty one', async ({
+        page,
+    }) => {
+        await goToLaptopProduct(page);
+        const productId = new URL(page.url()).pathname.split('/').pop() as string;
+
+        const dp = detailPage(page);
+        // Toggle Enabled to make the form dirty *without* touching any translation field.
+        const enabledSwitch = dp.formItem('Enabled').getByRole('switch');
+        const wasEnabled = (await enabledSwitch.getAttribute('data-state')) === 'checked';
```
biggamesmallworld: This is why `dashboard e2e (2)` is failing. I reproduced it locally — deterministic, fails on the retry too — and probed the DOM:

```
PROBE data-state : null        <- attribute doesn't exist
PROBE aria-check : "true"
PROBE isChecked  : true
PROBE wasEnabled : false  -> setChecked( true )
PROBE btn disabled: true       <- no-op, form never dirtied

(forcing a real click)
PROBE btn disabled after click: false
```

`Switch` doesn't render a `data-state` attribute — it exposes `aria-checked`. So `getAttribute('data-state')` is `null`, `wasEnabled` is `false`, and `setChecked(true)` is a no-op on an already-checked switch. Nothing gets clicked, the form never becomes dirty, and `Update` stays disabled until `clickUpdate` times out.

```suggestion
        const wasEnabled = await enabledSwitch.isChecked();
```

`isChecked()` reads `aria-checked` and returned `true` correctly in the probe.

Worth noting the failure is silent in two ways: `setChecked` doesn't complain about a no-op, and the `expect(Update).toBeDisabled()` at the end of `goToLaptopProduct` passes for the wrong reason — it's satisfied by `!isDirty`, so it never proves the button *can* enable. An `await expect(enabledSwitch).not.toBeChecked()` (or `.toBeChecked()`) right after the toggle would turn a future regression here into a clear failure instead of a 10s timeout further down.

The fix itself is unaffected — forcing a real click enables Update immediately, and the other 7 tests in the file pass, including the create-path one guarding the `dirtyFields` Proxy subscription.

grolmus: Applied verbatim in `1c09196fd` — thanks for the full DOM probe, that pinned it exactly. `isChecked()` reads `aria-checked` and the toggle now dirties the form as intended.

Also took your robustness note: added `await expect(Update).toBeEnabled()` right after the toggle, so a future no-op fails loudly at the toggle instead of timing out 10s later in `clickUpdate`. CI is green now, `dashboard e2e (2)` included.

That test-only change is the only diff since your approval (it got dismissed on the push) — mind re-approving so I can merge?

grolmus:
@biggamesmallworld the `dashboard e2e (2)` failure is fixed in `1c09196fd` — applied your `isChecked()` suggestion verbatim (plus an `expect(Update).toBeEnabled()` right after the toggle, per your robustness note, so a future no-op fails loudly instead of timing out). CI is green now, `dashboard e2e (2)` included, and that test-only change is the only diff since your approval.

Your approval got dismissed as stale on that push — could you give it another quick look and re-approve so I can merge? Thanks again for the diagnosis.

biggamesmallworld (approved):

PR #4964, opened 2026-07-14
fix(dashboard): Derive radius scale from the --radius token

### LINKED ISSUE #4865: Dashboard theming: `--radius` token has no effect — radius scale is hardcoded to 0.2rem via `@theme inline`
**Describe the bug**

The Dashboard theming docs state that `--radius` is the *"Base border radius"* — the single token that controls all corner roundness. In practice `--radius` (and any radius override passed via `vendureDashboardPlugin({ theme })`) has **no effect** on the `rounded-*` utilities. Every corner is locked to `0.2rem` regardless of the theme.

**Root cause**

`packages/dashboard/vite/vite-plugin-theme.ts` → `generateThemeInlineBlock()` emits the radius scale into a Tailwind v4 **`@theme inline`** block, sourced from `@vendure-io/design-tokens`:

```ts
// vite-plugin-theme.ts
const radiusLines = Object.entries(radii).map(([key, value]) => `    --radius-${key}: ${value};`);
// ...
return `@theme inline {\n${allLines.join('\n')}\n}`;
```

Two compounding problems:

1. **`@theme inline` bakes literals into the utilities.** Because the block uses the `inline` keyword, Tailwind substitutes the resolved value directly into each utility instead of emitting a `var()` reference. The compiled CSS is:

   ```css
   .rounded-lg { border-radius: .2rem }   /* literal — not var(--radius-lg) */
   .rounded-md { border-radius: .2rem }
   ```

   So overriding `--radius` (or even `--radius-md`) at `:root` / `.dark` — which is exactly and only what the `theme` option does — can never change the utilities. `--radius` ends up in `:root` but nothing consumes it.

2. **In `@vendure-io/design-tokens`, every radius size is hardcoded to `0.2rem`** (`sm`/`md`/`lg`/`xl`/`2xl`/`3xl`/`4xl` are all `0.2rem` in `radii`), and `darkTheme` defines no `radius` key at all. So even the `lg`/`xl` utilities are visually identical to `sm`, and there is no single base value the scale derives from.

Net effect: the documented `--radius` token is dead with respect to corner roundness.

**To Reproduce**

1. Scaffold a dashboard and pass a radius override:
   ```ts
   vendureDashboardPlugin({
     vendureConfigPath: '...',
     theme: { light: { radius: '1rem' }, dark: { radius: '1rem' } },
   })
   ```
2. Build the dashboard and inspect the emitted CSS, or open the UI.
3. Cards/buttons/inputs still render at `0.2rem`. The compiled CSS contains `.rounded-lg{border-radius:.2rem}` and an unused `--radius:1rem` in `:root`.

**Expected behavior**

Setting `--radius` (via the `theme` option) changes the corner roundness across the dashboard, as the theming docs describe — i.e. the `rounded-*` scale derives from `--radius` (the standard shadcn ladder), e.g.:

```CSS
  --radius-sm: calc(var(--radius) * 0.6);
  --radius-md: calc(var(--radius) * 0.8);
  --radius-lg: var(--radius);
  --radius-xl: calc(var(--radius) * 1.4);
  --radius-2xl: calc(var(--radius) * 1.8);
  --radius-3xl: calc(var(--radius) * 2.2);
  --radius-4xl: calc(var(--radius) * 2.6);
```

**Actual behavior**

All corners are fixed at `0.2rem`; `--radius` has no visible effect. The radius part of the `theme` option is silently inert.

**Suggested fix**

Make the radius scale derive from `--radius` so the documented token actually drives roundness. Either:

- Emit the `--radius-*` entries as `var(--radius)` / `calc(var(--radius) ± Npx)` expressions (referencing the runtime `--radius`) instead of inlined literals, and/or move the radius tokens out of the `@theme inline` block so utilities reference the variable rather than baking in a literal; and
- Define `radii` in `@vendure-io/design-tokens` as a calc ladder off a single base radius, and give `darkTheme` a `radius` key.

**Workaround** (for anyone hitting this): add a Vite plugin *before* `vendureDashboardPlugin` that appends an overriding `@theme inline` block re-deriving `--radius-*` from `var(--radius)`. It must run before the dashboard's Tailwind plugin so Tailwind compiles it; a later `@theme` declaration wins, restoring variable-driven utilities.

**Screenshots/Videos**

n/a (CSS output included above).

- `@vendure/dashboard` version: 3.6.3 (mechanism unchanged on `main` — see `packages/dashboard/vite/vite-plugin-theme.ts`, `generateThemeInlineBlock`)
- `@vendure-io/design-tokens` version: as bundled with 3.6.3 (`radii` all `0.2rem`)
- Nodejs version: 22
- Reference: theming docs — https://docs.vendure.io/guides/extending-the-dashboard/theming/

grolmus:
## Summary

Makes the dashboard's documented `--radius` theme token actually drive corner roundness. Previously `--radius` (and any `vendureDashboardPlugin({ theme: { light/dark: { radius } } })` override) had **no effect** — every `rounded-*` utility was locked to the literal `0.2rem`. (#4865 / OSS-577)

## Root cause

`vite-plugin-theme.ts` → `generateThemeInlineBlock()` emitted the radius scale into a Tailwind v4 `@theme inline` block using the **literal** values from `@vendure-io/design-tokens` `radii` (all `0.2rem`):

```ts
const radiusLines = Object.entries(radii).map(([key, value]) => `--radius-${key}: ${value};`);
```

Because the block uses `@theme inline`, Tailwind **bakes the value** into each utility → `.rounded-lg { border-radius: .2rem }` (a literal, not `var(--radius-lg)`). So the `--radius` that the `theme` option writes into `:root` is never consumed, and overriding it can't change anything.

## Fix

Emit the `sm…4xl` scale as `var(--radius)` / `calc(var(--radius) * N)` expressions (the standard shadcn ladder) instead of literals. `@theme inline` then bakes the **var reference** into the utilities (`.rounded-lg { border-radius: var(--radius) }`), so they resolve against the runtime `--radius` and a `theme.radius` override takes effect. `none` (`0`) and `full` (`9999px`) stay literal — they're fixed semantic values, not part of the scalable ladder.

Default look is unchanged: `lightTheme.radius` is `0.2rem`, so `var(--radius)` resolves to the same `0.2rem` by default.

## Tests

`plugin-hooks.spec.ts` — the stale test that asserted the baked-literal behaviour is replaced with one asserting the scale is `var(--radius)`/`calc(...)`-based (and that `none`/`full` stay literal), plus a test that a `theme.light.radius` override lands in `:root { --radius }`. Full theme-plugin suite green (75/75), `tsc` clean.

## Out of scope / follow-up

The issue's second half lives in the **separate `@vendure-io/design-tokens` package** (not in this monorepo): its `radii` are all identical `0.2rem` and `darkTheme` has no `radius` key. Ideally that package defines `radii` as a calc ladder off a single base and gives `darkTheme` a `radius`. This PR fixes the dashboard side so `--radius` is honoured now regardless; the design-tokens defaults are a follow-up in that repo.

Fixes #4865

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4964"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786616123&installation_id=128408429&pr_number=4964&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4964&signature=916c3e2ef4529f28fd8ba5ef2f484283073a91c186eda4aab9c837342851ad1b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


biggamesmallworld:
**PR description: "Default look is unchanged" isn't accurate**

Only `rounded-lg` holds at `0.2rem`. Every other step moves, because today `radii.sm`–`radii.4xl` are all literally `0.2rem`:

`sm` 0.2→0.12 · `md` 0.2→0.16 · `lg` 0.2→0.2 · `xl` 0.2→0.28 · `2xl` 0.2→0.36 · `3xl` 0.2→0.44 · `4xl` 0.2→0.52rem

That touches ~135 `rounded-*` usages in `packages/dashboard/src` and ~78 in `@vendure-io/ui` (Button, Input, Dialog, Card).

Not a regression — #4865 names the flat scale as part of the bug and asks for exactly this ladder — but the body currently claims the opposite, and anyone with visual snapshots will go red on release.

Suggested replacement for that paragraph:

> **Visual impact:** the base radius is unchanged (`--radius` stays `0.2rem`), but the scale becomes differentiated as #4865 requires — `rounded-lg` is unchanged, `sm`/`md` tighten slightly and `xl`+ open up. This is the intended fix for the issue's second complaint ("even the `lg`/`xl` utilities are visually identical to `sm`"). Expect visual-regression snapshots covering rounded corners to need updating.

Also worth noting in the body that the multipliers are lifted from the issue's "Expected behavior" block (and match @kevmtt's workaround) — in the diff they read as magic numbers. A one-line source comment next to `radiusScale` would save the next reader the archaeology.

biggamesmallworld (approved):
I would approve this pending that @dlhck validates that it doesn't interfere with the current design changes going on with vendure design tokens. It does fix the original reported issue, but I want to double check that

dlhck:
Closing this, because it is not necessary anymore with the design system changes coming in the next minor release. border radii are tokenized there again. thanks @biggamesmallworld for checking in.

PR #4970, opened 2026-07-14
fix(core): Record numeric stockOnHand at active channel's stock location

grolmus:
## Summary

With the default `MultiChannelStockLocationStrategy`, creating a `ProductVariant` while operating in a **non-default channel** and supplying a numeric `stockOnHand` (as the Dashboard's inline create-variant table does) wrote the stock to the **global default** `StockLocation` instead of the **active channel's** `StockLocation`. The channel-filtered `ProductVariant.stockLevels` then resolved to `0` for that channel — the stock the user entered was effectively lost from the channel's perspective.

Split out from #4741 (the reporter's secondary symptom: "variant created successfully, but its stock … cannot be [seen]"). The *disabled Update button* half of #4741 is a separate, already-merged fix (#4864).

## Root cause

The create path passes a numeric `stockOnHand` to `StockMovementService.adjustProductVariantStock`, whose numeric branch always targeted the global default location:

```ts
// stock-movement.service.ts
const defaultStockLocation = await this.stockLocationService.defaultStockLocation(ctx);
```

`StockLocationService.defaultStockLocation(ctx)` returned the globally-oldest `StockLocation` (bare `.find({ order: { createdAt: 'ASC' } })`), ignoring `ctx.channelId`. So in a secondary channel the entered stock landed on another channel's location and was invisible where it was entered. `#4864` guaranteed a `StockLevel` *row* exists for the channel (fixing the disabled button) but did not route the entered *quantity* to it.

## Change

`StockLocationService.defaultStockLocation(ctx)` is now channel-aware: it prefers the oldest `StockLocation` associated with the active channel (`innerJoin` on `stockLocation.channels`), falling back to the global oldest location when the channel has none associated. Both branches order by `createdAt` then `id` for determinism. The method's only caller is the numeric branch of `adjustProductVariantStock`, and default/single-channel behaviour is unchanged (the default location is associated with the default channel).

## Test plan

**Automated**
- New regression e2e: `packages/core/e2e/stock-location.e2e-spec.ts` (`multi channel` block) — creates a variant in a secondary channel with a numeric `stockOnHand` and asserts the exact `stockLevels` entry (`stockOnHand: 25` at the channel's location). Fails without the fix (stock lands on the default location → `0` in-channel), passes with it.
- `stock-location.e2e-spec.ts` 15/15 and `product.e2e-spec.ts` 100/100 pass; `tsc` build clean.

**Manual (React Dashboard, MultiChannelStockLocationStrategy, Postgres)**
- Channel `channel-b` with its own stock location "Warehouse B". Operating in `channel-b`, created a product + variant with Stock on Hand = 50 via the inline table.
- Before fix: variant shows Stock level **0** in the channel. After fix: **50** at Warehouse B. Screenshots are attached on the Linear issue.

Relates to #4741

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4970"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786629465&installation_id=128408429&pr_number=4970&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4970&signature=6a083076c65d40ac68ec88627152725fa79ab5447951a9d4c152b1cde864a7e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
01-create-variant-stock-50-channel-b
02-before-fix-variant-stock-0-in-channel
03-after-fix-variant-stock-50-in-channel

biggamesmallworld (approved):
Tests are passing, and it reads as the test covers the case reported in the issue 🏄‍♂️  Approving

PR #4978, opened 2026-07-16
chore(dev-server): Add worktree-safe Portless workflow

dlhck:
## What changed

- replace the dev-server's fixed-port `dev all` workflow with a Portless-supervised API and Dashboard workflow
- build required packages and a clean server-served Dashboard before starting
- watch `@vendure/common`, `@vendure/core`, and both Dashboard TypeScript outputs, restarting only the process that loaded rebuilt code
- keep `/dashboard` intact while exposing the HMR Dashboard through its own worktree-aware Portless URL
- inject the matching API and Dashboard URLs through the existing environment variable names
- make the worker opt-in and protect it with a repository-wide lock in the primary checkout's `.vendure` directory
- add an agent lifecycle contract with worktree-local state, structured events, readiness checks, status waiting, and scoped shutdown
- retain `dev:direct` as a fixed-port escape hatch
- remove the discontinued Angular Admin UI from the dev-server config
- document the workflow and ignore local agent plans

## Why

Fixed API and Dashboard ports make parallel human-driven development across Git worktrees impractical. The Docker-backed database and supporting services can remain shared, while Portless gives each worktree isolated application URLs and dynamically allocated backing ports.

Agents also need a deterministic way to know when the supervised processes are ready and which URLs belong to the current worktree. The machine-readable lifecycle avoids parsing human logs and prevents one worktree from stopping another worktree's processes.

The worker is intentionally separate because all worktrees share the same database and job queue, so running multiple workers would process the same jobs concurrently.

## Developer impact

From `packages/dev-server`:

- `bun run dev` starts the human-oriented API, Dashboard, and watcher workflow
- `bun run dev:agent` starts the same workflow with lifecycle state in `.vendure/dev-server.json`
- `bun run dev:status --wait --json` waits for watcher and HTTP readiness and returns all worktree URLs
- `bun run dev:stop` stops only the current worktree's agent-managed supervisor
- `bun run dev:worker` starts the single repository worker when needed
- `bun run dev:direct` uses the traditional ports 3000 and 5173

## Validation

- `bun run test:dev-workflow` (7 tests)
- `bun install --frozen-lockfile`
- Prettier check for changed code and configuration files
- Node syntax checks for the new scripts
- full `DB=postgres bun run dev` smoke test against the existing Postgres container
- full `DB=postgres bun run dev:agent` lifecycle smoke test
- verified `building`, `starting`, `ready`, and `stopping` structured events
- verified `dev:status --wait --json` returns only after all watchers and the three HTTP readiness endpoints are healthy
- verified duplicate agent startup reports the active supervisor PID
- verified the Portless API route, HMR Dashboard route, and server-served `/dashboard` route
- verified `dev:stop` removes the status file and leaves no active Portless routes

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4978"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786780184&installation_id=128408429&pr_number=4978&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4978&signature=76c2001068911da5b24df0e8b00329b7f7424bd422c08883e3e32ff613f650c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


michaelbromley:
## Code review

Reviewed the PR diff (not the local working tree). Overall this is decent dev tooling — the extraction of pure logic into testable modules (`dev-network-config.mjs`, `dev-state.mjs`, `worker-lock.mjs`) is the right instinct and the associated specs are solid. But the biggest, riskiest file — `dev-workflow.mjs` (513 lines of process orchestration) — has zero test coverage, and that's where the bugs are.

### Must fix before merge

**1. Stale `'ready'` status during watcher-triggered restarts.** In `RestartableProcess.restart()`, when a rebuild triggers a restart, the child is killed and later respawned but nothing moves `lifecycle.status` off `'ready'` during that window. An agent polling `dev:status --wait` mid-restart sees `status: 'ready'` and starts hitting a server that's mid-restart or dead. This breaks the exact scenario the whole lifecycle-contract feature exists to solve.

**2. Two dead `localhost:4201` email URLs in `dev-config.ts`.** Port 4201 was the old Angular admin-ui dev server port. This PR removes `AdminUiPlugin`, so nothing listens on 4201 anymore. Only `passwordResetUrl` was updated to use `dashboardUrl`; `verifyEmailAddressUrl` and `changeEmailAddressUrl` still point at the dead port, so those links are broken in every dev-environment email. Fix all three or none.

**3. `portlessCliPath` resolution relies on undocumented package internals.**
```js
const portlessCliPath = path.join(path.dirname(fileURLToPath(import.meta.resolve('portless'))), 'cli.js');
```
This assumes `cli.js` sits next to whatever the `portless` main/exports entry resolves to. `bun.lock` already gives the stable contract: `"bin": { "portless": "dist/cli.js" }`. The bin field is the documented interface; the main-entry directory is an implementation detail that can move in any patch release. Resolve via `node_modules/.bin/portless`, or `require.resolve('portless/package.json')` and follow `bin.portless`.

### Should fix

**4. Duplicated locking primitive.** `worker-lock.mjs` and `dev-state.mjs` copy-paste `isProcessAlive` verbatim, and `acquireWorkerLock`/`claimDevStatus` are structurally identical (openSync `wx`, catch `EEXIST`, check liveness, reclaim-if-stale). The only genuine difference is the root resolver (`git rev-parse --git-common-dir` vs `--show-toplevel`), which is correct and deliberate. Extract a shared `acquireLock({ path, pid, processIsAlive, ... })` parameterised by the two resolvers so future staleness/retry fixes only need applying once.

**5. `dev-stop.mjs` never escalates to SIGKILL.** After the 15s SIGTERM grace period, if the process is still alive the script just logs and exits 1 — it never sends `SIGKILL`, leaving a stuck supervisor that the caller has to hunt down by PID. Escalate to `SIGKILL` after the timeout.

**6. Locale-fragile `tsc` watch-readiness detection.**
```js
if (line.includes('Found 0 errors. Watching for file changes.')) { ... }
```
`tsc`'s watch output is localized. On a non-English-locale machine this string never appears, the readiness promise never resolves, and `dev:agent` hangs indefinitely. This is also in the one new file with no `.spec.mjs`. Use a machine-readable build signal (or at minimum add a timeout with a clear error and a comment flagging the fragility).

**7. `'auto'` sentinel for `VITE_ADMIN_API_PORT` is a magic string in three files** — set in `dev-network-config.mjs`, checked in `vite.config.mts`, re-checked in the spec — with no shared constant. Since these are plain `.mjs` with no cross-module type-checking, renaming it in one place silently breaks the others. Pull it into one exported constant.

### Nitpicks / follow-up

- **Scope**: this bundles three independent behavioural changes — Portless worktree URLs, the agent lifecycle/status contract, and *the worker no longer starting by default under `bun run dev`*. That last one silently surprises every existing contributor and only got a README paragraph. Worth splitting the worker change out as its own reviewable PR.
- **Magic timing numbers** with no rationale: `200` (restart debounce), `250` (poll interval, duplicated), `15_000` (SIGTERM grace), `180_000` (HTTP wait), `300` (default status timeout), `5_000` (startup deadline). Name them if any might need tuning later.
- The `.gitignore` addition for `plans/` under "Agentic engineering" looks unrelated to Portless worktrees — confirm it isn't stray scope.

### What's good

The dependency-injected pure functions (`processIsAlive`, `ensurePortlessProxy`, `getPortlessUrl` all passed in rather than hardcoded) make the extracted modules genuinely unit-testable without spawning real processes. The specs use exact-value assertions (`assert.equal`/`assert.deepEqual`) against concrete expected values — not the weak `toBeTruthy()`/`toBeDefined()` pattern — and cover the staleness-reclaim and already-running-error paths, not just the happy path. Real test discipline.

dlhck:
@michaelbromley Thanks for the detailed review. I addressed the actionable feedback in d72b85ff6:

- Restart-triggered lifecycle state now moves back to `starting` and only returns to `ready` after the affected endpoints respond again. Concurrent and superseded restarts are coordinated, with focused tests.
- All three development email URLs now derive from `VENDURE_DASHBOARD_URL`; the remaining `localhost:4201` references are gone.
- The Portless executable is resolved from the package-declared `bin` metadata rather than assuming its internal file layout.
- Worker and dev-status ownership now use a shared atomic lock primitive. It also avoids reclaiming an incomplete lock that another process is still initializing.
- `dev:stop` escalates from SIGTERM to SIGKILL after the grace period.
- TypeScript watchers are forced to English output and now fail with a clear timeout instead of waiting indefinitely if the expected readiness signal is not emitted.
- The Dashboard API-port sentinel is now a shared exported constant.
- Worker lock conflicts now print the clean error message without a stack trace.
- Timing values that are part of the lifecycle behavior are named.

I also added orchestration coverage around lifecycle transitions, obsolete restart completions, watcher readiness timeout, Portless bin resolution, forced process termination, and lock initialization races. The focused suite now has 20 passing tests. The Lerna Dashboard dependency build and full dev-server build both pass locally.

Two scope notes:

- The worker remaining explicit is intentional. All worktrees share the database and job queue, so `bun run dev` should not silently start competing workers. The repository-wide worker lock remains the guard when worker behavior is needed.
- The `plans/` ignore is intentional and has been retained as part of the agentic-engineering setup.

michaelbromley:
## Re-review — head `d72b85f` (`fix(dev-server): Address workflow review feedback`)

Every must-fix and should-fix from the previous round is now **Fixed**, and fixed at the root rather than papered over:

- **Stale `ready` on restart** → new `RestartReadinessCoordinator` (per-label sequence tokens, handles the obsolete-restart race), with tests. A proper redesign.
- **Dead `localhost:4201` email URLs** → all three `globalTemplateVars` now derive from `dashboardUrl`.
- **`portlessCliPath`** → new `resolvePackageBin()` reads the package's declared `bin` field, with its own test.
- **Duplicated lock primitive** → both callers delegate to `claimExclusiveLock()` in `exclusive-lock.mjs`, keeping only the genuinely-different root resolvers separate.
- **SIGKILL escalation** → `process-termination.mjs` `terminateProcess()` (SIGTERM → poll → SIGKILL), tested both paths.
- **Locale-fragile `tsc` string** → `LC_ALL=C`/`LANG=C` on watchers + explicit `--locale en` + a readiness timeout so it errors instead of hanging.
- **`'auto'` magic string** → single exported `DASHBOARD_API_PORT_FROM_PAGE` constant.
- **Magic timing numbers** → all named exported constants.

Nothing here **blocks merge**. Two new/minor items from this same commit, worth a quick pass:

**A. README now self-contradicts on the default DB.** The example was changed to `DB=postgres bun run dev` and the follow-up reworded to "changed to `DB=mysql` or `DB=sqlite` when needed", but the sentence above still says "The default database is MySQL" and `getDbConfig()` still defaults to `mysql`. As written a reader can't tell which is actually the default. One-line wording fix.

**B. `exclusive-lock.mjs` has no spec of its own.** The centralized primitive's edge cases (the final `getClaimError` throw when both retries lose the race) are only covered indirectly via its two consumers. Now that the logic is centralized, its tests should be too. Non-blocking follow-up.

PR #4986, opened 2026-07-17
fix(core): Hydrate relations missing from only some array elements

### LINKED ISSUE #4537: EntityHydrator does not hydrate object trees if partially present
**Describe the bug**
The entity hydrator does not hydrate reliably. The reason is that if the it finds an object early in the tree with the requested relation, it will stop. However, VERY often in Vendure the tree is full hydrated.

E.g. an orderLine > productVariant > product.

I created an order interceptor and for a cart with 2 items, the first items had the relationship, the second item didn't. I hydrate the object to have the product relation and I get an error when I access the variable bc it wasn't hydrated. This is SO frustrating.

It should not matter what happened before hydration. Hydration should simply ensure relations being loaded no matter what. This is likely an optimization related bug. The assumption being: "Oh, we have 1 productVariant.product relation", the rest will be there as well.

**To Reproduce**

1. The easiest reproduction is writing a test with an incomplete object and hydrating it

**Expected behavior**
Hydration should guarantee relations being loaded

**Actual behavior**
It cannot be relied upon.

- Database (mysql/postgres etc): Postgres

**Additional context**
We have a heavily customized Vendure version with tons of plugins. That is likely a contributing factor, but it's not an excuse.

- When did this issue start occurring? It's been around since forever
- Does it happen consistently or intermittently? When it fails, it tends to fail always
- Any recent changes to your setup? No

ryandiginomad:
# Description

Fixes #4537.

`EntityHydrator.hydrate()` promises that the requested relations are populated, regardless of how the entity was loaded beforehand. It does not hold for array relations that are loaded unevenly.

`getMissingRelations()` walks each relation path with a single cursor, and descends into an array by sampling its first element:

```ts
entity = Array.isArray(entity[part]) ? entity[part][0] : entity[part];
```

So if `order.lines[0].productVariant` happens to be loaded, the whole `lines.productVariant` path is treated as present: nothing is reported missing, no query is issued, and `lines[1..n].productVariant` stay unloaded. As the reporter puts it, the assumption is _"Oh, we have 1 productVariant.product relation, the rest will be there as well"_ — which nothing actually guarantees, since the lines of an order need not have travelled the same loading path.

## The fix

Check **every** entity at each depth of the path rather than just the first, reporting the path as missing if any one of them lacks the relation.

This is the approach the sibling method `getRelationEntityAtPath()` in the same file already takes (`for (const item of target) { visit(item, ...) }`), after it was fixed for the same class of partial-array bug in #4661. This change brings `getMissingRelations()` in line with it.

Semantics deliberately kept unchanged:

- `null` (fetched, but null in the DB) is still terminal and is not reported missing; `undefined` (not fetched) is. The only difference is that a `null` now skips just that entity instead of abandoning the whole path, so a relation genuinely missing from a *sibling* of a `null` one is now detected.
- A loaded-but-empty array still reports deeper path segments as missing, exactly as before. Walking every element would have let us skip that (arguably wasted) query, but that is a behaviour change beyond this bug, so I left it alone and pinned it with tests.

# Breaking changes

None. Hydration now issues a query in cases where it previously (incorrectly) skipped one, so unevenly-loaded relations get populated instead of staying `undefined`. Fully-loaded arrays are unaffected and still short-circuit without a query — covered by a test.

# Testing

**Unit** (`entity-hydrator.service.spec.ts`, 8 new cases, no DB):

- 3 cover the bug itself (relation missing from a later array element, an intermediate relation missing, and a nested array-within-array).
- 2 cover the `null` semantics, including a relation missing from the sibling of a `null` one.
- 3 pin behaviour that must *not* change: a fully-loaded array still reports nothing, plus the two empty-array cases above. These pass both before and after the fix.

**e2e** (`entity-hydrator.e2e-spec.ts`, 1 new case): builds an order with two lines from different products (T_1 Laptop, T_5 Curvy Monitor), drops the `product` relation from the second line's variant to model the partially-populated tree plugin code sees, then hydrates and asserts both lines end up with their own correct product.

What I ran:

| | Result |
|---|---|
| `entity-hydrator.service.spec.ts` before the fix | 4 failed / 5 passed |
| `entity-hydrator.service.spec.ts` after | 9 passed |
| `entity-hydrator.e2e-spec.ts` before the fix | 1 failed / 24 passed (only the new case) |
| `entity-hydrator.e2e-spec.ts` after | 25 passed |
| Full core unit suite | 74 files, 1043 passed |
| Full core e2e suite | 101 files, 1947 passed, 2 failed |

The 2 e2e failures (`custom-fields > validation > invalid datetime` and `ErrorHandlerStrategy > invokes the worker handler`) are unrelated to this change and reproduce identically on a clean `master` with my changes stashed. Same for a pre-existing type error in `order.service.ts:1460` that stops `npm run build` locally.

### 🤖 AI assistance

I used Claude Code while working on this. I have reviewed every line of the change myself, and the testing table above reflects test runs I actually performed and observed — including verifying that each new test fails without the fix and passes with it.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4986"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786850217&installation_id=128408429&pr_number=4986&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4986&signature=a30b14b28201acf67eea9f6f1abe6a0f901e4ebee67dedad544de8d32311245a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
@@ -115,11 +115,103 @@ describe('EntityHydrator', () => {
             // Spreading a very large array into push() (`push(...value)`) exceeds V8's argument
             // limit and throws "RangeError: Maximum call stack size exceeded"; a plain loop must
             // be used instead. Reproduces e.g. `collections.productVariants` on a big catalog.
+            //
+            // The path must run PAST the large array: a relation at the end of the path is only
+            // checked for presence, so its elements are never collected and the spread this test
+            // guards against would not be reached.
             const collection = {
-                productVariants: Array.from({ length: 200_000 }, (_, i) => ({ id: i })),
+                productVariants: Array.from({ length: 200_000 }, (_, i) => ({
+                    id: i,
+                    product: { id: i },
+                })),
             };
-            expect(() => getMissingRelations(collection, ['productVariants'])).not.toThrow();
-            expect(getMissingRelations(collection, ['productVariants'])).toEqual([]);
+            expect(() => getMissingRelations(collection, ['productVariants.product'])).not.toThrow();
+            expect(getMissingRelations(collection, ['productVariants.product'])).toEqual([]);
+        });
+
+        // A relation at the end of the path only needs to be checked for presence, so the walk
+        // must not descend into the entities it points to. Descending is wasted work proportional
+        // to the size of the relation, e.g. `collections.productVariants` on a big catalog.
+        it('does not walk into the elements of an array at the end of the path', () => {
+            class CountingArray extends Array {
+                walked = 0;
+                [Symbol.iterator]() {
+                    this.walked++;
+                    return super[Symbol.iterator]();
+                }
+            }
+            const productVariants = new CountingArray();
+            productVariants.push({ id: 1 }, { id: 2 }, { id: 3 });
+
+            expect(getMissingRelations({ productVariants }, ['productVariants'])).toEqual([]);
+            expect(productVariants.walked).toBe(0);
+        });
+
+        // Once one element is found to be missing the relation, the whole path is already known to
+        // be missing, so there is nothing to learn from the remaining elements.
+        it('stops checking array elements once one is found to be missing', () => {
+            const visited = new Set<number>();
+            // The relation is missing from an element in the middle rather than the first one, so
+            // this also fails an implementation that only samples element [0] — issue #4537 itself.
+            const missingAt = 7;
+            const lines = Array.from({ length: 100 }, (_, i) => ({
+                get productVariant() {
+                    visited.add(i);
+                    return i === missingAt ? undefined : { id: i };
+                },
+            }));
+
+            expect(getMissingRelations({ lines }, ['lines.productVariant'])).toEqual([
+                'lines',
+                'lines.productVariant',
+            ]);
+            expect(visited.size).toBe(missingAt + 1);
+        });
+
+        // An `undefined` element settles the path on its own, so the walk must stop at it rather
+        // than carry on reading the relation off every remaining element. Distinct from the test
+        // above: that one exits on a present element whose relation is undefined, this one exits
+        // on an element that is itself an `undefined` hole (issue #4955).
+        it('stops walking the frontier at an undefined array element', () => {
+            const read = new Set<number>();
+            const lines: Array<Record<string, any> | undefined> = Array.from({ length: 100 }, (_, i) => ({
+                get productVariant() {
+                    read.add(i);
+                    return { id: i };
+                },
+            }));
+            lines[0] = undefined;
+
+            expect(getMissingRelations({ lines }, ['lines.productVariant'])).toEqual([
+                'lines',
+                'lines.productVariant',
+            ]);
+            expect(read.size).toBe(0);
+        });
+
+        // An empty array mid-path leaves nothing to check further down, which settles the rest of
+        // the path as missing, so the walk must stop there too.
+        it('stops walking the frontier at an empty array mid-path', () => {
+            const read = new Set<number>();
+            const lines: Array<Record<string, any>> = Array.from({ length: 100 }, (_, i) => ({
+                get assets() {
+                    read.add(i);
+                    return [{ asset: { id: i } }];
+                },
+            }));
+            lines[0] = {
+                get assets() {
+                    read.add(0);
+                    return [];
+                },
+            };
+
+            expect(getMissingRelations({ lines }, ['lines.assets.asset'])).toEqual([
+                'lines',
+                'lines.assets',
+                'lines.assets.asset',
+            ]);
+            expect(read.size).toBe(1);
         });
     });
 
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
@@ -98,6 +98,29 @@ describe('EntityHydrator', () => {
                 'lines.productVariant',
             ]);
         });
+
+        // https://github.com/vendurehq/vendure/issues/4955
+        // Relation arrays can contain `undefined` holes (e.g. payments/surcharges/shippingLines
+        // after OrderPlacedEvent). An `undefined` element was never fetched and must be reported
+        // missing; only `null` counts as loaded.
+        it('reports missing when the array has an undefined leading hole', () => {
+            const order = { payments: [undefined, { refunds: [{ id: 1 }] }] };
+            expect(getMissingRelations(order, ['payments.refunds'])).toEqual([
+                'payments',
+                'payments.refunds',
+            ]);
+        });
+
+        it('does not crash when a loaded relation array is very large', () => {
+            // Spreading a very large array into push() (`push(...value)`) exceeds V8's argument
+            // limit and throws "RangeError: Maximum call stack size exceeded"; a plain loop must
+            // be used instead. Reproduces e.g. `collections.productVariants` on a big catalog.
+            const collection = {
+                productVariants: Array.from({ length: 200_000 }, (_, i) => ({ id: i })),
+            };
+            expect(() => getMissingRelations(collection, ['productVariants'])).not.toThrow();
+            expect(getMissingRelations(collection, ['productVariants'])).toEqual([]);
+        });
     });
 
     describe('getRelationEntityAtPath()', () => {
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
@@ -3,6 +3,103 @@ import { describe, expect, it } from 'vitest';
 import { EntityHydrator } from './entity-hydrator.service';
 
 describe('EntityHydrator', () => {
+    describe('getMissingRelations()', () => {
+        function getMissingRelations(target: any, relations: string[]): string[] {
+            const hydrator = new EntityHydrator(undefined as any, undefined as any, undefined as any);
+            return (hydrator as any).getMissingRelations(target, { relations });
+        }
+
+        // https://github.com/vendurehq/vendure/issues/4537
+        it('detects a relation missing from a later array element', () => {
+            const order = {
+                lines: [
+                    { productVariant: { product: { id: 1 } } },
+                    { productVariant: { product: undefined } },
+                ],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.product'])).toEqual([
+                'lines',
+                'lines.productVariant',
+                'lines.productVariant.product',
+            ]);
+        });
+
+        // https://github.com/vendurehq/vendure/issues/4537
+        it('detects an intermediate relation missing from a later array element', () => {
+            const order = {
+                lines: [{ productVariant: { product: { id: 1 } } }, { productVariant: undefined }],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.product'])).toEqual([
+                'lines',
+                'lines.productVariant',
+                'lines.productVariant.product',
+            ]);
+        });
+
+        it('detects a relation missing from a nested array element', () => {
+            const order = {
+                lines: [
+                    {
+                        productVariant: {
+                            assets: [{ asset: { id: 1 } }, { asset: undefined }],
+                        },
+                    },
+                ],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.assets.asset'])).toEqual([
+                'lines',
+                'lines.productVariant',
+                'lines.productVariant.assets',
+                'lines.productVariant.assets.asset',
+            ]);
+        });
+
+        it('reports nothing when every array element has the relation', () => {
+            const order = {
+                lines: [
+                    { productVariant: { product: { id: 1 } } },
+                    { productVariant: { product: { id: 2 } } },
+                ],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.product'])).toEqual([]);
+        });
+
+        it('treats a relation that is null in the database as loaded', () => {
+            const order = {
+                lines: [{ productVariant: { product: null } }, { productVariant: { product: null } }],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.product'])).toEqual([]);
+        });
+
+        it('reports a relation missing from a sibling of a null relation', () => {
+            const order = {
+                lines: [{ productVariant: { product: null } }, { productVariant: { product: undefined } }],
+            };
+
+            expect(getMissingRelations(order, ['lines.productVariant.product'])).toEqual([
+                'lines',
+                'lines.productVariant',
+                'lines.productVariant.product',
+            ]);
+        });
+
+        it('reports nothing for a loaded empty array', () => {
+            expect(getMissingRelations({ lines: [] }, ['lines'])).toEqual([]);
+        });
+
+        it('reports deeper relations of a loaded empty array as missing', () => {
+            expect(getMissingRelations({ lines: [] }, ['lines.productVariant'])).toEqual([
+                'lines',
+                'lines.productVariant',
+            ]);
+        });
+    });
+
     describe('getRelationEntityAtPath()', () => {
         // https://github.com/vendurehq/vendure/issues/4661
         it('treats undefined intermediate relations as terminal values', () => {
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
@@ -225,33 +225,40 @@ export class EntityHydrator {
                             // path must be reported missing rather than skipped.
                             if (entity === undefined) {
                                 isMissing = true;
-                                continue;
+                                break;
                             }
                             // null = the relation has been fetched but was null in the database.
                             if (entity === null || entity[part] === null) {
                                 continue;
                             }
-                            if (entity[part]) {
-                                const value = entity[part];
-                                if (Array.isArray(value)) {
-                                    if (value.length === 0 && path.length < parts.length) {
+                            const value = entity[part];
+                            if (!value) {
+                                isMissing = true;
+                                break;
+                            }
+                            // At the last segment of the path we only need to know whether the
+                            // relation is present; the entities it points to are never inspected,
+                            // so there is no point collecting them.
+                            const isLastPart = path.length === parts.length;
+                            if (Array.isArray(value)) {
+                                if (value.length === 0) {
+                                    if (!isLastPart) {
                                         // An empty array leaves nothing to check further down the
                                         // path, so treat the rest of the path as missing.
                                         isMissing = true;
-                                    } else {
-                                        // Use a plain loop rather than push(...value): spreading a
-                                        // very large array (e.g. collections.productVariants on a
-                                        // big catalog) exceeds V8's argument limit and throws
-                                        // RangeError, even when everything is already loaded.
-                                        for (const element of value) {
-                                            nextEntities.push(element);
-                                        }
+                                        break;
                                     }
-                                } else {
-                                    nextEntities.push(value);
+                                } else if (!isLastPart) {
+                                    // Use a plain loop rather than push(...value): spreading a
+                                    // very large array (e.g. collections.productVariants on a
+                                    // big catalog) exceeds V8's argument limit and throws
+                                    // RangeError, even when everything is already loaded.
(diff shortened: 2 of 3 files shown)


biggamesmallworld (approved):
This fix looks solid for the case mentioned in the gh issue. Ran the tests locally myself and the suite is passing: 

Image

@michaelbromley as this touches entity hydrator which is important core functionality, can you give this a once over please? 

Approving in the meantime as the fix looks correct to me.

biggamesmallworld:
One other thing @ryandiginomad, I think we can actually improve the test by making it more similar to the reported GH issue. Summary here:

The e2e test builds the unevenly-loaded array with `delete (order!.lines[1].productVariant as any).product`, so it verifies the fix against a hand-constructed tree rather than one any code path produces. It would keep passing even if nothing ever generated that shape.

The shape is reachable through the public API. Core's own line-append path yields `[missing, present]` (a newly added line's variant carries `product` and is pushed last), which `master` already handles because it samples `lines[0]`. The `[present, missing]` ordering comes from a plugin hydrating a *single* element in place — the `OrderInterceptor` pattern documented in `order-interceptor.ts:168`, and what the reporter described in #4537.

Suggested replacement:

```ts
it('hydrates lines after a plugin hydrated one line\'s variant', async () => {
    await shopClient.asAnonymousUser();
    await shopClient.query(addItemToOrderDocument, { productVariantId: 'T_1', quantity: 1 });
    const { addItemToOrder } = await shopClient.query(addItemToOrderDocument, {
        productVariantId: 'T_5',
        quantity: 1,
    });
    orderResultGuard.assertSuccess(addItemToOrder);

    const internalOrderId = +addItemToOrder.id.replace(/^\D+/g, '');
    const ctx = await server.app.get(RequestContextService).create({ apiType: 'admin' });
    const hydrator = server.app.get(EntityHydrator);
    const order = await server.app
        .get(OrderService)
        .findOne(ctx, internalOrderId, ['lines.productVariant']);

    expect(order!.lines[0].productVariant.product).toBeUndefined();
    expect(order!.lines[1].productVariant.product).toBeUndefined();

    // A plugin acts on one line and hydrates just that line's variant.
    await hydrator.hydrate(ctx, order!.lines[0].productVariant, { relations: ['product'] });

    // The array is now unevenly loaded: [present, missing].
    expect(order!.lines[0].productVariant.product).toBeDefined();
    expect(order!.lines[1].productVariant.product).toBeUndefined();

    await hydrator.hydrate(ctx, order!, { relations: ['lines.productVariant.product'] });

    expect(order!.lines[0].productVariant.product).toBeDefined();
    expect(order!.lines[1].productVariant.product).toBeDefined();
    expect(order!.lines[0].productVariant.product.id).toBe(order!.lines[0].productVariant.productId);
    expect(order!.lines[1].productVariant.product.id).toBe(order!.lines[1].productVariant.productId);
});
```

Differences: the uneven state is produced by a real `hydrate()` call rather than by editing the entity, the two intermediate assertions pin that the shape actually arose, and `productVariant.productId` replaces the hardcoded `toBe(1)`/`toBe(2)` so the test isn't coupled to fixture CSV row order or the id strategy.

Verified locally: fails on `master`, passes on this branch, full `entity-hydrator.e2e-spec.ts` 26/26.

ryandiginomad:
@biggamesmallworld Appreciate you writing out the full replacement and verifying it — adopted it in 236ecba96.

The `delete`-based original built a [present, missing] shape no code path produces, so it would have kept passing even if the fix regressed. Driving the uneven state through a real hydrate() on a single line's variant is a genuine regression guard, and comparing each product.id to that line's own productId drops the coupling to fixture row order / id strategy.

Verified locally (sqljs): on master the new test fails at `expect(order.lines[1].productVariant.product).toBeDefined()` — after the whole-order hydrate, the second line's product stays undefined because the walker only samples lines[0] — and it passes on this branch, with entity-hydrator.e2e-spec.ts green.

biggamesmallworld (approved):

michaelbromley (commented):
Approach is right and the tests are good. Two problems though:

**1. Crash on large arrays.** `nextEntities.push(...value)` throws `RangeError: Maximum call stack size exceeded` when the array exceeds V8's argument limit (~125k elements). Hits e.g. `collections.productVariants` on a big catalog, and fires even when everything is already loaded, since the check now walks the whole array. Use a plain loop instead of the spread.

**2. `undefined` array holes are now treated as loaded.** Relation arrays can contain `undefined` holes (#4955 — `payments`/`surcharges`/`shippingLines` after `OrderPlacedEvent`). The `!entity` guard skips them, so nothing is reported missing and the hole survives hydration; on master a leading hole triggered a refetch that repaired it (pinned in `merge-deep.spec.ts`). This test passes on master and fails on this branch:

```ts
it('reports missing when the array has an undefined leading hole', () => {
    const order = { payments: [undefined, { refunds: [{ id: 1 }] }] };
    expect(getMissingRelations(order, ['payments.refunds'])).toEqual([
        'payments',
        'payments.refunds',
    ]);
});
```

Fix: treat an `undefined` element as missing, `null` as loaded:

```ts
if (entity === undefined) {
    isMissing = true;
    continue;
}
if (entity === null || entity[part] === null) {
    continue;
}
```

Separate follow-up (not this PR): `isTranslatable()` and the price applicator still sample element `[0]`, so e.g. a null `variants[0].featuredAsset` suppresses translation of the newly-hydrated `variants[1].featuredAsset`.

ryandiginomad:
@michaelbromley Both fixed in e98adcd2b — thanks for catching these, they're real regressions the frontier walk introduced.

**1. Large-array crash.** Swapped `push(...value)` for a plain loop, so walking a large loaded array (e.g. `collections.productVariants`) no longer spreads past V8's argument limit.

**2. Undefined holes.** The walker now treats an `undefined` element as missing (never fetched) and only `null` as loaded, so a hole like `payments: [undefined, {...}]` is reported and repaired instead of surviving hydration.

Both pinned as unit tests in `getMissingRelations()` — your undefined-hole case (red on the branch before the fix, green after) and a 200k-element array for the crash. Full `entity-hydrator` unit (11) + e2e (25) green, `merge-deep.spec` unchanged.

Left the `isTranslatable()` / price-applicator `[0]` sampling for a separate PR as you suggested.

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
```diff
@@ -208,18 +208,55 @@ export class EntityHydrator {
...
-                        break;
+                    if (!isMissing) {
+                        const nextEntities: Array<Record<string, any>> = [];
+                        for (const entity of entities) {
+                            // undefined = this array element (or relation) was never fetched,
+                            // e.g. an `undefined` hole in a relation array, so the rest of the
+                            // path must be reported missing rather than skipped.
+                            if (entity === undefined) {
+                                isMissing = true;
+                                continue;
```
biggamesmallworld: Answer for this path is known once `isMissing` is set; remaining elements can't change it. `continue` keeps walking them: ~1.9ms wasted per call on a 100k-element relation.

```suggestion
                                break;
```

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
```diff
@@ -208,18 +208,55 @@ export class EntityHydrator {
...
+                                        for (const element of value) {
+                                            nextEntities.push(element);
+                                        }
+                                    }
+                                } else {
+                                    nextEntities.push(value);
+                                }
+                            } else {
+                                isMissing = true;
+                            }
```
biggamesmallworld: Last segment of a path only needs a presence check. The entities it points to are never read, so building `nextEntities` there is wasted work proportional to the size of the relation.

`relations: ['productVariants']` over a loaded 100k array: **0.410ms -> 0.000ms**. Transient heap per call on the deep 100k fixture: **9.4MB -> 5.2MB**. Also breaks early on not-found.

```suggestion
                            const value = entity[part];
                            if (!value) {
                                isMissing = true;
                                break;
                            }
                            // At the last segment of the path we only need to know whether the
                            // relation is present; the entities it points to are never inspected,
                            // so there is no point collecting them.
                            const isLastPart = path.length === parts.length;
                            if (Array.isArray(value)) {
                                if (value.length === 0) {
                                    if (!isLastPart) {
                                        // An empty array leaves nothing to check further down the
                                        // path, so treat the rest of the path as missing.
                                        isMissing = true;
                                        break;
                                    }
                                } else if (!isLastPart) {
                                    // Use a plain loop rather than push(...value): spreading a
                                    // very large array (e.g. collections.productVariants on a
                                    // big catalog) exceeds V8's argument limit and throws
                                    // RangeError, even when everything is already loaded.
                                    for (const element of value) {
                                        nextEntities.push(element);
                                    }
                                }
                            } else if (!isLastPart) {
                                nextEntities.push(value);
                            }
```

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
```diff
@@ -3,6 +3,126 @@ import { describe, expect, it } from 'vitest';
...
+        it('does not crash when a loaded relation array is very large', () => {
+            // Spreading a very large array into push() (`push(...value)`) exceeds V8's argument
+            // limit and throws "RangeError: Maximum call stack size exceeded"; a plain loop must
+            // be used instead. Reproduces e.g. `collections.productVariants` on a big catalog.
+            const collection = {
+                productVariants: Array.from({ length: 200_000 }, (_, i) => ({ id: i })),
+            };
+            expect(() => getMissingRelations(collection, ['productVariants'])).not.toThrow();
+            expect(getMissingRelations(collection, ['productVariants'])).toEqual([]);
+        });
```
biggamesmallworld, line 215: Two regression tests for the walk's cost, matching the style of the ones above. Deterministic: they count what the walker touches rather than timing it, so nothing to tune.

Red on this branch as it stands, green with the two suggestions above.

```
npx vitest run --config vitest.config.mts src/service/helpers/entity-hydrator/
```

```suggestion
        });

        // A relation at the end of the path only needs to be checked for presence, so the walk
        // must not descend into the entities it points to. Descending is wasted work proportional
        // to the size of the relation, e.g. `collections.productVariants` on a big catalog.
        it('does not walk into the elements of an array at the end of the path', () => {
            class CountingArray extends Array {
                walked = 0;
                [Symbol.iterator]() {
                    this.walked++;
                    return super[Symbol.iterator]();
                }
            }
            const productVariants = new CountingArray();
            productVariants.push({ id: 1 }, { id: 2 }, { id: 3 });

            expect(getMissingRelations({ productVariants }, ['productVariants'])).toEqual([]);
            expect(productVariants.walked).toBe(0);
        });

        // Once one element is found to be missing the relation, the whole path is already known to
        // be missing, so there is nothing to learn from the remaining elements.
        it('stops checking array elements once one is found to be missing', () => {
            const visited = new Set<number>();
            const lines = Array.from({ length: 100 }, (_, i) => ({
                get productVariant() {
                    visited.add(i);
                    // The relation is missing from the very first element.
                    return i === 0 ? undefined : { id: i };
                },
            }));

            expect(getMissingRelations({ lines }, ['lines.productVariant'])).toEqual([
                'lines',
                'lines.productVariant',
            ]);
            expect(visited.size).toBe(1);
        });
```

biggamesmallworld (commented):
Both of Michael's points check out. Verified independently: the reworked e2e test is a real guard (fails on master at `expect(order!.lines[1].productVariant.product).toBeDefined()`, passes here), `entity-hydrator.e2e-spec.ts` 25/25, full core e2e 1943 passed / 0 failed.

One thing left. The frontier walk is O(entities) and allocates: a `hydrate()` that finds **nothing** missing now walks and materialises the whole relation graph. Master did three property reads.

Two suggestions below fix the avoidable parts (`continue` should be `break`; the last path segment builds an array nothing reads). A third adds regression tests for both.

Median ms per `getMissingRelations()` call, fully loaded. Real service file swapped per variant, 5 fresh node processes x 15 timed reps, medians of per-process medians.

| fixture | master | this PR | with suggestions |
|---|---|---|---|
| order, 5 lines | 0.001 | 0.001 | 0.001 |
| order, 5 lines, uneven | 0.001 :x: | 0.002 | 0.002 |
| `productVariants` x1k | 0.000 | 0.019 | **0.000** |
| `productVariants` x10k | 0.000 | 0.186 | **0.000** |
| `productVariants` x100k | 0.000 | 0.410 | **0.000** |
| `lines.productVariant.product` x10k | 0.000 | 0.659 | 0.524 |
| same, x100k | 0.000 | 9.793 | 7.399 |
| 5 relation paths x100k | 0.001 | 13.615 | 9.921 |
| missing on element 0 of 100k | 0.001 | 1.918 | **0.392** |
| missing on element 99,999 of 100k | 0.000 :x: | 1.898 | 1.291 |
| **transient heap, 1 call, deep x100k** | 0.0MB | **9.4MB** | **5.2MB** |

:x: = wrong result. Master is the pre-fix floor, not an option. Heap is a forced-GC `heapUsed` delta around a single call, 5 runs, spread <=0.4MB.

**When it matters.** Same 3-segment path by array size:

| elements | this PR | with suggestions |
|---|---|---|
| 10 | 0.0015 | 0.0014 |
| 100 | 0.0106 | 0.0093 |
| 1,000 | 0.036 | 0.023 |
| 10,000 | 0.303 | 0.212 |
| 50,000 | 1.655 | 1.181 |
| 100,000 | 3.548 | 2.497 |

Below ~1k elements it's all under 0.04ms and irrelevant next to the query `hydrate()` would issue anyway. Real at 10k, clear by 50k, which in practice means `Collection.productVariants` and `ProductVariant.stockMovements`, not `Order.lines`. Single-segment paths are the exception with no threshold to argue about: suggestion 2 makes them free at any size instead of linear.

**Correctness:**

| | master | this PR | with suggestions |
|---|---|---|---|
| existing unit (hydrator + merge-deep) | - | 16/16 | 16/16 |
| with the 2 new tests below | - | **2 failed** | 18/18 |
| `entity-hydrator.e2e-spec.ts` | **1 failed / 24** | 25/25 | 25/25 |
| full core e2e (101 files) | - | - | **1943 passed, 0 failed** |

Apply order: suggestion 3 first (tests go red), then 1 and 2 (green). All three are one-click.

Two unrelated things from your notes: `order-multi-vendor.e2e-spec.ts` fails here too (`EntityMetadataNotFoundError: No metadata for "Zone"`, no Redis on 6379) on your branch unmodified, so it's environmental. And core built clean here, so `order.service.ts:1460` doesn't look like it's on the branch.

biggamesmallworld:
On the `isTranslatable()` / price-applicator follow-up you and Michael agreed to split out - one note for whoever picks it up.

The `undefined`-hole change in this PR widens who can hit it. If element `[0]` of a relation array is a hole or `null`, `getRelationEntityAtPath()` pushes it into the result, and the `entity[0] instanceof ProductVariant` sample then skips pricing (and `isTranslatable()` skips translation) for the **entire** array, not just that element. Same root cause, more reachable now.

Nothing to do here, just worth carrying over.

biggamesmallworld:
@ryandiginomad thanks again for following up here. As this is an inherently good fix for accuracy, but touches a piece of functionality that is widely used across the Vendure community, we wanted to be sure that we are as performant as possible to reduce the impact this may have when people begin using the updated method. Let me know if the comments above make sense, I've tried to benchmark them against some edge cases I'd found

ryandiginomad:
@biggamesmallworld All three applied in `1413db4d2`, along with three test fixes the suggestions turned up. Thanks for benchmarking this properly instead of asserting it — the per-process-median methodology, the heap deltas, and marking your own master column ✗ where the result is wrong is what made it straightforward to act on.

**On the suggestions themselves.** I checked that the walk changes are output-preserving rather than assuming it: both versions transcribed side by side and run over 300k randomised targets — arrays with `undefined` holes, `null` elements, empty arrays, arrays of arrays, single- and multi-segment paths, duplicate relations — plus a targeted set aimed specifically at the last-segment skip. Zero divergence in the returned array, compared both as an ordered list and as a set. The structural reason is that `isMissing` is monotone per relation string, so once it is set the `if (!isMissing)` guard makes `entities` dead; and at the last segment there is no next iteration to consume `nextEntities` at all.

The one input class where the two versions can differ is a relation property backed by a non-idempotent getter, since the old code evaluated `entity[part]` three times and the new one twice. Not reachable here: no `lazy: true` relations in core, entities are not Proxy-wrapped, and all 37 entity getters are `@Calculated()` value accessors rather than relations. Where it would differ, the new version's set of property reads is a subset of the old one's, so it cannot throw where the old one didn't.

**Three test gaps, all fixed in the same commit.**

1. Suggestion 2 silently neutered the RangeError regression test from the previous round. `entity-hydrator.service.spec.ts:114` used the single-segment path `['productVariants']`, so with `isLastPart` in place the collection loop is never entered on that input, and `push(...value)` could be restored with the whole suite green. The path now runs past the array (`['productVariants.product']`), which puts the guard back under test.

2. Neither of the new `break`s was pinned. Reverting suggestion 1's `entity === undefined` break, or the empty-array break, left the suite green. The `stops checking array elements once one is found to be missing` test exits through suggestion 2's `if (!value)` break, so it reaches neither. Added a counting test for each, in the style of the two you supplied.

3. Widened your second test. The missing relation sat on element `[0]`, which an implementation that samples only `[0]` — issue #4537 itself — also passes. It is now on element `[7]`, asserting `visited.size === 8`.

Mutation results, measured against both test sets:

| reverted edit | your tests | with the three fixes |
|---|---|---|
| suggestion 1: undefined-hole `break` → `continue` | survived | killed |
| suggestion 2a: falsy-value `break` → `continue` | killed | killed |
| suggestion 2b: empty-array `break` → `continue` | survived | killed |
| suggestion 2c: remove the `isLastPart` skip | killed | killed |
| round A: reintroduce `push(...value)` | survived | killed |
| round A: drop the `undefined`-hole guard | killed | killed |
| original #4537: sample element `[0]` only | killed | killed |

Verification: entity-hydrator unit 20/20, full core unit 1049 passed across 74 files, `entity-hydrator.e2e-spec.ts` 25/25, eslint and prettier clean.

On `order.service.ts:1460` — it does reproduce here on this branch. `tsc -p build/tsconfig.build.json` fails with `TS2322: CouponRemovedDuringCheckoutError is not assignable to ErrorResultUnion<AddPaymentToOrderResult, Order>`. It is unrelated to this PR (the file is byte-identical to master on this branch) and `tsc` still emits, so the build completes as far as the e2e run needs. Most likely your base is newer than mine — happy to rebase if that helps.

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
ryandiginomad: Applied. Also added a counting test for this one specifically — reverting it on its own left the whole suite green, so nothing was pinning it. Full write-up in [the summary comment](https://github.com/vendurehq/vendure/pull/4986#issuecomment-5097926804).

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.ts
ryandiginomad: Applied. This one also turned out to neuter the RangeError regression test from the previous round: that test used the single-segment path `['productVariants']`, so with the skip in place the collection loop is never entered and `push(...value)` could be restored with the suite still green. Fixed by running the path past the array. Full write-up in [the summary comment](https://github.com/vendurehq/vendure/pull/4986#issuecomment-5097926804).

packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
ryandiginomad: Both in. I widened the second one to put the missing relation on element `[7]` rather than `[0]`, so it now also fails an implementation that samples only the first element — issue #4537 itself. Full write-up in [the summary comment](https://github.com/vendurehq/vendure/pull/4986#issuecomment-5097926804).

biggamesmallworld:
@ryandiginomad thanks for the comprehensive tests and follow up! Happy to approve this now under the tests we've run.

biggamesmallworld (approved):

michaelbromley (approved):

PR #4988, opened 2026-07-17
fix(dashboard): Preserve numeric-looking values in string list inputs

### LINKED ISSUE #4987: Collection filter with arg of type string and list does not add number values
**Describe the bug**
When creating a custom collection filter with `type`: `string` and `list`: `true`, it does not add number values to the array of values.

**To Reproduce**

1. Create a custom collection filter with `type`: `'string'`, and `list`: `true` like below:
```ts
catalogOptions: {
    collectionFilters: [
        new CollectionFilter({
            code: 'external-category-ids-filter',
            description: [{ languageCode: LanguageCode.en, value: 'Filter by external category IDs' }],
            apply: (qb, args) => {
                return qb;
            },
            args: {
                externalCategoryIds: {
                    type: 'string',
                    list: true,
                },
            },
        }),
    ],
},
```
2. Create a new collection in the admin dashboard (or use an existing one)
3. Add the filter to the collection, and try to enter some numbers

**Actual behavior**
Numbers do not get added to the array of values for that collection filter

**Screenshots/Videos**

https://github.com/user-attachments/assets/3693b051-7cd4-4d76-9cff-d1480d04b401

- @vendure/core version: latest
- Nodejs version: 24.13.0
- Database (mysql/postgres etc): mysql 8 + postgres 12
- Browser (if applicable): Librewolf
- Package manager (npm/yarn/pnpm): bun

**Configuration**

Custom collection filter:
```ts
catalogOptions: {
    collectionFilters: [
        new CollectionFilter({
            code: 'external-category-ids-filter',
            description: [{ languageCode: LanguageCode.en, value: 'Filter by external category IDs' }],
            apply: (qb, args) => {
                return qb;
            },
            args: {
                externalCategoryIds: {
                    type: 'string',
                    list: true,
                },
            },
        }),
    ],
},
```

**Minimal reproduction**
See steps to reproduce

**Workaround**
Don't use `list: true`, and split on comma or semicolon. Or add a string value like `foo`, then add some numbers, when done, remove `foo`.

~I already created a fix for this, PR will follow soon~ PR created: #4988

kwerie:
# Description

String list args (type 'string', list: true) were routed to ConfigurableOperationListInput, whose parseArrayValue uses JSON.parse to determine array shape. A numeric-looking string like "3249" is itself valid JSON, so it was unwrapped from the array and passed to StringListInput as a scalar, collapsing its internal items to []. The value never rendered and the next addition overwrote it.

Route string lists to the self-contained StringListInput directly via a new selectListInputComponent helper, in both native and json-string value modes.

Fixes #4987

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4988"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786865323&installation_id=128408429&pr_number=4988&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4988&signature=991b8449fe453d6cfc4c2a03e12ce1249cc742b5617e7450fecfb9e618a208a7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


biggamesmallworld:
Confirmed the fix resolves #4987 as reported. I traced the value flow and the diagnosis holds, including why the reporter's "add `foo` first" workaround works (`JSON.parse("foo,3249")` throws, `.includes(',')` on an array is false, so the whole array falls through as a single item and happens to render). The routing change is minimal and faithful: `string` + `json-string` is the only combination whose behaviour changes.

One request on the test. `list-input-selection.spec.ts` asserts the four `if` statements it just introduced — if the routing table were conceptually wrong, the test would be wrong in the same direction and still pass. More to the point, the bug in #4987 was never in the routing table; it was in the value flow, so this test would have passed both before and after the bug was introduced.

Since this is a frontend behaviour change, a Playwright test is the right fit. The infrastructure is already there — `e2e/tests/catalog/collections.spec.ts` for the suite, and `e2e/fixtures/e2e-shared-config.ts` already exports fixtures (`e2eCustomFields`, `e2ePaymentMethodHandlers`) that `global-setup.ts` merges into the server config, so adding a filter with a string list arg is about ten lines:

```ts
new CollectionFilter({
    code: 'string-list-test-filter',
    description: [{ languageCode: LanguageCode.en, value: 'Filter by external IDs' }],
    args: { externalIds: { type: 'string', list: true } },
    apply: qb => qb,
})
```

Then the test drives the two symptoms from the issue video directly: type `3249`, press Enter, assert the badge renders; add `5`, assert both are present rather than the first being overwritten; save and reload, assert both survive. That fails on the current `master` for the right reasons, and it keeps working if someone later solves this in `parseArrayValue` instead — which a routing unit test wouldn't.

Happy for the unit test to be dropped in favour of that.

biggamesmallworld:
One more thing before I approve: the same bug is still live for `int`/`float` lists.

`parseArrayValue` ([configurable-operation-list-input.tsx:42](https://github.com/vendurehq/vendure/blob/master/packages/dashboard/src/lib/components/data-input/configurable-operation-list-input.tsx#L42)) receives a value that `FormControlAdapter` has *already* parsed, so it's an array, not a string. `JSON.parse([1])` → `1` → not an array → re-wrapped as `['1']` → re-enters the adapter as a scalar and collapses, exactly as it did for strings. `[1, 2]` survives only because `JSON.parse("1,2")` throws and the fallback happens to pass the array through — same numeric-vs-not coin flip you just fixed.

The PR currently has a test asserting int lists still route to `configurable-operation`, so this is locked in rather than just latent. Teaching `parseArrayValue` to return an already-parsed array as-is is roughly three lines and kills the whole class, including the string case.

If you'd rather keep the scope tight, that's fair — just note it in the PR description so it doesn't get lost. Either way I'm happy to approve once this and the test are addressed.

kwerie:
@biggamesmallworld thanks for the thorough review — both points addressed in c6b807d:

- **Test:** dropped `list-input-selection.spec.ts` and added a Playwright test in `catalog/collections.spec.ts` that drives the symptoms through a real `string` + `list` collection-filter arg — type `3249`, add `5`, save, reload, assert both persist as `["3249","5"]`. It fails on `master` (the `3249` badge is dropped) and passes on the branch, and asserts behaviour so it survives the fix later moving into `parseArrayValue`.
- **int/float:** fixed at the root — `parseArrayValue` received the already-parsed array from `FormControlAdapter` and re-serialized it; it now returns the array as-is, so the numeric-collapse is gone for the non-string lists too.

One separate limitation I noticed but am **not** addressing in this PR: `ConfigurableOperationListInput` has no add-item control (it only maps existing items, and each shares one `onChange`), so int/float lists still can't be built up or edited through that component. That reads as a pre-existing gap independent of #4987 — happy to open a follow-up for it.

biggamesmallworld (approved):
Thanks @kwerie !

PR #4998, opened 2026-07-17
fix(dashboard): Refresh customer history after updating customer

### LINKED ISSUE #4997: Dashboard: customer history timeline not refreshed after updating the customer
**Describe the bug**
On the dashboard customer detail page, the history timeline does not update after saving changes to the customer. New history entries (e.g. `CUSTOMER_DETAIL_UPDATED`) only appear after a full page reload.

**Cause**
The customer history is fetched by `useCustomerHistory` under the query key `['CustomerHistory', customerId]` (infinite query). The update mutation's `onSuccess` in `packages/dashboard/src/app/routes/_authenticated/_customers/customers_.$id.tsx` only shows a toast and resets the form — it never invalidates that query:

```ts
onSuccess: async data => {
    if (data.__typename === 'Customer') {
        toast.success(...)
        resetForm()
        ...
    }
}
```

**Expected behavior**
After a successful update, the timeline should reflect the new history entries, e.g. via `queryClient.invalidateQueries({ queryKey: ['CustomerHistory', params.id] })` in `onSuccess`. Since history entries written by event subscribers land slightly after the mutation response, a short delayed second invalidation (or a refetch on window focus) would make this robust for plugin-generated entries too.

**Environment**
- @vendure/dashboard 3.6.2 (verified unchanged on current master)

oliverstreissi:
# Description

On the customer detail page, the history timeline was not refreshed after saving changes to the customer — new entries (e.g. `CUSTOMER_DETAIL_UPDATED`) only appeared after a full page reload. The update mutation's `onSuccess` now invalidates the `['CustomerHistory', id]` query so the timeline refetches immediately.

Fixes #4997

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4998"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786884034&installation_id=128408429&pr_number=4998&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4998&signature=0a9f08be770ee04b941ec46a303b5767f93544b4f9b0bc8355f93a19f2b9c2d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/customers/customers.spec.ts
+++ b/packages/dashboard/e2e/tests/customers/customers.spec.ts
@@ -1,4 +1,7 @@
+import { expect, test } from '@playwright/test';
+
 import { createCrudTestSuite } from '../../utils/crud-test-factory.js';
+import { VendureAdminClient } from '../../utils/vendure-admin-client.js';
 
 createCrudTestSuite({
     entityName: 'customer',
@@ -14,4 +17,38 @@ createCrudTestSuite({
     ],
     searchTerm: 'TestCustomer',
     updateFields: [{ label: 'Last name', value: 'TestCustomerUpdated' }],
+});
+
+// #4997 — the history timeline must refresh after updating the customer,
+// without requiring a full page reload
+test('should show new history entries after updating the customer', async ({ page }) => {
+    const client = new VendureAdminClient(page);
+    await client.login();
+    const result = await client.gql(
+        `mutation CreateCustomerForHistoryTest($input: CreateCustomerInput!) {
+            createCustomer(input: $input) {
+                ... on Customer { id }
+                ... on ErrorResult { errorCode message }
+            }
+        }`,
+        {
+            input: {
+                firstName: 'History',
+                lastName: 'RefreshTest',
+                emailAddress: `history-refresh-test-${Date.now()}@example.com`,
+            },
+        },
+    );
+    const customerId = result.createCustomer.id;
+    expect(customerId).toBeTruthy();
+
+    await page.goto(`/customers/${customerId}`);
+    await expect(page.getByRole('heading', { name: 'History RefreshTest' })).toBeVisible();
+    await expect(page.getByText('Customer details updated')).toHaveCount(0);
+
+    await page.getByLabel('Last name').fill('RefreshTestUpdated');
+    await page.getByRole('button', { name: 'Update' }).click();
+    await expect(page.getByText('Successfully updated customer')).toBeVisible();
+
+    await expect(page.getByText('Customer details updated').first()).toBeVisible();
 });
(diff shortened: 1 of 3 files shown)


michaelbromley (changes requested):
Thanks for the fix — the diagnosis is right and the e2e test is a good one (asserting the "Customer details updated" entry is absent before the update proves the timeline actually refreshes, rather than the entry having been there all along).

One change I'd like before merging: the invalidation key should be derived from a shared helper rather than hand-written as a literal.

Right now the key is spelled out by hand in two separate places:

```ts
// customers_.$id.tsx (this PR)
await queryClient.invalidateQueries({ queryKey: ['CustomerHistory', data.id] });

// use-customer-history.ts:85 (the producer)
queryKey: ['CustomerHistory', customerId],
```

The invalidation only works because both files happen to spell `'CustomerHistory'` identically and order the tuple the same way. Nothing enforces that. If the key in the hook ever changes (a rename, or a third tuple element like a page size or channel token), the invalidation here silently stops matching and this exact bug comes back with no type error to catch it.

The order-history code already solves this — it exports a factory that both the producer and the consumer call:

```ts
// use-order-history.ts:58
export function orderHistoryQueryKey(orderId: string): QueryKey {
    return ['OrderHistory', orderId];
}
```

```ts
// use-order-history.ts:90 (producer)
queryKey: orderHistoryQueryKey(orderId),
// order-detail-shared.tsx:115 (consumer)
queryClient.refetchQueries({ queryKey: orderHistoryQueryKey(entity.id) }),
```

Could you mirror that for customers? Two small edits:

1. In `use-customer-history.ts`, add and use a factory:

```ts
export function customerHistoryQueryKey(customerId: string): QueryKey {
    return ['CustomerHistory', customerId];
}
// ...
queryKey: customerHistoryQueryKey(customerId),
```

2. In `customers_.$id.tsx`, import and call it instead of the literal:

```ts
await queryClient.invalidateQueries({ queryKey: customerHistoryQueryKey(data.id) });
```

That gives the key a single source of truth, matches the existing order-history convention, and makes the coupling type-checked rather than string-matched.

Minor, while you're in the file: `use-customer-history.ts:67` has a copy-paste comment `// Fetch order history` inside the customer hook — worth correcting to "customer history" since you're editing that file anyway.

michaelbromley (approved):

PR #5005, opened 2026-07-20
chore(repo): Resolve master to minor back-merge

grolmus:
Resolves the conflicted back-merge reported in #4996, following the same approach as #4982.

## Conflicts and how they were resolved

**27 × `packages/dashboard/src/i18n/locales/*.po`** — took the `minor` side, then re-ran `lingui extract` against the merged tree. The `master` side was pure extract churn: shifted line-number references plus one new `msgid` (`"The default currency is chosen from this list."`) with an empty `msgstr`. No translations are lost, and that new string is present in the regenerated catalogs.

**`packages/dashboard/src/lib/components/shared/multi-select.tsx`** — took the `minor` side.

This is the only semantic conflict. `master` carries #4198, which fixed required-field validation on the channel create form by making the popover controlled (`const [open, setOpen] = useState(false)`) and calling `setOpen(false)` after a single-select pick. `minor` carries the Design System v2 migration (#4957), which replaced that hand-rolled `Popover` entirely — all four branches of the component now delegate to `Combobox`, `UiMultiSelect` and `Select`.

The #4198 fix is therefore structurally obsolete rather than dropped: the code it patched no longer exists, and the behaviour it introduced is now native. Base UI's combobox closes on item press for a single select and keeps a multi-select open — `AriaCombobox.js`:

```js
if (multiple) {
  // ... toggle selection, popup stays open
} else {
  setSelectedValue(itemValue, eventDetails);
  setOpen(false, eventDetails);
}
```

Same semantics as #4198. The non-search single-select path uses `Select`, which closes on selection natively.

## Verification

Run against a clean `bun install --frozen-lockfile` with `build:core-common` (`@vendure-io/ui` 2.0.0-beta.15):

| Check | Result |
|---|---|
| `@vendure/dashboard` `check-types` | 0 errors |
| `@vendure/core` unit tests | 1108 passed (76 files) |
| `@vendure/dashboard` unit tests | 585 passed, 4 skipped (43 files) |

Closes #4996.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5005"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787134995&installation_id=128408429&pr_number=5005&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5005&signature=ba71d9fa73d00b96848c99554756686ec8c0aac34d97f4370562d7993596f4dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


grolmus:
Some notes for review, plus two things that need a decision rather than a review.

## ⚠️ Please merge this with **Create a merge commit**, not squash

This is the highest-consequence detail on the PR. #4982 landed as a real merge commit with two parents. If #5005 is squashed, the merge commit is flattened, `master` never becomes an ancestor of `minor`, and the very next push to `master` reproduces this identical conflict set. The repo has squash, rebase and merge-commit all enabled and squash is the default, hence the flag.

## What was actually conflicted

28 files: the 27 i18n catalogs plus `multi-select.tsx`.

The catalogs were mechanical — took the `minor` side, re-ran `lingui extract`. The `master` side was pure extract churn (shifted line references plus one new msgid with an empty msgstr). Verified after regenerating: `de`/`cs`/`fr` each carry 1399 non-empty `msgstr`, same as the `minor` tip, and the new validation msgids are present across all 27.

`multi-select.tsx` was resolved toward `minor` — the result is byte-identical to the `minor` tip. Worth being precise about why, because my original PR description overstated it:

- The #4198 **fix** was never in this file. It is `extendSchema` + `superRefine` in `channels_.$id.tsx` and `form-schema-tools.ts`, both of which merged clean and are untouched.
- What the `master` side had here was a UX affordance riding along: a controlled `Popover` with manual `open` state so a single-select closes on pick. The DS v2 migration (#4957) deleted that hand-rolled Popover outright.
- That affordance is now native. For **Default currency** the relevant path is `Select`, not `Combobox` — `multi-select.tsx` gates the combobox on `items.length > 10`, and the available-currency list is well under that. Base UI's `Select` closes on item press.
- One correction to something I wrote earlier: a DS v2 multi-select does **not** unconditionally stay open on selection. `AriaCombobox.js:522-535` closes it when a filter is active and the input sits outside the popup, which is exactly this component's chips input.

Net: no behaviour lost, but via a different component than I first claimed.

## The part that needed real work — and a question for you

`packages/dashboard/e2e/tests/settings/channels.spec.ts` came over from `master` with #4198 as a **clean addition — it never appeared as a conflict**. It asserts against the old MultiSelect DOM (`[data-slot="popover-content"]`, options as buttons, a `Remove <currency>` button), none of which DS v2 renders. Two tests failed after the merge; that is what the red `dashboard e2e (3)` was.

I retargeted them at the Base UI combobox, matching the conventions the `Channels CRUD` block in the same file already uses. Along the way, in response to review:

- the empty-list assertion was vacuous — `toHaveCount(0)` on options passes equally well if the field never rendered, so it now anchors positively on the trigger placeholder
- option queries are scoped to the open popup (several `select-content` nodes are mounted at once on this form)
- the multi-select value is read from the chip, not the form item as a whole
- both tests leaked a channel, so they only passed against a cold database — added an `afterAll`
- `closeDropdown` moved onto `BaseDetailPage`, and `selectPopoverOption` now routes through it

**The question:** this is arguably debt from #4957 rather than from the back-merge — the DS v2 migration rewrote the component but not the e2e tests aimed at its old DOM. I fixed it here because otherwise the back-merge is red, but it does make this PR bigger than a pure conflict resolution. Happy to split it out if you would rather review it separately.

## One related thing I did *not* touch

`channels.spec.ts:61` asserts `[data-slot="popover-content"]` is not visible after a MultiSelect interaction. Nothing in DS v2 renders that slot, so it passes vacuously. It is pre-existing on `minor`, not something this merge broke, so I left it out of the diff — flagging it rather than expanding scope. It is the last stale MultiSelect DOM assertion in the repo; I checked the other 11 `popover-content` usages and they all belong to components that genuinely still use a Popover (faceted filter, relation selector, DateTimePicker, address selector), and the specs that do drive a MultiSelect (`roles`, `administrators`, `global-settings`) are role-based and DS v2-safe.

## Verification

Clean `bun install --frozen-lockfile` + `build:core-common`:

| Check | Result |
|---|---|
| `@vendure/dashboard` `check-types` | 0 errors |
| `@vendure/core` unit | 1108 passed (76 files) |
| `@vendure/dashboard` unit | 585 passed, 4 skipped (43 files) |
| `channels.spec.ts` | 14/14, three consecutive runs, `--retries=0` |
| `roles` + `administrators` (share the touched page object) | 18/18 |
| CI `dashboard e2e` shards 1-4 | all green |

`deploy` is red because this is a fork PR without `VERCEL_TOKEN` — unrelated.

Closes #4996.

grolmus:
### 📋 Two things to get right at merge time

Pulling these out of the comment above so they are next to the merge button.

**1. Merge with "Create a merge commit" — not squash.**

Squashing flattens the merge commit, so `master` never becomes an ancestor of `minor` and the very next push to `master` reproduces this identical conflict set. #4982 landed as a real merge commit with two parents. The repo has squash as the default, hence the flag.

**2. Close #4996 by hand afterwards — and not before.**

The `Closes #4996` in the body will not do it: GitHub's closing keywords only auto-close issues, never pull requests. #4981 was closed manually by @dlhck in the same minute #4982 merged, which is the pattern to follow.

Closing it *earlier* would be actively unhelpful. `back_merge.yml` only skips creating a new PR while an open one already exists:

```sh
EXISTING_PR=$(gh pr list --base "$TARGET" --head "$SOURCE" --state open --json number --jq '.[0].number // empty')
if [ -n "$EXISTING_PR" ]; then ... # skip creation
```

So while `minor` is still conflicted, #4996 is the only thing stopping the bot opening another one. Close it once this merges, not before.

---

For completeness on #4996 itself: it was opened by the `github-actions` bot and nobody from the team had touched it — all six comments on it are automated (4× "New conflicts detected", plus Vercel and the docs preview), and it had no assignee. It is also unmergeable by construction, since its head is the `master` branch itself, so the resolution cannot be pushed into it. That is why this PR exists separately, mirroring #4981 → #4982.

michaelbromley (approved):

PR #5006, opened 2026-07-20
fix(core): Guard relation custom field resolution against a missing entity id

### LINKED ISSUE #5004: relation-type custom field on OrderAddress throws "loader.load() ... got: undefined" in v3.7 (DataLoader called with an undefined entity id)
**Describe the bug**

A `relation`-type custom field defined on the `Address` entity crashes when it is selected on an
`OrderAddress` (i.e. `Order.shippingAddress` / `Order.billingAddress`) via the built-in
`customFields.<relation>` field. The query fails with the DataLoader error
`The loader.load() function must be called with a value, but got: undefined.`

`OrderAddress` is an embedded value object (a JSON snapshot stored on the order), so it has no
database `id`. The custom-field relation resolver introduced in v3.7
(`CustomFieldRelationResolverService`) batches the lookup through a DataLoader keyed by the parent
entity id — which is `undefined` for an `OrderAddress` — and calls `loader.load(undefined)`, which
DataLoader rejects.

Because the failure comes back as a GraphQL error inside an otherwise-`200` response (not a thrown
server error), it is invisible in server logs and can silently break storefront checkout when the
active-order query selects the field.

**To Reproduce**

1. Define a `relation`-type custom field on `Address` (any target entity; the crash is independent
   of which):

   ```typescript
   customFields: {
     Address: [
       { name: 'pickupLocation', type: 'relation', entity: Country, nullable: true },
     ],
   }
   ```

2. Via the Shop API, create an active order and set a shipping address that includes the custom
   field:

   ```graphql
   mutation {
     addItemToOrder(productVariantId: "1", quantity: 1) { ... on Order { id } }
   }
   mutation {
     setOrderShippingAddress(input: {
       streetLine1: "1 Test St", city: "Test", countryCode: "US",
       customFields: { pickupLocationId: "1" }
     }) { ... on Order { id } }
   }
   ```

3. Query the relation custom field on the order's shipping address:

   ```graphql
   query {
     activeOrder {
       shippingAddress {
         customFields {
           pickupLocation { id }
         }
       }
     }
   }
   ```

4. See error (the mutation/query returns a GraphQL error rather than data).

**Expected behavior**

The field resolves to the related entity (or `null` when unset), the same as it did before v3.7. An
embedded value object with no `id` should resolve the relation as `null` (or `[]` for a list field),
not throw.

**Actual behavior**

The query returns a GraphQL error and no data for the field:

```
The loader.load() function must be called with a value, but got: undefined.
```

**Screenshots/Videos**

N/A (Shop API / GraphQL).

**Error logs**

```
The loader.load() function must be called with a value, but got: undefined.
    at DataLoader.load (.../node_modules/dataloader/index.js)
    at CustomFieldRelationResolverService.resolveRelation (.../@vendure/core/dist/api/common/custom-field-relation-resolver.service.js:42)
```

Root cause, for reference (paths are on `master` and unchanged from `3.7.1`):

- `packages/core/src/api/config/generate-resolvers.ts` — the `customFields` resolver attaches the
  parent id so relations can be resolved:
  ```typescript
  const customFieldResolver = source => ({ ...source.customFields, [ENTITY_ID_KEY]: source.id });
  ```
  For an `OrderAddress`, `source.id` is `undefined`, so `ENTITY_ID_KEY` is `undefined`.
- The generated relation field resolver falls through to:
  ```typescript
  const entityId = source[ENTITY_ID_KEY];          // undefined for OrderAddress
  return customFieldRelationResolverService.resolveRelation({ ctx, fieldDef, entityName, entityId });
  ```
- `packages/core/src/api/common/custom-field-relation-resolver.service.ts` — `resolveRelation` calls
  the DataLoader with no null guard:
  ```typescript
  const loader = this.getLoader(ctx, entityName, fieldDef);
  const batchResult = await loader.load(entityId);   // entityId === undefined -> throws
  ```

- @vendure/core version: 3.7.1 (also present on current `master` — `resolveRelation` still calls
  `loader.load(entityId)` with no null check)
- Nodejs version: 24.14.1
- Operating System (Windows/macOS/Linux): macOS (also reproduces on Linux CI)
- Browser (if applicable): N/A (GraphQL Shop API)

**Configuration**

```typescript
import { Country, VendureConfig } from '@vendure/core';

  // ...standard config...
  customFields: {
    Address: [
      {
        name: 'pickupLocation',
        type: 'relation',
        entity: Country, // any entity reproduces it
        nullable: true,
      },
    ],
  },
};
```

**Minimal reproduction**

The config + the three Shop API operations above reproduce it on a vanilla Vendure 3.7.x project.
Happy to put together a standalone repo if useful.

**Workaround**

Expose the relation through a hand-written field resolver on `OrderAddress` that reads the raw
foreign-key scalar from `customFields` (`<name>Id`) and loads the target entity itself, and avoid
selecting `OrderAddress.customFields.<relation>` in queries (use the hand-written top-level field
instead). Sketch:

```typescript
@Resolver('OrderAddress')
export class OrderAddressRelationResolver {
  constructor(private connection: TransactionalConnection) {}

  @ResolveField()
  pickupLocation(@Ctx() ctx: RequestContext, @Parent() address: Address) {
    const id = address?.customFields?.pickupLocationId;
    if (!id) return null;
    return this.connection.getRepository(ctx, Country).findOne({ where: { id } });
  }
}
```

**Additional context**

- Introduced by #4923 (*perf(core): resolve relation custom fields using request-scoped DataLoader
  batching*, merged into `master` 2026-07-13). That PR is where `CustomFieldRelationResolverService`
  started resolving relation custom fields through `loader.load(entityId)`. Before it, these
  relations resolved without DataLoader and returned `null` on an `OrderAddress` (which has no id)
  instead of throwing — so this is a behavioral regression for embedded/idless types, despite the
  PR being labelled "no breaking changes". Started for us after upgrading from 3.6.x to 3.7.1.
- Happens **consistently/deterministically** whenever a `relation` custom field is selected on an
  `OrderAddress` whose `customFields` object is non-null (i.e. when any address custom field is set).
- **Invisible in server logs** — it surfaces only as a GraphQL error in the response body, so it can
  silently break checkout flows that read the order's shipping/billing address.

**Suggested fix**

Guard `entityId` in `resolveRelation` — an embedded value object legitimately has no id, so the
relation should resolve to `null`/`[]`:

```typescript
async resolveRelation(config: ResolveRelationConfig) {
    const { ctx, entityId, entityName, fieldDef } = config;
    if (entityId == null) {
        return fieldDef.list ? [] : null;
    }
    const loader = this.getLoader(ctx, entityName, fieldDef);
    const batchResult = await loader.load(entityId);
    // ...unchanged...
}
```

brmk:
## Description

Relation-type custom fields selected on an embedded value object with no database `id` — the canonical case being `OrderAddress` (`Order.shippingAddress` / `billingAddress`) — crash with:

```
The loader.load() function must be called with a value, but got: undefined.
```

This is a regression from #4923 (*perf(core): resolve relation custom fields using request-scoped DataLoader batching*): relation custom fields are now resolved through a DataLoader keyed by the parent entity id. For an `OrderAddress` that id is `undefined` (it is a JSON snapshot with no entity id), so `DataLoader.load(undefined)` throws. It surfaces as a GraphQL error (HTTP 200 with an `errors` array), so it is invisible in server logs and can silently break checkout flows that read the order's shipping/billing address. Before #4923 these relations resolved without DataLoader and returned `null` on an `OrderAddress` rather than throwing.

## Fix

Guard the undefined `entityId` in `CustomFieldRelationResolverService.resolveRelation` and resolve to `null` (or `[]` for a list field) — an embedded value object legitimately has no id. `ResolveRelationConfig.entityId` is widened to `ID | undefined` to reflect what the single call site (`generate-resolvers.ts`, `source[ENTITY_ID_KEY]`) actually passes.

## Reproduction

1. Define a `relation` custom field on `Address` (any target entity).
2. Shop API: create an active order, then `setOrderShippingAddress` with the custom field set.
3. Query `activeOrder { shippingAddress { customFields { <relation> { id } } } }` → GraphQL error.

## Tests

Added an e2e regression test to `custom-field-relations.e2e-spec.ts` (`Address entity` group) that drives the order shipping-address path. Verified locally:

- **With the fix:** the `custom-field-relations` e2e suite passes (62/62).
- **Without the fix:** the new test fails with `The loader.load() function must be called with a value, but got: undefined.`

Fixes #5004

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5006"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787136217&installation_id=128408429&pr_number=5006&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5006&signature=b50f71c9b64d8e10f725a4070d6b8f2930b10c08daa2d1d54cdb708dd79ca7e3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/custom-field-relations.e2e-spec.ts
+++ b/packages/core/e2e/custom-field-relations.e2e-spec.ts
@@ -454,6 +454,51 @@ describe('Custom field relations', () => {
 
                 assertCustomFieldIds(updateCustomerAddress.customFields, 'T_3', ['T_2', 'T_4']);
             });
+
+            // Regression: an OrderAddress (Order.shippingAddress / billingAddress) is an embedded
+            // value object with no `id`, so its relation custom fields cannot be resolved by a parent
+            // entity id. Previously this threw "The loader.load() function must be called with a
+            // value, but got: undefined"; it must now resolve to null / [] instead of crashing.
+            it('order shippingAddress resolves relation custom fields without throwing', async () => {
+                await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+                await shopClient.query(gql`
+                    mutation {
+                        addItemToOrder(productVariantId: "T_1", quantity: 1) {
+                            ... on Order {
+                                id
+                            }
+                        }
+                    }
+                `);
+                await shopClient.query(gql`
+                    mutation {
+                        setOrderShippingAddress(
+                            input: {
+                                countryCode: "GB"
+                                streetLine1: "Test Street"
+                                customFields: { singleId: "T_1", multiIds: ["T_1", "T_2"] }
+                            }
+                        ) {
+                            ... on Order {
+                                id
+                            }
+                        }
+                    }
+                `);
+
+                const { activeOrder } = await shopClient.query(gql`
+                    query {
+                        activeOrder {
+                            shippingAddress {
+                                ${customFieldsSelection}
+                            }
+                        }
+                    }
+                `);
+
+                expect(activeOrder.shippingAddress.customFields.single).toBeNull();
+                expect(activeOrder.shippingAddress.customFields.multi).toEqual([]);
+            });
         });
 
         describe('Collection entity', () => {
--- a/packages/core/src/api/common/custom-field-relation-resolver.service.ts
+++ b/packages/core/src/api/common/custom-field-relation-resolver.service.ts
@@ -16,7 +16,7 @@ import { RequestContext } from './request-context';
 
 export interface ResolveRelationConfig {
     ctx: RequestContext;
-    entityId: ID;
+    entityId: ID | undefined;
     entityName: string;
     fieldDef: RelationCustomFieldConfig;
 }
@@ -39,6 +39,15 @@ export class CustomFieldRelationResolverService {
      */
     async resolveRelation(config: ResolveRelationConfig): Promise<VendureEntity | VendureEntity[] | null> {
         const { ctx, entityId, entityName, fieldDef } = config;
+
+        // Embedded value objects such as OrderAddress (Order.shippingAddress / billingAddress) have
+        // no `id`, so `entityId` can be undefined. Passing it to DataLoader.load() would throw
+        // ("The loader.load() function must be called with a value, but got: undefined"); there is
+        // nothing to resolve by id, so return an empty result, matching the behaviour before
+        // relation custom fields were resolved via DataLoader.
+        if (entityId == null) {
+            return fieldDef.list ? [] : null;
+        }
 
         const loader = this.getLoader(ctx, entityName, fieldDef);
         const batchResult = await loader.load(entityId);


brmk:

grolmus (approved):
Thanks @brmk — clean, honest, well-scoped fix. Reviewed the diff, traced the root cause, and ran it locally.

**Root cause confirmed:** `generate-resolvers.ts` attaches `[ENTITY_ID_KEY]: source.id` to the custom-fields resolver value, and for an `OrderAddress` (an embedded value object with no `id`) that's `undefined`, which then reached `loader.load(undefined)`. Guarding `entityId == null` and returning `[]`/`null` is the right fix, in the right place:

- It sits after the eager-load short-circuit, so no id-bearing entity can wrongly hit the null branch.
- `== null` correctly matches only null/undefined (Vendure ids are never a falsy-but-valid `0`).
- Returning null restores the documented pre-3.7 behaviour — the stored FK scalar (`singleId`) isn't consulted by either resolution path even today (the DataLoader path joins the parent entity table by id, which an embedded object doesn't have), so null is the honest result here rather than something being silently dropped.
- The type widening to `ID | undefined` is accurate and non-breaking (`ResolveRelationConfig` isn't part of the public API).

**Testing:** the new e2e reproduces the real crash path (set a shipping address with relation custom fields, then query them). Ran the full `custom-field-relations` suite locally — 62 passed.

Two optional, non-blocking follow-ups (no need to hold this PR):
1. A one-line assertion on `customFields.primitive` in the new test would confirm the guard doesn't collaterally affect scalar custom fields on the snapshot.
2. Actually resolving `OrderAddress` relation custom fields from the stored `<name>Id` scalar would be a nice future enhancement — it's a separate feature from this crash fix, so worth a follow-up issue if the team wants it.

Approving. 🚀

PR #5007, opened 2026-07-20
fix(core): prevent duplicate translations from concurrent saves

### LINKED ISSUE #4884: Duplicate translations (same entity + languageCode) created by concurrent/double-click Admin save; entity then becomes uneditable
**Describe the bug**

Translatable entities can end up with two translation rows for the same `languageCode`. Vendure only
enforces "one translation per language" in application code (`TranslatableSaver` loads the existing
translations, then inserts any language that isn't already there), and there is no unique constraint on
the translation table. So if two updates for the same entity run close together, before either commits,
both see "no `en` row yet" and both insert one.

Once that happens the entity can no longer be edited from the Admin UI. Saving appears to do nothing,
because with two `en` rows the read and write paths can resolve to different rows. The only way to fix it
is to delete the extra row directly in the database.

It lives in the shared `TranslatableSaver`, so any translatable entity is affected. We've hit it on
`product_translation`, `product_variant_translation` and `collection_translation`.

**To Reproduce**

In the Angular Admin the Update button isn't disabled while a save is in flight, so a double-click sends
the mutation twice:

1. Open a product that has no English translation yet.
2. Switch the language selector to English and fill in the name.
3. Double-click Update.
4. The DB now has two `product_translation` rows with `languageCode = 'en'` for that product.
5. Edit the product again — changes no longer save. It stays uneditable until one row is deleted in the DB.

It doesn't depend on the UI: two `updateProduct` mutations run in parallel, each adding a language the
product doesn't have yet, give the same result.

**Expected behavior**

An entity should never have more than one translation per language. Concurrent saves should be idempotent
or fail, and the entity should stay editable.

**Actual behavior**

Two rows are created for the same `(baseId, languageCode)`, no error is raised, and afterwards the entity
can't be updated through the Admin UI until the duplicate is removed in the database.

**Screenshots/Videos**

Video attached (Angular Admin): a double-click creates the duplicate rows, after which the product can no
longer be updated.

https://github.com/user-attachments/assets/3f912d39-3be4-47ae-82e4-f0bb6d5e6734

**Error logs**

```
None — it fails silently.
```

- Nodejs version: 24
- Database (mysql/postgres etc): PostgreSQL 15
- Operating System (Windows/macOS/Linux): Linux (server) / macOS (dev)

**Configuration**

Default `AdminUiPlugin` and Postgres. Nothing custom is needed to reproduce.

**Minimal reproduction**

```ts
const input = (t) => ({ id: productId, translations: [t] });
await Promise.all([
  adminClient.query(UPDATE_PRODUCT, { input: input({ languageCode: LanguageCode.en, name: 'X', slug: 'x' }) }),
  adminClient.query(UPDATE_PRODUCT, { input: input({ languageCode: LanguageCode.en, name: 'X', slug: 'x' }) }),
]);
// two product_translation rows with languageCode = 'en'
```

**Workaround**

Delete the extra rows, keeping one per language:

```sql
DELETE FROM product_translation
WHERE id IN (
  SELECT id FROM (
    SELECT id, ROW_NUMBER() OVER (
      PARTITION BY "baseId", "languageCode" ORDER BY "updatedAt" DESC, id DESC) AS rn
    FROM product_translation
  ) s WHERE rn > 1
);
```

Find them with:

```sql
SELECT "baseId", "languageCode", COUNT(*)
FROM product_translation
GROUP BY "baseId", "languageCode" HAVING COUNT(*) > 1;
```

**Additional context**

We know the Angular Admin is being deprecated, and the React dashboard does avoid the double-click (its
save button is disabled while the mutation is pending). But this isn't really a UI bug, and a UI change
won't fix it. The Admin UI is only one caller — custom CRM connectors, import scripts and other plugins go
through the same `update` APIs, and two of them running close together hit the same race. Leaving it
unprotected at the database level will keep causing duplicates regardless of which UI is used, so the real
fix has to be backend enforcement.

For us it's intermittent, roughly once per few hundred products entered by hand. It only happens the first
time a language is added to an entity; updating an existing translation is fine.

Where it comes from:

- `TranslatableSaver.update()` loads existing translations and `TranslationDiffer` inserts any language
  not already present, with no lock and no unique constraint
  (`packages/core/src/service/helpers/translatable-saver/translatable-saver.ts`, `translation-differ.ts`).
- The mutation runs in one `@Transaction()` (`api/resolvers/admin/product.resolver.ts`), so under READ
  COMMITTED two overlapping transactions don't see each other's uncommitted insert.
- There is no unique index on the translation `(baseId, languageCode)`.

A unique constraint on `(<entity>Id, languageCode)` across the translation tables would stop it at the
source. Ideally `TranslatableSaver.update` would also handle the conflict cleanly (upsert, or a clear
error) so a losing concurrent save doesn't 500 or leave the entity stuck.

brmk:
# Description

Translatable entities (`Product`, `Collection`, `ProductVariant`, etc.) could end up with two translation rows for the same `(baseId, languageCode)` pair. `TranslatableSaver.update()` loads existing translations, then `TranslationDiffer` inserts any language not already present — with no locking and no unique constraint. Two concurrent updates that both add the same new language (e.g. a double-click on Update in the Admin UI) can both read "no `en` row yet" before either commits, and both insert one. Once that happens, the entity becomes uneditable through the Admin UI.

This PR:

- **Registers a unique constraint on `(languageCode, base)` dynamically at bootstrap** for every translation entity, in `preBootstrapConfig()` alongside `registerCustomEntityFields()`. Any entity implementing `Translation<T>` (detected from the TypeORM metadata by its `languageCode` column and `base` relation) receives the constraint — including custom translatable entities defined by plugins, with no decorator needed on their part. This replaces the earlier revision's `@Unique` decorator on each of the 13 core translation entities, per review feedback. The generated constraint names are identical to what the decorator produced, since TypeORM derives them from the table and columns.
- Adds `isUniqueConstraintViolationError()` to `db-errors.ts` (Postgres/MySQL/SQLite), mirroring the existing `isForeignKeyViolationError()`.
- Updates `TranslationDiffer.applyDiff()` so that when an insert hits this constraint (a concurrent request already inserted the row), the losing request converges instead of throwing a 500. The insert runs in a savepoint (so the enclosing Postgres transaction survives the violation), and the recovery **looks the winning row up on the raw connection and adopts it as-is**. The lookup must happen outside the request's transaction because under REPEATABLE READ (the MySQL/MariaDB default) the transaction's snapshot predates the concurrent commit; writing to the row from inside the transaction is not possible either, because MariaDB ≥ 11.6.2 (`innodb_snapshot_isolation=ON` by default) rejects any write to a row that changed after the snapshot was taken. For the practical trigger of this race — the same update submitted twice via double-click — the payloads are identical, so adopting the winner's row is equivalent to updating it.
- **`vendure migrate generate` handles existing duplicates automatically.** Adds `deduplicateTranslations()` to `migration-utils` (keeps the most recently updated row per `(baseId, languageCode)`, deletes the rest; works on Postgres/MySQL/MariaDB/SQLite). `generateMigration()` detects which translation tables the generated DDL adds the unique constraint to — core or plugin-defined — and emits a `deduplicateTranslations()` call for exactly those tables at the top of `up()`, ahead of the DDL. No manual step, nothing to discover from a JSDoc. The generated file looks like:

  ```ts
  import {MigrationInterface, QueryRunner} from "typeorm";
  import { deduplicateTranslations } from "@vendure/core";

  export class AddUnique1700000000000 implements MigrationInterface {
     public async up(queryRunner: QueryRunner): Promise<any> {
          // Remove any duplicate (baseId, languageCode) translation rows left over from before the
          // unique constraint existed, keeping the most recently updated row of each pair. This must
          // run before the constraint is created below. See https://github.com/vendurehq/vendure/issues/4884
          await deduplicateTranslations(queryRunner, ['collection_translation', 'product_translation']);
          await queryRunner.query(`ALTER TABLE "collection_translation" ADD CONSTRAINT "UQ_858c112351f7714960a2dfcff91" UNIQUE ("languageCode", "baseId")`, undefined);
          await queryRunner.query(`ALTER TABLE "product_translation" ADD CONSTRAINT "UQ_dcc35f0d2b8d422634e878b813c" UNIQUE ("languageCode", "baseId")`, undefined);
     }
     ...
  ```
- Tightens `isUniqueConstraintViolationError()`: no bare `SQLITE_CONSTRAINT` match (it covers FK/not-null/check failures too), message fallback narrowed to `unique constraint`, with unit tests for both classifiers.
- Adds unit tests for the dynamic constraint registration (detection, idempotence, inheritance, plugin-style entities, all 13 core entities) and for the recovery branch of `applyDiff()` — the latter because on sql.js concurrent requests serialize on a single connection, so the e2e race cannot fire there deterministically.
- Adds an e2e regression test (two concurrent `updateProduct` calls adding the same new language) asserting the invariant: no duplicate translation row is created and the product remains editable afterwards.

## Validation

`product.e2e-spec.ts` run locally against all backends (101/101 in each case):

| backend | result |
|---|---|
| sql.js | ✅ 101/101 |
| Postgres 16 | ✅ 101/101 |
| MySQL 8.0.46 | ✅ 101/101 |
| MariaDB 11.5 (CI image) | ✅ 101/101 |
| MariaDB 12.3.2 (`innodb_snapshot_isolation=ON`) | ✅ 101/101 |

On MariaDB ≥ 11.6.2 the losing concurrent request can still be rejected with `Record has changed since last read` — on the **product row itself**, before any translation code runs. That is MariaDB's snapshot-isolation feature applying to any concurrent update of the same row anywhere in Vendure (it is why CI pins `mariadb:11.5`), and is out of scope here. The invariant this PR guarantees holds there too: no duplicate rows, and the entity stays editable.

Plugin coverage was verified end-to-end: the `campaign_translation` table from the `PluginIssue2453` e2e fixture (a plugin-defined translatable entity with no decorator) receives `UNIQUE ("languageCode", "baseId")` on a real Postgres via the dynamic registration.

Fixes #4884

# Breaking changes

Adds a unique constraint to the translation tables (the 13 core ones, plus any plugin-defined translation tables, since the dynamic registration covers those too), so upgrading requires a generated migration. Databases that already contain duplicate rows for the same `(baseId, languageCode)` pair have those duplicates removed by the generated migration (see above). No changes to any public API surface otherwise (`deduplicateTranslations` is a new, additive export).

## Migration guide for users

1. Upgrade `@vendure/core` and run `vendure migrate generate <name>` as usual.
2. If any translation table in your database gains the new constraint, the generated migration starts with a `deduplicateTranslations()` call for those tables (see above). Review it like any other generated migration — it deletes the redundant duplicate rows, keeping the most recently updated row of each `(baseId, languageCode)` pair, so no translation content is lost.
3. Run the migration.

Only if you write migrations by hand do you need to call `deduplicateTranslations()` yourself, before the constraint DDL. Worth a line in the release notes, but there is no failure mode that requires the user to go looking for it.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/e2e/product.e2e-spec.ts
+++ b/packages/core/e2e/product.e2e-spec.ts
@@ -1064,6 +1064,47 @@ describe('Product resolver', () => {
             ),
         );
 
+        // #4884 — concurrent updates adding the same not-yet-existing language must not
+        // create two translation rows for that (product, languageCode) pair
+        it('concurrent updates adding the same new language do not create duplicate translations', async () => {
+            const { createProduct } = await adminClient.query(createProductDocument, {
+                input: {
+                    translations: [
+                        {
+                            languageCode: LanguageCode.en,
+                            name: 'en Concurrent Potato',
+                            slug: 'en-concurrent-potato',
+                            description: 'A potato',
+                        },
+                    ],
+                },
+            });
+
+            const input = {
+                id: createProduct.id,
+                translations: [
+                    {
+                        languageCode: LanguageCode.de,
+                        name: 'de Concurrent Potato',
+                        slug: 'de-concurrent-potato',
+                        description: 'Ein Erdapfel',
+                    },
+                ],
+            };
+
+            await Promise.all([
+                adminClient.query(updateProductDocument, { input }),
+                adminClient.query(updateProductDocument, { input }),
+            ]);
+
+            const { product } = await adminClient.query(getProductWithVariantsDocument, {
+                id: createProduct.id,
+            });
+            productGuard.assertSuccess(product);
+            const deTranslations = product.translations.filter(t => t.languageCode === LanguageCode.de);
+            expect(deTranslations.length).toBe(1);
+        });
+
         it('addOptionGroupToProduct adds an option group', async () => {
             const optionGroup = await createOptionGroup('Quark-type', ['Charm', 'Strange']);
             const result = await adminClient.query(addOptionGroupToProductDocument, {
--- a/packages/core/e2e/product.e2e-spec.ts
+++ b/packages/core/e2e/product.e2e-spec.ts
@@ -1092,10 +1092,17 @@ describe('Product resolver', () => {
                 ],
             };
 
-            await Promise.all([
+            // On MariaDB >= 11.6.2 (innodb_snapshot_isolation=ON by default) the losing
+            // request can be rejected with "Record has changed since last read" when both
+            // transactions update the same product row — a database-level concurrency
+            // conflict, not the bug under test. The invariant here is not that both
+            // requests succeed, but that no duplicate translation row is created and the
+            // product remains editable afterwards.
+            const results = await Promise.allSettled([
                 adminClient.query(updateProductDocument, { input }),
                 adminClient.query(updateProductDocument, { input }),
             ]);
+            expect(results.some(result => result.status === 'fulfilled')).toBe(true);
 
             const { product } = await adminClient.query(getProductWithVariantsDocument, {
                 id: createProduct.id,
@@ -1103,6 +1110,29 @@ describe('Product resolver', () => {
             productGuard.assertSuccess(product);
             const deTranslations = product.translations.filter(t => t.languageCode === LanguageCode.de);
             expect(deTranslations.length).toBe(1);
+
+            // Before the unique constraint, a duplicate pair of rows made the entity
+            // uneditable (#4884) — verify a subsequent update still lands.
+            await adminClient.query(updateProductDocument, {
+                input: {
+                    id: createProduct.id,
+                    translations: [
+                        {
+                            languageCode: LanguageCode.de,
+                            name: 'de Concurrent Potato zwei',
+                            slug: 'de-concurrent-potato',
+                            description: 'Ein Erdapfel',
+                        },
+                    ],
+                },
+            });
+            const { product: productAfter } = await adminClient.query(getProductWithVariantsDocument, {
+                id: createProduct.id,
+            });
+            productGuard.assertSuccess(productAfter);
+            const deAfter = productAfter.translations.filter(t => t.languageCode === LanguageCode.de);
+            expect(deAfter.length).toBe(1);
+            expect(deAfter[0].name).toBe('de Concurrent Potato zwei');
         });
 
         it('addOptionGroupToProduct adds an option group', async () => {
--- a/packages/core/src/bootstrap.ts
+++ b/packages/core/src/bootstrap.ts
@@ -21,6 +21,7 @@ import {
     getEntityNamesWithCustomFields,
     registerCustomEntityFields,
 } from './entity/register-custom-entity-fields';
+import { registerTranslationEntityUniqueConstraints } from './entity/register-translation-unique-constraints';
 import { runEntityMetadataModifiers } from './entity/run-entity-metadata-modifiers';
 import { setEntityIdStrategy } from './entity/set-entity-id-strategy';
 import { setMoneyStrategy } from './entity/set-money-strategy';
@@ -337,6 +338,7 @@ export async function preBootstrapConfig(
     patchTypeOrmEmbeddedRelationColumns();
     patchTypeOrmRelationIdLoader();
     registerCustomEntityFields(config);
+    registerTranslationEntityUniqueConstraints(entities);
     setEntityIdStrategy(entityIdStrategy, entities);
     const moneyStrategy = config.entityOptions.moneyStrategy;
     setMoneyStrategy(moneyStrategy, entities);
--- a/packages/core/src/entity/api-key/api-key-translation.entity.ts
+++ b/packages/core/src/entity/api-key/api-key-translation.entity.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { DeepPartial } from '@vendure/common/lib/shared-types';
-import { Column, Entity, Index, ManyToOne } from 'typeorm';
+import { Column, Entity, Index, ManyToOne, Unique } from 'typeorm';
 
 import { Translation } from '../../common/types/locale-types';
 import { HasCustomFields } from '../../config/custom-field/custom-field-types';
@@ -10,6 +10,7 @@ import { CustomApiKeyFieldsTranslation } from '../custom-entity-fields';
 import { ApiKey } from './api-key.entity';
 
 @Entity()
+@Unique(['languageCode', 'base'])
 export class ApiKeyTranslation extends VendureEntity implements Translation<ApiKey>, HasCustomFields {
     constructor(input?: DeepPartial<Translation<ApiKey>>) {
         super(input);
--- a/packages/core/src/entity/api-key/api-key-translation.entity.ts
+++ b/packages/core/src/entity/api-key/api-key-translation.entity.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { DeepPartial } from '@vendure/common/lib/shared-types';
-import { Column, Entity, Index, ManyToOne, Unique } from 'typeorm';
+import { Column, Entity, Index, ManyToOne } from 'typeorm';
 
 import { Translation } from '../../common/types/locale-types';
 import { HasCustomFields } from '../../config/custom-field/custom-field-types';
@@ -10,7 +10,6 @@ import { CustomApiKeyFieldsTranslation } from '../custom-entity-fields';
 import { ApiKey } from './api-key.entity';
 
 @Entity()
-@Unique(['languageCode', 'base'])
 export class ApiKeyTranslation extends VendureEntity implements Translation<ApiKey>, HasCustomFields {
     constructor(input?: DeepPartial<Translation<ApiKey>>) {
         super(input);
--- a/packages/core/src/entity/asset/asset-translation.entity.ts
+++ b/packages/core/src/entity/asset/asset-translation.entity.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { DeepPartial } from '@vendure/common/lib/shared-types';
-import { Column, Entity, Index, ManyToOne, Unique } from 'typeorm';
+import { Column, Entity, Index, ManyToOne } from 'typeorm';
 
 import { Translation } from '../../common/types/locale-types';
 import { HasCustomFields } from '../../config/custom-field/custom-field-types';
@@ -10,7 +10,6 @@ import { CustomAssetFieldsTranslation } from '../custom-entity-fields';
 import { Asset } from './asset.entity';
 
 @Entity()
-@Unique(['languageCode', 'base'])
 export class AssetTranslation extends VendureEntity implements Translation<Asset>, HasCustomFields {
     constructor(input?: DeepPartial<Translation<Asset>>) {
         super(input);
--- a/packages/core/src/entity/asset/asset-translation.entity.ts
+++ b/packages/core/src/entity/asset/asset-translation.entity.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { DeepPartial } from '@vendure/common/lib/shared-types';
-import { Column, Entity, Index, ManyToOne } from 'typeorm';
+import { Column, Entity, Index, ManyToOne, Unique } from 'typeorm';
 
 import { Translation } from '../../common/types/locale-types';
 import { HasCustomFields } from '../../config/custom-field/custom-field-types';
@@ -10,6 +10,7 @@ import { CustomAssetFieldsTranslation } from '../custom-entity-fields';
 import { Asset } from './asset.entity';
 
 @Entity()
+@Unique(['languageCode', 'base'])
 export class AssetTranslation extends VendureEntity implements Translation<Asset>, HasCustomFields {
     constructor(input?: DeepPartial<Translation<Asset>>) {
         super(input);
--- a/packages/core/src/entity/collection/collection-translation.entity.ts
+++ b/packages/core/src/entity/collection/collection-translation.entity.ts
@@ -1,6 +1,6 @@
 import { LanguageCode } from '@vendure/common/lib/generated-types';
 import { DeepPartial } from '@vendure/common/lib/shared-types';
-import { Column, Entity, Index, ManyToOne } from 'typeorm';
+import { Column, Entity, Index, ManyToOne, Unique } from 'typeorm';
 
 import { Translation } from '../../common/types/locale-types';
 import { HasCustomFields } from '../../config/custom-field/custom-field-types';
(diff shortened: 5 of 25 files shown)


michaelbromley:
**T3: Systemic** — resolves an open question by picking one interpretation (pattern 7).

- New `@Unique(['languageCode', 'base'])` constraint; stores with existing duplicate rows hit a failing migration
- Recovery is a manual `deduplicateTranslations()` call in the generated migration, discoverable only from a JSDoc comment
- Decision needed on the upgrade path: manual step / auto-inject into generated migration / bundled automatic core migration

ryandiginomad (commented):
I've been working in the translation code recently (#5059 touches `translateDeep`), and this PR overlaps areas I had to reason about there, so I gave it a proper look. Overall the shape is right: constraint at the DB layer, recovery at the application layer, and an escape hatch for dirty data. Since the main CI workflows haven't run on this PR yet (they need maintainer approval), I ran what I could locally against sql.js and read the rest against the TypeORM sources.

**Verified locally:**

- The savepoint mechanism from the third commit does what its comment says. `withTransaction()` on a ctx that already carries a transaction reaches `TransactionWrapper.executeInTransaction()`, which inherits the query runner and calls `startTransaction()` again; the postgres, mysql and sqlite drivers all declare `transactionSupport = "nested"`, so this issues a real `SAVEPOINT`, and the rollback path rolls back to it leaving the outer transaction usable.
- All 13 `*-translation.entity.ts` files in core get the `@Unique`; there is no 14th translatable entity left out (counted).
- `deduplicateTranslations()`: the ROW_NUMBER window with the derived-table wrapper runs on all three supported drivers, identifiers go through `driver.escape()`, and keep-newest by `(updatedAt, id)` is a sensible survivor rule.
- `product.e2e-spec.ts` on this branch: 101/101 (sql.js).

**1. On MySQL, the recovery path likely cannot see the concurrent row.** The fallback `findOne` after the violation runs inside the outer transaction. MySQL's default isolation is REPEATABLE READ, and the outer transaction's consistent snapshot was taken before the concurrent commit, so that `findOne` should return null and the code throws `InternalServerError` — the constraint still prevents the duplicate, but the losing request gets a 500 instead of converging, which is the behaviour the description promises to fix. Postgres (fresh snapshot per statement under READ COMMITTED) and SQLite are fine. One way out: run the fallback lookup via `rawConnection` so it reads outside the outer snapshot; by the time `ER_DUP_ENTRY` surfaces, the competing row is committed (the insert blocks on the conflicting lock until the other transaction resolves), so a fresh read will find it. Flagging honestly: this comes from reading the TypeORM and MySQL semantics, not from a reproduction against a real MySQL instance.

**2. The new e2e cannot reproduce the bug on sql.js.** I checked out the parent commit's `src` (old differ, no constraint), rebuilt the dist, and ran just the new test: it passes there too. With sql.js the two mutations share one connection and serialize end to end, so the second update always sees the first one's committed row and takes the update path — the race never fires. It should fire on the postgres/mysql CI backends, but those haven't run yet. A deterministic, backend-independent complement would be a unit test on `TranslationDiffer.applyDiff()` with a repository stub whose `save()` throws a unique-violation error once: that pins the retry branch, which currently has no test exercising it.

Happy to re-run anything if that helps. Tooling note: I used Claude Code while working through this; every check above is something I ran and read myself.

michaelbromley:
Hi, thanks for the PR. Here are my thoughts:

- the intent is correct, this is a real problem to be solved
- the solution with adding a unique index is good, BUT it will cause a DB migration for anyone who updates. Therefore it should not target `master`. We could target it for `minor` as we do allow smaller schema changes like this on a feature release.
- I'd like to have a better solution than manually decorating every translatable entity. Custom entities will then also need to add a similar decorator to align with core entities. I think we can achieve the same effect by dynalically defining the unique constraint during the bootstrap process, which we already do in `registerCustomEntityFields()`. A similar mechanism of introspecting and dynamically modifying the TypeORM metadata can be used for this purpose too.
- The point about MySQL/MariaDB compat is confirmed by the e2e tests failing for those 2 DB backends.

ryandiginomad:
Following up on my review from last week, now that CI has run and confirms the MySQL/MariaDB point. I reproduced both failures locally and worked out what it takes to clear them, because one of the results says something about the approach rather than about the code.

Setup: MySQL 8.0.46 and MariaDB 12.3.2 in Docker at their default REPEATABLE READ, Postgres 16 at READ COMMITTED, plus sql.js. `packages/core` built from this branch, e2e run against the built `dist`.

**The failure has two layers, and my review only named the first one.** The fallback `findOne` at `translation-differ.ts:89-96` runs on `ctx`, so it reads inside the request's transaction and its snapshot predates the competing commit. Move that lookup off `ctx` and the next line fails instead: the `save()` at `:101-103` carries the id we just assigned, TypeORM runs an existence SELECT before choosing insert vs update, and that SELECT sits in the same snapshot. It finds nothing, treats the row as new, and inserts.

```
brmk head            → Duplicate entry 'de-21' for key 'product_translation.IDX_dcc35f0d2b8d422634e878b813'
+ lookup off ctx     → Duplicate entry '22' for key 'product_translation.PRIMARY'
+ update() at :101   → green on MySQL 8
```

Replacing that `save()` with an `update()` against the id we already hold (a plain UPDATE requires no prior read) clears the whole file on three of the four backends:

| backend | this branch | both changes |
|---|---|---|
| sql.js | 101/101 | 101/101 |
| Postgres 16 | 101/101 | 101/101 |
| MySQL 8.0.46 | 1 failed / 100 passed | **101/101** |
| MariaDB 12.3.2 | 1 failed / 100 passed | **1 failed / 100 passed** |

MariaDB is the one worth looking at. With both changes it stops failing on a duplicate and fails on this instead:

```
Record has changed since last read in table 'product_translation'; try restarting transaction
```

The lookup now finds the row, and MariaDB refuses the write because the row has moved on since this transaction's snapshot was taken. MySQL 8 allows the same UPDATE, resolving it against the current version. So the split is not MySQL-family versus Postgres the way the CI matrix reads, it is MariaDB on its own, and what it rules out is wider than one statement: no write to that row from inside the request's transaction will land there, however it is phrased. MariaDB's own advice, restarting the transaction, points somewhere above this function.

One thing I could not settle: I tried collapsing the insert and the recovery into a single upsert to avoid the read-then-write altogether, and failed to get it working. It breaks on MySQL 8 with `Cannot update entity because entity id is not set`, which is my error and not the database's, so I have no evidence either way on that route. Saying so explicitly, since the table above should not be read as ruling it out.

None of this touches the diagnosis or the constraint, which both still look right to me. It does mean the converge-on-conflict behaviour needs to live somewhere other than this catch block if MariaDB is to be supported.

On tooling: Claude Code was part of my workflow here, and each run in the table is one I performed and read myself.

brmk:
Thanks @michaelbromley and @ryandiginomad for the thorough reviews — all points addressed:

**Retargeted to `minor`** and rebased on it.

**Dynamic constraint registration** — the 13 `@Unique` decorators are gone. `registerTranslationEntityUniqueConstraints()` now runs in `preBootstrapConfig()` next to `registerCustomEntityFields()` and adds the `(languageCode, base)` constraint to every entity implementing `Translation<T>`, detected from the TypeORM metadata, so plugin-defined translatable entities are covered with no decorator of their own. Verified end-to-end: `campaign_translation` from the PluginIssue2453 e2e fixture gets `UNIQUE ("languageCode","baseId")` on Postgres. Constraint names are unchanged, and migration generation goes through `preBootstrapConfig()` too, so `vendure migrate generate` picks the constraints up as before.

**MySQL/MariaDB recovery** — @ryandiginomad's two-layer diagnosis reproduced exactly on my side, thanks for the groundwork. The catch block no longer writes at all: it looks the winning row up on `rawConnection` (outside the request transaction, where the concurrent commit is visible on every backend) and adopts it as-is. For the double-click trigger the payloads are identical, so adopting equals updating; and under MariaDB's snapshot isolation no in-transaction write could do better anyway. Local results for `product.e2e-spec` (101 tests): sql.js ✅, Postgres 16 ✅, MySQL 8.0.46 ✅, MariaDB 11.5 ✅, MariaDB 12.3.2 with `innodb_snapshot_isolation=ON` ✅.

The residual MariaDB ≥ 11.6.2 failure turned out to be on the **product row itself** — both transactions update the base row, and snapshot isolation rejects the loser before any translation code runs. That applies to any concurrent same-row update anywhere in Vendure (presumably why CI pins `mariadb:11.5`) and is out of scope here; the e2e now asserts the actual invariant instead: no duplicate row, at least one request succeeds, and the product remains editable afterwards.

**Unit tests** added for the previously uncovered recovery branch of `applyDiff()` (with a repository stub throwing a unique violation, as suggested) and for the dynamic registration (detection, idempotence, inheritance, plugin-style entities, all 13 core entities).

**On the upgrade path question**: with dynamic registration the generated migration includes the constraints automatically; the manual `deduplicateTranslations()` step remains only for databases that already contain duplicates. Auto-injecting that call into the generated migration would remove even that step — happy to explore it here or as a follow-up, whichever you prefer.

biggamesmallworld (changes requested):
Requesting changes on two points — happy to approve once they're addressed:

1. **Upgrade path** — Please get an explicit decision on manual JSDoc step vs auto-inject into `generateMigration` vs bundled core migration (as Michael flagged). “Read the JSDoc when migrate fails” isn’t enough for a framework schema change.
2. **`isUniqueConstraintViolationError`** — Tighten detection (drop bare `SQLITE_CONSTRAINT`; narrow the `/\bunique\b/i` fallback to match the FK helper’s specificity) and add unit tests for the classifier.

brmk:
Thanks @biggamesmallworld , both done.

1. `vendure migrate generate` now takes care of it. When the generated migration adds the unique constraint to a translation table, it also adds a `deduplicateTranslations()` call for that table at the top of `up()`. Users just run `migrate generate` and `migrate run` as usual. This is covered by an integration test on sqlite (old schema + duplicate rows → generated migration removes the duplicates and adds the constraint), and I checked the same flow on Postgres by hand. Also works for plugin translation tables.

2. Dropped the bare `SQLITE_CONSTRAINT` match and narrowed the message check to `unique constraint`. Added tests for both classifiers.

biggamesmallworld (approved):
Both points from my last review are addressed, and CI is green across all four e2e backends.

1. `migrate generate` now injects `deduplicateTranslations()` ahead of the constraint DDL for exactly the tables gaining it, so there is no manual step to discover.
2. `isUniqueConstraintViolationError` no longer matches a bare `SQLITE_CONSTRAINT`, the message fallback is narrowed to `unique constraint`, and both classifiers have unit tests.

Verified locally at `a3d73d92`: `migrate.spec.ts`, `db-errors.spec.ts`, `register-translation-unique-constraints.spec.ts` and `translation-differ.spec.ts` pass, 44 tests. The `deploy` failure is the Vercel fork-secrets issue, not this branch.

Approving. Two non-blocking notes for whenever you next touch this code:

**Logging on the recovery branch.** In `translation-differ.ts`, the losing request adopts the winning row and returns success. If two concurrent requests add the same new language with different content, the loser's content is discarded with no trace. In practice the trigger is a double-click, where the payloads are identical and nothing is lost, so this is not worth another round. But a single `Logger.warn` on that branch, with the entity id and `languageCode`, would make the rare non-identical case greppable rather than invisible. Cheap insurance, since the surviving row looks entirely legitimate afterwards.

**A SQLite quirk in `getTranslationTablesGainingUniqueConstraint`.** It matches on DDL text, and SQLite recreates the whole table for any schema change, carrying the pre-existing `UNIQUE ("languageCode", "baseId")` clause into the emitted `CREATE TABLE "temporary_..."`. So a future SQLite migration that only adds a custom field to a translation table will still get a `deduplicateTranslations()` call prepended. No runtime effect, but the emitted comment about rows left over from before the constraint existed will not be true. Checking whether the table already carries the constraint, rather than regex-matching the statement, would avoid it.

One for the release notes rather than the code: the fix is inert until a user runs `migrate generate` and `migrate run`, so anyone who upgrades without generating a migration keeps getting duplicates and never sees `deduplicateTranslations`.

PR #5010, opened 2026-07-21
docs(dashboard): Document contributing UI translations to the Dashboard

### LINKED ISSUE #3989: Document how to translate the new React Admin Dashboard
I want to translate the UI into a language that is not supported by default. There is no documentation for that. I would like to add this translation 

The only other way is to send a pr and wait until it is merged and new version is released. 

i would like to at least test the translation before sending the pr

grolmus:
## Summary

Adds documentation for translating the built-in React Dashboard UI into a new language, a contributor workflow that was undocumented. Resolves #3989.

## Background

Per Michael's note on the issue, the extension-localization docs already exist (`extending-the-dashboard/localization`), and CONTRIBUTING pointed translation contributors at the **legacy Angular Admin UI** readme. The gap was the **core React Dashboard** contributor workflow: which `.po` catalogs to edit in `packages/dashboard`, how to run the dashboard locally in a target locale to test, and the PR process. That guidance did not exist.

## Change

- **Localization guide** — new "Contributing a translation of the Dashboard itself" section, distinct from the existing extension-localization content it follows:
  - catalogs live at `packages/dashboard/src/i18n/locales/{locale}.po`
  - improving an existing translation (fill in `msgstr`)
  - adding a new language: add the locale to `packages/dashboard/lingui.config.js` and run `npm run i18n:extract`; the loader globs the directory (`import.meta.glob('*.po')`), so nothing else needs wiring
  - testing locally: dev-server, then the user menu → **Display language**
  - opening the PR (a translation is a `fix`)
- **CONTRIBUTING.md** — the translations section now points at both the Dashboard guide and the legacy Admin UI one, instead of only the latter.

## Test plan

- `docs` `test:mdx` — 830 files, 0 errors
- Every claim verified against the source rather than assumed:
  - the loader auto-discovers catalogs via `import.meta.glob('../../i18n/locales/*.po')` (`load-i18n-messages.ts`)
  - `i18n:extract` script exists in `packages/dashboard/package.json`
  - the "Display language" control exists in `language-dialog.tsx`
  - the dev-server serves the dashboard (`DashboardPlugin`, route `dashboard`)
- Docs-only change; no code affected.

Fixes #3989

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5010"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787222283&installation_model_id=20193&pr_number=5010&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5010&signature=f2a4c07ca6b0aa09b615ad33b387bae95c83f865e15325a630031e624756774e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/docs/guides/extending-the-dashboard/localization/index.mdx
+++ b/docs/docs/guides/extending-the-dashboard/localization/index.mdx
@@ -91,3 +91,59 @@ open up the `de.po` file and add German translations for each of the strings, by
 msgid "Welcome to Dashboard"
 msgstr "Willkommen zum Dashboard" # [!code highlight]
 ```
+
+## Contributing a translation of the Dashboard itself
+
+The sections above cover localizing **your own** extensions. If instead you want to translate the
+built-in Dashboard UI into a new language — or improve an existing translation — those catalogs live
+in the Vendure repository and the change is contributed via a pull request.
+
+:::note
+This is for the **React Dashboard** (`@vendure/dashboard`). The legacy Angular Admin UI has its own
+[separate translation workflow](https://github.com/vendurehq/vendure/blob/master/packages/admin-ui/README.md#localization).
+:::
+
+The Dashboard's own message catalogs live at `packages/dashboard/src/i18n/locales/{locale}.po`, one
+`.po` file per supported language. The English catalog (`en.po`) is the source and is always complete;
+the others are translations of it.
+
+### Improving an existing translation
+
+Open the catalog for the language you want to improve — for example `de.po` for German — and fill in
+or correct the `msgstr` values. An empty `msgstr` means that string is not yet translated and falls
+back to English at runtime.
+
+### Adding a new language
+
+1. Add the [locale code](https://lingui.dev/misc/faq#which-locales-and-formats-are-supported) to the
+   `locales` array in `packages/dashboard/lingui.config.js`.
+2. From `packages/dashboard`, generate the catalog:
+
+   ```bash
+   npm run i18n:extract
+   ```
+
+   This creates `src/i18n/locales/{locale}.po` containing every UI string with empty `msgstr` values.
+3. Fill in the `msgstr` values with your translations. There is nothing else to wire up — the
+   Dashboard loads catalogs by globbing that directory, so the new file is picked up automatically.
+
+### Testing your translation locally
+
+Run the development server so you can see your translations in the running Dashboard:
+
+```bash
+cd packages/dev-server
+bun run populate   # first time only, to seed test data
+bun run dev
+```
+
+Open the Dashboard (`http://localhost:3000/dashboard`), then switch the UI language from the user menu
+at the bottom of the sidebar → **Display language**, and choose your locale. The interface re-renders
+with your catalog; any string you left untranslated shows the English source.
+
+### Opening the pull request
+
+Commit the changed `.po` file(s) (and the `lingui.config.js` change, for a new language) and open a PR
+against `master`. A translation change is a `fix`, so title it accordingly, e.g.
+`fix(dashboard): Add Ukrainian UI translations`. Translations are always welcome, including partial
+ones — you don't have to translate every string to contribute.
--- a/docs/docs/guides/extending-the-dashboard/localization/index.mdx
+++ b/docs/docs/guides/extending-the-dashboard/localization/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: 'Localization'
 metaTitle: "Localizing Vendure Dashboard Extensions with Lingui"
-metaDescription: "Add multi-language support to your Vendure Dashboard extensions using Lingui for string extraction, translation catalogs, and ICU MessageFormat."
+metaDescription: "Localize your Vendure Dashboard extensions with Lingui, and contribute UI translations of the Dashboard itself via a pull request."
 ---
 
 :::note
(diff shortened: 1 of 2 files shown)


docs/docs/guides/extending-the-dashboard/localization/index.mdx
```diff
@@ -91,3 +91,59 @@ open up the `de.po` file and add German translations for each of the strings, by
...
+   `locales` array in `packages/dashboard/lingui.config.js`.
+2. From `packages/dashboard`, generate the catalog:
+
+   ```bash
+   npm run i18n:extract
+   ```
+
+   This creates `src/i18n/locales/{locale}.po` containing every UI string with empty `msgstr` values.
+3. Fill in the `msgstr` values with your translations. There is nothing else to wire up — the
+   Dashboard loads catalogs by globbing that directory, so the new file is picked up automatically.
```
biggamesmallworld: `lingui.config.js` alone isn't enough — the locale must also go in `defaultAvailableLanguages` (`packages/dashboard/vite/constants.ts`) or it never appears in the selector. And `i18n:extract` is `lingui extract && node ./scripts/translate/i18n-tool.js extract`; the second half writes a non-gitignored `missing-translations.txt`, right before this guide says to commit.

~~~suggestion
1. Add the locale code to the `locales` array in `packages/dashboard/lingui.config.js`. It must be a
   valid [`LanguageCode`](https://docs.vendure.io/reference/typescript-api/common/language-code/)
   value, which uses underscores for region variants (`zh_Hans`, `pt_BR`).
2. Add the same locale to `defaultAvailableLanguages` in `packages/dashboard/vite/constants.ts`. The
   **Display language** selector is built from this list, so a catalog missing from it can never be
   selected.
3. From `packages/dashboard`, generate the catalog:

   ```bash
   npx lingui extract
   ```

   This creates `src/i18n/locales/{locale}.po` with every UI string and empty `msgstr` values. Use
   `npx lingui extract` rather than `npm run i18n:extract`, which also runs an LLM helper that writes
   an untracked `missing-translations.txt`.
4. Fill in the `msgstr` values. The Dashboard globs that directory, so nothing else needs registering.
~~~

docs/docs/guides/extending-the-dashboard/localization/index.mdx
```diff
@@ -91,3 +91,59 @@ open up the `de.po` file and add German translations for each of the strings, by
...
+
+Run the development server so you can see your translations in the running Dashboard:
+
+```bash
+cd packages/dev-server
+bun run populate   # first time only, to seed test data
+bun run dev
+```
+
+Open the Dashboard (`http://localhost:3000/dashboard`), then switch the UI language from the user menu
```
biggamesmallworld: `bun run dev` runs `dev-workflow.mjs portless`, and `dev-network-config.mjs` sets `VENDURE_SERVE_STATIC_DASHBOARD: 'false'`, serving the Dashboard from Vite at a Portless origin (or `localhost:5173/dashboard` under `dev:direct`). Port 3000 is the static path this workflow turns off.

```suggestion
`bun run dev` prints the Dashboard URL when it starts — open that. It serves from Vite, so `.po` edits
are picked up on reload; the static build served by `DashboardPlugin` needs a rebuild. Then switch the
UI language from the user menu
```

docs/docs/guides/extending-the-dashboard/localization/index.mdx
```diff
@@ -91,3 +91,59 @@ open up the `de.po` file and add German translations for each of the strings, by
...
+at the bottom of the sidebar → **Display language**, and choose your locale. The interface re-renders
+with your catalog; any string you left untranslated shows the English source.
+
+### Opening the pull request
+
+Commit the changed `.po` file(s) (and the `lingui.config.js` change, for a new language) and open a PR
```
biggamesmallworld: Follow-on from the `defaultAvailableLanguages` change above.

```suggestion
Commit the changed `.po` file(s) (and, for a new language, the `lingui.config.js` and
`vite/constants.ts` changes) and open a PR
```

biggamesmallworld (changes requested):
Useful gap to fill, but a few instructions don't hold against the source. Three suggestions inline.

Main one: `lingui.config.js` is not sufficient. The **Display language** selector is built from `uiConfig.i18n.availableLanguages` (`language-dialog.tsx`), which defaults to the hardcoded `defaultAvailableLanguages` in `packages/dashboard/vite/constants.ts`. A catalog missing from that array can never be selected — `ko.po` is in neither list, and `ja` is in `lingui.config.js` with a populated `ja.po` but absent from `defaultAvailableLanguages`, so this failure mode has already happened twice.

Also worth adding:

- **CI gates** a translation PR must pass: `dashboard i18n sync` (`scripts/check-i18n-sync.sh`, fails if catalogs differ after `lingui extract`) and `i18n:check` (per-locale script validation).
- **Consistency**: `npm run i18n:extract` here, `bun run` a few lines later; CONTRIBUTING.md uses `bun run` throughout.

Non-blocking nit: contributor workflow on a page whose `metaDescription` is about Dashboard *extensions* — description is now stale.

grolmus:
Thanks @biggamesmallworld — all applied in `5b9134b7e`:

- **`defaultAvailableLanguages`** — added as an explicit step (the selector is built from `vite/constants.ts`, so a catalog missing there can't be selected). Good catch; that's exactly the `ko`/`ja` gap.
- **`npx lingui extract`** instead of `npm run i18n:extract` — with a note that the latter also runs the `i18n-tool.js` helper that writes an untracked `missing-translations.txt`.
- **Dev URL** — replaced the hardcoded `localhost:3000` with "open the URL `bun run dev` prints"; noted it serves from Vite (`.po` picked up on reload) vs the static `DashboardPlugin` build needing a rebuild.
- **Commit step** — now includes the `vite/constants.ts` change for a new language.
- **CI gates** — added a paragraph on `dashboard i18n sync` (`check-i18n-sync.sh`) and `i18n:check`, with `npx lingui extract` to keep the sync check green.
- **metaDescription** — updated to cover contributing Dashboard translations, not just extensions.

Verified each against the source (`vite/constants.ts`, the `i18n:*` scripts, `check-i18n-sync.sh` in the workflow). Mind another look?

biggamesmallworld:
one nit:  `i18n:check` isn't a part of the CI gate, but that is fairly trivial. Can approve now, consider addressing it

biggamesmallworld (approved):

PR #5014, opened 2026-07-22
fix(job-queue-plugin): Fix BullMQ job list query pagination, ordering & index maintenance

michaelbromley:
## Description

The `jobs` list query in the BullMQ job queue plugin had several defects, mostly in the `getJobsByType` Lua script and the `JobListIndexService` which maintains the indexed sorted sets that the script reads.

### Defects fixed

**In the Lua script / query path:**

- **Jobs were silently dropped from pages.** When the result set spanned both list-stored states (`wait`, `active`, `paused`) and sorted-set-stored states (`completed`, `failed`, etc.) — which is always the case for the unfiltered admin UI view — the `skip` offset was applied twice. In a test with 5 waiting and 30 completed jobs, paging with `take: 10` returned pages of 10, 5, 0 and 0 items, leaving 20 of the 35 jobs unreachable through the API.
- **Ordering was inconsistent.** List-stored jobs were always emitted before sorted-set-stored jobs regardless of creation time, and the raw scores of the `delayed` set (which BullMQ encodes as `timestamp * 4096 + n`) were merged with plain timestamps, so delayed jobs always sorted above everything else.
- **The query crashed on busy queues.** The script passed the entire wait list through Lua's `unpack()`, which is limited to roughly 8000 arguments, so the jobs list errored with `too many results to unpack` exactly when the queue backlog was largest.
- **Repeatable-job configs were counted as jobs.** `ALL_JOB_TYPES` included `'repeat'`, whose Redis key holds repeatable-job configuration entries rather than job ids, inflating `totalItems`.
- **Filtering by `RETRYING` returned nothing.** The filter mapped to the `'repeat'` job type, while jobs awaiting a retry actually live in the `'delayed'` state (which is also how the single-job query reports them).
- **Filtering by multiple queue names used only the first name.** `queueName: { in: [a, b] }` silently dropped every name after the first.

**In the index maintenance:**

- **The index was never updated when a custom Redis prefix was configured.** The `QueueEvents` instance was created without the `prefix` option, so it listened on the default `bull`-prefixed events stream which the queue never writes to. The index was built once at worker startup by the migration and then never updated again.
- **Removed jobs left stale index entries.** By the time the `removed` event is handled, the job data has already been deleted, so the handler's job lookup failed and it returned without removing anything. This made `totalItems` keep counting removed jobs.
- **State transitions could race.** The event handlers ran concurrently and the remove-then-add sequence was not atomic, so a fast `active` → `completed` transition could leave a job indexed under a stale state.

### Approach

- The Lua script is rewritten to be read-only. Since Redis scripts execute atomically, candidates are merged and sorted in Lua memory instead of in temporary Redis keys (the old script performed `ZADD`/`ZUNIONSTORE`/`DEL` writes on every poll of the jobs list). Only the first `skip + take` entries of each state structure are fetched, so cost is bounded by page depth rather than queue length. Creation timestamps are read from the job hashes to give one consistent newest-first ordering across structures whose native scores are not comparable.
- The `JobListIndexService` now passes the configured prefix to `QueueEvents`, performs each index update as a single atomic pipeline, serialises operations per job id to preserve event order, and maintains a small registry set of known queue names so that a removed job's id can be cleared from all indexed sets even though the job data is gone. The `QueueEvents` connection is now also closed when the strategy is destroyed.

### Tests

Two new e2e spec files demonstrate the defects (all of these failed before the fix commits and pass after):

- `get-jobs-by-type-script.e2e-spec.ts` runs the Lua script directly against Redis with data seeded in exactly the layout BullMQ and the index service use, covering pagination completeness, ordering, the delayed-score problem, repeat-config counting and the 9000-job `unpack` crash.
- `job-list.e2e-spec.ts` exercises the full stack through the GraphQL admin API with jobs held in a stable mixture of states (completed, retrying, running, queued behind a blocked worker), covering state mapping, ordering, pagination, the `RETRYING` filter, multi-queue-name filtering, index updates under a custom prefix, and total consistency after job removal.

The stale `e2e-wip` npm script (which pointed at a no-longer-existing jest config) is replaced with a working `e2e` script following the same pattern as the other packages. The tests skip themselves when no Redis instance is available, as before.

### Breaking changes

None. The Lua script's argument layout changed, but it is an internal detail between the strategy and the script it defines.

### Known limitations (unchanged by this PR)

- The script accesses multiple keys derived from a prefix passed via `KEYS[1]`, so it is still not Redis Cluster-safe, though it is now at least read-only.
- The `sort` parameter of `JobListOptions` is still ignored; the list is always ordered newest-first by creation time.
- `findMany` still performs per-job lookups for the returned page; returning states from the script could halve those round trips in a future change.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5014"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787315051&installation_model_id=20193&pr_number=5014&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5014&signature=09d95b4c64a36b679e06e39a7d1c0fefee74a777d84f80f9665f25d9e03cf0aa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/job-queue-plugin/e2e/get-jobs-by-type-script.e2e-spec.ts
+++ b/packages/job-queue-plugin/e2e/get-jobs-by-type-script.e2e-spec.ts
@@ -1,0 +1,279 @@
+import Redis from 'ioredis';
+import { afterAll, beforeEach, describe, expect, it } from 'vitest';
+
+import { ALL_JOB_TYPES } from '../src/bullmq/constants';
+import { getJobsByType } from '../src/bullmq/scripts/get-jobs-by-type';
+
+/**
+ * These tests exercise the `getJobsByType` Lua script directly against a real Redis instance,
+ * with data laid out exactly as BullMQ and the JobListIndexService store it:
+ *
+ * - `<prefix>wait`, `<prefix>active`, `<prefix>paused` are **lists** of job ids, newest at the head
+ * - `<prefix>completed`, `<prefix>failed` are **sorted sets** scored by the finished timestamp
+ * - `<prefix>delayed` is a **sorted set** scored by an encoded value (`ms * 4096 + n`)
+ * - `<prefix>prioritized` is a **sorted set** scored by priority
+ * - `<prefix>repeat` is a **sorted set** of repeatable-job config keys (not job ids!)
+ * - `<prefix><jobId>` is a hash of job data including the `timestamp` (creation time) field
+ * - `<prefix>queue:<queueName>:<state>` are the plugin's own indexed sorted sets,
+ *   scored by creation timestamp
+ *
+ * Each test seeds this full layout and asserts the intended contract of the jobs list:
+ * a paginated, newest-first (by creation time) listing which never drops, duplicates
+ * or invents jobs. Several of these tests currently fail, demonstrating known defects
+ * in the script. They should all pass once the script is fixed.
+ */
+
+const redisHost = '127.0.0.1';
+const redisPort = process.env.CI ? +(process.env.E2E_REDIS_PORT || 6379) : 6379;
+
+const PREFIX = 'script-test:vendure-job-queue:';
+
+interface SeedJob {
+    id: string;
+    queueName: string;
+    state: 'wait' | 'active' | 'paused' | 'completed' | 'failed' | 'delayed' | 'prioritized';
+    createdAt: number;
+    settledAt?: number;
+    /** For delayed jobs: the time at which the job becomes due */
+    delayUntil?: number;
+}
+
+const LIST_STATES = ['wait', 'active', 'paused'];
+
+describe('getJobsByType Lua script', () => {
+    const redis = new Redis({ host: redisHost, port: redisPort, maxRetriesPerRequest: null });
+    redis.defineCommand(getJobsByType.name, {
+        numberOfKeys: getJobsByType.numberOfKeys,
+        lua: getJobsByType.script,
+    });
+
+    function callScript(
+        skip: number,
+        take: number,
+        filterName: string,
+        states: string[] = ALL_JOB_TYPES,
+    ): Promise<[number, string[]]> {
+        return (redis as any).getJobsByType(PREFIX, skip, take, filterName, ...states);
+    }
+
+    async function deleteTestKeys() {
+        let cursor = '0';
+        do {
+            const [nextCursor, keys] = await redis.scan(cursor, 'MATCH', 'script-test:*', 'COUNT', 1000);
+            cursor = nextCursor;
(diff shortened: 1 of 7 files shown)


michaelbromley:
Addressed the review feedback in b427383e6:

**CodeRabbit's major point (unfiltered path could omit newer jobs)** — fixed as suggested. The unfiltered path no longer reads BullMQ's native state structures at all. Instead it resolves the set of known queue names from a registry (a Redis set the JobListIndexService now maintains for the removed-event handling) and runs the same indexed-set merge as the name-filtered path. Both paths are now a single code path, uniformly ordered by creation timestamp, so no state's native ordering (finish time, encoded delay, or priority) can cut a job from a page before the creation-time sort sees it. There is a new script-level test covering the worst case: 30 prioritized jobs whose priority order is the reverse of their creation order, with a page size smaller than the backlog.

**CodeRabbit's minor point (skip + take <= 0)** — fixed. Range reads are guarded so that a zero or negative page size returns the total with an empty item list instead of translating into a full-range read (a range end of -1 means the whole structure to Redis). Also covered by a new test.

Two consequences of the unified approach worth noting for reviewers:

- The jobs list now depends entirely on the worker-maintained index, including for unfiltered queries. To make sure the index converges after a worker has been offline, the startup migration now always merges existing jobs into the indexed sets (the zadds are idempotent) rather than skipping any set that already exists — previously, jobs added while no worker was listening would have stayed invisible forever.
- Index updates now use MULTI/EXEC instead of a plain pipeline, so the remove-from-old-states/add-to-new-state sequence is atomic against other Redis clients as well, which matters for multi-worker deployments.

Also tightened the Lua script's argument typing from a loose `Array<string | number>` back to a proper tuple.

Full package e2e suite (16 tests across 3 files) passes against a live Redis, and `tsc --noEmit` and eslint are clean.

PR #5017, opened 2026-07-22
fix(core): Add channel-scope guard to StockLocation & Asset update()

### LINKED ISSUE #5003: Cross-channel (cross-tenant) authorization bypass in StockLocation and Asset update (incomplete fix of channel-scope enforcement)
reported via email on 13 June 2026 - no response, trying here:

Affected Versions: confirmed on v3.6.4 

### Summary

An Administrator restricted to one channel can modify `StockLocation` (and `Asset`) entities that belong only to another channel, on which they have no permissions. The entity is loaded by raw global id with no channel filter, so the per-channel role isolation that Vendure Channels provide is bypassed. This is an incomplete fix of the channel-scope enforcement change (PR #4821, unreleased), which added `{ channelId: ctx.channelId }` guards to six services but missed `StockLocation` and `Asset`. Confirmed by booting the `@vendure/core` server and driving the Admin GraphQL API: a Channel-B-only admin overwrote a Channel-A stock location.

### Details

Sink: `packages/core/src/service/services/stock-location.service.ts:160`:

```ts
async update(ctx, input) {
    const stockLocation = await this.connection.getEntityOrThrow(ctx, StockLocation, input.id); // NO channelId
    const updatedStockLocation = patchEntity(stockLocation, input);
    await this.connection.getRepository(ctx, StockLocation).save(updatedStockLocation);          // commits globally
```

Sibling with the identical defect: `packages/core/src/service/services/asset.service.ts` `update()` loads `getEntityOrThrow(ctx, Asset, input.id)` with no `channelId`, then `getRepository(ctx, Asset).save(asset)` (Asset is `ChannelAware`; higher impact, catalog images/files).

Flow: Admin GraphQL `updateStockLocation`/`updateAsset` -> resolver `@Allow(Permission.UpdateStockLocation)` / `@Allow(UpdateCatalog, UpdateAsset)`. The AuthGuard (`auth-guard.ts:84` `canAccess`) checks the permission scoped to the active channel selected by the `vendure-token` header, so a Channel-B-only admin passes with the Channel-B token. The service then loads the target by raw global id (`getEntityOrThrowInternal`, `transactional-connection.ts:322-332`, which does a plain `findOne({where:{id}})` across all channels when `channelId` is omitted) and overwrites it.

The boundary crossed is multi-channel tenant isolation (the purpose of Vendure Channels + per-channel Roles). The mutation returns a GraphQL non-null error because the return `findOne()` is channel-scoped (null in Channel B), but the write commits first, a silent successful tamper masked by an error response. PR #4821 added `getEntityOrThrow(ctx, Entity, input.id, { channelId: ctx.channelId })` to PaymentMethod, Facet, FacetValue, Collection, ProductOptionGroup, and ProductOption, but not to StockLocation or Asset.

### PoC

```
# as an admin whose Role grants Update only on Channel B, using the Channel-B token:
mutation { updateStockLocation(input:{ id:"<Channel-A stock location id>", name:"PWNED" }) { id } }
# write commits to the Channel-A entity (response shows a non-null error, but the row is changed)
```

Validated with the official `@vendure/testing` `createTestEnvironment`, which boots the built `@vendure/core` NestJS server (AuthGuard, resolvers, services) over the Admin GraphQL API against sqljs:

```
LOGIN (bob-channelb): channels:[{code:"channel-b", permissions:["Authenticated","ReadStockLocation","UpdateStockLocation"]}]
CHANNEL-B VISIBLE STOCK LOCATIONS: []          (attacker cannot even SEE the Channel-A entity)
UPDATE THREW: Cannot return null for non-nullable field Mutation.updateStockLocation.
CHANNEL-A LOCATION AFTER ATTACK (read as SuperAdmin):
  {"id":"T_2","name":"PWNED-BY-CHANNEL-B","description":"TAMPERED-CROSS-CHANNEL"}
```

The Channel-A StockLocation was renamed/overwritten by an admin with no rights on Channel A.

### Impact

In a multi-channel deployment, any channel-scoped (non-SuperAdmin) admin with Update permission on any one channel can tamper with and corrupt StockLocation and Asset entities belonging to other channels/tenants, defeating channel isolation. Integrity (and availability via data corruption) impact across tenants; only multi-channel deployments with channel-scoped admins are affected.

### Remediation

Add the same guard PR #4821 used to both missed services:

```ts
// stock-location.service.ts update()
await this.connection.getEntityOrThrow(ctx, StockLocation, input.id, { channelId: ctx.channelId });
// asset.service.ts update()
const asset = await this.connection.getEntityOrThrow(ctx, Asset, input.id, { channelId: ctx.channelId });
```

Audit every service `update()`/`save()` that loads by `input.id` without `{ channelId }` for `ChannelAware` entities, and re-check `delete` paths and any other unfixed siblings.

---

Ryrahul:
Fixes #5003

# Description

`StockLocationService.update()` and `AssetService.update()` loaded entities via `getEntityOrThrow()` without passing `{ channelId: ctx.channelId }`, allowing a channel-scoped admin to modify entities belonging to other channels. This is the same class of defect fixed for six other services in PR #4821, which missed these two.

**Changes:**
- Add `{ channelId: ctx.channelId }` guard to `StockLocationService.update()` and `AssetService.update()`
- Add cross-channel update protection e2e tests for StockLocation, Asset, Collection, ProductOptionGroup, ProductOption, and Facet

# Breaking changes

No breaking changes.

# Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5017"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787340217&installation_model_id=20193&pr_number=5017&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5017&signature=385b9c7609dc7b201fcae7ba94a4eabd7fc94541aa73cff5febff369e9e7e7f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/core/e2e/asset-channel.e2e-spec.ts
+++ b/packages/core/e2e/asset-channel.e2e-spec.ts
@@ -16,6 +16,7 @@ import {
     deleteAssetDocument,
     getAssetDocument,
     getProductWithVariantsDocument,
+    updateAssetDocument,
     updateCollectionDocument,
     updateProductDocument,
 } from './graphql/shared-definitions';
@@ -306,5 +307,72 @@ describe('Collection related assets', () => {
             id: collectionFeaturedAssetId,
         });
         expect(asset?.id).toEqual(collectionFeaturedAssetId);
+    });
+});
+
+// Cross-channel update protection
+// Verifies that a channel-scoped admin cannot update an Asset belonging to
+// another channel. This is the guard that should use { channelId: ctx.channelId }
+// in asset.service.ts update().
+describe('cross-channel update protection', () => {
+    const CHANNEL_A_TOKEN = 'asset-cross-channel-a';
+    const CHANNEL_B_TOKEN = 'asset-cross-channel-b';
+    let targetAssetId: string;
+
+    beforeAll(async () => {
+        await adminClient.asSuperAdmin();
+        adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+        // Create two isolated channels
+        for (const [code, token] of [
+            ['asset-cross-a', CHANNEL_A_TOKEN],
+            ['asset-cross-b', CHANNEL_B_TOKEN],
+        ]) {
+            await adminClient.query(createChannelDocument, {
+                input: {
+                    code,
+                    token,
+                    defaultLanguageCode: LanguageCode.en,
+                    currencyCode: CurrencyCode.GBP,
+                    pricesIncludeTax: true,
+                    defaultShippingZoneId: 'T_1',
+                    defaultTaxZoneId: 'T_1',
+                },
+            });
+        }
+        // Create an Asset in Channel A via file upload
+        adminClient.setChannelToken(CHANNEL_A_TOKEN);
+        const filesToUpload = [path.join(__dirname, 'fixtures/assets/pps2.jpg')];
+        const { createAssets } = await adminClient.fileUploadMutation({
+            mutation: createAssetsDocument,
+            filePaths: filesToUpload,
+            mapVariables: filePaths => ({
+                input: filePaths.map(() => ({ file: null })),
+            }),
+        });
+        targetAssetId = createAssets[0].id;
+    });
+
+    afterAll(() => {
+        adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+    });
+
+    it('cannot update an Asset belonging to another channel', async () => {
+        // Channel B does not contain this Asset; the update must be rejected.
+        adminClient.setChannelToken(CHANNEL_B_TOKEN);
+        await expect(
+            adminClient.query(updateAssetDocument, {
+                input: {
+                    id: targetAssetId,
+                    name: 'PWNED-BY-CHANNEL-B',
+                },
+            }),
+        ).rejects.toThrow(/No Asset with the id .* could be found/);
+
+        // Verify the original Asset in Channel A is completely unchanged.
+        adminClient.setChannelToken(CHANNEL_A_TOKEN);
+        const { asset } = await adminClient.query(getAssetDocument, {
+            id: targetAssetId,
+        });
+        expect(asset?.name).toBe('pps2.jpg');
     });
 });
--- a/packages/core/e2e/stock-location.e2e-spec.ts
+++ b/packages/core/e2e/stock-location.e2e-spec.ts
@@ -360,5 +360,76 @@ describe('Stock location', () => {
 
             expect(productVariant?.stockLevels.length).toBe(0);
         });
+
+        // Cross-channel update protection
+        // Verifies that a channel-scoped admin cannot update a StockLocation belonging
+        // to another channel. This is the guard added by the channelId fix.
+        describe('cross-channel update protection', () => {
+            const CHANNEL_A_TOKEN = 'stock-loc-cross-channel-a';
+            const CHANNEL_B_TOKEN = 'stock-loc-cross-channel-b';
+            let targetLocationId: string;
+
+            beforeAll(async () => {
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                // Create two isolated channels
+                for (const [code, token] of [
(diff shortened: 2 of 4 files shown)


packages/core/e2e/asset-channel.e2e-spec.ts
```diff
@@ -308,3 +309,70 @@ describe('Collection related assets', () => {
...
+                },
+            }),
+        ).rejects.toThrow(/No Asset with the id .* could be found/);
+
+        // Verify the original Asset in Channel A is completely unchanged.
+        adminClient.setChannelToken(CHANNEL_A_TOKEN);
+        const { asset } = await adminClient.query(getAssetDocument, {
+            id: targetAssetId,
+        });
+        expect(asset?.name).not.toBe('PWNED-BY-CHANNEL-B');
```
biggamesmallworld: Negative assertion — this passes even if the name got corrupted some other way. Assert the real value, like the other five tests in this PR do.

```suggestion
        expect(asset?.name).toBe('pps2.jpg');
```

biggamesmallworld (approved):
Fix is correct and matches the pattern from #4821. I checked that both entities are `ChannelAware` and that both `create()` paths go through `channelService.assignToCurrentChannel`, which assigns the default channel alongside `ctx.channel` — so the stricter read doesn't break a default-channel admin updating a sub-channel entity. `findOneInChannel` uses `leftJoin` rather than `leftJoinAndSelect`, so `entity.channels` stays undefined and the following `save()` won't clobber the relation.

I also swept the rest of `service/services/` for the same defect. Every remaining `update()` that calls `getEntityOrThrow` without `channelId` (`tag`, `zone`, `seller`, `customer-group`) is on a non-ChannelAware entity, so this closes the class for `update()`.

Two notes:

**Extraneous tests.** Four of the six new suites — Collection, ProductOptionGroup, ProductOption and Facet — cover guards that already shipped in #4821 and aren't touched by this PR. They're not wrong, but that's ~250 lines and six extra channels of e2e setup for an eight-line change. I'd prefer this PR stay scoped to StockLocation and Asset. Not blocking — if you'd rather not churn it, leave them.

**Delete paths.** #5003 also asked to re-check deletes. The same shape is still there on ChannelAware entities in `facet-value.service.ts:228`, `product-option.service.ts:146` and `promotion.service.ts:196`. Out of scope here; I'll open a separate issue for it.

The only CI failure is `Deploy Dashboard`, which is unrelated infra.

Thanks for the detailed repro on the issue — made this easy to verify.

Ryrahul:
Thanks @biggamesmallworld for the detailed review!

Fixed the negative assertion — now asserts the actual filename (`pps2.jpg`). Also removed the extraneous Collection, ProductOptionGroup, ProductOption, and Facet test suites to keep the PR scoped to the actual fix.

Happy to pick up the delete paths as well — it's the same one-line `{ channelId: ctx.channelId }` change on `facet-value.service.ts:228`, `product-option.service.ts:146`, and `promotion.service.ts:196`. Want me to add them to this PR or open a follow-up?

Also should the cross-channel update protection tests for Collection, ProductOptionGroup, ProductOption, and Facet go into that follow-up PR? They cover the existing #4821 guards but have no regression tests today.

biggamesmallworld:
I think a follow up is appropriate, let's keep this small. A follow up to encapsulate the delete paths and regression tests works for me. Thanks!

biggamesmallworld (approved):

PR #5023, opened 2026-07-23
feat(core): Route ManyToOne custom property filters via EXISTS subqueries

RubenDarioGuerreroNeira:
Routes ManyToOne custom property filters to use EXISTS subqueries, improving performance and skipping unnecessary joins.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5023"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787420708&installation_model_id=20193&pr_number=5023&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5023&signature=59107dfeb5fddfd1171f8cdb29d9e26d6c9bdf4a2fe59d143b724ee30f9fef1a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/service/helpers/list-query-builder/parse-filter-params.ts
+++ b/packages/core/src/service/helpers/list-query-builder/parse-filter-params.ts
@@ -215,7 +215,7 @@ function getToManyRelationCustomProperties<T extends VendureEntity>(
         const relationName = pathParts[0];
         const relationMetadata = metadata.findRelationWithPropertyPath(relationName);
 
-        if (relationMetadata && (relationMetadata.isOneToMany || relationMetadata.isManyToMany)) {
+        if (relationMetadata && (relationMetadata.isOneToMany || relationMetadata.isManyToMany || relationMetadata.isManyToOne)) {
             toManyProperties.add(property);
         }
     }


michaelbromley:
Thanks for this, and for the follow-up work in #5034.

Closing this one as superseded. #5034 covers the same code path and takes the opposite position on the substantive line here: this PR adds `isManyToOne` to the EXISTS-eligible relation types, whereas #5034 documents that ManyToOne relations should stay on a plain `WHERE` clause and adds e2e coverage proving it. Since the two conflict, it makes sense to carry the discussion forward in the tested one.

For what it is worth, tracing `buildExistsSubquery` suggests this change is likely a no-op for the generated SQL in any case — only the `isManyToMany` and `isOneToMany` branches build a real EXISTS clause, and everything else falls through to `return null`, after which `compileFilter` uses the same standard `andWhere`/`orWhere` that was built regardless of the `isExistsCondition` flag. So the join it aims to skip is not actually being skipped.

No issue with the work itself, it is purely that #5034 is where this should land. Happy to reopen if I have misread the relationship between the two.

PR #5027, opened 2026-07-24
fix(dashboard): optional transaction ID for manual payment

knoid:
# Description

When adding a new payment to an order, the dialog has transactionId field marked as required but in the GraphQL type it represents it's optional. This change reflects that by removing the "required" condition set forthis specific field.

# Breaking changes

No breaking changes.

# Screenshots

This is the dialog being changed.
image

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5027"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787450439&installation_model_id=20193&pr_number=5027&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5027&signature=fdc3a713d23db349eebb327f0191ed400669109e3ecb8a747b9896547c6eb7b2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


HouseinIsProgramming (changes requested):
Thanks, the core change is right, the server never required this field.

Requested changes:

1. Send `data.transactionId || undefined` in `add-manual-payment-dialog.tsx` instead of the raw form value.
The form default is `''`, which currently gets stored as an empty string in a nullable column.
2. Replace the toast assertion with an order query that checks the payment was actually added.
The failure toast is titled "Failed to add payment", which doesn't match `/error/i`, so the current assertion also passes when the mutation fails.
3. Drop `packages/dev-server/graphql/graphql-env.d.ts` from the branch.
It's local codegen churn unrelated to this change.

Will approve once that's in.

knoid:
@HouseinIsProgramming I've addressed your comments, please re-check!

HouseinIsProgramming (approved):
All three points addressed: blank submits now send `undefined`, the test asserts the payment via the Admin API, and the codegen churn is reverted. Approving.

PR #5034, opened 2026-07-24
feat(core): Add WHERE-based filtering for ManyToOne custom-property filters

RubenDarioGuerreroNeira:
## Summary of changes

- Clarifies the behavior of custom-property filtering for ManyToOne relations.
- Keeps ManyToOne filters on standard JOIN + WHERE rather than routing them through EXISTS.
- Tightens the EXISTS subquery builder to throw on unsupported paths or relation types instead of silently returning `null`.

## Related issue

- None

## Breaking changes

- None

## Checklist

- [x] I have updated the PR title to reflect the actual change.
- [x] I have provided a clear description using the Vendure PR template.
- [x] I have verified the fix aligns with the intended behavior of ManyToOne filtering.


RubenDarioGuerreroNeira:
Quick note on the direction change here: the original approach tried 
to route ManyToOne custom-property filters through EXISTS subqueries. 
CodeRabbit correctly flagged that `buildExistsSubquery()` never had a 
ManyToOne branch implemented, so those filters were silently falling 
back to `null` — meaning the EXISTS routing wasn't actually happening 
despite being marked as supported.

Rather than build out a new EXISTS branch for ManyToOne (which 
@grolmus separately flagged as a high-blast-radius change affecting 
`orders` and `taxRates`, not just collections), this PR keeps 
ManyToOne filters on the existing JOIN + WHERE path and tightens the 
EXISTS builder to throw instead of silently no-op on unsupported 
paths. That's the reason for the title change from "Route through 
EXISTS" to "WHERE-based filtering."

This is the filter-only half of what was previously in #4636 (now 
closed) — batch loading is split out in #5035.

biggamesmallworld:
Hi @RubenDarioGuerreroNeira thanks for the attention here.

On master today, getToManyRelationCustomProperties already gates on relationMetadata.isOneToMany || relationMetadata.isManyToMany, so a ManyToOne path like parent.label never becomes an EXISTS candidate and never reaches buildExistsSubquery. The silent return null CodeRabbit flagged was reachable only via the guard change that was in https://github.com/vendurehq/vendure/pull/4636. With that change gone, master already does what this PR describes.

So I think the split resolved to one PR rather than two, and there's nothing left here to merge. I will close this in the meantime. I believe PR https://github.com/vendurehq/vendure/pull/5035 has a comment for you as well 👍

PR #5036, opened 2026-07-25
feat(dashboard): Support multi-channel selection in assign-to-channel follow-up

### LINKED ISSUE #4812: Feature Request: Multi-Channel Support for Existing Channel Bulk Actions
## Is your feature request related to a problem? Please describe.

The Dashboard bulk actions for entities like Products, Collections, Facets, Shipping Methods, Payment Methods, Promotions, etc. currently support assigning/removing items to/from a **single channel at a time** via the **"Assign to channel"** / **"Remove from channel"** bulk actions.

In multi-channel setups (marketplaces, regional storefronts, B2B/B2C splits), it's common to need the same set of products, collections, and other entities available across many channels.

Today the only way to do this in the UI is:

1. Select the items
2. Click **Assign to channel**
3. Pick one channel
4. Repeat the entire flow *N* times for *N* channels

This is tedious, error-prone, and doesn't scale well.

We've ended up writing a custom plugin and custom bulk action simply to expose a multi-channel version of this workflow, which suggests there may be a gap worth addressing in core.

---

## Describe the solution you'd like

Extend the existing channel-related bulk actions (at minimum for Products, but ideally for every entity that already supports `assignToChannel` / `removeFromChannel`) so the channel selector supports **multiple channel selection** in a single action.

#### Assign to Channel

* The existing **Assign to channel** dialog should allow selecting multiple channels.
* All selected items should be assigned to all selected channels in a single operation.

#### Remove from Channel

* The existing **Remove from channel** dialog should allow selecting multiple channels.
* All selected items should be removed from all selected channels in a single operation.

#### Option A: API-level support

Extend mutations to accept multiple channel IDs:

```graphql
assignProductsToChannel(
  input: {
    productIds: [ID!]!
    channelIds: [ID!]!
  }
)
```

#### Option B: UI-level support

Keep the existing API unchanged and have the Admin UI:

* Accept multiple channel selections
* Fan out requests in parallel (one request per channel)
* Aggregate success and error reporting for the user

---

## Affected entities

This would be useful for any entity that currently exposes channel assignment bulk actions, including:

* Products
* Collections
* Facets
* Shipping Methods
* Payment Methods
* Promotions
* etc
* Any other channel-aware entity supporting bulk channel assignment

---

### Custom plugin

Register a custom `BulkAction` that opens a multi-select channel dialog and calls `assignProductsToChannel` once per selected channel.

This works, but every team using Vendure must reinvent the same functionality.

### Admin API scripts

Automate channel assignment through scripts.

While effective, this is not accessible to non-technical administrators.

### Repeating the existing workflow

Use the current bulk action multiple times, once per channel.

This becomes increasingly slow and error-prone as the number of channels grows.

---

## Additional context

* This is a **feature request**, not a bug report.
* Current behavior matches the documented single-channel design.
* We're already shipping this capability in production via a custom plugin.
* I'd be happy to contribute a PR if the maintainers agree on the preferred UX:

  * Extend the existing action with multi-select support, or
  * Introduce a separate **"Assign to multiple channels..."** bulk action.

---

## Environment

* **@vendure/core version:**3.6.4
* **Node.js version:** 22.22.3
* **Database:** postgres
* **OS:** macOS

* **Package manager:** pnpm

Ryrahul:
# Description

Extends the "Assign to channel" bulk action dialog to support selecting **multiple channels** in a single operation, replacing the previous single-select dropdown.

Fixes #4812

## What changed

The `AssignToChannelDialog` shared component (used by all 9 entity types with channel assignment bulk actions) now uses the existing `MultiSelect` component instead of the single-select `Select` dropdown. When the user selects multiple channels, the existing per-channel mutation is fanned out in parallel via `Promise.allSettled` inside `useMutation`, following the same pattern as `cancel-jobs-bulk-action.tsx`.

**Affected entities (zero per-entity changes needed):**
Products, Product Variants, Collections, Facets, Promotions, Shipping Methods, Payment Methods, Stock Locations, Product Option Groups

## Why only "Assign to channel" and not "Remove from channel"

The "Remove from channel" action has a fundamentally different UX: it currently has no dialog — it removes from the **active channel** with an inline confirmation. Supporting multi-channel removal would require:

- A new `RemoveFromChannelDialog` component (~80 LOC)
- Changing the `buildInput` signature from `() => Record` to `(channelId: string) => Record` (breaking change to all 9 entity callers)
- Updating all 9 per-entity bulk action files
- Handling 2 callers (facets, option-groups) with custom `onSuccess` result inspection

This is a ~4x larger scope change. It can be done as a follow-up if there's interest, but the assign side covers the primary pain point described in #4812 (assigning entities across many channels).

## Implementation approach

**Option B from the issue (UI-level fan-out)** was chosen over Option A (API-level `channelIds` array) because:

- **Zero API/GraphQL schema changes** — no breaking changes for plugins or external consumers
- **Simpler implementation** — 1 file changed vs 18+ input types, 10 resolvers, 10 services

## Toast behavior

- **All succeed**: `"Successfully assigned N entityType to M channel(s)"`
- **All fail**: `"Failed to assign N entityType to M channel(s)"`
- **Partial**: `"Assigned N entityType to M channel(s), failed for F"` (uses `toast.warning`, matching the existing pattern in `use-refund-order.ts`)

# Breaking changes

None. The `AssignToChannelDialogProps` interface is unchanged. The `buildInput(channelId, additionalData)` callback signature is unchanged — the dialog calls it once per selected channel instead of once. All existing callers (9 bulk actions + `AssignedChannels` detail page) work without modification.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5036"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787599895&installation_model_id=20193&pr_number=5036&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5036&signature=e0a8fd91dd8536d5c132dcd6d5d24e3d991633a37bfd8dde8c0bad96601e3402"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
+++ b/packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
@@ -1,8 +1,10 @@
-import { toast } from '@/vdb/components/ui/sonner.js';
+import { useMutation } from '@tanstack/react-query';
 import { ReactNode, useEffect, useState } from 'react';
+import { toast } from '@/vdb/components/ui/sonner.js';
 
 import { ChannelCodeLabel } from '@/vdb/components/shared/channel-code-label.js';
 import { MultiSelect } from '@/vdb/components/shared/multi-select.js';
+import { assignToChannels } from '@/vdb/components/shared/assign-to-channels.js';
 import { Button } from '@/vdb/components/ui/button.js';
 import {
     Dialog,
@@ -14,6 +16,7 @@ import {
 } from '@/vdb/components/ui/dialog.js';
 import { Input } from '@/vdb/components/ui/input.js';
 import { ResultOf } from '@/vdb/graphql/graphql.js';
+import { plural } from '@lingui/core/macro';
 import { Trans, useLingui } from '@lingui/react/macro';
 
 import { useChannel } from '@/vdb/hooks/use-channel.js';
@@ -55,7 +58,6 @@ export function AssignToChannelDialog({
 }: Readonly<AssignToChannelDialogProps>) {
     const { t } = useLingui();
     const [selectedChannelIds, setSelectedChannelIds] = useState<string[]>([]);
-    const [isAssigning, setIsAssigning] = useState(false);
     const { channels, activeChannel } = useChannel();
     const entityIdsLength = entityIds.length;
 
@@ -63,7 +65,7 @@ export function AssignToChannelDialog({
         if (!open) setSelectedChannelIds([]);
     }, [open]);
 
-    // Filter out the currently selected channel from available options
+    // Filter out the currently active channel from available options
     const availableChannels = channels.filter(channel => channel.id !== activeChannel?.id);
 
     const selectItems = availableChannels.map(ch => ({
@@ -72,42 +74,64 @@ export function AssignToChannelDialog({
         display: <ChannelCodeLabel code={ch.code} />,
     }));
 
-    const handleAssign = async () => {
-        setIsAssigning(true);
-        try {
-            const results = await Promise.allSettled(
-                selectedChannelIds.map(channelId =>
-                    mutationFn({ input: buildInput(channelId, additionalData) }),
-                ),
-            );
+    const { mutate, isPending } = useMutation({
+        mutationFn: () =>
+            assignToChannels(
+                selectedChannelIds,
+                channelId => buildInput(channelId, additionalData),
+                mutationFn,
+            ),
+        onSuccess: ({ succeeded, failed }) => {
+            if (failed.length) {
+                console.error('Failed to assign to channels', failed);
+            }
 
-            const failedChannelIds = results.flatMap((r, i) =>
-                r.status === 'rejected' ? [selectedChannelIds[i]] : [],
-            );
-            const succeededCount = results.length - failedChannelIds.length;
+            const failedCodes = failed
+                .map(f => channels.find(c => c.id === f.channelId)?.code ?? f.channelId)
+                .join(', ');
 
-            if (succeededCount > 0) {
+            if (failed.length === 0) {
+                toast.success(
+                    plural(succeeded.length, {
+                        one: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channel`,
+                        other: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channels`,
+                    }),
+                );
+                setSelectedChannelIds([]);
+                onSuccess?.();
+                onOpenChange(false);
+            } else if (succeeded.length === 0) {
+                toast.error(
+                    plural(failed.length, {
+                        one: `Failed to assign ${entityIdsLength} ${entityType} to ${failed.length} channel`,
+                        other: `Failed to assign ${entityIdsLength} ${entityType} to ${failed.length} channels`,
+                    }),
+                    { description: failedCodes },
+                );
+            } else {
+                toast.warning(
+                    plural(succeeded.length, {
+                        one: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channel`,
+                        other: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channels`,
+                    }),
+                    { description: t`Failed for: ${failedCodes}` },
+                );
+                // Keep the dialog open and narrow the selection to the failures, so retrying
+                // is one click and doesn't re-assign the channels that already succeeded.
+                setSelectedChannelIds(failed.map(f => f.channelId));
                 onSuccess?.();
             }
+        },
+        onError: () => {
+            toast.error(t`Failed to assign ${entityIdsLength} ${entityType} to the selected channels`);
+        },
+    });
 
-            if (failedChannelIds.length === 0) {
-                toast.success(
-                    t`Successfully assigned ${entityIdsLength} ${entityType} to ${selectedChannelIds.length} channels`,
-                );
-                onOpenChange(false);
-            } else {
-                const firstRejected = results.find(r => r.status === 'rejected');
-                const firstReason = firstRejected?.status === 'rejected' ? firstRejected.reason : undefined;
-                const description = firstReason instanceof Error ? firstReason.message : undefined;
-                setSelectedChannelIds(failedChannelIds);
-                toast.error(
-                    t`Failed to assign ${entityIdsLength} ${entityType} to ${failedChannelIds.length} of ${selectedChannelIds.length} channels`,
-                    { description },
-                );
-            }
-        } finally {
-            setIsAssigning(false);
+    const handleAssign = () => {
+        if (selectedChannelIds.length === 0) {
+            return;
         }
+        mutate();
     };
 
     return (
@@ -135,6 +159,7 @@ export function AssignToChannelDialog({
                             onChange={setSelectedChannelIds}
                             placeholder={t`Select one or more channels`}
                             searchPlaceholder={t`Search channels...`}
+                            showSearch
                             className="w-full"
                         />
                     </div>
@@ -144,7 +169,7 @@ export function AssignToChannelDialog({
                     <Button variant="outline" onClick={() => onOpenChange(false)}>
                         <Trans>Cancel</Trans>
                     </Button>
-                    <Button onClick={handleAssign} disabled={selectedChannelIds.length === 0 || isAssigning}>
+                    <Button onClick={handleAssign} disabled={selectedChannelIds.length === 0 || isPending}>
                         <Trans>Assign</Trans>
                     </Button>
                 </DialogFooter>
--- a/packages/dashboard/src/i18n/locales/ar.po
+++ b/packages/dashboard/src/i18n/locales/ar.po
@@ -541,6 +541,14 @@ msgstr "إزالة الرابط"
 msgid "This field is readonly"
 msgstr "هذا الحقل للقراءة فقط"
 
+#. placeholder {0}: succeeded.length
+#. placeholder {1}: succeeded.length
+#. placeholder {2}: succeeded.length
+#: src/lib/components/shared/assign-to-channel-dialog.tsx
+#: src/lib/components/shared/assign-to-channel-dialog.tsx
+msgid "{0, plural, one {Assigned {entityIdsLength} {entityType} to {1} channel} other {Assigned {entityIdsLength} {entityType} to {2} channels}}"
+msgstr ""
+
 #: src/lib/framework/dashboard-widget/metrics-widget/index.tsx
 #: src/lib/framework/dashboard-widget/metrics-widget/index.tsx
 msgid "Order Count"
@@ -2658,6 +2666,10 @@ msgstr "عرض البيانات"
 msgid "Delete View"
 msgstr "حذف العرض"
 
+#: src/lib/components/shared/assign-to-channel-dialog.tsx
+msgid "Failed to assign {entityIdsLength} {entityType} to the selected channels"
+msgstr ""
+
 #: src/app/routes/_authenticated/_customers/customers_.$id.tsx
 msgid "Add a new address to the customer."
 msgstr "أضف عنوانًا جديدًا للعميل."
@@ -3068,10 +3080,9 @@ msgstr "حفظ التغييرات"
 msgid "Address updated"
 msgstr "تم تحديث العنوان"
 
-#. placeholder {0}: selectedChannelIds.length
 #: src/lib/components/shared/assign-to-channel-dialog.tsx
-msgid "Successfully assigned {entityIdsLength} {entityType} to {0} channels"
-msgstr ""
+#~ msgid "Successfully assigned {entityIdsLength} {entityType} to {0} channels"
+#~ msgstr ""
 
 #: src/app/routes/_authenticated/_orders/hooks/use-refund-order.ts
 msgid "Refund processed successfully"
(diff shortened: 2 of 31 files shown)


Ryrahul:
**Note on Option A (as mentioned in #4812):** The foundation for API-level multi-channel support already exists but is currently unused. `ChannelService.assignToChannels()` already accepts `channelIds: ID[]` (an array), and TypeORM's `relation.add(newChannelIds)` can insert multiple junction rows in a single query. However, all 20 call sites across the entire codebase always pass a single-element array `[input.channelId]` because the GraphQL input types only accept `channelId: ID!`. Option A would unlock this unused capability and reduce DB operations, but requires schema changes across 18 input types a much larger, breaking change better suited for a `major` branch. **Would love @grolmus  or @michaelbromley  feedback on whether Option A is worth pursuing closing this PR, and whether multi-channel "Remove from channel" bulk actions are also desired.**

packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
```diff
@@ -13,7 +14,6 @@ import {
     DialogTitle,
 } from '@/vdb/components/ui/dialog.js';
 import { Input } from '@/vdb/components/ui/input.js';
-import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/vdb/components/ui/select.js';
 import { ResultOf } from '@/vdb/graphql/graphql.js';
 import { Trans, useLingui } from '@lingui/react/macro';
```
biggamesmallworld, line 20: Needs `plural` for the toast messages below.

```suggestion
import { plural } from '@lingui/core/macro';
import { Trans, useLingui } from '@lingui/react/macro';
```

packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
```diff
@@ -55,33 +55,56 @@ export function AssignToChannelDialog({
...
+                    t`Assigned ${entityIdsLength} ${entityType} to ${succeeded} channel(s), failed for ${failed}`,
+                );
+            }
+
+            if (succeeded > 0) {
+                onSuccess?.();
+                onOpenChange(false);
+            }
         },
     });
```
biggamesmallworld, line 128: Five issues here, all the same root cause — the fan-out knows which channels failed and then forgets. One suggestion since they overlap:

1. **`channel(s)` isn't pluralization.** This repo ships `ar.po`, `ru.po`, `uk.po`, `pl.po` — 3-6 plural forms each. A translator receiving `"to {succeeded} channel(s)"` has no plural category to work with. `cancel-jobs-bulk-action.tsx` (which the PR description cites) uses `plural()` from `@lingui/core/macro` for exactly this, and note the comment in that file: don't nest `t` inside the arms, it breaks extraction. See also #4870.
2. **`failed for ${failed}`** — failed for 3 *what*? The noun is missing. `use-refund-order.ts` puts a short title in the toast and the detail in `description`; same shape here.
3. **Rejected reasons are discarded.** Every `reason` is thrown away, so a partial failure gives the admin a count and nothing else — no channel codes, no error, no console trace. Keep them and name the failed channels.
4. **`onError` was removed.** `Promise.allSettled` never rejects, but if `buildInput` throws synchronously inside the `.map()` the whole `mutationFn` rejects — and with no `onError` the user gets no toast at all, just an un-disabled button and an open dialog.
5. **Partial failure closes the dialog.** The user can't see or retry the channels that failed without re-running against the ones that already succeeded. Keep it open and narrow the selection to the failures.

Also switched to sequential. `cancel-jobs` can fan out because cancelling N jobs touches N disjoint rows; here N mutations converge on the *same* entity rows, so parallel buys lock contention plus N overlapping reindex jobs. A bulk admin action has no latency budget worth defending.

```suggestion
    const { mutate, isPending } = useMutation({
        mutationFn: async () => {
            // Sequential: each mutation writes channel relations for the same entity rows,
            // so a parallel fan-out only buys lock contention and N overlapping reindex jobs.
            const succeeded: string[] = [];
            const failed: Array<{ channelId: string; reason: unknown }> = [];
            for (const channelId of selectedChannelIds) {
                try {
                    await mutationFn({ input: buildInput(channelId, additionalData) });
                    succeeded.push(channelId);
                } catch (reason) {
                    failed.push({ channelId, reason });
                }
            }
            return { succeeded, failed };
        },
        onSuccess: ({ succeeded, failed }) => {
            if (failed.length) {
                console.error('Failed to assign to channels', failed);
            }
            const failedCodes = failed
                .map(f => channels.find(c => c.id === f.channelId)?.code ?? f.channelId)
                .join(', ');

            if (failed.length === 0) {
                toast.success(
                    plural(succeeded.length, {
                        one: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channel`,
                        other: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channels`,
                    }),
                );
                setSelectedChannelIds([]);
                onSuccess?.();
                onOpenChange(false);
            } else if (succeeded.length === 0) {
                toast.error(
                    plural(failed.length, {
                        one: `Failed to assign ${entityIdsLength} ${entityType} to ${failed.length} channel`,
                        other: `Failed to assign ${entityIdsLength} ${entityType} to ${failed.length} channels`,
                    }),
                    { description: failedCodes },
                );
            } else {
                toast.warning(
                    plural(succeeded.length, {
                        one: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channel`,
                        other: `Assigned ${entityIdsLength} ${entityType} to ${succeeded.length} channels`,
                    }),
                    { description: t`Failed for: ${failedCodes}` },
                );
                // Keep the dialog open and narrow the selection to the failures, so retrying
                // is one click and doesn't re-assign the channels that already succeeded.
                setSelectedChannelIds(failed.map(f => f.channelId));
                onSuccess?.();
            }
        },
        onError: () => {
            toast.error(t`Failed to assign ${entityIdsLength} ${entityType} to the selected channels`);
        },
    });
```

packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
```diff
@@ -93,39 +116,35 @@ export function AssignToChannelDialog({
...
+                            multiple={true}
+                            value={selectedChannelIds}
+                            onChange={setSelectedChannelIds}
+                            items={availableChannels.map(ch => ({
+                                value: ch.id,
+                                label: ch.code,
+                                display: <ChannelCodeLabel code={ch.code} />,
+                            }))}
+                            placeholder={t`Select channels`}
+                            searchPlaceholder={t`Search channels...`}
```
biggamesmallworld, line 161: `MultiSelect` only renders the search box when `showSearch === true || items.length > 10`, so on an install with 10 or fewer channels this placeholder is dead config.

```suggestion
                            searchPlaceholder={t`Search channels...`}
                            showSearch
```

packages/dashboard/src/lib/components/shared/assign-to-channel-dialog.tsx
```diff
@@ -55,33 +55,56 @@ export function AssignToChannelDialog({
...
 }: Readonly<AssignToChannelDialogProps>) {
     const { t } = useLingui();
-    const [selectedChannelId, setSelectedChannelId] = useState<string>('');
+    const [selectedChannelIds, setSelectedChannelIds] = useState<string[]>([]);
     const { channels, activeChannel } = useChannel();
     const entityIdsLength = entityIds.length;
 
-    // Filter out the currently selected channel from available options
+    // Filter out the currently active channel from available options
     const availableChannels = channels.filter(channel => channel.id !== activeChannel?.id);
```
biggamesmallworld, line 69: Good catch fixing the stale `currently selected` → `currently active` comment.

Pre-existing, but multi-select makes it sharper: channels the entity already belongs to are still offered. On the `AssignedChannels` detail page a user can now tick a batch that includes channels already assigned. Worth filtering out, or deciding explicitly not to — but not worth blocking on.

biggamesmallworld (changes requested):
Thanks for this — reworking the one shared dialog so all nine entity types get multi-select for free, with `buildInput` and the props unchanged, is the right call. Both callers (`assign-to-channel-bulk-action.tsx`, `assigned-channels.tsx`) work untouched, which I've verified.

The error handling needs work before merge. Inline suggestions cover it; two things that aren't suggestable:

**Reset selection on close.** `selectedChannelIds` is never cleared, so reopening the dialog pre-checks the previous selection. Line 111 is outside the diff so I can't suggest it — please change:

```tsx
<Dialog open={open} onOpenChange={next => { if (!next) setSelectedChannelIds([]); onOpenChange(next); }}>
```

**Tests.** There's currently zero coverage for this component (no e2e spec, no unit spec, no test references anywhere for assign-to-channel), and this PR adds a three-way branch — all-succeed / all-fail / partial — where the partial path is the one nobody exercises manually.

The dashboard has no `@testing-library/react`, so a render test would mean the `renderToStaticMarkup` + `vi.mock` harness from `page-layout.spec.tsx`, which can't click anything. Cheaper and better: pull the loop out of the component and test it as plain logic.

New file `packages/dashboard/src/lib/components/shared/assign-to-channels.ts`:

```ts
export interface AssignToChannelsResult {
    succeeded: string[];
    failed: Array<{ channelId: string; reason: unknown }>;
}

export async function assignToChannels(
    channelIds: string[],
    buildInput: (channelId: string) => Record<string, any>,
    mutationFn: (variables: any) => Promise<unknown>,
): Promise<AssignToChannelsResult> {
    const result: AssignToChannelsResult = { succeeded: [], failed: [] };
    for (const channelId of channelIds) {
        try {
            await mutationFn({ input: buildInput(channelId) });
            result.succeeded.push(channelId);
        } catch (reason) {
            result.failed.push({ channelId, reason });
        }
    }
    return result;
}
```

New file `packages/dashboard/src/lib/components/shared/assign-to-channels.spec.ts`:

```ts
import { describe, expect, it, vi } from 'vitest';

import { assignToChannels } from './assign-to-channels.js';

const buildInput = (channelId: string) => ({ channelId });

describe('assignToChannels', () => {
    it('reports all successes', async () => {
        const result = await assignToChannels(['a', 'b'], buildInput, vi.fn().mockResolvedValue({}));
        expect(result).toEqual({ succeeded: ['a', 'b'], failed: [] });
    });

    it('reports all failures with their reasons', async () => {
        const error = new Error('nope');
        const result = await assignToChannels(['a'], buildInput, vi.fn().mockRejectedValue(error));
        expect(result.succeeded).toEqual([]);
        expect(result.failed).toEqual([{ channelId: 'a', reason: error }]);
    });

    it('records which channels failed on partial failure', async () => {
        const mutationFn = vi.fn(({ input }) =>
            input.channelId === 'b' ? Promise.reject(new Error('boom')) : Promise.resolve({}),
        );
        const result = await assignToChannels(['a', 'b', 'c'], buildInput, mutationFn);
        expect(result.succeeded).toEqual(['a', 'c']);
        expect(result.failed.map(f => f.channelId)).toEqual(['b']);
    });

    it('continues with remaining channels after a failure', async () => {
        const mutationFn = vi.fn().mockRejectedValue(new Error('x'));
        await assignToChannels(['a', 'b', 'c'], buildInput, mutationFn);
        expect(mutationFn).toHaveBeenCalledTimes(3);
    });

    it('propagates synchronous buildInput errors to the caller', async () => {
        const throwing = () => {
            throw new Error('bad input');
        };
        await expect(assignToChannels(['a'], throwing, vi.fn())).rejects.toThrow('bad input');
    });
});
```

The last case is why `onError` has to come back — a synchronous `buildInput` throw rejects the whole `mutationFn`, and without `onError` the user sees nothing at all.

With that extracted, the `mutationFn` in my inline suggestion collapses to `assignToChannels(selectedChannelIds, id => buildInput(id, additionalData), mutationFn)`.

Ryrahul:
Need to close this ig  #4687 landed the same multi-channel select feature on `master` while this was in review. Nice work @casperiv0!

@biggamesmallworld's [review](https://github.com/vendurehq/vendure/pull/5036#pullrequestreview-4809119564) on this PR flagged several improvements that aren't in #4687's implementation. Summarizing here for visibility (@michaelbromley):

- **Sequential execution** instead of parallel `Promise.allSettled` — each mutation writes channel relations for the same entity rows, so parallel fan-out causes lock contention and overlapping reindex jobs
- **`plural()` from `@lingui/core/macro`** — the current `t` template strings don't produce proper plural forms for languages with 3+ categories (ar, ru, uk, pl)
- **`useMutation`** instead of manual `isAssigning` state — matches the dominant codebase pattern (~80 usages)
- **`onError` handler** — without it, a synchronous `buildInput` throw gives the user no feedback at all
- **Extracted `assignToChannels()` function + 5 unit tests** — covers all-success, all-failure, partial failure, continuation after failure
- **`showSearch`** on MultiSelect  search box is hidden below 10 channels without it

Happy to open a follow-up PR for these if there's interest or can drop it. Let me know.

biggamesmallworld:
Let's keep this PR open and rebase onto master rather than closing it.

#4687 landed the feature, but the improvements from my review still apply on top of what's now on master — sequential execution instead of `Promise.allSettled`, `plural()`, `useMutation` + `onError` (a sync `buildInput` throw currently escapes `handleAssign` with no feedback), `showSearch`, and the extracted `assignToChannels()` with unit tests. Only the reset-on-close point is already handled.

Rebase onto master, drop the `.po` churn, and the diff becomes just those changes. Still valuable.

biggamesmallworld (approved):
🏄‍♂️

PR #5039, opened 2026-07-27
fix(core): Report when migration patterns match no files

### LINKED ISSUE #5001: npx vendure migrate --run silently no-ops (exits 0, ~3s) instead of running pending migrations
**Describe the bug**

Running `npx vendure migrate --run` with pending migrations present exits `0` after roughly 3 seconds without applying anything, printing no error and no indication that no migrations ran. It looks identical to a successful "already up to date" run, so it silently masks a database that is out of sync with the entity schema.

**To Reproduce**

1. In a Vendure project with `@vendure/cli` installed, add a new TypeORM migration file under the configured `migrations` glob (e.g. `src/migrations/`) that has not yet been applied to the target database.
2. Run `npx vendure migrate --run`.
3. Observe: the command exits `0` in ~3s.
4. Inspect the `migrations` table in the database, or re-run a schema diff — the new migration was never applied.

**Expected behavior**

The command should apply the pending migration(s), print which migration(s) were run, and exit `0`. If it cannot run for any reason (e.g. a config-loading failure), it should exit non-zero and print an error, not silently no-op.

**Actual behavior**

The process exits `0` almost immediately, with no log output indicating a migration ran or was skipped, and the pending migration remains unapplied.

**Root cause (as diagnosed)**

Tracing the CLI, the process dies during config-loading — before it ever reaches `runMigrationsOperation()` — and this failure is swallowed rather than surfaced. The suspected culprit is the `@vendure/dashboard` package's `.d.ts` files being picked up under Node's native TS-stripping (`--experimental-strip-types`-style loading) when the CLI loads `vendure-config.ts`; something in that load path throws or returns early, and the CLI's error handling for that path is missing or non-fatal, so it falls through to a clean exit instead of propagating the error.

Note: `npx vendure migrate --revert` exhibits the same silent no-op behavior, for the same reason (it goes through the same config-loading path before reaching its migration operation).

**Workaround**

Bypass the `vendure` CLI entirely and run migrations via a raw TypeORM `DataSource`:

```typescript
// scripts/run-migrations-raw.ts
import 'dotenv/config';
import path from 'path';
import { DataSource } from 'typeorm';

async function main() {
    const dataSource = new DataSource({
        type: 'postgres',
        host: process.env['DB_HOST']!,
        port: +(process.env['DB_PORT'] ?? 5432),
        username: process.env['DB_USERNAME']!,
        password: process.env['DB_PASSWORD']!,
        database: process.env['DB_NAME']!,
        migrations: [path.join(__dirname, '../src/migrations/*.+(js|ts)')],
        logging: true,
    });

    await dataSource.initialize();
    try {
        const applied = await dataSource.runMigrations();
        console.log(`Applied ${applied.length} migration(s):`, applied.map(m => m.name));
    } finally {
        await dataSource.destroy();
    }
}

main().catch(err => {
    console.error(err);
    process.exit(1);
});
```

Run with `npx ts-node scripts/run-migrations-raw.ts` (or equivalent). This correctly applies pending migrations and records them in the `migrations` table, confirming the migrations themselves are valid and the bug is isolated to the CLI wrapper, not to TypeORM or the migration files.

- @vendure/core version: 3.7.1 (also @vendure/cli 3.7.1, @vendure/dashboard 3.7.1)
- Nodejs version: v22.22.3
- Database (mysql/postgres etc): PostgreSQL 17.10
- Operating System (Windows/macOS/Linux): reproduced on both macOS (local dev, Darwin 25.5.0) and Linux (Ubuntu 24.04 LTS, production/staging server)
- Browser (if applicable): N/A (CLI issue)

**Additional context**

- Every migration file's header comment that cites `npx vendure migrate --run` as the way to verify/apply it is currently misleading, since the command does not reliably do what it claims.
- This is a correctness/data-safety issue, not just a DX annoyance: a silent no-op that looks like success can leave a production database out of sync with entity definitions with no visible warning.
- Would like a root-cause fix (either the config-loading failure should be fixed, or at minimum the CLI should fail loudly instead of silently exiting 0) before trusting the `vendure migrate` CLI path again, especially against staging/production databases.

alex-hahn:
## Description

`runMigrations()` returns an empty array in two different situations:

1. every migration has already been applied, and
2. the configured `migrations` glob patterns matched no files at all.

Both CLI entry points derive their report from that array alone, so the second case is
presented as `No pending migrations found`. A database that is out of sync with the entity
schema is therefore indistinguishable from an up-to-date one, and `vendure migrate --run`
exits `0` having applied nothing.

The patterns are resolved relative to the current working directory, so this happens when the
command is run from an unexpected directory, or when the patterns point at compiled output
(e.g. `dist/migrations/*.js`) that has not been built.

## Approach

The case is detectable from the migration classes TypeORM actually loaded (`connection.migrations`),
which is empty when the globs matched nothing. Where to report it needs some care:

- **Logging it directly does not work.** `log()` in `migrate.ts` is a no-op while running from
  the CLI, because a spinner is active for the duration of the call and writing to stdout
  would corrupt it. That suppression is why the condition is currently invisible.
- **Throwing would break the default project template.** The scaffold in
  `packages/create/templates/vendure-config.hbs` configures
  `migrations: [path.join(__dirname, './migrations/*.+(js|ts)')]` before any migration file
  exists, so a freshly created project would fail `vendure migrate --run` with a non-zero exit.

So the diagnostic is handed to the caller through a new optional
`RunMigrationsOptions.onNoMigrationsFound` callback, and both CLI entry points surface it in
place of the misleading default message. This is additive and non-breaking; the exit code is
deliberately unchanged.

If you would prefer a non-zero exit here, that is a one-line change — happy to adjust.

## Tests

- `packages/core/src/migrate.spec.ts` (new file) — 7 tests covering the message builder:
  loaded migrations, unconfigured/empty patterns, class-valued entries, the object form of the
  `migrations` option, and the reported patterns and cwd.
- `packages/cli/src/commands/migrate/migration-operations-reporting.spec.ts` (new file) —
  3 tests covering what the CLI actually reports. Verified that the first fails without this
  change (`expected 'No pending migrations found' to be 'No migration files matched…'`) while
  the other two stay green.

Full `packages/cli` suite passes (275 tests), and `tsc --noEmit` reports no new errors in
either package.

Fixes #5001

### DIFF
--- a/packages/cli/src/commands/migrate/migration-operations-reporting.spec.ts
+++ b/packages/cli/src/commands/migrate/migration-operations-reporting.spec.ts
@@ -1,0 +1,71 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const runMigrationsMock = vi.fn();
+
+vi.mock('@clack/prompts', () => ({
+    log: { info: vi.fn(), error: vi.fn(), success: vi.fn() },
+}));
+vi.mock('../../shared/project-validation', () => ({
+    validateVendureProjectDirectory: vi.fn(),
+}));
+vi.mock('../../shared/shared-prompts', () => ({
+    analyzeProject: vi.fn().mockResolvedValue({ project: {}, tsConfigPath: 'tsconfig.json' }),
+}));
+vi.mock('../../shared/vendure-config-ref', () => ({
+    VendureConfigRef: class {
+        getPathRelativeToProjectRoot() {
+            return 'src/vendure-config.ts';
+        }
+    },
+}));
+vi.mock('../../shared/load-vendure-config-file', () => ({
+    loadVendureConfigFile: vi.fn().mockResolvedValue({}),
+}));
+vi.mock('@vendure/core', () => ({
+    runMigrations: (...args: any[]) => runMigrationsMock(...args),
+    generateMigration: vi.fn(),
+    revertLastMigration: vi.fn(),
+}));
+
+// eslint-disable-next-line import/first
+import { runMigrationsOperation } from './migration-operations';
+
+describe('runMigrationsOperation() reporting', () => {
+    beforeEach(() => {
+        runMigrationsMock.mockReset();
+    });
+
+    it('reports the reason when no migration files were found', async () => {
+        // #5001 — a non-matching `migrations` glob is reported as success, which makes an
+        // out-of-sync database look like an up-to-date one.
+        runMigrationsMock.mockImplementation((config: unknown, options: any) => {
+            options?.onNoMigrationsFound?.('No migration files matched the configured patterns.');
+            return Promise.resolve([]);
+        });
+
+        const result = await runMigrationsOperation();
+
+        expect(result.success).toBe(true);
+        expect(result.message).toBe('No migration files matched the configured patterns.');
+        expect(result.message).not.toBe('No pending migrations found');
+    });
+
+    it('still reports "No pending migrations found" when migration files exist', async () => {
+        runMigrationsMock.mockResolvedValue([]);
+
+        const result = await runMigrationsOperation();
+
+        expect(result.success).toBe(true);
+        expect(result.message).toBe('No pending migrations found');
+    });
+
+    it('reports the number of migrations that ran', async () => {
+        runMigrationsMock.mockResolvedValue(['1700000000000-first', '1700000000001-second']);
+
+        const result = await runMigrationsOperation();
+
+        expect(result.success).toBe(true);
+        expect(result.message).toBe('Successfully ran 2 migrations');
+        expect(result.migrationsRan).toHaveLength(2);
+    });
+});
--- a/packages/cli/src/commands/migrate/migration-operations.ts
+++ b/packages/cli/src/commands/migrate/migration-operations.ts
@@ -78,11 +78,14 @@ export async function runMigrationsOperation(configFile?: string): Promise<Migra
         const config = await loadVendureConfigFile(vendureConfig);
 
         log.info('Running migrations...');
-        const migrationsRan = await runMigrations(config);
+        let noMigrationsFoundMessage: string | undefined;
+        const migrationsRan = await runMigrations(config, {
+            onNoMigrationsFound: message => (noMigrationsFoundMessage = message),
+        });
 
         const report = migrationsRan.length
             ? `Successfully ran ${migrationsRan.length} migrations`
-            : 'No pending migrations found';
+            : (noMigrationsFoundMessage ?? 'No pending migrations found');
 
         return {
             success: true,
--- a/packages/cli/src/commands/migrate/run-migration/run-migration.ts
+++ b/packages/cli/src/commands/migrate/run-migration/run-migration.ts
@@ -23,10 +23,13 @@ async function runRunMigration(configFile?: string): Promise<CliCommandReturnVal
 
     const runSpinner = spinner();
     runSpinner.start('Running migrations...');
-    const migrationsRan = await runMigrations(config);
+    let noMigrationsFoundMessage: string | undefined;
+    const migrationsRan = await runMigrations(config, {
+        onNoMigrationsFound: message => (noMigrationsFoundMessage = message),
+    });
     const report = migrationsRan.length
         ? `Successfully ran ${migrationsRan.length} migrations`
-        : 'No pending migrations found';
+        : (noMigrationsFoundMessage ?? 'No pending migrations found');
     runSpinner.stop(report);
     return {
         project,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -18,7 +18,7 @@ export * from './event-bus/index';
 export * from './health-check/index';
 export * from './i18n/index';
 export * from './job-queue/index';
-export { generateMigration, revertLastMigration, runMigrations } from './migrate';
+export { RunMigrationsOptions, generateMigration, revertLastMigration, runMigrations } from './migrate';
 export * from './migration-utils/index';
 export * from './plugin/index';
 export * from './process-context/index';
--- a/packages/core/src/migrate.spec.ts
+++ b/packages/core/src/migrate.spec.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest';
+
+import { getNoMigrationsFoundMessage } from './migrate';
+
+describe('getNoMigrationsFoundMessage()', () => {
+    class SomeMigration {}
+
+    it('returns undefined when migration files were loaded', () => {
+        expect(getNoMigrationsFoundMessage(3, ['src/migrations/*.ts'], '/project')).toBeUndefined();
+    });
+
+    it('returns undefined when no migrations are configured', () => {
+        expect(getNoMigrationsFoundMessage(0, undefined, '/project')).toBeUndefined();
+        expect(getNoMigrationsFoundMessage(0, [], '/project')).toBeUndefined();
+    });
+
+    it('returns undefined when migrations are configured as classes', () => {
+        // Classes are passed directly rather than resolved from disk, so a zero count
+        // cannot be attributed to non-matching glob patterns.
+        expect(getNoMigrationsFoundMessage(0, [SomeMigration], '/project')).toBeUndefined();
+    });
+
+    it('reports the patterns and the cwd they resolve against', () => {
+        const message = getNoMigrationsFoundMessage(0, ['dist/migrations/*.js'], '/project');
+
+        expect(message).toContain('No migration files matched');
+        expect(message).toContain('/project');
+        expect(message).toContain('dist/migrations/*.js');
+    });
+
+    it('reports every configured pattern', () => {
+        const message = getNoMigrationsFoundMessage(
+            0,
+            ['dist/migrations/*.js', 'plugins/*/migrations/*.js'],
+            '/project',
+        );
+
+        expect(message).toContain('dist/migrations/*.js');
+        expect(message).toContain('plugins/*/migrations/*.js');
+    });
+
+    it('supports the object form of the migrations option', () => {
+        const message = getNoMigrationsFoundMessage(0, { first: 'dist/migrations/*.js' } as any, '/project');
+
+        expect(message).toContain('dist/migrations/*.js');
+    });
+
+    it('ignores class entries when reporting patterns', () => {
+        const message = getNoMigrationsFoundMessage(0, [SomeMigration, 'dist/migrations/*.js'], '/project');
+
+        expect(message).toContain('dist/migrations/*.js');
+        expect(message).not.toContain('SomeMigration');
+    });
+});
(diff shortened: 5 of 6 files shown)


biggamesmallworld (dismissed):
Thanks for this — the underlying observation is a good catch, and the PR description does a genuinely useful thing by writing down *why* the obvious fixes (log directly, throw) don't work. The `onNoMigrationsFound` seam is the right shape. Two things need to change before this lands, and I think one of them makes the PR considerably more valuable rather than smaller.

## The signal is one step away from a better one

I built `packages/core` from this branch and ran the CLI e2e suite (`e2e/vitest.e2e.config.mts`), which the unit suite doesn't cover. One existing test fails:

```
FAIL  runMigrationsOperation > should report no pending migrations when none exist
AssertionError: expected 'No migration files matched the config…' to contain 'No pending migrations found'

- No pending migrations found
+ No migration files matched the configured `migrations` patterns, so no migrations can be run.
+ Patterns are resolved relative to the current directory (…/e2e/fixtures/test-project):
+  - …/e2e/fixtures/test-project/migrations/*.ts
```

That isn't a stale assertion to update — it's the regression. The fixture has a valid glob over an empty directory, which is what every project looks like before its first migration is authored. `connection.migrations.length === 0` overwhelmingly means "no migrations written yet", not "your globs are broken", so as written this prints an alarming, factually-true non-problem to exactly the users least able to evaluate it. You identified this scenario in the description (the `create` template configures the glob before any migration exists) and concluded it must not *throw* — but not throwing isn't the same as not being wrong.

There is a signal that discriminates exactly, and it's proof rather than heuristic: **zero migration classes loaded while the `migrations` table has rows** can only mean the globs stopped matching. That's also the dangerous real-world case — a deploy pointed at `dist/migrations/*.js` that was never built, against a database with history.

```ts
import { MigrationExecutor } from 'typeorm';

const executed = await new MigrationExecutor(connection).getExecutedMigrations();
if (!connection.migrations.length && executed.length) {
    // globs definitively broken
}
```

With that gate, the failing test above passes untouched, and the warning only fires when it's true.

I checked the fresh-database case since it's the one that would bite: against a sqlite DB with no `migrations` table, `getExecutedMigrations()` returns `[]` rather than throwing, so no special-casing is needed. It does *create* the `migrations` table as a side effect — which is fine at this call site, because `connection.runMigrations()` on the next line creates it anyway, but worth knowing before reusing the call somewhere read-only.

## The bigger hole is already computed and thrown away

While you're in here: `checkMigrationStatus()` (`migrate.ts:87`) already computes `builderLog.upQueries` and emits *precisely* the warning #5001 asks for —

> Your database schema does not match your current configuration. Generate a new migration for the following changes:

— via `log()`, which is a hard no-op whenever `VENDURE_RUNNING_IN_CLI` is set (`migrate.ts:312`). So on every single `vendure migrate --run`, the "your DB is out of sync" diagnostic is calculated and discarded. That is the data-safety gap in the issue, it costs nothing extra to detect, and the callback you've just added is the natural way to surface it. Routing that through the same seam would be the higher-value half of this change.

Worth noting the same suppression applies to `revertLastMigration` (`migrate.ts:155`), which the issue explicitly calls out as having identical symptoms.

## `Fixes #5001` isn't accurate

The reporter's glob matches — their workaround `DataSource` uses the identical pattern and applies migrations successfully — so `connection.migrations.length` is non-zero in their scenario and `onNoMigrationsFound` never fires. Their symptom is exit 0 in ~3s with no output at all, before any migration work happens.

I tried to reproduce that and could not. Every config-load failure I could construct against the e2e fixture exits non-zero with a message:

```
config = undefined            → ■ Cannot read properties of undefined (reading 'plugins')          EXIT=1
config with top-level await   → ■ require() cannot be used on an ESM graph with top-level await    EXIT=1
```

So that root cause is still open and needs a repro from the reporter (their tsconfig, plus `NODE_DEBUG=module npx vendure migrate --run`). Could you retarget this to `Relates to #5001`? Landing it as `Fixes` would close a live data-safety report that remains unaddressed.

## Test that would have caught this

Nothing currently executes `warnIfNoMigrationsFound` — the CLI spec mocks `@vendure/core` wholesale, so `runMigrations` is a `vi.fn()`, and the core spec only exercises the message builder. The e2e fixture drives real TypeORM against real sqlite and is the right home. This passes on your branch and fails on master:

```ts
// #5001 — a `migrations` glob that resolves to nothing must not be reported as "up to date"
it('should report unmatched migration patterns when migrations have already been applied', async () => {
    process.chdir(TEST_PROJECT_DIR);

    // Apply a migration so the `migrations` table is non-empty
    const generateResult = await generateMigrationOperation({
        name: 'TestMigration',
        outputDir: MIGRATIONS_DIR,
    });
    expect(generateResult.success).toBe(true);
    expect((await runMigrationsOperation()).migrationsRan?.length).toBeGreaterThan(0);

    // Simulate the glob resolving to nothing (wrong cwd, or unbuilt `dist/migrations/*.js`)
    await fs.emptyDir(MIGRATIONS_DIR);

    const result = await runMigrationsOperation();

    expect(result.migrationsRan).toHaveLength(0);
    expect(result.message).toContain('No migration files matched');
});
```

Together with the existing `'should report no pending migrations when none exist'`, those two pin the discriminator from both sides.

One warning about verifying this locally: the CLI e2e suite resolves `@vendure/core` through `packages/core/dist`, so it will happily pass against a build that predates your change and tell you nothing. Run `npm run build` in `packages/core` first.

## Smaller things

- **Drop the two class-configuration tests** in `migrate.spec.ts`. If a migration class is in `options.migrations`, TypeORM loads it, so `connection.migrations.length` can't be `0` — those states are unreachable in production. They only typecheck because `getNoMigrationsFoundMessage` takes the count and the config as independent parameters; passing the connection (or just the two derived booleans) would make them impossible to express.
- **The report-building block is now byte-identical** in `migration-operations.ts:81-88` and `run-migration/run-migration.ts:26-32`, and `'No pending migrations found'` appears in three places plus an assertion in `migration-operations.spec.ts:194`. Worth extracting into one `formatMigrationReport()`.
- **Wrong channel.** `run-migration.ts:33` passes a multi-line message to `runSpinner.stop()` and `migrate.ts:87` renders it with `log.success()` — clack prefixes only the first line, so the pattern list hangs outside the box, and a warning gets a green success glyph with exit 0. If it's worth reporting, it isn't a success.
- **Public API.** `RunMigrationsOptions` is exported from `packages/core/src/index.ts` with `@docsCategory` and `@since`, and its whole payload is a pre-formatted English string a programmatic consumer can only print. If it stays public, hand over data (`{ patterns, cwd }`, or the pending-change list) and let each caller format. Also prefer `export type { RunMigrationsOptions }` — the current value-export breaks downstream consumers building with `isolatedModules`.
- `migration-operations-reporting.spec.ts:50` (`expect(...).not.toBe('No pending migrations found')`) can't fail if line 49 passes. And that file duplicates the existing `describe('runMigrationsOperation()')` in `migration-operations.spec.ts` under a different mocking regime — worth merging.
- The comment at `migrate.ts:141` says `log()` is suppressed because a spinner is active. It's actually suppressed whenever `VENDURE_RUNNING_IN_CLI` is set, spinner or not — the `migrate --run` path (`migrate.ts:62`) has no spinner at all. Same for the ten-line block at `migrate.ts:104`; most of it is commit-message material, and the one durable sentence is "TypeORM resolves migration globs relative to `process.cwd()` and silently yields zero classes when nothing matches."

Happy to review again once the gate changes — the diagnostic seam itself is good work and I'd like to see it land carrying the schema-drift warning too.

biggamesmallworld:
I have taken this over, @alex-hahn. It had been sitting for four weeks with no response, and the diagnostic seam is worth landing. Your commit is preserved as the base of the branch and you are credited as co-author on the follow-up, so the PR stays yours.

What changed from your version:

- **The gate.** The warning now fires only when zero migration classes loaded **and** the `migrations` table has rows. Zero classes on its own is just a project that has not written its first migration yet, which is how the `create` scaffold ships, so the previous version warned on the normal case. That is also what broke `should report no pending migrations when none exist` in the CLI e2e suite. With the gate it passes untouched.
- **The schema-drift warning now goes through your callback.** `checkMigrationStatus()` was already computing exactly the warning #5001 asks for and throwing it away on every `vendure migrate --run`, because `log()` is suppressed under `VENDURE_RUNNING_IN_CLI`. That is the data-safety half of the issue and your seam was the natural place for it.
- **`onNoMigrationsFound(message)` became `onDiagnostic(diagnostic)`.** It hands over `patterns`, `cwd` and `queries` rather than a pre-formatted English string, so a programmatic consumer can act on the condition instead of only printing it.
- **Tests.** Added an e2e case that drives real TypeORM against real sqlite and fails on `master`. Nothing previously executed the detection path, since the CLI spec mocks `@vendure/core` wholesale. Dropped the two class-configuration unit tests, which described states TypeORM cannot produce.
- **Retargeted to `Relates to #5001`.** The reporter's glob matches, so this code path never fires in their scenario. Their root cause is still open.

Also folded the duplicated report-building block into one `buildMigrationReport()`, and a warning no longer renders with a green success glyph.

Happy to hand it back if you would rather finish it yourself. Otherwise I will get it reviewed and merged.

biggamesmallworld:
One blocker I cannot clear from this side: the CLA check is still failing. Your original commit is in this branch and always will be, so the signature is needed regardless of who finishes the work. @alex-hahn, please post the line the bot asked for on this PR when you get a chance:

> I have read the CLA Document and I hereby sign the CLA

If you would rather not, tell me and I will rewrite the change so it carries none of your code, and credit you in the description instead.

biggamesmallworld:
Correction to my previous two comments: rather than push to your branch, I have opened #5174 and reverted this one back to exactly the commit you left it at. Sorry for the churn on your fork.

Closing this in favour of #5174. The reasoning is yours and it is credited as such in the commit message and at the top of the new description. What changed from your version:

- **The gate.** The warning now fires only when zero migration classes loaded **and** the `migrations` table has rows. Zero classes on its own is just a project that has not written its first migration yet, which is how the `create` scaffold ships. That is also what broke `should report no pending migrations when none exist` in the CLI e2e suite, which the unit tests here could not catch because they mock `@vendure/core` wholesale.
- **The schema-drift warning now goes through your callback.** `checkMigrationStatus()` was already computing exactly the warning #5001 asks for and throwing it away on every `vendure migrate --run`, because `log()` is suppressed whenever `VENDURE_RUNNING_IN_CLI` is set. That is the data-safety half of the issue and your seam was the natural place for it.
- **`onNoMigrationsFound(message)` became `onDiagnostic(diagnostic)`**, handing over `patterns`, `cwd` and `queries` rather than a pre-formatted English string.
- **Retargeted to `Relates to #5001`.** The reporter's glob matches, so this code path never fires in their scenario and their root cause is still open.

To be clear about why this is a new PR and not a merge of yours: the CLA on this branch was never signed, so it could not have been merged as it stands. If you would still like to sign it, say so and I am happy to reopen this and land it under your name instead. Either way, thanks for the original catch.

PR #5043, opened 2026-07-28
fix(core): Add channel-scope guard to delete paths & cross-channel re…

### LINKED ISSUE #5042: fix(core): Cross-channel delete bypass on FacetValue, ProductOption & Promotion
**Describe the bug**
`FacetValueService.delete()`, `ProductOptionService.delete()`, and `PromotionService.softDeletePromotion()` load entities via `getEntityOrThrow()` without passing `{ channelId: ctx.channelId }`, allowing a channel-scoped admin to delete entities belonging to other channels. This is the same class of defect as #5003, but on delete paths instead of update paths.

**To Reproduce**
1. Create a multi-channel Vendure deployment with channel-scoped admin roles
2. Create a FacetValue (or ProductOption or Promotion) in Channel A
3. As an admin with delete permissions only on Channel B, call the delete mutation with the Channel-A entity ID using the Channel-B token
4. The entity is deleted from Channel A despite the admin having no permissions on that channel

**Expected behavior**
The delete mutation should throw `EntityNotFoundError` ("No {Entity} with the id {id} could be found") because the entity does not belong to the caller's active channel.

**Actual behavior**
The entity is loaded globally (across all channels) and the delete commits successfully, bypassing channel isolation.

**Environment:**
- Affected services: `facet-value.service.ts:228`, `product-option.service.ts:146`, `promotion.service.ts:196`

**Additional context**
Follow-up to #5003. The reviewer on #5017 identified these three delete paths as having the same missing `{ channelId: ctx.channelId }` guard.

Ryrahul:
Fixes #5042
Relates to #5003

# Description

Add `{ channelId: ctx.channelId }` guard to `delete()` in `FacetValueService`, `ProductOptionService`, and `PromotionService.softDeletePromotion()`. These had the same cross-channel bypass defect as the `update()` paths fixed in #5017.

Also adds cross-channel e2e regression tests for:
- **Delete protection:** FacetValue, ProductOption, Promotion
- **Update protection:** Collection, ProductOptionGroup, ProductOption, Facet (PR #4821 guards that had no e2e coverage)

# Breaking changes

No breaking changes.

# Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5043"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787821940&installation_model_id=20193&pr_number=5043&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5043&signature=11e1796e0f7b5db4ba14d88e35b97761ee05919c7094ea2a1e31666a8557e258"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/collection.e2e-spec.ts
+++ b/packages/core/e2e/collection.e2e-spec.ts
@@ -2161,6 +2161,76 @@ describe('Collection resolver', () => {
         });
     });
 
+    describe('cross-channel update protection', () => {
+        const CHANNEL_A_TOKEN = 'coll-cross-channel-a';
+        const CHANNEL_B_TOKEN = 'coll-cross-channel-b';
+        let targetCollectionId: string;
+
+        beforeAll(async () => {
+            adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+            for (const [code, token] of [
+                ['coll-cross-a', CHANNEL_A_TOKEN],
+                ['coll-cross-b', CHANNEL_B_TOKEN],
+            ]) {
+                await adminClient.query(createChannelDocument, {
+                    input: {
+                        code,
+                        token,
+                        defaultLanguageCode: LanguageCode.en,
+                        currencyCode: CurrencyCode.USD,
+                        pricesIncludeTax: true,
+                        defaultShippingZoneId: 'T_1',
+                        defaultTaxZoneId: 'T_1',
+                    },
+                });
+            }
+            adminClient.setChannelToken(CHANNEL_A_TOKEN);
+            const { createCollection } = await adminClient.query(createCollectionDocument, {
+                input: {
+                    filters: [],
+                    translations: [
+                        {
+                            languageCode: LanguageCode.en,
+                            name: 'Channel-A Collection',
+                            description: 'Belongs to Channel A only',
+                            slug: 'channel-a-collection',
+                        },
+                    ],
+                },
+            });
+            targetCollectionId = createCollection.id;
+        });
+
+        afterAll(() => {
+            adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+        });
+
+        it('cannot update a Collection belonging to another channel', async () => {
+            adminClient.setChannelToken(CHANNEL_B_TOKEN);
+            await expect(
+                adminClient.query(updateCollectionDocument, {
+                    input: {
+                        id: targetCollectionId,
+                        translations: [
+                            {
+                                languageCode: LanguageCode.en,
+                                name: 'PWNED-BY-CHANNEL-B',
+                                description: 'TAMPERED',
+                                slug: 'pwned',
+                            },
+                        ],
+                    },
+                }),
+            ).rejects.toThrow(/No Collection with the id .* could be found/);
+
+            adminClient.setChannelToken(CHANNEL_A_TOKEN);
+            const { collection } = await adminClient.query(getCollectionDocument, {
+                id: targetCollectionId,
+            });
+            expect(collection?.name).toBe('Channel-A Collection');
+        });
+    });
+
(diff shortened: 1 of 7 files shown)


latifniz:
Two of the three fixes here match exactly what I'd have suggested: ProductOptionService.delete() and PromotionService.softDeletePromotion() both look complete.

One small thing worth a look on FacetValueService.delete() though: checkFacetValueUsage(ctx, [id]) still runs before the new channel check, in its original position. That means for a FacetValue outside the caller's Channel, the usage count (how many Products/Variants reference it) still gets computed and could leak out before the ownership check ever gets a chance to reject the request. It's a small thing, just two numbers, not real data, but since we're already touching this function, might be worth moving the getEntityOrThrow(..., { channelId: ctx.channelId }) call to the very first line, ahead of checkFacetValueUsage, so nothing runs against an out-of-scope entity at all before ownership is confirmed.

grolmus (approved):
Thanks @Ryrahul — this is a tight, well-scoped fix. The three guards copy the established `{ channelId: ctx.channelId }` pattern from the update paths (#4821/#5017) verbatim, and the regression tests genuinely fail against the pre-fix code, so they're real protection rather than decoration.

I verified the guard doesn't break legitimate deletes:
- **Default channel / superadmin:** every entity is assigned to the default channel on creation (`assignToCurrentChannel` unions in the default channel), so a default-channel admin still resolves and deletes entities created in sub-channels. No regression.
- **Multi-channel entities:** membership is an OR across the channel join, so an entity in {A, B} stays deletable from either.
- The rejection surfaces as `EntityNotFoundError` rather than `Forbidden`, which matches the sibling update guards and the Vendure convention of not leaking cross-tenant existence.

Confirmed these three were the last delete paths still using the bare `getEntityOrThrow(ctx, Entity, id)` form — the rest of the ChannelAware surface is already guarded — so this completes that set. CI is green (only the fork `deploy` job fails, which is the usual missing `VERCEL_TOKEN`).

Approving. A few **non-blocking** notes for a follow-up push if you like:

- The co-located "protection" blocks assert the entity *survived* a cross-channel attempt, which proves rejection but wouldn't catch an over-strict guard that rejected *every* delete. Same-channel happy-path deletes are covered elsewhere in each suite, so coverage is fine overall — but a one-line same-channel positive delete inside these blocks would nail it down.
- A couple of the `describe`/test blocks are titled "cross-channel **update** protection" but contain delete tests (and the `collection.e2e-spec.ts` block is update-only) — worth renaming so the titles match their contents.
- Minor: in `FacetValueService.delete`, the usage check (`checkFacetValueUsage`) runs before the new guard, so a rejected cross-channel call does a wasted usage query before throwing not-found. Moving the guard first skips that.

@michaelbromley — looping you in as the owner of the channel-scoping / security area: this is a T2 multi-tenancy fix and it's good to merge from my side after review + a deep pass. There's also a **related, out-of-scope** finding in the same area that I've written up separately (internal) rather than here — I'll share it with you directly.

Thanks again @Ryrahul — solid, disciplined work.

michaelbromley (approved):

PR #5049, opened 2026-07-29
fix(core): Enforce channel scope on product variant soft-deletion

grolmus:
## Description

`ProductVariantService.softDelete` resolved the variants to delete with a channel-unaware lookup (`getRepository(...).find({ where: { id: In(ids) } })`). Since it is reachable directly through the `deleteProductVariant` / `deleteProductVariants` admin mutations (`@Allow(DeleteCatalog, DeleteProduct)`), an administrator scoped to one channel could soft-delete a `ProductVariant` belonging only to **another** channel — a cross-channel data-integrity bypass. `ProductVariant` is `ChannelAware`.

This is the delete-path counterpart to the channel-scoping already enforced on `ProductService.softDelete` (which resolves the Product via `getEntityOrThrow(..., { channelId })`).

### Fix

- `ProductVariantService.softDelete` now scopes the lookup to the active channel via `findByIdsInChannel(ctx, ProductVariant, ids, ctx.channelId, {})` and throws `EntityNotFoundError` for any requested id that is not in the active channel — the same convention (`EntityNotFoundError`, not `ForbiddenError`) used by the sibling channel-scoped lookups, so cross-channel existence isn't leaked.
- The trusted **product-deletion cascade** (`ProductService.softDelete`) opts out (`checkChannel: false`): deleting a Product is a global operation and must still cascade to every one of its variants, including any that were individually removed from the active channel (via `removeProductVariantsFromChannel`). Without the opt-out, deleting such a product from a sub-channel would incorrectly throw.

### Tests

Adds cross-channel regression tests to `product-channel.e2e-spec.ts`:
- a second-channel admin **cannot** soft-delete a variant that lives only in the default channel (throws), the variant **survives**, and it **can** still be deleted from its own channel;
- a channel-scoped **product** delete still cascades to a variant that was removed from that channel (verifies the cascade opt-out);
- an unknown variant id **throws** rather than reporting `DELETED`;
- `deleteProductVariants` **discards the whole batch** when one id is outside the active channel — the variant the admin *could* have deleted survives. This asserts the all-or-nothing guarantee below and is the test that would fail if the `@Transaction()` wrapper were ever dropped. It is skipped on sqljs, whose transaction rollback is a no-op (same pattern as `database-transactions.e2e-spec.ts`), so a partial delete would otherwise persist and mask the guarantee.

Verified red→green for the direct-mutation guard and the cascade (each test fails against the pre-fix / pre-opt-out code). Full `product-channel` suite green on postgres (35 tests, bulk case included) and on sqljs (34 + 1 skipped).

### Behavioural note

`deleteProductVariant` / `deleteProductVariants` now throw `EntityNotFoundError` for an id that is unknown or not in the active channel, where they previously returned `DELETED` silently. This aligns variant deletion with how `deleteProduct` already behaves.

`deleteProductVariants` is `@Transaction()`-wrapped, so the throw makes the **bulk** mutation all-or-nothing: a batch containing any unknown or out-of-channel id is rejected in full rather than partially applied, and the mutation returns a top-level error instead of a `DeletionResponse[]`. Consequently the Dashboard bulk action (`product-variant-list-bulk-actions.ts:101`) no longer reaches its per-item `.message` path on such a failure. This matches `deleteProducts`, and a partial cross-channel delete is worse than a failed one.

Relates to OSS-666.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5049"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787915960&installation_model_id=20193&pr_number=5049&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5049&signature=e7117681144f3b0e7514111186c203738d17f866dac3eab55772f80095d49d1d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/product-channel.e2e-spec.ts
+++ b/packages/core/e2e/product-channel.e2e-spec.ts
@@ -25,6 +25,7 @@ import {
     createRoleDocument,
     deleteProductDocument,
     deleteProductVariantDocument,
+    deleteProductVariantsDocument,
     getChannelsDocument,
     getProductVariantListDocument,
     getProductWithVariantsDocument,
@@ -1102,7 +1103,9 @@ describe('ChannelAware Products and ProductVariants', () => {
                 await adminClient.asUserWithCredentials('deleter2@test.com', 'test');
                 adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
                 await adminClient.query(deleteProductVariantDocument, { id: defaultOnlyVariantId });
-            }, 'No ProductVariant with the id'),
+                // Deferred so `defaultOnlyVariantId` (set in beforeAll) is read at throw time, not
+                // at test-registration time when it is still undefined.
+            }, () => `No ProductVariant with the id "${defaultOnlyVariantId.replace(/^T_/, '')}" could be found`),
         );
 
         it('the variant still exists after the cross-channel delete attempt', async () => {
@@ -1208,6 +1211,103 @@ describe('ChannelAware Products and ProductVariants', () => {
             });
             expect(productVariants.items).toEqual([]);
         });
+
+        it(
+            'deleting an unknown variant id throws rather than reporting DELETED',
+            assertThrowsWithMessage(async () => {
+                await adminClient.asSuperAdmin();
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                await adminClient.query(deleteProductVariantDocument, { id: 'T_999999' });
+            }, 'No ProductVariant with the id "999999" could be found'),
+        );
+
+        // deleteProductVariants is @Transaction()-wrapped, so a batch containing an id outside the
+        // active channel is rejected in full rather than partially applied. Skipped on sqljs, whose
+        // transaction rollback is a no-op (see database-transactions.e2e-spec.ts), so a partial
+        // delete would persist and mask the guarantee this test asserts.
+        it.skipIf(!process.env.DB || process.env.DB === 'sqljs')(
+            'deleteProductVariants discards the whole batch when one id is not in the active channel',
+            async () => {
+                await adminClient.asSuperAdmin();
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+
+                // A variant which lives only in the default channel, plus one assigned to T_2 as well.
+                const { createProduct } = await adminClient.query(createProductDocument, {
+                    input: {
+                        translations: [
+                            {
+                                languageCode: LanguageCode.en,
+                                name: 'OSS-666 Bulk Product',
+                                slug: 'oss-666-bulk',
+                                description: '',
+                            },
+                        ],
+                    },
+                });
+                const { createProductVariants } = await adminClient.query(createProductVariantsDocument, {
+                    input: [
+                        {
+                            productId: createProduct.id,
+                            sku: 'OSS-666-BULK',
+                            optionIds: [],
+                            translations: [{ languageCode: LanguageCode.en, name: 'OSS-666 Bulk Variant' }],
+                        },
+                    ],
+                });
+                productVariantGuard.assertSuccess(createProductVariants[0]);
+                const defaultOnlyId = createProductVariants[0].id;
+
+                const { createProduct: sharedProduct } = await adminClient.query(createProductDocument, {
+                    input: {
+                        translations: [
+                            {
+                                languageCode: LanguageCode.en,
+                                name: 'OSS-666 Bulk Shared Product',
+                                slug: 'oss-666-bulk-shared',
+                                description: '',
+                            },
+                        ],
+                    },
+                });
+                const { createProductVariants: sharedVariants } = await adminClient.query(
+                    createProductVariantsDocument,
+                    {
+                        input: [
+                            {
+                                productId: sharedProduct.id,
+                                sku: 'OSS-666-BULK-SHARED',
+                                optionIds: [],
+                                translations: [
+                                    { languageCode: LanguageCode.en, name: 'OSS-666 Bulk Shared Variant' },
+                                ],
+                            },
+                        ],
+                    },
+                );
+                productVariantGuard.assertSuccess(sharedVariants[0]);
+                const sharedId = sharedVariants[0].id;
+                await adminClient.query(assignProductToChannelDocument, {
+                    input: { channelId: 'T_2', productIds: [sharedProduct.id] },
+                });
+
+                // A T_2 admin deletes both in one call. Only one of them is in T_2.
+                await adminClient.asUserWithCredentials('deleter2@test.com', 'test');
+                adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
+                await expect(
+                    adminClient.query(deleteProductVariantsDocument, { ids: [sharedId, defaultOnlyId] }),
+                ).rejects.toThrow('No ProductVariant with the id');
+
+                // The batch was rejected in full: the variant the admin *could* have deleted survives.
+                await adminClient.asSuperAdmin();
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                const { productVariants } = await adminClient.query(getProductVariantListDocument, {
+                    options: { filter: { id: { in: [sharedId, defaultOnlyId] } } },
+                });
+                expect(productVariants.items.map(v => v.id).sort()).toEqual(
+                    [sharedId, defaultOnlyId].sort(),
+                );
+            },
+        );
     });
 });
 
--- a/packages/core/e2e/product-channel.e2e-spec.ts
+++ b/packages/core/e2e/product-channel.e2e-spec.ts
@@ -1,4 +1,4 @@
-import { CurrencyCode, LanguageCode, Permission } from '@vendure/common/lib/generated-types';
+import { CurrencyCode, DeletionResult, LanguageCode, Permission } from '@vendure/common/lib/generated-types';
 import type { ErrorResultGuard } from '@vendure/testing';
 import { createErrorResultGuard, createTestEnvironment, E2E_DEFAULT_CHANNEL_TOKEN } from '@vendure/testing';
 import { fail } from 'assert';
@@ -23,6 +23,8 @@ import {
     createProductOptionGroupDocument,
     createProductVariantsDocument,
     createRoleDocument,
+    deleteProductDocument,
+    deleteProductVariantDocument,
     getChannelsDocument,
     getProductVariantListDocument,
     getProductWithVariantsDocument,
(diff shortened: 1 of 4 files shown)


michaelbromley (commented):
### One change worth making: reuse `idsAreEqual` for the ID comparison

In `product-variant.service.ts` the new code builds a Set of stringified IDs and compares against it:

```ts
const foundIds = new Set(variants.map(variant => `${variant.id}`));
const missingId = ids.find(candidate => !foundIds.has(`${candidate}`));
```

This file already imports `idsAreEqual` from `../../common/utils` and uses it in several other places to compare IDs regardless of whether they come through as strings or numbers. Better to use it here too, so there's a single consistent way of comparing IDs in the file:

```ts
const missingId = ids.find(candidate => !variants.some(v => idsAreEqual(v.id, candidate)));
```

Not a bug — the string version works fine for both numeric and UUID IDs — just a consistency thing.

### On the `checkChannel` boolean argument: decision for the record

We looked at whether `checkChannel: boolean = true` should be an options object instead (`{ checkChannel: false }`), since the call site `softDelete(ctx, ids, false)` doesn't tell you what the `false` means on its own.

We decided to keep the positional boolean, because that's already how the codebase handles this kind of flag. Around a dozen service methods take a trailing boolean flag, and the closest matches — the `delete(ctx, id, force = false)` methods on Facet, FacetValue, ProductOptionGroup, PaymentMethod and Asset — all use a positional boolean with a default, exactly like this one. The options-object style only shows up on a few newer methods that have several optional settings, where a bag of options actually pays off. For a single flag it doesn't add much, and switching to it here would make this method the odd one out.

The one downside is the cascade opt-out reading as `softDelete(ctx, ids, false)`. That's a single internal call with a comment next to it explaining why, so the ambiguity is contained. If we want to remove any doubt at that call site, a named local or a `/* checkChannel */ false` inline comment does the job.

If we ever want to move away from the boolean-flag style properly, that's a wider cleanup across all the methods that use it (the two-boolean `AssetService.delete` being the main one), not something to do differently on just this fix.

grolmus:
Done — swapped the stringified-id Set for `ids.find(candidate => !variants.some(v => idsAreEqual(v.id, candidate)))`, so ID comparison in the service is now consistent.

Agreed on keeping the positional `checkChannel` boolean per your decision-for-the-record — matches the `delete(ctx, id, force)` methods on Facet/FacetValue/etc. The one cascade call site in `product.service.ts` already carries a comment explaining what the `false` means, so the ambiguity there is covered.

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -773,11 +773,31 @@ export class ProductVariantService {
         }
     }
 
-    async softDelete(ctx: RequestContext, id: ID | ID[]): Promise<DeletionResponse> {
+    async softDelete(
```
biggamesmallworld, line 789: Nit. The boolean stays, so it wants a docblock rather than only the body comment: `ProductVariantService` carries `@docsCategory services`, so this signature is generated into the docs and `checkChannel` appears there undescribed.

```suggestion
    /**
     * @description
     * Soft-deletes the ProductVariant(s) with the given id(s).
     *
     * The lookup is scoped to the active Channel, so an id which does not belong to
     * `ctx.channelId` throws an {@link EntityNotFoundError}. Previously an unknown or
     * out-of-channel id was silently reported as deleted.
     *
     * @param checkChannel - Set to `false` only for trusted internal cascades which have
     * already verified the parent entity's Channel, such as `ProductService.softDelete`.
     * A global Product deletion must reach every one of its variants, including any which
     * were individually removed from the active Channel.
     */
    async softDelete(
```

packages/core/src/service/services/product-variant.service.ts
```diff
@@ -773,11 +773,31 @@ export class ProductVariantService {
...
+        // with the channel-scoped lookups used elsewhere (e.g. `ProductService.softDelete`).
+        // The check is skipped for the trusted product-deletion cascade (see `ProductService.softDelete`),
+        // which has already verified the parent Product's channel and must cascade to every one of
+        // its variants regardless of their individual channel membership.
+        let variants: ProductVariant[];
+        if (checkChannel) {
+            variants = await this.connection.findByIdsInChannel(ctx, ProductVariant, ids, ctx.channelId, {});
+            const missingId = ids.find(candidate => !variants.some(v => idsAreEqual(v.id, candidate)));
+            if (missingId != null) {
+                throw new EntityNotFoundError('ProductVariant', missingId);
```
biggamesmallworld, line 800: This throw is what changes for existing installs, and combined with the `@Transaction()` on `deleteProductVariants` it is what makes the bulk mutation all-or-nothing. Both are the intended behaviour and neither is covered.

Add to `packages/core/e2e/graphql/shared-definitions.ts`, next to `deleteProductVariantDocument`:

```ts
export const deleteProductVariantsDocument = graphql(`
    mutation DeleteProductVariants($ids: [ID!]!) {
        deleteProductVariants(ids: $ids) {
            result
            message
        }
    }
`);
```

Then, in the `cross-channel ProductVariant deletion protection` describe, importing `deleteProductVariantsDocument` alongside `deleteProductVariantDocument`:

```ts
it(
    'deleting an unknown variant id throws rather than reporting DELETED',
    assertThrowsWithMessage(async () => {
        adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
        await adminClient.query(deleteProductVariantDocument, { id: 'T_999999' });
    }, 'No ProductVariant with the id "999999" could be found'),
);

// deleteProductVariants is @Transaction()-wrapped, so a batch containing an id outside the
// active channel is rejected in full rather than partially applied.
it('deleteProductVariants discards the whole batch when one id is not in the active channel', async () => {
    adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);

    // A variant which lives only in the default channel, plus one assigned to T_2 as well.
    const { createProduct } = await adminClient.query(createProductDocument, {
        input: {
            translations: [
                {
                    name: 'OSS-666 Bulk Product',
                    slug: 'oss-666-bulk',
                    description: '',
                },
            ],
        },
    });
    const { createProductVariants } = await adminClient.query(createProductVariantsDocument, {
        input: [
            {
                productId: createProduct.id,
                sku: 'OSS-666-BULK',
                optionIds: [],
                translations: [{ languageCode: LanguageCode.en, name: 'OSS-666 Bulk Variant' }],
            },
        ],
    });
    productVariantGuard.assertSuccess(createProductVariants[0]);
    const defaultOnlyId = createProductVariants[0].id;

    const { createProduct: sharedProduct } = await adminClient.query(createProductDocument, {
        input: {
            translations: [
                {
                    name: 'OSS-666 Bulk Shared Product',
                    slug: 'oss-666-bulk-shared',
                    description: '',
                },
            ],
        },
    });
    const { createProductVariants: sharedVariants } = await adminClient.query(
        createProductVariantsDocument,
        {
            input: [
                {
                    productId: sharedProduct.id,
                    sku: 'OSS-666-BULK-SHARED',
                    optionIds: [],
                    translations: [{ languageCode: LanguageCode.en, name: 'OSS-666 Bulk Shared Variant' }],
                },
            ],
        },
    );
    productVariantGuard.assertSuccess(sharedVariants[0]);
    const sharedId = sharedVariants[0].id;
    await adminClient.query(assignProductToChannelDocument, {
        input: { channelId: 'T_2', productIds: [sharedProduct.id] },
    });

    // A T_2 admin deletes both in one call. Only one of them is in T_2.
    await adminClient.asUserWithCredentials('deleter2@test.com', 'test');
    adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
    await expect(
        adminClient.query(deleteProductVariantsDocument, { ids: [sharedId, defaultOnlyId] }),
    ).rejects.toThrow('No ProductVariant with the id');

    // The batch was rejected in full: the variant the admin *could* have deleted survives.
    adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
    const { productVariants } = await adminClient.query(getProductVariantListDocument, {
        options: { filter: { id: { in: [sharedId, defaultOnlyId] } } },
    });
    expect(productVariants.items.map(v => v.id).sort()).toEqual([sharedId, defaultOnlyId].sort());
});
```

The second one is the more valuable of the two: it is the only test that would fail if the transaction wrapper were ever dropped from the resolver, which is what currently makes the guard hold across a batch.

packages/core/e2e/product-channel.e2e-spec.ts
```diff
@@ -1027,6 +1029,186 @@ describe('ChannelAware Products and ProductVariants', () => {
...
+            adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+        });
+
+        it(
+            'a second-channel admin cannot delete a variant that lives only in the default channel',
+            assertThrowsWithMessage(async () => {
+                await adminClient.asUserWithCredentials('deleter2@test.com', 'test');
+                adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
+                await adminClient.query(deleteProductVariantDocument, { id: defaultOnlyVariantId });
+            }, 'No ProductVariant with the id'),
```
biggamesmallworld: Nit. The expected message stops before the id, so this passes even if the server names a different variant.

```suggestion
            }, `No ProductVariant with the id "${defaultOnlyVariantId.replace(/^T_/, '')}" could be found`),
```

biggamesmallworld (changes requested):
Re-read at `6e5f22f4` against the feedback on this PR.

**Addressed** — the stringified-id Set is gone and `idsAreEqual` is used, consistent with the other seven uses in the file.

**Settled** — the positional `checkChannel` boolean stays, per the earlier decision.

**Settled** — `deleteProductVariants` going all-or-nothing is the behaviour we want. The resolver is `@Transaction()`, so one out-of-channel or unknown id rolls the whole batch back and the mutation returns `null` with a top-level error rather than `DeletionResponse[]`. That matches `deleteProducts`, and a partial cross-channel delete is worse than a failed one. No code change needed. Two consequences to write down rather than leave implicit:

1. The Behavioural note in the PR description covers the single mutation but not the bulk one. Add that `deleteProductVariants` now discards the entire batch if any id is unknown or outside the active channel, and that `product-variant-list-bulk-actions.ts:101` therefore no longer reaches its per-item `.message` path on failure.
2. Neither that nor the unknown-id throw has a test. The four new tests cover the block path and the cascade path, and they hold up under mutation, but the two behaviours that change for existing installs are untested. Runnable tests for both are in the inline comments; paste them in and they should give 35/35.

Two nits inline as well, both optional.

CI is green at this head on all four databases.

packages/core/src/service/services/product-variant.service.ts
grolmus: Added the docblock verbatim in `f57bc300f` and dropped the now-duplicated body comment.

packages/core/src/service/services/product-variant.service.ts
grolmus: Both tests added in `3d86d06ac` (plus `deleteProductVariantsDocument` in shared-definitions). The bulk one is skipped on sqljs — its transaction rollback is a no-op, so a partial delete would persist and mask the guarantee (same pattern as database-transactions.e2e-spec.ts). Verified on postgres: full 6-test batch green; sqljs: 34 + 1 skipped.

packages/core/e2e/product-channel.e2e-spec.ts
grolmus: Applied, with one change: `assertThrowsWithMessage` evaluates its message arg at test-registration time, when `defaultOnlyVariantId` (set in beforeAll) is still undefined — so the literal form threw `Cannot read properties of undefined`. Passed it as `() => \`...\`` instead (the helper already accepts a thunk), deferring the read to throw time.

grolmus:
All three addressed in `f57bc300f` + `3d86d06ac`, and the behavioural note in the description now covers the bulk all-or-nothing case and the `product-variant-list-bulk-actions.ts:101` consequence.

- **Docblock** added to `softDelete`; removed the duplicated body comment.
- **Tests**: both added (unknown-id throw + bulk discard), plus `deleteProductVariantsDocument`. The bulk test is `skipIf` sqljs, whose transaction rollback is a no-op — otherwise a partial delete persists and masks the guarantee (same pattern as `database-transactions.e2e-spec.ts`).
- **Message nit**: applied as a thunk — `assertThrowsWithMessage` evaluates the message at registration time, when `defaultOnlyVariantId` is still undefined, so the literal form threw.

Verified: postgres 35/35 (bulk included), sqljs 34 + 1 skipped.

biggamesmallworld (approved):
Re-read at `3d86d06a`. All three addressed: docblock, thunked message assertion, both tests plus `deleteProductVariantsDocument`, and the behavioural note now covers the bulk case.

Skipping the bulk test on sqljs is right — a no-op rollback would let it pass while the guarantee it asserts was broken. CI sets `DB` explicitly, so it still runs on postgres, mysql and mariadb. It is also the only test that fails if the `@Transaction()` wrapper ever leaves the resolver.

Two red checks, neither this PR's code. `dashboard e2e (3)` never started (HTTP 429 fetching `setup-bun`), so it just needs a re-run. `conventional commit format` is failing on the PR **title**: commitlint `subject-case` rejects `Enforce`. That gate landed on master in `3e182107f` after this PR was opened. Lowercase to `fix(core): enforce channel scope on ProductVariant soft-deletion`.

Approving.

PR #5051, opened 2026-07-29
feat(core): Encrypt secret custom fields and config args at rest

### LINKED ISSUE #2648: Ability to encrypt custom field / config arg data at rest
Some data which is stored as config args or custom fields might be sensitive and would benefit from being stored encrypted in the DB.
For instance, the API keys of a payment plugin.

I will investigate the feasibility of a new property on custom fields and configurable operation args, `encrypt: true` which uses some kind of symmetric encryption using a secret key (provided by an env var) to encrypt/decrypt these values on save/load.

Storing the values as env vars in the first place. This limits the utility of certain functions, e.g. we cannot then have channel-aware payment methods using different api keys for the same payment service.

michaelbromley:
## Summary

Adds a `secret: true` flag for custom fields and configurable-operation args. Values marked secret are **encrypted at rest** in the database and only returned in decrypted form via the API to callers permitted to see them (by default, holders of the new `ReadSecret` permission); everyone else receives a redaction placeholder.

Fixes #2648

## How it works

**Encryption at rest.** A configurable `EncryptionStrategy` (default: `DefaultEncryptionStrategy`, AES-256-GCM) encrypts secret values before they are stored. Custom-field values are encrypted via a TypeORM value transformer; config-arg values are encrypted centrally in `ConfigArgService.parseInput`. Ciphertext carries a versioned `enc:v1:` prefix, and legacy plaintext is passed through unchanged so enabling `secret` on existing data does not break it.

**Gating decrypted access, centrally.** Rather than wiring redaction into each resolver, a single field resolver on the shared `ConfigurableOperation` GraphQL type gates the args of *every* configurable operation in the API — core or custom, on core or custom entities — with no per-operation-type wiring. Custom-field values are gated in the generic custom-field resolver. Both consult a `SecretAccessStrategy` (default: `PermissionSecretAccessStrategy`, which requires `ReadSecret`). Its input is a discriminated union on `kind` (`customField` | `configArg`), so each context carries only the information actually available to it.

**Editing without re-entering the secret.** Because the API returns a placeholder, submitting that placeholder back on an update preserves the stored value; submitting a new value replaces it; on a create the placeholder is rejected.

**Bootstrap safety.**
- If a `secret` field/arg is configured but no usable key is available, the server refuses to start, and the error shows the exact config to add.
- The first time a key is used against a database, a key-check value is stored via the settings store and verified on every startup. If the configured key later differs (e.g. a database was restored into an environment with a different key), the server fails to start with a clear message instead of failing later with scattered decryption errors.

**Config-driven secret.** The strategy does not read `process.env` itself. Following the framework convention, the secret is read in the config and passed in explicitly:

```ts
systemOptions: {
    encryptionStrategy: new DefaultEncryptionStrategy({ secret: process.env.VENDURE_ENCRYPTION_KEY }),
}
```

## Dashboard

Secret custom fields and secret configurable-operation args render by default as a masked, revealable password input in the Dashboard, so it is visually clear that the value is sensitive even to a user who is permitted to read it. The same masking is applied to the read-only value summaries, so a redacted or decrypted secret is never shown in plain text there either. An explicit `ui.component` override still takes precedence over this default.

The masked input reuses the existing `password-form-input` component, and the default is resolved centrally in `resolveInputComponentId`, shared between the form control and the operation summary. For this to work the Dashboard's custom-field and config-arg definition queries now select the `secret` flag.

When the current user is not permitted to read a secret, the API returns the redaction placeholder rather than the value. The Dashboard recognises this placeholder and renders the field as a read-only "Hidden" input with no reveal control, so the internal placeholder is never shown to the user. The placeholder remains the field's form value, so saving the entity still round-trips it and preserves the stored secret unchanged.

## Security hardening

An adversarial review of the design surfaced several issues, now addressed:

- **Key-check exposure.** The bootstrap key-check settings-store entry was readable by any authenticated user, which handed out an offline brute-force oracle for the secret. Read access is now restricted to the SuperAdmin.
- **Config-arg fail-open.** The systemic resolver redacted based on whether the stored value *looked* encrypted, so a secret arg holding legacy plaintext leaked to callers without `ReadSecret`. Redaction is now driven by the arg's `secret` definition, so such values are redacted regardless of their stored form.
- **Weak key derivation.** Key derivation moved from a single unsalted SHA-256 to scrypt (memory-hard), and a startup warning is emitted for a short/low-entropy secret.

Known follow-up (not in this PR): binding each ciphertext to its location with AAD would prevent an attacker with database *write* access from transplanting one secret's ciphertext into another field. Doing this at row-level granularity requires moving encryption into a layer that knows the entity's identity, so it is left as a separate hardening task.

## Limitations

- Key rotation (re-encrypting existing data under a new key) is not supported; the key must remain stable. Moving data between environments therefore requires the same key, or scrubbing the secret columns in the dump. This is documented in the new guide.
- Decrypted values are available to all server-side code (events, jobs, plugins); redaction applies only to the GraphQL API. Take care not to re-leak secrets via logs, exports, job payloads or the search index.

## Testing

- New unit specs for the encryption strategy (round-trip, random IV, legacy passthrough, key mismatch, `isEncrypted`, does-not-read-env) and the key-check verifier (first-boot write, same-key pass, different-key fail, graceful skip).
- New e2e coverage for secret custom fields and secret config args, including a cross-operation-type case (collection filter) proving the systemic redaction works for a type with no bespoke wiring, and a legacy-plaintext redaction case.
- Full related e2e sweep (collection, promotion, shipping, payment method, configurable operation, entity duplicator, custom fields) passes.

## Docs

New "Secret fields" developer guide covering configuration, the access strategy, editing behaviour, moving data between environments, and limitations.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5051"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787927191&installation_model_id=20193&pr_number=5051&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5051&signature=9317ab004d6a88915c89ef4a90dd78255cb86b6656aef0410053d7d9e2075210"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


dlhck (changes requested):
Requesting changes for four correctness and authorization issues in the secret-field implementation.

1. Repeated configurable operations are matched to previous values only by code. Duplicate promotion conditions/actions and collection filters can therefore preserve the first instance’s encrypted secret into later instances. Preserve by list position or stable instance identity.

2. SecretAccessStrategy receives the custom-fields wrapper rather than the documented owning VendureEntity, preventing reliable entity-aware authorization.

3. secret: true is silently accepted on relation custom fields because validation occurs only in the non-relation branch; such fields are neither encrypted nor redacted.

4. Legacy plaintext is passed unconditionally to EncryptionStrategy.decrypt() in custom-field transformers and configurable-operation execution, although the public strategy contract only requires decrypting values produced by encrypt(). Check isEncrypted() first.

packages/core/src/service/services/promotion.service.ts
```diff
@@ -177,12 +177,20 @@ export class PromotionService {
                 p.priorityScore = this.calculatePriorityScore(input);
                 if (input.conditions) {
                     p.conditions = input.conditions.map(c =>
-                        this.configArgService.parseInput('PromotionCondition', c),
+                        this.configArgService.parseInput(
+                            'PromotionCondition',
+                            c,
+                            promotion.conditions.find(previous => previous.code === c.code),
```
dlhck: Repeated configurable operations are matched to their previous value only by `code`. Duplicate promotion conditions/actions are allowed, so when two entries share a code, every placeholder-bearing entry preserves the first entry’s encrypted secret and silently overwrites the others. Collection filters have the same issue. Please preserve by list position or introduce stable instance identity rather than using `.find()` by code.

packages/core/src/api/config/generate-resolvers.ts
```diff
@@ -311,7 +312,20 @@ function generateCustomFieldResolvers(
...
-                    return source[fieldDef.name];
+                    const value = source[fieldDef.name];
+                    if (fieldDef.secret === true && value != null) {
+                        const secretAccessStrategy = configService.systemOptions.secretAccessStrategy;
+                        const canReveal = secretAccessStrategy
+                            ? await secretAccessStrategy.canAccessSecret(ctx, {
+                                  kind: 'customField',
+                                  entityType: entityName,
+                                  fieldName: fieldDef.name,
+                                  entity: source,
```
dlhck: `source` here is not the owning `VendureEntity`; the parent custom-field resolver constructs it as `{ ...source.customFields, __entityId__ }`. This breaks the public `SecretAccessInput` contract and prevents entity-aware authorization strategies from inspecting fields such as the entity ID, channel, or seller. Please retain and pass the actual owning entity.

packages/core/src/entity/register-custom-entity-fields.ts
```diff
@@ -100,6 +101,40 @@ function registerCustomFieldsForEntity(
                     ) {
                         options.precision = 6;
                     }
+                    if (customField.secret === true) {
```
dlhck, line 183: The `secret` compatibility validation lives entirely inside the non-relation branch. Consequently, `{ type: "relation", secret: true }` is accepted, registered as an ordinary relation, and neither encrypted nor redacted by the relation resolver. Please validate unsupported secret field types before branching on `relation`.

packages/core/src/entity/value-transformers.ts
```diff
@@ -12,3 +14,38 @@ export class DecimalTransformer implements ValueTransformer {
...
+            return value;
+        }
+        return this.strategy().encrypt(String(value));
+    }
+
+    from(value: any): any {
+        if (value == null || value === '') {
+            return value;
+        }
+        return this.strategy().decrypt(String(value));
```
dlhck: This unconditionally calls `decrypt()` for legacy plaintext values. The public `EncryptionStrategy` contract only requires `decrypt()` to accept values produced by `encrypt()`, while the guide promises that enabling `secret` on existing plaintext data continues to work. `argsArrayToHash()` has the same issue for config args. Please check `isEncrypted()` first and return plaintext unchanged when false.

michaelbromley:
Thanks @dlhck — all four addressed.

**1. Duplicate operations matched by code only** (`0a6c2b1b0`)
Added `ConfigArgService.parseInputList`, which matches each input to its previous value by code *and* by position among entries that share a code (consuming previous entries in order). Duplicate promotion conditions/actions and collection filters now each preserve their own secret rather than all matching the first. Promotion and collection update paths use it. Regression test added: two collection filters sharing a code, each resubmitting the placeholder, each preserving its own distinct secret.

**2. Custom-fields wrapper passed instead of the owning entity** (`e34b3d0b2`)
The owning entity is now retained on the customFields wrapper (`__entity__`) alongside `__entityId__`, and passed as `input.entity`, so entity-aware strategies receive the real `VendureEntity` (id, channel, etc.). Test added: a capturing strategy asserts the owning entity — not the wrapper — is provided.

**3. `secret: true` silently accepted on relation fields** (`20d17fad9`)
The type/list validation moved ahead of the relation/column branch, so `{ type: 'relation', secret: true }` (and any other unsupported type) is now rejected at registration instead of being registered without encryption or redaction.

**4. Legacy plaintext passed to `decrypt()`** (`54dea2dc6`)
The custom-field transformer and `argsArrayToHash()` now check `isEncrypted()` first and return the plaintext unchanged otherwise, so `decrypt()` only ever receives values produced by `encrypt()`. This keeps custom strategies from having to tolerate non-ciphertext and preserves the guide's promise that enabling `secret` on existing plaintext data keeps working. Unit test added with a strict strategy that throws if handed a non-ciphertext value.

All secret and related e2e suites (config-arg, custom-field, collection, promotion) and the unit specs pass. Ready for another look.

dlhck (approved):

gabriellbui:
## Review: secret-field write paths + access-strategy entity plumbing

Two issues found (one blocking). Both are new — neither overlaps the four points already fixed in this round.

---

### 🔴 HIGH — Secret custom fields are silently corrupted on write paths the placeholder interceptor doesn't recognize

**Blocking.** Ordinary edits by operators *without* `ReadSecret` permanently overwrite stored secrets, with no error.

**Root cause.** The "resubmit the placeholder to preserve the stored value" contract depends entirely on `CustomFieldProcessingInterceptor` stripping `REDACTED_SECRET_PLACEHOLDER` before it reaches the DB. But the interceptor only acts on input types whose name it *generated* from a fixed pattern. In the constructor (`custom-field-processing-interceptor.ts:42-45`):

```ts
Object.keys(configService.customFields).forEach(entityName => {
    this.createInputsWithCustomFields.add(`Create${entityName}Input`);
    this.updateInputsWithCustomFields.add(`Update${entityName}Input`);
});
```

`hasCustomFields()` (`:84-90`) is then an exact set-membership check plus one hardcoded case. Any mutation that carries an entity's custom fields under a **differently-named input** is invisible to it, so the placeholder is never stripped — it flows into `patchEntity(...customFields)`, the `EncryptedFieldTransformer.to()` encrypts the literal placeholder string, and the real secret is gone.

There's a second layer that fails for the same reason: even if such an input reached `handleSecretCustomFields()` (`:117-138`), it re-derives the entity name by slicing prefix/suffix (`getEntityNameFromInputType`, `:249-258`), so `UpdateActiveAdministratorInput` → `"ActiveAdministrator"`, which isn't a key in `configService.customFields`, and it early-returns.

**Confirmed corrupting paths:**

1. **`updateActiveAdministrator`** — input type is `UpdateActiveAdministratorInput`, not `UpdateAdministratorInput`, so it never enters the generated set. Custom fields are attached to it separately in `graphql-custom-fields.ts:403` (`extend input UpdateActiveAdministratorInput { customFields: UpdateAdministratorCustomFieldsInput }`) — a path the interceptor's generator knows nothing about. The resolver forwards to `administratorService.update` → `patchEntity(administrator, input)` (`administrator.service.ts:170`). **Any admin lacking `ReadSecret` who saves their own profile destroys their secret custom fields** — the common case.
2. **`modifyOrder`** — input type `ModifyOrderInput`; order-level `customFields` applied at `order-modifier.ts:652`. Not in the set.
3. **OrderLine custom fields** (`addItemToOrder` / `adjustOrderLine`) — `OrderLineCustomFieldsInput` *is* special-cased in `hasCustomFields`, but `handleSecretCustomFields` bails: `getEntityNameFromInputType('OrderLineCustomFieldsInput')` returns the string unchanged (no key), and the input *is* the customFields object rather than a `{ customFields }` wrapper, so the `inputVariable?.customFields` guard also fails.

**Repro (path 1):** Add `{ name: 'apiToken', type: 'string', secret: true }` to `Administrator`. As an admin with `UpdateAdministrator` but not `ReadSecret`:

```graphql
query { activeAdministrator { customFields { apiToken } } }
# → "__vendure_secret_unchanged_v1__"

mutation {
  updateActiveAdministrator(input: {
    firstName: "x",
    customFields: { apiToken: "__vendure_secret_unchanged_v1__" }
  }) { id }
}
```

Stored value is now `encrypt("__vendure_secret_unchanged_v1__")`. The real token is silently, irreversibly lost.

**Why it wasn't caught:** the e2e suites only exercise `Product` (`CreateProductInput` / `UpdateProductInput`) — exactly the shape the generator handles. The feature works on the one entity that's tested and corrupts data on the ones that follow different input-naming conventions.

**Fix direction:** the pattern generator can't discover these inputs on its own, so register the alias types → owning entity explicitly in *both* `hasCustomFields` and the entity resolution used by `handleSecretCustomFields` (`UpdateActiveAdministratorInput` → `Administrator`, `ModifyOrderInput` → `Order`, `OrderLineCustomFieldsInput` → `OrderLine`), and handle the unwrapped order-line shape. Worth adding e2e coverage for a non-`Product` secret field (Administrator profile self-edit is the sharpest regression test).

---

### 🟡 LOW — Custom access strategies can receive `entity: undefined`

The owning entity is attached to the resolver source only when an entity-level custom-field resolver runs (`generate-resolvers.ts:670-678`, which sets `__entity__`). For GraphQL types that expose a structured `customFields` field but resolve it via GraphQL's default resolver — e.g. some plugin-defined types, or variants not wired through the special-cased resolvers — `source` has no `__entity__`, so a custom `SecretAccessStrategy` inspecting `input.entity` receives `undefined`.

The built-in `PermissionSecretAccessStrategy` ignores `entity`, so redaction itself still holds and there's no leak with the default config — this only bites operators who write entity-aware strategies (checking entity id, channel, seller). Partly suspected: the exact set of affected types wasn't enumerated; what's confirmed is that `__entity__` presence depends on the entity-level resolver running.

biggamesmallworld (commented):
Reviewed at 54dea2d, re-checked at 23cb84b — everything below still applies. dlhck's four points check out as fixed. Below is only what isn't already covered by dlhck or @gabriellbui.

Lint output and the unit specs (14/14 pass) were run locally; everything else was read from the source at 23cb84b.

## Blocking

**1. `previewCollectionVariants` and the order-testing service throw on the placeholder**

`collection.service.ts:494` calls `getCollectionFiltersFromInput(input)` with no `previous`, so when the UI posts back current form values containing `REDACTED_SECRET_PLACEHOLDER`, `processSecretArgs` hits `if (!previousArg) throw new UserInputError('error.secret-value-required')`. Collection preview is broken for any collection with a secret filter arg. Line 576 (update) got the fix, 494 was missed.

Same at `order-testing.service.ts:58-59` — `testShippingMethod` / `testEligibleShippingMethods` parse a caller-supplied checker/calculator with no `previous`. Test-shipping-method breaks identically.

Both are read-only previews and shouldn't touch the secret machinery at all. That's the underlying issue: `parseInput` does double duty as "coerce for immediate use" and "prepare for persistence", and only the second needs it.

**2. Lint fails**

```
configurable-operation-entity.resolver.ts
  55:39  error  Forbidden non-null assertion   @typescript-eslint/no-non-null-assertion
  55:39  error  This assertion is unnecessary  @typescript-eslint/no-unnecessary-type-assertion
```

`encryptionStrategy!.decrypt(...)` — and one of the two errors says the assertion does nothing, which checks out: `RuntimeVendureConfig.systemOptions` is `Required<SystemOptions>`, so `encryptionStrategy` is already non-optional there. Also six new `import/order` warnings (`value-transformers.ts:1,3` and `service.module.ts:32,35,36,37`) — none of them present on `minor`.

**3. `isEncrypted ||` contradicts the class docblock, and opens a corruption path**

`configurable-operation-entity.resolver.ts:33-34`:

```ts
const isEncrypted = arg.value != null && encryptionStrategy?.isEncrypted(arg.value) === true;
const isSecret = isEncrypted || this.configArgService.hasSecretArg(operation.code, arg.name);
```

The docblock ten lines above says an arg is secret "if its definition declares `secret: true`". I assume the OR is there to handle orphaned ciphertext after someone removes `secret: true` — if so, say that, because the OR has a cost: any non-secret string arg whose value happens to start with `enc:v1:` gets redacted for unprivileged callers and throws a GCM auth failure for privileged ones, killing the query. And since `processSecretArgs` won't preserve it (the def isn't secret), the unprivileged user's next save writes the literal `__vendure_secret_unchanged_v1__` into the arg. Gate on `hasSecretArg` and handle the orphan case explicitly.

**4. `SecretAccessInput` omits the operation `code`** — public API, awkward to change later

The `configArg` variant carries `entityType: string | undefined` and `field: string | undefined`, both derived from `info.path` walking, but not `operation.code`, which is always defined and is the only thing that identifies which operation's arg this is. "Reveal `apiKey` on the Stripe handler but not PayPal" isn't expressible today, and two Collection filters both declaring `apiKey` produce identical `{ entityType: 'Collection', field: 'filters' }`. Adding `code` also raises the question of whether `deriveOwner` is still worth the path archaeology.

Related: `hasSecretArg` searches every def type for a matching code, but codes are only unique within a type — a `CollectionFilter` and a `PromotionCondition` sharing a code will cross-contaminate redaction. It's also unmemoized (nested loop per arg, per operation, per response) when all the config is available at construction time. A `Set<'defType:code:argName'>` built in the constructor fixes both, plus an early bail when no secret args are configured at all.

## Should fix

**5. `settingsStoreService.set()` return value ignored** — `encryption-key-verifier.service.ts:658`. `set` catches internally and returns `{ result: false, error }` rather than throwing, so if the first-boot key write fails the mismatch guard silently never arms. The spec's fake always returns `{ result: true }`, so it can't catch this. Check `result` and log.

**6. `decrypt()` failures fail the whole list query** — no catch in `EncryptedFieldTransformer.from()` or the resolver. `ConfigArg.value` is `String!`, so one row corrupted by a bad restore or manual SQL edit 500s the entire `paymentMethods`/`collections` page. The key verifier covers the global key-mismatch case; single-row corruption isn't. Log the entity id and return the placeholder.

**7. Test coverage is on the easy half**

- `value-transformers.spec.ts` has three tests, all calling `from()`. `to()` — the encrypt-on-write path, i.e. whether secrets actually reach the DB as ciphertext — has no unit test. Nor does the `strategy()` throw-when-unconfigured branch.
- No GCM tamper test. Wrong-key is covered; flipping a byte of ciphertext or auth tag isn't, and integrity is the whole point of AEAD.
- `strategy.decrypt('enc:v1:')` → `split(':')` gives undefined parts → `Buffer.from(undefined, 'base64url')` throws a raw `TypeError` instead of the helpful key-mismatch message.
- `deriveOwner`'s `info.path` walking has zero coverage. The e2e `CapturingSecretAccessStrategy` only ever captures the `customField` kind — no test asserts a derived `entityType`/`field` for a `configArg`. That's the code most likely to start silently returning `undefined` on a graphql-js upgrade.
- `parseInputList` has e2e coverage but no unit test, despite being a pure function over two arrays and having been got wrong once already.
- `custom-field-secret.e2e-spec.ts:173-175` asserts `expect(entity?.id).toBeTruthy()` and `expect(...customFields).toBeDefined()` for a test titled "passes the owning entity (not the customFields wrapper)". The id is known to be `T_1`. As written it passes for any object with a truthy id and a `customFields` key — including the wrapper it's meant to rule out. (`23cb84b91` fixed the related vacuous-pass problem; the assertions themselves are still loose.)
- The "does not read the secret from the environment" test sets `VENDURE_ENCRYPTION_KEY`, which the class never references, and mutates global `process.env`. It asserts the absence of code nobody wrote, and would still pass if env-reading were added under another name.

**8. Dashboard: `secret: true` is unusable as shipped** — a secret string field renders as a plain text input containing the literal `__vendure_secret_unchanged_v1__`, on screen. `password-form-input` already exists in `packages/dashboard/src/lib/components/data-input/`, but there's no default `ui.component` mapping and no docs note. Fine as a follow-up, but the PR body's "Editing without re-entering the secret" section reads as though the editing experience ships.

**9. Reject foreign-version placeholders** — the preserve-on-update mechanism depends on an exact string match with `__vendure_secret_unchanged_v1__`. Introduce `_v2_` and any older client still submitting `_v1_` is treated as a new value and overwrites the stored secret with the literal placeholder. Since it's versioned, reject anything matching `/^__vendure_secret_unchanged_/` that isn't the current constant.

## Smaller

- Secret-field validation lives in `register-custom-entity-fields.ts` (registration time) and `config.module.ts` (bootstrap) with four near-identical error strings and no sharing. The config-arg version reaches its data by casting past the type — `argDef as { type?: string; list?: boolean; secret?: boolean }` at `config.module.ts:35`, and `arg as { secret?: boolean }` at :69. `ConfigArgs` is ours — worth fixing the type rather than casting past it.
- `unique: true` and an explicit `length` on a secret field throw, but a `defaultValue` is silently `delete`d from the column options. Same class of config error, so it should throw too.
- `SecretAccessInput.entity` is typed `VendureEntity`, but `generate-resolvers.ts:232-236` assigns the same custom-field resolver to `ShippingMethodQuote` and `PaymentMethodQuote`, where `source[__entity__]` is a plain quote object. Widen the type or exclude quotes.
- `error.secret-value-required` reads "A value must be provided for the secret argument" and is reused verbatim for custom fields — `custom-field-secret.e2e-spec.ts:690` asserts on a field being called an argument. Needs a second key. `secret-value-required` is also out of alphabetical order in `en.json`.
- `payment-method-entity.resolver.ts` is now `@Resolver('PaymentMethod')` around an empty class with an unused injected service.
- `RECOMMENDED_SECRET_LENGTH = 24` counts characters while the warning says "high-entropy"; 24 `a`s passes.
- `output.push({ ...arg })` for non-secret args vs `{ name, value }` for secret ones — worth picking one shape.

## Credit where due

Gating redaction on the single shared `ConfigurableOperation` type rather than wiring nine operation types individually is the right call, as are the versioned `enc:v1:` prefix and the legacy-plaintext passthrough. The comments are unusually good — `KEY_DERIVATION_SALT` explaining *why* a fixed salt (sync bootstrap), the `Permission.SuperAdmin` note on the brute-force oracle, and "only values produced by encrypt() may be passed to decrypt()" all explain non-obvious decisions rather than restating the code.

michaelbromley:
Thanks both — this was a genuinely valuable pass. Addressed across four commits (`66a2b69`, `4d6c1fa`, `6ce4cb9`, `246f330`). Full secret + related e2e sweep (custom-fields, permissions, relations, order-line, administrator, collection, promotion, shipping, configurable-operation, duplicate-entity) is green, lint has no errors.

## @gabriellbui

**HIGH — secret custom fields corrupted on alias write paths.** Fixed (`66a2b69`). The interceptor now maps *every* custom-field-carrying input to its entity — the standard `Create/Update<Entity>Input` plus the aliases (`UpdateActiveAdministratorInput`, `ModifyOrderInput`, `RegisterCustomerInput`, `UpdateOrderAddressInput`, and the order-line inputs) — and strips the placeholder on all of them, handling the unwrapped order-line shape. The placeholder strip now runs as its own pass over all these inputs, independent of the (narrower) defaults/validation gating, so it can't be missed. Regression test added: an Administrator secret field preserved through a self-edit via `updateActiveAdministrator` (the common case). I verified it fails without the fix.

**LOW — `entity: undefined` for default-resolved types.** The `customField` variant's `entity` is now typed `VendureEntity | undefined`, documented as absent for types whose `customFields` use the default resolver (quotes, some plugin types).

## @biggamesmallworld

1. **preview/test throw on the placeholder** — fixed (`4d6c1fa`). Added `parseInputForExecution`, which validates/orders args without touching the secret machinery; `previewCollectionVariants` and `testShippingMethod` use it. Your framing was right — `parseInput` was doing double duty; this splits "coerce for immediate use" from "prepare for persistence".
2. **Lint** — the non-null assertion is gone (narrowed properly), and the three `import/order` warnings are fixed.
3. **`isEncrypted ||` corruption path** — fixed (`6ce4cb9`). Redaction is now gated purely on the `secret` definition (`hasSecretArg`), never on whether the value looks encrypted. Docblock updated to describe the orphaned-ciphertext case (passed through until re-saved).
4. **`code` on `SecretAccessInput` + `hasSecretArg`** — `code` added (always defined). `hasSecretArg` is now an O(1) lookup against a `Set` built at construction with an early bail; a comment notes the cross-type code-uniqueness assumption and that a collision fails closed.
5. **`set()` result ignored** — now checked and logged (`6ce4cb9`).
6. **decrypt failures fail the whole query** — the transformer and the resolver now catch a single undecryptable value, log it, and fall back to the placeholder rather than 500-ing the list.
7. **Tests** — added: transformer `to()` / unconfigured / undecryptable-value; GCM tamper + malformed-ciphertext; config-arg strategy input asserts `code` + derived `entityType`/`field` (deriveOwner coverage); strengthened the owning-entity assertion to rule out the wrapper. `parseInputList` is covered by the duplicate-preservation e2e (verified red without the fix); I didn't add a separate unit test.
8. **Dashboard** — agreed, follow-up. I'll soften the PR body so "Editing without re-entering the secret" doesn't imply the input renders as a password field yet.
9. **Foreign-version placeholders** — rejected on write for both custom fields and config args via `isForeignSecretPlaceholder`.

**Malformed ciphertext** now throws a clear "not a well-formed ciphertext" error instead of a raw `TypeError`.

**Smaller:** `defaultValue` on a secret field now throws; the empty `PaymentMethodEntityResolver` is removed; the config-arg error message is split from the custom-field one (`secret-value-required` vs `secret-custom-field-value-required`, alphabetised in `en.json`); arg output shape unified.

**Not yet done (deferred):** sharing the four near-identical validation error strings, the `ConfigArgs` cast in `config.module`, and `RECOMMENDED_SECRET_LENGTH` measuring length rather than entropy — happy to fold these in, just kept this pass focused on correctness.

docs/docs/guides/developer-guide/secret-fields/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+Some data stored by Vendure is sensitive — payment-gateway API keys, webhook signing secrets and
+similar. The `secret` flag lets you mark a custom field or a config arg so that its value is
+**encrypted in the database** and is **only returned in decrypted form via the API to users who are
+permitted to see it**.
+
+## Enabling encryption
+
+Encryption requires a stable secret key. Configure an [`EncryptionStrategy`](/reference/typescript-api/configuration/encryption-strategy/)
+in your VendureConfig and pass it the secret. The strategy does not read the environment itself —
```
michaelbromley: "The strategy does not read the environment itself —
following Vendure's convention, you read the environment variable in your config and pass the value
in explicitly:"

waffle. Just say "the value is passed in to the constructor of the strategy"

docs/docs/guides/developer-guide/secret-fields/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+args: {
+    apiKey: { type: 'string', secret: true },
+},
+```
+
+This works for any configurable operation — payment method handlers and eligibility checkers,
+shipping calculators and checkers, promotion conditions and actions, collection filters, and any
+custom `ConfigurableOperationDef` you define yourself. The encryption, redaction and
+preservation-on-edit behaviour is handled centrally, so a custom operation does not need to
+implement any of it. `secret` is only supported on non-list `string` args.
```
michaelbromley: > `secret` is only supported on non-list `string` args.

Is this a result of the design? Is it not possible?

PR #5035, opened 2026-07-24
perf(core): Add batch loading for collection product variants

RubenDarioGuerreroNeira:
## Description

This PR fixes an N+1 query problem when fetching `productVariants` for
collections. Previously, each collection resolved its variants with a
separate query, causing performance to degrade linearly with the number
of collections returned in a single request.

This adds batch loading: variants for all requested collections are
fetched in a single query and then grouped in memory, before applying
per-collection pagination (skip/take).

This PR is a follow-up to #4636, split out per maintainer feedback to
keep concerns isolated. It complements #5034, which addresses ManyToOne
filter handling for a related N+1 pattern.

### What changed

- **`collection.service.ts`**: Added `getProductVariantsForCollections` for batch-loading variants across multiple collections in one query. The method fetches all matching variants in a single DB query (using `ignoreQueryLimits`), groups them by `collection.id` in memory, and applies per-collection `skip`/`take` pagination after grouping — so a global limit never truncates later collections.
- **`collection.resolver.ts`**: When the `productVariants` field is in the GraphQL selection, the resolver now pre-caches a `Promise<Map<collectionId, ProductVariant[]>>` via `RequestContextCacheService` (same pattern already used for `productVariantCount`), so field resolvers consume the batched result instead of issuing N queries.
- **`constants.ts`**: Added `CacheKey.CollectionVariants` and `COLLECTION_VARIANTS_CACHE_RELATIONS` (`['taxCategory']`) for the new cache entry and default relations.
- **`tax-rate.service.ts`**: Fixed a bug where filtering tax rates by `categoryId` mapped the filter to `customPropertyMap.zoneId` instead of `customPropertyMap.categoryId`, causing the category filter to silently not apply.
- **`list-query-builder.ts`**: Minor formatting fix in `escapeTablePath` helper (no behavioral change).
- **`many-to-one-filters.e2e-spec.ts`** (new): Added e2e coverage for: many-to-one filter creation, batch loading correctness (variants grouped by collection), per-collection pagination behavior, empty-input handling, and overlapping collections (variant in multiple collections).

### Why

When querying `collections { productVariants { ... } }`, the admin API
resolved `productVariants` per collection with a separate DB query each.
A request returning 20 collections executed 20+ variant queries. With
batch loading, all variants are fetched in **1 query** regardless of
collection count, then grouped and paginated in memory.

Additionally, the `tax-rate.service.ts` fix corrects a silent bug where
the `categoryId` filter on tax rates was mapped to the wrong property
[...]

PR #5055, opened 2026-07-29
fix(dashboard): Use updateActiveAdministrator for profile page

### LINKED ISSUE #5037: Dashboard profile page saves via updateAdministrator, so admins without the UpdateAdministrator permission cannot edit their own profile
**Describe the bug**

The Dashboard's profile page (`/profile`) saves changes using the `updateAdministrator` mutation, which is gated by `Permission.UpdateAdministrator`. As a result, any administrator whose role does not include `UpdateAdministrator` cannot edit their own first name, last name, email address, or password — the page loads and the form is editable, but saving fails with a permission error.

The Admin API already has a mutation intended for exactly this case: `updateActiveAdministrator`, which is gated by `Permission.Owner` and resolves the target id from `ctx.activeUserId`, so it can only ever modify the calling user's own administrator. The Dashboard does not use it anywhere.

Note the asymmetry within the same page: the `activeAdministrator` **query** that loads the page *is* correctly gated by `Permission.Owner`, so the page renders fine for every logged-in admin. Only the save path requires the elevated permission.

Relevant source (v3.6.5):

- `packages/dashboard/src/app/routes/_authenticated/_profile/profile.graphql.ts` — declares `updateAdministratorDocument` using `updateAdministrator`.
- `packages/dashboard/src/app/routes/_authenticated/_profile/profile.tsx:51` — passes it as `updateDocument` to `useDetailPage`.
- `packages/core/src/api/resolvers/admin/administrator.resolver.ts` — `updateActiveAdministrator` is `@Allow(Permission.Owner)`; `updateAdministrator` is `@Allow(Permission.UpdateAdministrator)`.

**To Reproduce**

1. Create a Role that does **not** include `UpdateAdministrator` — for example a shop-staff role with catalog/order permissions only, plus `Permission.Authenticated`.
2. Create an Administrator and assign that role.
3. Log in to the Dashboard as that administrator.
4. Go to `/profile` (user menu → Profile). The page loads correctly and shows the current first name, last name, email address and authentication methods.
5. Change the first name (or type a new password) and click **Update**.
6. A toast appears: `Failed to update profile` with the description `You are not currently authorized to perform this action`. Nothing is saved.

**Expected behavior**

Every logged-in administrator can edit their own profile, regardless of whether their role grants `UpdateAdministrator`. Editing one's own name, email address and password is self-service and should only require being authenticated — which is what `Permission.Owner` and the existing `updateActiveAdministrator` mutation express.

**Actual behavior**

The save is rejected with a `FORBIDDEN` error unless the administrator's role includes `UpdateAdministrator`. Because `UpdateAdministrator` is the permission that governs editing *other* administrators, the only way to let a user edit their own profile today is to also let them edit other administrators.

**Error logs**

GraphQL response from the `UpdateAdministrator` mutation:

```json
{
  "errors": [
    {
      "message": "You are not currently authorized to perform this action",
      "extensions": { "code": "FORBIDDEN" },
      "path": ["updateAdministrator"]
    }
  ],
  "data": null
}
```

Server log:

```
warn <timestamp> [GraphQL] You are not currently authorized to perform this action
```

- @vendure/core version: 3.6.5 (`@vendure/dashboard` 3.6.5)
- Nodejs version: 24.18.0
- Operating System (Windows/macOS/Linux): Windows 11 (also applies on Linux — this is not platform-specific)
- Package manager (npm/yarn/pnpm): npm 12.0.1

**Configuration**

No special configuration is required to reproduce — this reproduces on a default config. The only requirement is a Role that omits `UpdateAdministrator`, e.g.:

```typescript
await roleService.create(ctx, {
  code: 'shop-staff',
  description: 'Shop staff',
  channelIds: [channel.id],
  permissions: [
    Permission.Authenticated,
    Permission.ReadCatalog,
    Permission.UpdateCatalog,
    Permission.ReadOrder,
    Permission.UpdateOrder,
    // no Permission.UpdateAdministrator
  ],
});
```

**Minimal reproduction**

No repository needed — the two steps above (create a role without `UpdateAdministrator`, log in as a member of it, try to save `/profile`) reproduce it on a stock Vendure install.

**Suggested fix**

Point the profile page at the mutation that already exists for this purpose. In `profile.graphql.ts`:

```diff
-export const updateAdministratorDocument = graphql(`
-    mutation UpdateAdministrator($input: UpdateAdministratorInput!) {
-        updateAdministrator(input: $input) {
+export const updateActiveAdministratorDocument = graphql(`
+    mutation UpdateActiveAdministrator($input: UpdateActiveAdministratorInput!) {
+        updateActiveAdministrator(input: $input) {
             id
         }
     }
 `);
```

and update the import and `updateDocument` reference in `profile.tsx` accordingly.

One non-obvious detail: `setValuesForUpdate` in `profile.tsx` currently returns `id: entity.id`, and `useDetailPage` submits the form values verbatim as the mutation `input`. Since `UpdateActiveAdministratorInput` has no `id` field, that line has to be dropped as part of the change or every save fails schema validation:

```diff
         setValuesForUpdate: entity => {
             return {
-                id: entity.id,
                 firstName: entity.firstName,
```

`UpdateActiveAdministratorInput` exposes exactly the five fields the profile form edits (`firstName`, `lastName`, `emailAddress`, `password`, `customFields`) and deliberately omits `roleIds`, so this also removes the theoretical possibility of the profile form touching role assignments.

**Workaround**

We patch the two Dashboard source files in `node_modules/@vendure/dashboard` from a `postinstall` script, applying the diff shown above (swap the mutation document, update the import and `updateDocument`, and drop `id` from `setValuesForUpdate`). With that patch applied, `npm run build:dashboard` succeeds and administrators on restricted roles can edit their own profiles without being granted `UpdateAdministrator`.

**Additional context**

- Happens consistently, on every save attempt, for every role lacking `UpdateAdministrator`.
- Not triggered by any recent change on our side — it is the behaviour of a stock 3.6.5 Dashboard.
- Context for why this matters to us: we run a multi-tenant setup where each merchant store is a Channel, and merchant-facing roles are deliberately scoped. Granting those roles `UpdateAdministrator` purely to unblock self-service profile editing would give them a considerably broader capability than intended, so the profile page is effectively unusable for every non-superadmin user.

Ryrahul:
# Description

The profile page (`/profile`) uses the `updateAdministrator` mutation, which requires `Permission.UpdateAdministrator`. This means any administrator whose role does not include that permission cannot edit their own first name, last name, email, or password — the form loads fine (the query uses `activeAdministrator` gated by `Permission.Owner`), but saving fails with a `FORBIDDEN` error.

The Admin API already has `updateActiveAdministrator` for exactly this case — it's gated by `Permission.Owner` and resolves the target from `ctx.activeUserId`, so it can only ever modify the calling user's own record.

This PR switches the profile page to use `updateActiveAdministrator`.

Fixes #5037

## What changed

| File | Change |
|---|---|
| `profile.graphql.ts` | `updateAdministrator` → `updateActiveAdministrator` (mutation name + input type) |
| `profile.tsx` | Updated import, `updateDocument` reference, removed `id` from `setValuesForUpdate` (`UpdateActiveAdministratorInput` has no `id` field — the server infers it from the active user) |

## Why this is safe

- `updateActiveAdministrator` is `@Allow(Permission.Owner)` — any authenticated admin can call it
- The resolver resolves the target admin from `ctx.activeUserId` (cannot modify other admins)
- `UpdateActiveAdministratorInput` has no `roleIds` field — prevents privilege escalation
- `customFields` is dynamically added via `graphql-custom-fields.ts` `extend input`, so custom fields on the profile page continue to work
- The separate admin detail page (`/administrators/:id`) still uses `updateAdministrator` with `Permission.UpdateAdministrator` — unchanged

# Breaking changes

None.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5055"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787931250&installation_model_id=20193&pr_number=5055&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5055&signature=5d99f75da7c4ffe78f653138db3c0fc39b764a9ae37b315719c58aab7eccc348"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/settings/profile.spec.ts
+++ b/packages/dashboard/e2e/tests/settings/profile.spec.ts
@@ -1,5 +1,7 @@
 import { expect, test } from '@playwright/test';
 
+import { VendureAdminClient } from '../../utils/vendure-admin-client.js';
+
 test.describe('Profile', () => {
     test('should display profile page with form fields', async ({ page }) => {
         await page.goto('/profile');
@@ -86,5 +88,66 @@ test.describe('Profile', () => {
 
         // Reset to original value (don't submit)
         await firstNameInput.fill(originalValue);
+    });
+
+    // #5037 — an admin without UpdateAdministrator permission must still be able to
+    // edit their own profile (the page uses updateActiveAdministrator, gated by Owner)
+    test('should allow an admin without UpdateAdministrator permission to update own profile', async ({
+        page,
+        browser,
+    }) => {
+        const suffix = Date.now();
+        const emailAddress = `restricted-${suffix}@test.com`;
+        const password = 'test-password';
+
+        const client = new VendureAdminClient(page);
+        await client.login();
+        const { createRole } = await client.gql(
+            `mutation ($input: CreateRoleInput!) { createRole(input: $input) { id } }`,
+            {
+                input: {
+                    code: `restricted-${suffix}`,
+                    description: 'No admin permissions',
+                    permissions: ['ReadCatalog'],
+                },
+            },
+        );
+        await client.gql(
+            `mutation ($input: CreateAdministratorInput!) { createAdministrator(input: $input) { id } }`,
+            {
+                input: {
+                    firstName: 'Restricted',
+                    lastName: 'Admin',
+                    emailAddress,
+                    password,
+                    roleIds: [createRole.id],
+                },
+            },
+        );
+
+        // Fresh context so we browse as the restricted admin, not the superadmin
+        const context = await browser.newContext({ storageState: { cookies: [], origins: [] } });
+        const restrictedPage = await context.newPage();
+        await restrictedPage.goto('/login');
+        await restrictedPage.getByPlaceholder('Email').fill(emailAddress);
+        await restrictedPage.getByPlaceholder('Password').fill(password);
+        await restrictedPage.getByRole('button', { name: 'Sign in' }).click();
+        await expect(restrictedPage).not.toHaveURL(/\/login/, { timeout: 15_000 });
+
+        await restrictedPage.goto('/profile');
+        const firstNameField = restrictedPage.locator('[data-slot="field"]').filter({
+            has: restrictedPage.locator('[data-slot="field-label"]').getByText('First name', { exact: true }),
+        });
+        await firstNameField.getByRole('textbox').fill('Renamed');
+        await restrictedPage.getByRole('button', { name: 'Update' }).click();
+
+        await expect(
+            restrictedPage.locator('[data-sonner-toast]').filter({ hasText: 'Successfully updated profile' }),
+        ).toBeVisible({ timeout: 10_000 });
+
+        await restrictedPage.reload();
+        await expect(firstNameField.getByRole('textbox')).toHaveValue('Renamed');
+
+        await context.close();
     });
 });
--- a/packages/dashboard/src/app/routes/_authenticated/_profile/profile.graphql.ts
+++ b/packages/dashboard/src/app/routes/_authenticated/_profile/profile.graphql.ts
@@ -21,9 +21,9 @@ export const activeAdministratorDocument = graphql(`
     }
 `);
 
-export const updateAdministratorDocument = graphql(`
-    mutation UpdateAdministrator($input: UpdateAdministratorInput!) {
-        updateAdministrator(input: $input) {
+export const updateActiveAdministratorDocument = graphql(`
+    mutation UpdateActiveAdministrator($input: UpdateActiveAdministratorInput!) {
+        updateActiveAdministrator(input: $input) {
             id
         }
     }
--- a/packages/dashboard/src/app/routes/_authenticated/_profile/profile.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_profile/profile.tsx
@@ -20,7 +20,7 @@ import { useLocalFormat } from '@/vdb/hooks/use-local-format.js';
 import { Trans, useLingui } from '@lingui/react/macro';
 import { createFileRoute } from '@tanstack/react-router';
 import { toast } from 'sonner';
-import { activeAdministratorDocument, updateAdministratorDocument } from './profile.graphql.js';
+import { activeAdministratorDocument, updateActiveAdministratorDocument } from './profile.graphql.js';
 
 const pageId = 'profile';
 
@@ -49,11 +49,10 @@ function ProfilePage() {
     const { form, submitHandler, isPending, entity } = useDetailPage({
         queryDocument: activeAdministratorDocument,
         entityField: 'activeAdministrator',
-        updateDocument: updateAdministratorDocument,
+        updateDocument: updateActiveAdministratorDocument,
         pageId,
         setValuesForUpdate: entity => {
             return {
-                id: entity.id,
                 firstName: entity.firstName,
                 lastName: entity.lastName,
                 emailAddress: entity.emailAddress,


biggamesmallworld:
Thanks @Ryrahul,  checked out the branch and reviewed it. The change is correct:

- `updateActiveAdministrator` is `@Allow(Permission.Owner)` and resolves the target from `ctx.activeUserId`, so it can only ever touch the caller's own record.
- Dropping `id` from `setValuesForUpdate` is required, not cosmetic — `UpdateActiveAdministratorInput` has no `id`, and `useDetailPage` submits the form values verbatim, so nothing else needed changing.
- Custom fields still work: `addActiveAdministratorCustomFields` is wired unconditionally for the Admin API.
- This matches what the Angular admin-ui has always done, so it's a dashboard parity fix rather than a new pattern.

Two things before merge:

**1. The commit message says `Fixes #5035`** (a collection batch-loading perf issue). The PR body correctly says `#5037`. Please amend the commit so the squash body doesn't close the wrong issue.

**2. Needs a test.** The existing `packages/dashboard/e2e/tests/settings/profile.spec.ts` tests all run as superadmin, who already has `UpdateAdministrator` — they passed before this change and pass after, so they don't protect against a regression. The test needs to be "a *restricted* admin can save their profile".

I wrote one locally and confirmed it fails on `master` and passes with this PR. Please add it to the existing spec file, along with the import `import { VendureAdminClient } from '../../utils/vendure-admin-client.js';`:

```ts
    // #5037 — an admin without UpdateAdministrator permission must still be able to
    // edit their own profile (the page uses updateActiveAdministrator, gated by Owner)
    test('should allow an admin without UpdateAdministrator permission to update own profile', async ({
        page,
        browser,
    }) => {
        const suffix = Date.now();
        const emailAddress = `restricted-${suffix}@test.com`;
        const password = 'test-password';

        const client = new VendureAdminClient(page);
        await client.login();
        const { createRole } = await client.gql(
            `mutation ($input: CreateRoleInput!) { createRole(input: $input) { id } }`,
            {
                input: {
                    code: `restricted-${suffix}`,
                    description: 'No admin permissions',
                    permissions: ['ReadCatalog'],
                },
            },
        );
        await client.gql(
            `mutation ($input: CreateAdministratorInput!) { createAdministrator(input: $input) { id } }`,
            {
                input: {
                    firstName: 'Restricted',
                    lastName: 'Admin',
                    emailAddress,
                    password,
                    roleIds: [createRole.id],
                },
            },
        );

        // Fresh context so we browse as the restricted admin, not the superadmin
        const context = await browser.newContext({ storageState: { cookies: [], origins: [] } });
        const restrictedPage = await context.newPage();
        await restrictedPage.goto('/login');
        await restrictedPage.getByPlaceholder('Email').fill(emailAddress);
        await restrictedPage.getByPlaceholder('Password').fill(password);
        await restrictedPage.getByRole('button', { name: 'Sign in' }).click();
        await expect(restrictedPage).not.toHaveURL(/\/login/, { timeout: 15_000 });

        await restrictedPage.goto('/profile');
        const firstNameField = restrictedPage.locator('[data-slot="field"]').filter({
            has: restrictedPage.locator('[data-slot="field-label"]').getByText('First name', { exact: true }),
        });
        await firstNameField.getByRole('textbox').fill('Renamed');
        await restrictedPage.getByRole('button', { name: 'Update' }).click();

        await expect(
            restrictedPage.locator('[data-sonner-toast]').filter({ hasText: 'Successfully updated profile' }),
        ).toBeVisible({ timeout: 10_000 });

        await restrictedPage.reload();
        await expect(firstNameField.getByRole('textbox')).toHaveValue('Renamed');

        await context.close();
    });
```

Run with:

```bash
CI=true VITE_TEST_PORT=5176 bunx playwright test --config e2e/playwright.config.ts e2e/tests/settings/profile.spec.ts --reporter=list
```

Happy to approve once the test is in and the commit trailer is fixed.

biggamesmallworld (approved):

PR #5057, opened 2026-07-29
fix(dashboard): treat readonly custom fields as optional in Zod schema

### LINKED ISSUE #5045: [dashboard] Readonly non-nullable custom field permanently disables the create/update button
**Describe the bug**
When a custom field is configured with both `readonly: true` and `nullable: false`, the Dashboard's generated Zod validation schema treats it as a **required** form value: `applyCustomFieldModifiers()` 

However, readonly custom fields can never receive a value from the form: the server excludes readonly custom fields from the Create...Input / Update...Input GraphQL types (graphql-custom-fields.ts in @vendure/core) and the Dashboard strips them from the mutation payload via` removeReadonlyAndLocalizedCustomFields()` before submitting, and renders the input disabled and empty

The result is an unsatisfiable schema the form is permanently invalid, `!form.formState.isValid` keeps the Create/Update button disabled

**To Reproduce**

1. Add a custom field like:
```js
customFields: {
    Promotion: [
        {
            name: 'type',
            type: 'string',
            nullable: false,
            readonly: true,
            defaultValue: 'manually-created',
            label: [{ languageCode: LanguageCode.en, value: 'Promotion Type' }],
        },
    ],
},
```
2. Try to create a new promotion, fill in all visible required fields
3. The Create button remains disabled and no error is shown anywhere on the page 

**Expected behavior**
A readonly custom field should not block form submission

**Actual behavior**
The generated Zod schema marks the field as required, the form is permanently invalid, and the Create/Update button is disabled with no way to satisfy the validation

**Environment:**

- @vendure/core: 3.6.4
- @vendure/dashboard: 3.6.4

**Additional context**
For context: the angular Admin UI never derived a required-validator from nullable (custom field controls got initial values only, no validators), so the same field config worked there

Ryrahul:
## Breaking changes

No breaking changes. This is a purely additive fix — the only behavioral change is that readonly custom fields with `nullable: false` no longer block form submission. All existing validation behavior for non-readonly fields is unchanged.

## Summary

Fixes #5045

When a custom field is configured with both `readonly: true` and `nullable: false`, the Dashboard's Zod validation schema treats it as **required**. But readonly fields can never receive a value from the form:

- The server excludes them from `Create*Input` / `Update*Input` GraphQL types
- The Dashboard strips them via `removeReadonlyAndLocalizedCustomFields()` before submitting
- The input is rendered disabled

The result is an unsatisfiable schema — the form is permanently invalid and the Create/Update button stays disabled.

**Fix:** A single-condition change in `applyCustomFieldModifiers()` — readonly fields are now always treated as optional regardless of their `nullable` setting.

```diff
- if (customField.nullable !== false) {
+ if (customField.nullable !== false || customField.readonly) {
```

## Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5057"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787948570&installation_model_id=20193&pr_number=5057&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5057&signature=5c0e1971205dee60f87f5655105b4970075f70d55c4fe6c18aa790b212919b6b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.spec.ts
+++ b/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.spec.ts
@@ -41,6 +41,7 @@ const createMockCustomField = (
         datetimeMax?: string;
         list?: boolean;
         nullable?: boolean;
+        readonly?: boolean;
     } = {},
 ) => ({
     name,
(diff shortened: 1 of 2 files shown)


HouseinIsProgramming (changes requested):
Nice one, clean fix, and thanks for adding a test.

One thing before I approve: the test covers the readonly case but not the opposite one. I had a fiddle locally, and if you swap the whole condition for `if (true)` the spec still passes. So nothing's actually guarding the non-nullable behaviour any more.

Worth adding a case where a plain `nullable: false` field (no readonly) still gets rejected.

HouseinIsProgramming (approved):
Guard test added and verified locally. Approving.

PR #5058, opened 2026-07-30
fix(core): Price every ProductVariant in relation arrays, avoid spread RangeError

ryandiginomad:
# Description

This is the first half of the follow-up agreed in [#4986](<https://github.com/vendurehq/vendure/issues/4986>). @michaelbromley wrote there:

> Separate follow-up (not this PR): `isTranslatable()` and the price applicator still sample element `[0]`, so e.g. a null `variants[0].featuredAsset` suppresses translation of the newly-hydrated `variants[1].featuredAsset`.

and @biggamesmallworld added a note on reachability:

> The `undefined`-hole change in this PR widens who can hit it. If element `[0]` of a relation array is a hole or `null`, `getRelationEntityAtPath()` pushes it into the result, and the `entity[0] instanceof ProductVariant` sample then skips pricing (and `isTranslatable()` skips translation) for the **entire** array, not just that element. Same root cause, more reachable now.

**This PR covers the price-applicator half, plus a sibling spread overflow in the same class.** The `isTranslatable()` half needs a change in `translate-entity.ts` to avoid regressing `null` relations into `undefined`, so it goes in companion PR vendurehq/vendure#5059 to keep each review focused.

## Fix 1: `[0]` sampling in the price application (second commit)

Whether an array-valued relation received prices was decided by `entity[0] instanceof ProductVariant`, but `entity.map(...)` then called `applyChannelPriceAndTax()` on **every** element — and `applyChannelPriceAndTax()` dereferences its argument on its first statement (`variant.productVariantPrices`). Since `getRelationEntityAtPath()` deliberately pushes `null`/`undefined` elements into its result, the `[0]` sample was wrong in both directions:

| relation array | master | this PR |
| -- | -- | -- |
| `[null, variant]` | silently prices nothing | prices `variant` |
| `[variant, null]` | `TypeError` out of `hydrate()` | prices `variant` |
| `[variant, variant]` | prices both | prices both (unchanged) |

So `[0]` was not acting as a guard — it was a coin flip on where the hole sits. The variants to price are now selected with a per-element type-guard filter (new `getProductVariantsToPrice()` helper), which fixes both directions at once. Note that widening the gate alone (e.g. `entity.some(e => e instanceof ProductVariant)`) would have made things *worse*: it converts both silent-skip rows above into crashes because the body still maps over every element. I verified that variant against the new tests and it fails both of them.

On reachability, to be precise: every core `@ManyToOne` into `ProductVariant` is non-nullable, so pure core code mostly cannot produce the mixed array; the live path is the `undefined` hole on entities held in plugin/event-handler code (the vendurehq/vendure#4955 lineage), which [#4986](<https://github.com/vendurehq/vendure/issues/4986>)'s hole-preservation made more visible, as noted in the quote above.

## Fix 2: spread `RangeError` in `getRelationEntityAtPath()` (first commit)

`getRelationEntityAtPath()` collects a terminal relation array with `result.push(...target)` — the exact defect you had me fix in `getMissingRelations()` in [#4986](<https://github.com/vendurehq/vendure/issues/4986>) (spreading a very large array into call arguments exceeds V8's stack budget and throws `RangeError: Maximum call stack size exceeded`). **I missed this sibling occurrence in the same class when writing that fix**; found it while scoping this follow-up. Replaced with the same `for...of` loop merged in [#4986](<https://github.com/vendurehq/vendure/issues/4986>). `forEach` would also avoid the `RangeError` but silently skips holes (measured: 999,999 elements collected instead of 1,000,000), which matters precisely because downstream code now handles those holes per element.

The two fixes are in one PR because they are causally linked: fix 2 un-blocks arrays above \~110k elements so they now *reach* the price application instead of aborting `hydrate()` — landing them together means the path fix 2 opens is already safe on arrival.

# Breaking changes

None. Arrays whose element `[0]` was a hole now get their real ProductVariants priced (the documented intent of `applyProductVariantPrices`), and very large terminal relation arrays no longer throw. A variant that master's `[0]` sample skipped and this PR now prices takes exactly the exposure master already takes today on every array whose `[0]` is a variant — `getRequiredProductVariantRelations()` appends `.taxCategory`/`.productVariantPrices` for those paths under the same option flag. Cost-wise, the filter adds one `instanceof` check per element, only under `applyProductVariantPrices: true`, on an array the code already walks to price.

# Testing

**Unit** (`entity-hydrator.service.spec.ts`, 7 new cases, no DB):

* 2 call-site tests drive the real `hydrate()` (stubbed query builder, real `ProductPriceApplicator` with stubbed strategies) and are the regression guards for fix 1 — one per failure direction. RED on master: `[null, variant]` leaves `listPrice` unset; `[variant, null]` rejects with `TypeError: Cannot read properties of null (reading 'productVariantPrices')`.
* 1 pins the fully-populated `[variant, variant]` case; passes before and after.
* 3 pin the semantic edges of the new helper (`[]`, `[null, undefined]`, non-ProductVariant entities). These are spec-locks for the helper, not regression proofs — stated here so the coverage isn't overread.
* 1 guards fix 2: a 1M-element terminal relation array with a deliberate hole. RED on master with `RangeError` at the `push(...target)` line; the hole assertion is what rules out a `forEach`-based fix. Costs \~35 ms / \~58 MB transient on an M-series Mac (measured), cheaper than the neighbouring 200k-distinct-objects fixture.

**Robustness checks on the tests themselves**: I mutation-tested the suite — reverting each fix individually, substituting the `.some()` gate for fix 1 and `forEach` for fix 2 — and confirmed every mutant turns at least one of the new tests red (after asserting the mutant actually applied).

What I ran:

|  | Result |
| -- | -- |
| `entity-hydrator.service.spec.ts` before the fixes | 6 failed / 16 passed |
| `entity-hydrator.service.spec.ts` after | 22 passed |
| Full core unit suite | 75 files, 1144 passed |
| `entity-hydrator.e2e-spec.ts` (rebuilt dist) | 25 passed |

No new e2e: no core relation can produce the mixed array (non-nullable FKs, see reachability note), so an e2e would need a permanent nullable-FK fixture change purely to demonstrate a shape core cannot produce, and would only exercise one failure direction per run depending on row order. The DB-free call-site tests cover both directions deterministically. The pre-existing type error in `order.service.ts:1460` that stops `npm run build` locally is unrelated (file untouched, reproduces on clean master).

### 🤖 AI assistance

# Checklist

📌 Always:

- [X] I have [checked my own PR](<##>)

👍 Most of the time:

---

![View with \[code\]smith](https://camo.githubusercontent.com/6e22944985996f6a1201f1d35788e9dc855f2e6b7b781e7ed1c22d7fb14217e8/68747470733a2f2f70722d636f6d6d656e74732d6173736574732e626c61636b736d6974682e73682f636f6465736d6974682f766965772d776974682d636f6465736d6974682d6461726b2d76322e737667) ![Autofix with \[code\]smith](https://camo.githubusercontent.com/8e7d836069a9e09b18eb0f68d9afc9beb6c2b13211745830b36a5d88a500d653/68747470733a2f2f70722d636f6d6d656e74732d6173736574732e626c61636b736d6974682e73682f636f6465736d6974682f6175746f6669782d776974682d636f6465736d6974682d6461726b2e737667) Need help on this PR? Tag `@codesmith-bot` with what you need. Autofix is disabled.

### DIFF
--- a/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
+++ b/packages/core/src/service/helpers/entity-hydrator/entity-hydrator.service.spec.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from 'vitest';
 
+import { Asset } from '../../../entity/asset/asset.entity';
+import { ProductVariant } from '../../../entity/product-variant/product-variant.entity';
+import { ProductPriceApplicator } from '../product-price-applicator/product-price-applicator';
+
 import { EntityHydrator } from './entity-hydrator.service';
 
 describe('EntityHydrator', () => {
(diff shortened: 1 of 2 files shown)


HouseinIsProgramming (commented):
Solid follow-up to #4986. One ask, one suggestion.

**1. Missing tests.** `getProductVariantsToPrice()` is only tested with arrays. The non-array branch, which handles pricing a single variant (e.g. `hydrate(ctx, orderLine, { relations: ['productVariant'], applyProductVariantPrices: true })`), has no coverage: it could regress and no test would fail. Please add two cases: bare `ProductVariant` returns `[variant]`, `undefined` returns `[]`.

**2. There is a problem that your fix might expose, would be nice if we guard against it.**
The pricing step now runs `applyChannelPriceAndTax()` for all variants **concurrently** via `Promise.all`, without any limit.
Before your fix, it was being stopped by the `RangeError` crash which capped it. Now that the limit is gone, we would need to bound the `Promise.all` somehow.
I see no upside to the concurrency here, correct me if I am wrong.
So consider adding a limit or sequentialising this. `ProductVariantService.assignProductVariantsToChannel` has a similar pattern you can follow.

Will approve once both points above are addressed, thanks for the thorough PR.

ryandiginomad:
@HouseinIsProgramming Both points are in.

1. Tests: af19e29d2 adds the two non-array cases (bare ProductVariant wraps to `[variant]`, `undefined` returns `[]`). I also checked they bite: turning the non-array branch into `return []` fails the new bare-variant case.

2. Sequential pricing: b56c030cb replaces the Promise.all with a per-variant await loop, same shape as `assignProductVariantsToChannel()`. Agreed there is no upside to the concurrency. The Promise.all itself predates this PR, but the RangeError fix is what lets very large arrays reach it now, so bounding it belongs here.

What I ran after the changes: hydrator spec 24/24, full core unit suite 1146, `entity-hydrator.e2e-spec.ts` 25/25 against a rebuilt dist.

Thanks for the quick pickup and for pointing at the existing pattern to follow.

HouseinIsProgramming (approved):
Good fix. The tests fail for the right reasons in both directions, and the newly-priced variants already had their price relations fetched, so there is no new exposure. The isTranslatable half is covered in #5059.

PR #5059, opened 2026-07-30
fix(core): Check every element in isTranslatable, harden translateDeep against holes

ryandiginomad:
# Description

This is the second half of the follow-up agreed in #4986 — the `isTranslatable()` half of what @michaelbromley described there:

> Separate follow-up (not this PR): `isTranslatable()` and the price applicator still sample element `[0]`, so e.g. a null `variants[0].featuredAsset` suppresses translation of the newly-hydrated `variants[1].featuredAsset`.

The price-applicator half is #5058. This one is separate because fixing `isTranslatable()` correctly requires changes in `translate-entity.ts` — shared code with five call sites (`translator.service.ts`, `facet.service.ts` ×2, `facet-value.service.ts`, plus the hydrator) — and that scope deserves its own review rather than riding along with two mechanical single-file fixes.

## Why the gate change alone would be a regression (commit order)

Making `isTranslatable()` element-aware sends relation arrays that contain `null`/`undefined` entries into `translateDeep()` — which is exactly the case it was never asked to handle before. On master, `translateDeep()`/`translateLeaf()` have three defects on such arrays, each reproducible today **without** the hydrator change (unit tests for all three are RED on clean master):

1. `translateLeaf()` maps a relation array straight into `translateEntity()`, which dereferences `.translations` — an `undefined`/`null` element throws a `TypeError` out of `translateDeep()`.
2. The two-segment array branch assigns `object[property]` where `object` is the array element itself — a hole there throws `Cannot set properties of undefined`. Not reachable via `hydrate()` (`getRelationEntityAtPath()` throws first on that shape — verified by driving it), but `facet.service.ts` passes exactly this shape (`[['values', 'facet']]`) with plain TypeORM query results.
3. The same branch writes the result of `translateLeaf()` back unconditionally. `translateLeaf()` returns `undefined` for a `null` relation, so a relation that is **null in the database gets rewritten to `undefined`** — destroying the `undefined` = never-fetched / `null` = fetched-and-null distinction that `getMissingRelations()` relies on since #4986, which makes `hydrate()` lose idempotency and re-query such relations on every call. This one is live on master right now whenever the asset sits at element `[0]`.

So the PR is structured as two commits in dependency order: **commit 1** hardens `translateDeep()`/`translateLeaf()` (fixes the three defects above, independently valuable), **commit 2** makes `isTranslatable()` check every element (`.some()` over loaded `translations`), which is only safe on top of commit 1. If you'd rather land them separately, they split cleanly at the commit boundary.

**Updated after review (third commit):** I originally left the shared `if (object && property)` assignment guard at the end of `translateDeep()`'s loop untouched, reasoning that the non-array branches only ever receive defined values. That holds inside `hydrate()`, whose `isTranslatable()` gate filters null leaves — but not for the other four callers: `translate(collection, ctx, ['parent'])` reaches the single-segment branch with a `null` parent on the root collection and rewrote it to `undefined`. Coderabbit's review flagged this; I confirmed both non-array shapes with failing tests and applied the one-line guard (`value !== undefined`) — the same null-preservation rule the array branch already follows — pinned by two regression tests.

**Scope and perf notes (added per review):**

- The hardening stops at two path segments, which is all `DeepTranslatableRelations` can express: a three-segment path like `lines.productVariant.featuredAsset` is not translated by `translateDeep()`, before or after this PR.
- An incidental fix that fell out of the guard: a three-segment path used to leave a junk own-property on the entity (the stringified path array as key, e.g. `"variants,featuredAsset,x"`), because the fall-through branch assigned `undefined` under that key. Now pinned by a small test.
- Preserving `null` is a perf fix as well as a correctness one. `getMissingRelations()` treats `undefined` as never-fetched, so the old null-to-undefined rewrite made every subsequent `hydrate()` call re-query that relation. On master this fires whenever a relation is null at element `[0]`, i.e. on hot paths like order lines and variant lists.

# Breaking changes

None. Arrays whose element `[0]` was a hole now get their real entities translated (the documented behavior); `null` relations stay `null` instead of flipping to `undefined`; and `translateDeep()` no longer throws on arrays containing holes — every call site keeps its current behavior for fully-populated arrays, pinned by the existing 31 tests in `translate-entity.spec.ts` which all still pass.

# Testing

**Unit — `translate-entity.spec.ts` (6 new cases).** The last two (third commit) pin null-preservation for the single-segment and two-segment non-array paths — RED before the shared-assignment guard. The first four RED on clean master, one per defect plus one per failure direction: `TypeError (reading 'translations')` on an `undefined` hole and on a `null` element; `TypeError (setting 'options')` on a level-0 hole; and `null` featuredAsset rewritten to `undefined`. GREEN with commit 1 alone — they document live bugs independent of the `isTranslatable()` change, which is why they sit in the first commit.

**Unit — `entity-hydrator.service.spec.ts` (4 new cases).** Two are the regression guards for `isTranslatable()` (translatable entity behind a `null` / behind an `undefined` leading element — RED on master, `false` instead of `true`); two pin the `some` semantics (`[]` and no-translatable-element stay `false`) and pass before and after, stated so the coverage isn't overread.

**e2e — `entity-hydrator.e2e-spec.ts` (1 new case), run against a rebuilt dist.** No fixture changes: the full-products CSV gives Laptop four variants whose `featuredAsset` is null; the test gives exactly one variant the fixture asset, hydrates `variants.featuredAsset`, and asserts (a) the asset's `LocaleString` name is translated and (b) the other three variants are still `null`, not `undefined`. RED on master — on the sqlite row order the asset lands at the end of the array, so the `[0]` sample sees `null` and skips translation entirely (`name` stays `undefined`, the exact symptom from the quote above). The other ordering (asset at `[0]`, nulls rewritten to `undefined`) is covered deterministically by the unit tests. The shared-fixture mutation is restored in a `finally` block, the expected counts are derived from the fixture rather than hard-coded, and the expected asset name is the fixture literal, so the test is self-contained regardless of what runs after it.

What I ran:

| | Result |
|---|---|
| `translate-entity.spec.ts` before commit 1 | 4 failed / 31 passed |
| `entity-hydrator.service.spec.ts` before commit 2 | 2 failed / 17 passed |
| non-array null-preservation tests before commit 3's guard | 2 failed / 35 passed |
| All after (incl. the junk-own-property pin added in review) | 57 passed |
| Full core unit suite | 75 files, 1148 passed |
| `entity-hydrator.e2e-spec.ts` on master dist + new test | 1 failed (the new test) / 25 passed |
| `entity-hydrator.e2e-spec.ts` on fixed dist | 26 passed |

The pre-existing type error in `order.service.ts:1460` that stops `npm run build` locally is unrelated (reproduces on clean master); the import-order and require-await eslint warnings in the touched files reproduce identically on the master versions of those files.

### 🤖 AI assistance

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5059"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787968901&installation_model_id=20193&pr_number=5059&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5059&signature=0bb3222119610abca134f56284dc44dda787cb298f4bc1bec90b6e2dd0fbbe5d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/entity-hydrator.e2e-spec.ts
+++ b/packages/core/e2e/entity-hydrator.e2e-spec.ts
@@ -10,6 +10,7 @@ import {
     OrderLine,
     OrderService,
     Product,
+    ProductVariant,
     RequestContext,
     RequestContextService,
     TransactionalConnection,
@@ -522,6 +523,44 @@ describe('Entity hydration', () => {
         const child = entity.childrenPropertyWithAVeryLongNameThatExceedsPostgresLimitsEasilyByItself[0];
         expect(child.image1).toBeDefined();
         expect(child.image2).toBeDefined();
+    });
+
+    // Follow-up to #4986: isTranslatable() decided from element [0] whether an entire relation
+    // array was translatable. With a null featuredAsset at [0], nothing in the array was
+    // translated (Asset.name is a LocaleString, so it stayed undefined); with the asset at [0],
+    // translation ran but every other element's null featuredAsset was rewritten to undefined.
+    // The assertions below fail under either ordering, so the test does not depend on DB row
+    // order or on the id strategy.
+    it('translates a relation reached past a null array element', async () => {
+        const ctx = await server.app.get(RequestContextService).create({ apiType: 'admin' });
+        const connection = server.app.get(TransactionalConnection).rawConnection;
+        const asset = await connection.getRepository(Asset).findOne({ where: {} });
+        const assetName = asset!.translations.find(t => t.languageCode === ctx.languageCode)!.name;
+
+        const laptop = await connection
+            .getRepository(Product)
+            .findOne({ where: { id: 1 }, relations: ['variants'] });
+        // Give exactly one of the four variants a featuredAsset; the other three keep the
+        // featuredAsset that is null in the database.
+        await connection
+            .getRepository(ProductVariant)
+            .update(laptop!.variants[laptop!.variants.length - 1].id, { featuredAsset: asset! });
+
+        const product = await connection
+            .getRepository(Product)
+            .findOne({ where: { id: 1 }, relations: ['variants'] });
+
+        await server.app.get(EntityHydrator).hydrate(ctx, product!, {
+            relations: ['variants.featuredAsset'],
+        });
+
+        const withAsset = product!.variants.filter(v => v.featuredAsset != null);
+        expect(withAsset.length).toBe(1);
+        expect(withAsset[0].featuredAsset.name).toBe(assetName);
+        // The other three were null in the database and must still be null, not undefined:
+        // getMissingRelations() treats undefined as never-fetched, so hydrate() would re-query
+        // them on every subsequent call.
+        expect(product!.variants.filter(v => v.featuredAsset === null).length).toBe(3);
     });
 });
 
--- a/packages/core/e2e/entity-hydrator.e2e-spec.ts
+++ b/packages/core/e2e/entity-hydrator.e2e-spec.ts
@@ -534,33 +534,45 @@ describe('Entity hydration', () => {
     it('translates a relation reached past a null array element', async () => {
         const ctx = await server.app.get(RequestContextService).create({ apiType: 'admin' });
         const connection = server.app.get(TransactionalConnection).rawConnection;
-        const asset = await connection.getRepository(Asset).findOne({ where: {} });
-        const assetName = asset!.translations.find(t => t.languageCode === ctx.languageCode)!.name;
+        // The expected name is the known fixture literal (also asserted near the top of this
+        // file), so the expectation is independent of the data the translation path reads
+        const assetName = 'derick-david-409858-unsplash.jpg';
+        const assets = await connection.getRepository(Asset).find();
+        const asset = assets.find(a => a.translations.some(t => t.name === assetName));
 
+        // Raw-connection queries take the numeric DB id; 'T_1' exists only at the API layer
         const laptop = await connection
             .getRepository(Product)
             .findOne({ where: { id: 1 }, relations: ['variants'] });
-        // Give exactly one of the four variants a featuredAsset; the other three keep the
+        const variantWithAssetId = laptop!.variants[laptop!.variants.length - 1].id;
+        const nullAssetVariantCount = laptop!.variants.length - 1;
+        // Give exactly one of the variants a featuredAsset; the others keep the
         // featuredAsset that is null in the database.
-        await connection
-            .getRepository(ProductVariant)
-            .update(laptop!.variants[laptop!.variants.length - 1].id, { featuredAsset: asset! });
+        await connection.getRepository(ProductVariant).update(variantWithAssetId, { featuredAsset: asset! });
+        try {
+            const product = await connection
+                .getRepository(Product)
+                .findOne({ where: { id: 1 }, relations: ['variants'] });
 
-        const product = await connection
-            .getRepository(Product)
-            .findOne({ where: { id: 1 }, relations: ['variants'] });
+            await server.app.get(EntityHydrator).hydrate(ctx, product!, {
+                relations: ['variants.featuredAsset'],
+            });
 
-        await server.app.get(EntityHydrator).hydrate(ctx, product!, {
-            relations: ['variants.featuredAsset'],
-        });
-
-        const withAsset = product!.variants.filter(v => v.featuredAsset != null);
-        expect(withAsset.length).toBe(1);
-        expect(withAsset[0].featuredAsset.name).toBe(assetName);
-        // The other three were null in the database and must still be null, not undefined:
-        // getMissingRelations() treats undefined as never-fetched, so hydrate() would re-query
-        // them on every subsequent call.
-        expect(product!.variants.filter(v => v.featuredAsset === null).length).toBe(3);
+            const withAsset = product!.variants.filter(v => v.featuredAsset != null);
+            expect(withAsset.length).toBe(1);
+            expect(withAsset[0].featuredAsset.name).toBe(assetName);
+            // The others were null in the database and must still be null, not undefined:
+            // getMissingRelations() treats undefined as never-fetched, so hydrate() would
+            // re-query them on every subsequent call.
+            expect(product!.variants.filter(v => v.featuredAsset === null).length).toBe(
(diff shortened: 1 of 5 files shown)


ryandiginomad:
The coderabbit finding above ("Null-to-undefined rewrite still possible for non-array relation paths") is confirmed and addressed in 0006a5f85.

I verified it rather than taking it as-is: on this branch, both non-array path shapes did still rewrite `null` → `undefined` (reproduced with two failing tests, now in the suite as regression tests), and the shapes are reachable today outside `hydrate()` — e.g. `translate(collection, ctx, ['parent'])` reaches the single-segment branch with a `null` parent on the root collection. My original "deliberately unchanged" note reasoned from `hydrate()`'s `isTranslatable()` gate, which does filter those shapes — but since this PR's premise is that `translateDeep()` has four other call sites without that gate, the guard belongs here.

Applied as the suggested one-line change. `translate-entity.spec.ts` 37/37, full core unit suite 1147 passed, `entity-hydrator.e2e-spec.ts` 26/26 against a rebuilt dist. PR description updated to match.

biggamesmallworld:
**Nice fix — diagnosis and commit ordering are both right.** Verified locally: the new tests are RED against `db8482a7f` and 56/56 GREEN on head. `translateLeaf` returning `nested2` rather than a blanket `null` (so `null` stays `null`, `undefined` stays `undefined`) is the correct call, and `value !== undefined` is provably narrow — `translateLeaf` only returns `undefined` when `object[property]` is falsy, so no truthy relation's assignment changes.

## Must, before merge

- `e2e/entity-hydrator.e2e-spec.ts` — the test sets `featuredAssetId` on a shared Laptop variant and never restores it. "Runs last in the file" is an ordering constraint nothing enforces; the next test appended here inherits a mutated fixture. Wrap in `try/finally` (or use a dedicated product).
- Same test — `expect(...length).toBe(3)` is hard-coded next to a dynamically-picked variant; derive it as `laptop.variants.length - 1`. And `where: { id: 1 }` should be `'T_1'` like the rest of the file. Also prefer the literal `'derick-david-409858-unsplash.jpg'` (already asserted ~line 91) over deriving `assetName` from the same data the production path reads — a self-derived expectation can pass while both sides are wrong.

## Should

- Drop the four `as any` casts in `translate-entity.spec.ts` (167-169, 192-194, 208, 217-219) — `tsc --noEmit` is clean without them, and they disable type checking on the exact API under test.
- The fix stops at two path segments (`translate-entity.ts:211`): a 3-segment path like `lines.productVariant.featuredAsset` still translates nothing. Worth a line in the description so the next person doesn't rediscover it. Incidentally this PR also fixes a junk own-property (`"variants,featuredAsset,x"`) on that path — free, but untested.
- Comments carry PR narrative into the code: `translate-entity.ts:247-251` is a near-copy of `:225-229` ending in "same as in the array branch above"; the `isTranslatable()` doc block ends with a changelog of the old behaviour (already in the commit message); `entity-hydrator.service.spec.ts:264-265` says "these pass before and after the fix" — unidentifiable in six months. Keep the invariants, drop the postmortems.
- Rename the three `does not throw when...` cases — they each assert length, translated value, and hole preservation, not just absence of a throw.

## Performance

- `isTranslatable()` goes O(1) → O(n) per relation, but `.some()` short-circuits on the first translatable element and the body is a `hasOwnProperty` call. Worst case is an all-null relation array, which is a full scan of in-memory objects sitting next to a DB round-trip. Non-issue.
- The null-preservation guard is a net perf **win**, worth stating in the description: `getMissingRelations()` treats `undefined` as never-fetched (`entity-hydrator.service.ts:226`) and `null` as fetched-and-null (`:231`). Rewriting `null` → `undefined` made every subsequent `hydrate()` re-issue the query for that relation. On master this fires whenever the relation is null at element `[0]` — so this fixes a repeated query per hydrate call on hot paths (order lines, variant lists), not just a correctness edge case.
- Widening the gate does mean more arrays now reach `translateDeep()`, including 3-segment paths that no-op. Bounded by relation-array length, no extra queries.

ryandiginomad:
@biggamesmallworld All of it is in, across cbad09534 (e2e) and e72652b61 (tests and comments).

Musts: the fixture mutation is now restored in a `finally`, and both expectations are derived rather than hard-coded — the null-variant count from `laptop.variants.length - 1`, the asset name from the fixture literal you pointed at. One correction on the `'T_1'` detail: that test queries through `rawConnection`, where ids are the numeric DB values. `'T_1'` only exists at the API layer (`TestingEntityIdStrategy` encodes on the way out), so a raw `where: { id: 'T_1' }` would match nothing. I kept the numeric id and added a comment at the site saying why.

Shoulds: the four casts on the `translateDeep()` paths argument are gone (nothing new from `tsc --noEmit` at those call sites; the two `localeName` errors it does report reproduce on master). The three tests are renamed to what they actually assert, the narrative comments are cut down to the invariants they protect, and the junk own-property you spotted is now pinned by a small test instead of staying incidental. The 3-segment limitation and your perf framing of the null-preservation guard are both in the description now.

After the changes: `translate-entity.spec.ts` 38, `entity-hydrator.service.spec.ts` 19, full core suite 1148, `entity-hydrator.e2e-spec.ts` 26/26 against a rebuilt dist.

Thanks for verifying the RED/GREEN yourself and for the perf analysis — the description quotes its substance almost directly.

HouseinIsProgramming (approved):
Good fix, and the commit ordering makes the dependency clear. Checked all five translateDeep call sites, no behaviour change beyond the declared ones.

PR #5065, opened 2026-07-30
fix(core): Propagate field descriptions to generated Filter/Sort parameters

### LINKED ISSUE #5032: Propagate GraphQL field descriptions to auto-generated FilterParameter / SortParameter
When a custom GraphQL object type defines field descriptions (via SDL """...""" or equivalent), Vendure’s generateListOptions() builds {Type}FilterParameter and {Type}SortParameter without copying those descriptions.

As a result, GraphiQL / GraphQL introspection shows helpful docs on the output type (e.g. QuantsStockBasic.tsCode), but the matching filter/sort input fields (e.g. QuantsStockBasicFilterParameter.tsCode) have empty descriptions, even though they represent the same business fields.

This is less a runtime bug and more a DX / schema documentation gap in list-options generation.

To Reproduce

Extend Shop (or Admin) API with a custom type that has field descriptions, e.g.:

```typescript

type ProductNote implements Node {
  id: ID!
  """SKU or product code used for lookup"""
  code: String!
  """Human-readable note body"""
  body: String
}

type ProductNoteList implements PaginatedList {
  items: [ProductNote!]!
  totalItems: Int!
}

input ProductNoteListOptions

extend type Query {
  productNotes(options: ProductNoteListOptions): ProductNoteList!
}
```

2. Start Vendure so generateListOptions creates ProductNoteFilterParameter / ProductNoteSortParameter.
3. Open GraphiQL (or run introspection) and inspect ProductNote.code → description is present.
4. Inspect ProductNoteFilterParameter.code (and ProductNoteSortParameter.code) → description is missing.

**Expected behavior**

Auto-generated list filter/sort input fields should inherit the description from the corresponding field on the source object type (when present), so API docs, GraphiQL, MCP/OpenAPI generators, and codegen all stay consistent.

**Actual behavior**

createFilterParameter / createSortParameter in @vendure/core only set type on each generated input field and omit description, so introspection returns null for those fields even when the source object field has a description.

Relevant code (v3.7.1): dist/api/config/generate-list-options.js — createSortParameter / createFilterParameter field config currently looks like:

```ts

const fieldConfig = {
  type: SortOrder, // or filterType
  // description is not set
};

```
Screenshots/Videos

GraphiQL Docs for *FilterParameter shows field names + operator types only; no description text under fields that are documented on the parent object type.

Error logs

N/A — schema builds successfully; only documentation metadata is missing.

Environment (please complete the following information):

@vendure/core version: 3.7.1
Nodejs version: v24.14.1
Database (mysql/postgres etc): MySQL (not relevant)
Operating System (Windows/macOS/Linux): macOS
Browser (if applicable): GraphiQL in browser
Package manager (npm/yarn/pnpm): Yarn 4.17.1
Configuration

Not configuration-specific. Any plugin/schema that:

Defines an object type with field descriptions, and
Declares input XxxListOptions for Vendure to expand
will reproduce this.

Minimal reproduction

See “To Reproduce” SDL above. After boot, introspection:

```ts
{
  __type(name: "ProductNote") {
    fields { name description }
  }
  __type(name: "ProductNoteFilterParameter") {
    inputFields { name description }
  }
}
```

Expect: ProductNote.code.description is non-null; ProductNoteFilterParameter.code.description is null today.

**Workaround**

Local patch of @vendure/core generate-list-options to pass through descriptions:

```ts
 const fieldConfig = {
   type: SortOrder, // or filterType
+  description: field.description ?? undefined,
 };

```
Applied in both createSortParameter and createFilterParameter.

Additional context

Consistent / always reproducible on current Vendure list-options generation.
Especially painful for data APIs where filter inputs are the main surface developers explore in GraphiQL.
ListOptions wrapper fields (skip / take / filter / sort / filterOperator) already have descriptions; only per-entity filter/sort fields lose the source field docs.
Happy to open a PR with the two-line change if this is accepted as intended behavior.
Proposed change

In createSortParameter and createFilterParameter, when building each fieldConfig, set:

```ts
description: field.description ?? undefined
```
so generated *FilterParameter / *SortParameter fields mirror the documented meaning of the entity fields.

grolmus:
## Summary

Auto-generated `{Type}FilterParameter` and `{Type}SortParameter` input types now inherit the field descriptions from their source object type, so introspection / GraphiQL / codegen stay consistent with the documented object fields.

## Root cause

`generateListOptions()` builds each generated filter/sort input field from only the source field's `type`, omitting its `description`. The `ListOptions` wrapper fields (`skip`/`take`/`sort`/`filter`) carry descriptions, but the per-entity filter/sort fields did not — so a documented field (e.g. `Product.name`) produced `ProductFilterParameter.name` / `ProductSortParameter.name` with a `null` description.

## Change

In `createSortParameter` and `createFilterParameter` (`packages/core/src/api/config/generate-list-options.ts`), copy `field.description` onto each generated `fieldConfig`. Two lines, mirroring the source field's documentation. The synthetic `_and` / `_or` filter fields are untouched and remain undescribed.

## Test plan

**Automated** — added a regression test in `generate-list-options.spec.ts` asserting that a source type with `"""..."""` field descriptions produces matching descriptions on both the generated `*SortParameter` and `*FilterParameter` fields, and that the synthetic `_and` field stays undescribed. Fails without the fix (descriptions come back `undefined`).

```
✓ src/api/config/generate-list-options.spec.ts (8 tests)
✓ src/api/config/graphql-custom-fields.spec.ts (15 tests)
✓ src/service/helpers/list-query-builder/parse-sort-params.spec.ts (7 tests)
✓ src/service/helpers/list-query-builder/parse-filter-params.spec.ts (29 tests)
```

**Manual** — no runtime behaviour change beyond schema documentation metadata; the generated SDL now shows the source field docs on filter/sort inputs. Verified no existing schema/introspection snapshot depended on the previously-empty descriptions.

Fixes vendurehq/vendure#5032

---

![View with \[code\]smith](https://camo.githubusercontent.com/6e22944985996f6a1201f1d35788e9dc855f2e6b7b781e7ed1c22d7fb14217e8/68747470733a2f2f70722d636f6d6d656e74732d6173736574732e626c61636b736d6974682e73682f636f6465736d6974682f766965772d776974682d636f6465736d6974682d6461726b2d76322e737667) ![Autofix with \[code\]smith](https://camo.githubusercontent.com/8e7d836069a9e09b18eb0f68d9afc9beb6c2b13211745830b36a5d88a500d653/68747470733a2f2f70722d636f6d6d656e74732d6173736574732e626c61636b736d6974682e73682f636f6465736d6974682f6175746f6669782d776974682d636f6465736d6974682d6461726b2e737667) Need help on this PR? Tag `@codesmith-bot` with what you need. Autofix is disabled.


HouseinIsProgramming (changes requested):
Thanks, the fix is correct, including for fields merged from a pre-existing `*FilterParameter`/`*SortParameter` input.

One gap: the new test only covers fields from the source object type. The same code also runs for fields pulled in from a pre-declared filter/sort input (`generate-list-options.ts` L165-167), and that path has no coverage. That includes the name-collision case, where the existing input's field silently wins over the source type's, description included.

Could you add a test for that? One assertion that descriptions on fields from a pre-declared `PersonFilterParameter`/`PersonSortParameter` survive, and one for the collision precedence.

Will approve once that's in.

packages/core/src/api/config/generate-list-options.spec.ts
```diff
@@ -350,4 +350,32 @@ describe('generateListOptions()', () => {
...
+        const sortParameter = result.getType('PersonSortParameter') as any;
+        expect(sortParameter.getFields().name.description).toBe("The person's full name");
+        expect(sortParameter.getFields().age.description).toBe('Age in years');
+
+        const filterParameter = result.getType('PersonFilterParameter') as any;
+        expect(filterParameter.getFields().name.description).toBe("The person's full name");
+        expect(filterParameter.getFields().age.description).toBe('Age in years');
+        // synthetic fields without a source description remain undescribed
+        expect(filterParameter.getFields()._and.description).toBeUndefined();
+    });
```
biggamesmallworld, line 394: The production change is correct — I reverted both `description:` lines locally and confirmed this test genuinely fails without them. Two things on the test itself.

**Use `printType` against exact SDL, like every other test in this file.** All 350 lines above assert via `printType(...)` + `removeLeadingWhitespace(...)`; this one reaches into `getFields().name.description` behind `as any`. `printType` renders the descriptions as docstrings, so a single comparison per generated type asserts both fields, both types, *and* that `_and`/`_or` carry no docstring — in the form the rest of the suite uses, with no `as any`.

It also drops `expect(..._and.description).toBeUndefined()`, which pins an implementation detail: graphql-js or `stitchSchemas` normalising an absent description to `null` would fail that assertion for reasons unrelated to the behaviour. With `printType` the question doesn't arise — either a docstring prints or it doesn't. The `// synthetic fields ... remain undescribed` comment goes away with it.

**Cover the pre-existing-input merge path.** `createSortParameter` / `createFilterParameter` fold in fields from a hand-written `{Type}SortParameter` / `{Type}FilterParameter` when one exists (`generate-list-options.ts:126-128`, `:167-169`), and Vendure ships several — `OrderFilterParameter`, `OrderSortParameter`, `ProductFilterParameter`. Those fields flow through the same new `field.description` line, so documented hand-written SDL fields start emitting descriptions too. Currently untested. Second test below covers it (note the merged fields print *before* the source-type fields — that's the existing merge order, asserted as-is).

Verified locally:
- GREEN with the fix: `Tests 9 passed (9)`
- RED without it (both `description:` lines deleted from `generate-list-options.ts`): `Tests 2 failed | 7 passed` — both new tests fail

```suggestion
        const result = generateListOptions(buildSchema(input));

        expect(printType(result.getType('PersonSortParameter')!)).toBe(
            removeLeadingWhitespace(`
                   input PersonSortParameter {
                     """The person's full name"""
                     name: SortOrder

                     """Age in years"""
                     age: SortOrder
                   }`),
        );

        expect(printType(result.getType('PersonFilterParameter')!)).toBe(
            removeLeadingWhitespace(`
                   input PersonFilterParameter {
                     """The person's full name"""
                     name: StringOperators

                     """Age in years"""
                     age: NumberOperators
                     _and: [PersonFilterParameter!]
                     _or: [PersonFilterParameter!]
                   }`),
        );
    });

    it('propagates descriptions from a pre-existing sort and filter parameter', () => {
        const input = `
               ${COMMON_TYPES}
               type Query {
                   people: PersonList
               }

               type Person {
                   name: String!
               }

               input PersonSortParameter {
                   """Sort by relevance score"""
                   score: SortOrder
               }

               input PersonFilterParameter {
                   """Filter by nickname"""
                   nickname: StringOperators
               }
           `;

        const result = generateListOptions(buildSchema(input));

        expect(printType(result.getType('PersonSortParameter')!)).toBe(
            removeLeadingWhitespace(`
                   input PersonSortParameter {
                     """Sort by relevance score"""
                     score: SortOrder
                     name: SortOrder
                   }`),
        );

        expect(printType(result.getType('PersonFilterParameter')!)).toBe(
            removeLeadingWhitespace(`
                   input PersonFilterParameter {
                     """Filter by nickname"""
                     nickname: StringOperators
                     name: StringOperators
                     _and: [PersonFilterParameter!]
                     _or: [PersonFilterParameter!]
                   }`),
        );
    });
```

biggamesmallworld:
Agree with @HouseinIsProgramming , can approve after the tests are updated

grolmus:
@HouseinIsProgramming added the two tests you asked for in 587dff1d6: one asserting that descriptions on fields merged from a pre-declared `PersonFilterParameter`/`PersonSortParameter` survive, and one for the name-collision precedence (the pre-declared input's field wins over the source type's, description included) — for both the filter and sort paths. Whole spec green (10 tests). Re-requested your review.

biggamesmallworld (commented):
The new pre-declared-input tests close the gap I raised, thanks.

Verified locally: green at 587dff1d6 (10 passed), red with both `description:` lines deleted (3 failed). The collision test failing too is correct, since the reduce rebuilds each field config from scratch, so merged fields lose their description as well.

Four things before merge:

**1. Regenerate codegen and docs.** This changes the emitted SDL for every paginated entity, and the repo commits that output. `order.type.graphql:17` documents `code`, but `generated-types.ts:4310` and `input-types.mdx:2660` show no description on `OrderFilterParameter.code`. Comments-only churn, nothing breaks, but the committed artifacts should match. CI runs `bun run codegen` without a diff check, so green doesn't tell us they're current. Please run `bun run codegen` and `bun run docs:generate-graphql-docs` and commit the result.

**2. Drop `?? undefined`** at `generate-list-options.ts:149` and `:208`. `description` is typed `Maybe<string>`, so plain `field.description` is identical. The coalesce only exists to make `.toBeUndefined()` pass, which is the test shaping the implementation.

**3. `spec.ts:412-413`** — the two `toBeDefined()` assertions pass with or without the fix, and field existence is already covered above. Assert type and description, or drop them.

**4. Still asking for `printType`.** 11 assertions in the file use it; `as any` appears only in the three new tests. It asserts what the issue is actually about, that descriptions render as `"""..."""`, catches ordering and type changes in the same comparison, and removes the casts.

Separately, and not for this PR: `createSortParameter` keeps only `type` and `description`, so a pre-declared `*SortParameter` field with `@deprecated` loses its `deprecationReason`. Pre-existing, same class of bug, worth its own issue.

grolmus:
All four addressed:

1. **Codegen + docs regenerated.** Ran `bun run codegen` and `bun run docs:generate-graphql-docs`; the committed generated types, `schema-*.json` and the GraphQL API docs now carry the propagated descriptions (e.g. `OrderFilterParameter.code`). The raw codegen output needed a `prettier --write` pass to match the committed formatting — without it `generated-shop-types.ts` showed ~5k lines of pure whitespace churn — so the diff is now just the description additions.
2. **Dropped `?? undefined`** at both call sites; plain `field.description` it is.
3. Gone — the `toBeDefined()` pair went away with the printType rewrite (field existence is asserted by the printed SDL).
4. **Switched all three new tests to `printType`.** Both the source-type and the pre-declared-input cases assert exact SDL now, and the collision test asserts the pre-declared field's docstring wins — no more `as any`, and the `_and.description` assertion is dropped as you suggested.

Verified: green (10 passed), and red (3 failed) with the two `description:` lines removed.

Good catch on `createSortParameter` dropping `deprecationReason` — agreed that's a separate pre-existing issue; happy to open one for it.

HouseinIsProgramming (approved):
Round 2 looks good. The merge-path tests are non-vacuous (they fail without the source change), and the printType assertion is a nice tightening. Approving.

PR #5066, opened 2026-07-30
fix(dashboard): Forward transformQueryKey and view options from ListPage

### LINKED ISSUE #5026: ListPage does not forward transformQueryKey to PaginatedListDataTable, so transformVariables-injected state serves a stale cache
### Describe the bug

`PaginatedListDataTable` accepts `transformQueryKey` and uses it to build the react-query key:

```ts
// packages/dashboard/src/lib/components/shared/paginated-list-data-table.tsx (3.7.1, ~line 465)
const queryKey = transformQueryKey ? transformQueryKey(defaultQueryKey) : defaultQueryKey;
```

`ListPage` wraps that component, and its `commonTableProps` object forwards 22 props — but not `transformQueryKey`:

```ts
// packages/dashboard/src/lib/framework/page/list-page.tsx (3.7.1, ~lines 564-603)
const commonTableProps = {
    listQuery,
    deleteMutation,
    transformVariables,       // <- forwarded
    customizeColumns,
    ...
    transformData,
    registerRefresher,
};                            // <- transformQueryKey is not in the object
```

`ListPage` does forward `transformVariables`. The two belong together: the default query key contains `page`, `itemsPerPage`, `sorting`, the column filters and the debounced search term, and nothing else. Any state a page injects through `transformVariables` — a language selector, a status filter in the action bar, a parent entity id — is therefore invisible to react-query.

The result is a list that fetches once and then serves that first response forever, for every value of the injected state. There is no error and no empty table: the rows simply never change, which makes it easy to ship and hard to notice in review.

`disableViewOptions` and `includeSelectionColumn` are missing from `ListPage` in the same way, though those fail visibly rather than silently.

### To Reproduce

1. Build a `ListPage` with a toolbar `<Select>` held in component state.
2. Inject it into the query with `transformVariables={vars => ({ ...vars, someFilter: value })}`.
3. Change the Select.

Expected: the list refetches with the new variable. Actual: no network request; the previously cached page is re-rendered.

### Expected behavior

`ListPage` forwards `transformQueryKey` (and ideally `disableViewOptions` / `includeSelectionColumn`) to `PaginatedListDataTable`, so a page that customises variables can keep the query key in sync.

### Environment

- `@vendure/dashboard` 3.7.1
- Verified by reading the installed package source; the prop appears in `paginated-list-data-table.tsx` only.

### Additional context

The workaround is to drop `ListPage` and render `PaginatedListDataTable` directly, which means taking ownership of `page` / `itemsPerPage` / `sorting` state and the URL wiring `ListPage` otherwise provides. Happy to open a PR adding the prop to `ListPageProps` and to `commonTableProps` if that is the shape you would want.

grolmus:
## Summary

`ListPage` now forwards `transformQueryKey`, `disableViewOptions` and `includeSelectionColumn` to the underlying `PaginatedListDataTable`, so pages built on `ListPage` can actually use them.

## Root cause

`ListPage` builds a `commonTableProps` object and spreads it into `PaginatedListDataTable`, forwarding most props — but three real `PaginatedListDataTable` props were missing from that object: `transformQueryKey`, `disableViewOptions` and `includeSelectionColumn`.

The harmful one is `transformQueryKey`. The default react-query key only reflects `page`, `itemsPerPage`, `sorting`, the column filters and the debounced search term. Any state a page injects through `transformVariables` (a language selector, an action-bar status filter, a parent entity id) is invisible to the cache key. Without a way to also transform the key, the list fetches once and then serves that first response for every value of the injected state — no error, no empty table, the rows simply never change. `disableViewOptions` / `includeSelectionColumn` were missing the same way but fail visibly.

## Change

Add the three props in `list-page.tsx` in the three places a forwarded prop must appear: the `ListPageProps` interface (with JSDoc), the component destructuring, and the `commonTableProps` object literal. Types mirror `PaginatedListDataTable` exactly. No behaviour change unless a prop is set.

## Test plan

**Automated** — `list-page.spec.tsx` (new) mocks `PaginatedListDataTable` and asserts `ListPage` forwards the three props (identity for `transformQueryKey`, non-default `false` for `includeSelectionColumn`), plus a negative case asserting they stay `undefined` when not provided. This is a RED→GREEN guard for the exact omission: without the fix, `transformQueryKey` arrives `undefined`.

```
✓ src/lib/framework/page/list-page.spec.tsx (2 tests)
Full dashboard unit suite: 479 passed | 4 skipped
check-types: clean
```

The downstream refetch behaviour lives in `PaginatedListDataTable` + react-query (unchanged here), which already consumes `transformQueryKey` to build the query key.

**Manual** — a `ListPage` using `transformVariables` + `transformQueryKey` to inject toolbar state now refetches when that state changes instead of serving the stale first page.

Fixes #5026

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5066"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787998716&installation_model_id=20193&pr_number=5066&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5066&signature=c89f99bea29b7b6ee93d372effead45b20fe4c595ee9797d4cb1f99376bce384"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
+++ b/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
@@ -46,12 +46,22 @@ describe('ListPage prop forwarding', () => {
                 ...baseProps,
                 transformQueryKey,
                 disableViewOptions: true,
-                includeSelectionColumn: true,
+                // false is the non-default value, so this asserts the value is really forwarded
+                // rather than coinciding with PaginatedListDataTable's own default of true.
+                includeSelectionColumn: false,
             }),
         );
 
         expect(captured.props?.transformQueryKey).toBe(transformQueryKey);
         expect(captured.props?.disableViewOptions).toBe(true);
-        expect(captured.props?.includeSelectionColumn).toBe(true);
+        expect(captured.props?.includeSelectionColumn).toBe(false);
+    });
+
+    it('does not inject the forwarded props when they are not provided', () => {
+        renderToStaticMarkup(React.createElement(ListPage as any, { ...baseProps }));
+
+        expect(captured.props?.transformQueryKey).toBeUndefined();
+        expect(captured.props?.disableViewOptions).toBeUndefined();
+        expect(captured.props?.includeSelectionColumn).toBeUndefined();
     });
 });
--- a/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
+++ b/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { ListPage } from './list-page.js';
 
 // Capture the props ListPage forwards to PaginatedListDataTable.
 const captured: { props?: Record<string, any> } = {};
@@ -28,40 +29,34 @@ vi.mock('@tanstack/react-router', () => ({
     useNavigate: () => vi.fn(),
 }));
 
-// eslint-disable-next-line import/first
-import { ListPage } from './list-page.js';
-
 describe('ListPage prop forwarding', () => {
     const baseProps = {
-        route: { useSearch: () => ({}), fullPath: '/' },
+        route: { useSearch: () => ({}), fullPath: '/' } as any,
         title: 'Test',
         listQuery: {} as any,
     };
+
+    beforeEach(() => {
+        captured.props = undefined;
+    });
 
     it('forwards transformQueryKey, disableViewOptions and includeSelectionColumn to PaginatedListDataTable', () => {
         const transformQueryKey = (queryKey: any[]) => [...queryKey, 'extra'];
 
         renderToStaticMarkup(
-            React.createElement(ListPage as any, {
-                ...baseProps,
-                transformQueryKey,
-                disableViewOptions: true,
+            <ListPage
+                {...baseProps}
+                transformQueryKey={transformQueryKey}
+                disableViewOptions={true}
                 // false is the non-default value, so this asserts the value is really forwarded
                 // rather than coinciding with PaginatedListDataTable's own default of true.
-                includeSelectionColumn: false,
-            }),
+                includeSelectionColumn={false}
+            />,
         );
 
+        expect(captured.props).toBeDefined();
         expect(captured.props?.transformQueryKey).toBe(transformQueryKey);
         expect(captured.props?.disableViewOptions).toBe(true);
         expect(captured.props?.includeSelectionColumn).toBe(false);
-    });
-
-    it('does not inject the forwarded props when they are not provided', () => {
-        renderToStaticMarkup(React.createElement(ListPage as any, { ...baseProps }));
-
-        expect(captured.props?.transformQueryKey).toBeUndefined();
-        expect(captured.props?.disableViewOptions).toBeUndefined();
-        expect(captured.props?.includeSelectionColumn).toBeUndefined();
     });
 });
--- a/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
+++ b/packages/dashboard/src/lib/framework/page/list-page.spec.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { describe, expect, it, vi } from 'vitest';
+
+// Capture the props ListPage forwards to PaginatedListDataTable.
+const captured: { props?: Record<string, any> } = {};
+
+vi.mock('@/vdb/components/shared/paginated-list-data-table.js', () => ({
+    PaginatedListDataTable: (props: Record<string, any>) => {
+        captured.props = props;
+        return null;
+    },
+}));
+
+vi.mock('../layout-engine/page-layout.js', () => ({
+    Page: ({ children }: any) => <>{children}</>,
+    PageTitle: ({ children }: any) => <>{children}</>,
+    PageActionBar: ({ children }: any) => <>{children}</>,
+    PageLayout: ({ children }: any) => <>{children}</>,
+    FullWidthPageBlock: ({ children }: any) => <>{children}</>,
+}));
+
+vi.mock('@/vdb/hooks/use-user-settings.js', () => ({
+    useUserSettings: () => ({ setTableSettings: vi.fn(), settings: {} }),
+}));
+
+vi.mock('@tanstack/react-router', () => ({
+    useNavigate: () => vi.fn(),
+}));
+
+// eslint-disable-next-line import/first
+import { ListPage } from './list-page.js';
+
+describe('ListPage prop forwarding', () => {
+    const baseProps = {
+        route: { useSearch: () => ({}), fullPath: '/' },
+        title: 'Test',
+        listQuery: {} as any,
+    };
+
+    it('forwards transformQueryKey, disableViewOptions and includeSelectionColumn to PaginatedListDataTable', () => {
+        const transformQueryKey = (queryKey: any[]) => [...queryKey, 'extra'];
+
+        renderToStaticMarkup(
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/lib/framework/page/list-page.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    });
+
+    it('does not inject the forwarded props when they are not provided', () => {
+        renderToStaticMarkup(React.createElement(ListPage as any, { ...baseProps }));
+
+        expect(captured.props?.transformQueryKey).toBeUndefined();
+        expect(captured.props?.disableViewOptions).toBeUndefined();
+        expect(captured.props?.includeSelectionColumn).toBeUndefined();
+    });
+});
```
biggamesmallworld: Deletes the no-op second test, drops `as any` so the interface half is type-checked, adds a `beforeEach` reset of the shared `captured` object (matching `page-layout.spec.tsx`), asserts `captured.props` is actually defined, and hoists the import / removes the bogus eslint suppression.

```suggestion
import { renderToStaticMarkup } from 'react-dom/server';
import { beforeEach, describe, expect, it, vi } from 'vitest';

import { ListPage } from './list-page.js';

// Capture the props ListPage forwards to PaginatedListDataTable.
const captured: { props?: Record<string, any> } = {};

vi.mock('@/vdb/components/shared/paginated-list-data-table.js', () => ({
    PaginatedListDataTable: (props: Record<string, any>) => {
        captured.props = props;
        return null;
    },
}));

vi.mock('../layout-engine/page-layout.js', () => ({
    Page: ({ children }: any) => <>{children}</>,
    PageTitle: ({ children }: any) => <>{children}</>,
    PageActionBar: ({ children }: any) => <>{children}</>,
    PageLayout: ({ children }: any) => <>{children}</>,
    FullWidthPageBlock: ({ children }: any) => <>{children}</>,
}));

vi.mock('@/vdb/hooks/use-user-settings.js', () => ({
    useUserSettings: () => ({ setTableSettings: vi.fn(), settings: {} }),
}));

vi.mock('@tanstack/react-router', () => ({
    useNavigate: () => vi.fn(),
}));

describe('ListPage prop forwarding', () => {
    const baseProps = {
        route: { useSearch: () => ({}), fullPath: '/' } as any,
        title: 'Test',
        listQuery: {} as any,
    };

    beforeEach(() => {
        captured.props = undefined;
    });

    it('forwards transformQueryKey, disableViewOptions and includeSelectionColumn to PaginatedListDataTable', () => {
        const transformQueryKey = (queryKey: any[]) => [...queryKey, 'extra'];

        renderToStaticMarkup(
            <ListPage
                {...baseProps}
                transformQueryKey={transformQueryKey}
                disableViewOptions={true}
                // false is the non-default value, so this asserts the value is really forwarded
                // rather than coinciding with PaginatedListDataTable's own default of true.
                includeSelectionColumn={false}
            />,
        );

        expect(captured.props).toBeDefined();
        expect(captured.props?.transformQueryKey).toBe(transformQueryKey);
        expect(captured.props?.disableViewOptions).toBe(true);
        expect(captured.props?.includeSelectionColumn).toBe(false);
    });
});
```

packages/dashboard/src/lib/framework/page/list-page.tsx
```diff
@@ -322,6 +322,27 @@ export interface ListPageProps<
      * that is not often required.
      */
     transformData?: (data: any[]) => any[];
+    /**
+     * @description
+     * Allows the react-query cache key to be transformed. Use this together with
+     * {@link ListPageProps.transformVariables} when a page injects extra state into the query
```
biggamesmallworld: `{@link ListPageProps.transformVariables}` is the only dotted member `@link` in `src/lib` — every other `{@link X}` targets a bare type name, and the docgen has no handling for members, so this likely renders as a dead link.

```suggestion
     * Allows the react-query cache key to be transformed. Use this together with the
     * `transformVariables` prop when a page injects extra state into the query
```

biggamesmallworld (commented):
Fix looks right — I diffed every `PaginatedListDataTableProps` prop against `commonTableProps` and after this change all 27 are accounted for. The `@default true` claim on `includeSelectionColumn` checks out against `use-generated-columns.tsx:82`.

Two issues in the spec, both with committable suggestions below.

**The second test can never fail.** I ran the spec with the fix reverted: `Tests 1 failed | 1 passed` — `does not inject the forwarded props when they are not provided` passes identically before and after. It asserts `undefined === undefined`, and because of the `captured.props?.` optional chaining it would also pass if `ListPage` rendered nothing at all.

**`React.createElement(ListPage as any, ...)` casts away the half of the bug that mattered.** The omission was in three places, one of which was the `ListPageProps` interface. With `as any`, the test goes green even if you add the props to `commonTableProps` and forget the interface — public API still broken, guard still says fine. JSX with the props inline gets `check-types` covering that half.

Also a stray `// eslint-disable-next-line import/first` — that rule isn't configured in `.eslintrc.js` (only `import/order` is), and vitest hoists `vi.mock` so the bottom-of-file import is unnecessary anyway.

Verified the suggested version locally:
- GREEN: `1 passed`
- Runtime RED (drop the three props from `commonTableProps`): `AssertionError: expected undefined to be [Function transformQueryKey]`
- Type RED (drop `transformQueryKey` from `ListPageProps` only): `list-page.spec.tsx(49,17): error TS2322: ... not assignable to ... ListPageProps<...>`

Both halves of the regression now guarded.

---

Non-blocking, for a follow-up: this class of bug exists because every pass-through prop has to be hand-maintained in three places — the `ListPageProps` interface, the destructuring, and the `commonTableProps` literal. #5026 wasn't "someone forgot `transformQueryKey`", it was "the shape guarantees someone forgets something eventually". Deriving the pass-through subset via `Pick<PaginatedListDataTableProps<...>, ...>` and forwarding by rest-spread would make it unrepresentable. Not asking for that in a `master` bug fix, but worth an issue.

A behavioural test (render the real `PaginatedListDataTable` with a stubbed `useQuery`, assert the query key changes when injected state changes) would cover the actual #5026 symptom rather than just the forwarding. Optional — the forwarding guard is the regression that matters.

biggamesmallworld (approved):
Approving on the production change — the prop forwarding is correct and complete (all 27 `PaginatedListDataTableProps` props accounted for in `commonTableProps` after this).

Please commit the spec suggestion above before merging. As it stands the second test passes with the fix reverted, and the `as any` means the `ListPageProps` interface half of the regression is unguarded — so the current spec would not catch a recurrence of half of #5026. The suggested version is verified RED→GREEN on both halves locally.

packages/dashboard/src/lib/framework/page/list-page.spec.tsx
grolmus: Applied in 8d16b2245 — dropped the no-op second test, removed the `as any` so the interface is type-checked, added a `beforeEach` reset of `captured` plus a `toBeDefined` guard, and hoisted the import (removing the eslint suppression; vitest hoists the `vi.mock` calls above it).

packages/dashboard/src/lib/framework/page/list-page.tsx
grolmus: Fixed in c6a84c06c — replaced the dotted-member `{@link ListPageProps.transformVariables}` with a plain `transformVariables` reference.

biggamesmallworld (approved):

PR #5067, opened 2026-07-30
feat(core): Expose previous entity state on CollectionEvent updates

grolmus:
## Summary

Adds a `previousEntity` to `CollectionEvent`, populated for `updated` events, so subscribers can diff a collection against its pre-update state. Builds on and supersedes #4402 by @flobacher (their original commit is preserved).

## Root cause / motivation

`CollectionEvent` only exposed the post-update entity, so a subscriber wanting to react to *what changed* on a collection update had no access to the previous values.

The original PR (#4402) added `previousEntity` but loaded it via a bare `getEntityOrThrow` with no relations, leaving `translations` (and the other relations) empty — so the previous snapshot was not actually usable for diffing. It also introduced a second collection load alongside the existing active-channel guard.

## Change

- `collection-event.ts` — new `previousEntity?: Collection` constructor field, documented with `@since 3.8.0`. The doc notes its relation set is not identical to `entity`, so reliable diffing is limited to `translations`, `featuredAsset`, `assets` and scalar columns.
- `collection.service.ts` — `update()` loads the pre-update collection **once**, with `relations: ['featuredAsset', 'assets', 'channels', 'parent', 'translations']`. This single load both enforces the active-channel guard (unchanged semantics — `findOneInChannel` still filters by channel) and captures the previous snapshot, which is passed to the `updated` `CollectionEvent`. The redundant second load is removed. `productVariants`/`children` are deliberately excluded (potentially huge).
- Mutation-safety: TypeORM has no persistent identity map and `translatableSaver.update` builds a brand-new entity for the save, so `previousEntity` is a detached pre-update snapshot in all update paths (including filter changes).

## Test plan

**Automated** — extends `entity-event-update-state.e2e-spec.ts` (the suite that already covers post-update state for Promotion/ShippingMethod/Administrator/PaymentMethod events) with a `CollectionEvent` case: update a collection's name, assert `event.type === 'updated'`, `event.entity` has the new name, `event.previousEntity` has the **old** name, and that a non-translation relation (`channels`, `parent`) is loaded on the snapshot (guards against silently dropping a relation).

Verified against a freshly-built core:
```
✓ entity-event-update-state.e2e-spec.ts (13 tests)   # incl. the new CollectionEvent case
✓ collection.e2e-spec.ts (78 tests)                  # update() path regression
```
Without the fix, `event.previousEntity` is undefined and the new assertions fail.

Closes #4402

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5067"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788005244&installation_model_id=20193&pr_number=5067&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5067&signature=2ad18335624c0d0ac3b58b42ec58795f5dcd7ceacb2419feb96f791244c30eb3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/entity-event-update-state.e2e-spec.ts
+++ b/packages/core/e2e/entity-event-update-state.e2e-spec.ts
@@ -1,5 +1,6 @@
 import {
     AdministratorEvent,
+    CollectionEvent,
     defaultShippingCalculator,
     defaultShippingEligibilityChecker,
     dummyPaymentHandler,
@@ -22,10 +23,12 @@ import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-conf
 
 import {
     createAdministratorDocument,
+    createCollectionDocument,
     createPromotionDocument,
     createRoleDocument,
     createShippingMethodDocument,
     updateAdministratorDocument,
+    updateCollectionDocument,
     updatePromotionDocument,
     updateShippingMethodDocument,
 } from './graphql/shared-definitions';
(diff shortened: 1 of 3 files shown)


michaelbromley:
Hi Martin,

The PR this comes from is rightly labeled T3: systemic. That means it should not be blindly implemented.

This is a perfect case in fact, because we should not be ad-hoc solving this for a single event. We need to step back and think about the general problem of "how do we convey the specific changes to an entity via the event system?"

grolmus:
@michaelbromley ah u are right. i didnot notice that it is T3

thanks for your check

PR #5068, opened 2026-07-30
fix(core): Scope customFields auto-init to this server's entities

grolmus:
## Summary

Scopes the `config.customFields` auto-initialisation (added in #4965 / OSS-408) to the entities actually registered with this server, instead of every entity present in the process-global TypeORM metadata.

## Root cause

`getEntityNamesWithCustomFields()` read the global `getMetadataArgsStorage()`, which holds every entity imported *anywhere* in the process — including entities not registered with this server (a second server in the same process, or an imported-but-uninstalled plugin). Those produced phantom `config.customFields[EntityName]` keys, which surface as fields on the deprecated `ServerConfig.customFieldConfig` type.

## Change

- `getEntityNamesWithCustomFields(entities)` now takes the server's entity list and only seeds names present in it (the translation-entity exclusion is unchanged; its set stays process-global as it is only used to exclude).
- `runPluginConfigurations()` derives that list with `getAllEntities(config)` (core entities + this config's plugin entities). This is done from `config` rather than `dbConnectionOptions.entities` so **every** caller is covered — not only the `preBootstrapConfig`/`bootstrap()` path, but also the CLI/dashboard schema generators and codegen, which call `runPluginConfigurations` with a config that never set `dbConnectionOptions.entities`. `getEntityNamesWithCustomFields` is internal (not re-exported), so its signature change is safe.

## Test plan

**Automated** — `bootstrap.spec.ts`:
- New: an entity whose `customFields` embedded is present only in the global metadata (not in the server's list) must **not** get a phantom key, while a real entity (Product) still does. Fails without the fix.
- New: a config with no `dbConnectionOptions.entities` (the schema-generator/codegen caller shape) still seeds core entities — locks out the regression of reading the list from `dbConnectionOptions`.

```
✓ src/bootstrap.spec.ts (10 tests)   # RED→GREEN on the two new cases
eslint: 0 errors
```

**Manual** — no runtime behaviour change for a normal single-server bootstrap beyond the removal of phantom `customFieldConfig` fields for unregistered entities.

Relates to #4965

## Behaviour change (CLI / codegen callers)

`runPluginConfigurations` now derives its entity list via `getAllEntities(config)`, which throws `error.entity-name-conflict` when a plugin declares an entity whose name collides with an existing (core or other-plugin) entity. The direct callers — CLI/dashboard schema generators and codegen — previously did not go through this path, so a config with such a name conflict could still generate a schema; it now fails fast on those configs. Only already-ambiguous configs are affected.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5068"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788008513&installation_model_id=20193&pr_number=5068&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5068&signature=d7d258f622cfd682cc6a2382d4cf62375d16c17d0302fa208bf005f88cb498ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/bootstrap.spec.ts
+++ b/packages/core/src/bootstrap.spec.ts
@@ -5,9 +5,9 @@ import { runPluginConfigurations } from './bootstrap';
 import { CustomFieldConfig } from './config/custom-field/custom-field-types';
 import { RuntimeVendureConfig } from './config/vendure-config';
 // Importing the core entities registers their `customFields` embedded columns in the
-// TypeORM metadata; `coreEntitiesMap` also provides the concrete entity list a real server
-// is configured with, which getEntityNamesWithCustomFields() now scopes its seeding to.
-import { coreEntitiesMap } from './entity/entities';
+// TypeORM metadata, and populates `coreEntitiesMap` (which `getAllEntities`, and therefore the
+// auto-init seeding, reads). Imported for its side effect only.
+import './entity/entities';
 import { registerCustomEntityFields } from './entity/register-custom-entity-fields';
 import { VendurePlugin } from './plugin/vendure-plugin';
 
@@ -46,12 +46,7 @@ function makeConfig(partial: {
     plugins?: RuntimeVendureConfig['plugins'];
     customFields?: Record<string, CustomFieldConfig[]>;
 }): RuntimeVendureConfig {
-    return {
-        plugins: [],
-        customFields: {},
-        dbConnectionOptions: { entities: Object.values(coreEntitiesMap) },
-        ...partial,
-    } as unknown as RuntimeVendureConfig;
+    return { plugins: [], customFields: {}, ...partial } as unknown as RuntimeVendureConfig;
 }
 
 describe('runPluginConfigurations()', () => {
@@ -62,6 +57,17 @@ describe('runPluginConfigurations()', () => {
         await runPluginConfigurations(config);
         expect(config.customFields.Product).toEqual([]);
         expect(config.customFields.Customer).toEqual([]);
+    });
+
+    // OSS-653: runPluginConfigurations is also called directly (CLI/dashboard schema generators,
+    // codegen) with a config that never passed through preBootstrapConfig, so `dbConnectionOptions`
+    // carries no `entities`. Seeding must still work there — the entity list is derived from
+    // `config` (core entities + plugin entities) via getAllEntities, not from dbConnectionOptions.
+    it('seeds core entities when the config has no dbConnectionOptions.entities', async () => {
+        const config = makeConfig({});
+        expect((config as any).dbConnectionOptions?.entities).toBeUndefined();
+        await runPluginConfigurations(config);
+        expect(config.customFields.Product).toEqual([]);
     });
 
     // OSS-408: translation entities also declare a `customFields` embedded (for localized
--- a/packages/core/src/bootstrap.spec.ts
+++ b/packages/core/src/bootstrap.spec.ts
@@ -106,6 +106,33 @@ describe('runPluginConfigurations()', () => {
         }
     });
 
+    // OSS-408: the core point of the feature — a plugin's OWN entity that supports custom fields
+    // gets a seeded key too, so the plugin's `configuration` callback can extend it without a
+    // defensive guard. Contrast with the phantom-entity test above: seeding happens here because
+    // the entity is registered with this server (via the plugin's `entities`), not merely present
+    // in the global metadata.
+    it('seeds customFields for a plugin-registered entity', async () => {
+        const storage = getMetadataArgsStorage();
+        class Oss408PluginEntity {}
+        storage.embeddeds.push({
+            target: Oss408PluginEntity,
+            propertyName: 'customFields',
+            prefix: undefined,
+            type: () => Oss408PluginEntity,
+        } as any);
+
+        @VendurePlugin({ entities: [Oss408PluginEntity] })
+        class TestPlugin {}
+
+        try {
+            const config = makeConfig({ plugins: [TestPlugin] });
+            await runPluginConfigurations(config);
+            expect(config.customFields.Oss408PluginEntity).toEqual([]);
+        } finally {
+            storage.embeddeds.pop();
+        }
+    });
+
     it('does not overwrite an existing customFields entry', async () => {
         const existing: CustomFieldConfig[] = [{ name: 'foo', type: 'string' }];
         const config = makeConfig({ customFields: { Product: existing } });
--- a/packages/core/src/bootstrap.spec.ts
+++ b/packages/core/src/bootstrap.spec.ts
@@ -5,9 +5,9 @@ import { runPluginConfigurations } from './bootstrap';
 import { CustomFieldConfig } from './config/custom-field/custom-field-types';
 import { RuntimeVendureConfig } from './config/vendure-config';
 // Importing the core entities registers their `customFields` embedded columns in the
-// TypeORM metadata, which is how getEntityNamesWithCustomFields() detects the entities
-// that support custom fields. Imported for its side effect only.
-import './entity/entities';
+// TypeORM metadata; `coreEntitiesMap` also provides the concrete entity list a real server
+// is configured with, which getEntityNamesWithCustomFields() now scopes its seeding to.
+import { coreEntitiesMap } from './entity/entities';
 import { registerCustomEntityFields } from './entity/register-custom-entity-fields';
 import { VendurePlugin } from './plugin/vendure-plugin';
 
@@ -46,7 +46,12 @@ function makeConfig(partial: {
     plugins?: RuntimeVendureConfig['plugins'];
     customFields?: Record<string, CustomFieldConfig[]>;
 }): RuntimeVendureConfig {
-    return { plugins: [], customFields: {}, ...partial } as unknown as RuntimeVendureConfig;
+    return {
+        plugins: [],
+        customFields: {},
+        dbConnectionOptions: { entities: Object.values(coreEntitiesMap) },
+        ...partial,
+    } as unknown as RuntimeVendureConfig;
 }
 
 describe('runPluginConfigurations()', () => {
@@ -68,6 +73,31 @@ describe('runPluginConfigurations()', () => {
         await runPluginConfigurations(config);
         expect(config.customFields.ProductTranslation).toBeUndefined();
         expect(config.customFields.CollectionTranslation).toBeUndefined();
+    });
+
+    // OSS-653: seeding must be scoped to the entities registered with THIS server, not the global
+    // TypeORM metadata storage. An entity whose `customFields` embedded is present in the process
+    // (e.g. a second test server in the same process, or an imported-but-uninstalled plugin) but
+    // is not in this server's entity list must not produce a phantom `config.customFields` key.
+    it('does not seed customFields for entities not registered with this server', async () => {
+        const storage = getMetadataArgsStorage();
+        class Oss653PhantomEntity {}
+        storage.embeddeds.push({
+            target: Oss653PhantomEntity,
+            propertyName: 'customFields',
+            prefix: undefined,
+            type: () => Oss653PhantomEntity,
+        } as any);
+        try {
+            const config = makeConfig({});
+            await runPluginConfigurations(config);
+            // a real, registered entity is still seeded…
+            expect(config.customFields.Product).toEqual([]);
+            // …but the phantom entity present only in the global metadata is not
+            expect(config.customFields.Oss653PhantomEntity).toBeUndefined();
+        } finally {
+            storage.embeddeds.pop();
+        }
     });
 
     it('does not overwrite an existing customFields entry', async () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/bootstrap.spec.ts
```diff
@@ -59,6 +59,17 @@ describe('runPluginConfigurations()', () => {
...
+    // codegen) with a config that never passed through preBootstrapConfig, so `dbConnectionOptions`
+    // carries no `entities`. Seeding must still work there — the entity list is derived from
+    // `config` (core entities + plugin entities) via getAllEntities, not from dbConnectionOptions.
+    it('seeds core entities when the config has no dbConnectionOptions.entities', async () => {
+        const config = makeConfig({});
+        expect((config as any).dbConnectionOptions?.entities).toBeUndefined();
+        await runPluginConfigurations(config);
+        expect(config.customFields.Product).toEqual([]);
+    });
+
```
biggamesmallworld, line 72: This test can't fail. I reverted the `.filter(name => registeredEntityNames.has(name))` line and it stayed green.

Its config is `makeConfig({})` — byte-identical to the test 10 lines above — and `expect(config.customFields.Product).toEqual([])` is byte-identical too. And line 68 asserts that `makeConfig`, a helper this file defines four lines earlier, doesn't set `dbConnectionOptions` — a test asserting on its own fixture, which tells us nothing about `runPluginConfigurations`.

Suggest deleting it; the case it's aiming at is covered properly in the suggestion below.

```suggestion
```

packages/core/src/bootstrap.spec.ts
```diff
@@ -70,6 +81,31 @@ describe('runPluginConfigurations()', () => {
...
+            const config = makeConfig({});
+            await runPluginConfigurations(config);
+            // a real, registered entity is still seeded…
+            expect(config.customFields.Product).toEqual([]);
+            // …but the phantom entity present only in the global metadata is not
+            expect(config.customFields.Oss653PhantomEntity).toBeUndefined();
+        } finally {
+            storage.embeddeds.pop();
+        }
+    });
```
biggamesmallworld, line 134: The phantom test is a genuine red-green test — good. But nothing here covers the *positive* half of the change: that a **plugin-registered** entity still gets seeded. That's the half plugin authors actually depend on (`config.customFields.MyEntity.push()` without a guard), and it's the half that now routes through `getEntitiesFromPlugins`. If that path regresses, the phantom test stays green and plugin authors get a `TypeError` at bootstrap.

The suggestion below adds it, and folds in the no-`dbConnectionOptions.entities` case the deleted test was reaching for — same config shape, but with an assertion that can fail. Verified red without the fix (blank out the plugin entities and only this new test goes red).

Also extracts the embedded-registration into a helper, since it's now needed twice, and swaps the `pop()` cleanup for `splice(indexOf(...))` — LIFO cleanup silently corrupts global metadata if anything ever interleaves.

```suggestion
    // OSS-653: registers a `customFields` embedded for `target` in the global metadata, as an
    // entity imported into the process would. Returns a cleanup fn.
    function registerCustomFieldsEmbedded(target: new () => any): () => void {
        const storage = getMetadataArgsStorage();
        const embedded = { target, propertyName: 'customFields', prefix: undefined, type: () => target };
        storage.embeddeds.push(embedded as any);
        return () => storage.embeddeds.splice(storage.embeddeds.indexOf(embedded as any), 1);
    }

    // OSS-653: an entity present in the process metadata but not in this server's entity list
    // (a second server in the same process, or an imported-but-uninstalled plugin) must not
    // produce a phantom `config.customFields` key.
    it('does not seed customFields for entities not registered with this server', async () => {
        class Oss653PhantomEntity {}
        const cleanup = registerCustomFieldsEmbedded(Oss653PhantomEntity);
        try {
            const config = makeConfig({});
            await runPluginConfigurations(config);
            expect(config.customFields.Product).toEqual([]);
            expect(config.customFields.Oss653PhantomEntity).toBeUndefined();
        } finally {
            cleanup();
        }
    });

    // OSS-653: the other half — an entity registered by a plugin IS in this server's entity list,
    // so it must still be seeded. Note the config has no `dbConnectionOptions.entities`, the shape
    // the CLI/dashboard schema generators pass; the list comes from `config` via `getAllEntities`.
    it('seeds customFields for a plugin-registered entity', async () => {
        class Oss653PluginEntity {}
        const cleanup = registerCustomFieldsEmbedded(Oss653PluginEntity);

        @VendurePlugin({ entities: [Oss653PluginEntity as any] })
        class TestPlugin {}

        try {
            const config = makeConfig({ plugins: [TestPlugin] });
            await runPluginConfigurations(config);
            expect(config.customFields.Oss653PluginEntity).toEqual([]);
        } finally {
            cleanup();
        }
    });
```

packages/core/src/bootstrap.ts
```diff
@@ -383,8 +383,14 @@ export async function runPluginConfigurations(config: RuntimeVendureConfig): Pro
...
-    // entities nobody extends. See OSS-408.
-    for (const entityName of getEntityNamesWithCustomFields()) {
+    // entities nobody extends. See OSS-408. Seeding is scoped to this server's entities
+    // (core entities plus this config's plugin entities, via `getAllEntities`) rather than the
+    // global TypeORM metadata, to avoid phantom keys from entities imported into the process but
+    // not registered with this server — a second server in the same process, or an
+    // imported-but-uninstalled plugin (OSS-653). Derived from `config` so every caller of
+    // `runPluginConfigurations` is covered, not only the `preBootstrapConfig` path.
+    const entities = getAllEntities(config);
+    for (const entityName of getEntityNamesWithCustomFields(entities)) {
```
biggamesmallworld, line 393: The OSS-653 rationale is written out three times — here, in `getEntityNamesWithCustomFields`, and again above the new spec — plus a fourth in the PR description. Three copies drift; it belongs in the function that implements the rule (see the suggestion on `register-custom-entity-fields.ts`), with a pointer here.

Also, "Derived from `config` so every caller of `runPluginConfigurations` is covered, not only the `preBootstrapConfig` path" justifies a decision, which belongs in the commit message rather than the source. What the next reader needs is the one fact that stops them "simplifying" it back: the other callers never set `dbConnectionOptions.entities`.

```suggestion
    // entities nobody extends. See OSS-408. The entity list is derived from `config` rather than
    // `config.dbConnectionOptions.entities`, since callers other than `preBootstrapConfig`
    // (the CLI/dashboard schema generators) never set the latter.
    for (const entityName of getEntityNamesWithCustomFields(getAllEntities(config))) {
```

packages/core/src/entity/register-custom-entity-fields.ts
```diff
@@ -47,8 +47,16 @@ const MAX_STRING_LENGTH = 65535;
  * translation entities as the target of a `translations` relation, the same signal
  * `registerCustomEntityFields` uses to locate the translation type.
  */
-export function getEntityNamesWithCustomFields(): string[] {
+export function getEntityNamesWithCustomFields(entities: Array<Type<any>>): string[] {
+    // Scope to the entities actually registered with this server. The global metadata storage
+    // holds every entity imported anywhere in the process — including ones not registered here
+    // (a second server in the same process, or an imported-but-uninstalled plugin) — which would
+    // otherwise seed phantom `config.customFields` keys.
+    const registeredEntityNames = new Set(entities.map(entity => entity.name));
```
biggamesmallworld, line 55: The JSDoc still opens "Returns the names of all registered entities…" with no `@param`. "Registered" now carries specific meaning — registered where, by whom — and supplying this argument correctly is the one thing a caller can get wrong, so it should be documented rather than left to the inline comment.

Second point: `new Set(entities.map(entity => entity.name))` matches by **name**, not identity, so the comment overstates what the code does. The headline scenario in this PR — two servers in one process — is precisely where two distinct classes can share a name and this filter can't tell them apart. Identity matching isn't available (embedded targets may be bare strings), so name matching is the right pragmatic choice; it just shouldn't read as stronger than it is.

```suggestion
 *
 * @param entities the entities registered with this server (core entities plus its plugins'
 * entities). Results are scoped to these: the global metadata storage holds every entity
 * imported anywhere in the process — including ones not registered here, e.g. a second server
 * in the same process or an imported-but-uninstalled plugin — which would otherwise produce
 * phantom `config.customFields` keys (OSS-653).
 */
export function getEntityNamesWithCustomFields(entities: Array<Type<any>>): string[] {
    // Matched by name rather than identity, since an embedded target may be a bare string.
    // Two same-named entity classes from different servers are therefore indistinguishable.
    const registeredEntityNames = new Set(entities.map(entity => entity.name));
```

biggamesmallworld (commented):
The fix itself is right, and lands in the correct place: filtering by the server's own entity list is the correct scoping, and deriving that list from `config` via `getAllEntities` rather than `dbConnectionOptions.entities` is the right call — `preBootstrapConfig` overwrites `dbConnectionOptions.entities` with `getAllEntities()` anyway, so the two are equivalent on the bootstrap path, and the config-derived version is the only one that works for the CLI/dashboard schema generators. `getEntityNamesWithCustomFields` isn't re-exported from `src/index.ts` or `src/entity/index.ts`, so the signature change is safe.

Two things on the tests, and some comment trimming — suggestions inline, all verified locally (10/10 green, eslint clean).

One thing without a suggestion: `getAllEntities` isn't pure — it throws `InternalServerError('error.entity-name-conflict')` on a name clash. It now runs twice on the bootstrap path (once in `preBootstrapConfig`, once here), which is harmless, but it also now runs inside `packages/cli/src/commands/schema/generate-schema/generate-schema.ts` and `packages/dashboard/vite/utils/schema-generator.ts`, which never triggered that validation before. Neither has an i18n layer, so a plugin author with a name clash gets the raw message key `error.entity-name-conflict` out of `vendure schema generate`. That's arguably an improvement (fail earlier, fail everywhere) but it's an accidental one — worth either owning in the PR description or dodging with `runPluginConfigurations(config, entities = getAllEntities(config))` so `preBootstrapConfig` passes the list it already has.

HouseinIsProgramming (dismissed):
Fix is correct. Checked the callers: `getAllEntities` returns a fresh array each call, and the seeding now also covers the CLI/codegen paths that never set `dbConnectionOptions.entities`.

Two things:

1. All the seeding tests use core entities. There's no test where a plugin's own entity gets a seeded key, which is the main point of OSS-408. Can you add one?

2. Note in the description: `getAllEntities` throws on entity name conflicts, so CLI/codegen now fail on configs that previously still generated a schema. Fine, but it's new behaviour for those callers.

Follow-up for a separate issue: `getCustomFieldsForAllEntities` in `config.service.ts` overlaps with this seeding and uses a different translation-exclusion heuristic.

Will approve once that's in.

grolmus:
@HouseinIsProgramming addressed both:

1. Added a test (`37cd3d3cd8`) that a **plugin-registered** entity with a `customFields` embedded gets a seeded key — the OSS-408 core case — contrasted with the phantom (not-registered) entity that stays unseeded.
2. Added a **Behaviour change** note to the description covering `getAllEntities` throwing `error.entity-name-conflict` for the CLI/codegen callers.

The overlapping `getCustomFieldsForAllEntities` heuristic you flagged is exactly what #5069 (stacked on this) consolidates. Re-requested your review.

biggamesmallworld (approved):
Approving. Re-verified at 37cd3d3.

The scoping matches what the runtime already does: `ConfigService.getCustomFieldsForAllEntities` filters by `dbConnectionOptions.entities`, and `preBootstrapConfig` sets that to `getAllEntities(userConfig)`. Seeding now uses the same set, so this is an alignment rather than a new rule. Seeding still runs before the plugin `configuration` loop, so the no-guard `config.customFields.X.push()` contract is intact. `getAllEntities` returns a fresh array per call, so the second call mutates nothing shared; the only effect is running the name-conflict check twice, which the description now owns.

Tests: 11/11 green. Mutation-checked both directions locally: dropping the `.filter` fails the phantom test, and swapping `getAllEntities(config)` for `dbConnectionOptions.entities` fails five.

Resolved my four threads. Two were fixed by the new plugin-entity test and the description note. On the other two:

- "This test can't fail" was overstated. `seeds core entities when the config has no dbConnectionOptions.entities` does fail under the mutation it targets. It is redundant rather than dead, since four other tests catch the same mutation. Not worth another round.
- The JSDoc on `getEntityNamesWithCustomFields` still opens "Returns the names of all registered entities" with no `@param`, which is now the one thing a caller can get wrong. Non-blocking nit, happy for it to ride along with #5069.

One thing neither of us flagged: the entity list is snapshotted before the plugin `configuration` loop, so a plugin that injects another plugin from its own `configuration` callback will not have that plugin's entities seeded. Those entities are not registered with TypeORM in that case either, so seeding them was wrong to begin with. Noting it rather than asking for a change.

Only red check is `deploy`, which fails on an empty `VERCEL_TOKEN` for a fork.

PR #5069, opened 2026-07-30
refactor(core): consolidate the translation-entity detection heuristics

grolmus:
## Summary

Core had three independent notions of "is this a translation entity / does it support custom fields", which could classify the same entity differently and drift over time. This collapses them onto one relation-based signal: an entity is a translation entity when something points a `translations` relation at it.

## Root cause

The three definitions:

1. `getEntityNamesWithCustomFields()`: relation-based, via `getRelationTargetName()`.
2. `ConfigService`'s `getCustomFieldsForAllEntities()` getter: a `*Translation` name suffix plus a `languageCode` column.
3. `registerCustomEntityFields()`: an inline `translations` relation lookup.

The getter skipped keys that already existed and the bootstrap-time auto-init pre-empted it, so any entity the two heuristics classified differently changed behaviour depending on which ran first.

## Change

- Extracted the relation-based signal into `getTranslationEntityNames()`, private to `register-custom-entity-fields.ts`, and used by `getEntityNamesWithCustomFields`.
- `getRelationTargetName()` now resolves all three shapes TypeORM allows for a relation target: a constructor closure, a bare string name, and a closure returning a string. The previous code called the target unconditionally, so a bare-string `translations` relation anywhere in the process threw `relation.type is not a function` and aborted bootstrap.
- Shared the `'translations'` literal as `TRANSLATIONS_RELATION_PROPERTY`, used by both the detection helper and the inline lookup in `registerCustomEntityFields`. That lookup stays inline because it needs the relation itself, not a set of names.
- **Deleted the seeding loop in `ConfigService.getCustomFieldsForAllEntities()`** rather than rewiring it, per @gabriellbui's review. `runPluginConfigurations()` already seeds an empty array for every entity that supports custom fields, and it runs inside `preBootstrapConfig()`, which completes before the Nest app and so before any `ConfigService` exists. The CLI and dashboard schema generators call `runPluginConfigurations()` directly and never build a `ConfigService`. Every key the getter would have added is therefore already present. The getter now returns `activeConfig.customFields` unchanged.

Deleting the loop removes the second seeding site, and with it the need to keep two entity-list sources (`dbConnectionOptions.entities` and `getAllEntities(config)`) in step. One definition of "supports custom fields" and one place that acts on it.

The one case where the deleted loop still did something: a plugin `configuration` callback that replaces `config.customFields` with a new object without spreading the old one. Such a callback already discards every user-configured custom field, so it is broken either way.

## Behaviour changes

The removed name-plus-column heuristic and the relation-based one disagree on two edge cases, in both directions:

1. **Now excluded:** a `translations` relation target whose name does not end in `Translation`. The getter would previously have seeded it, risking the duplicate `*TranslationInput.customFields` schema collision.
2. **Now seeded:** an entity whose name ends in `Translation` and which has a `languageCode` column, but which nothing points a `translations` relation at. The old heuristic excluded it; it is now treated as a normal custom-field-bearing entity.

Behaviour is unchanged for every core and standard-plugin entity. An entity could only diverge by wiring its base-to-translation link under a property name other than `translations`, which breaks Vendure's `Translatable` contract everywhere else. Both directions are covered by tests, and the translatable guide documents the detection rule for plugin authors, since the `refactor(core):` title keeps it out of the changelog.

## Known remaining duplication

`registerTranslationEntityUniqueConstraints` in `packages/core/src/entity/register-translation-unique-constraints.ts` landed on `minor` after this PR was opened. It detects translation entities by a `languageCode` column plus a `base` relation, so core still has two definitions that do not agree on the same set of entities. Deliberately left out of this PR: it widens the scope and the two checks run at different points in bootstrap. Tracked as a follow-up in #5327.

## Test plan

**Automated.** All of it in `bootstrap.spec.ts`, which now covers both directions of the detection rule:

- a lookalike translation entity, name ending in `Translation` plus a `languageCode` column but no incoming `translations` relation, is seeded
- real translation entities (`ProductTranslation`, `CollectionTranslation`) are not
- all three relation-target shapes (constructor closure, bare string, closure returning a string) resolve without throwing, and each target is still excluded
- an existing `customFields` entry is not overwritten
- a plugin-registered entity is seeded, a phantom entity present only in the global metadata is not

```
✓ packages/core unit suite: 1519 tests, 109 files
✓ e2e (sqlite): custom-fields 79, plugin + list-query-builder 241
eslint: 0 errors
tsc: no new errors
```

Rebased onto `minor`.

Relates to #4965

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5069"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788010544&installation_model_id=20193&pr_number=5069&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5069&signature=2e6e5f3fd971269bbc2b7aa6f5f3ed6f86fdeb2baf2e5f9ac8ae52c31164192b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld:
Reviewed the OSS-654 commits only (`c5260a594..HEAD`). Right idea — the three heuristics needed collapsing. Some things to fix before it lands.

### Must fix

**1. `config.service.spec.ts` "does not overwrite an explicitly-configured entity" is vacuous.**

`entities: []` means `getEntityNamesWithCustomFields([])` returns `[]`, so the loop body never executes and the guard under test is never reached. Confirmed by replacing

```ts
if (!definedCustomFields[entityName]) {
    definedCustomFields[entityName] = [];
}
```

with an unconditional `definedCustomFields[entityName] = [];` — both tests still pass.

Put `Product` in the `entities` array so the loop runs, and assert identity (`toBe(existing)`, as `bootstrap.spec.ts:113` does) rather than `toEqual` on a literal, which would also pass against a fresh clone.

**2. The new spec doesn't typecheck.**

```
src/config/config.service.spec.ts(61,21): error TS2540: Cannot assign to 'embeddeds' because it is a read-only property.
src/config/config.service.spec.ts(64,21): error TS2540: Cannot assign to 'relations' because it is a read-only property.
```

It gets past CI because the build tsconfig compiles from `src/index.ts` and vitest transpiles without checking, but `tsc -p tsconfig.json` and any editor will flag it. Use `splice`/`pop` for teardown like the other specs.

**3. `getTranslationEntityNames`'s doc comment describes sharing that doesn't exist.**

The JSDoc says it's "shared across custom-field registration (`getEntityNamesWithCustomFields` and the `ConfigService.customFields` getter)". `ConfigService` imports `getEntityNamesWithCustomFields`, not this. Only caller in the repo is line 56 of the same file. Either unexport it (it's an internal detail of `getEntityNamesWithCustomFields`) and cut the doc down, or fix the wording.

Related: `registerCustomEntityFields` (~line 349) still does its own inline `translations` relation lookup. Fine in itself — it needs the specific relation, not a name set — but it means the description's "three notions collapsed into one" is really two call sites plus a third inline lookup left alone. Worth stating, or sharing the `'translations'` literal via a constant.

### Should fix

**4. Test setup duplication.** `bootstrap.spec.ts:21` already has `registerTranslationRelation()` for exactly this. The new spec re-rolls it inline with a fourth teardown idiom. Extract it to a shared test util and use it from both — four bespoke unwind paths over process-global TypeORM metadata is how you get a cross-file failure that costs an afternoon.

**5. The two call sites still source their entity list differently.** `bootstrap.ts:392` uses `getAllEntities(config)`; `config.service.ts:143` uses `this.dbConnectionOptions.entities` — the field the preceding commit (`c5260a594`) moved away from precisely because it's empty in the CLI/dashboard/codegen paths. In practice `preBootstrapConfig` populates it before any `ConfigService` exists, so it's likely unreachable, but nothing in the code enforces that ordering. `getAllEntities` living in `bootstrap.ts` makes the direct fix an import cycle, so at minimum drop the "cannot diverge" framing and leave a comment naming the invariant.

**6. The two guards aren't the same.** `bootstrap.ts:394` uses `Object.prototype.hasOwnProperty.call(...)`; `config.service.ts:147` uses truthiness. They differ on an explicit `undefined`/`null` entry. Harmless today, but odd in a PR about these two not diverging.

**7. Comments carrying history.** `config.service.ts:137-142` and the spec header both explain what the code used to do. Git has that. The *why* ("delegate to the shared detector so this and the bootstrap auto-init stay in sync") is the part worth keeping.

### Good

- `getRelationTargetName` handling the string / closure / constructor shapes, with a doc that names the actual failure mode (`relation.type is not a function` killing bootstrap), is a genuinely useful comment.
- The first test is a real test: RED without the change, exact assertions, one deterministic path.
- Scoping to `registeredEntityNames` before applying the exclusion set, and documenting why a superset exclusion set is safe, is the right call.

grolmus:
@biggamesmallworld addressed all five points:

1. **Vacuous "does not overwrite" test** — it now registers a supporting entity in `entities` so the seeding loop actually runs and reaches the guard, and asserts `toBe(existing)` (identity), not `toEqual` on a literal.
2. **TS2540 teardown** — the shared helper removes exactly what it pushed via reference-matched `splice` (order-independent); no more reassigning `storage.embeddeds`/`.relations`. `tsc -p tsconfig.json` is clean.
3. **Doc + centralization** — `getTranslationEntityNames` is now `private` (only ever used internally by `getEntityNamesWithCustomFields`), with the misleading "shared with the getter" wording removed. The `'translations'` literal is now a `TRANSLATIONS_RELATION_PROPERTY` constant used by both it and the inline lookup in `registerCustomEntityFields`.
4. **Test-util duplication** — extracted `registerCustomFieldEntityMetadata` into `src/testing/custom-field-metadata-test-utils.ts`; both `bootstrap.spec.ts` and `config.service.spec.ts` use it, and I converted the remaining hand-rolled `push`/`pop()` blocks in `bootstrap.spec.ts` (including the triple-`pop()` one) onto it, so there is a single teardown path.
5. **Diverging entity-list source** — dropped the "cannot diverge" framing. The comment now states that the classification logic is shared while the entity-list *source* differs (`dbConnectionOptions.entities` vs `getAllEntities(config)`), names the invariant (`preBootstrapConfig` populates the entities before any `ConfigService` exists), and notes that unifying the source would create a `bootstrap.ts` import cycle.

Verification: unit specs green (12, including all three relation-target shapes and the now-non-vacuous guard test), lint + `tsc` clean, and the `custom-fields` e2e suite (56 tests — real server boot with `localeString` custom fields on translatable entities) passes, so the `*TranslationInput` schema path is exercised end-to-end.

@michaelbromley this one is `@vendure/core`, so I'd value your eyes on it — could you check when you get a chance that the approach is sound? Requested you as a reviewer.

HouseinIsProgramming (dismissed):
Good fix, agreed on the approach.

Requested changes:

1. The description only mentions one direction of the behaviour change. Also declare the other one, and add a test for it: an entity whose name ends in `Translation` and has a `languageCode` column, but is not the target of a `translations` relation, was excluded before and now gets seeded.
   Without a test, a later refactor can flip this back without anyone noticing.

2. Mention the behaviour change somewhere plugin authors will see it, either a changelog note or a line in the custom-fields docs.
   The PR title starts with `refactor(core):`, so it won't appear in the changelog at all.

Will approve once that's in.

grolmus:
@HouseinIsProgramming addressed both points:

1. Declared the **other direction** in the description ("Behaviour changes (both directions)") and added a regression test (`b20d33c35`): an entity whose name ends in `Translation` with a `languageCode` column but no `translations` relation pointing at it is now **seeded** (was excluded by the old name+column heuristic). Guards against a flip back.
2. Documented the detection rule where plugin authors will see it — a note in the **translatable guide** (`289f72241`), since the `refactor(core):` title keeps it out of the changelog.

Re-requested your review.

biggamesmallworld:
@grolmus this is looking good, but would need to rebase before we can approve and merge 👍

biggamesmallworld (approved):

biggamesmallworld:
Rebased onto `minor` and took over the branch, since this had been sitting conflicted.

**What changed in the rebase**

- Dropped the first two commits (`ad9742eb`, `c5260a59`). Their content already landed on `minor` via #5068, in a later form that derives the seed entities from `getAllEntities(config)` rather than `dbConnectionOptions`.
- Replayed the nine translation-detection commits cleanly on top.
- Added one commit: the plugin-registered-entity spec that came in with #5068 pushed its `customFields` embedded onto the global metadata storage by hand, which no longer compiled once this branch moved that import into the shared test util. It now goes through `registerCustomFieldEntityMetadata` like the rest of the file.

`packages/core` unit tests pass: 1521 tests, 110 files.

**One thing for the reviewers**

The premise of this PR is consolidating the several "is this a translation entity?" checks onto one signal, the target of a `translations` relation. Since the PR was opened, `minor` gained another one: `registerTranslationEntityUniqueConstraints` in `packages/core/src/entity/register-translation-unique-constraints.ts` detects translation entities by a `languageCode` column plus a `base` relation. So after this merges there are still two independent definitions in core, and they do not agree on the same set of entities.

I have deliberately not folded that in here, since it widens the scope and the two checks run at different points in bootstrap. Worth deciding whether it is a follow-up or belongs in this PR.

gabriellbui (commented):
@biggamesmallworld before this merges, can you check whether the seeding loop in `ConfigService.getCustomFieldsForAllEntities()` can be deleted instead of rewired?

Since #5068, `runPluginConfigurations()` already seeds an empty `customFields` array for every entity that supports them, and it runs in `preBootstrapConfig()` before the Nest app (and so any `ConfigService`) exists. The CLI and dashboard schema generators call `runPluginConfigurations()` directly and never build a `ConfigService`. So by the time the getter runs, every key it would add is already there.

I tried it locally: replaced the method body with `return this.activeConfig.customFields;`, rebuilt core, and ran `custom-fields`, `custom-field-relations`, `plugin` and `list-query-builder` e2e on sqlite. All 274 tests pass, same as on the branch. The core unit suite passes too, except the two new tests in `config.service.spec.ts` that test the loop itself.

If that holds, the PR gets smaller: no second seeding site to keep in sync, no comment about the ordering between `dbConnectionOptions.entities` and `getAllEntities()`, and `config.service.spec.ts` plus the shared test util are not needed. The "lookalike Translation entity is seeded" test could move to `bootstrap.spec.ts`. The `getTranslationEntityNames` helper, the `'translations'` constant and the docs note would stay.

I'm not certain there is no path I missed, so I'd like a second pair of eyes on it. The only cases I found where the loop still does something are a `ConfigService` built without bootstrap (unit tests only) and a plugin `configuration` callback that replaces `config.customFields` with a new object without spreading the old one.

biggamesmallworld:
@gabriellbui you were right, the loop is dead. I traced it before deleting:

- `ConfigService` is only ever constructed by Nest DI, and the Nest app is built after `preBootstrapConfig()` returns (`bootstrap.ts:211`, `bootstrap.ts:282`).
- `preBootstrapConfig()` calls `runPluginConfigurations()` at `bootstrap.ts:335`, which seeds the keys.
- Both sites read the same entity set: `preBootstrapConfig` sets `dbConnectionOptions.entities` from `getAllEntities(userConfig)` at `bootstrap.ts:317`, which is the same list `runPluginConfigurations` derives its own from.

So every key the getter would add is present before any `ConfigService` exists, and nothing after bootstrap can add an entity to either list.

Deleted, and the PR is smaller for it:

- `ConfigService.getCustomFieldsForAllEntities()` and the `allCustomFieldsConfig` cache are gone. The getter returns `activeConfig.customFields`, with a comment naming who seeds it and when.
- `config.service.spec.ts` deleted. Its one case that tests the shared detection rather than the loop, the lookalike translation entity that now gets seeded, moved to `bootstrap.spec.ts`.
- `src/testing/custom-field-metadata-test-utils.ts` deleted. `bootstrap.spec.ts` was its only remaining consumer, so the helper moved back into that file.
- The comment about the ordering between `dbConnectionOptions.entities` and `getAllEntities()` is gone with the code it described.

The one case where the loop still did something: a plugin `configuration` callback that replaces `config.customFields` wholesale without spreading. That already discards every user-configured custom field, so it is broken either way. I did not add a guard for it.

Also applied @vendure-oss-bot's suggestion on the translatable guide, and rebased onto `minor`.

Verification: `packages/core` unit suite 1519 tests / 109 files, e2e on sqlite `custom-fields` 79 and `plugin` + `list-query-builder` 241, eslint 0 errors, `tsc -p tsconfig.json` no new errors.

On the second heuristic I raised earlier: `registerTranslationEntityUniqueConstraints` is still a separate definition and still disagrees with this one. Opened #5327 for it rather than widening this PR. Description updated to say so.

@HouseinIsProgramming @gabriellbui re-requesting review, since the shape of the change moved since you last looked.

PR #5072, opened 2026-07-30
fix(core): Enforce channel permission check on Promotion assign/remove

### LINKED ISSUE #5071: Any Administrator with UpdatePromotion permission can assign a Promotion to, or remove one from, a Channel they have no permission on
Description

The assignPromotionsToChannel mutation lets any Administrator holding the UpdatePromotion permission assign an existing Promotion to any Channel in the system, by id.

This works regardless of whether the requester has any permission at all in the target Channel. The mutation only checks whether the requester has UpdatePromotion somewhere, in whichever Channel their session is currently active in. It never checks whether they actually have permission in the Channel they are trying to push the Promotion into.

This is a different kind of gap from a typical missing ownership check. Rather than reaching out to touch something that already belongs to someone else, this lets an administrator push something they legitimately own into a space they have no right to be in.

Several sibling entities in the same codebase, Collections, Facets, ProductOptionGroups, ShippingMethods, PaymentMethods, and ProductVariants, all correctly check permission on the target channel before allowing this kind of assignment. Promotion is the one place where that check appears to be missing, which suggests an oversight rather than a deliberate design choice.

How to reproduce

Setup:

As SuperAdmin, create a Role scoped to one Channel (for example shop-a), with permissions CreatePromotion and UpdatePromotion. Assign it to a test Administrator (for example shop-a-owner).
Note the numeric id of a second, unrelated Channel (for example shop-b). This is visible directly in its settings page URL in the dashboard, no special permission needed to find it this way.
Reproduction script (replace EMAIL, PASSWORD, and SHOPB_ID with your own test values):

#!/bin/bash
set +H
BASE_URL="http://localhost:3000/admin-api"
SHOP_A_TOKEN="shop-a"
EMAIL="shop-a-owner@example.com"
PASSWORD="REPLACE_ME"
SHOPB_ID="REPLACE_ME"
COOKIE=$(mktemp)
trap 'rm -f "$COOKIE"' EXIT

echo "== STEP 1: Log in as shop-a-owner (only scoped to shop-a) =="
curl -s -X POST "$BASE_URL" -H "Content-Type: application/json" -H "vendure-token: $SHOP_A_TOKEN" -c "$COOKIE" \
  -d "{\"query\":\"mutation(\$u:String!,\$p:String!){ login(username:\$u,password:\$p){ __typename ... on CurrentUser { identifier } ... on ErrorResult { message } } }\",\"variables\":{\"u\":\"$EMAIL\",\"p\":\"$PASSWORD\"}}"
echo -e "\n"

echo "== STEP 2: shop-a-owner creates a promotion, legitimately, inside his own shop-a =="
PROMO_RESP=$(curl -s -X POST "$BASE_URL" -H "Content-Type: application/json" -H "vendure-token: $SHOP_A_TOKEN" -b "$COOKIE" \
  -d '{"query":"mutation { createPromotion(input:{enabled:true, translations:[{languageCode:en,name:\"ShopA Promo\",description:\"\"}], couponCode:\"SHOPAPROMOUNIQUE\", conditions:[], actions:[{code:\"order_percentage_discount\",arguments:[{name:\"discount\",value:\"10\"}]}]}){ __typename ... on Promotion { id name } ... on ErrorResult { message } } }"}')
echo "$PROMO_RESP"
PROMO_ID=$(echo "$PROMO_RESP" | sed -n 's/.*"id":"\([^"]*\)".*/\1/p')
echo "promotion id: $PROMO_ID"
echo

echo "== STEP 3: push shop-a-owner's promotion into shop-b, a channel he has zero relation to =="
curl -s -X POST "$BASE_URL" -H "Content-Type: application/json" -H "vendure-token: $SHOP_A_TOKEN" -b "$COOKIE" \
  -d "{\"query\":\"mutation(\$pid:ID!,\$cid:ID!){ assignPromotionsToChannel(input:{promotionIds:[\$pid], channelId:\$cid}){ id name } }\",\"variables\":{\"pid\":\"$PROMO_ID\",\"cid\":\"$SHOPB_ID\"}}"
echo -e "\n"
Result observed, running this against a real instance:

{"data":{"createPromotion":{"__typename":"Promotion","id":"3","name":"ShopA Promo"}}}
{"data":{"assignPromotionsToChannel":[{"id":"3","name":"ShopA Promo"}]}}
No error at any step. shop-a-owner successfully pushed his own Promotion into shop-b, a Channel he has no role or permission on whatsoever.

Suggested fix

Add the same target channel permission check that already exists in the equivalent methods for Collection, Facet, ProductOptionGroup, ShippingMethod, PaymentMethod, and ProductVariant. Before assigning, verify that the requester holds the required permission specifically on input.channelId, not just somewhere in their current session, and throw a ForbiddenError if not.

Files and line numbers

packages/core/src/service/services/promotion.service.ts
Around lines 207 to 236, assignPromotionsToChannel and removePromotionsFromChannel, missing the target channel permission check
packages/core/src/api/resolvers/admin/promotion.resolver.ts
Around lines 138 to 153, gated only by @Allow(Permission.UpdatePromotion), with no check against the target channel

latifniz:
Fixes #5071

An administrator who only manages one shop could push their own promotion into a completely different shop, or pull an existing promotion out of it, even with zero permission there. The action never checked whether the person doing it was actually allowed to touch that other shop, it only checked that they were allowed to manage promotions somewhere.

What this fix does

Adds one check before either action runs: does the person making the request actually have permission in the specific shop they're trying to push into or pull from. If not, it's rejected. Same idea already used correctly elsewhere in the codebase for similar actions on other entity types, just applied here too.

Tests

Added two tests that create a real, narrowly scoped staff account (not a superadmin) and confirm it gets correctly blocked when trying to touch a shop it has no permission on. Confirmed these tests actually fail without the fix and pass with it, and nothing else in the existing test file broke.

Breaking changes

None. Nothing changes for anyone acting within a shop they're already allowed to manage.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5072"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788026438&installation_model_id=20193&pr_number=5072&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5072&signature=91ed6b7a1e2c38bf73112de66dd03c8c24ffaac80a0fe967442fc61ba5b77fc2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/promotion.e2e-spec.ts
+++ b/packages/core/e2e/promotion.e2e-spec.ts
@@ -1,4 +1,10 @@
-import { CurrencyCode, DeletionResult, ErrorCode, LanguageCode } from '@vendure/common/lib/generated-types';
+import {
+    CurrencyCode,
+    DeletionResult,
+    ErrorCode,
+    LanguageCode,
+    Permission,
+} from '@vendure/common/lib/generated-types';
 import { pick } from '@vendure/common/lib/pick';
 import { ConfigService, type PromotionAction, PromotionCondition, PromotionOrderAction } from '@vendure/core';
 import {
@@ -17,8 +23,10 @@ import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-conf
 
 import {
     assignPromotionsToChannelDocument,
+    createAdministratorDocument,
     createChannelDocument,
     createPromotionDocument,
+    createRoleDocument,
     deletePromotionDocument,
     getAdjustmentOperationsDocument,
     getPromotionDocument,
@@ -384,6 +392,63 @@ describe('Promotion resolver', () => {
                 id: promotion.id,
             });
             expect(result?.name).toBe('test promotion');
+        });
+
+        describe('cannot assign/remove without permission on the target channel', () => {
+            const RESTRICTED_ADMIN_EMAIL = 'promotion-channel-permission-test@e2e.com';
+            const RESTRICTED_ADMIN_PASSWORD = 'test';
+
+            beforeAll(async () => {
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                await adminClient.asSuperAdmin();
+                const { createRole } = await adminClient.query(createRoleDocument, {
+                    input: {
+                        code: 'promotion-channel-permission-test-role',
+                        description: '',
+                        channelIds: [secondChannel.id],
+                        permissions: [Permission.UpdatePromotion],
+                    },
+                });
+                await adminClient.query(createAdministratorDocument, {
+                    input: {
+                        firstName: 'Restricted',
+                        lastName: 'Admin',
+                        emailAddress: RESTRICTED_ADMIN_EMAIL,
+                        password: RESTRICTED_ADMIN_PASSWORD,
+                        roleIds: [createRole.id],
+                    },
+                });
+                adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
+                await adminClient.asUserWithCredentials(RESTRICTED_ADMIN_EMAIL, RESTRICTED_ADMIN_PASSWORD);
+            });
+
+            afterAll(async () => {
+                await adminClient.asSuperAdmin();
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+            });
+
+            it('cannot assign a Promotion to a channel it has no permission on', async () => {
+                // the restricted admin only has UpdatePromotion on secondChannel, not on the default channel
+                await expect(
+                    adminClient.query(assignPromotionsToChannelDocument, {
+                        input: {
+                            channelId: 'T_1',
+                            promotionIds: [promotion.id],
+                        },
+                    }),
+                ).rejects.toThrow(/not currently authorized/);
+            });
+
+            it('cannot remove a Promotion from a channel it has no permission on', async () => {
+                await expect(
+                    adminClient.query(removePromotionsFromChannelDocument, {
+                        input: {
+                            channelId: 'T_1',
+                            promotionIds: [promotion.id],
+                        },
+                    }),
+                ).rejects.toThrow(/not currently authorized/);
+            });
         });
     });
 
--- a/packages/core/e2e/promotion.e2e-spec.ts
+++ b/packages/core/e2e/promotion.e2e-spec.ts
@@ -401,6 +401,18 @@ describe('Promotion resolver', () => {
             beforeAll(async () => {
                 adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
                 await adminClient.asSuperAdmin();
+
+                // The restricted admin can only act on Promotions it can already see in its
+                // own active channel (secondChannel). Put the Promotion there first, as
+                // SuperAdmin, so the tests below exercise the real assign/remove logic
+                // instead of being stopped earlier by the channel-visibility filter.
+                await adminClient.query(assignPromotionsToChannelDocument, {
+                    input: {
+                        channelId: secondChannel.id,
+                        promotionIds: [promotion.id],
+                    },
+                });
+
                 const { createRole } = await adminClient.query(createRoleDocument, {
                     input: {
                         code: 'promotion-channel-permission-test-role',
@@ -425,6 +437,25 @@ describe('Promotion resolver', () => {
             afterAll(async () => {
                 await adminClient.asSuperAdmin();
                 adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                await adminClient.query(removePromotionsFromChannelDocument, {
+                    input: {
+                        channelId: secondChannel.id,
+                        promotionIds: [promotion.id],
+                    },
+                });
+            });
+
+            it('can assign a Promotion to a channel it has permission on', async () => {
+                // Positive control: proves the restricted admin's permissions genuinely work,
+                // so the rejections below can be trusted to be specifically about the target
+                // channel, not some other authorization gate.
+                const { assignPromotionsToChannel } = await adminClient.query(assignPromotionsToChannelDocument, {
+                    input: {
+                        channelId: secondChannel.id,
+                        promotionIds: [promotion.id],
+                    },
+                });
+                expect(assignPromotionsToChannel).toEqual([{ id: promotion.id, name: promotion.name }]);
             });
 
             it('cannot assign a Promotion to a channel it has no permission on', async () => {
--- a/packages/core/e2e/promotion.e2e-spec.ts
+++ b/packages/core/e2e/promotion.e2e-spec.ts
@@ -402,10 +402,8 @@ describe('Promotion resolver', () => {
                 adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
                 await adminClient.asSuperAdmin();
 
-                // The restricted admin can only act on Promotions it can already see in its
-                // own active channel (secondChannel). Put the Promotion there first, as
-                // SuperAdmin, so the tests below exercise the real assign/remove logic
-                // instead of being stopped earlier by the channel-visibility filter.
+                // Make the Promotion visible in secondChannel so the restricted admin can
+                // reach it below, rather than being stopped by the visibility filter.
                 await adminClient.query(assignPromotionsToChannelDocument, {
                     input: {
                         channelId: secondChannel.id,
@@ -446,9 +444,7 @@ describe('Promotion resolver', () => {
             });
 
             it('can assign a Promotion to a channel it has permission on', async () => {
-                // Positive control: proves the restricted admin's permissions genuinely work,
-                // so the rejections below can be trusted to be specifically about the target
-                // channel, not some other authorization gate.
+                // Positive control for the rejections below.
                 const { assignPromotionsToChannel } = await adminClient.query(assignPromotionsToChannelDocument, {
                     input: {
                         channelId: secondChannel.id,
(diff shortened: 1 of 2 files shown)


michaelbromley:
The service-level fix looks right, and it matches what `assignProductVariantsToChannel` and `asset.service.ts` already do.

The two new e2e tests don't test what they claim to, though. By the time that `describe` block runs, an earlier test in the same file has already removed the shared `promotion` from `secondChannel`, so it only exists in the default channel. The restricted admin's session is on `secondChannel`, and both mutations look the promotion up via `findByIdsInChannel(ctx, Promotion, ids, ctx.channelId, {})`, so the promotion is invisible to it and the lookup comes back empty.

Revert the service change and run them — they do still fail, but like this:

```
AssertionError: promise resolved "{ assignPromotionsToChannel: [] }" instead of rejecting
```

The mutation never gets as far as the line that performs the cross-channel assign. All that's being exercised is the channel visibility filter.

To hit the actual path, the restricted admin needs a promotion it legitimately owns — created in or assigned to `secondChannel` — and then attempts to push that into `T_1`. That version genuinely succeeds before the fix and is rejected after it.

packages/core/e2e/promotion.e2e-spec.ts
```diff
@@ -385,6 +393,63 @@ describe('Promotion resolver', () => {
...
+                await expect(
+                    adminClient.query(removePromotionsFromChannelDocument, {
+                        input: {
+                            channelId: 'T_1',
+                            promotionIds: [promotion.id],
+                        },
+                    }),
+                ).rejects.toThrow(/not currently authorized/);
+            });
+        });
```
biggamesmallworld, line 479: Positive control added: a non-empty result proves the guard was passed, so the two rejections are attributable to `PromotionService`. Also needs its own fixture — the shared `promotion` was already removed from `secondChannel` above, and `assignPromotionsToChannel` looks up by `ctx.channelId`, so a positive control against it would resolve to `[]`. `DEFAULT_CHANNEL_ID` const because bare `'T_1'` sits 40 lines from `defaultTaxZoneId: 'T_1'`.

Check it bites: revert `promotion.service.ts` → both negatives must fail. Drop `channelIds: [secondChannel.id]` → positive must fail.

```suggestion
        describe('channel permissions', () => {
            const RESTRICTED_ADMIN_EMAIL = 'promotion-channel-permission-test@e2e.com';
            const RESTRICTED_ADMIN_PASSWORD = 'test';
            const DEFAULT_CHANNEL_ID = 'T_1';
            let channelScopedPromotion: PromotionFragment;

            beforeAll(async () => {
                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
                await adminClient.asSuperAdmin();

                // A dedicated Promotion which lives in secondChannel, so that the positive control
                // below does real work rather than resolving to an empty result set.
                const { createPromotion } = await adminClient.query(createPromotionDocument, {
                    input: {
                        enabled: true,
                        couponCode: 'CHANNEL_PERMS',
                        translations: [
                            {
                                languageCode: LanguageCode.en,
                                name: 'channel permission test promotion',
                            },
                        ],
                        conditions: [
                            {
                                code: promoCondition.code,
                                arguments: [{ name: 'arg', value: '500' }],
                            },
                        ],
                        actions: [
                            {
                                code: promoAction.code,
                                arguments: [{ name: 'facetValueIds', value: '["T_1"]' }],
                            },
                        ],
                    },
                });
                promotionGuard.assertSuccess(createPromotion);
                channelScopedPromotion = createPromotion;
                await adminClient.query(assignPromotionsToChannelDocument, {
                    input: {
                        channelId: secondChannel.id,
                        promotionIds: [channelScopedPromotion.id],
                    },
                });

                // The role grants UpdatePromotion on secondChannel only, and the client's active
                // channel is secondChannel. So the resolver's @Allow(UpdatePromotion) guard always
                // passes, and any rejection below can only come from the channel permission check
                // inside PromotionService.
                const { createRole } = await adminClient.query(createRoleDocument, {
                    input: {
                        code: 'promotion-channel-permission-test-role',
                        description: '',
                        channelIds: [secondChannel.id],
                        permissions: [Permission.UpdatePromotion],
                    },
                });
                await adminClient.query(createAdministratorDocument, {
                    input: {
                        firstName: 'Restricted',
                        lastName: 'Admin',
                        emailAddress: RESTRICTED_ADMIN_EMAIL,
                        password: RESTRICTED_ADMIN_PASSWORD,
                        roleIds: [createRole.id],
                    },
                });
                adminClient.setChannelToken(SECOND_CHANNEL_TOKEN);
                await adminClient.asUserWithCredentials(RESTRICTED_ADMIN_EMAIL, RESTRICTED_ADMIN_PASSWORD);
            });

            afterAll(async () => {
                await adminClient.asSuperAdmin();
                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
            });

            // Positive control: proves the resolver guard is passed by this admin, so the two
            // rejections below are attributable to the service-level check and not to @Allow.
            it('allows removal from a channel the admin does have permission on', async () => {
                const { removePromotionsFromChannel } = await adminClient.query(
                    removePromotionsFromChannelDocument,
                    {
                        input: {
                            channelId: secondChannel.id,
                            promotionIds: [channelScopedPromotion.id],
                        },
                    },
                );

                expect(removePromotionsFromChannel).toEqual([
                    { id: channelScopedPromotion.id, name: channelScopedPromotion.name },
                ]);
            });

            it(
                'rejects assignment to a channel the admin has no permission on',
                assertThrowsWithMessage(
                    () =>
                        adminClient.query(assignPromotionsToChannelDocument, {
                            input: {
                                channelId: DEFAULT_CHANNEL_ID,
                                promotionIds: [channelScopedPromotion.id],
                            },
                        }),
                    'You are not currently authorized to perform this action',
                ),
            );

            it(
                'rejects removal from a channel the admin has no permission on',
                assertThrowsWithMessage(
                    () =>
                        adminClient.query(removePromotionsFromChannelDocument, {
                            input: {
                                channelId: DEFAULT_CHANNEL_ID,
                                promotionIds: [channelScopedPromotion.id],
                            },
                        }),
                    'You are not currently authorized to perform this action',
                ),
            );
        });
```

biggamesmallworld (commented):
Service fix looks correct — permission matches the resolver `@Allow`, check runs before any read inside the transaction, no DI cycle.

Tests need a positive control. Both assertions match `/not currently authorized/`, which is also what the resolver's `@Allow(UpdatePromotion)` guard throws — so they can't tell the new service check from the pre-existing guard. They only hit the new code by coincidence of the setup. Suggestion inline.

biggamesmallworld:
Three non-blocking notes:

**1.** Six of seven sibling services use the array form — `userHasAnyPermissionsOnChannel(ctx, input.channelId, [Permission.UpdatePromotion])`. Identical behaviour (`role.service.ts:172-178` delegates), but already shaped for an umbrella permission later.

**2.** Every sibling `removeXFromChannel` gates on a Delete permission; Promotion's resolver only requires `UpdatePromotion`, and this PR correctly mirrors that rather than making a breaking `@Allow` change. Worth a line in the description so it's a recorded decision, not an oversight.

**3.** Follow-up issue, not this PR: the `userHasAnyPermissionsOnChannel` + `throw new ForbiddenError()` pair now appears ~16 times across the service layer. That duplication is why Promotion got missed.

latifniz:
let me take a closer look at all those feedbacks and do the fixes.

latifniz:
Thanks for the review, both. Pushed two commits for this.

What changed:

The restricted admin now has a promotion it can actually reach. It gets assigned into secondChannel before the admin logs in.
Because of that, the two rejection tests now exercise the real assign/remove code. Before, they were stopped early by the channel visibility check instead.
Added a test proving the admin can successfully assign into a channel it does have permission on. This shows the rejections are caused by the new channel check, not something else.
Switched the permission check to the array form, matching the other services.

@biggamesmallworld — I tried your suggested version first. I kept most of it, but changed the positive control from remove back to assign.

Remove pulls the promotion out of the admin's own channel as a side effect.
That made it unreachable for the two tests right after.
When I reverted the fix to double check, those two tests failed for the wrong reason again (an empty result, not a real rejection).

Assign doesn't have that side effect, so I kept it instead.

Once this merges, I'd like to pick up the follow up issue about the duplicated permission check pattern. Feel free to assign it to me.

biggamesmallworld (approved):
Re-reviewed the two new commits. The test setup fix is right — assigning the promotion into `secondChannel` before the restricted admin logs in means the rejection tests now exercise the actual permission check rather than dying at the channel-visibility filter. I verified this empirically: with the check neutralised in the service, both negative tests fail as expected, so they genuinely guard the bypass.

The service fix matches the pattern in `assignProductVariantsToChannel` and `asset.service.ts`, and the check sits ahead of any mutation with `@Transaction()` on the resolver, so a rejection can't leave partial state behind.

One nit for whenever you next touch this: `packages/core/e2e/promotion.e2e-spec.ts` doesn't pass `npx prettier --check` (the `adminClient.query(assignPromotionsToChannelDocument, {` call needs the multi-line argument form). Not blocking.

Two things this surfaced that are out of scope here — filing separately:
- `removePromotionsFromChannel` has no default-channel guard, unlike `shipping-method.service.ts` and `stock-location.service.ts`.
- `userHasAnyPermissionsOnChannel` rejects SuperAdmin on channels created programmatically via `channelService.create()`, since the SuperAdmin role is only assigned in `ChannelResolver.createChannel`. Pre-existing across all six services using this pattern.

LGTM.

PR #5074, opened 2026-07-30
feat(core): add index support for custom fields

### LINKED ISSUE #3444: Support indexes on customFields
Allow settings `index: true` on a customField so we can have indexes on customFields.

Indexes for customFields for improved query performance 

Manually adding the index in the migrations => works, but warnings on startup and will be removed in the next migration that gets generated.

Hemang-ai:
# feat(core): Add index support for custom fields

## Description

Adds an `index: true` option for non-list custom fields so frequently filtered or
sorted custom data can be backed by a database index.

- Registers a TypeORM index for scalar and to-one relation custom fields.
- Avoids registering a redundant index when `unique: true` already supplies one.
- Excludes list and struct fields from the type-safe option because their storage and
  indexing semantics are not portable across Vendure's supported databases.
- Adds end-to-end coverage that checks the generated entity metadata.

Fixes #3444.

## Why

Today an index added manually in a migration is absent from Vendure's entity metadata,
which can produce schema warnings or cause a later generated migration to remove it.
Declaring the index with the custom field keeps configuration, ORM metadata, and
migrations aligned.

## Verification

- `build:core-common` — passed
- `packages/core/e2e/custom-fields.e2e-spec.ts` — 57 passed
- ESLint — passed
- Prettier check — passed

## Breaking changes

None. The new option is additive and defaults to the current behavior.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5074"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788031048&installation_model_id=20193&pr_number=5074&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5074&signature=865b1823ab3f787a2ab09c157917a5447823951d6ad5f1b14542a6efd47e1012"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Hemang-ai:

Hemang-ai:
recheck

biggamesmallworld (changes requested):
The mechanism is right and the diff is small. What's missing is the guard rails around it, and the two riskiest lines have no test.

### 1. `type: 'text'` with `index: true` type-checks and then breaks MySQL

`getColumnType` maps `text` and `localeText` to `longtext` on mysql/mariadb. MySQL rejects an index on a TEXT/LONGTEXT column with no prefix length:

```
CREATE INDEX IDX_x ON a (c);   -- c LONGTEXT
ERROR 1170 (42000): BLOB/TEXT column 'c' used in key specification without a key length
```

So this config compiles clean and fails at synchronize or migration on two of the four engines in the CI matrix:

```ts
{ name: 'notes', type: 'text', index: true }
```

The docblock excludes `struct` because "their database representation is not portable across the supported database engines". `text` has the same problem and is waved through.

To be fair, this is a pre-existing hole rather than one the PR opens: `unique: true` on a `text` field already reaches the mysql `Index({ unique: true })` branch and fails with the same 1170. So it is fine to fix it here or file it separately, but do one of the two rather than widening the hole silently.

### 2. The validation belongs in the validator, and there isn't any

`packages/core/src/entity/validate-custom-fields-config.ts` exists for this job. It runs from `bootstrap.ts:325` where `config.dbConnectionOptions.type` is in scope, it has a spec file, and `assetNonNullablesHaveDefaults` is the direct precedent: a cross-option rule that produces a readable error naming the entity and field.

This matters because the type-level exclusions are compile-time only. I checked: `{ list: true, index: true }` and `{ type: 'struct', index: true }` are both rejected for object literals, but neither is rejected for a config built dynamically, spread, or widened to `CustomFieldConfig[]`, which is how plugins routinely assemble custom fields. Runtime then calls `Index()` on a `simple-json` or `json` column with no error.

### 3. Put `index` on `BaseTypedCustomFieldConfig`

That is where `unique` and `nullable` live, and it is the same class of option: a column-level database concern. Hanging it off `TypedCustomSingleFieldConfig` with a conditional `never` is what forces this:

```ts
const indexed = 'index' in customField && customField.index === true;
```

The `in` check exists only to satisfy narrowing across the union. Move the property, enforce the list/struct/text rules in the validator, and it collapses to `customField.index === true`.

### 4. The relation branch has zero coverage

```ts
EntityId({ nullable: true })(instance, `${name}Id`);
if (indexed) {
    Index()(instance, name);
}
```

The index is declared on the relation property while the line above declares an id property on the same database column. Whether TypeORM resolves the index to the join column, and how it interacts with `${name}Id`, is exactly what a test is for. `{ type: 'relation', index: true }` type-checks, so the branch is reachable, and nothing in the diff exercises it.

### 5. The unique-dedup claim is untested

Bullet two of the description is this line, and its correctness is engine-dependent:

```ts
} else if (indexed && customField.unique !== true) {
```

mysql/mariadb take the `Index({ unique: true })` branch above; postgres and sqlite get the constraint from `options.unique`. No config in the suite sets both `unique` and `index`, so the branch never runs.

It also can't be asserted cleanly in this e2e spec, because Postgres records the unique constraint in `metadata.uniques` rather than `metadata.indices`, and branching on engine inside the test gives you something that passes on all four engines and verifies nothing. Call `registerCustomFieldsForEntity` once per `dbEngine` in a unit test and read `getMetadataArgsStorage().indices`. No server, no database, four engines in one run.

### 6. The one test that exists is near-tautological

```ts
const indexedCustomField = productMetadata.indices.find(index =>
    index.columns.some(column => column.propertyPath === 'customFields.indexedString'),
);
expect(indexedCustomField?.columns.map(column => column.propertyPath)).toEqual([
    'customFields.indexedString',
]);
```

The `find` predicate already guarantees that column is present, so the assertion re-checks the predicate and adds one new fact: the index has a single column. It does not assert there is exactly one such index, nor that it isn't unique.

```ts
const indices = productMetadata.indices.filter(
    i => i.columns.length === 1 && i.columns[0].propertyPath === 'customFields.indexedString',
);
expect(indices.length).toBe(1);
expect(indices[0].isUnique).toBe(false);
```

### 7. Docs are not updated

`docs/docs/guides/developer-guide/custom-fields/index.mdx` hand-maintains the option list: an anchor list around line 265 and one `####` section per option, with `unique` at line 487. A new option that isn't in there does not exist as far as users are concerned. Add the anchor and a section with the `<CustomFieldProperty required={false} type="boolean" />` block, mirroring `unique`, and note that adding `index: true` to an existing field is a schema change, so a migration must be generated and run.

### Minor

The docblock sentence "their database representation is not portable across the supported database engines" names no mechanism and reaches no consequence. Say what happens: mysql and mariadb store struct fields as `json`, which can't be indexed directly. Writing that honestly also forces the `text` question in point 1.

`// #3444 — indexed custom fields should be represented in database metadata` sits directly above an `it()` that says the same thing. Keep the issue number, drop the restatement.

Hemang-ai:
Thanks for the detailed review. I addressed the requested changes and merged the latest `minor` branch.

Changes made:

- moved `index` to the base custom-field configuration;
- added database-aware runtime validation before metadata registration;
- reject indexes on list, struct, and secret fields;
- reject `text` and `localeText` indexes or unique constraints on MySQL and MariaDB;
- added scalar, relation, and `unique + index` coverage across MySQL, MariaDB, PostgreSQL, and SQLite;
- strengthened the end-to-end assertion to require exactly one non-unique index;
- documented supported field types, database limitations, and migration requirements;
- resolved the conflicts with the current `minor` branch while preserving the newer secret-field and relation-option changes.

The two focused test files now pass all 41 tests. GitHub currently shows six workflows awaiting maintainer approval.

Please take another look when convenient. Thank you.

biggamesmallworld (approved):
Nice work! 👍

Hemang-ai:
Thank you! I also merged the latest `minor` branch and resolved the new bootstrap conflict.

The E2E matrix exposed duplicate TypeORM index metadata when the same custom-field configuration is registered more than once during the test bootstrap lifecycle. I added an idempotent registration guard and regression coverage for scalar, unique, and relation indexes across MySQL, MariaDB, PostgreSQL, and SQLite configurations.

The focused unit suite passes all 41 tests. CI is now running against commit `8984a6555`.

biggamesmallworld (approved):

Issue #4820, opened 2026-06-09
Extract configurable-operation translations out of core .ts files into lingui catalogs


BibiSebi:
### Background

PR #4806 added Czech (`LanguageCode.cs`) translations for configurable operations. To do so it had to edit ~20 core source files (`ShippingCalculator`, `ShippingEligibilityChecker`, all promotion actions/conditions, entity duplicators, payment/fulfillment handlers, collection filters) and append a `{ languageCode, value }` entry to every inline `LocalizedString[]` array.

```ts
// packages/core/src/config/shipping-method/default-shipping-calculator.ts
description: [
    { languageCode: LanguageCode.en, value: 'Default Flat-Rate Shipping Calculator' },
    { languageCode: LanguageCode.cs, value: 'Výchozí kalkulátor dopravy s pevnou sazbou' }, // ← added
],
```

### Problem

Inline `LocalizedString[]` arrays mean **every new locale linearly bloats every config file**. The translation content (a translator's concern) is interleaved with operation logic (a developer's concern), in the wrong file type, with no tooling. This pattern doesn't scale past a couple of languages and diverges from how the rest of the packages handle i18n (lingui `.po` catalogs — e.g. `packages/dashboard/src/i18n/locales/cs.po`).

### Proposed approach

Extract configurable-operation labels/descriptions out of the `.ts` source into dedicated locale catalogs, keep `LanguageCode.en` inline as the source/default, and use **lingui** (consistent with the dashboard & other packages). At build time, copy/merge the compiled catalogs into the dashboard.

Sketch:
- Source `.ts` keeps only the English default (or a stable message key).
- Translations live in per-locale catalog files owned by translators, not developers.
- A build step compiles + copies catalogs into the dashboard bundle.
- Server resolves `LocalizedString` for a given `languageCode` from the catalog rather than from the inline array.

### Open questions / decisions

- **Key strategy** — derive translation keys from operation `code` + arg name (e.g. `default-shipping-calculator.args.rate.label`), or use the English string as the msgid (lingui default)?
- **Backwards compatibility** — third-party plugins still ship inline `LocalizedString[]`. The resolver must fall back to inline arrays so existing plugins keep working.
- **Server vs dashboard** — these labels are consumed both server-side (Admin API responses) and in the dashboard UI. Where does resolution happen, and how do catalogs reach both?
- **Build wiring** — exact mechanism for the "copy to dashboard at build" step.

### Scope

- [ ] Define catalog format + key strategy
[...]

PR #5079, opened 2026-07-31
feat(core): Resolve configurable operation strings from i18n catalogs

michaelbromley:
Configurable operation descriptions and arg labels are currently defined inline in each operation's `.ts` file as a `LocalizedStringArray`. Supporting a new language means editing every one of those files, which is why core has never gone beyond a handful of languages for these strings while the dashboard has reached 28.

This adds a second source for those strings: the server-side i18n message catalogs core already ships. There is no user-visible change in this PR — the mechanism lands, and the 1242 translations it enables are a follow-up.

## Approach

Resolution happens server-side in `toGraphQlType`, so the GraphQL contract is unchanged and existing API consumers keep working.

The keys are derived from the operation itself, so `ConfigurableOperationDefOptions` is untouched and every existing plugin works unmodified:

```text
configurableOperation.<type>.<code>.description
configurableOperation.<type>.<code>.args.<argName>.label
configurableOperation.<type>.<code>.args.<argName>.description
configurableOperation.<type>.<code>.args.<argName>.options.<optionValue>.label
```

`<type>` is the registry the operation belongs to. It has to be part of the key because a code is only unique within its own registry — core itself uses `buy_x_get_y_free` for both a `PromotionCondition` and a `PromotionAction`.

English stays inline. The catalogs hold translations only, so a missing translation falls back to the English already in the source file rather than to a raw key, and adding a 29th language adds zero lines to any `.ts` file.

### Resolution order

For each language in turn — the requested language, then the Channel's default, then English — the catalog is checked first, then the inline array:

```ts
for (const languageCode of [ctx.languageCode, ctx.channel.defaultLanguageCode, DEFAULT_LANGUAGE_CODE]) {
    const fromCatalog = this.translator?.getConfigurableOperationTranslation(...);
    if (fromCatalog) return fromCatalog;
    const fromInline = inline?.find(x => x.languageCode === languageCode);
    if (fromInline) return fromInline.value;
}
return inline?.[0]?.value;
```

Checking both sources per language rather than exhausting the catalogs first matters: the other ordering would reach core's English catalog entry before a plugin's own German inline string, silently breaking every plugin that ships non-English strings today.

Catalog beating inline for the same language is deliberate, and is what allows a user to override a third-party operation's wording via `addTranslation()` without forking it.

### Why the lookup bypasses `t()`

Six core descriptions are templates rather than sentences, such as `Discount order by { discount }%`. The Admin UI substitutes the live argument values client-side, so the braces have to survive the round trip untouched. Core's i18next is initialised with `.use(ICU)`, which would treat `{ discount }` as an argument reference.

`I18nService.getConfigurableOperationTranslation()` therefore reads the resource straight out of the store and walks the path segment by segment. That also sidesteps i18next's key splitting, so a plugin is free to use a code containing dots such as `acme.shipping.handler`.

## Changes

- Resolution and key derivation on `ConfigurableOperationDef`, covering the operation description, each arg's label and description, and `ui.options[].label`
- `getConfigurableOperationTranslation()` on `I18nService`, plus a `configurableOperation` member on `VendureTranslationResources` (optional, so existing typed resource objects still compile)
- A `Symbol` token in `common/constants.ts`, registered by `I18nModule` as `useExisting: I18nService` and captured in the base `init(injector)`. A token rather than the class itself keeps the low-level `ConfigurableOperationDef` from importing the i18n module.
- `defType` on the nine `ConfigurableOperationDef` subclasses. `ConfigDefType` in the registry is now an alias of the new `ConfigurableOperationDefType` union, so indexing `ConfigDefTypeMap` by it keeps the two lists in step.
- `getTranslationKeys()`, which returns each key path with its English source value. Plugin authors can call it to check a key; the extraction tooling uses it as its input.
- `npm run i18n:extract` / `i18n:apply` in `packages/core`, modelled on the dashboard's tooling and sharing its `locale-profiles.js` script validation, so a batch labelled `ru` with no Cyrillic in it is refused rather than written to disk.
- Guide docs under "Server message translations".

## Behaviour changes

Two small ones, both improvements:

- An operation with an empty `description: []` array previously threw inside `localizeString`. It now resolves to an empty string.
- An arg `label` or `description` is now resolved even when the operation defines no inline array for it, so a label can be supplied purely from a catalog.

## Testing

- 17 unit tests for the resolution chain in `configurable-operation.spec.ts`, including the case that proves existing plugins are not broken, and the case proving the shared `ui` config is not mutated across requests
- 6 unit tests for the catalog lookup against a real i18next instance, covering dotted codes and verbatim placeholders
- 3 e2e tests in `translations.e2e-spec.ts` exercising the token-based DI wiring end to end, which is the part that would fail silently if the token were wrong
- Full core unit suite (1219 passed) and the e2e suites that exercise `toGraphQlType`: translations, promotion, order-promotion, promotion-side-effects, shipping-method, shipping-method-eligibility, collection, payment-method, payment-process, configurable-operation-secrets (235 passed)

## Follow-ups

1. Generate and land the `configurableOperation` translations for all 27 non-English locales. `i18n:extract` reports 46 distinct strings across 23 core operations, so 1242 translations. This is the point at which a Czech administrator finally sees Czech.
2. Backfill core's `error` / `errorResult` / `message` namespaces from 8 locales to 28. They have already drifted — `en.json` has 58 `error` keys, `de.json` has 45, `ru.json` has 44.
3. Custom field `label` and `description`. The same underlying pain, but a different architecture: `CustomField.label` goes over the wire as a full `LocalizedString[]` and is resolved client-side, so covering it means a breaking schema change.

Relates to #4820

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5079"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788091021&installation_model_id=20193&pr_number=5079&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5079&signature=3547e45ac3fe2a972c9d1e5de7d4269263cd0be2a7596d19beede3c2b4b62bca"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -400,3 +400,7 @@ package-lock.json
 
 # Agentic engineering
 plans/
+
+# Scratch files produced by the translation tooling
+missing-translations.txt
+translations.txt
--- a/bun.lock
+++ b/bun.lock
@@ -366,7 +366,7 @@
         "@types/react": "^19.2.14",
         "@types/react-dom": "^19.0.4",
         "@uidotdev/usehooks": "^2.4.1",
-        "@vendure-io/design-tokens": "^2.0.0-beta.7",
+        "@vendure-io/design-tokens": "^2.0.0-beta.9",
         "@vendure-io/ui": "^2.0.0-beta.16",
         "@vitejs/plugin-react": "^5.2.0",
         "acorn": "^8.16.0",
@@ -2649,7 +2649,7 @@
 
     "@vendure-io/design-lint": ["@vendure-io/design-lint@1.0.0-beta.2", "", { "peerDependencies": { "eslint": ">=9" }, "optionalPeers": ["eslint"] }, "sha512-h9q2QgY3oavSbt9llwYp8/6c+GWfydfYjOF4dgyhipfsEJQzWfCDGDyL1eT00KIifltKXTKLPQHliN2IrmpVFQ=="],
 
-    "@vendure-io/design-tokens": ["@vendure-io/design-tokens@2.0.0-beta.7", "", { "dependencies": { "@fontsource-variable/geist-mono": "^5.2.8", "@fontsource-variable/inter": "^5.2.8", "@fontsource-variable/public-sans": "^5.2.7", "@tailwindcss/typography": "^0.5.16", "shadcn": "^3.8.5", "tailwindcss": "^4.1.18", "tw-animate-css": "1.4.0" } }, "sha512-2cEDAJzieEMgqcANG7UVMm7GsIi3ONID35QF4Aql/p5zmUphfSLj4Z1EVeiX1Jb0SbqHaQyajb3qgQMEJd4TMg=="],
+    "@vendure-io/design-tokens": ["@vendure-io/design-tokens@2.0.0-beta.9", "", { "dependencies": { "@fontsource-variable/geist-mono": "^5.2.8", "@fontsource-variable/inter": "^5.2.8", "@fontsource-variable/public-sans": "^5.2.7", "tailwindcss": "^4.1.18", "tw-animate-css": "1.4.0" } }, "sha512-lwYVjhUJ0X8mr5xr4l0natwKgzPPtfiZSjDEzGPejPStQ0IzvJoY3QgFJ9TlQJWKClDD1CxGZJjJ01PKxosR4w=="],
 
     "@vendure-io/docs-generator": ["@vendure-io/docs-generator@0.1.1", "", { "dependencies": { "fs-extra": "^11.0.0", "klaw-sync": "^6.0.0", "typescript": "^5.0.0" } }, "sha512-tn1BUZuacKM4d99CuVQHb+rttZJANy5kuhfWmjwhyzXMx2Nvv41WguBxVpF9njT/E7iLfyvRtwin8dom4zJF2A=="],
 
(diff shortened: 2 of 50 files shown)


grolmus (commented):
Second-eyes review @michaelbromley — this is really cleanly built. The token-based `CONFIGURABLE_OPERATION_TRANSLATOR` (rather than importing `I18nService` into the low-level base class), reading via `getResourceBundle` + manual path walk instead of `t()` to preserve `{ arg }` placeholders, and the `ConfigDefType = ConfigurableOperationDefType` alias with the compile-error-on-omission guarantee are all the right calls. The resolution order (per-language: catalog → inline, so a plugin's own non-English inline string isn't beaten by core's English catalog) is subtle and correctly handled, and the test suite genuinely covers it — including the shared-`ui`-mutation case with the follow-up assertion that the source array isn't truncated. Nothing blocking below; one thing worth confirming and a few minor notes.

### Worth confirming: `ui.options[].label` shape narrowing

Previously `toGraphQlType` passed `ui: arg.ui` through verbatim, so `ui.options[].label` reached the client as the full multi-language `LocalizedStringArray`. `localizeUiOptions` now resolves it server-side to a single-entry array:

```ts
? option
: { ...option, label: [{ languageCode: ctx.languageCode, value: label }] };
```

Two consequences:

1. **The stamped language can be inaccurate on fallback.** The entry is always stamped with `ctx.languageCode`, even when the resolved value came from the channel default or the English inline fallback — e.g. a `de` request with only an English inline option label returns `[{ languageCode: de, value: 'Automatic' }]`. Harmless for display, but the array no longer tells the truth about what language the string is in.
2. **Client-side language switching no longer re-localizes option labels.** The dashboard consumer (`select-with-options.tsx`) does `label.find(t => t.languageCode === displayLanguage)`. At fetch time `languageCode === displayLanguage`, so it matches and works today. But a cache-then-render-at-a-different-language flow would now fall through to `label[0]` instead of picking the right language from a full array. This is *consistent* with how `description` and arg `label` already resolve to a single string server-side, so it's a narrowing rather than a regression — just want to confirm no admin-ui/dashboard flow relies on receiving all languages for option labels in one fetch. (The mutation-safety e2e covers cross-request isolation well, but not this cache-then-switch angle.)

### Minor

- `if (fromCatalog)` in `localizeString` is a truthiness check, so an intentionally empty-string catalog entry is treated as "no translation" and a catalog can't blank out an inline string. Almost certainly intended — just flagging it's truthiness, not `!== undefined`.
- The PR ships the mechanism with no populated core catalogs (`src/i18n/messages/<lng>.json` aren't in the diff), so every core op still resolves to its inline strings until the tooling is run and committed. Assuming that's the intended "plumbing now, content follow-up" split — worth a line in the description confirming it.
- Doc nit: a few docblocks / the guide say `getTranslationKeys()` is used to "print" the key paths — it *returns* them.
- Test gap: `getTranslationKeys` is covered for a normal op and the no-`defType` case, but not for an arg whose `ui` has no options (non-select) — `getUiOptions` handles it, but there's no direct assertion that such an arg contributes no option keys.

Thanks — nice mechanism, and the "no user-visible change, translations are a follow-up" framing makes it easy to reason about.

michaelbromley:
Good catch, you were right. The dashboard sends `languageCode` from `contentLanguage`, but resolves option labels against `displayLanguage` — two different settings — so narrowing them server-side broke anyone whose two differ. Now the catalog value gets merged into the existing array instead of replacing it, so the other languages survive.

Other three fixed too: the truthiness check has a comment saying it's deliberate, doc nit, and a test for a non-select `ui`.

PR #5085, opened 2026-08-03
fix(core): Prevent exponential re-merge of shared entity references

### LINKED ISSUE #5083: Hydrate regression in 3.7.1 due to mergeDeep changes hangs nodejs
**Describe the bug**

When I modify an order within one of my custom plugins nodejs hangs at 100% and stops responding. It works fine in 3.6.5 but fails in 3.7.1; due to some SMTP plugin weirdness I could not test in 3.7.0.

I've used Kimi K3 to track down and patch locally, having confirmed via node inspector that the stack trace was recursion on mergeDeep fn caused by an entity hydrate in the order path.

The text that follows is generated by Kimi K3 explaining the issue - which makes sense to me but I'm not fully across the vendure code:

# mergeDeep() exponential re-merge: server pins at 100% CPU on order modifications (3.7.x regression)

## Summary

`mergeDeep()` (`packages/core/src/service/helpers/entity-hydrator/merge-deep.ts`) was changed in
**v3.7.0** (PR #4935, commit `aa5859e1c`) to track only the *current recursion path* in its
`visited` WeakSet, instead of every source object seen during one `hydrate()` call. This fixes the
reported case (shared instance merged into every target), but makes the merge **exponential** on any
entity graph containing shared references (diamonds): a shared subgraph is re-merged once per
*path* that reaches it, and path counts multiply with each level of sharing.

On real entity graphs — e.g. an `Order` whose `lines[*].productVariant.channels` all reference the
same Channel instances — `EntityHydrator.hydrate()` effectively never terminates: the Node process
pins at 100% CPU inside `mergeDeep` and the HTTP request never completes. Genuine cycles are still
caught correctly (they are always on the current path), so this is purely a shared-reference
(diamond) blowup, not a cycle bug.

**Affected:** `@vendure/core` ≥ 3.7.0 (verified on 3.7.1; still present on `master` at time of
writing — the `visited.delete(b)` is still there; #4961 and #4986 only fixed adjacent issues).
**Last known good:** 3.6.5.

## Real-world symptoms (observed on 3.7.1)

Any order-modifying Shop API mutation — `addItemToOrder`, `adjustOrderLine`,
`removeItemFromOrder`, `setOrderShippingMethod` — hangs the server at 100% CPU. The price
recalculation transaction commits *before* the hang (the `hydrate()` happens afterwards), so the
modification is visible after a server restart, which makes this look like a "ghost" failure.

Stack sampled from the hung process via the Node inspector (5 samples, all identical):

```
at mergeDeep (merge-deep.js:57:26)
at mergeDeep (merge-deep.js:66:41)
at mergeDeep (merge-deep.js:66:41)   × ~60 frames — depth stays shallow, breadth explodes
...
at hydrate (entity-hydrator.service.js:129)
```

The environment that triggered it: an Order with 3 lines, whose ProductVariants each reference the
same ~31 Channel entities (multi-channel marketplace setup). Every Channel instance is a shared
node under every line — exactly the diamond shape the repro below synthesizes.

## Root cause

v3.6.5 — each source object is merged **once** per `hydrate()` call (linear):

```ts
if (isObject(b)) {
    if (visited.has(b)) {
        return a;
    }
    visited.add(b);   // never removed
}
```

v3.7.0+ — `b` is removed from `visited` on the way out, so the same source object is re-merged for
every distinct path that reaches it:

```ts
let addedToPath = false;
if (isObject(b)) {
    if (visited.has(b)) {
        return a;
    }
    visited.add(b);
    addedToPath = true;
}
// ... recurse ...
if (addedToPath) {
    visited.delete(b);   // <-- shared subgraph will be fully re-merged for every other path
}
```

With `k` levels of "two references to the same child", the leaf subgraph is merged `2^k` times.
Call count follows exactly `2^(depth+1) − 1` (see repro output).

## Minimal reproduction

[repro.js](https://github.com/user-attachments/files/30639110/repro.js) contains both implementations copied verbatim from the compiled
`@vendure/core` dist (only `isObject`/`safeAssign` inlined) and benchmarks them against a
layered-diamond graph (each level references the same child instance twice; **no cycles**):

```js
function makeDiamondGraph(depth) {
    let shared = { id: 'leaf', value: 'x' };
    for (let i = 0; i < depth; i++) {
        shared = { id: `lvl${i}`, left: shared, right: shared, value: 'x' };
    }
    return { id: 'root', a: shared, b: shared };
}
// distinct nodes: depth + 2; distinct paths to leaf: 2^(depth+1)
mergeDeep(makeDiamondGraph(depth), makeDiamondGraph(depth));
```

Measured output (Node 24):

```
mergeDeep regression repro: layered-diamond graph (shared instances, no cycles)

  v3.6.5 depth=10  calls=          12  0.2ms
  v3.7.1 depth=10  calls=        4095  4.3ms

  v3.6.5 depth=14  calls=          16  0.0ms
  v3.7.1 depth=14  calls=       65535  28.3ms

  v3.6.5 depth=18  calls=          20  0.0ms
  v3.7.1 depth=18  calls=     1048575  584.1ms

  v3.6.5 depth=22  calls=          24  0.1ms
  v3.7.1 depth=22  calls=    16777215  12130.2ms

  v3.6.5 depth=24  calls=          26  0.1ms
  v3.7.1 depth=24  calls>20,000,000  10934.5ms  *** EXPLODED: aborted: >20000000 mergeDeep calls
```

v3.6.5: `depth + 2` calls (linear). v3.7.1: `2^(depth+1) − 1` calls (exponential). A hydrated
Order in a multi-channel setup easily exceeds the equivalent of depth 25+, i.e. tens of millions of
re-merges — indistinguishable from a hang.

## Proposed fixes

1. **Revert to the 3.6.5 semantics** (drop `visited.delete(b)`), accepting that #4935's case
   (shared instance merged into every target) regresses. Linear, battle-tested, but throws away the
   #4935 fix.
2. **Memoize merged sources** (preferred): keep the path-scoped cycle detection, but additionally
   record each fully-merged source in a `WeakMap<source, mergedTarget>`; when a source is
   re-encountered off-path, return the memoized result instead of re-merging. This keeps #4935's
   correctness ("every referencing target gets the hydrated content") at linear cost. Note the
   aliasing consideration: targets would share the memoized sub-object reference, which matches how
   the shared *source* instance was already shared before hydration.

Either way, a regression test should cover "shared instance reachable via multiple paths in a
cycle-free graph" with a call-count or time bound, not just correctness.

## Workaround for anyone hitting this

Patch the compiled `node_modules/@vendure/core/dist/service/helpers/entity-hydrator/merge-deep.js`
to remove the `visited.delete(b)` block (restores 3.6.5 behavior).

TheHypnoo:
# Description

Fixes #5083.

`mergeDeep()` re-merges a source object once for every path that reaches it, so hydration cost grows with the number of paths through the entity graph rather than the number of objects in it.

## Cause

Before #4945, `visited` accumulated every source object seen during a `hydrate()` call, so each object was merged once. #4945 changed it to track only the current recursion path (`visited.delete(b)` on the way out) so that an instance shared by several targets is merged into each of them, which is correct — but it also means a shared instance is re-merged once per path that reaches it, and path counts multiply with each level of sharing.

The graphs this runs on are full of shared instances by construction: `EntityHydrator.hydrate()` loads with `relationLoadStrategy: 'query'`, which gives every referencing parent the same instance of a related entity. That is the same property the e2e test added in #4945 relies on (`order.lines[0].productVariant === order.lines[1].productVariant`).

## Fix

Memoise each `(target, source)` pair. Merging a source into a target is idempotent, so a given pair only needs merging once, and the cost goes back to being a function of the graph size. The path-scoped cycle detection and the shared-instance behaviour from #4945 are both unchanged.

A pair is only recorded once it merged in full. The cycle guard skips keys whose source is on the current recursion path, so a pair first reached through a cycle is partial; recording it would freeze it that way and a later route avoiding the cycle would short-circuit on the memo instead of merging the skipped keys. Completeness is tracked through the recursion so an ancestor of a partial merge is not recorded either.

Two alternatives were considered and rejected:

- Reverting to the pre-#4945 semantics fixes the cost but reintroduces #4935.
- Memoising on the source alone (`WeakMap<source, mergedResult>`) is also linear, but aliases the merged sub-object between targets, so a target loses data it already had. The second unit test covers this.

## Measurements

Real data, sqljs e2e: an Order with 3 lines whose ProductVariants reference the same N Channels, each Channel sharing one Zone. `mergeDeep` calls during the second `hydrate()` of the Order — the state a plugin sees after an order mutation leaves one line without the relation.

| Channels | Objects in graph | pre-#4945 | current master | this PR |
| --- | --- | --- | --- | --- |
| 10 | 305 | 167 | 676 | 196 |
| 31 | 515 | 272 | 1810 | 364 |
| 100 | 1205 | 617 | 5536 | 916 |
| 200 | 2205 | 1117 | 10936 | 1716 |
| 500 | 5205 | 2617 | 27136 | 4116 |

Master's cost relative to the pre-#4945 implementation keeps climbing with the amount of sharing (4.0x, 6.7x, 9.0x, 9.8x, 10.4x). This PR settles at ~1.6x and stops growing; that remainder is the work #4945 requires, since merging a shared source into each distinct target is strictly more merging than skipping it.

The e2e seed data only has two levels of sharing, so it shows a large constant factor rather than a hang. The shape that stops completing is covered by the first unit test: on a graph with `depth + 2` objects and `2 ^ (depth + 1)` paths and no cycles, master takes 11.6s at depth 22 and this PR takes under a millisecond.

## Tests

- `should merge each object once rather than once per path` — bounds the cost on a cycle-free graph whose path count is exponential in its depth. Fails by assertion (11.6s) without the fix, not by timeout.
- `should keep the existing data of every target a shared source is merged into` — guards that the memo is keyed on the pair rather than on the source alone.

Verified green: 1139 core unit tests, 25 `entity-hydrator` e2e tests, 197 `order` / `shop-order` e2e tests.

No e2e test is added. With a channel count small enough for CI it would not fail against the regression, so it would add runtime without detecting anything; the exponential shape is what the unit test pins down.

# Breaking changes

None. The exported `mergeDeep()` signature is identical to master — the memo lives in a private recursive function behind it.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5085"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788338410&installation_model_id=20193&pr_number=5085&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5085&signature=1a621ea23fc739fc0e6b55e15104cd0e1e09f8f3d815b5d0c0d042bae8a0635a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


packages/core/src/service/helpers/entity-hydrator/merge-deep.spec.ts
```diff
@@ -103,6 +103,44 @@ describe('mergeDeep()', () => {
...
+        let shared: any = { id: 'leaf', value: 'x' };
+        for (let i = 0; i < 22; i++) {
+            shared = { id: `level-${i}`, left: shared, right: shared, value: 'x' };
+        }
+
+        const start = Date.now();
+        mergeDeep({ id: 'root', a: shared, b: shared }, { id: 'root', a: shared, b: shared });
+
+        // Merging per object is sub-millisecond here; merging per path takes tens of seconds.
+        expect(Date.now() - start).toBeLessThan(2000);
```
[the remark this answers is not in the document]
TheHypnoo: Good catch, applied in 274e0257b.

The second half of the point was the stronger one: the elapsed-time bound only showed the merge was not catastrophically slow, it never asserted that each pair is merged exactly once. Counting enumerations of a shared leaf asserts the actual invariant.

Two adjustments to the proposed diff:

- The graph is now built twice, so the target and the source are independent object graphs rather than one graph merged into itself. That matches what `hydrate()` actually does, where the target is the caller's entity and the source comes from a separate query, and it means the counted leaf is only reachable as a source.
- Only `expect(leafMerges).toBe(1)` is kept as the assertion. The `throw` inside the trap stays because it is what makes a regression fail immediately instead of grinding through the remaining 2 ^ 23 paths first.

Verified in both directions against the pre-fix implementation: it now fails in 5ms with `the shared leaf was merged more than once`, where the timing assertion took 11.6s to fail.

packages/core/src/service/helpers/entity-hydrator/merge-deep.ts
```diff
@@ -12,11 +12,28 @@ export function mergeDeep<T extends { [key: string]: any }>(
...
+        let mergedSources = merged.get(a);
+        if (mergedSources?.has(b)) {
+            return a;
+        }
+        if (!mergedSources) {
+            mergedSources = new WeakSet();
+            merged.set(a, mergedSources);
+        }
+        mergedSources.add(b);
+    }
```
biggamesmallworld: The memo records the pair before the merge runs, so a pair merged partially under the cycle guard is frozen that way for every later route to it.

```ts
const P: any = { id: 'P', tag: 'p' };
const b: any = { id: 'b', tag: 'b', back: P };
P.child = b;
const source = { p: P, direct: { c: b } };

const a: any = { id: 'b', existing: 'yes' };
const target = { p: { id: 'P', child: a }, direct: { c: a } };

mergeDeep(target, source).direct.c.back?.id;
// master:   'P'
// this PR:  undefined
```

First entry into `(a, b)` happens with `P` on the path, so `back` is skipped by the cycle guard below — but the pair is recorded as done, so the cycle-free route through `direct` short-circuits and `back` never merges. `Collection.parent`/`children`, `Zone` and `Facet` all produce this shape once both directions are hydrated.

Fix: only record a pair that merged in full, and propagate incompleteness up. This also gets `merged` out of the public signature, which removes the "can't collapse the two params" problem from the description:

```ts
export function mergeDeep<T extends { [key: string]: any }>(
    a: T | undefined,
    b: T,
    visited: WeakSet<object> = new WeakSet(),
): T {
    return mergeDeepRecursive(a, b, visited, new WeakMap()).value;
}

/**
 * `mergedPairs` records the (target, source) pairs already merged. Merging a source into a target
 * is idempotent, so each pair only needs doing once. Relations load with the 'query' strategy, so
 * every referencing parent gets the same instance; without this the pair is re-merged once per path
 * reaching it, and paths multiply with each level of sharing. See #5083.
 *
 * A pair is only recorded once it merged in full: the cycle guard below skips keys whose source is
 * on the current path, and a pair left partial that way must stay mergeable via a path that does
 * not cross the cycle. `complete` reports that upwards so an ancestor of a partial merge is not
 * recorded either.
 */
function mergeDeepRecursive<T extends { [key: string]: any }>(
    a: T | undefined,
    b: T,
    visited: WeakSet<object>,
    mergedPairs: WeakMap<object, WeakSet<object>>,
): { value: T; complete: boolean } {
    if (!a) {
        return { value: b, complete: true };
    }

    if (mergedPairs.get(a)?.has(b)) {
        return { value: a, complete: true };
    }
```

then in the body:

```ts
    if (isObject(b)) {
        if (visited.has(b)) {
            return { value: a, complete: false };   // was: return a;
        }
```

```ts
    let complete = true;                            // new, above the loop
    for (const [key, value] of Object.entries(b)) {
```

```ts
                if (isObject(value) && visited.has(value)) {
                    complete = false;               // new
                    continue;
                }
```

```ts
                    const child = mergeDeepRecursive(a[key], b[key], visited, mergedPairs);
                    complete &&= child.complete;
                    safeAssign(a, key, child.value);
```

and the tail:

```ts
    if (addedToPath) {
        visited.delete(b);
    }

    if (complete && typeof a === 'object' && typeof b === 'object' && b !== null) {
        let sources = mergedPairs.get(a);
        if (!sources) {
            sources = new WeakSet();
            mergedPairs.set(a, sources);
        }
        sources.add(b);
    }

    return { value: a, complete };
}
```

Skips only happen in cyclic graphs, so the linear cost on the #5083 shape is unchanged — I ran it: the layered-diamond repro still merges the leaf exactly once, and the existing tests pass.

Public signature stays byte-identical to master, so no breaking change. `mergedPairs` reads better than `merged` too — two screens away in the spec, `merged` is the *result* of a merge.

packages/core/src/service/helpers/entity-hydrator/merge-deep.spec.ts
```diff
@@ -103,6 +103,54 @@ describe('mergeDeep()', () => {
...
+        const layeredDiamond = (leaf: object) => {
+            let shared: any = leaf;
+            for (let i = 0; i < 22; i++) {
+                shared = { id: `level-${i}`, left: shared, right: shared, value: 'x' };
+            }
+            return { id: 'root', a: shared, b: shared };
+        };
+        let leafMerges = 0;
+        // mergeDeep() enumerates the source once per merge. Throwing rather than counting up stops
+        // a regression here from working through the other 2 ^ 23 paths to the leaf first.
```
biggamesmallworld: `depth` doesn't exist, and the next comment hardcodes `2 ^ 23`, so the two disagree on the shape.

````suggestion
        const depth = 22;
        // `depth + 2` distinct objects, 2 ^ (depth + 1) distinct paths to the leaf, no cycles.
        const layeredDiamond = (leaf: object) => {
            let shared: any = leaf;
            for (let i = 0; i < depth; i++) {
                shared = { id: `level-${i}`, left: shared, right: shared, value: 'x' };
            }
            return { id: 'root', a: shared, b: shared };
        };
        let leafMerges = 0;
        // mergeDeep() enumerates the source once per merge, so this is one `ownKeys` per merge —
        // changing how the source is enumerated invalidates the count. Throwing rather than
        // counting up stops a regression here from working through the other paths to the leaf first.
````

File: packages/core/src/service/helpers/entity-hydrator/merge-deep.spec.ts [hunk 4]
```diff
@@ -103,6 +103,54 @@ describe('mergeDeep()', () => {
...
+            },
+        );
+
+        mergeDeep(layeredDiamond({ id: 'leaf', value: 'x' }), layeredDiamond(countedLeaf));
+
+        expect(leafMerges).toBe(1);
+    });
+
+    // Rules out memoising on the source alone, which is also linear but aliases the merged
+    // sub-object between the targets.
```
biggamesmallworld: State the invariant — the rejected alternative is already in the PR body, and this comment argues with code that isn't in the tree.

````suggestion
// Each target keeps its own pre-existing data when a shared source is merged into several of them.
````

**Review by biggamesmallworld — CHANGES_REQUESTED** [review]
Verified the fix against the repro in #5083: the layered-diamond graph goes from `2^23 - 1` merges down to 24, exactly matching 3.6.5's `depth + 2`. Also built the real-world shape from the issue (3 order lines -> variants -> the same 31 shared Channels): 40 merges, instant. The hang is genuinely gone, and memoising on the `(target, source)` pair rather than the source alone is the right call, since source-only would re-break #4935.

One blocker below: the memo also introduces silent data loss on self-referential graphs, which #5083 doesn't cover. I implemented and tested the suggested fix locally before posting it — the existing 25 entity-hydrator tests pass, the #5083 repro stays at one leaf merge, and the frozen-pair case is resolved.

One more, not inline because the file isn't in the diff — `entity-hydrator.service.ts:143`:

```ts
for (const prop of propertiesToAdd) {
    (target as any)[prop] = mergeDeep((target as any)[prop], hydrated[prop]);
}
```

Each iteration gets a fresh memo, so two top-level relations reaching the same shared `Channel` still merge it twice. Linear either way, so fine to leave — but worth a line saying it's deliberate, otherwise the next reader tries to hoist it.

packages/core/src/service/helpers/entity-hydrator/merge-deep.ts
TheHypnoo: Confirmed and applied in cb2f9511a. Thanks for catching this, and for the tested patch.

I ran your repro against three implementations before changing anything:

| | `direct.c.back?.id` |
| --- | --- |
| master | `"P"` |
| this PR, before | `undefined` |
| pre-#4945 | `undefined` |

So my "matches pre-#4945 behaviour, not a new limitation" note in the description was measured against the wrong baseline. What 3.6.5 did is not a licence to reintroduce it on top of master. I have dropped that section.

Your patch is in as written, plus a regression test (`should complete a pair whose first merge was cut short by the cycle guard`) built from your repro. It fails against the previous implementation and passes against both this one and master.

One thing worth recording. `complete` propagating to the root switches the memo off for every ancestor, so on a graph that is both cyclic and shared it puts the exponential back:

```
back-edge from leaf to root, layered diamond
  depth  8   1ms
  depth 12  15ms
  depth 16 195ms
  depth 18 770ms      (x4 per +2 depth, same growth as unfixed master)
```

I could not reach that through `hydrate()`. I rebuilt the e2e harness and walked the merged graph looking for an object reachable from itself, including the shapes you named:

```
channels.defaultTaxZone/defaultShippingZone.members   nodes=5205  merges=4116  cycle=none
collections.parent.children                           nodes=79    merges=54    cycle=none
facetValues.facet.values                              nodes=167   merges=54    cycle=none
product.variants.product                              nodes=5368  merges=4116  cycle=none
```

With the 'query' strategy TypeORM hands back a fresh instance per parent rather than the same one, so `product.variants[i].product` is a different object from the parent and there is no back-edge to skip on. `merges=4116` at 500 channels is identical to the previous implementation, so the completeness check costs nothing on real hydration.

If a source graph ever does come back cyclic, the memo could carry the set of path nodes that caused the skip and be reused only while those are still on the path, which keeps both properties. That is more machinery than this fix needs, and I have no case that reaches it, so I have left it out. Happy to open a follow-up issue if you would rather have it written down.

TheHypnoo:
Comment added at `entity-hydrator.service.ts:143` in cb2f9511a:

```ts
// Each call starts its own memo, so an entity shared by two top-level relations is
// merged once per relation. Deliberate: bounded by the relation count. See #5083.
```

Hoisting it is also awkward now that `mergedPairs` is behind the private recursive function, which seemed like the right place for it to stay.

packages/core/src/service/helpers/entity-hydrator/merge-deep.spec.ts
TheHypnoo: Applied as suggested in cb2f9511a. The two really did disagree, `2 ^ 23` was left over from an earlier version of the test.

packages/core/src/service/helpers/entity-hydrator/merge-deep.spec.ts
TheHypnoo: Applied as suggested in cb2f9511a.

packages/core/src/service/helpers/entity-hydrator/merge-deep.ts
```diff
@@ -13,8 +13,32 @@ export function mergeDeep<T extends { [key: string]: any }>(
...
+    visited: WeakSet<object>,
+    mergedPairs: WeakMap<object, WeakSet<object>>,
+): { value: T; complete: boolean } {
     if (!a) {
-        return b;
+        return { value: b, complete: true };
+    }
+
+    if (mergedPairs.get(a)?.has(b)) {
+        return { value: a, complete: true };
```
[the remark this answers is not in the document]
TheHypnoo: I ran your probe against three implementations before deciding.

| | `merged.last.value` | `first === middle === last` |
| --- | --- | --- |
| this PR | `B` | true |
| master | `A` | true |
| pre-#4945 | `B` | true |

Three things follow from that.

The three properties are the same object in every implementation, so there is no state in which they hold different values. `sharedTarget` gets `'A'` written, then `'B'`, and whichever write lands last is what all three properties read. Master's `'A'` is not a correct answer that this PR loses, it is just master's last write. Reorder the source keys to `first, last, middle` and master returns `'B'` too.

The case where the memo could actually drop data is disjoint keys, and it does not. Merging `{x: 1}` and `{y: 2}` into the same aliased target gives `x=1 y=2` on this PR, on master and on 3.6.5. Skipping the repeat of `(sharedTarget, sourceA)` skips work that has already been done, so nothing is lost. Only a second source writing the *same* key changes the outcome, and that outcome is order dependent everywhere.

The shape also cannot come out of `hydrate()`. It needs `a.first === a.middle` while `b.first !== b.middle`. Both graphs are TypeORM results for the same entity loaded with `relationLoadStrategy: 'query'`, which hands every referencing parent the same instance for a given row, so if two properties of the target are one instance then the matching properties of the source are one instance as well.

On the suggested remedy: invalidating memo entries when the target or a merged descendant changes would invalidate on every merge, since merging always mutates the target. That is the memo removed rather than corrected, and #5083 comes back with it. The identity-aligned-graph alternative is what `hydrate()` already provides, and the `mergeDeepRecursive` docblock now states it.

Leaving this as is unless someone has a case that reaches it from a real hydration.

biggamesmallworld (approved):
Blocker fixed in cb2f9511a, comment nits in f243a43c. Approving.

The new cycle-guard test is a real guard: it fails at 274e0257b with `Cannot read properties of undefined (reading 'id')` and passes here. Full core unit suite green (1140 tests).

## On the remaining limitation

Incompleteness taints every ancestor, so a cycle below a shared node disables the memo along that spine. Synthetic layered diamond with a leaf back-ref to the root, depth 18: master 1212ms, this PR 580ms — still exponential.

Unreachable through `hydrate()` in practice. I instrumented `mergeDeep` with counters and ran real hydrate calls: hydrate once to populate, drop one line's relation, hydrate again (the #5083 shape — a single hydrate on a bare target barely merges at all). Three order lines, relation specs picked to name both directions of a relation, since a cycle is what the limitation needs. Merges during the second hydrate:

| relations | master | PR | cycle skips | tainted pairs | cycles |
| --- | --- | --- | --- | --- | --- |
| `lines.productVariant.product.variants` | 123 | 123 | 0 | 0 | none |
| `…product.variants.product` | 165 | 141 | 0 | 0 | none |
| `…product.variants.product.variants.product` | 531 | 227 | 0 | 0 | none |
| `lines.productVariant.collections.productVariants` | 43 | 43 | 0 | 0 | none |
| `…product.facetValues.facet.values` | 285 | 124 | 0 | 0 | none |
| `lines.productVariant.{product,channels,collections}` | 69 | 65 | 0 | 0 | none |
| `lines.productVariant.channels` + `product.variants.channels` | 171 | 147 | 0 | 0 | none |
| `Collection` `parent.children.parent` + `children.parent.children` | 9 | 9 | 0 | 0 | none |

No hydrated graph contains a cycle: `relationLoadStrategy: 'query'` builds a fresh instance per relation load, so `product.variants[i]` is never the same object as `lines[j].productVariant`. Sharing is within a load, not across levels. Guard never fires. No test needed for it — the trigger can't be built from a real hydrate, and the cyclic path's correctness is already covered by the two tests in the diff.

Merge counts lower or equal everywhere, never higher, and conservative (this PR counts memo hits as merges). Correctness fix, no measured perf cost on any shape.

PR #5087, opened 2026-08-03
fix(core): Invalidate stock location channel id cache correctly

### LINKED ISSUE #3324: MultiChannelStockLocationStrategy Fails to Clear Cached Channel IDs on StockLocationEvent
**Describe the bug**
`MultiChannelStockLocationStrategy` is not clearing cached channel ids on `StockLocationEvent`

**Expected behavior**
`MultiChannelStockLocationStrategy` should clear cached channel ids on `StockLocationEvent`

 - @vendure/core version: 3.1.1
 - Nodejs version 22
 - Database (mysql/postgres etc): sqlite, mysql

**Additional context**
~~[This](https://github.com/vendure-ecommerce/vendure/blob/4c6773ec7a8357407caaff71484db11d8e7c1c81/packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts#L147) should be~~
~~const channelIds = await this.channelIdCache.get(this.getCacheKey(stockLevel.stockLocationId),~~
~~since we are trying to clear cached channel ids by using `this.getCacheKey(stockLocationId)` in [here](https://github.com/vendure-ecommerce/vendure/blob/4c6773ec7a8357407caaff71484db11d8e7c1c81/packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts#L66)~~
[This](https://github.com/vendure-ecommerce/vendure/blob/e22c60171aa8bb33403c0cf389a07fdd322b17f5/packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts#L66) should be
```ts
.subscribe(({ entity }) => this.channelIdCache.delete(entity.id));
```
Since [`Cache.delete`](https://github.com/vendure-ecommerce/vendure/blob/e22c60171aa8bb33403c0cf389a07fdd322b17f5/packages/core/src/cache/cache.ts#L101) expects `id` rather than the whole cache key.

ryandiginomad:
# Description

Fixes #3324.

`MultiChannelStockLocationStrategy` caches each StockLocation's channel ids for 7 days, but the cache was never actually invalidated, for two independent reasons:

**1. The delete used a double-prefixed key** (the defect reported in the issue). The `StockLocationEvent` subscription passed `this.getCacheKey(entity.id)` into `Cache.delete()`, but `Cache.delete()` applies the configured `getKey` function itself — the same way `Cache.get()` does — so the delete targeted `MultiChannelStockLocationStrategy:StockLocationChannelIds:MultiChannelStockLocationStrategy:StockLocationChannelIds:<id>`, which never exists, and the real entry lived out its full TTL. The fix is the one-liner proposed in the issue: pass the raw id. (I checked the rest of core for the same pattern — the other `Cache` consumers, e.g. `customer-group-condition.ts` and `facet-value-checker.ts`, already pass raw ids; this was the only occurrence.)

**2. The operation that changes channel membership never emitted an invalidating event at all** — surfaced by the follow-up investigation in [this comment](https://github.com/vendurehq/vendure/issues/3324#issuecomment-5134242455). `assignStockLocationsToChannel()` / `removeStockLocationsFromChannel()` delegate to `ChannelService.assignToChannels()` / `removeFromChannels()`, which emit `ChangeChannelEvent`, not `StockLocationEvent`. So even with the key fix, assigning a StockLocation to a new Channel leaves that channel reading the stale pre-assignment channel list: in a multi-channel setup, every newly added channel can show all products as `OUT_OF_STOCK` for up to 7 days (or until a server restart clears the in-memory cache).

The fix adds a second subscription in the strategy's `init()`, on `ChangeChannelEvent` filtered to `entityType === StockLocation`. Subscribing to the event — rather than patching the two service methods — keeps the invalidation inside the strategy that owns the cache, and covers every path that changes a StockLocation's channels, including plugins calling `ChannelService` directly.

Both changes live in the same `init()` block and are only meaningful together (the first makes deletes actually work, the second makes the right operation trigger them), so they ship as one PR.

Both subscribers now route the delete through a small helper that catches and logs a rejecting cache backend. Previously the returned promise was dropped, so a failed invalidation would have been an unhandled rejection with no log line, leaving the stale list in place for its full TTL.

# Breaking changes

None. The only behavioural change is that saleable-stock reads now see channel-membership changes immediately, instead of after cache expiry or a server restart.

# Testing

**Unit** (`multi-channel-stock-location-strategy.spec.ts`, new file, 7 cases, no DB): uses the real `Cache` wrapper over an in-memory Map — so the key-prefixing interaction under test behaves exactly as in production — plus a Subject-backed EventBus stub. On master, four cases fail: the three invalidation cases (update / assign / remove: the first because the delete misses its key, the other two because no subscription exists) plus the failure-logging case below. With the fix all 7 pass. The other three pin behaviour that should not change: channel ids are cached after the first lookup, `created` events don't invalidate, and `ChangeChannelEvent`s for other entity types don't invalidate.

**E2E** (`stock-location.e2e-spec.ts`, new describe, 2 cases): both reproduce the sequence from the issue thread. The first queries a variant's saleable stock via the Shop API in a new channel's context, which caches the location's pre-assignment channel list, then calls `assignStockLocationsToChannel` and queries again; on master's dist it reports `OUT_OF_STOCK` where `IN_STOCK` is expected. The second covers the removal direction, where a stale positive list lets a location that was removed from the channel keep contributing saleable stock: it warms a positive read, removes the location, then polls until the Shop query reports `OUT_OF_STOCK`, which on master's dist never happens. Both pass with the fix, and the 16 existing tests in the file are unaffected in both runs.

What I ran:

| | Result |
|---|---|
| New unit spec against master code | 4 failed / 3 passed |
| New unit spec after the fix | 7 passed |
| Full core unit suite | 76 files, 1144 passed |
| `stock-location.e2e-spec.ts` against master dist | 2 failed (both new cases) / 16 passed |
| `stock-location.e2e-spec.ts` against rebuilt dist | 18 passed |
| `stock-control.e2e-spec.ts` + `stock-control-multi-location.e2e-spec.ts` | 46 passed |

The pre-existing type error in `order.service.ts:1460` that stops `npm run build` locally is unrelated (file untouched, reproduces on clean master).

### 🤖 AI assistance

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5087"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788392762&installation_model_id=20193&pr_number=5087&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5087&signature=8aab0d08de8c48f97493929d84145e809cc744c92f4ffccb67d77c52263d2b2d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/stock-location.e2e-spec.ts
+++ b/packages/core/e2e/stock-location.e2e-spec.ts
@@ -32,7 +32,11 @@ import {
     createProductDocument,
     createProductVariantsDocument,
 } from './graphql/shared-definitions';
-import { localUpdatedOrderFragment, testOrderFragment } from './graphql/shop-definitions';
+import {
+    getProductWithStockLevelDocument,
+    localUpdatedOrderFragment,
+    testOrderFragment,
+} from './graphql/shop-definitions';
 
 describe('Stock location', () => {
     const defaultStockLocationId = 'T_1';
@@ -390,15 +394,12 @@ describe('Stock location', () => {
                 }
                 // Create a StockLocation in Channel A
                 adminClient.setChannelToken(CHANNEL_A_TOKEN);
-                const { createStockLocation } = await adminClient.query(
-                    testCreateStockLocationDocument,
-                    {
-                        input: {
-                            name: 'Channel-A Location',
-                            description: 'Belongs to Channel A only',
-                        },
+                const { createStockLocation } = await adminClient.query(testCreateStockLocationDocument, {
+                    input: {
+                        name: 'Channel-A Location',
+                        description: 'Belongs to Channel A only',
                     },
-                );
+                });
                 targetLocationId = createStockLocation.id;
             });
 
@@ -421,14 +422,86 @@ describe('Stock location', () => {
 
                 // Verify the original entity in Channel A is completely unchanged.
                 adminClient.setChannelToken(CHANNEL_A_TOKEN);
-                const { stockLocations } = await adminClient.query(
-                    testGetStockLocationsListDocument,
-                );
-                const target = stockLocations.items.find(
-                    (sl: any) => sl.id === targetLocationId,
-                );
+                const { stockLocations } = await adminClient.query(testGetStockLocationsListDocument);
+                const target = stockLocations.items.find((sl: any) => sl.id === targetLocationId);
                 expect(target?.name).toBe('Channel-A Location');
                 expect(target?.description).toBe('Belongs to Channel A only');
+            });
+        });
+
+        // Repro for #3324: the MultiChannelStockLocationStrategy caches each StockLocation's
+        // channel ids for 7 days. Assigning a StockLocation to a Channel must invalidate that
+        // cache entry, otherwise the new channel sees no saleable stock until the entry expires
+        // or the server restarts.
+        describe('channelId cache invalidation (#3324)', () => {
+            const CACHE_TEST_CHANNEL_TOKEN = 'stock-loc-cache-test';
+            let cacheTestChannelId: string;
+            let cacheTestLocationId: string;
+
+            beforeAll(async () => {
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                const { createStockLocation } = await adminClient.query(testCreateStockLocationDocument, {
+                    input: {
+                        name: 'Cache test location',
+                    },
+                });
+                cacheTestLocationId = createStockLocation.id;
+                await adminClient.query(testSetStockLevelInLocationDocument, {
+                    input: {
+                        id: 'T_1',
+                        stockLevels: [
+                            {
+                                stockLocationId: cacheTestLocationId,
+                                stockOnHand: 10,
+                            },
+                        ],
+                    },
+                });
+                const { createChannel } = await adminClient.query(createChannelDocument, {
+                    input: {
+                        code: 'cache-test-channel',
+                        token: CACHE_TEST_CHANNEL_TOKEN,
+                        defaultLanguageCode: LanguageCode.en,
+                        currencyCode: CurrencyCode.GBP,
+                        pricesIncludeTax: true,
+                        defaultShippingZoneId: 'T_1',
+                        defaultTaxZoneId: 'T_1',
+                    },
+                });
+                channelGuard.assertSuccess(createChannel);
+                cacheTestChannelId = createChannel.id;
+                await adminClient.query(assignProductToChannelDocument, {
+                    input: {
+                        channelId: cacheTestChannelId,
+                        productIds: ['T_1'],
+                    },
+                });
+            });
+
+            afterAll(() => {
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                shopClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+            });
+
+            it('saleable stock becomes visible once the StockLocation is assigned to the channel', async () => {
+                // Populate the strategy's channelId cache for the location while it is
+                // not yet assigned to the new channel
+                shopClient.setChannelToken(CACHE_TEST_CHANNEL_TOKEN);
+                const before = await shopClient.query(getProductWithStockLevelDocument, { id: 'T_1' });
+                expect(before.product?.variants[0].stockLevel).toBe('OUT_OF_STOCK');
+
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                await adminClient.query(testAssignStockLocationToChannelDocument, {
+                    input: {
+                        stockLocationIds: [cacheTestLocationId],
+                        channelId: cacheTestChannelId,
+                    },
+                });
+                // Wait a bit for the ChangeChannelEvent subscriber to invalidate the cache
+                await new Promise(resolve => setTimeout(resolve, 100));
+
+                const after = await shopClient.query(getProductWithStockLevelDocument, { id: 'T_1' });
+                expect(after.product?.variants[0].stockLevel).toBe('IN_STOCK');
             });
         });
     });
--- a/packages/core/e2e/stock-location.e2e-spec.ts
+++ b/packages/core/e2e/stock-location.e2e-spec.ts
@@ -435,8 +435,27 @@ describe('Stock location', () => {
         // or the server restarts.
         describe('channelId cache invalidation (#3324)', () => {
             const CACHE_TEST_CHANNEL_TOKEN = 'stock-loc-cache-test';
+            const CACHE_REMOVAL_CHANNEL_TOKEN = 'stock-loc-cache-removal';
             let cacheTestChannelId: string;
             let cacheTestLocationId: string;
+
+            // The cache delete runs in an async event subscriber, so the shop query is
+            // polled with a deadline rather than assuming a fixed delay. Returns the
+            // last observed value so a timeout produces a clear assertion failure.
+            async function pollShopStockLevel(expected: string, timeoutMs = 10_000): Promise<string> {
+                const deadline = Date.now() + timeoutMs;
+                let lastSeen = '';
+                for (;;) {
+                    const { product } = await shopClient.query(getProductWithStockLevelDocument, {
+                        id: 'T_1',
+                    });
+                    lastSeen = product?.variants[0].stockLevel ?? '';
+                    if (lastSeen === expected || Date.now() > deadline) {
+                        return lastSeen;
+                    }
+                    await new Promise(resolve => setTimeout(resolve, 50));
+                }
+            }
 
             beforeAll(async () => {
                 adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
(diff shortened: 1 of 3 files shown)


packages/core/e2e/stock-location.e2e-spec.ts
```diff
@@ -421,15 +422,87 @@ describe('Stock location', () => {
...
+                    },
+                });
+            });
+
+            afterAll(() => {
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                shopClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+            });
+
+            it('saleable stock becomes visible once the StockLocation is assigned to the channel', async () => {
```
Refaat-alktifan, line 505: This proves the assignment direction only. I think the removal direction deserves an equivalent e2e test, since stale positive membership can keep a removed location contributing stock in the channel it was removed from, which is the higher risk case. The unit spec constructs a removed event, but it does not drive the real removeStockLocationsFromChannel path through the Shop projection.

A sibling case could warm IN_STOCK in the channel, remove the location, assert the relation is gone while the source stock stays positive, then poll the Shop query with a deadline until it returns OUT_OF_STOCK.

Refaat-alktifan:
Hi Ryan, I independently investigated #3324 before I saw this PR and arrived at the same production design: pass the raw StockLocation ID to Cache.delete() and invalidate on ChangeChannelEvent only when entityType === StockLocation.
My warm-cache assignment and removal reproductions also failed on master and passed with this repair, so the production direction looks right to me. Using the real Cache wrapper in the unit spec is a particularly good choice, since it proves the key transformation itself.
I left one substantive test suggestion inline. The public E2E covers assignment, while removal is currently exercised only by constructing the event in the unit test. A warm-positive removal E2E would prove the real service path and Shop projection in the higher risk stale-positive direction.

CodeRabbit has already covered bounded polling, so I have not repeated that. I am happy to share the isolated removal-test shape from my local reproduction if it helps.

packages/core/e2e/stock-location.e2e-spec.ts
```diff
@@ -421,15 +422,87 @@ describe('Stock location', () => {
...
+
+                adminClient.setChannelToken(E2E_DEFAULT_CHANNEL_TOKEN);
+                await adminClient.query(testAssignStockLocationToChannelDocument, {
+                    input: {
+                        stockLocationIds: [cacheTestLocationId],
+                        channelId: cacheTestChannelId,
+                    },
+                });
+                // Wait a bit for the ChangeChannelEvent subscriber to invalidate the cache
+                await new Promise(resolve => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
ryandiginomad: Adopted in ad3e83c48: the fixed delay is replaced by a poll helper with a 10s deadline that returns the last observed value, so a timeout fails the assertion showing the actual stock level. Both the assignment test and the new removal test use it.

packages/core/e2e/stock-location.e2e-spec.ts
ryandiginomad: Added in ad3e83c48. The test warms IN_STOCK in a fresh channel, removes the location, checks the location's own stockOnHand is untouched, then polls until OUT_OF_STOCK through the real removeStockLocationsFromChannel path.

One ordering detail that turned out to matter: the StockLevel row is created only after the channel assignment. Any admin response that resolves variant stock in between (assignProductsToChannel does, via the ProductVariant.stockOnHand field) runs the strategy and seeds the cache with the pre-assignment list, and on master that makes the warm-positive state unreachable. With the ordering right, I verified the test against master's dist: it fails exactly at the stale-positive point (poll stuck at IN_STOCK after removal) and passes on this branch.

Thanks for running an independent reproduction and for pushing on this direction. You were right that it is the higher-risk case.

ryandiginomad:
@Refaat-alktifan Thanks for verifying this independently against master, and for the review. Both suggestions are in ad3e83c48: bounded polling replaces the fixed delay, and the removal direction now has its own e2e that reaches OUT_OF_STOCK through the real service path. Details on the test shape and one ordering subtlety are in the inline thread.

biggamesmallworld (commented):
**This change is correct and I have verified it. Nothing below blocks merge** - the two points are nits, and the third is a follow-up for us.

The diagnosis is right, and the two-part split is the useful part of it: the `ChangeChannelEvent` half is what actually resolves #3324, since the double-prefixed delete key alone would not have produced the reported symptom. Invalidating from inside the strategy that owns the cache, rather than patching `StockLocationService`, is the correct seam - it also covers plugins that call `ChannelService` directly.

How I checked the tests are real: I reverted the strategy source to the master version and the three regression tests failed while the three control tests passed; reverting *only* the `getCacheKey` double-prefix fix failed exactly one. So the two defects are independently covered rather than one test standing in for both. Same result at the e2e level against a master-patched `dist`.

---

**Nit 1, non-blocking: the verification table in the description does not match this diff.** It reports 17 e2e tests in `stock-location.e2e-spec.ts` both before and after. `origin/master` has 16 and this branch has 18 (`git grep -c '^\s*it(' <ref> -- packages/core/e2e/stock-location.e2e-spec.ts`), so that run predates the removal test. Worth re-running and updating the numbers, or dropping the table. Description text only, no code impact.

**Nit 2, non-blocking: both subscribers drop the promise from `Cache.delete()`** - see the two inline comments, each with a ready-to-apply suggestion. This only bites with a cache backend that can reject, and the pattern already ships on the first subscriber on master, so merging as-is introduces no new failure mode. Nice to clean up while the file is open. If you take the suggestions, add one import:

```ts
import { Logger } from '../logger/vendure-logger';
```

---

**Follow-up on our side, nothing for you to do here.** The `Cache` API is what invites this class of bug: `createCache()` applies the configured `getKey` inside `get`, `set` and `delete`, so all three take a raw id. A class that also keeps its own `getCacheKey()` helper to build that config then has an ambiguous key-builder in scope at every call site, and passing it to `delete()` typechecks, double-prefixes and silently no-ops forever - which is exactly the bug you found. `RoleService` (`role.service.ts:77`) has the same shape today and is harmless only because its `getKey` ignores its argument; it becomes a live bug the day that key goes id-dependent. I will open an issue against core to fix the API, and this PR should not grow to cover it.

packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts
```diff
@@ -59,11 +59,21 @@ export class MultiChannelStockLocationStrategy extends BaseStockLocationStrategy
...
 
-        // When a StockLocation is updated, we need to invalidate the cache
+        // When a StockLocation is updated, we need to invalidate the cache.
+        // Note: `Cache.delete()` applies the configured `getKey` function itself,
+        // so it must be passed the raw id, not the result of `getCacheKey()`.
         this.eventBus
             .ofType(StockLocationEvent)
             .pipe(filter(event => event.type !== 'created'))
-            .subscribe(({ entity }) => this.channelIdCache.delete(this.getCacheKey(entity.id)));
+            .subscribe(({ entity }) => this.channelIdCache.delete(entity.id));
```
biggamesmallworld: Non-blocking nit. `Cache.delete()` returns a promise that is neither awaited nor caught. If the cache backend rejects (a Redis blip, say), this is an unhandled rejection with no log line and the invalidation is silently lost - the stale channel-id list then lives out its full 7-day TTL, which is the exact symptom of #3324.

```suggestion
            .subscribe(({ entity }) => {
                void this.channelIdCache
                    .delete(entity.id)
                    .catch(err =>
                        Logger.error(
                            `Failed to invalidate StockLocation channel id cache for id ${entity.id}: ${
                                err instanceof Error ? err.message : String(err)
                            }`,
                            'MultiChannelStockLocationStrategy',
                        ),
                    );
            });
```

Needs `import { Logger } from '../logger/vendure-logger';` at the top of the file.

packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts
```diff
@@ -59,11 +59,21 @@ export class MultiChannelStockLocationStrategy extends BaseStockLocationStrategy
...
+
+        // Assigning a StockLocation to a Channel (or removing it) does not emit a
+        // StockLocationEvent, so we also need to invalidate the cache on ChangeChannelEvents
+        // which relate to StockLocations.
+        this.eventBus
+            .ofType(ChangeChannelEvent)
+            .pipe(filter(event => event.entityType === StockLocation))
+            .subscribe(({ entity }) => this.channelIdCache.delete(entity.id));
```
biggamesmallworld: Non-blocking nit. Same floating promise here on the new subscriber.

```suggestion
            .subscribe(({ entity }) => {
                void this.channelIdCache
                    .delete(entity.id)
                    .catch(err =>
                        Logger.error(
                            `Failed to invalidate StockLocation channel id cache for id ${entity.id}: ${
                                err instanceof Error ? err.message : String(err)
                            }`,
                            'MultiChannelStockLocationStrategy',
                        ),
                    );
            });
```

biggamesmallworld:
Happy to approve after you take a look at these nits! @ryandiginomad 🙏

packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts
ryandiginomad: Adopted in 15401dad. Rather than repeat the block on both subscribers, the delete now goes through a private `invalidateChannelIdCache()` helper, and the context string is a `loggerCtx` const to match `assert-public-url.ts`.

Both of these lines were `@typescript-eslint/no-misused-promises` warnings on the branch before the change; the file is now down to the one pre-existing `no-floating-promises` error on `super.init()` that master has too.

packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts
ryandiginomad: Covered by the same helper in 15401dad, so both subscribers get the catch and the log line.

226c62c5 adds a unit case for the path: with the backing store rejecting the delete, the failure has to reach `Logger.error`. Removing the `.catch` fails that case and no other.

ryandiginomad:
@biggamesmallworld Thanks for reverting the strategy and checking that each defect is independently covered. Testing that the two halves fail separately is not something I had verified directly, and it is the right check on a two-part fix.

Both nits are in, across 15401dad and 226c62c5.

**Nit 1.** You are right. The table was written when the PR opened and never updated after the removal case was added in the first review round. Re-ran everything against the current branch:

| | Result |
|---|---|
| New unit spec against master code | 4 failed / 3 passed |
| New unit spec after the fix | 7 passed |
| Full core unit suite | 76 files, 1144 passed |
| `stock-location.e2e-spec.ts` against master dist | 2 failed (both new cases) / 16 passed |
| `stock-location.e2e-spec.ts` against rebuilt dist | 18 passed |
| `stock-control.e2e-spec.ts` + `stock-control-multi-location.e2e-spec.ts` | 46 passed |

The description now carries these numbers.

**Nit 2.** Both subscribers route through one `invalidateChannelIdCache()` helper instead of repeating the block, with the context string as a `loggerCtx` const to match `assert-public-url.ts`. I added a unit case for it, since otherwise nothing would fail if the catch handler were dropped later; removing the `.catch` fails that case and no other.

Writing it surfaced something in the spec that is worth calling out, since it is in the diff: `init()` subscribes and nothing tears the subscription down, so each test was firing its event into every strategy built before it. Harmless while every assertion was on cache state, but it made the log-call count depend on test ordering. The event stream is now replaced per test.

**Follow-up.** Agreed, and `RoleService` does have the same shape today. Happy to leave it out of this PR and pick it up against the issue once you open it.

biggamesmallworld (approved):
Both nits are addressed and I have re-verified at `226c62c5`. Nothing outstanding from my side.

**Nit 1, the description table.** Updated, and the new numbers match what I see: the unit spec is 7 cases at this head, and the e2e file is 18.

**Nit 2, the floating promises.** The `invalidateChannelIdCache()` helper is the better shape than what I suggested - one place to change instead of two copies of the same block, and the `loggerCtx` const matches the rest of core. I re-ran your mutation check rather than take it on trust: replacing the helper body with a bare `void this.channelIdCache.delete(stockLocationId)` fails exactly one case, the new `logs rather than throws` one, and leaves the other six green. So the catch handler is genuinely pinned and cannot be dropped later without a test going red.

The shared `Subject` you found in the spec is a real fix, not incidental cleanup. `init()` subscribes with no teardown, so a module-level stream fans every event out to every strategy built so far in the file, and any future assertion on call counts rather than cache state would have been order-dependent. Rebuilding the stream per test is the right call.

Gates I ran at this head: `tsc -p build/tsconfig.build.json --noEmit` clean, the unit spec 7/7, plus the mutation check above. CI is green on all four e2e databases; the only red job is `deploy`, which cannot pass from a fork because `VERCEL_TOKEN` is not exposed to fork runs, and is unrelated to this change.

One last step on my side before approval, and it is on me, not you: this is labelled `T2: Elevated risk`, so I want one manual run through the assign-then-read path against a live server rather than approving on the test suite alone. I will do that and come back.

The `Cache` API follow-up is mine to open - I will link it here once it exists, and it stays out of this PR.

PR #5090, opened 2026-08-04
fix(dashboard): keep compiled files inside outputPath for upward imports

### LINKED ISSUE #5086: Dashboard config compiler emits compiled files outside outputPath when the config imports from above its directory
**Describe the bug**

Follow-up to #4979 and #4913, as offered in
https://github.com/vendurehq/vendure/issues/4979#issuecomment-5105508820
(my own comment there — filing it separately so the point is not lost now
that #4979 is closed).
Verified against master (`611658f`) by running the real compiler code on
Node 24.18.0 / Windows 11 (platform-independent mechanism).

In `compileTypeScript` (`packages/dashboard/vite/utils/compiler.ts`):

```ts
const sourceRoot = customSourceRoot ?? path.dirname(inputPath);
// ...
const relativePath = path.relative(sourceRoot, filePath);
const outputFilePath = path.join(outputPath, relativePath).replace(/\.tsx?$/, '.js');
```

`sourceRoot` defaults to the config's directory, so any collected file above
it (e.g. `../shared/util.ts`) yields a relative path with a leading `..` and
is emitted outside `outputPath`, past the compiler-written
`{"type": "commonjs"}` guard (`writePackageJson`). Verified on disk: for a
config at `<root>/src/vendure-config.ts` importing `../shared/util.js`, the
emitted `util.js` lands at `<outputPath>/../shared/util.js`.

**Impact**

1. Custom `tempCompilationDir` outside `node_modules`. With a config that
   has an upward local import and an enclosing `"type": "module"` scope
   (e.g. the consuming project's root `package.json`), the identical error
   from #4979 returns:

   ```
   ReferenceError: exports is not defined in ES module scope
   ... '<project>/package.json' contains "type": "module".
   ```

   Reproduced with `tempCompilationDir: <project>/.vendure-temp` in a
   project whose root `package.json` sets `"type": "module"`.

2. Escaped files are written outside the per-instance `outputPath`, and
   `compileInternal` only removes `outputPath`, so they persist across
   builds in the parent `compiler/` directory.

Note on the default location: the #4913 default
(`<project>/node_modules/.cache/vendure-dashboard-temp`) is protected even
in `"type": "module"` projects, because Node's package-scope lookup stops at
the `node_modules` boundary — verified directly: the same escaped file loads
correctly under `node_modules/.cache` and fails with the ReferenceError when
placed outside `node_modules` in the same project. So exposure is limited to
custom `tempCompilationDir` values — which is also why the constraint now
documented on that option exists in the first place. Fixing the emit would
remove the need for it.

**To Reproduce**

1. In a project whose root `package.json` sets `"type": "module"`, create a
   config at `src/vendure-config.ts` importing a local module from above its
   directory (e.g. `import { appName } from '../shared/util.js'`).
2. Build the React Dashboard with
   `tempCompilationDir: path.join(process.cwd(), '.vendure-temp')`.
3. Config loading fails with the `ReferenceError` above.

**Expected behavior**

All files collected during compilation are emitted inside `outputPath`,
where the `{"type": "commonjs"}` guard applies, regardless of where the
imported sources live relative to the config.

**Suggested direction**

Make the default `sourceRoot` the common ancestor of all collected source
files, so no emit can escape `outputPath`. Constraints:

- The default `getCompiledConfigPath` joins `outputPath` with the config
  basename, assuming the config lands flat at the output root. Once
  `sourceRoot` moves, the compiled config path needs to be derived from the
  actual emit layout.
- `configFileName` should remain a basename — custom `PathAdapter`
  implementations rely on that contract.
- An explicit `pathAdapter.sourceRoot` (monorepo setups) should keep its
  current behavior.

Happy to pick this up.

TheStreamCode:
# Description

Fixes #5086.

`compileTypeScript` defaulted its source root to the config file's own directory, so a config importing from above it — `../shared/util.js` — produced a relative path with a leading `..` and the compiled file was written outside `outputPath`. That places it past the `package.json` the compiler writes to declare the output's module type, so Node resolves it against the enclosing project instead and fails with `exports is not defined in ES module scope`.

The approach here was proposed by @Refaat-alktifan in #5086: derive the source root from the nearest common ancestor of the collected source graph, so parent imports keep their relative topology and stay inside `outputPath`.

Three changes:

- **Source root.** Defaults to the deepest directory containing the config and every file it imports. The config's own directory takes part in that calculation, so a project whose imports all sit alongside or below the config gets exactly the previous root and an unchanged layout. It only widens where the old default would have emitted outside `outputPath`. An explicit `pathAdapter.sourceRoot` still wins, and a widened default is announced with a warning log.
- **Compiled config path.** Unchanged for custom adapters: `getCompiledConfigPath` keeps deciding where the config is imported from. When no adapter supplies it, the default follows the emit — the config's path relative to the resolved source root, which is the bare filename unless the root widened.
- **Guard.** Emitting outside `outputPath` now throws at compile time before anything is written, instead of surfacing later as an unrelated module-scope error.

Tests cover the upward import, a downward-import control asserting the layout does not change, an explicit `sourceRoot`, a custom compiled-config path beside a widened root, the multi-root error, and a filename beginning with dots (which must not be mistaken for an escape). Fixtures place an ESM `package.json` above `outputPath`, so on `master` the escaped emit fails to load with the reported `exports is not defined in ES module scope`; the leading-dot case loads fine there and is covered by layout assertions.

# Breaking changes

One behavioural addition to call out in the release notes: a compile that would emit outside `outputPath` now throws instead of writing anyway. Projects that never escape are unaffected, and custom `getCompiledConfigPath` adapters keep their contract — but a CommonJS setup with a deliberately narrow `pathAdapter.sourceRoot` that relied on the escaped file still resolving will now need to widen that root. Also note the default root can widen above the config directory (announced with a warning log), which moves emitted files deeper; tsconfig `baseUrl` is remapped accordingly.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5090"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788448833&installation_model_id=20193&pr_number=5090&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5090&signature=9eb67e21aad9d8a388dcdb9259606f9ac199faa9c7b1773314374e63b06b0439"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/vite/tests/source-root.spec.ts
+++ b/packages/dashboard/vite/tests/source-root.spec.ts
@@ -192,6 +192,12 @@ describe('#5086 compiler source root', () => {
             }),
         ).rejects.toThrow(/outside the output directory/);
 
+        // Rejecting is only half of it: the file that would have escaped must
+        // not have been written on the way to the error.
+        expect(listFiles(path.dirname(outputPath))).toEqual(
+            listFiles(outputPath).map(file => `${path.basename(outputPath)}/${file}`),
+        );
+
         await fs.remove(outputRoot);
         await fs.remove(root);
     }, 120_000);
--- a/packages/dashboard/vite/tests/source-root.spec.ts
+++ b/packages/dashboard/vite/tests/source-root.spec.ts
@@ -4,7 +4,7 @@ import os from 'node:os';
 import path from 'node:path';
 import { describe, expect, it } from 'vitest';
 
-import { compile } from '../utils/compiler.js';
+import { commonAncestorDir, compile } from '../utils/compiler.js';
 
 // #5086 — the source root defaulted to the config's own directory, so a config
 // importing from above it produced a relative path with a leading `..` and the
@@ -72,6 +72,28 @@ async function compileProject(configPath: string, sourceRoot?: string) {
 
     return { outputRoot, outputPath, result };
 }
+
+describe('commonAncestorDir', () => {
+    it('returns the deepest shared directory', () => {
+        const base = path.resolve(path.sep, 'projects', 'shop');
+        expect(commonAncestorDir([path.join(base, 'src'), path.join(base, 'shared')])).toBe(base);
+    });
+
+    it('returns an absolute root when the paths only share the filesystem root', () => {
+        // Joining a prefix that stops at the root yields "" on POSIX and a
+        // drive-relative "C:" on Windows; neither may be returned as-is, or the
+        // emitted paths stop resolving under outputPath.
+        const root = path.parse(path.resolve(path.sep)).root;
+        const ancestor = commonAncestorDir([path.resolve(path.sep, 'alpha'), path.resolve(path.sep, 'beta')]);
+
+        expect(ancestor).toBe(root);
+        expect(path.isAbsolute(ancestor as string)).toBe(true);
+    });
+
+    it('returns undefined for an empty list', () => {
+        expect(commonAncestorDir([])).toBeUndefined();
+    });
+});
 
 describe('#5086 compiler source root', () => {
     it('keeps a config importing from above its directory inside outputPath', async () => {
--- a/packages/dashboard/vite/tests/source-root.spec.ts
+++ b/packages/dashboard/vite/tests/source-root.spec.ts
@@ -2,21 +2,31 @@ import fs from 'fs-extra';
 import { randomUUID } from 'node:crypto';
 import os from 'node:os';
 import path, { posix, win32 } from 'node:path';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
-import { commonAncestorDir, compile } from '../utils/compiler.js';
+import type { Logger } from '../types.js';
+import { commonAncestorDir, compile, resolveDefaultSourceRoot } from '../utils/compiler.js';
 
 // #5086 — the source root defaulted to the config's own directory, so a config
 // importing from above it produced a relative path with a leading `..` and the
 // compiled file was written outside outputPath, past the package.json the
 // compiler writes there to declare the module type.
 
-const silentLogger = {
-    info: () => undefined,
-    warn: () => undefined,
-    debug: () => undefined,
-    error: () => undefined,
-};
+interface RecordingLogger extends Logger {
+    errors: string[];
+    warns: string[];
+}
+
+function createRecordingLogger(): RecordingLogger {
+    const logger = { errors: [], warns: [] };
+    return {
+        ...logger,
+        info: () => undefined,
+        debug: () => undefined,
+        warn: (message: string) => logger.warns.push(message),
+        error: (message: string) => logger.errors.push(message),
+    };
+}
 
 function listFiles(dir: string): string[] {
     if (!fs.existsSync(dir)) {
@@ -35,6 +45,18 @@ function listFiles(dir: string): string[] {
     };
     walk(dir);
     return files.sort();
+}
+
+// Plugin discovery can log unrelated resolution notes on some setups; what
+// must stay silent is the #5086 load-time symptom an escaped emit produces.
+function expectNoEscapeSymptom(logger: RecordingLogger): void {
+    expect(logger.errors.join('\n')).not.toMatch(/exports is not defined|Error loading config/);
+}
+
+async function removeDirs(...dirs: string[]): Promise<void> {
+    for (const dir of dirs) {
+        await fs.remove(dir);
+    }
 }
 
 async function createProject(imported: 'above' | 'below') {
@@ -58,19 +80,29 @@ async function createProject(imported: 'above' | 'below') {
     return { root, configPath: path.join(root, 'src', 'vendure-config.ts') };
 }
 
-async function compileProject(configPath: string, sourceRoot?: string) {
+async function compileProject(configPath: string, options?: { sourceRoot?: string; id?: string }) {
+    const id = options?.id ?? randomUUID();
     const outputRoot = await fs.mkdtemp(path.join(os.tmpdir(), 'vendure-source-root-out-'));
     // Nested, as in production: <tempCompilationDir>/compiler/<id>
-    const outputPath = path.join(outputRoot, 'compiler', randomUUID());
+    const outputPath = path.join(outputRoot, 'compiler', id);
+    // An ESM scope above outputPath is what makes an escaped emit fail to
+    // load in production (#5086). Without it, an escaped file resolves as
+    // CommonJS and only the layout assertions would catch the escape.
+    await fs.outputFile(path.join(outputRoot, 'compiler', 'package.json'), '{"type":"module"}');
 
-    const result = await compile({
-        vendureConfigPath: configPath,
-        outputPath,
-        logger: silentLogger,
-        ...(sourceRoot ? { pathAdapter: { sourceRoot } } : {}),
-    });
-
-    return { outputRoot, outputPath, result };
+    const logger = createRecordingLogger();
+    try {
+        const result = await compile({
+            vendureConfigPath: configPath,
+            outputPath,
+            logger,
+            ...(options?.sourceRoot ? { pathAdapter: { sourceRoot: options.sourceRoot } } : {}),
+        });
+        return { outputRoot, outputPath, result, logger };
+    } catch (error) {
+        await removeDirs(outputRoot);
+        throw error;
+    }
 }
 
 describe('commonAncestorDir', () => {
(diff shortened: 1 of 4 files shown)


packages/dashboard/vite/utils/compiler.ts
```diff
@@ -18,10 +18,34 @@ import { findTsConfigPaths } from './tsconfig-utils.js';
...
+    const segments = dirs.map(dir => path.resolve(dir).split(path.sep));
+    const [first] = segments;
+    let shared = 0;
+    while (shared < first.length && segments.every(parts => parts[shared] === first[shared])) {
+        shared++;
+    }
+    if (shared === 0) {
+        return undefined;
+    }
+    return first.slice(0, shared).join(path.sep) || path.sep;
```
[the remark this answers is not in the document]
TheStreamCode: Fixed in `a8e35d4`. `commonAncestorDir` now returns the parsed root whenever the joined prefix is not absolute, which covers both the POSIX empty string and the Windows drive-relative `C:`. Added tests asserting the result is absolute on either platform.

packages/dashboard/vite/tests/source-root.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+            ? path.join(root, 'shared', 'util.ts')
+            : path.join(root, 'src', 'lib', 'util.ts');
+
+    await fs.outputFile(utilPath, 'export const appName = "shop";\n');
+    await fs.outputFile(
+        path.join(root, 'src', 'vendure-config.ts'),
+        [
+            `import type { VendureConfig } from '@vendure/core';`,
+            `import { appName } from '${importPath}';`,
+            `export const config: VendureConfig = { apiOptions: { port: 3000 }, __n: appName } as any;`,
```
Refaat-alktifan: Minor: __n: appName as any works to keep the import live, but it leans on a bogus config field. Putting the value somewhere real, for example apiOptions.hostname, would exercise the loaded config without the cast. Not important.

packages/dashboard/vite/tests/source-root.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+        vendureConfigPath: configPath,
+        outputPath,
+        logger: silentLogger,
+        ...(sourceRoot ? { pathAdapter: { sourceRoot } } : {}),
+    });
+
+    return { outputRoot, outputPath, result };
+}
+
+describe('commonAncestorDir', () => {
```
Refaat-alktifan, line 124: Good that this is exported and unit tested now. The gap is that these run under the native path, so on CI (Linux) they only ever exercise POSIX, and the case that actually bites, drive-letter and segment casing on Windows, is exactly what stays uncovered. Since commonAncestorDir uses the module path directly, a Windows test needs either a path implementation threaded through the function or mocking path as path.win32 in the test, the way the asset-server Windows spec (local-asset-storage-strategy on Windows) does it. A couple of cases (mixed drive casing, mixed segment casing) would lock this down without needing a Windows runner.

packages/dashboard/vite/tests/source-root.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+        const { outputRoot, outputPath, result } = await compileProject(configPath, root);
+
+        expect(listFiles(outputPath)).toEqual(['package.json', 'shared/util.js', 'src/vendure-config.js']);
+        expect(result.vendureConfig.apiOptions?.port).toBe(3000);
+
+        await fs.remove(outputRoot);
+        await fs.remove(root);
+    }, 120_000);
+
+    it('does not mistake a leading-dot filename for an escape', async () => {
```
Refaat-alktifan: Good edge to cover. A filename that starts with dots is exactly what a naive .. prefix check gets wrong, nice that it is explicit.

packages/dashboard/vite/utils/compiler.ts
```diff
@@ -18,10 +18,42 @@ import { findTsConfigPaths } from './tsconfig-utils.js';
...
+ */
+export function commonAncestorDir(dirs: string[]): string | undefined {
+    if (dirs.length === 0) {
+        return undefined;
+    }
+    const resolved = dirs.map(dir => path.resolve(dir));
+    const segments = resolved.map(dir => dir.split(path.sep));
+    const [first] = segments;
+    let shared = 0;
+    while (shared < first.length && segments.every(parts => parts[shared] === first[shared])) {
```
Refaat-alktifan: This comparison is case-sensitive, and that is the one thing I would still fix here. Windows path casing is not stable in practice: process.cwd(), a resolved relative import, and a tsconfig alias resolution can each hand back a different drive letter or segment casing for the same folder. Running your current commonAncestorDir under path.win32:

 - C:\repo\app\src and c:\repo\shared return undefined. That falls through to configDir below, so the upward file gets a leading .. again and the guard throws. A valid project fails to compile on Windows, which is where this issue was reported.
 - C:\Repo\app\src and C:\repo\shared return C:\. It only works because the Windows filesystem is case-insensitive, and it widens all the way to the drive root instead of the real shared folder.

The drive-root join you just fixed and this are the same underlying thing: hand-rolled segment math that does not follow Windows rules. Two clean options: lowercase the segments when path.sep === '\\' before comparing, or drop the manual walk and use path.relative, which already compares case-insensitively on win32 and gets UNC and drive roots right for free. We have a small helper that does the latter with an injectable path implementation, glad to share it.

packages/dashboard/vite/utils/compiler.ts
```diff
@@ -239,8 +275,18 @@ async function compileTypeScript({
...
+    // The default is the deepest directory containing the config and every file
+    // it imports. For a config whose imports all sit alongside or below it, that
+    // is exactly its own directory, so the layout is unchanged. It only widens
+    // when a collected file lives above the config — otherwise `path.relative`
+    // below would produce a leading `..` and emit that file outside outputPath,
+    // past the package.json written to mark the output's module type.
+    const configDir = path.dirname(inputPath);
+    const sourceRoot =
+        customSourceRoot ??
+        commonAncestorDir([configDir, ...sourceFiles.map(file => path.dirname(file))]) ??
```
Refaat-alktifan: Small point tied to the casing one: when commonAncestorDir returns undefined, this quietly falls back to configDir, which guarantees an upward import escapes and then trips the guard later with a generic message. It reads like a safe default, but it turns a resolvable layout into a hard failure. A dedicated error here ("config imports span multiple roots, set pathAdapter.sourceRoot") would tell the user what to do instead of surfacing downstream as an escape.

packages/dashboard/vite/types.ts
```diff
@@ -17,6 +17,17 @@ export type GetCompiledConfigPathFn = (params: {
...
+     * @description
+     * The config's path relative to `outputPath`, reflecting where the compiler
+     * actually emitted it. This equals `configFileName` unless the source root
+     * was widened because the config imports a file from above its own
+     * directory, in which case the config is nested to keep that file inside
+     * `outputPath`.
+     *
+     * @since 3.8.0
+     */
+    configOutputRelativePath?: string;
```
Refaat-alktifan: Not blocking, more of a design call. This widens the public GetCompiledConfigPathFn for every adapter author, and the fix does not strictly need it: the default compiled path can be derived internally from the resolved sourceRoot, and a custom getCompiledConfigPath already owns its own layout so it can be left untouched. The upside of your version is that custom adapters now receive the relative path too, which is a genuine hook. Just worth deciding whether that hook earns the extra public surface and the version bump, versus keeping the whole change internal.

Refaat-alktifan (commented):
Thanks for taking this on and for the credit. The approach is right, the guard is a good addition, and exporting commonAncestorDir with unit tests is the structure this needed.
The one thing I would fix before merge is the Windows handling in commonAncestorDir. The segment comparison is still case-sensitive, so on Windows a drive-letter mismatch (C: vs c:) returns undefined and fails through the guard, and a segment casing mismatch over-widens to the drive root. I left the path.win32 output inline. Comparing case-insensitively on Windows, or switching to path.relative, fixes both, and the new commonAncestorDir tests are POSIX-only right now so a couple of win32 cases would keep it from regressing quietly.

The rest is minor: a note on the silent configDir fallback, and a soft one on the new public param. We hit this same bug in our own setup, so I will run your branch against the Oronts config as you asked and report back the emitted layout. I have a small tested helper and win32 tests if they are useful, no pressure to take them.

TheStreamCode:
Thanks — the casing point was right, and it broke exactly where this was reported. Under `path.win32`, `C:\repo\app\src` and `c:\repo\shared` returned `undefined`, fell through to the config directory and tripped the guard; `C:\Repo\app\src` and `C:\repo\shared` widened to `C:\`. Both fixed.

I took the `path.relative` option rather than lowercasing. Walking up with it removes the segment maths entirely and leaves casing, drive roots and UNC to the platform. `commonAncestorDir` now takes an optional path implementation, so the Windows cases are asserted with `path.win32` and run on Linux CI: drive-letter casing, segment casing, drive root, and separate drives.

Also from your review:

- **Silent fallback** — no common root is now a dedicated error naming `pathAdapter.sourceRoot`, instead of falling back to the config directory and surfacing later as a generic escape.
- **Public surface** — you were right that the fix does not need it. `GetCompiledConfigPathFn` is unchanged; the default compiled path is derived internally from the resolved source root, and a custom adapter keeps owning its own layout. No version bump.
- **Test config field** — now sets `apiOptions.hostname` rather than a bogus one.

CodeRabbit also asked the guard test to assert that nothing escaped before the rejection, which is a fair half of that contract, so that is in too.

Thanks for the offer of the helper — the injectable-path shape you described is what this ended up with, so it is covered. The Oronts run is the part I cannot reproduce here, so that would still be valuable whenever you get to it.

Refaat-alktifan:
> Thanks — the casing point was right, and it broke exactly where this was reported. Under `path.win32`, `C:\repo\app\src` and `c:\repo\shared` returned `undefined`, fell through to the config directory and tripped the guard; `C:\Repo\app\src` and `C:\repo\shared` widened to `C:\`. Both fixed.
> 
> I took the `path.relative` option rather than lowercasing. Walking up with it removes the segment maths entirely and leaves casing, drive roots and UNC to the platform. `commonAncestorDir` now takes an optional path implementation, so the Windows cases are asserted with `path.win32` and run on Linux CI: drive-letter casing, segment casing, drive root, and separate drives.
> 
> Also from your review:
> 
> * **Silent fallback** — no common root is now a dedicated error naming `pathAdapter.sourceRoot`, instead of falling back to the config directory and surfacing later as a generic escape.
> * **Public surface** — you were right that the fix does not need it. `GetCompiledConfigPathFn` is unchanged; the default compiled path is derived internally from the resolved source root, and a custom adapter keeps owning its own layout. No version bump.
> * **Test config field** — now sets `apiOptions.hostname` rather than a bogus one.
> 
> CodeRabbit also asked the guard test to assert that nothing escaped before the rejection, which is a fair half of that contract, so that is in too.
> 
> Thanks for the offer of the helper — the injectable-path shape you described is what this ended up with, so it is covered. The Oronts run is the part I cannot reproduce here, so that would still be valuable whenever you get to it.

This all looks right to me now. I re-ran commonAncestorDir under path.win32 on your latest and the four cases land where they should: drive-letter and segment casing both resolve to the real shared folder (C:\repo and C:\Repo), the drive root returns C:\, and separate drives return undefined. Moving to path.relative and threading the path implementation through was the right call, it hands casing, drive roots and UNC to the platform instead of re-deriving them.

The dedicated multi-root error, deriving the compiled path internally with no public change, and the hostname test field all read well too.

On the Oronts run: the layout that triggered this for us is a monorepo where the config imports a shared package from above its directory, which is exactly the case your above-import test covers. I will run your branch against it and post the emitted tree here so you have a real-world confirmation alongside the synthetic fixture.

biggamesmallworld (changes requested):
Approach is right. I applied the patch locally and confirmed it fixes the real symptom. Cleanup needed before merge.

**Blocking**

1. `compiler.ts:127-128` — the `getCompiledConfigPath` const is now dead (line 168 calls the adapter directly). That also makes `defaultPathAdapter.getCompiledConfigPath` at line 21 unreachable, and it still states the old default. Delete both, or restore the commit-1 design.
2. `vite/types.ts:94` — "Defaults to the directory containing the `vendureConfigPath` file" is exactly what this PR changed. Ships to the docs site (`vendure-dashboard-plugin.mdx:297`).
3. The description documents `configOutputRelativePath` on `GetCompiledConfigPathFn`. That was reverted, it is not in the diff. Also "all but the downward-import control fail on master": the leading-dot test passes on master too.

**Non-blocking**

4. `assertWithinOutputPath` re-implements `isWithin` with the arguments reversed. Use `if (!isWithin(outputPath, outputFilePath))`.
5. The guard throws mid-loop, leaving partially written output. Compute all output paths, validate, then write.
6. `compileTypeScript` returns `{ sourceRoot }`, echoing one of its own parameters, and `compileInternal` re-derives the config path from it. Resolve the root in `compileInternal` and pass it down.
7. No log of the resolved root, and it can widen a long way. With a tsconfig alias into a sibling tree I get `a/b/c/d/app/src/vendure-config.js` and `x/y/z/other/util.js` under `outputPath`. Correct per the design, but add a `logger.debug` of the root and a `logger.warn` when it lands above the config directory.

**Tests**

8. None of them reproduce #5086. They use a tmpdir with no enclosing `package.json`, so the escaped file resolves as CommonJS and loads fine, and they fail on layout instead. Write `{"type":"module"}` at `outputRoot` in the fixture and the "above" test fails on master for the real reason. The symptom goes to `logger.error`, the thrown error is `Could not find a variable exported as VendureConfig with the name "config"`, so assert on the logger.
9. `toContain('..util.js')` where the sibling tests pin the full list. Use `toEqual(['..util.js', 'package.json', 'vendure-config.js'])`.
10. Two assertions derive expected from actual (`listFiles(dirname(outputPath))` vs `listFiles(outputPath).map(...)`). Pin a literal.
11. Uncovered: the `pathAdapter.getCompiledConfigPath` branch with a widened root, and the multiple-filesystem-roots throw through `compile()`.
12. Nits: file mode is `100755`; timeouts are positional `}, 120_000)` where every sibling spec uses `it('...', { timeout: 60_000 }, ...)`; cleanup is trailing statements, so a failed assertion leaks a temp dir.

Verified locally: patch applied, `vite/tests` passes 122 / 4 skipped, and the fix resolves `exports is not defined in ES module scope` once an ESM `package.json` sits above `outputPath`.

TheStreamCode:
@biggamesmallworld Thank you for applying the patch locally and confirming the real symptom — the ESM enclosure was exactly what our fixtures were missing.

All twelve points are addressed on the branch (`c1ccc44ec`, `06fec0d74`, `6be28fd89`):

**Blocking**

1. Both removed — the dead const and the unreachable `defaultPathAdapter.getCompiledConfigPath` entry, with the `Pick` type narrowed to match.
2. The `sourceRoot` JSDoc now describes the ancestor-based default, and `vendure-dashboard-plugin.mdx` was regenerated from it.
3. Description corrected: the reverted `configOutputRelativePath` mention is gone, and the `master` claim now reflects reality — the leading-dot case loads fine there and is covered by layout assertions.

**Non-blocking**

4. `assertWithinOutputPath` is now `!isWithin(outputPath, outputFilePath)`.
5. Transpilation first collects every emit, then validates all destinations, then writes — a rejected compile leaves `outputPath` empty, which the guard test now pins.
6. Root resolution lives in a `resolveSourceContext` step called from `compileInternal`; `compileTypeScript` takes `sourceRoot` and returns nothing.
7. `logger.debug` reports the resolved root, and when the default widens above the config directory `logger.warn` announces it ("Source root widened to …"). A test pins the warning.

**Tests**

8. Every fixture writes `{"type":"module"}` into the enclosing `compiler/` directory, so on `master` an escaped emit fails exactly as reported, through `logger.error`. The success specs assert that symptom stays absent alongside the loaded hostname.
9. Pinned: `['..util.js', 'package.json', 'vendure-config.js']`.
10. Derived-from-actual assertions are gone; each case uses a fixed compile id so even the parent-directory listing is a literal.
11. The adapter branch is covered by a spy asserting `getCompiledConfigPath` is called once with `{inputRootDir, outputPath, configFileName}` and that its return value is what gets imported. For multi-root: a cross-drive layout cannot be constructed portably on POSIX CI, so the throw is exercised through the newly exported `resolveDefaultSourceRoot` with `path.win32` injected — the same seam `commonAncestorDir` already uses. Happy to thread a deeper seam through `compile()` if you would rather see it end-to-end.
12. Executable bit dropped, timeouts switched to the sibling object style, temp-dir cleanup moved into `finally`.

Locally verified: `vite/tests` 505 tests with only the pre-existing Windows-local failures untouched by this change, source-root spec 15/15 green, `check-types` clean, Prettier applied.

TheStreamCode:
One CI note: `dashboard e2e (3)` failed on the latest push with a strict-mode violation (`[role="dialog"] option /Test Payment/` resolving to two elements) plus a draft-order click timeout — while shards 1, 2 and 4 are green, including the rest of the orders suite. That signature looks like test-data pollution rather than anything this branch touches (the change is confined to the config compiler's emit layout). I don't have rerun rights on this repository, so if the failures reproduce as flake, could someone re-run the failed jobs?

biggamesmallworld (changes requested):
Sorry I missed this on the first review, but I ran a few tests and came to the following conclusions.

Twelve points all addressed. One new problem, then this is an approve.

**Blocking**

1. Widening the source root breaks tsconfig path aliases.

`registerTsConfigPaths` (`compiler.ts:501`) registers `baseUrl: outputPath`, but the patterns are relative to the tsconfig `baseUrl`. Those only line up when the emit is laid out relative to that directory. `sourceRoot` used to be the config directory, so they lined up for anyone with a tsconfig sitting there. Widening breaks it.

Config at `<root>/src/vendure-config.ts`, `<root>/src/tsconfig.json` with `baseUrl: "."` and `paths: {"@plugins/*": ["plugins/*"]}`, importing `@plugins/foo` and `../shared/util.js`. No `pathAdapter`.

master:
```
[package.json, plugins/foo.js, vendure-config.js]
loaded { port: 4242, hostname: "from-above" }
```
this branch:
```
[package.json, shared/util.js, src/plugins/foo.js, src/vendure-config.js]
Error loading config: Cannot find module '@plugins/foo'
```

Same on `3b4c093`, so it is the common-ancestor approach itself, not anything from the last round.

Fix in `registerTsConfigPaths`, threading `sourceRoot` in from `compileInternal`:

```ts
const emittedBaseUrl = isWithin(sourceRoot, tsConfigInfo.baseUrl)
    ? path.join(outputPath, path.relative(sourceRoot, tsConfigInfo.baseUrl))
    : outputPath;
```

Repro loads, suite stays at 123.

Keep the `isWithin` guard. I tried folding the tsconfig `baseUrl` into the ancestor set instead: no effect, it only widens a root that is already too wide. Both together do fix the repro, but they push every project with `paths` above its config a directory deeper and fail 8 tests, including the `/fixtures-path-alias` strip in `path-alias.spec.ts`, which turns into a double correction. That is a layout change for all alias users, so it is a `minor` change, not this one.

Add a test for the repro. No current fixture catches it, they all put config, tsconfig and source root in the same directory.

**Nits**

2. "Breaking changes: None" is wrong. `assertWithinOutputPath` is a new throw. A CommonJS project with a narrow `pathAdapter.sourceRoot` works today because the escaped emit still resolves, and throws after this. Right behaviour, but it needs to reach the release notes.

3. `it(name, fn, { timeout })` is the deprecated third-argument form, seven warnings per run. Siblings use `it('...', { timeout: 60_000 }, async () => {})`.

Verified locally: `vite/tests` 123 / 4 skipped, `check-types` clean. `dashboard e2e (3)` is unrelated flake in `orders.spec.ts`, needs a re-run.

TheStreamCode:
Update: the rerun reproduced the same failure, so it is not transient — and it does not look specific to this branch. Over the last 24 hours `dashboard e2e (3)` has failed on three other PRs (#5039, #5173, #5171) with an identical strict-mode violation, while that shard matrix does not run at all on master pushes. The signature is a data-pollution pattern: `orders.spec.ts` resolves the `/Test Payment/` locator against two methods (`e2e-test-payment` plus the standard `test-payment`). On those branches it surfaces as a single failed test; on this branch it additionally cascaded into a click timeout in the same file. Since several unrelated branches hit it within one day, it looks like shard-3 ordering/cleanup in the suite rather than anything any of these changes touch. Happy to help isolate the leaking setup step if that's useful.

biggamesmallworld:
You called it, and I was wrong to put that down to flake. #5175 fixes it.

Options are labelled `${name} (${code})`, and `payment-methods.spec.ts` creates "E2E Test Payment", so `/Test Payment/` matched both while that spec sat between its create and update steps. The cascade is `mode: 'serial'` on the `Orders` describe: the retry re-ran the draft order test against an already populated database, timed out, and skipped the other 17. Same signature on #5171 and #5173.

Nothing on this branch caused it, so it should not count against this PR.

Still open from the review:

1. The alias regression, with the `isWithin` fix in `registerTsConfigPaths`.
2. "Breaking changes: None".
3. The deprecated third-argument timeouts.

TheStreamCode:
Root cause identified — it is not shard-3 cleanup ambiguity after all, and none of the affected branches are involved. Upstream commit `06e6fcfd6` ("test(dashboard): check there is one payment after manual payment is added", merged to `master` on Aug 18) introduced an orders-lifecycle test whose payment-method lookup is a regex, `/Test Payment/`. That pattern also matches `E2E Test Payment` — the method `settings/payment-methods.spec.ts` keeps alive between its create and bulk-delete steps. With `fullyParallel: true` and four workers sharing one Vendure server/database, the two specs can overlap inside shard 3, and the strict-mode locator then resolves to two elements. The serial `Order lifecycle` describe aborts its remaining tests afterwards, which produces the `did not run` entries. Since the dashboard e2e shard matrix does not run on `master` pushes, this never fires there; it only surfaces on pull-request merge refs, which now include that commit for every open PR.

A minimal fix would be matching by exact accessible name (`name: 'Test Payment'` as a string performs a full match, so it cannot collide with `E2E Test Payment`) instead of the regex. Happy to send that as a small standalone PR if useful.

Edit: #5175 already landed this a few hours ago — it matches the full accessible name ('Test Payment (test-payment)'), which is even tighter than the suggestion above. Shard 3 should recover for every affected PR once merge refs pick that up.

TheStreamCode:
All three are addressed on the branch (`e47ce81b1`):

1. `registerTsConfigPaths` now re-expresses the tsconfig `baseUrl` in emitted coordinates whenever it sits inside the resolved source root (`path.join(outputPath, path.relative(sourceRoot, baseUrl))`, falling back to `outputPath` when it does not). Added a repro to the specs along the lines of yours: config in `src/`, sibling `tsconfig.json` with `baseUrl: "."` mapping `@plugins/*`, importing both the alias and `../shared/util.js`. Without the remap the alias import fails to load and the compile surfaces the generic export-not-found throw; with it the alias resolves against the nested emit and the loaded config carries the plugin value, which the spec pins. Local suite: 486 passed / 506 total, only the pre-existing Windows-local failures.

2. "Breaking changes" no longer claims none — it now names the new emit-time throw and the widened-root layout for the release notes.

3. The seven timeouts moved to the second-argument position.

biggamesmallworld (approved):
Looking good! 👍

PR #5091, opened 2026-08-05
feat(core): Translate configurable operation strings and resolve them from Accept-Language

michaelbromley:
Two follow-ups to #5079, which landed the mechanism for resolving configurable operation strings from the i18n catalogs without shipping any translations.

## 1. The translations themselves

Adds the `configurableOperation` namespace to every message catalog: 46 distinct strings across core's 23 shipped operations, 55 key paths per locale, 1485 entries in total. Twenty of the 27 locales had no core catalog at all before this, and they are picked up automatically because `I18nService` derives its supported-language list from the filenames in the messages directory.

Generated with `npm run i18n:extract` / `i18n:apply`, so the key paths come from `getTranslationKeys()` rather than from anything hand-written. Terminology is anchored to the Dashboard's existing `.po` catalogs, so an administrator sees the same wording in the interface chrome and in the operation descriptions.

English is unchanged. It stays inline in each operation's `.ts` file, which is what makes a missing translation fall back to a real sentence rather than a raw key path.

## 2. Resolving them against the reader's language, not the content language

These strings describe the operation itself rather than the data it acts on, so they belong in the language the administrator reads. They were being resolved against `ctx.languageCode`, which both admin clients set from their content-language setting. An administrator reading the interface in Japanese while editing the German translation of a product saw German labels.

That conflation predates #5079 — `localizeString` has always used `ctx.languageCode`, and the server has never been told anything else. What #5079 changed is visibility: before it, hardly any operation had non-English inline strings, so nearly everyone saw English regardless of which axis was used. With 27 locales in place the wrong axis is now something you can see.

### Approach

The server now takes the two languages from different places:

- `?languageCode=` selects which translation of the **content** to return, exactly as before.
- `Accept-Language` states which language the **response's own text** should be written in.

`Accept-Language` is the standard HTTP semantic for this, so the server contract is "we honour `Accept-Language` for server-generated text" rather than anything specific to the Dashboard. Any consumer expresses it the same way. It was also already the third link in the chain `i18next-http-middleware` uses to produce `req.t`; it never won because both admin clients always send the querystring parameter.

`RequestContext` gains `acceptedLanguageCodes`, and the resolution chain becomes the accepted languages in header order, then `ctx.languageCode`, then the Channel default, then English. Prepending rather than replacing is what keeps every existing client working: a client which sends no header resolves exactly as it does today, including the whole Shop API, where a visitor's reading language and content language are the same thing anyway.

The GraphQL contract is untouched. `ConfigurableOperationDefinition.description` stays `String!` and `ConfigArgDefinition.label` stays `String`.

### Both clients send the header

The Dashboard and the Angular Admin UI now send `Accept-Language` from their display-language setting. This has to land together with the server change rather than after it, because browsers send their own `Accept-Language` derived from the OS locale. Honouring the header without the clients overriding it would have replaced one wrong answer with another, and an administrator with a German browser and an English interface would suddenly have seen German labels.

### Option labels

Select option labels were merged into the array keyed by the content language, while the client looks them up by display language, so when the two settings differed the merged value was never found. This is why the tax-setting options stayed English in a Japanese interface even when everything around them had translated.

They are now tagged with the language the client asked for rather than the language the catalog entry was found under. Every client does an exact match and then falls back to the first entry in the array, so a match reached by truncating `pt_BR` to `pt`, or from the `languageCode` / channel-default / English tail of the chain, would otherwise be tagged with a language the client never looks for and the translation would go unused.

## Header parsing

`parseAcceptLanguage()` handles quality ordering, drops `q=0` entries and the wildcard, and truncates each tag subtag by subtag, so `zh-Hans-CN` tries `zh_Hans_CN`, then `zh_Hans`, then `zh`. Casing is normalised to the convention `LanguageCode` follows, so a client sending `pt-br` still matches `pt_BR`. Tags which are not well-formed language tags are dropped rather than passed through, on the same reasoning as the existing format check on the `languageCode` query parameter.

It is capped at ten tags. Every accepted language lengthens the chain walked for each string on each request, so an arbitrarily long header would otherwise multiply the lookup work.

## Custom fields are deliberately untouched

`CustomField.label` is already `[LocalizedString!]` and the Dashboard already resolves it against the display language, so custom fields are on the correct axis and need no change here. Their remaining gap is that the strings can only come from inline arrays rather than the catalogs, which the same catalog-merge treatment used for option labels would fix without a schema change. They are also typically written by the project rather than shipped by core, which makes the inline array a reasonable place for them.

The two therefore end up on different models — configurable operations resolved server-side, custom fields resolved client-side — which is a real inconsistency to settle in a major, and not one worth a breaking change now.

## Behaviour change for other browser-based Admin API clients

Browsers send an `Accept-Language` header of their own on every request, derived from the operating system locale. A browser-based Admin API client which is not the Dashboard or the Admin UI, and which has been relying on the `languageCode` query parameter alone for these strings, will therefore start resolving them against that OS locale.

This is the correct behaviour rather than an accident of the design, since the header is what states the reader's language, but it is a change for those clients and should go in the release notes. The fix on their side is one line: send `Accept-Language` from whatever language the interface is displayed in, which is what both first-party clients now do.

Server-to-server clients are unaffected, since they only send the header if they choose to. So is the Shop API, where a visitor's reading language and content language are the same thing.

## Testing

- 20 unit tests for the header parser, covering quality ordering, subtag truncation, casing, whitespace, malformed and out-of-range weights, the tag cap and the subtag bound
- 6 new unit tests for the resolution order, including the case where the header and `languageCode` disagree, and one for an option label matched by truncation
- 4 new e2e tests exercising the real header end to end, plus one added earlier that proves core's own shipped catalogs are loaded and read
- 2 new tests covering `acceptedLanguageCodes` through `serialize`/`deserialize`, including a payload queued before the field existed
- Full core unit suite (1251 passed) and the e2e suites which exercise `toGraphQlType`: translations, promotion, order-promotion, promotion-side-effects, shipping-method, shipping-method-eligibility, collection, payment-method (227 passed, 1 skipped)
- Dashboard type check passes

Verified in the browser: with the content language set to German and the display language to Japanese, the shipping method's calculator description, arg labels and option labels all render in Japanese while the product data stays German.

Relates to #4820

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5091"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788525035&installation_model_id=20193&pr_number=5091&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5091&signature=902e682da80f45efd67631974a6648715a52a5baf65545ef6f8ddf6721af4560"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


HouseinIsProgramming (changes requested):
Design is right, and splitting the two languages onto separate transports is the correct call. Six things before it lands.

### What I verified

- `_acceptedLanguageCodes` round-trips through `serialize` / `deserialize` (`request-context.ts:214`, `:253`), and the `?? []` means job payloads queued before the upgrade behave exactly as before.
- Every producer funnels through the single `languagePreference()` path, so `RequestContext.empty()`, the importer's hand-built context and `RequestContextService.create()` without a `req` all keep the old `[languageCode, channelDefault, en]` chain.
- The per-language interleaving of catalog and inline lookup at `configurable-operation.ts:562-575` is the right shape, and the non-mutating merge plus the leak test at `translations.e2e-spec.ts:122` is good discipline.

### Requested changes

**1. Both new unknown-language e2e cases currently assert nothing.**

`translations.e2e-spec.ts:203` and `:212` send `xx-unsupported`. The `-unsupported` subtag is 11 characters, so it fails the `{1,8}` bound in the regex at `parse-accept-language.ts:29` and is dropped at parse time rather than at catalog lookup.

- `:203` is left with one surviving tag, so nothing is ordered.
- `:212` parses to `[]`, which is byte-for-byte the no-header case.

For `:212`, `xx` alone is enough. For `:203`, `xx;q=0.9, ja;q=0.8` is not, because header position and quality order still agree, so a parser that ignored `q` entirely would pass. Use two languages that both have shipped catalogs with the preferred one second, for example `de;q=0.8, ja;q=0.9` asserting the Japanese description.

**2. A tag with trailing whitespace is silently discarded.**

`parse-accept-language.ts:26` is `entry.trim().split(';')`, which only strips the outer edges, so the RFC 9110-legal `de ;q=0.5` yields the tag `'de '` and fails the regex at `:29`. Change it to `entry.split(';').map(s => s.trim())` and drop the now-redundant `.map(p => p.trim())` at `:32`.

While you are in there, `p.startsWith('q=')` at `:32` is case-sensitive. Parameter names are not, so `Q=0.5` is read as `q=1`, which promotes the tag to full weight and reorders the whole chain. Both want a spec case.

**3. `MAX_TAGS` does not bound what the function returns.**

The comment at `:7` justifies the cap by the length of the resulting lookup chain, but the slice at `:43` caps tags while the expansion loop at `:44-49` emits one code per subtag prefix, and the regex at `:29` puts no limit on subtag repetitions. A single tag can therefore return thousands of codes.

- `languagePreference()` rebuilds that array once per resolved string, and `lookupCatalog` runs once per code per string. `promotionActions` resolves dozens of strings per request.
- The array is also copied verbatim into every serialized `RequestContext` that reaches the job queue.

Bound it, cheapest as `(-[a-zA-Z0-9]{1,8}){0,3}` in the regex.

**4. Option labels are merged under the matched language, not the requested one.**

`configurable-operation.ts:646` tags the merged label with the `languageCode` the catalog hit came from. The comment at `:613-614` says that "is the language the client will look for it under", which does not hold whenever the match arrived via subtag truncation or via the `ctx.languageCode` / channel-default / `en` tail of the chain.

Every client does an exact match and then falls back to `label[0]`: `select-with-options.tsx:52`, `configurable-operation-arg-summary.tsx:198`, `custom-field-label.pipe.ts:34`. So a plugin that ships translations under `pt` while the admin reads in `pt_BR` silently shows English option labels, even though a translation was found.

Tag the merged entry with the requested code, `ctx.acceptedLanguageCodes[0] ?? ctx.languageCode`. That is exactly what the client looks up, since both admin clients send their display language as the header.

**5. Extend the `RequestContext` serialization tests.**

`request-context.spec.ts` has a field-by-field `serialize` / `deserialize` suite at `:12-84` and a `copy` getter suite at `:103-114`. A new serialized field was added at `request-context.ts:26`, `:214` and `:253` with no case in either. Two three-line tests.

**6. Tighten the parser spec assertions.**

- `parse-accept-language.spec.ts:58` uses `toHaveLength(10)`, which passes whichever ten survive. Assert the exact array.
- There is no case proving the cap keeps the highest-quality tags rather than the first ten. Put the high-`q` tags late in the header.
- No case for malformed quality: `q=`, `q=abc`, `q=1.5`.

### One note for the description

`configurable-operation.ts:584` says `ctx.languageCode` is consulted after the header "so that a client which sends no header behaves as it always has". Browsers always send `Accept-Language`, so any browser-based admin client that is not the updated Admin UI or Dashboard now has its OS locale override the `languageCode` query param it has been sending. That is the right trade-off, but it should be in the PR description and the release notes rather than implied away in a comment.

Will approve once that's in.

michaelbromley:
@HouseinIsProgramming thanks for the review - fixed those things now.

PR #4940, opened 2026-07-10
fix(dashboard): Show loading indicator when uploading assets

latifniz:
## Description

When uploading assets via Catalog → Assets, there was zero visual feedback during the upload process. The UI appeared completely frozen until all files finished uploading, then they appeared all at once. This was confusing especially when uploading multiple large files.

## What changed

Added upload progress feedback to the `AssetGallery` component:

- The Upload button shows a spinner and becomes disabled while upload is in progress, preventing accidental double-submissions
- A semi-transparent overlay with a spinner and "Uploading assets..." message appears over the gallery grid during upload
- Once the upload completes the overlay disappears and the gallery refreshes automatically

## Root cause

`useMutation` from React Query exposes an `isPending` state but it was never being used. The fix simply destructures `isPending` and wires it up to the UI.

## Notes

True per-file progress (e.g. percentage bar) is not possible with the current architecture because `api.mutate` uses the Fetch API which does not expose upload progress events. This fix gives the best feedback possible without rewriting the upload mechanism to use XHR.

The overlay also shows in the asset picker dialog (used when selecting assets on product/category pages) which is correct behaviour.

## Checklist

- [x] I have set a clear title
- [x] My PR is small and contains a single change
- [x] I have checked my own PR

Fixes issue #4936 

here is video after fix ( this is local host so it takes less than a second but loader is visible, the issue i reported was in prod so it took a lot of time , see video in the issue as well)

https://github.com/user-attachments/assets/c5e5ef53-4514-4bd4-a40c-4abfc7f8b164

---
[...]

Issue #4936, opened 2026-07-09
[Asset Manager] No upload progress indicator when uploading multiple assets


latifniz:
### Description
When uploading multiple assets via Catalog → Assets, there is no loading indicator, 
progress bar, or any visual feedback during the upload process. The UI appears 
as nothing has happend.

After all files finish uploading (which can take up to a minute for multiple files), 
the uploaded images suddenly appear all at once with no prior indication that 
anything was happening.

### Expected Behavior
A progress indicator, loading spinner, or upload progress bar should be shown 
while assets are being uploaded so the user knows the upload is in progress.

### Steps to Reproduce
1. Go to Catalog → Assets
2. Select and upload multiple image files at once
3. Observe: no feedback is shown during upload

### Environment
- Vendure version: 3.7.0
- Browser: Chrome
- OS: Debain 13

Here is the video, please watch full.

https://github.com/user-attachments/assets/d876be77-ea14-4558-99bf-a2e0bbae3b6e

RISO525:
Hi! I'd be interested in contributing if there is still any work remaining on this issue. I noticed the linked fix (#4940) and the follow-up enhancement (#5052), so I wanted to check whether this issue still needs any implementation or whether new contributions should target the follow-up instead.

PR #5092, opened 2026-08-05
feat(dashboard): add per file upload progress modal for assets

### LINKED ISSUE #5052: [Dashboard] Asset upload: per-file progress modal (follow-up to #4940)
### Summary

Follow-up enhancement to #4940 (which adds a single global "Uploading assets…" overlay + button spinner to the Asset Manager, resolving #4936).

Take the upload feedback further: when multiple assets are selected for upload, show a **modal that lists each selected file with its own per-file progress bar**, so the user can see each asset initialise and upload individually — rather than a single global overlay for the whole batch.

### Suggested behaviour

- On selecting multiple files, open a modal listing each file (name / thumbnail).
- Each row shows its own state: queued → uploading (with a progress bar) → done / error.
- Overall progress and a close/dismiss when finished.

### Notes / considerations

- The current `AssetGallery` upload path uses a single `createAssets` GraphQL multipart mutation, which doesn't expose per-file progress. Per-file progress will need either uploading files individually with `XMLHttpRequest` `upload.onprogress` events, or another mechanism to surface progress per file — so this is more involved than the global overlay in #4940.
- Should build on top of #4940 (keep the simple overlay as the fallback / until the modal lands).
- Dashboard component: `packages/dashboard/src/lib/components/shared/asset/asset-gallery.tsx`.

@latifniz volunteered to take this on (see #4940). Assigning accordingly.

Relates to #4936, #4940.

latifniz:
## Description

Follow-up to #4940. That PR added a global "Uploading assets..." overlay as
the immediate fix for #4936, but noted that true per-file progress wasn't
possible without moving off the fetch-based mutation. This PR does that
move and adds real per-file progress.

## What changed

Replaces the silent mutation-based upload in `AssetGallery` with a modal
that lists every selected file individually, showing:

- A live 0–100% progress bar per file
- A status icon (uploading / success / error) per file
- Any server-side error for that specific file (e.g. the 20 MB file size
  limit) — without blocking the other files in the batch

## Root cause / approach

`api.mutate` uses the Fetch API, which doesn't expose upload progress
events. This switches asset uploads to XHR (`upload.onprogress`) instead,
with one XHR per file, all files uploading in parallel via `Promise.all`.

## Notes

- Builds on top of #4940 (the loading-indicator overlay this replaces) —
  this branch is stacked on it, so it should merge after #4940.
- Targets `minor` since this is a new feature, not a bug fix.

Fixes #5052

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5092"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788525742&installation_model_id=20193&pr_number=5092&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5092&signature=3b32b17887bb542e5b0eb804b2ffec01175fccdfc4696c694921bf4fd41ce116"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/i18n/locales/ar.po
+++ b/packages/dashboard/src/i18n/locales/ar.po
@@ -82,6 +82,13 @@ msgstr "شقة، جناح، إلخ."
 msgid "No more items"
 msgstr "لا توجد عناصر أخرى"
 
+#. placeholder {0}: fileUploads.length
+#. placeholder {1}: fileUploads.length
+#. placeholder {2}: fileUploads.length
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
+msgid "{0, plural, one {Upload progress for {1} file} other {Upload progress for {2} files}}"
+msgstr ""
+
 #: src/lib/components/shared/rich-text-editor/link-bubble-menu.tsx
 msgid "Edit link"
 msgstr "تحرير الرابط"
@@ -443,7 +450,7 @@ msgstr "لا توجد أدوات لإضافتها"
 #: src/app/routes/_authenticated/_products/products_.$id_.variants.tsx
 #: src/app/routes/_authenticated/_products/products_.$id.tsx
 #: src/app/routes/_authenticated/_profile/profile.tsx
-#: src/lib/components/shared/asset/asset-gallery.tsx
+#: src/lib/components/shared/asset/asset-picker-dialog.tsx
 #: src/lib/components/shared/asset/asset-preview.tsx
 msgid "Unknown error"
 msgstr "خطأ غير معروف"
@@ -577,7 +584,7 @@ msgstr "مجموعة جديدة"
 msgid "Insights"
 msgstr "الرؤى"
 
-#: src/lib/components/shared/asset/asset-gallery.tsx
+#: src/lib/components/shared/asset/asset-picker-dialog.tsx
 msgid "Failed to upload assets"
 msgstr ""
 
@@ -1926,6 +1933,7 @@ msgstr "إدراج جدول {row} في {col}"
 #: src/lib/components/data-table/views-sheet.tsx
 #: src/lib/components/layout/manage-languages-dialog.tsx
 #: src/lib/components/shared/asset/asset-focal-point-editor.tsx
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
 #: src/lib/components/shared/assign-to-channel-dialog.tsx
 #: src/lib/components/shared/confirmation-dialog.tsx
 #: src/lib/components/shared/customer-address-form.tsx
@@ -2257,6 +2265,10 @@ msgstr "إزالة من المنطقة"
 #: src/lib/components/layout/manage-languages-dialog.tsx
 msgid "Select from globally available languages for this channel"
 msgstr "حدد من اللغات المتاحة عالميًا لهذه القناة"
+
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
+msgid "Upload timed out"
+msgstr ""
 
 #: src/app/routes/_authenticated/_customers/components/customer-address-card.tsx
 msgid "Are you sure you want to delete this address?"
@@ -3219,6 +3231,10 @@ msgstr "فشل إزالة العميل من المجموعة"
 
 #: src/lib/components/shared/configurable-operation-arg-summary.tsx
 msgid "1 property"
+msgstr ""
+
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
+msgid "Upload failed (HTTP {detail})"
 msgstr ""
 
 #: src/lib/components/layout/nav-user.tsx
@@ -4647,6 +4663,10 @@ msgstr "يساوي"
 msgid "Source File"
 msgstr "الملف المصدر"
 
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
+msgid "Upload failed — check your connection"
+msgstr ""
+
 #: src/app/routes/_authenticated/_products/components/create-product-options-dialog.tsx
 #: src/app/routes/_authenticated/_products/components/create-product-options-dialog.tsx
 #~ msgid "Create options"
@@ -5009,6 +5029,10 @@ msgstr ""
 msgid "A refund of {formattedDiff} is required. Select payment amounts and enter a note to proceed."
 msgstr "مطلوب استرداد بقيمة {formattedDiff}. حدد مبالغ الدفع وأدخل ملاحظة للمتابعة."
 
+#: src/lib/components/shared/asset/asset-upload-modal.tsx
+msgid "File exceeds the server upload size limit"
+msgstr ""
+
 #: src/app/routes/_authenticated/_orders/components/edit-order-table.tsx
 #: src/app/routes/_authenticated/_orders/components/order-table.tsx
 #: src/app/routes/_authenticated/_orders/components/refund-order-dialog.tsx
@@ -5134,10 +5158,9 @@ msgstr ""
 msgid "Failed to create variants"
 msgstr "فشل في إنشاء المتغيرات"
 
-#. placeholder {0}: fileUploads.length
 #: src/lib/components/shared/asset/asset-upload-modal.tsx
-msgid "Upload progress for {0} files"
-msgstr ""
+#~ msgid "Upload progress for {0} files"
+#~ msgstr ""
 
 #: src/app/routes/_authenticated/_global-settings/global-settings.tsx
 msgid "Sets the stock level at which this a variant is considered to be out of stock. Using a negative value enables backorder support. Can be overridden by product variants."
@@ -5876,7 +5899,7 @@ msgid "Failed to remove customers from group"
 msgstr "فشل في إزالة العملاء من المجموعة"
 
 #. placeholder {0}: failedAssets.length
-#: src/lib/components/shared/asset/asset-gallery.tsx
+#: src/lib/components/shared/asset/asset-picker-dialog.tsx
 msgid "Failed to upload {0} assets"
 msgstr ""
 
@@ -6526,7 +6549,7 @@ msgid "Strikethrough"
 msgstr "يتوسطه خط"
 
 #. placeholder {0}: createdAssets.length
-#: src/lib/components/shared/asset/asset-gallery.tsx
+#: src/lib/components/shared/asset/asset-picker-dialog.tsx
 msgid "Uploaded {0} assets"
 msgstr ""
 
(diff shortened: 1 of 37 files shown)


michaelbromley:
Hi, please clean up the commit history of this PR so that it only contains the relevant work, thanks!

latifniz:
Oh. sorry. let me do it quickly

HouseinIsProgramming (changes requested):
Good problem to solve, and the per-file UI is the right shape. The transport needs to move before this can land, and the two CI failures both have causes in this diff.

All line numbers below are in `asset-upload-modal.tsx` unless stated otherwise.

My agent found the following issues, listed in the comment below.

HouseinIsProgramming:
**1. Close the modal once every file has succeeded.**

`uploadAll` at `:170` calls `onComplete()` and stops. Nothing sets `uploadModalOpen` back to false: `onComplete` at `asset-gallery.tsx:801` only invalidates a query, the footer button at `:219` requires a click, and `onOpenChange` at `:183` is gated on `allUploadsFinished`. So a successful upload leaves a modal the admin has to dismiss by hand every time.

This is also the cause of the `dashboard e2e (1)` failure, so it is worth fixing first. `assets.spec.ts:23` uploads a PNG through the dropzone and then waits for its gallery card at `:40`. `@vendure-io/ui` dialogs are Base UI, which marks everything outside the dialog `inert` and `aria-hidden` while it is open, so the role query matches nothing and the test times out with "element(s) not found" rather than "not visible". The asset does upload and the grid does refetch. The modal is simply covering it.

Auto-close on all-success, and stay open only when at least one file failed.

**2. Treat an `ErrorResult` member as a failure.**

`parseGraphqlResponse` at `:91-98` inspects only `json.errors`. `createAssets` returns `Asset | MimeTypeError` and Vendure returns ErrorResults as data with HTTP 200 and no `errors[]`, so a rejected mime type currently shows a green check at 100% while no asset is created. Per-file status is the point of this PR, so it cannot report success for the most common per-file failure.

`createAssetsDocument` already selects `... on ErrorResult { message }` at `asset-gallery.tsx:76-78`, so no schema change is needed. Add `__typename` to that selection anyway and switch on it, matching the pattern at `auth.tsx:183`.

**3. Do not treat a non-JSON 200 as success.**

The bare `catch {}` at `:96-97` falls through to `return { success: true }`, so an HTML error page served with a 200 by a misconfigured proxy counts as a successful upload. Return a failure when the body does not parse.

**4. Move the upload into `api.ts` rather than hand-rolling a second transport.**

`buildApiUrl` at `:62-71` reimplements `api.ts:33-50`, and `applyAuthHeaders` at `:73-78` reimplements `api.ts:26-31`. Two things `api.ts` does that the XHR does not:

- `credentials: 'include'` (`api.ts:55`). The XHR never sets `xhr.withCredentials`, so cookie-session deployments with `api.host` on a separate origin fail every upload. CI does not catch this because the e2e config uses `tokenMethod: 'bearer'`.
- The `CHANNEL_NOT_FOUND` channel-token cleanup at `api.ts:73-79`.

Add something like `api.uploadWithProgress(document, variables, { onProgress, signal })` and have the modal call it. Sharing the code also means the next header added to `api.ts` does not silently miss the upload path.

**5. Break the circular import.**

`:18` imports `createAssetsDocument` from `./asset-gallery.js`, while `asset-gallery.tsx:37` imports `AssetUploadModal` from `./asset-upload-modal.js`. Move the document into its own module. Cycles like this survive until an unrelated bundler change turns them into a TDZ crash.

**6. Finish the i18n work and re-run the extract.**

Three strings are still hard-coded English:

- `:101` `'File exceeds the server upload size limit'`
- `:102` `` `Upload failed (HTTP ${xhr.status})` ``
- `:122` `'Upload failed — check your connection'`

None of the 28 catalogs contains any of them. Note `t` from `@lingui/react/macro` is a hook binding and will not work in these module-level functions, so either use `msg` from `@lingui/core/macro` and resolve at render, or return an error code from `parseServerResponse` and map it to `<Trans>` in the component. The second option also makes the helpers testable for item 9.

Separately, `dashboard i18n sync` fails because the old toast flow was deleted without re-running the extract. Three messages are still live in all 28 catalogs pointing at code that no longer exists (`en.po:580`, `:5830`, `:6474`):

- `Failed to upload assets`
- `Failed to upload {0} assets`
- `Uploaded {0} assets`

Run `npm run i18n:extract --workspace=@vendure/dashboard` and commit the result.

While you are in there, `<Trans>Upload progress for {fileUploads.length} files</Trans>` at `:190` renders "1 files". Use `<Plural>`, as `asset-gallery.tsx:645-649` already does.

**7. Abort in-flight XHRs on cleanup, add a Cancel button, and set `xhr.timeout`.**

Cleanup at `:175-177` only flips `active`; the requests keep running. Under `React.StrictMode`, which this app enables at `main.tsx:171`, every file uploads twice in dev and both succeed, creating duplicates. With no timeout, a stalled connection leaves the admin in a modal they cannot dismiss.

When you add the abort, give the XHR an `onabort` handler. `sendUploadRequest` at `:110-125` resolves only from `onload` and `onerror`, so an unhandled abort leaves that promise pending forever, hanging `Promise.all` at `:169` and disabling Close at `:219` permanently.

**8. Cap upload concurrency, and fix or drop the `'queued'` status.**

`Promise.all` at `:169` fires every file at once. With fifty files the browser queues past its connection limit while the UI shows all of them uploading at 0%. Cap it at 3 or 4.

`'queued'` is assigned at `:145` but overwritten by `'uploading'` at `:153` inside the same synchronous flush, so the `Clock` branch at `:230` never paints. Either render the queued state before starting, or remove the status and its icon.

**9. Fix the completion callback.**

- It is missing from the effect deps at `:178` and is a fresh inline arrow every render at `asset-gallery.tsx:801`, so the effect captures the `queryKey` from the render where `open` flipped true. Change page or search term mid-upload and the invalidation targets the old key.
- It invalidates only that key. The removed mutation also invalidated `[PaginatedListDataTableKey]`, which the other action in this component still does at `asset-gallery.tsx:265`, so the list view keeps showing stale data after an upload. The grid itself is fine, the fully parameterised key does match.
- It fires at `:170` even when every file failed, so "complete" carries no outcome. Pass the result summary.

A `useCallback`'d handler that invalidates the `['AssetGallery']` prefix plus `[PaginatedListDataTableKey]` fixes all three.

**10. Confirm the nested-dialog path.**

`asset-picker-dialog.tsx:109` renders `AssetGallery` without `onFilesDropped`, so dropping a file inside the picker now opens a second dialog inside the first. The picker binds `onOpenChange={onClose}` at `:98`, so a dismiss event escaping the inner dialog closes the picker and discards the selection. Either verify this is fine or pass `onFilesDropped` through from the picker.

**11. Add unit tests for the pure helpers.**

`parseServerResponse`, `parseGraphqlResponse` and `overallProgress` are pure and this is a new 235-line transport. Cover 200 with `errors[]`, 200 with an `ErrorResult`, a 413, and a non-JSON 200 body. There are 30 existing `.spec.ts` files in the dashboard, including `asset-bulk-actions.spec.ts` in this same directory.

### On the remaining CI failures

`conventional commit format` needs the subject fixed. The `deploy` failure is expected on PRs from forks and can be ignored. Shards 2 to 4, lint, unit tests, codegen and all four database e2e suites pass.

---

_Will approve once that's in._

HouseinIsProgramming:
@latifniz just let me know when you resolve them, and I will approve once the changes are adressed and CI is passing! Thanks!

latifniz:
ok let me take a closer look at all these feedbacks

latifniz:
Thanks for the thorough review, all 11 points addressed in the latest 2 commits:

1) Modal auto-closes on all-success; stays open if any file failed
2) __typename now checked, so a rejected file no longer shows as success
3) Non-JSON 200 responses treated as failures
4) Upload now goes through a shared api.uploadWithProgress() (fixes the missing credentials bug too)
5) Circular import broken. createAssetsDocument moved to its own module
6) Hardcoded strings translatable, <Plural> fixes "1 files", catalogs re-extracted
7) In-flight uploads actually abort on cancel/cleanup, added a Cancel button, added a timeout
8) Upload concurrency capped at 4. "queued" status now renders correctly
9) Completion callback invalidates by query key prefix plus the list view's key, skipped when nothing succeeded
10) Picker no longer opens a nested dialog when dropping files
11) Added unit tests for the pure helpers (17 tests)

Also fixed the PR title casing for commitlint. Should be green now, let me know if anything needs another pass.

HouseinIsProgramming (changes requested):
Good progress, all nine previous points are resolved and the api.ts refactor looks about right.

There are a couple of new open things left that my agents picked up, list in the comment below.

They are definetly worth looking into:

HouseinIsProgramming:
1. Check `signal.aborted` at the top of `uploadWithProgress` and resolve `ABORTED` immediately, and report the summary on the aborted path too.
An abort listener added to an already-aborted signal never fires, so everything still queued when the user cancels uploads to the server anyway. And files that finished before the cancel never refresh the gallery.

2. Return `INVALID_RESPONSE` from `parseUploadResponse` when `json.data` is missing, add a `.catch()` to the bare `uploadAll()` call, plus a spec case.
A 200 without `data` currently throws in `interpretUploadResult` as an unhandled rejection and the file is stuck on "uploading" forever.

3. Re-export `asset-documents.js` from `lib/index.ts`.
`createAssetsDocument` was public via the barrel, so `import { createAssetsDocument } from '@vendure/dashboard'` stops compiling.

4. Remove the default 60s `xhr.timeout` (make `timeoutMs` opt-in, no default).
It is a total-request deadline, so any upload slower than a minute fails where the old path had no deadline.

Will approve once that's in.

HouseinIsProgramming:
@latifniz There are some merge confilcts, please resolve them then I can approve of it

HouseinIsProgramming (approved):
Round 3 closes everything: cancel is checked before send, a 200 without `data` is reported, the `createAssetsDocument` export is back, and the upload deadline is gone. Verified the modal end to end at this head: per-file progress renders during upload and the dialog closes itself when done.

Approving. Note the branch has merge conflicts with `minor` now, needs a rebase before it can merge. The failing `unit tests (24.x)` is a `better-sqlite3` crash in core inherited from the base branch, not this PR.

latifniz:
it should be clean now

HouseinIsProgramming (approved):

PR #5099, opened 2026-08-07
fix(core): Bump file-type to ^21.3.1 (GHSA-5v7r-6r5c-r473)

HouseinIsProgramming:
## What

Bumps the `file-type` dependency from `^19.0.0` to `^21.3.1` in `@vendure/core` and `@vendure/asset-server-plugin`.

## Why

`file-type@19.6.0` is affected by [GHSA-5v7r-6r5c-r473](https://github.com/advisories/GHSA-5v7r-6r5c-r473): an infinite loop in the ASF parser on malformed input with a zero-size sub-header. Both packages call `fileTypeFromBuffer` on user-uploaded asset data, so the vulnerable path is reachable via asset upload. Patched in `file-type@21.3.1`.

The `fileTypeFromBuffer` API is unchanged across 19 to 21, so no call-site changes are needed. The repo already requires Node `^20.19.0 || >=22.12.0`, satisfying v21's Node >= 20 requirement.

One v21 breaking change is user-visible. Four MIME types were corrected to their formal IANA registrations:

| v19 | v21 |
| --- | --- |
| `audio/x-flac` | `audio/flac` |
| `video/x-matroska` | `video/matroska` |
| `application/x-apache-arrow` | `application/vnd.apache.arrow.file` |
| `application/x-parquet` | `application/vnd.apache.parquet` |

`getPermittedMimeTypeError` validates the content-detected MIME against `assetOptions.permittedFileTypes`. The defaults are wildcards (`image/*`, `video/*`, `audio/*`, `.pdf`) so they absorb the change, but anyone listing one of the old values explicitly will start rejecting valid uploads. Documented in the changelog.

Note: `@nestjs/common` pins `file-type@21.2.0` exactly, which is still inside the vulnerable range. That resolution is owned by NestJS and needs an upstream bump; it is not addressed here.

## Verification

- `packages/core` asset e2e suite: 26/26 passed (sqlite)
- `packages/asset-server-plugin` e2e suite: 35/35 passed (sqlite)
- `bun.lock` resolves `file-type` to 21.3.4 for both packages

Relates to OSS-692.

https://claude.ai/code/session_01PYg1vgCkngRoqhHRzUetgJ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+#### Security
+
+* **core** Bump `file-type` to `^21.3.1` to fix an infinite loop on malformed ASF input (#5099) [GHSA-5v7r-6r5c-r473](https://github.com/advisories/GHSA-5v7r-6r5c-r473)
+* **asset-server-plugin** Bump `file-type` to `^21.3.1` (#5099) [GHSA-5v7r-6r5c-r473](https://github.com/advisories/GHSA-5v7r-6r5c-r473)
+* **core** Note: `file-type` v21 renames four MIME types to their IANA registrations (`audio/x-flac` to `audio/flac`, `video/x-matroska` to `video/matroska`, `application/x-apache-arrow` to `application/vnd.apache.arrow.file`, `application/x-parquet` to `application/vnd.apache.parquet`). If you list any of the old values explicitly in `assetOptions.permittedFileTypes`, update them, otherwise those uploads will be rejected. The default wildcard config (`image/*`, `video/*`, `audio/*`, `.pdf`) is unaffected.
+
 ## <small>3.7.2 (2026-08-03)</small>
 
 #### Security
--- a/bun.lock
+++ b/bun.lock
@@ -59,7 +59,7 @@
     },
     "packages/admin-ui": {
       "name": "@vendure/admin-ui",
-      "version": "3.7.1",
+      "version": "3.7.2",
       "dependencies": {
         "@angular/animations": "^19.2.4",
         "@angular/cdk": "^19.2.7",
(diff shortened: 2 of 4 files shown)


biggamesmallworld (approved):
Verified locally:

- Advisory range checks out (introduced 13.0.0, fixed 21.3.1), so `^21.3.1` is the right bound.
- Reproduced the DoS: a crafted ASF header with a zero-size sub-object hangs `fileTypeFromBuffer` indefinitely on 19.6.0, and returns `undefined` immediately on 21.3.4. Reachable through `detectMimeTypeFromContents` in `asset.service.ts`, so this is worth landing.
- API signature is unchanged between 19 and 21, no call-site changes needed.

Nit: the PR body says the v20/v21 breaking changes don't affect usage. That's true for the default config, but not universally. Diffing the full MIME set between 19.6.0 and 21.3.4 turns up four renames:

| v19 | v21 |
| --- | --- |
| `audio/x-flac` | `audio/flac` |
| `video/x-matroska` | `video/matroska` |
| `application/x-apache-arrow` | `application/vnd.apache.arrow.file` |
| `application/x-parquet` | `application/vnd.apache.parquet` |

`getPermittedMimeTypeError` validates the content-detected MIME against `permittedFileTypes`. The defaults (`['image/*', 'video/*', 'audio/*', '.pdf']`) are wildcards so they absorb this, but anyone who configured an explicit `permittedFileTypes: ['audio/x-flac']` or `['video/x-matroska']` will silently start rejecting valid uploads after the bump.

Could you add a changelog note covering those renames? Nothing else blocking.

Non-blocking FYI: `@nestjs/common@11.1.11` pins `file-type@21.2.0` exactly, which is still in the vulnerable range, so audit scanners will keep flagging it. I checked and it's not reachable from our code (Nest only uses it in `FileTypeValidator`/`ParseFilePipe`, which we don't use), so it's a reporting artefact rather than an exposure. A `file-type` entry in the root `overrides` block would silence it if we want the audit clean, but that can be separate.

biggamesmallworld (approved):

Issue #4949, opened 2026-07-10
Flaky e2e tests on master CI: search-index race, scheduler/collection timing, worker-pool crash


michaelbromley:
## Summary

Auditing the last ~40 `master` runs of the **Build & Test** workflow (2026-06-24 → 2026-07-10) shows the e2e suite is intermittently red on roughly **1 in 4 master pushes**. The failures are overwhelmingly **flaky** (a single DB/Node job fails while the other 11 matrix cells pass), and cluster into a small number of root causes — most of them the same underlying disease: **assertions racing async background work**.

This issue captures the full tabulation so we can fix the causes systematically. There was also **one genuine (non-flaky) regression** in the window, already resolved, included for completeness.

### Scope

`bun run e2e` = `lerna run e2e` over 5 suites (core, asset-server-plugin, cli, graphiql-plugin, testing), matrixed across **DB {sqljs, mariadb, mysql, postgres} × Node {20, 22, 24}** = 12 job instances per push. Of ~35 completed master runs, 10 were red (1 was a non-e2e `dashboard i18n sync` check).

---

### A — Search-index race (`packages/core/e2e/default-search-plugin.e2e-spec.ts`) — dominant

Every failing test *does* call `awaitRunningJobs()` before asserting, but the helper (`packages/core/e2e/utils/await-running-jobs.ts`) races:

- It sleeps a fixed **100ms** for jobs to appear, then polls for jobs with `isSettled == false` until the count reaches 0.
- If the DB event → job enqueue takes **longer than 100ms** (real engines under CI load), the first poll sees **0 running jobs**, the loop exits immediately, and the assertion runs against a **stale index**.
- On its 5s timeout it only `console.log`s and continues, so the assertion runs against a **mid-reindex** snapshot.

Smoking gun: a partial price-range assertion `{ min: 5374 }` vs expected `{ min: 7896 }` — a half-applied index caught mid-write. Others show stale SKUs vs `*_updated`, and missing/extra channel IDs.

| Run (postgres/mysql/mariadb, Node) | # | Failing tests (`default-search-plugin > admin api > …`, line) |
|---|---|---|
| mysql 22 | 6 | ProductVariants changed (850), deleted (864), Collection created (1000), asset added to variant (1102 — `TypeError: null.id`), excludes deleted variants (1121), synthetic index removed (1334) |
| postgres 20 | 2 | Product deleted (913), Collection created (1000) |
| mariadb 24 | 1 | removing product variant from channel (1489) |
[...]

PR #5100, opened 2026-08-07
fix(core): Apply seller Order price adjustments in the seller's Channel

### LINKED ISSUE #4117: Channel-specific Order Promotion Leaks to Other Channels After Order Split (multivendor-plugin)
Environment
	•	Vendure version: 3.5 (tested on 3.5.2)
	•	multivendor-plugin: enabled
	•	Multi-channel: enabled
	•	Promotion type: Order discount
	•	Promotion scope: Channel-specific

⸻

❌ Issue Summary

When using multivendor-plugin, an Order Discount Promotion assigned to only one Channel is incorrectly applied to seller orders from other Channels after Vendure splits the order by vendor/channel.

This happens even though the promotion is configured for only one specific Channel.

⸻

🧪 Sample Order (Actual Case)

Parent Order Code:
GHPZYVMA9NFGDGAS

Vendure splits the order into two seller orders:
	1.	Seller Order: RWG5GS886AQHMR34
		•	Channel: Eazy
		•	Seller: Eazy Seller
		•	❌ Promotion applied (INCORRECT)
	2.	Seller Order: CUJ7V5AMW9KNHBSF
		•	Channel: Brand Collection
		•	Seller: Brand Collection Seller
		•	✅ Promotion should apply (CORRECT)

📌 Promotion configuration:
	•	Condition: minimum_order_amount (1.00$)
	•	Action: products_percentage_discount (10%)
	•	Assigned only to Channel: Brand Collection
	•	Not assigned to Eazy channel

⸻

✅ Expected Behavior
	1	Order discount promotion should apply only to
		•	Orders created in Brand Collection channel
	2	After order splitting:
		•	Only seller orders belonging to Brand Collection should receive the discount
		•	Seller orders from Eazy channel must NOT receive the promotion

⸻

❌ Actual Behavior
	1	After Vendure splits the order:
		•	Both seller orders receive the promotion
		•	The seller order under Eazy channel incorrectly gets the discount

This indicates that channel scoping is ignored during promotion evaluation after order splitting.

⸻

🔍 Technical Observation
	•	Promotion evaluation appears to run before or without validating order.channelId
	•	OrderLine / split seller orders do not re-check channel constraints
	•	Without multivendor-plugin, channel-restricted promotions behave correctly

⸻

💥 Business Impact
	•	Channel-exclusive promotions cannot be enforced
	•	Sellers receive unintended discounts
	•	Incorrect settlement and commission calculation
	•	Critical issue for real multi-vendor, multi-channel marketplaces

⸻

🛠 Suggested Fix
	•	During promotion evaluation (especially post-split):
	•	Validate promotion.channels against each seller order’s channelId
	•	Ensure multivendor-plugin does not bypass Vendure’s core channel guards
	•	Add regression test:
	•	Multi-vendor + multi-channel + order-level promotion

⸻

biggamesmallworld:
## Description

Fixes #4117.

`OrderSplitter.createSellerOrders` called `applyPriceAdjustments` with the aggregate Order's `RequestContext`. Promotions are resolved via `getActivePromotionsInChannel`, which filters on `ctx.channelId`, so every seller Order was priced against the *aggregate* Channel's Promotions. A Promotion assigned to a single seller Channel was therefore applied to seller Orders belonging to every other Channel.

### Root cause

- `order-splitter.ts` → `applyPriceAdjustments(ctx, sellerOrder)` with the aggregate `ctx`
- → `order.service.ts` → `getActivePromotionsInChannel(ctx)`
- → `promotion.service.ts` → `WHERE channel.id = :channelId` using `ctx.channelId`

The `adjustments` copied by `duplicateOrderLine` are *not* the cause. `OrderCalculator.applyOrderItemPromotions` clears and recalculates them. It is the promotion list that is wrong.

### Fix

Price adjustments are applied in a `RequestContext` scoped to the seller Channel. Everything other than the Channel is deliberately carried over:

- **`session`** — the built-in Promotion conditions do not read it (`customer-group-condition.ts` reads `order.customer`), but plugin-supplied conditions and strategies may, and dropping it would silently change their behaviour on seller Orders.
- **`currencyCode`** — otherwise a seller Channel with a different default currency trips the `ctx.currencyCode !== order.currencyCode` branch in `applyPriceAdjustments` and re-prices every line.
- **`languageCode`** — the seller Order is a copy of the customer's Order, so it keeps the language the customer ordered in rather than adopting the seller Channel's default.
- **the transaction manager** — `RequestContextService.create()` would drop the `TRANSACTION_MANAGER_KEY` symbol and move the writes outside the transaction.

The Channel is re-fetched via `ChannelService.findOne` because the Channel stored on the seller Order is a stub carrying only an id (`new Channel({ id })`), and `defaultTaxZone` is required to calculate the Order.

The multivendor example plugin made the same mistake when re-applying price adjustments after adding its platform fee surcharge, which overwrote the corrected result. The core fix alone does not fix the reported symptom, so the example plugin is fixed alongside.

### Shipping

Scoping the context to the seller Channel also scopes `ShippingMethodService.findOne`, which is Channel-scoped. A ShippingMethod assigned only to the default Channel was therefore not found for the seller Order, `OrderCalculator.applyShipping` dropped the ShippingLine, and `applyChangesToShippingLines` persisted the removal. No error, no log, and the seller Order silently lost its shipping charge.

`OrderService.applyPriceAdjustments` now takes a `recalculateShipping` option, which `OrderSplitter` and the example plugin pass as `false`. Seller Orders inherit their ShippingLines from the aggregate Order, where they were already calculated against the ShippingMethod, so there is nothing to recalculate.

The option gates `applyShipping` only. Shipping Promotions are still re-applied in the seller Channel, because `applyShippingPromotions` is the only caller of `ShippingLine.clearAdjustments()` and `duplicateShippingLine` copies the aggregate ShippingLine's `adjustments`. Skipping it would leave an aggregate-Channel shipping discount on every seller Order and make a seller-Channel shipping Promotion impossible to apply, which is the same leak as #4117 moved from OrderLines to ShippingLines.

### Scope

Order splitting only happens when an `OrderSellerStrategy` implements `splitOrder`. `DefaultOrderSellerStrategy` does not, so single-vendor installations never enter this path and are unaffected.

## Behaviour changes for the changelog

All of these apply only to split Orders, i.e. installations with an `OrderSellerStrategy` that implements `splitOrder`. Aggregate and seller Order totals can now diverge where they previously agreed, which matters for payout reconciliation.

Seller Orders are now calculated against the seller Channel rather than the aggregate Channel. That changes four things:

1. **Promotions.** A Promotion assigned only to the default Channel no longer applies to seller Orders. This is the fix.
2. **Tax zone.** Tax is computed against the seller Channel's `defaultTaxZone` (`order-calculator.ts:65-68`). For setups whose seller Channels share the default Channel's tax zone (as the multivendor example plugin creates them) this is a no-op. For setups with per-seller tax zones, tax on split Orders will change.
3. **`pricesIncludeTax`.** `order_percentage_discount`, `order_fixed_discount`, `product_percentage_discount`, `facet_values_percentage_discount` and `buy_x_get_y_free` all read `ctx.channel.pricesIncludeTax`, which decides whether they compute against `subTotal` or `subTotalWithTax`. A seller Channel configured differently from the default Channel will now discount against the other base.
4. **Shipping Promotions.** These are now resolved in the seller Channel too. ShippingLine *prices* are not recalculated, so a ShippingMethod which exists only in the default Channel keeps working.

## Testing

Two new e2e specs.

### `order-multi-vendor-promotion.e2e-spec.ts` — the Promotion leak

The spec exercises the same code path a real multivendor storefront does, over the wire:

- **A real server is booted.** `TestServer.bootstrap()` (`packages/testing/src/test-server.ts:53`) starts an actual Vendure application, not a stubbed module or an in-process mock of the ORM.
- **Requests go over real HTTP.** `SimpleGraphQLClient.fetch()` (`packages/testing/src/simple-graphql-client.ts:111`) issues a genuine `fetch()` against the Shop and Admin API URLs. Nothing about `OrderSplitter`, `OrderCalculator` or `PromotionService` is mocked or spied on.
- **The split is triggered by a genuine checkout,** not by calling `OrderSplitter` directly:

  ```
  registerNewSeller           (shop)   ×2
  assignProductsToChannel     (admin)  ×2
  createPromotion             (admin, channel-token header = seller A's channel)
  addItemToOrder              (shop)   ×2   — one variant from each seller
  setOrderShippingAddress     (shop)
  setOrderShippingMethod      (shop)
  transitionOrderToState      (shop)   → ArrangingPayment
  addPaymentToOrder           (shop)   → payment settles, split fires here
  order(id:)                  (admin)  → read back each seller Order
  ```

- **The assertion reads the public API response,** not internal state. Actual response bodies observed, before and after the fix:

  **Before** — Bob's Promotion has leaked onto Alice's Order:
  ```json
  [
    { "channels": ["__default_channel__", "bobs-parts"],
      "discounts": [{ "description": "bobs 10% off", "amount": -11581 }] },
    { "channels": ["__default_channel__", "alices-wares"],
      "discounts": [{ "description": "bobs 10% off", "amount": -22285 }] }
  ]
  ```

  **After** — the Promotion stays in its own Channel:
  ```json
  [
    { "channels": ["__default_channel__", "bobs-parts"],
      "discounts": [{ "description": "bobs 10% off", "amount": -11581 }] },
    { "channels": ["__default_channel__", "alices-wares"],
      "discounts": [] }
  ]
  ```

  The discount is asserted by exact `amount` and `amountWithTax`, and against Bob's own lines, not just by description. The values are golden values captured from a verified run; they are under a flat 10% of the line total because `subTotalWithTax` also carries the platform fee surcharge.

- **It cannot pass for the wrong reason.** Before checking out, the spec asserts the Promotion really is scoped to Bob's Channel: present in `promotions` when queried with Bob's channel token, absent with Alice's. So a green result cannot be explained by the Promotion never having existed or being mis-assigned.

### `order-split-shipping.e2e-spec.ts` — the shipping behaviour

Uses its own minimal `OrderSellerStrategy` rather than the multivendor example plugin, so it exercises core behaviour only. The example plugin assigns every ShippingMethod to its seller Channel (`mv.service.ts:95`), which is exactly the case that hides this bug.

- Asserts up front that the seed ShippingMethods exist in the default Channel and *not* in the seller Channel, so the spec cannot pass for the wrong reason.
- Asserts the seller Order's `shippingWithTax` equals the aggregate Order's exactly, rather than merely being non-zero. A regression that halves the price or double-applies tax fails.
- Adds a free-shipping Promotion assigned to the default Channel only, asserts it does discount the aggregate Order to 0, then asserts the seller Order keeps the full shipping price.

### Before/after

The specs fail on `master` and pass with this change. Reverting **either** changed file individually also reproduces the promotion failure:

| State | Result |
|---|---|
| Both fixes applied | passes |
| `order-splitter.ts` reverted | fails, aggregate Channel's Promotions used |
| `mv-order-seller-strategy.ts` reverted | fails, plugin overwrites the corrected result |

Also ran the `order-multi-vendor`, `promotion` and `order-promotion` e2e suites. (`order-promotion` is intermittently flaky per #4949; it failed once on the customer group test and passed on re-run, unrelated to this change.)

Kept as separate specs rather than folded into `order-multi-vendor.e2e-spec.ts`, since that suite is sequential and stateful and introducing a Promotion mid-flow shifts its existing total assertions.

### What these tests do not cover

Every real multivendor deployment supplies its **own** `OrderSellerStrategy`. These specs prove the fix for core and for the bundled example plugin. A third-party strategy which calls `applyPriceAdjustments` itself, exactly as the example plugin did, will still leak Promotions after upgrading, and these tests would not catch it.

A supported helper for deriving a seller-Channel-scoped `RequestContext` would stop strategy authors making this mistake. Per review, that lands on `minor` rather than here.

## Checklist

- [x] I have added tests covering the fix


grolmus (commented):
Reviewed this closely since it's order-pricing / promotion / tax code. The fix is sound and correctly targeted: scoping the `RequestContext` to the seller Channel fixes promotion resolution, tax zone and shipping in one move, all of which resolve channel-scoped off `ctx` in `applyPriceAdjustments`. I traced a few things that would be silent money bugs if wrong and they hold up:

- `ctx.copy()` + reassigning `_channel` mutates only the copy (`copy()` is a shallow `Object.assign`), so the aggregate `ctx` is unaffected.
- The `channelService.findOne` re-fetch is necessary, not defensive noise — `sellerOrder.channels` holds a `new Channel({ id })` stub with no `defaultTaxZone`, which the calculator requires.
- The `(ctx as any)._channel =` cast looks like a smell but is consistent with the existing `updateOrderCurrency` precedent (`(ctx as any)._currencyCode =`), for the same reason: `RequestContextService.create()` would drop the transaction manager and session. Pragmatic and correct here.

The e2e is genuinely strong — real server, over the wire, exact-value assertions, and it guards against a false pass by proving the promotion is absent from the other seller's Channel.

### Should fix before merge

1. **Throw-vs-silent-fallback inconsistency.** `order-splitter.ts` returns the original (aggregate) `ctx` when the channel isn't found, whereas `mv-order-seller-strategy.ts` throws `InternalServerError` for the same case. `SplitOrderContents.channelId` is a required `ID` sourced from the order lines' seller channels, so a missing channel is a "can't happen" — and silently falling back to the aggregate `ctx` reintroduces exactly the promotion-leak / wrong-tax-zone bug this PR fixes, with no signal. Prefer throwing (consistent with the sibling code), or drop the branch.
2. **The e2e asserts the discount description but not the amount.** `expect(bobs.discounts.map(d => d.description)).toEqual(['bobs 10% off'])` — `amount` is already fetched but discarded. The whole point is that the 10% is computed against Bob's *seller-order* subtotal; a regression applying the right promotion against the wrong base would pass this test unchanged. An exact `amount` assertion is the one that actually protects the money.

### Unrelated churn — please drop

`list-page.mdx`, `promotion-service.mdx` and `manifest.json` are a stray doc-regen: the list-page changes are the `transformQueryKey`/view-option docs from #5066, `promotion-service.mdx` reflects a `roleService` param already on `master`, and `manifest.json` only bumps timestamps. None of it relates to seller-order pricing — dropping it keeps the diff honest.

### Notes (non-blocking)

- The tax-zone behaviour change you flagged is the right call to surface. Since the title is `fix(core):`, it will land in the changelog — worth a one-line changelog/note so multi-vendor installs with per-seller tax zones know split-order tax may change.
- The `(ctx as any)._channel` pattern now appears in three places, which signals a missing first-class `ctx.withChannel(channel)` API — not this PR's job, but plugin authors currently have to reach into private fields to channel-scope a context.

Labelling this `T2: Elevated risk` — it's small but sits squarely in silently-wrong-money territory.

HouseinIsProgramming:
The fix swaps which shop (Channel) the pricing code looks at. That fixes discounts leaking, but the same switch also silently changes tax and shipping, and nobody checked those.

### The bug being fixed

When one customer order gets split into separate orders per seller, each seller's order was being priced using the **main shop's** settings. So a discount that belonged to seller A got applied to seller B's order too. Clearly wrong.

### The fix

Price each seller's order using **that seller's own shop** settings. Right idea.

### Why I'd hold it

"That seller's shop" doesn't only decide which discounts exist. It also decides which tax rate applies, and which shipping methods exist.

The shipping one is the problem:

**If the shipping method belongs to the main shop and not the seller's shop, the code can't find it, deletes the shipping charge off the seller's order, and saves that deletion. No error, no log.**

Before this change it always found it, because it was looking at the main shop.

The test here doesn't catch this because the example plugin happens to give every seller its own shipping method. A real store that shares one shipping method across sellers would just lose money quietly.

I traced this in the code but have not reproduced it, so I am asking for a test rather than asserting it is broken.

### What I'm asking for

One test: shipping method on the main shop only, does the seller's order keep its shipping charge? If yes, fine, merge. If no, that needs fixing first.

The other four items are small: make a silent failure throw instead, fix a comment that states a wrong reason, check the discount amount and not just its name, and write the tax and shipping change in the changelog so users aren't surprised. Details in the next comment.

HouseinIsProgramming:
Requested changes:

1. Add an e2e case where the ShippingMethod is assigned to the default Channel only, and assert the seller Order keeps its ShippingLine with a non-zero price.
Chain: `order-calculator.ts:104` always runs `applyShipping` because `OrderService.applyPriceAdjustments` passes no `options`. `order-calculator.ts:316-321` drops the ShippingLine when `shippingMethodService.findOne` returns `undefined`, and that lookup is channel-scoped at `shipping-method.service.ts:97-101`. The removal is persisted by `applyChangesToShippingLines` at `order.service.ts:2420-2436`. The current fixture cannot reach it because `mv.service.ts:95` assigns every ShippingMethod to its seller Channel.
If it passes, say so and this item drops. If it fails, it is a regression this PR introduces and we decide the resolution, not you: either keep shipping resolution on the aggregate context, or accept the new scoping and document it as a requirement that ShippingMethods on split Orders be assigned to each seller Channel.

2. Change `if (!sellerChannel) { return ctx; }` at `order-splitter.ts:104-107` to throw `InternalServerError`.
Returning the aggregate context silently reinstates the leak this PR fixes. Your copy of the same logic already throws at `mv-order-seller-strategy.ts:120-122`.

3. Fix the `session` clause in the doc comment at `order-splitter.ts:97`.
It claims Customer-dependent Promotion conditions need it. They do not: `customer-group-condition.ts` reads `order.customer`, and nothing in this path reads `ctx.session`. Preserving the session is still right for plugin-supplied conditions, so give that reason. The `currencyCode` and transaction clauses are accurate, leave them.

4. Assert the discount `amount` in `order-multi-vendor-promotion.e2e-spec.ts`, not just `description`.
`GET_ORDER_DISCOUNTS` already selects it. As written, a discount computed against the aggregate subtotal instead of Bob's seller subtotal still passes.

5. Replace `expect(createPromotion.id).toBeDefined()` with an assertion on `createPromotion.name`.
An `ErrorResult` currently fails with an unhelpful message instead of the actual error.

6. Add a changelog entry covering the full scope.
Seller Orders are now calculated against the seller Channel's tax zone (`order-calculator.ts:65-68`) and ShippingMethods, and Promotions assigned only to the default Channel no longer apply to them. Aggregate and seller Order totals can therefore diverge, which matters for payout reconciliation.

Two things you do not need to chase:

The `(sellerCtx as any)._channel` cast is fine on `master`. `RequestContextService.create()` drops the `TRANSACTION_MANAGER_KEY` symbol, so it would move the writes outside the transaction, and there is precedent at `order.service.ts:595-596`. We will add a supported channel-scoping helper on `minor`.

No need to fold the spec into `order-multi-vendor.e2e-spec.ts`. Your reasoning holds: a Promotion mid-flow would shift that suite's total assertions.

Will approve once that's in.

grolmus (approved):
Both points from the earlier pass are addressed, and CI is green across all four DBs:

- **Failure handling** — `createSellerChannelContext` now throws `InternalServerError` instead of silently returning the aggregate `ctx`, consistent with the mv-strategy. No more silent fallback to the pre-fix behaviour.
- **Money assertion** — the e2e now pins the exact discount (`amount` / `amountWithTax`) and asserts it lands on Bob's own lines (`linePriceWithTax - proratedLinePriceWithTax`), with a comment explaining why it isn't a flat 10% (the platform-fee surcharge). That's the assertion that actually protects the money, and it passes on sqljs/mariadb/mysql/postgres.

Non-blocking: the regenerated `docs/` files (`list-page.mdx`, `promotion-service.mdx`, `manifest.json`) are still unrelated to seller-order pricing — fine to leave, but dropping them would keep the diff on-topic.

Still worth a changelog line for the tax-zone behaviour change you flagged (split-order tax now computed against the seller Channel's zone), so multi-vendor installs with per-seller tax zones are aware.

LGTM on the code. 👍

biggamesmallworld:
@HouseinIsProgramming the new test I added should be reflective of the one channel, one seller state and flag any regression. Let me know when you have a chance to take a look again!

grolmus (dismissed):
Thanks for the continued work here — this is careful, and the tests are genuinely excellent: exact assertions with deliberate false-pass guards on both the promotion-scoping and the shipping-line cases (proving the promotion is Channel-scoped / the ShippingMethod really is absent before asserting the outcome). I re-verified the core mechanism: the new `recalculateShippingPromotions` flag is backward-compatible across every `applyPriceAdjustments` caller (no options → unchanged; `recalculateShipping: false` alone → unchanged; the split path → shipping held, shipping-promotions revalidated standalone, which is safe since `applyShippingPromotions` only reads existing `shippingLines`), and the transaction manager survives `ctx.copy()`, so the seller-Channel calc and its saves stay in the same transaction. CI is green across all DBs.

One correctness gap I'd like addressed before merge:

**Shipping tax is left in the aggregate Channel's tax zone.** With `recalculateShipping: false`, `applyShipping` is skipped, so the seller Order's `shippingLine.taxLines` remain the ones duplicated from the aggregate Order (`order-splitter` picks `taxLines`). But line-item taxes *are* recalculated for the seller Channel's zone — `order-calculator` sets `order.taxZoneId` to the seller zone and re-runs `applyTaxes` on the lines — and `applyShippingPromotions` only touches adjustments, not tax rates. So a seller Channel that resolves to a *different* tax zone than the aggregate Channel ends up with line items taxed in the seller zone but shipping taxed in the aggregate zone: internally inconsistent totals. It's invisible when seller Channels share the default tax zone, which is exactly why the tests don't catch it. Please at least document this limitation in the `applyPriceAdjustments` doc comment and the splitter comment; ideally re-derive the shipping tax rate for the seller zone even when the shipping *price* is held constant.

Non-blocking:

- **Scope:** the dashboard Vite JSX change (`compiler.ts` + `jsx-compilation.spec.ts` + fixtures) is unrelated to seller-Order pricing. It looks reasonable on its own, but it would be cleaner — and safer to review — as its own PR.
- **Maintainability:** `(ctx as any)._channel = ...` is now load-bearing in the price engine and duplicated (`order-splitter`, `mv-order-seller-strategy`, and the sibling `_currencyCode` poke in `updateOrderCurrency`). Consider promoting it to a typed `RequestContext.copyWithChannel(channel)` helper so a future refactor of the private field can't silently break the money paths.

Thanks again — the promotion-leak fix and the shipping-line preservation are solid improvements.

biggamesmallworld:
Thanks, the tax point is real and I've documented it. One correction on the mechanism, because it changes what "re-derive the shipping tax rate for the seller zone" would mean.

**Shipping tax does not come from the tax zone.** `ShippingCalculator.calculate()` returns the `taxRate` directly, and `defaultShippingCalculator` reads it from a ShippingMethod arg:

```ts
// packages/core/src/config/shipping-method/default-shipping-calculator.ts
taxRate: {
    type: 'float',
    defaultValue: 0,
    ...
},
calculate: (ctx, order, args) => ({
    price: args.rate,
    taxRate: args.taxRate,
    priceIncludesTax: getPriceIncludesTax(ctx, args.includesTax as any),
}),
```

`applyShipping` then writes that straight onto the ShippingLine as `{ description: 'shipping tax', taxRate: result.taxRate }`. The tax zone never enters it. So with the built-in calculator there is no seller-zone shipping rate to re-derive: the rate is a property of the ShippingMethod, and it is the same number whichever Channel you price in.

What *is* Channel-dependent is `priceIncludesTax` under `includesTax: 'auto'`, which reads `ctx.channel.pricesIncludeTax`. A seller Channel configured differently from the aggregate Channel does inherit the wrong basis there. A custom ShippingCalculator which resolves a TaxRate from the RequestContext would have the same problem, and more visibly.

**Re-deriving it isn't available to us here.** Getting a tax rate means running the ShippingMethod's calculator, which means resolving the ShippingMethod in the seller `ctx`:

```ts
const currentShippingMethod =
    shippingLine?.shippingMethodId &&
    (await this.shippingMethodService.findOne(ctx, shippingLine.shippingMethodId));
if (!currentShippingMethod) {
    order.shippingLines = order.shippingLines.filter(sl => sl !== shippingLine);
    continue;
}
```

That lookup is Channel-scoped, and its failing is precisely why we hold the prices: a ShippingMethod not assigned to the seller Channel would drop the ShippingLine off the seller Order entirely. So any path that recovers the tax rate also reintroduces the bug this PR is fixing. Recalculating tax but not price would need `applyShipping` split into a price half and a tax half, with the tax half tolerating an unresolvable method, and I don't think that belongs in a bugfix PR.

So I've documented it in both places, in 1f1821acb: the `applyPriceAdjustments` doc comment now states that `taxLines` and `listPriceIncludesTax` are held along with the prices and what that implies for a Channel on a different tax zone or `pricesIncludeTax` setting, and the splitter comment records why recalculating shipping tax is not possible there. Happy to raise a follow-up issue for the `includesTax: 'auto'` case if you think it warrants one.

On the other two:

- **Scope:** done. The compiler change, its test and the fixtures are out of this branch (9982d3016) and now sit on `fix/dashboard-jsx-compilation`. They were swept in by mistake, you were right to flag it twice.
- **`copyWithChannel`:** agreed in principle, but it adds public API to `RequestContext` and would want to absorb the `_currencyCode` poke in `updateOrderCurrency` too, so it reaches beyond this fix. I'd rather do it as its own PR than grow this one further. Say the word if you'd prefer it here.

grolmus (approved):
Thanks for the thorough follow-up — all my blocking points are resolved:

- **Shipping tax basis**: documented in both `applyPriceAdjustments` and the splitter call site, and you went further and bounded the real-world impact (the built-in `defaultShippingCalculator` takes its rate from a ShippingMethod arg, so this only bites `includesTax: 'auto'` and custom calculators that derive tax from the RequestContext). That's exactly the context a future reader needs.
- **Scope**: the dashboard Vite JSX change is out of the branch now — this PR is back to a single concern.
- The two new `order-calculator.spec.ts` cases (existing adjustments preserved vs. revalidated) pin the `recalculateShippingPromotions` flag nicely.

I re-verified the flag's backward-compat across every `applyPriceAdjustments` caller and the transaction propagation through `ctx.copy()` in the earlier pass; nothing changed there. Approving.

Two non-blocking notes before merge:
- `e2e (mysql)` is red, but it ran ~32m (3× the others) with no test failure in the log while mariadb/postgres/sqljs are all green — looks like a runner hang rather than a real failure. Worth a re-run to confirm green before merging.
- The `(ctx as any)._channel = …` poke is still duplicated across the price paths; a typed `RequestContext.copyWithChannel()` helper would be nicer, but that's a cleanup for another day, not a blocker.

grolmus (approved):
Re-approving — the only change since is a one-line comment clarification. LGTM.

PR #5101, opened 2026-08-07
fix(core): Prevent NaN discount amount on cancelled modified OrderLine

### LINKED ISSUE #5097: Money cannot represent non numeric value: NaN when viewing cancelled order with post-placement line modifications
**Describe the bug**

When an order line is added via order modification (after order placement), and the order is then cancelled, viewing the order in the admin dashboard throws `GraphQL Request Error: Money cannot represent non numeric value: NaN`.

The root cause is that the private helper methods in `OrderLine` entity (`_proratedUnitPrice()`, `_discountedUnitPrice()`, etc.) divide adjustment amounts by `quantity`. When `quantity === 0` (cancelled order), this produces `NaN` which the GraphQL `Money` scalar cannot serialize.

**To Reproduce**

1. Create an order with products and complete payment (COD or card)
2. Modify the order via admin to **add a new product line**
3. Complete the modification with payment
4. **Cancel the entire order**
5. Try to view the order in admin dashboard
6. See error: `Money cannot represent non numeric value: NaN`

**Expected behavior**

The cancelled order should display correctly in the admin dashboard with all line prices showing as 0.

**Actual behavior**

GraphQL error is thrown and the order cannot be viewed.

**Error logs**

```
GraphQL Request Error: Money cannot represent non numeric value: NaN
```
**Environment:**

- Nodejs version: 20.x
- Database: PostgreSQL
- Operating System: Linux
- Package manager: npm

**Root cause analysis**

The issue occurs specifically for order lines added via modification that have:
- `orderPlacedQuantity = 0` (line was added after order placement)
- `quantity = 0` (set to 0 when order is cancelled)

The computed getters like `proratedUnitPrice` call private methods that calculate `adjustment.amount / quantity`, causing division by zero → NaN.

PR #3009 fixed a similar issue in `OrderCalculator` but did not address the entity getters themselves.

**Proposed fix**

In `packages/core/src/entity/order-line/order-line.entity.ts`, the private helper methods should guard against `quantity === 0`:

```typescript
private _proratedUnitPrice(): number {
    if (this.quantity === 0) {
        return 0;
    }
    // existing calculation...
}
```
Same for `_proratedUnitPriceWithTax()`, `_discountedUnitPrice()`, `_discountedUnitPriceWithTax()`.

**Workaround**

Manually delete the problematic order line from the database:
```sql
DELETE FROM order_line_reference WHERE "orderLineId" = <line_id>;
DELETE FROM order_line WHERE id = <line_id>;
```
**Additional context**

- Related to #3008 which was fixed in PR #3009, but that fix only addressed the `OrderCalculator`, not the `OrderLine` entity getters
- This specifically affects lines added via modification (`orderPlacedQuantity = 0`) that are then cancelled

TheHypnoo:
# Description

Fixes #5097.

Viewing a cancelled order in the Admin dashboard fails with `Money cannot represent non numeric value: NaN` when the order contains a line that was added by an OrderModification after the order was placed.

## Cause

`OrderLine.discounts` prorates each adjustment over the quantity the line was placed at:

```ts
const unitAdjustmentAmount =
    (adjustment.amount / Math.max(this.orderPlacedQuantity, this.quantity)) * this.quantity;
```

`orderPlacedQuantity` is written once, on the transition into a placed state (`default-order-process.ts`), so a line added by a later modification keeps it at 0. Cancelling then sets `quantity` to 0 as well (`OrderModifier.cancelOrderByOrderLines`), which leaves both operands of the `Math.max` at zero. `amount / 0 * 0` is `NaN`, and the `Money` scalar refuses to serialize it. `Order.discounts` aggregates `line.discounts`, so the order-level field fails for the same reason.

The loop only runs if the line carries an adjustment, so this needs a promotion to have been applied to the order. Lines that were part of the original order are unaffected: their `orderPlacedQuantity` survives the cancellation and keeps the divisor non-zero.

The issue suggests guarding `_proratedUnitPrice()` and the other private helpers. Those are already safe — they go through `getUnitAdjustmentsTotal()`, which returns 0 when `quantity === 0`, and `getLineAdjustmentsTotal()` carries the same guard. `discounts` is the only place that divides without one. #3009 fixed the same symptom in `OrderCalculator`, which is a different code path.

## Fix

Lift the divisor out of the loop and return 0 when there are no units to prorate over. Everything else is unchanged: with a non-zero divisor the expression is identical to before, and a cancelled line that was part of the original order still reports 0 because it is multiplied by `this.quantity`.

## Tests

`order-line.entity.spec.ts` is new. It covers the getter directly: one case for the normal proration, one for a line with both quantities at zero. Without the fix the second one fails with `expected NaN not to be NaN`.

The e2e in `order-modification.e2e-spec.ts` walks the flow from the report — place an order with an order-level promotion, add a line via a modification, settle the additional payment, cancel the whole order, then read it back with the per-line discounts the dashboard asks for. Without the fix it fails with the reported error rather than an assertion, which is the point of it. The shared `OrderWithLines` fragment does not select `lines.discounts`, so the test uses its own document.

Verified green: 1139 core unit tests, and all 71 tests in `order-modification.e2e-spec.ts`.

# Breaking changes

None. A field that used to serialize as `NaN` — and in practice killed the whole query — now serializes as 0.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5101"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788687552&installation_model_id=20193&pr_number=5101&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5101&signature=3ab07451d639bbba15419a83b6293b0dac6aacb64ec3bfbdfd540d05e97246fa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


grolmus (approved):
Really clean fix, and good diagnosis — nice catch that the issue's proposed guard location (`_proratedUnitPrice()` et al.) was already safe via `getUnitAdjustmentsTotal()`, and the real divide-by-zero is only in `discounts`. Lifting the divisor out of the loop and returning 0 when there are no units to prorate is behaviour-preserving for the normal path, and correct for the edge case (a zero-quantity line's prorated discount is 0 anyway).

Tests are exactly what this needs: the unit spec pins both the normal proration and the zero-quantity case (NaN → 0), and the e2e walks the full flow from the report (place → modify-add-line → settle → cancel → read back per-line discounts). CI green.

LGTM 🎉

michaelbromley (approved):

michaelbromley:
Thanks for the fix!

PR #5102, opened 2026-08-07
fix(asset-server-plugin): harden SVG asset serving headers (GHSA-f4r3)

grolmus:
## Summary

Hardens the response headers on assets served by the asset-server-plugin so that a permitted-but-scriptable asset (SVG in particular) cannot execute script when opened directly or embedded. Addresses **GHSA-f4r3-h6jf-4m29** (OSS-681).

## Change

A shared `setAssetSecurityHeaders` helper, applied on **both** serve paths (`sendAsset` and the on-the-fly `generateTransformedImage`), sets:

- `X-Content-Type-Options: nosniff` on every asset — stops the browser MIME-sniffing a response into a more dangerous type than its declared `Content-Type`.
- `Content-Security-Policy: default-src 'none'; script-src 'none'; sandbox` (tightened from `default-src 'self'`) — denies scripts/subresources and sandboxes the document.
- `Content-Disposition: attachment` for markup mime types (`image/svg+xml` and other `*+xml`, `text/html`, `text/xml`, `application/xml`) — forces download rather than inline rendering for the types that can carry executable script.

`` previews are unaffected — the header is ignored for subresource loads and scripts in SVG never execute inside ``.

## Behaviour change (please confirm)

The `sandbox` directive applies to **all** assets, so opening a **PDF** URL directly in a Chromium-based browser downloads it instead of showing it in the built-in viewer, and inline `<embed>`/`<object>` PDF rendering is blocked. This follows the approach specified on OSS-681 and is deliberate hardening, but it is a visible change for setups that relied on inline PDF viewing — flagging it for the team, and it likely deserves a changelog note. Images and `` previews are unaffected.

Dropping SVG from the default `permittedFileTypes` (the advisory's other recommendation) is breaking, so it's intentionally out of scope here (would target `major`).

## Test plan

`asset-server-plugin` e2e (`asset-server-plugin.e2e-spec.ts`), against a real booted server over HTTP — 39 tests, incl. 4 new under `security headers`:

- SVG source → `Content-Disposition: attachment` + `nosniff` + the locked-down CSP (the actual GHSA vector, `*+xml` branch).
- HTML source → `attachment` + `nosniff` + CSP (the `text/html` markup branch; the test config permits `text/html` uploads to exercise it).
- JPEG source → **no** `Content-Disposition`, but still `nosniff` + CSP (non-markup negative case).
- Transformed image → `nosniff` + CSP (covers the `generateTransformedImage` serve path).

```
✓ asset-server-plugin.e2e-spec.ts (39 tests)
eslint: clean · tsc: clean (changed files)
```

Header assertions can't verify the browser-rendering effect of `sandbox` (e.g. the PDF-download behaviour above) — that's inherent to `fetch()`-level testing and is documented in the code comment.

Relates to OSS-681 (part of the security-hardening epic).

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5102"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788697081&installation_model_id=20193&pr_number=5102&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5102&signature=c42c2b1ef81c9888687d308ebc5396f21a681a9460657ced7c3d9bec6d83a1b7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
+++ b/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
@@ -40,6 +40,11 @@ describe('AssetServerPlugin', () => {
     const { server, adminClient } = createTestEnvironment(
         mergeConfig(testConfig(), {
             // logger: new DefaultLogger({ level: LogLevel.Info }),
+            // Permit text/html uploads so the security-headers tests can exercise the non-SVG
+            // markup branch; the defaults (image/* etc.) would reject it.
+            assetOptions: {
+                permittedFileTypes: ['image/*', 'video/*', 'audio/*', '.pdf', 'text/html'],
+            },
             plugins: [
                 AssetServerPlugin.init({
                     assetUploadDir: path.join(__dirname, TEST_ASSET_DIR),
@@ -118,6 +123,15 @@ describe('AssetServerPlugin', () => {
         it('serves an SVG source as an attachment with nosniff and a locked-down CSP', async () => {
             const svgAsset = await uploadAsset('test.svg');
             const res = await fetch(svgAsset.source);
+
+            expect(res.headers.get('content-disposition')).toContain('attachment');
+            expect(res.headers.get('x-content-type-options')).toBe('nosniff');
+            expect(res.headers.get('content-security-policy')).toBe(CSP);
+        });
+
+        it('serves an HTML asset as an attachment (non-SVG markup branch)', async () => {
+            const htmlAsset = await uploadAsset('test.html');
+            const res = await fetch(htmlAsset.source);
 
             expect(res.headers.get('content-disposition')).toContain('attachment');
             expect(res.headers.get('x-content-type-options')).toBe('nosniff');
--- a/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
+++ b/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
@@ -40,10 +40,10 @@ describe('AssetServerPlugin', () => {
     const { server, adminClient } = createTestEnvironment(
         mergeConfig(testConfig(), {
             // logger: new DefaultLogger({ level: LogLevel.Info }),
-            // Permit text/html uploads so the security-headers tests can exercise the non-SVG
-            // markup branch; the defaults (image/* etc.) would reject it.
+            // The types these tests actually upload. `text/html` is added on top of the usual
+            // image/pdf set so the security-headers tests can exercise the non-SVG markup branch.
             assetOptions: {
-                permittedFileTypes: ['image/*', 'video/*', 'audio/*', '.pdf', 'text/html'],
+                permittedFileTypes: ['image/*', '.pdf', 'text/html'],
             },
             plugins: [
                 AssetServerPlugin.init({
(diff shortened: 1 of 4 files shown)


biggamesmallworld (changes requested):
Right diagnosis, and the JSDoc explaining *why* is genuinely good. One blocking issue plus two test fixes.

## Must fix

**`sandbox` applies to every asset, including the ones that were never the vulnerability.**

I verified the PDF behaviour you flagged: a local server returning `application/pdf` renders in Chrome's built-in viewer normally, and with `default-src 'none'; script-src 'none'; sandbox` the navigation aborts and the file downloads instead. So this is a real, visible regression, and it is targeted at `master` (patch).

The markup types already get `Content-Disposition: attachment`, so they are never rendered as a document. The `sandbox` token is therefore doing its work only for cases where the mime lookup is wrong, and costing inline PDF viewing for everyone else.

Suggested fix: keep `default-src 'none'; script-src 'none'` global (that still covers a scriptable asset whose mime we misidentified, e.g. an SVG with no extension, which falls through to `application/octet-stream` at asset-server.ts:96 and gets no attachment header), and add `sandbox` only for `isMarkupMimeType`. That removes the PDF regression without losing the defence in depth.

If the team would rather keep the global `sandbox`, it needs an opt-out in `AssetServerOptions` plus a changelog entry, since a patch release should not silently kill inline PDF rendering.

## Test fixes

1. `expect(res.headers.get('content-disposition')).toContain('attachment')` (both the SVG and HTML tests): the header is set to the exact string `attachment`, so use `.toBe('attachment')`. `toContain` would also pass for `inline; filename="attachment.txt"`.

2. Neither the SVG nor the HTML test asserts `content-type`, so neither proves the branch its name claims. If mime lookup regressed to `text/html` the SVG test would still pass. Assert `res.headers.get('content-type')` contains `image/svg+xml` and `text/html` respectively.

## Nits

- Move the `security headers` describe below the `caching` block. Right now it sits before `cache initially empty`, and the `cache=false` comment documents that coupling rather than removing it.
- The CSP string and the markup mime list would be better in `src/constants.ts`, which already owns `DEFAULT_CACHE_HEADER`.
- The last JSDoc paragraph (the PDF behaviour note) is changelog material rather than something the next reader of the method needs, and it goes stale as soon as the must-fix above lands.
- The `permittedFileTypes` override in the e2e config duplicates `default-config.ts:146` plus one entry, so it will drift silently when the defaults change. Worth asking whether the `text/html` branch earns a global config change and a new fixture, given SVG already exercises the markup path.
- Not this PR, but noting it: with no `format` param, `getFileNameFromParameters` keeps the original extension, so `/foo.svg?w=100` is served as `image/svg+xml` for a body sharp has already rasterised to PNG (confirmed locally). This PR is the first thing to make a header decision from that value. Pre-existing, probably worth a follow-up issue.

## Checked and withdrawn

I initially suspected `nosniff` plus the `application/octet-stream` fallback might break existing `` usage via ORB. Tested in Chrome: PNG bytes served as `application/octet-stream` with `nosniff` and the new CSP still render fine. No issue there.

grolmus:
Thanks Will — all points addressed in `3edda6a`.

**Must fix — `sandbox` regression:** `sandbox` is now scoped to markup types only. Every asset still gets `nosniff` + `default-src 'none'; script-src 'none'` (covers a mis-identified scriptable asset); markup additionally gets `sandbox` + `Content-Disposition: attachment`. PDFs render inline in the built-in viewer again.

**Test fixes:**
- `content-disposition` now asserted with `.toBe('attachment')`.
- Added `content-type` assertions (`image/svg+xml` / `text/html`) so each test proves its branch.

**Nits:**
- CSP strings + markup mime list moved to `src/constants.ts`.
- `security headers` describe moved below the `caching` block; dropped the `cache=false` workaround.
- Removed the stale PDF paragraph from the JSDoc.
- `permittedFileTypes` override narrowed to what the tests actually upload (`['image/*', '.pdf', 'text/html']`) so it no longer mirrors — and drifts from — the default.

Also added a PDF test case asserting no `sandbox` / no `attachment`, which pins the exact regression the split fixes (the JPEG case didn't, since JPEG was never force-downloaded).

Filed the pre-existing `getFileNameFromParameters` content-type follow-up as a separate issue.

40/40 e2e green, eslint + tsc clean.

biggamesmallworld (approved):
Feedback addressed

Issue #5105, opened 2026-08-10
Support TypeORM v1 alongside v0.3


michaelbromley:
TypeORM 1.0 landed in May, and 1.1 followed in July. We're still on 0.3, and I want the next minor of Vendure to work on both.

It has to be both. Plenty of people running Vendure in production have no appetite for a database layer upgrade right now, and a version of Vendure that forces one on them isn't much use to them. Equally, anyone starting fresh shouldn't be pinned to a version of TypeORM that's now tagged `legacy` on npm. So `@vendure/core` needs to work against `^0.3.21` and `^1.0.0` from the same source, and we need to be able to prove it on every commit.

This issue tracks that work. I'll keep findings here as they come in.

## The test harness

`scripts/typeorm/` runs everything that touches a database against either version.

```sh
bun run typeorm:use 1          # switch the workspace to TypeORM v1
bun run build:for-tests        # build, continuing past type errors
bun run test:db                # unit + e2e, sqljs
DB=postgres bun run test:db    # unit + e2e, postgres
bun run typeorm:use --reset    # back to the committed versions
```

`typeorm` is a direct dependency of `@vendure/core` rather than a peer dependency, so you can't pick the version at install time. Each profile in `profiles.json` is applied as a set of root-level `overrides`, which is the same mechanism we already use to keep `graphql` on a single version. A profile covers more than `typeorm` itself: v1 needs `@nestjs/typeorm` 11.0.1 or later, because earlier releases declare a `typeorm` peer range of `^0.3.0`, and it needs `better-sqlite3` v12.

`typeorm:verify` checks that exactly one copy of TypeORM is installed and that it's the one you asked for. Entity metadata is registered against a single module instance, so a second copy nested somewhere in `node_modules` produces failures that read exactly like genuine version incompatibilities. `test:db` runs this check before it runs anything else.

`build:for-tests` runs each build stage separately instead of chaining them with `&&`. `@vendure/core` builds in three stages — the main compile, the CLI compile, and a copy of the static `.graphql` schema files — and while the migration is in progress the main compile fails, so `dist/` never gets its schema files. The server then refuses to boot with "No type definitions were found", which says nothing about the TypeORM incompatibility underneath. `tsc` emits its output even when it reports errors, so running the stages separately gets you a bootable server to test against. The strict compile still happens in the workflow's `build` job, so type errors are still reported.

[...]

PR #5107, opened 2026-08-10
refactor(core): Use object syntax for find options relations and select

michaelbromley:
Moves `@vendure/core` off the string-array form of the TypeORM find options `relations` and `select` properties. The object form is valid in both v0.3 and v1, so this lands on the current version and is proven by the existing Build & Test suite.

Stacked on #5106. Relates to #5105.

## Vendure keeps accepting arrays

`RelationPaths<T>` is a string array, it is public API, and the `@Relations()` decorator produces one. Plugins pass those straight into `getEntityOrThrow`, `findOneInChannel` and `findByIdsInChannel`, so those methods have to keep taking arrays or every plugin doing this breaks on v1 — in a minor release.

So `TransactionalConnection` now owns its find-options types instead of extending TypeORM's. `VendureFindOneOptions` and `VendureFindManyOptions` accept `relations` and `select` in either form and normalise to the object form before handing over to TypeORM. That fixed 62 call sites without touching any of them.

The remaining call sites go directly to a TypeORM repository, and those are converted here. A literal path list becomes an object literal; a list built at runtime from `RelationPaths` goes through `findOptionsArrayToObject`, which is exported for plugins doing the same thing.

## What the object form catches

The array form was an untyped string list, so nothing checked the paths. The object form is checked against the entity, which turned up a real one: three queries in `SessionService` request `relations: ['user', 'user.roles', 'user.roles.channels']` against the abstract `Session`, but `user` is declared on `AuthenticatedSession`. It resolves at runtime through single-table inheritance, which is why it has always worked, but the type was wrong and nothing said so.

I have left those three as they are, going through `findOptionsArrayToObject` with a comment, because changing which entity a session query targets is a behaviour change and does not belong in a syntax conversion. Worth a separate look.

`findByIdsInChannel` also had a real bug: it only applied the tree-relation join filtering when `relations` was an array, so passing the object form produced duplicate joins for Collection tree relations. It now filters both forms, matching `findOneInChannel`.

## Effect

`@vendure/core` type errors against TypeORM v1: **184 → 35**. The `relations`/`select` cluster is gone entirely. What is left is the `'sqlite'` driver type cases, `Connection`/`createConnection`, and the renamed deep import paths, all tracked in #5105.

## Checks done

On v0.3: clean typecheck, 1268 unit tests pass, and the `order`, `administrator` and `session-management` e2e suites pass (125 tests) — the last of those covers the three `Session` sites and `order` covers the largest cluster of changes.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5107"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788957024&installation_model_id=20193&pr_number=5107&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5107&signature=aa8c3c6a34fce692405d31fd26b60a6ebbce4cae33c420349e994427c5594a57"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,6 +10,10 @@ inputs:
         required: false
         # Pinned. Bump explicitly when re-validating against a newer Bun.
         default: '1.3.10'
+    typeorm:
+        description: 'TypeORM profile from scripts/typeorm/profiles.json. Leave empty to install the versions declared in the workspace.'
+        required: false
+        default: ''
 runs:
     using: 'composite'
     steps:
@@ -20,5 +24,14 @@ runs:
           with:
               bun-version: ${{ inputs.bun-version }}
         - name: Install dependencies
+          if: inputs.typeorm == ''
           shell: bash
           run: bun install --frozen-lockfile
+        # Applying a profile rewrites the root overrides, so the lockfile is
+        # expected to change and cannot be installed frozen.
+        - name: Install dependencies (TypeORM ${{ inputs.typeorm }})
+          if: inputs.typeorm != ''
+          shell: bash
+          run: |
+              node scripts/typeorm/use-version.mjs "${{ inputs.typeorm }}"
+              node scripts/typeorm/verify.mjs
--- a/.github/workflows/typeorm_v1.yml
+++ b/.github/workflows/typeorm_v1.yml
@@ -29,28 +29,9 @@ concurrency:
 # the e2e-* jobs because GitHub Actions supports neither YAML anchors nor
 # service reuse. Update all four copies together.
 jobs:
-    should-run:
-        name: should run
-        runs-on: ubuntu-latest
-        permissions:
-            contents: read
-        outputs:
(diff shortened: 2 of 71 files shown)


michaelbromley:

PR #5120, opened 2026-08-10
refactor(core): Read the TypeORM DataSource through a version-independent accessor

michaelbromley:
Third in the series for the TypeORM v1 work tracked in vendurehq/vendure#5105.

**Stacked on vendurehq/vendure#5107**, so the base branch here is `refactor/find-options-object-syntax` rather than `minor`. Only the last commit belongs to this PR. It needs retargeting to `minor` once #5107 merges.

## What this does

Adds `getDataSource()`, which returns the `DataSource` that a TypeORM object belongs to by reading whichever of the `dataSource` and `connection` properties the installed version of TypeORM provides. It is applied at every site in `@vendure/core` which reaches for a TypeORM object's `DataSource`.

## What was actually broken

The issue predicted that `.connection` exists only on 0.3 and `.dataSource` only on v1, so that every one of these sites needed the accessor. Reading the published typings for `typeorm@0.3.31` and `typeorm@1.1.0` gives a more precise picture:

| Receiver | 0.3.31 | 1.1.0 | Does `.connection` work on both? |
| -- | -- | -- | -- |
| `QueryBuilder` | `connection` | `dataSource`, plus a deprecated `connection` getter | yes |
| `QueryRunner` | `connection` | `dataSource`, plus a `connection` getter | yes |
| `EntityManager` | `connection` | `dataSource`, plus a `connection` getter | yes |
| `EntityMetadata` | `connection` | getter present in the JS, dropped from the `.d.ts` | at runtime, but not in the types |
| `RelationIdLoader` (private field) | `this.connection` | `this.dataSource` | no |

So of the eighteen sites, exactly one was genuinely broken: the `RelationIdLoader` patch in `packages/core/src/entity/typeorm-relation-id-loader-fix.ts`. That class holds its `DataSource` in a private field, which TypeORM renamed without leaving an alias, because nothing outside the class was supposed to be reading it. That single site produced the 110 or so `Cannot read properties of undefined (reading 'driver')` failures per backend reported on the issue.

The other seventeen sites work on v1 today through the deprecated getter. They are converted here as well, so that they do not break when TypeORM removes it.

The accessor prefers `dataSource` and falls back to `connection`. That ordering means the fallback branch stops being used the moment a workspace moves to v1, rather than keeping every call site on the deprecated property.

## The test

`get-data-source.spec.ts` covers both property shapes, the case where both are present, and the error thrown when neither is. The last test constructs a real `RelationIdLoader` from whichever TypeORM is installed and asserts that the accessor resolves its `DataSource`. That is the check which would have caught the original failure, and it is meaningful under both profiles because the same file runs against both.

The original failure produced no type error at all. The patch assigns through `proto[methodName]`, which is `any`, so there was nothing for the compiler to check. The patched method now declares `this: DataSourceHolder`, which states what it needs from its receiver instead of leaving it implicit, but a monkey-patch against an `any`-typed prototype cannot be fully typechecked, which is why the test matters more than the annotation.

## Results

Against `typeorm@0.3.31`:

- strict `tsc` on `@vendure/core` is clean
- the full unit suite passes, 90 files and 1275 tests
- the `list-query-builder`, `collection`, `custom-field-relations`, `asset` and `entity-hydrator` e2e suites pass on sqljs, 321 tests

Against `typeorm@1.1.0`:

- type errors in `@vendure/core` stay at 36, unchanged from the #5107 baseline
- the unit suite passes
- no `reading 'driver'` failures remain in e2e, down from around 110 per backend

## What this does not fix

The e2e suites still do not complete on v1, but they now stop somewhere else. `SqlJobQueueStrategy.connectionAvailable()` reads `this.rawConnection.isConnected`, and `DataSource.isConnected` was removed in v1, having been deprecated in 0.3 in favour of `isInitialized`. It evaluates to `undefined`, so every job queue call throws "Connection not available" and test data population hangs.

Applying that one-word change locally and rerunning `collection.e2e-spec.ts` on v1 gets the server booting and the suite running: 15 passed, 61 failed and 2 skipped, with the failures being differences in collection contents rather than crashes. That change belongs to the `Connection`/`createConnection` to `DataSource` item on the issue, so it is not included here, but it works on both versions and is worth taking whenever that item is picked up.

Relates to vendurehq/vendure#5105

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788981065&installation_model_id=20193&pr_number=5120&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5120&signature=309a0a39de6d795798976d6c9ece6c1943b70a6cf828ad6e1c97470748cd7c51"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/guides/core-concepts/collections/index.mdx
+++ b/docs/docs/guides/core-concepts/collections/index.mdx
@@ -68,7 +68,7 @@ added to it using the `.andWhere()` method.
 Here's an example of a collection filter which filters by SKU:
 
 ```ts title="src/config/sku-collection-filter.ts"
-import { CollectionFilter, LanguageCode } from '@vendure/core';
+import { CollectionFilter, getDataSource, LanguageCode } from '@vendure/core';
 
 export const skuCollectionFilter = new CollectionFilter({
     args: {
@@ -91,9 +91,9 @@ export const skuCollectionFilter = new CollectionFilter({
     // This is the function that defines the logic of the filter.
     apply: (qb, args) => {
         // Sometimes syntax differs between database types, so we use
-        // the `type` property of the connection options to determine
+        // the `type` property of the DataSource options to determine
         // which syntax to use.
-        const LIKE = qb.connection.options.type === 'postgres' ? 'ILIKE' : 'LIKE';
+        const LIKE = getDataSource(qb).options.type === 'postgres' ? 'ILIKE' : 'LIKE';
 
         return qb.andWhere(`productVariant.sku ${LIKE} :sku`, {
             sku: `%${args.sku}%`
--- a/docs/docs/reference/typescript-api/configuration/collection-filter.mdx
+++ b/docs/docs/reference/typescript-api/configuration/collection-filter.mdx
@@ -16,7 +16,7 @@ Here's a simple example of a custom CollectionFilter:
 *Example*
 
 ```ts
-import { CollectionFilter, LanguageCode } from '@vendure/core';
+import { CollectionFilter, getDataSource, LanguageCode } from '@vendure/core';
 
 export const skuCollectionFilter = new CollectionFilter({
   args: {
@@ -38,7 +38,7 @@ export const skuCollectionFilter = new CollectionFilter({
 
   // This is the function that defines the logic of the filter.
   apply: (qb, args) => {
-    const LIKE = qb.connection.options.type === 'postgres' ? 'ILIKE' : 'LIKE';
+    const LIKE = getDataSource(qb).options.type === 'postgres' ? 'ILIKE' : 'LIKE';
     return qb.andWhere(`productVariant.sku ${LIKE} :sku`, { sku: `%${args.sku}%` });
   },
 });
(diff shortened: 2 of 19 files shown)


michaelbromley:

PR #5121, opened 2026-08-11
refactor(core): Replace the removed TypeORM Connection API with DataSource

michaelbromley:
Fourth in the series for the TypeORM v1 work tracked in vendurehq/vendure#5105.

**Stacked on vendurehq/vendure#5120**, so the base branch here is `mbromley/oss-696-support-typeorm-v1-alongside-v03` rather than `minor`. The two commits on top of that branch are this PR. It needs retargeting once the PRs below it merge.

## What this does

Replaces every use of the TypeORM connection API that v1 removed. All of it is a straight rename to APIs which already exist in 0.3.28, so this lands on `minor` and works on both versions with no version-specific code.

| Removed in v1 | Replacement | Where |
| -- | -- | -- |
| `Connection` | `DataSource` | `migrate.ts`, `bootstrap.ts`, `transaction-subscriber.ts`, `sql-job-queue-strategy.ts`, `injector.ts`, `parse-channel-param.ts` |
| `createConnection()` | `new DataSource().initialize()` | `migrate.ts`, `@vendure/testing`'s `clear-all-tables.ts` |
| `DataSource.isConnected` | `DataSource.isInitialized` | `sql-job-queue-strategy.ts` |
| `DataSource.close()` | `DataSource.destroy()` | `migrate.ts`, `clear-all-tables.ts` |
| `DataSourceOptions.name` | dropped, see below | `migrate.ts` |

I found `close()` and `DataSourceOptions.name` by diffing the published `DataSource` typings between 0.3.31 and 1.1.0 rather than working from the error list, since the error list only shows what the compiler can see. The full set of members present in 0.3 and gone in v1 is `close`, `connect`, `isConnected`, `getCustomRepository` and `name`. Of those, only `close` and `isConnected` were in use.

## Why dropping the connection `name` is safe

`provisionShadowDatabase()` gave its two temporary connections explicit names. That was necessary under `createConnection()`, which calls `getConnectionManager().create(options)` and throws `AlreadyHasActiveConnectionError` if a connection with the same name is already initialised, defaulting to `"default"`.

`new DataSource(options).initialize()` does not touch the global connection manager at all, so there is no registry for the names to collide in and nothing reads them back. All three sites in `migrate.ts` are covered: the shadow-admin connection, the shadow connection, and the in-memory sqlite options. Nothing in the repository uses `getConnection()` or `getConnectionManager()`, so removing the names loses no behaviour on 0.3 and removes a piece of global state.

This also accounts for the two `as DataSourceOptions` casts noted on the issue as "no longer overlapping, most likely fallout from the driver union changing". It was not the driver union: `name` was removed from `BaseDataSourceOptions` in v1, and both cast errors disappear once it goes.

## The one public type change

`TransactionSubscriberEvent.connection` is declared as `Connection` and becomes `DataSource`. In 0.3, `Connection` is an empty deprecated subclass of `DataSource` which adds no members, so nothing a consumer could be calling is lost. TypeORM constructs `DataSource` and never `Connection`, so the value carried by that event was already a plain `DataSource` and the declared type was too narrow. The property keeps its name, so no consumer code changes.

## Results

Against `typeorm@0.3.28`:

- strict `tsc` clean on both `@vendure/core` and `@vendure/testing`
- full unit suite passes, 90 files and 1277 tests
- on sqljs, the `job-queue`, `job-queue-concurrency`, `database-transactions`, `migrate-asset-translations` and `migrate-product-option-groups` e2e suites pass
- on postgres, the same suites pass, plus `migrate-from-empty`, which is the only coverage of the shadow database code and is skipped on sqljs. It asserts the shadow database is both created and cleaned up, which is the behaviour the `name` removal touches.

Against `typeorm@1.1.0`:

- type errors in `@vendure/core` drop from 36 to 23, with the `Connection`/`createConnection` cluster and both `migrate.ts` cast errors gone
- e2e suites now run to completion with no infrastructure failures at all: zero "Connection not available", zero `reading 'driver'`, zero unhandled rejections
- across `collection`, `list-query-builder`, `asset`, `entity-hydrator`, `custom-field-relations`, `job-queue` and `database-transactions`, **133 tests now pass on v1**, against zero at the original baseline. `job-queue` and `database-transactions` pass outright.

## Two findings for the issue, neither addressed here

**vendurehq/vendure#5107 is not quite complete, and it is the cause of most of the remaining e2e failures.** 19 occurrences of `String-array "relations" syntax has been removed` still appear, from two places:

- `EntityHydratorService` passes a string array into `setFindOptions({ relations })`, which fails all 26 `entity-hydrator` e2e tests
- `FacetService.findByCode()` passes a string array to a raw repository `findOne()`. The importer calls it, so seed data import fails with "1 errors encountered when importing product data", FacetValues are never created, and dozens of downstream assertions then fail on missing data. This is why `collection.e2e-spec.ts` reports "Could not find a FacetValue with the code electronics" and `asset.e2e-spec.ts` sees 1 asset instead of 4.

These are the same sites as the 8 remaining `string[]` is not assignable to `FindOptionsRelations` type errors. They are invisible on 0.3, where the string array form still typechecks and still works. Worth fixing in #5107 before it merges, since the seed data failure makes every other v1 number look worse than it is.

**One failure that looks like a genuine v1 behavioural difference.** `custom-field-relations.e2e-spec.ts` dies during `ChannelService.initChannels` with `Cannot read properties of undefined (reading 'profile')` inside `OrmUtils.deepValue`, reached from `loadRelation` in `SelectQueryBuilder`. That is a `relationLoadStrategy: 'query'` path and does not look like string-array residue. It belongs with the behavioural-changes item on the issue.

Relates to vendurehq/vendure#5105

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5121"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789024366&installation_model_id=20193&pr_number=5121&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5121&signature=cbebe5be922c6ddea712415904d559242498cc5b63c9d39843c76c6e54a4abef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/typeorm_v1.yml
+++ b/.github/workflows/typeorm_v1.yml
@@ -14,11 +14,10 @@ on:
         # Nightly, so regressions in either TypeORM v1 or Vendure surface without
         # anyone having to remember to trigger a run.
         - cron: '0 3 * * *'
+    # No base-branch filter: the `typeorm-v1` label is the only gate, applied per job
+    # below. A pull request stacked on another feature branch does not target master,
+    # major or minor, and filtering on the base would silently skip every one of them.
     pull_request:
-        branches:
-            - master
-            - major
-            - minor
         types: [opened, reopened, synchronize, labeled]
 
 env:
--- a/packages/core/e2e/fixtures/test-payment-methods.ts
+++ b/packages/core/e2e/fixtures/test-payment-methods.ts
@@ -133,7 +133,7 @@ export const singleStageRefundFailingPaymentMethod = new PaymentMethodHandler({
     createRefund: async (ctx, input, amount, order, payment, args) => {
         const paymentWithRefunds = await connection
             .getRepository(ctx, Payment)
-            .findOne({ where: { id: payment.id }, relations: ['refunds'] });
+            .findOne({ where: { id: payment.id }, relations: { refunds: true } });
         const isFirstRefundAttempt = paymentWithRefunds?.refunds.length === 0;
         const metadata = isFirstRefundAttempt ? { errorMessage: 'Service temporarily unavailable' } : {};
         return {
--- a/packages/core/e2e/fixtures/test-plugins/hydration-test-plugin.ts
+++ b/packages/core/e2e/fixtures/test-plugins/hydration-test-plugin.ts
@@ -36,7 +36,7 @@ export class TestAdminPluginResolver {
     async hydrateProduct(@Ctx() ctx: RequestContext, @Args() args: { id: ID }) {
         const product = await this.connection.getRepository(ctx, Product).findOne({
             where: { id: args.id },
-            relations: ['facetValues'],
+            relations: { facetValues: true },
         });
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         await this.entityHydrator.hydrate(ctx, product!, {
--- a/packages/core/e2e/fixtures/test-plugins/issue-2453/services/campaign.service.ts
+++ b/packages/core/e2e/fixtures/test-plugins/issue-2453/services/campaign.service.ts
@@ -54,7 +54,7 @@ export class CampaignService {
     findOne(ctx: RequestContext, id: ID): Promise<Translated<Campaign> | undefined | null> {
         return this.connection
             .getRepository(ctx, Campaign)
-            .findOne({ where: { id }, relations: ['promotion'] })
+            .findOne({ where: { id }, relations: { promotion: true } })
             .then(campaignItem => {
                 return campaignItem && translateDeep(campaignItem, ctx.languageCode, ['promotion']);
             });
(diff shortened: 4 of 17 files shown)


michaelbromley:

PR #5139, opened 2026-08-14
feat(cli): Add plugin API to register and override commands

dlhck:
# Description

Adds a plugin API to `@vendure/cli` so external packages can:

1. **Add** new top-level commands
2. **Replace / wrap** existing commands (e.g. Cloud overriding `vendure dev` while still calling `builtinCommands.dev.action(...)`)

Discovery is hybrid: auto-load direct dependencies that declare `vendure.cliPlugin`, with optional project `vendure.cli.plugins` / `exclude` allow/deny lists. Built-in command actions now return exit codes so the host owns `process.exit` and wrapping works cleanly.

Also exports a public library surface (`defineCliPlugin`, `builtinCommands`, types) via `package.json` `exports`, plus docs and unit tests.

# Breaking changes

No. Additive API; existing CLI usage is unchanged when no plugins are installed.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5139"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789317296&installation_model_id=20193&pr_number=5139&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5139&signature=11541a140332ea2eeb2ad6757fb9c9737166a00af6b4d196967897564181d3a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


dlhck:
@coderabbitai review

michaelbromley (commented):
Two things about plugin discovery I'd want resolved before this merges. The rest looks good — the `CliCommandExit` work is the right fix for the `process.exit()` problem, and scanning only direct dependencies rather than the whole tree was a good call.

## 1. Implicit mode runs code from every direct dependency

In `resolveCliPlugins()` (`packages/cli/src/shared/resolve-cli-plugins.ts`), with no `vendure.cli.plugins` set, the candidate list is every entry in `dependencies`, `devDependencies` and `optionalDependencies`. Any of those shipping a `vendure.cliPlugin` field gets `require()`d at startup and can replace `dev`, `build` or `migrate`. Nobody has to approve it.

So if any direct dependency is compromised, the attacker gets code execution the next time someone types `vendure`, and can quietly proxy commands that people run against production databases. npm lifecycle scripts had the same problem, and both npm and pnpm have since moved them to opt-in.

I'd keep discovery implicit but make activation explicit. The CLI already knows which installed packages declare a plugin, so it can print:

```
@vendure/cloud provides CLI commands. Add it to vendure.cli.plugins in package.json to enable them.
```

and load nothing until told. If we keep the current default instead, the docs need to say outright that installing a package which declares a CLI plugin lets it replace built-in commands.

## 2. Override precedence is alphabetical

Implicit mode sorts candidates with `localeCompare`:

```ts
const candidateNames = allowlist
    ? [...allowlist]
    : [...directDependencyNames].sort((a, b) => a.localeCompare(b));
```

`CommandRegistry.register()` is last-write-wins, so when two plugins both register a `dev` command, the winner is whichever package name sorts later. It's stable, so it works fine until someone renames a package or adds an unrelated dependency, and then the CLI behaves differently with no code change and nothing obvious to grep for. The allowlist path preserves declared order, but that isn't the default. I assume the sort is there to make the scan deterministic, which is fine, but we need a real precedence rule as well.

The replacement notice makes this worse:

```ts
process.stderr.write(
    pc.dim(`Replaced command "${command.name}" via ${source}\n`),
);
```

In implicit mode that line is the only clue that your `vendure dev` isn't the one described in the docs, and people don't read dim stderr output. It should either not be printed on every run, or not be dim.

michaelbromley:
Following up on the trust model point above, since I asked for a change without showing how cheap it is. Also the non-interactive story, which matters here because a lot of CLI usage is now agents.

## What explicit activation would cost

Everything needed is already a dependency. `@clack/prompts` is in `packages/cli/package.json`, `isNonInteractiveEnvironment()` and `withInteractiveTimeout()` are in `packages/cli/src/utilities/utils.ts`, `fs-extra` handles the write, and the deny path can use the existing `vendure.cli.exclude`.

Shape I'd suggest:

- `vendure plugins` — list discovered packages with status (enabled, not enabled, excluded, failed to resolve), with an interactive multiselect to toggle them.
- `vendure plugins add <pkg>` and `vendure plugins remove <pkg>` — non-interactive, writing to `vendure.cli.plugins` and `vendure.cli.exclude`.
- No prompting during normal runs. When unapproved plugins are found, one line: `2 packages provide CLI commands. Run "vendure plugins" to review them.`

Close to what pnpm does with `pnpm approve-builds` writing to `pnpm.onlyBuiltDependencies`.

Three things to watch on the write:

1. `plugins` is currently an exclusive allowlist, so the first write into it flips the project from scan-everything to only-these and silently disables whatever was auto-loading before. Auto-add has to land with the default change, not before it.
2. `fs.writeJson(..., { spaces: 2 })` reformats the whole file. Detect the existing indent and keep the trailing-newline convention.
3. Print the absolute path being written. `resolveCliProjectRoot()` walks up to the first ancestor mentioning `@vendure/cli`, so running from `packages/api` in a monorepo writes to the repo root.

## Non-interactive and agents

**Never prompt, never hang.** `isNonInteractiveEnvironment()` already covers `CI`, `VENDURE_CLI_NON_INTERACTIVE` and non-TTY stdin or stdout. The approval flow should go through it.

**Make the unknown-command error actionable.** Cheapest high-value item here. `abortIfNonInteractive()` already sets the precedent by printing the flags to use instead of just refusing. Agents will land on unknown-command constantly:

```
Unknown command "cloud-deploy".
It is provided by @vendure/cloud, which is installed but not enabled.
Enable it with: vendure plugins add @vendure/cloud
```

Otherwise an agent gets "unknown command" with no way to discover that an installed package provides it, and burns a lot of turns guessing.

**`vendure plugins --json`**, with status and the reason for any skip. That also fixes something I missed in the review: when `defaultResolvePackage()` exhausts all three fallback tiers it returns `null` and implicit mode skips silently, so an installed plugin with an unusual layout never loads and nothing explains why.

**No blanket-approve flag.** No `--yes-to-all`, no `VENDURE_CLI_APPROVE_ALL`. Agents are very good at finding the flag that makes an error go away, and they will use it, which removes the boundary for exactly the people most likely to be running dependencies nobody reviewed. `VENDURE_CLI_NON_INTERACTIVE` is fine because it only makes the CLI stricter.

**Explicit activation helps CI.** The obvious objection to opt-in is that it breaks automation, but approvals live in a committed package.json, so CI resolves the same command set on every run regardless of what ends up in `node_modules`. That is more deterministic than loading whatever happens to be installed, and it removes the alphabetical precedence problem for any project that has been through the flow.

dlhck:
@michaelbromley Thanks — addressed in 929d38b.

**Explicit activation.** Discovery still scans direct deps for `vendure.cliPlugin`, but nothing is loaded until it’s listed in `vendure.cli.plugins`. Unapproved packages get a one-line hint (`N packages provide CLI commands. Run "vendure plugins" to review them.`) and are never `require()`’d on normal runs.

**`vendure plugins`.** As you sketched:
- `vendure plugins` — interactive multiselect (TTY only)
- `vendure plugins --json` — status + skip reason, no prompts
- `vendure plugins add|remove <pkg>` — non-interactive writes to `plugins` / `exclude`
- Preserves package.json indent + trailing newline, prints the absolute path written
- No blanket-approve flag

**Unknown commands.** If the package declares optional `vendure.cliCommands`, we can point at the provider without loading it (`It is provided by @vendure/cloud… Enable with: vendure plugins add …`). Otherwise we fall back to the inactive-package hint.

**Precedence.** Overrides follow allowlist order (last wins). The replacement notice is yellow on stderr instead of dim.

Would appreciate another look when you have a moment.

michaelbromley (commented):
Re-reviewed at 979103c. Both points from my earlier review are properly addressed.

`resolveCliPlugins()` now returns nothing unless a package is listed in `vendure.cli.plugins`, and `discoverCliPlugins()` reads package.json only and never calls `require()`, so discovery executes no plugin code. Driving the unknown-command hint off a declarative `vendure.cliCommands` array is a better answer than I had — I assumed you would need to load a plugin to know what commands it provides. Precedence is now the declared order of the allowlist, with a test covering it. The `plugins` command, `--json` output with skip reasons, indent and trailing-newline preservation, and the `exclude` deny path are all in and tested.

Two things I would fix before this merges.

## 1. The interactive manager silently reorders `plugins`

In `runInteractiveManager()` (`packages/cli/src/commands/plugins/plugins.ts`), the saved array is built from `toggleable`, which comes from `discoverCliPlugins()`. That sorts direct dependencies with `localeCompare`, and `Map.set` on an existing key keeps the original insertion position, so the result stays alphabetical.

A project with `plugins: ["@z/late", "@a/early"]`, ordered that way so `@a/early` wins the `dev` override, gets rewritten to `["@a/early", "@z/late"]` as soon as anyone opens the interactive manager and saves. `@z/late` then wins instead, with nothing in the output to say so.

Declared order has just become the precedence rule, so this path needs to preserve it. `addCliPluginToProjectConfig()` already gets this right by pushing onto the existing array.

## 2. `writeUnknownCommandHelp()` guesses when there is one inactive plugin

```ts
if (inactive.length === 1) {
    process.stderr.write(`It may be provided by ${inactive[0]}, which is installed but not enabled.\n`);
    process.stderr.write(`Enable it with: vendure plugins add ${inactive[0]}\n`);
```

Typing `vendure buidl` tells you to enable `@vendure/cloud`. The `declaredCommands` match in `findInactivePluginProvidingCommand()` has already run and found nothing, so this branch fires exactly when there is no evidence for it.

Agents follow instructions in error output literally, so this turns a typo into an unintended plugin activation, which is what explicit activation is meant to prevent. The `declaredCommands` path is good and should stay. The guess should go.

## Should fix

**`discoverCliPlugins()` is not memoized and runs up to three times per invocation.** `cli.ts` calls it via `maybeWriteInactivePluginsHint()` on every command, then again through `findInactivePluginProvidingCommand()` and `listInactiveCliPluginPackages()` on the unknown-command path. Each run calls `resolvePackage()` for every direct dependency, which is a `require.resolve` plus a `readJsonSync`. When a dependency does not resolve, which is routine for uninstalled `optionalDependencies` such as platform-specific binaries, `defaultResolvePackage()` falls through all three tiers and the last one walks every ancestor directory to the filesystem root calling `existsSync`. That is now on the startup path of `vendure dev` and `vendure build`. A per-process cache would fix it.

**The interactive save drops `exclude` entries it cannot see.** `writeCliPluginProjectConfig({ plugins, exclude })` replaces `exclude` wholesale with names derived from `toggleable`, so anything already in `exclude` that no longer appears in discovery, because it was uninstalled or stopped declaring `vendure.cliPlugin`, is deleted from the user's config without mention. Worth merging rather than replacing.

**`vendure.cliCommands` is never validated against what the plugin registers.** A plugin declaring `cliCommands: ["deploy"]` while registering `cloud-deploy` produces a hint that tells you to enable it, after which the command still does not exist. `resolveCliPlugins()` has both the declared list and the loaded `plugin.commands`, so warning on a mismatch there is cheap.

**`plugins` is itself an overridable built-in.** It is in `builtinCommandDefs`, so an enabled plugin can replace the command that manages activation. An enabled plugin already has code execution so this is not an escalation, but I would reserve the name rather than let the activation control surface be replaceable.

## Nitpicks

- `abortIfNonInteractive()` inside `runInteractiveManager()` is unreachable, since `pluginsCommand()` already returns `printTextList(...)` when `isNonInteractiveEnvironment()` is true.
- With more than one inactive plugin, an unknown command prints "N packages provide CLI commands. Run `vendure plugins` to review them." twice, once from the pre-parse hint and once from the unknown-command handler.
- Overriding `command:*` may suppress commander's built-in "did you mean" suggestion. Worth checking `showSuggestionAfterError` still fires.
- `readCliProjectPackageJson()` returns `null` for both a missing and a malformed package.json, so a JSON syntax error reports "Could not find a project package.json".
- The comment in `writeCliPluginProjectConfig()` says it keeps `plugins` when empty "so the project stays in explicit-activation mode after the first write". There is no other mode now, and `undefined` and `[]` both load nothing.
- `example-cli-plugin.ts` is still imported by nothing. It has grown into a reasonable documentation example, but it lives in `__tests__/fixtures/` and its comment still says "see unit tests for fixture-based discovery".

gabriellbui (commented):
Reviewed at `979103c`. Everything already covered by @michaelbromley's two reviews is left out, so this is only what is not on those lists.

## Blocks merge

### One unusable entry in `vendure.cli.plugins` takes out every command, including the one that recovers it

`packages/cli/src/cli.ts` loads plugins before it registers commands and before it parses argv:

```ts
const registry = new CommandRegistry();
registry.registerAll(builtinCommandDefs);

const plugins = resolveCliPlugins();              // :40  throws here
...
registerCommands(program, registry.toArray());    // :45  never reached
await program.parseAsync(process.argv);           // :55  never reached
```

`resolveCliPlugins()` throws when an allowlisted package is not a direct dependency (`resolve-cli-plugins.ts:243-247`), cannot be resolved (`:250-254`), declares no `vendure.cliPlugin` (`:256-261`), has a missing entry file (`:264-266`), or fails to load (`loadCliPluginModule`, `:421-433`). The only handler is `main().catch` (`cli.ts:117-120`) — print the message, exit 1.

Three states that produce a total outage:

- **Production install.** `vendure.cli.plugins` lists a package that lives in `devDependencies`. After `npm ci --omit=dev` the name is still in package.json, so the direct-dependency check at `:242-247` passes, but resolution fails and `:250-254` throws. `vendure migrate --run` and `vendure start` both fail in the deploy image.
- **Unbuilt workspace plugin.** `vendure.cliPlugin: "./dist/cli-plugin.js"` on a workspace package that has not been built yet → `fs.existsSync` false → `:264-266` throws. Every `vendure` invocation in the repo fails until someone hand-edits package.json.
- **Plugin fails to load.** Throws at require time, fails `assertCliPlugin`, or is ESM-only — `loadCliPluginModule` uses `require`, and nothing in the docs says the entry has to be CommonJS.

In all three, the documented recovery (`vendure plugins remove <pkg>`) is itself unreachable, and the error text suggests no way out. `discoverCliPlugins()` already classifies the same four conditions as `status: 'failed'` with a `reason` and no throw (`:166-202`) — startup could use that and keep the `plugins` command alive, or at minimum name the recovery command in the error.

Related, same file: discovery never loads the module, so a plugin that fails `assertCliPlugin` is reported as `enabled` by `vendure plugins --json` while the CLI is actually unusable.

## Should fix

### `plugins add` / `plugins remove` combined with `--json` writes nothing and exits 0

`packages/cli/src/commands/plugins/plugins.ts:32-51` — the `--json` branch returns before `add`, before `remove`, and before the unknown-action check:

```ts
if (options.json) {
    printJson(discoverCliPlugins());
    return;                                     // :34
}
if (normalizedAction === 'add')    { ... }      // :37
if (normalizedAction === 'remove') { ... }      // :42
if (normalizedAction) { log.error(`Unknown plugins action ...`); exitCliCommand(1); }  // :47-51
```

`vendure plugins add @vendure/cloud --json` prints the discovery list, never writes `vendure.cli.plugins`, and resolves `undefined` → exit 0. The JSON it prints still shows the package as `not-enabled`, so a script that polls until the status changes never terminates. `vendure plugins remove <pkg> --json` is the same silent no-op, and `vendure plugins bogus --json` swallows the invalid action entirely. `skills/vendure-cli/commands/plugins.md` tells agents to prefer `--json`, `add` and `remove`, which is exactly this combination.

### `exclude` silently beats an explicit `plugins` entry, and no hint will name the package

`resolveCliPlugins()` skips an allowlisted name that also appears in `exclude`, with no output at all:

```ts
for (const packageName of allowlist) {
    if (exclude.has(packageName)) {
        continue;                               // resolve-cli-plugins.ts:238-240
    }
```

`removeCliPluginFromProjectConfig()` is what puts names into `exclude` (`cli-plugin-project-config.ts:125-145`), and the guide shows hand-editing that block. So: run `vendure plugins remove @vendure/cloud`, later hand-add it back to `plugins`, and nothing loads, `vendure --help` does not list its commands, and `vendure cloud-deploy` prints a bare `Unknown command` — because `findInactivePluginProvidingCommand()` (`:293-297`) and `listInactiveCliPluginPackages()` (`:280-282`) both filter `status === 'not-enabled'` and this package is `'excluded'`. Precedence is stated in `packages/cli/src/README.md` but not in `docs/docs/guides/developer-guide/cli/index.mdx` and not in the `ProjectCliPluginConfig` JSDoc (`cli-command-definition.ts:54-58`).

### `plugins remove` validates nothing, so a typo reports success and writes a permanent bogus `exclude` entry

```ts
async function removePlugin(packageName: string): Promise<void> {   // plugins.ts:78-82
    const result = removeCliPluginFromProjectConfig(packageName);
    log.success(`Disabled CLI plugin ${pc.cyan(packageName)}`);
    log.info(`Wrote ${result.packageJsonPath}`);
}
```

No counterpart to `addPlugin`'s `assertPackageCanBeEnabled()` (`:84-106`). `vendure plugins remove @typo/nmae` writes `"exclude": ["@typo/nmae"]`, prints `Disabled CLI plugin @typo/nmae`, exits 0 — the real plugin stays enabled while the user believes it is off. The bogus entry is then invisible: discovery `continue`s on unresolvable names (`resolve-cli-plugins.ts:112-114`), and the interactive save drops it without mention.

### Project root resolution makes a plugin installed in a workspace package unreachable

```ts
if (pkg.vendure?.cli || hasDirectDependency(pkg, '@vendure/cli')) {
    return current;                             // resolve-cli-plugins.ts:77-79
}
```

With `@vendure/cli` hoisted to the workspace root as a devDependency — the normal layout — that returns the repo root. A plugin in `packages/app/dependencies` is then never discovered (`listDirectDependencyNames` reads the root package.json at `:106`), produces no startup hint, and never appears in `vendure --help`. `vendure plugins add @acme/cli-plugin` from `packages/app` hits `assertPackageCanBeEnabled` reading the same root file and prints *"is not a direct dependency of /ws. Install it first"* — telling the user to install a package that is already installed, with no working next step. Working around it by adding `vendure.cli` to `packages/app/package.json` makes the finder stop there on the first iteration instead, so the tool and the hand-edit disagree about which package.json owns the config. The guide states the CLI detects monorepo structure automatically.

### `plugins add` blames the plugin author when the real problem is the install

```ts
const discovered = discoverCliPlugins().find(p => p.packageName === packageName);
if (!discovered) {
    log.error(`Package "${packageName}" does not declare vendure.cliPlugin in its package.json, ...`);
    exitCliCommand(1);                          // plugins.ts:99-105
}
```

But the discovery scan drops unresolvable packages *before* it looks at that field:

```ts
const resolved = resolvePackage(packageName);
if (!resolved) { continue; }                    // :111-114  resolution failure
const entryRel = resolved.packageJson.vendure?.cliPlugin;
if (!entryRel || ...) { continue; }             // :115-118  genuinely absent
```

So a package that does declare `vendure.cliPlugin` but sits in a broken or unusual install is reported as missing the field. The user cannot tell "you need a newer plugin version" from "your install is broken".

### The interactive save also drops allowlist entries it classified as `failed`

Separate from the reordering and the `exclude` replacement already raised: `toggleable` is `discovered.filter(status !== 'failed')` (`plugins.ts:127`), and the saved `plugins` array is built only from it (`:169-171`). An allowlisted plugin whose entry file is momentarily missing is therefore deleted from `vendure.cli.plugins` by a user who only opened the manager to look. Currently masked by the startup throw above; it goes live as soon as that is made non-fatal. This write path has no test — there is no `plugins.spec.ts`.

### The package.json rewrite preserves indent and the trailing newline, but not the rest of the formatting

```ts
const indent = detectJsonIndent(project.raw);                                    // cli-plugin-project-config.ts:82
const trailingNewline = project.raw.endsWith('\n') ? '\n' : '';                  // :83
const serialized = `${JSON.stringify(nextPackageJson, null, indent)}${trailingNewline}`;
```

`JSON.stringify` with a non-empty indent puts every array element and object property on its own line, and Prettier collapses short arrays that fit `printWidth` — so `"keywords": ["vendure", "shop"]`, `"files": ["dist"]` and a one-line `exports` map all become multi-line in the commit diff of a command the user ran to toggle one flag. A CRLF file passes the `endsWith('\n')` check but every interior line ending is re-emitted as `\n`, so on Windows the write touches 100% of the lines. A minified single-line package.json has no leading whitespace for `detectJsonIndent` (`:15-21`) to match, falls back to `2`, and is fully expanded. Key order and non-ASCII content are fine.

## Smaller items

**The third tier of `defaultResolvePackage()` is unreachable in one case, and its comment describes the wrong thing** (`resolve-cli-plugins.ts:379-418`). Tier 3 lives in the `catch` of tier 2's `try`, so it runs only when `requireFromProject.resolve(packageName)` itself throws. When tier 2 resolves fine but no ancestor package.json has a matching `name`, control falls out of the `try` to `return null` at `:417` and the `node_modules` walk never runs — which is what an npm alias dependency (`"myalias": "npm:real-pkg@1"`, whose `package.json` says `name: "real-pkg"`) produces. That feeds the "was not found" throw above, or the misleading `plugins add` message above. The comment at `:393-396` presents the walk as the general fallback for a package that "may expose neither its root nor package.json", which is not the case it actually guards. `command-registry-store.spec.ts:380-409` covers tier 1 → tier 2 only.

dlhck:
@gabriellbui Thanks for the thorough pass — all items addressed in 2e1e70f.

**Blocker — one broken plugin takes out the CLI.** `resolveCliPlugins()` no longer throws. Each enabled plugin loads independently and failures come back as data; startup prints the reason plus `Skipping it. Fix the issue or disable it with: vendure plugins remove <pkg>` and continues with built-ins and the remaining plugins, so the recovery command is always reachable. On your related point: `vendure plugins` (list/`--json`/interactive) now actually loads enabled plugins, so one that fails `require()` or `assertCliPlugin` is reported as `failed` with the reason instead of `enabled`.

**`exclude` is gone entirely** (David's call): with an opt-in allowlist, disabling a plugin is simply not listing it. That removes both the exclude-beats-plugins precedence trap and the permanent bogus exclude entry from a typo'd `remove`. On top, `remove` now validates the package is actually enabled and errors with the enabled list otherwise — no more silent success. Tradeoff we accepted: a reviewed-and-declined plugin keeps its one-line startup hint.

**`add`/`remove` + `--json`:** actions now run before `--json` is considered, so the write happens first and the updated state is printed after; `plugins bogus --json` errors instead of being swallowed.

**Monorepo root resolution:** discovery scans direct dependencies of every package.json from the cwd up to the project root (nearest declaration wins), so a plugin in `packages/app/dependencies` is found and enablable with `@vendure/cli` hoisted. The allowlist still lives in the root config.

**`add` error attribution:** three distinct messages now — not a dependency, dependency-but-unresolvable (broken install / unbuilt workspace package), and genuinely missing `vendure.cliPlugin`.

**Interactive save:** merges the selection into the current allowlist via `mergeEnabledPluginSelection()`, preserving entries classified as `failed` and the existing order (which is override precedence). That merge logic is unit-tested, covering the write path you flagged.

**Resolver tier 3:** flattened out of the nested catch so it's reachable when tier 2 resolves but the name never matches (npm aliases), and it now trusts the path-based `node_modules` walk instead of re-checking `name`. Comment corrected.

**package.json rewrite:** CRLF files are now re-emitted with CRLF (plus the existing indent/trailing-newline detection). Honest caveat: `JSON.stringify` still expands Prettier-collapsed short arrays; matching Prettier's exact layout would need a real formatter, which felt out of proportion here.

Verified with 302 passing unit tests and an end-to-end smoke of the lifecycle: hint → add → run → break the entry file → CLI stays usable → remove.

PR #5141, opened 2026-08-15
fix(cli): resolve hoisted dependencies in monorepo workspaces

### LINKED ISSUE #5140: `vendure doctor`: dependency check reports the DB driver as "not installed" in npm/yarn/pnpm workspaces monorepos (hoisted packages aren't resolved)
**Describe the bug**

In a workspaces monorepo (npm/yarn/pnpm), `vendure doctor`'s `Dependencies` check reports the configured database driver package (e.g. `pg` for Postgres) as **not installed**, even though it is correctly installed and hoisted to the workspace root `node_modules`, resolvable by Node at runtime, and the app itself connects to the database and boots without any issue.

The check does a hard-coded filesystem probe for `<cwd>/node_modules/<driver>/package.json` instead of resolving the package the way Node actually would (e.g. `require.resolve`), so it can never find a driver hoisted to an ancestor directory. `doctor` already detects the monorepo layout in an earlier check (it prints `Monorepo detected (...)`), but that information isn't passed into the dependency check.

**To Reproduce**

1. Create an npm workspaces monorepo: root `package.json` with `"workspaces": ["apps/*"]` and no dependencies of its own.
2. Add a Vendure app at `apps/server` (a normal `@vendure/create`-style project) with `dbConnectionOptions.type: 'postgres'` and `pg` declared as a dependency in `apps/server/package.json`.
3. From the repo root, run `npm install` (or `npm ci`). npm hoists `pg` to `<root>/node_modules/pg`; nothing is installed at `apps/server/node_modules/pg`.
4. `cd apps/server && npx vendure doctor`.

**Expected behavior**

The `Dependencies` check should pass, reporting the Postgres driver (`pg`) as installed — it's present and resolvable via Node's module resolution from `apps/server`, and the `Database` check (which performs a real connection) confirms this by connecting successfully in the same run.

**Actual behavior**

The `Dependencies` step fails with:

```
DB driver "pg" not installed (required for dbConnectionOptions.type: "postgres")
```

even though every other check passes (Project, Config, Schema, and Database — including a live DB connection), and the application itself starts and runs correctly, including generating/running migrations. `doctor` exits non-zero, which breaks its use as a CI/pre-deploy gate (we run `doctor --profile production --strict`).

**Screenshots/Videos**

N/A — terminal output only, see Error logs.

**Error logs**

```
● Vendure Doctor

●  Project         pass  Vendure config found at src\vendure-config.ts
●                    Monorepo detected (apps)
●                    Package manager: npm
●                    Node.js v24.18.0
●  Dependencies    fail  Dependency issues detected
●                    No duplicate singleton dependencies
●                    DB driver "pg" not installed (required for dbConnectionOptions.type: "postgres")
●  Config          pass  Vendure config loaded and validated successfully
●                    Config loaded from vendure-config.ts
●                    Custom fields validated
●                    Plugin configuration completed
●  Schema          pass  Admin and Shop API schemas validated successfully
●                    Admin API schema validated successfully
●                    Shop API schema validated successfully
●  Database        pass  Successfully connected to postgres database
●                    Database type: postgres
■  Result: failed (1 failure)
```

- @vendure/core version: 3.7.2 (`@vendure/cli`: 3.7.2 — the package containing the bug)
- Nodejs version: v24.18.0
- Operating System (Windows/macOS/Linux): Windows 11 — the faulty logic is a plain `fs.existsSync` path check with no OS-specific behavior, so it should reproduce identically on macOS/Linux
- Package manager (npm/yarn/pnpm): npm workspaces (npm 12.0.1); the root cause (no module-resolution lookup) is not npm-specific, so yarn/pnpm workspaces likely reproduce it too

**Configuration**

```typescript
// apps/server/src/vendure-config.ts (minimal, generic example)
    dbConnectionOptions: {
        type: 'postgres',
        // host/port/credentials ...
    },
    // ...
};
```

```json
// root package.json
{
  "name": "my-monorepo",
  "private": true,
  "workspaces": ["apps/*"]
}
```

```json
// apps/server/package.json (excerpt)
{
  "name": "server",
  "dependencies": {
    "pg": "^8.x",
    "@vendure/core": "3.7.2"
  }
}
```

**Minimal reproduction**

Root cause, found by inspecting the installed `@vendure/cli@3.7.2` package:

- `dist/commands/doctor/checks/dependency-check.js` — `checkDbDriver(cwd, modulesDir)` resolves the driver's `package.json` with:

  ```js
  const driverPath = path.join(modulesDir, driverPkg, 'package.json');
  if (fs.existsSync(driverPath)) {
      // pass
  } else {
      // fail: `DB driver "${driverPkg}" not installed ...`
  }
  ```

  where `modulesDir` defaults to `path.join(process.cwd(), 'node_modules')`. This never uses Node's module resolution algorithm (e.g. `require.resolve(driverPkg, { paths: [cwd] })`), so a driver hoisted to a workspace root (or any ancestor `node_modules`) is invisible to this check. The same fixed-`modulesDir` pattern is also used by the duplicate-singleton and installed-Vendure-version checks in the same file (`findDuplicatePackages`, `getInstalledVendureVersions`), so they're likely susceptible to the same false-positive class.

- `dist/commands/doctor/doctor.js` calls `runDependencyCheck()` with no arguments, so `modulesDir` always falls back to `cwd/node_modules`.

- The CLI already has monorepo detection — `dist/utilities/monorepo-utils.js`'s `detectMonorepoStructure()` — and `dist/commands/doctor/checks/project-check.js` uses it (that's what prints `Monorepo detected (apps)` and picks the right directory for lockfile discovery). But the detected root is only used locally inside `project-check.js`; it's never returned from that check or forwarded into `runDependencyCheck()`, so the dependency check can't benefit from information the CLI has already computed earlier in the same `doctor` run.

**Workaround**

None needed to actually run the app — `vendure migrate`, `dev:server`, `build:server`, and the running server all work correctly, since the driver genuinely resolves at runtime through normal Node module resolution. The only impact is that `vendure doctor` (and `doctor --strict`, used as a CI/production gate) reports a false failure and exits non-zero. Teams using `doctor`/`doctor:prod` as a CI gate currently have to skip or otherwise tolerate this specific failure.

**Additional context**

- Encountered while converting a Vendure project into an npm workspaces monorepo (moving the Vendure app from the repo root into `apps/<name>` so a second, unrelated app could live alongside it). Every other verification (server build, dashboard typecheck, a full live server+worker boot, DB connectivity, `vendure migrate`) passed correctly after the move — this is isolated to `doctor`'s dependency-check step.
- Reproduces consistently, every `doctor` run, as long as the driver package is hoisted rather than present in the app's own `node_modules`.
- Suggested fix direction: have `checkDbDriver` (and the other checks in `dependency-check.js`) resolve packages via `require.resolve(pkg, { paths: [cwd] })` (or an equivalent upward walk), the same way Node itself resolves the import at runtime — rather than a single fixed-directory `fs.existsSync` probe. Alternatively, thread the monorepo root already computed in `project-check.js` through to `runDependencyCheck()` so it can check both `cwd/node_modules` and the workspace root.

Ryrahul:
Fixes #5140

# Description

Fixes `vendure doctor` falsely reporting DB driver packages as not installed in monorepo workspaces where dependencies are hoisted to the workspace root `node_modules`.

**Root cause:** `checkDbDriver()` used `fs.existsSync(path.join(cwd, 'node_modules', driverPkg))` — a fixed-directory probe that can't find packages hoisted to an ancestor `node_modules`. The same issue affected `getInstalledVendureVersions()` and `findDuplicatePackages()`.

**Fix:**
- `checkDbDriver()` now uses `createRequire().resolve()` —  which walks up ancestor directories and correctly finds hoisted packages.
- `getInstalledVendureVersions()` and `findDuplicatePackages()` now accept multiple `node_modules` directories (local + monorepo root) so hoisted `@vendure/*` packages and singleton dependencies are found.
- The monorepo root detected by the project check is now threaded into the dependency check via `monorepoRoot` on `CheckResult`.

# Breaking changes

None. The `runDependencyCheck()` function signature changed from `(nodeModulesPath?: string)` to `(options?: DependencyCheckOptions)` but this is an internal API not exposed to consumers.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5141"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789378395&installation_model_id=20193&pr_number=5141&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5141&signature=d20d90f3c12056d90412be25ad438e696aef3b9b5f88dddc334f50f068a78f82"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
+++ b/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
@@ -135,4 +135,58 @@ describe('dependency-check', () => {
 
         expect(result.details?.some(d => d.includes('No duplicate singleton dependencies'))).toBe(true);
     });
+
+    it('finds @vendure/* packages hoisted to monorepo root', async () => {
+        vi.mocked(fs.existsSync).mockImplementation((p: any) => {
+            const pathStr = String(p);
+            // Local node_modules exists but has no @vendure packages
+            if (pathStr === '/app/node_modules') return true;
+            // Monorepo root node_modules exists
+            if (pathStr === '/root/node_modules') return true;
+            // @vendure packages only at root
+            if (pathStr.startsWith('/root/node_modules/@vendure')) return true;
+            return false;
+        });
+        vi.mocked(fs.readJsonSync).mockImplementation((p: any) => {
+            const pathStr = String(p);
+            if (pathStr.startsWith('/root/node_modules/@vendure')) {
+                return { version: '3.7.2' };
+            }
+            return {};
+        });
+        vi.mocked(fs.readdirSync).mockReturnValue([]);
+        vi.mocked(fs.readFileSync).mockReturnValue('');
+
+        const result = await runDependencyCheck({
+            nodeModulesPath: '/app/node_modules',
+            monorepoRoot: '/root',
+        });
+
+        expect(result.details?.some(d => d.includes('All @vendure/* packages at 3.7.2'))).toBe(true);
+    });
+
+    it('finds DB driver via require.resolve when hoisted', async () => {
+        // Mock createRequire to return a resolver that finds the driver
+        const { createRequire: mockCreateRequire } = await import('node:module');
+        vi.mocked(mockCreateRequire).mockReturnValue({
+            resolve: vi.fn().mockReturnValue('/root/node_modules/pg/package.json'),
+        } as any);
+
+        vi.mocked(fs.existsSync).mockReturnValue(true);
+        vi.mocked(fs.readJsonSync).mockImplementation((p: any) => {
+            if (String(p) === '/root/node_modules/pg/package.json') {
+                return { version: '8.20.0', name: 'pg' };
+            }
+            return { version: '3.7.2' };
+        });
+        vi.mocked(fs.readdirSync).mockReturnValue([]);
+        // Config file with postgres type
+        vi.mocked(fs.readFileSync).mockReturnValue(
+            'dbConnectionOptions: { type: \'postgres\' }',
+        );
+
+        const result = await runDependencyCheck({ nodeModulesPath: '/app/node_modules' });
+
+        expect(result.details?.some(d => d.includes('DB driver pg') && d.includes('8.20.0'))).toBe(true);
+    });
 });
--- a/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
+++ b/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
@@ -26,13 +26,22 @@ describe('dependency-check', () => {
         fs.mkdirpSync(appDir);
         fs.writeJsonSync(path.join(appDir, 'package.json'), {
             name: 'api',
-            dependencies: { pg: '8.20.0' },
+            dependencies: { '@vendure/core': '3.7.2', pg: '8.20.0' },
         });
         // Config file that references postgres
         fs.writeFileSync(
             path.join(appDir, 'vendure-config.ts'),
             'export const config = { dbConnectionOptions: { type: \'postgres\' } };',
         );
+        // @vendure/core hoisted to workspace root (needed so the "not installed" early return doesn't fire)
+        const corePkgDir = path.join(tmpDir, 'node_modules', '@vendure', 'core');
+        fs.mkdirpSync(corePkgDir);
+        fs.writeJsonSync(path.join(corePkgDir, 'package.json'), {
+            name: '@vendure/core',
+            version: '3.7.2',
+            main: 'index.js',
+        });
+        fs.writeFileSync(path.join(corePkgDir, 'index.js'), '');
         // pg hoisted to workspace root
         fs.mkdirpSync(path.join(tmpDir, 'node_modules', 'pg'));
         fs.writeJsonSync(path.join(tmpDir, 'node_modules', 'pg', 'package.json'), {
--- a/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
+++ b/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
@@ -12,6 +12,15 @@ vi.mock('fs-extra', () => ({
     },
 }));
 
+// Mock node:module's createRequire (used by checkDbDriver)
+vi.mock('node:module', () => ({
+    createRequire: vi.fn(() => ({
+        resolve: vi.fn(() => {
+            throw new Error('Cannot find module');
+        }),
+    })),
+}));
+
 import fs from 'fs-extra';
 
 describe('dependency-check', () => {
@@ -26,7 +35,7 @@ describe('dependency-check', () => {
     it('returns fail when node_modules does not exist', async () => {
         vi.mocked(fs.existsSync).mockReturnValue(false);
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.status).toBe('fail');
         expect(result.message).toContain('node_modules not found');
@@ -38,7 +47,7 @@ describe('dependency-check', () => {
         vi.mocked(fs.readdirSync).mockReturnValue([]);
         vi.mocked(fs.readFileSync).mockReturnValue('');
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.status).toBe('pass');
         expect(result.details?.some(d => d.includes('All @vendure/* packages at 3.6.3'))).toBe(true);
@@ -52,14 +61,13 @@ describe('dependency-check', () => {
         let callCount = 0;
         vi.mocked(fs.readJsonSync).mockImplementation(() => {
             callCount++;
-            // Return different patch version for one package
             if (callCount === 3) {
                 return { version: '3.6.2' };
             }
             return { version: '3.6.3' };
         });
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.status).toBe('warn');
         expect(result.details?.some(d => d.includes('Mismatched') && d.includes('patch'))).toBe(true);
@@ -73,14 +81,13 @@ describe('dependency-check', () => {
         let callCount = 0;
         vi.mocked(fs.readJsonSync).mockImplementation(() => {
             callCount++;
-            // Return different minor version for one package
             if (callCount === 3) {
                 return { version: '3.7.0' };
             }
             return { version: '3.6.3' };
         });
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.status).toBe('fail');
         expect(result.details?.some(d => d.includes('Mismatched') && d.includes('minor/major'))).toBe(true);
@@ -89,13 +96,9 @@ describe('dependency-check', () => {
     it('detects duplicate singleton packages', async () => {
         vi.mocked(fs.existsSync).mockImplementation((p: any) => {
             const pathStr = String(p);
-            // node_modules exists
             if (pathStr === '/fake/node_modules') return true;
-            // @vendure dir exists
             if (pathStr.includes('@vendure')) return true;
-            // root graphql exists
             if (pathStr === '/fake/node_modules/graphql/package.json') return true;
-            // nested graphql exists
             if (pathStr.includes('msw/node_modules/graphql/package.json')) return true;
             return false;
         });
@@ -116,7 +119,7 @@ describe('dependency-check', () => {
         });
         vi.mocked(fs.readFileSync).mockReturnValue('');
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.status).toBe('warn');
         expect(result.details?.some(d => d.includes('Multiple graphql versions'))).toBe(true);
@@ -128,7 +131,7 @@ describe('dependency-check', () => {
         vi.mocked(fs.readdirSync).mockReturnValue([]);
         vi.mocked(fs.readFileSync).mockReturnValue('');
 
-        const result = await runDependencyCheck('/fake/node_modules');
+        const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });
 
         expect(result.details?.some(d => d.includes('No duplicate singleton dependencies'))).toBe(true);
     });
--- a/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
+++ b/packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
@@ -1,192 +1,189 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import fs from 'fs-extra';
+import os from 'node:os';
+import path from 'node:path';
 
 import { runDependencyCheck } from './dependency-check';
 
-// Mock fs-extra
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/doctor/checks/dependency-check.ts
```diff
@@ -303,19 +324,24 @@ function checkDbDriver(
...
+        // ancestor node_modules directories -- correctly finding packages
+        // hoisted to a workspace root.
+        const localRequire = createRequire(path.join(cwd, 'package.json'));
+        const resolvedPath = localRequire.resolve(`${driverPkg}/package.json`);
+        const version = readPackageVersion(resolvedPath);
         return {
             status: 'pass',
             message: `DB driver ${driverPkg}${version ? ` (${version})` : ''} installed for type "${dbType}"`,
         };
+    } catch {
```
biggamesmallworld, line 416: `ERR_PACKAGE_PATH_NOT_EXPORTED` means the package was found but does not export `./package.json` (mysql2 3.0.0-3.5.x). Reporting that as not installed is the bug this PR fixes, by another route.

```suggestion
    } catch (e) {
        // Some packages (e.g. mysql2 < 3.6.0) ship an `exports` map with no
        // "./package.json" entry, so resolving that subpath throws even though
        // the package is installed.
        if ((e as NodeJS.ErrnoException).code === 'ERR_PACKAGE_PATH_NOT_EXPORTED') {
            return {
                status: 'pass',
                message: `DB driver ${driverPkg} installed for type "${dbType}"`,
            };
        }
```

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -56,7 +59,15 @@ export async function doctorCommand(options?: DoctorOptions) {
 
     // Check 2: Dependency version alignment, singleton duplication, DB driver
     if (checksToRun.includes('dependencies')) {
-        results.push(await runDependencyCheck());
+        // If the project check didn't run, detect monorepo root directly
+        // so hoisted dependencies are still found.
+        if (!monorepoRoot) {
+            const monorepoInfo = detectMonorepoStructure(process.cwd());
```
biggamesmallworld: `detectMonorepoStructure` matches the path string alone, so `/home/me/apps/my-shop` returns root `/home/me` and `runDependencyCheck` then reads `/home/me/node_modules`. Check the candidate root for a workspace marker before using it, and walk up to the nearest one.

packages/cli/src/commands/doctor/checks/dependency-check.spec.ts
```diff
@@ -12,6 +12,15 @@ vi.mock('fs-extra', () => ({
     },
 }));
 
+// Mock node:module's createRequire (used by checkDbDriver)
+vi.mock('node:module', () => ({
```
biggamesmallworld: This mock is never reached: `readFileSync` returns `''` in every test, so `checkDbDriver` exits at `dbType === undefined` before calling `createRequire`. Neither of its branches is covered. This test covers the hoisted case and passes on this head (needs `import { createRequire } from 'node:module';` at the top):

```ts
it('finds a DB driver hoisted to an ancestor node_modules', async () => {
    const resolve = vi.fn((id: string) => {
        if (id === 'pg/package.json') {
            return '/workspace/node_modules/pg/package.json';
        }
        throw new Error('Cannot find module');
    });
    vi.mocked(createRequire).mockReturnValue({ resolve } as any);
    vi.mocked(fs.existsSync).mockImplementation((p: any) => {
        const pathStr = String(p);
        return (
            pathStr === '/fake/node_modules' ||
            pathStr.endsWith('vendure-config.ts') ||
            pathStr === '/workspace/node_modules/pg/package.json'
        );
    });
    vi.mocked(fs.readdirSync).mockReturnValue([]);
    vi.mocked(fs.readFileSync).mockReturnValue('dbConnectionOptions: { type: "postgres" }');
    vi.mocked(fs.readJsonSync).mockReturnValue({ version: '8.13.0' });

    const result = await runDependencyCheck({ nodeModulesPath: '/fake/node_modules' });

    expect(result.details?.some(d => d.includes('DB driver pg (8.13.0) installed'))).toBe(true);
});
```

A second test passing `{ monorepoRoot: '/workspace' }` would cover the other half.

biggamesmallworld (changes requested):
Diagnosis is right and `createRequire().resolve()` is the correct mechanism. Three things before it lands.

**1. The new resolve fails on packages whose `exports` map omits `./package.json`.** `mysql2` 3.0.0-3.5.x is one: resolve throws `ERR_PACKAGE_PATH_NOT_EXPORTED` and the bare `catch` reports the driver as not installed. For a project on `mysql2@3.5.x` that is a regression, since `existsSync` finds it today. `pg`, `better-sqlite3` and `sql.js` are unaffected at current versions. Suggestion inline.

**2. `monorepoRoot` is a path-substring guess that now selects directories to read.** Any path containing `/packages/`, `/apps/`, `/libs/`, `/services/` or `/modules/` counts, and the root is the prefix before the first match, so `/home/me/apps/my-shop` yields `/home/me`. If `~/node_modules` exists, its `graphql` is unioned into `findDuplicatePackages` and the check turns from `No duplicate singleton dependencies` into `Multiple graphql versions found`. Require a workspace marker at the candidate root (`pnpm-workspace.yaml`, `workspaces` in `package.json`, `lerna.json`, `nx.json`, `turbo.json`), and take the nearest match rather than the first split.

**3. Nothing new is tested.** The six changed tests only swap the argument shape, and the added `node:module` mock is never reached: `readFileSync` returns `''`, so `dbType` is `undefined` and `checkDbDriver` returns before `createRequire`. I made the mock throw on call and all six still passed. No test passes `monorepoRoot` either. A test for the case this PR exists to fix is inline.

Minor: `monorepoRoot` on `CheckResult` is serialised by `formatJsonReport`, so `doctor --json` now prints an absolute path like `/Users/alice/dev/acme-shop`. Every other path in that report goes through `path.relative(cwd, ...)`.

Ran at `d642d6c`: `tsc --noEmit -p packages/cli/tsconfig.json` clean for the changed files, `vitest run src/commands/doctor` 46/46, `eslint` no new errors.

biggamesmallworld (changes requested):
Closer. Two things from last round are still open, plus one new bug.

**1. The new code isn't tested.** Both behaviours added in `b1792ebb` are uncovered. I checked by breaking them:

- `throw` at the top of the `exports` fallback in `checkDbDriver`: 48/48 pass.
- `hasWorkspaceMarker` hardcoded to `true`: 48/48 pass.
- `formatJsonReport` reverted to printing the raw report: 48/48 pass.

The two tests you did add are real, they fail if I revert the driver lookup to `existsSync` or drop the root `node_modules` from the union. But the driver test mocks `createRequire` to hand back a fixed path, so it asserts the mock rather than the thing under test, which is that Node walks up ancestor `node_modules`. A temp directory tests it for real:

```ts
const root = fs.mkdtempSync(path.join(os.tmpdir(), 'doctor-'));
fs.outputJsonSync(path.join(root, 'node_modules/pg/package.json'), { name: 'pg', version: '8.20.0' });
// resolve from path.join(root, 'packages/api'), assert 8.20.0
```

Please also cover the fallback branch (a package whose `exports` map omits `./package.json`) and `hasWorkspaceMarker` in both directions.

**2. `readJsonSync` in the fallback catch reports an installed driver as missing.**

```ts
const pkg = fs.readJsonSync(pkgJsonPath);
```

That throws on a malformed `package.json` anywhere on the walk up, and lands in the catch returning `DB driver "x" not installed`. `require.resolve` has already succeeded by then, so the driver is installed. Use `readPackageVersion`, which swallows exactly this, or move the resolve out of the try.

**3. `hasWorkspaceMarker` is copy-pasted into `doctor.ts` and `project-check.ts`.** Move it to `utilities/monorepo-utils.ts`.

More to the point, the substring bug is still in `detectMonorepoStructure`, and `ast-utils.ts:86` calls it and gets the same wrong root. The guard rejects a bad root but doesn't find the right one: `/home/me/apps/acme/libs/api` still resolves to `/home/me`, fails the marker check, and the hoisting fix silently doesn't apply. Walking up from cwd to the nearest marker fixes the utility for all three callers, and then the guard isn't needed at the call sites.

A question rather than a request: does the `require.resolve` approach in `checkDbDriver` make the local + root `node_modules` list redundant? It works for `@vendure/*` and the singletons too, and it handles nested workspaces and pnpm layouts that "local plus root" misses. If it does, `monorepoRoot` on `CheckResult`, `DependencyCheckOptions.monorepoRoot`, both copies of `hasWorkspaceMarker` and the json formatter change all go away. Worth deciding before you write more tests around the two-directory version.

Last thing: `conventional commit format` is failing on the PR title. Lowercase the subject, `fix(cli): resolve hoisted dependencies in monorepo workspaces`. The dashboard e2e failure is unrelated flake.

Ran at `b1792ebb`: `vitest run src/commands/doctor` 48/48, mutations as above. `tsc --noEmit` and `eslint` show no new problems in the changed files.

Ryrahul:
Thanks for the detailed review  all three points are valid, and I'll address them.

### Point 4

On your question regarding point 4: `require.resolve` makes most of the two-directory machinery redundant, so I'll go that route:

* **Version alignment (`@vendure/*`) and DB driver checks** will both use `require.resolve` from `cwd`. This is strictly more correct because it handles nested workspaces and pnpm layouts that the "local + root" approach can miss, while also resolving the exact version Node loads.
* **Duplicate singleton detection** is the one case where `require.resolve` isn't sufficient, since it returns a single resolved path rather than all installed copies. That will remain a tree scan, rooted via a fixed `detectMonorepoStructure` instead of the current substring-based guess.

This will allow me to remove:

* `DependencyCheckOptions.monorepoRoot`
* `monorepoRoot` from `CheckResult`
* Both `hasWorkspaceMarker` copies
* The JSON formatter workaround

### `detectMonorepoStructure`

Agreed on the substring bug. I'll rewrite it to walk upward from `cwd` and find the nearest directory containing a workspace marker:

Note: the substring bug in detectMonorepoStructure is pre-existing — it's called via getTsMorphProject() → analyzeProject() by every CLI command (add, migrate, schema, doctor), so all of them have the same false-positive risk . And will change this behaviour for all this callsite

* `pnpm-workspace.yaml`
* `workspaces` in `package.json`
* `lerna.json`
* `nx.json`
* `turbo.json`

This fixes the utility for all three callers, including `ast-utils.ts`, and removes the need for a guard at the call sites.

### `readJsonSync` Fallback

Good catch on `readJsonSync` in the fallback `catch`. `require.resolve` has already succeeded by that point, so a malformed `package.json` encountered during the walk-up shouldn't cause the driver to be reported as missing.


I'll switch that logic to `readPackageVersion`.

### Tests

You're right that the current `createRequire` mock only asserts the mock behavior rather than the actual walk-up behavior. Had asked claude to write it up so completely missed from my side to verify that , my bad 

I'll switch the tests to use real temporary directories, including a case where a hoisted `pg` is resolved from a nested subdirectory. I'll also add coverage for:

* The exports-map fallback branch
* `hasWorkspaceMarker` in both directions

I'll also lowercase the PR title to follow conventional commits.

### One Question

For the duplicate scan, do you have a preference between:

1. Rooting it at the detected monorepo root, or
2. Rooting it at the resolved `@vendure/core` location and scanning outward?

Happy to go with whichever you find cleaner to review.

biggamesmallworld (approved):
Nice work!

Issue #3098, opened 2024-10-01
Discount calculation on line discount is not as expected when modify line quantity


TeikSean:
**Describe the bug**
The discount calculation is not as expected when i try to modify the order line quantity (to lesser amount). For example as the screenshot shown, the order discount i apply is 10% which after modify the quantity to 7 it should be $279.30 instead of $244.39
<img width="945" alt="Screenshot 2024-10-01 at 3 03 44 PM" src="https://github.com/user-attachments/assets/ef02d14a-a09f-4e67-bc4f-dffb38ad4eeb">

I believe this is due to the unitAdjustmentAmount which use Math.max(this.orderPlacedQuantity, this.quantity) which cause the discount to be calculated as not expected, not sure the use case to find the max quantity
<img width="1089" alt="Screenshot 2024-10-01 at 3 05 39 PM" src="https://github.com/user-attachments/assets/5073f7cd-7841-4a79-8d24-64f4a1ba0f3d">

**To Reproduce**
1. Go to Promotions and add a promotion which condition order total greater than 0 and action discount order by 10%
<img width="815" alt="Screenshot 2024-09-27 at 6 10 30 PM" src="https://github.com/user-attachments/assets/b8b13060-5316-4c92-bc52-69b509b4450f">

2. Create a product with price 399.00

3. Create an order with 8 quantity of the product, add customer, addresses and shipping method, then create draft order

4. Add payment to order

<img width="1130" alt="Screenshot 2024-10-01 at 3 10 21 PM" src="https://github.com/user-attachments/assets/5d2b2a62-3c05-4a42-9121-fba3bfca9981">

5. Modify order, can edit the quantity from 8 to 7, then click on preview changes, you should see as the screenshot attached

<img width="930" alt="Screenshot 2024-10-01 at 3 11 01 PM" src="https://github.com/user-attachments/assets/041f599d-b511-49e2-b69d-ffe9ae7fe497">

**Expected behavior**
The discount after edit the quantity should be $279.30 instead of $244.39 

**Environment (please complete the following information):**
 - @vendure/core version: v3.0.0
 - Nodejs version: 18.20
 - Database (mysql/postgres etc): Postgres 15.3

**Additional context**
This might related to my previous issue ticket https://github.com/vendure-ecommerce/vendure/issues/3047
https://github.com/vendure-ecommerce/vendure/issues/3091

PR #5143, opened 2026-08-16
fix(core): reconcile read/write quantity basis for OrderLine discounts (#5127)

### LINKED ISSUE #5127: Promotion discount is scaled by quantity/orderPlacedQuantity after reducing an OrderLine quantity (continuation of #3098)
**Describe the bug**

When an `OrderLine`'s quantity is reduced after the order has been placed (partial cancellation / refund via order modification), the promotion discount applied to that line is silently scaled down by `quantity / orderPlacedQuantity`. A 40% promotion becomes 4% on a line reduced from 20 units to 2.

This is the same underlying bug as #3098 (reported against v3.0.0, closed as stale in June 2025, no fix attached). It is still reproducible on **v3.6.1**. The reporter of #3098 already identified the cause: the `Math.max(this.orderPlacedQuantity, this.quantity)` divisor.

This report adds two things #3098 did not have:

1. A deterministic ratio proof: `actual / expected === quantity / orderPlacedQuantity`.
2. The reason a plugin **cannot** work around it in userland: `OrderLine.addAdjustment()` caps the adjustment on a *different* quantity basis than the one used to read it back, so an action that pre-compensates for the divisor has its amount clamped.

The two bases are:

```ts
// addAdjustment() — WRITE side: capped on `quantity`
const maxDiscount = (this.listPriceIncludesTax ? this.proratedLinePriceWithTax : this.proratedLinePrice) * -1;

// getUnitAdjustmentsTotal() — READ side: divided by max(orderPlacedQuantity, quantity)
adjustment.amount / Math.max(this.orderPlacedQuantity, this.quantity)

// getLineAdjustmentsTotal() — READ side: same
sum * (this.quantity / Math.max(this.orderPlacedQuantity, this.quantity))
```
So an adjustment is limited on creation using one quantity basis and interpreted using another.

**To Reproduce**

1. Create a Promotion with condition "order total greater than 0" and an action that discounts by a percentage. Reproducible both with the built-in `order_percentage_discount` (as in #3098) and with a custom `PromotionLineAction` (see Minimal reproduction below). Use 40%.
2. Create a product variant priced at 29.75 (tax-inclusive channel).
3. Place an order with **quantity 20** of that variant, so the promotion applies. Confirm the discount is 40%.
4. Settle the payment so the order leaves the active state (this is what sets `orderPlacedQuantity = 20`).
5. Modify the order and reduce that line's quantity from **20 to 2**. Click "Preview changes".
6. Observe the line discount: it is **4%**, not 40%.

**Expected behavior**

The remaining units should keep the configured discount: `2 × 29.75 × 40% = 23.80`, giving a discounted line price of `35.70`.

**Actual behavior**

The line discount is `2.38`, giving a discounted line price of `57.12` — i.e. **4%** instead of 40%.

The error is not random; it is exactly the quantity ratio:

```
actual / expected === quantity / orderPlacedQuantity

#3098 (order-level promo):   244.39 / 279.30 = 0.875 =  7 / 8   -> 10% became  8.75%
this report (line-level), A:   2.38 /  23.80 = 0.10  =  2 / 20  -> 40% became  4%
this report (line-level), B: 541.50 / 570.00 = 0.95  = 19 / 20  -> 40% became 38%
```
Lines whose quantity was never modified (`orderPlacedQuantity === quantity`) are unaffected, which is why this only shows up on partially cancelled/refunded orders.

**Screenshots/Videos**

N/A — reproducible from the numbers above.

**Error logs**

No error is thrown; the discount is silently wrong. Instrumenting a `PromotionLineAction` shows the action returning the correct amount while the persisted adjustment is smaller:

```
[B2BDiscountAction] inputs: line=54155 pct=40 qty=2 orderPlacedQuantity=20
  listPrice=2975 listPriceIncludesTax=true unitPriceWithTax=2975
  lineTotal=59500 => returned amount=-23800
```
The action returned `-23800` (40% of 20 units), but the stored adjustment is `-5950`:

```
 id    | quantity | orderPlacedQuantity | listPrice | adjustments
-------+----------+---------------------+-----------+--------------------------------------------------
 54155 |        2 |                  20 |      2975 | [{"amount":-5950,"type":"PROMOTION",...}]
 54160 |       19 |                  20 |      7500 | [{"amount":-60000,"type":"PROMOTION",...}]
```
`-5950` is exactly `proratedLinePriceWithTax` for the *current* quantity (`2 × 2975`) — the `addAdjustment()` cap. Read back through the `max(orderPlacedQuantity, quantity)` divisor it yields `5950 / 20 = 297.5` per unit ≈ **9.98%**.

Line `54160` (19 of 20) is under the cap (`0.4 × 20 = 8 ≤ 19`), so it stores the full `-60000` and renders a correct 40% — which is why the bug appears inconsistent across lines of the same order.

- @vendure/core version: `3.6.1`
- Nodejs version: `<fill in: node -v>`
- Database (mysql/postgres etc): `postgres <fill in version>`
- Operating System (Windows/macOS/Linux): `Linux`
- Browser (if applicable): `<fill in>`
- Package manager (npm/yarn/pnpm): `npm`

**Configuration**

Nothing unusual is required. The channel is tax-inclusive (`pricesIncludeTax: true`, `listPriceIncludesTax` is `true` on the lines), and the default `OrderPlacedStrategy` / `orderItemPriceCalculationStrategy` are in use.

```typescript
// Only the promotion action registration is relevant:
promotionOptions: {
    promotionActions: [...defaultPromotionActions, linePercentageDiscount],
},
```
**Minimal reproduction**

A plain percentage line action is enough to reproduce:

```typescript
import { LanguageCode, PromotionLineAction } from '@vendure/core';

export const linePercentageDiscount = new PromotionLineAction({
    code: 'line_percentage_discount',
    description: [{ languageCode: LanguageCode.en, value: 'Discount line by { discount }%' }],
    args: { discount: { type: 'int', ui: { component: 'number-form-input', suffix: '%' } } },
    execute(ctx, orderLine, args) {
        const unitPrice = orderLine.listPriceIncludesTax ? orderLine.unitPriceWithTax : orderLine.unitPrice;
        return -(unitPrice * orderLine.quantity) * (args.discount / 100);
    },
});
```
Place an order with quantity 20, settle payment, then modify the line down to quantity 2 and preview. The discount will be 4%.

Attempting to compensate in userland by returning `-(unitPrice * Math.max(orderLine.orderPlacedQuantity, orderLine.quantity)) * (discount / 100)` does **not** work: the required `-23800` exceeds the `addAdjustment()` cap of `-5950` and is clamped, yielding ~10%. The maximum discount representable on such a line is therefore `quantity / orderPlacedQuantity` of the intended percentage.

**Workaround**

Two changes are needed together, since fixing either alone is insufficient:

1. In the promotion action, base the amount on `Math.max(orderPlacedQuantity, quantity)` so it matches the read-side divisor.
2. Patch `addAdjustment()` so the cap uses the same basis. Basing it on the *remaining* prorated unit value rather than `listPrice` preserves the existing protection when adjustments stack:

```diff
     addAdjustment(adjustment) {
-        const maxDiscount = (this.listPriceIncludesTax ? this.proratedLinePriceWithTax : this.proratedLinePrice) * -1;
+        const divisor = Math.max(this.orderPlacedQuantity, this.quantity);
+        const remainingUnitValue = this.listPriceIncludesTax
+            ? this._proratedUnitPriceWithTax()
+            : this._proratedUnitPrice();
+        const maxDiscount = this.quantity === 0 ? 0 : remainingUnitValue * divisor * -1;
         const limitedAdjustment = { ...adjustment, amount: Math.max(maxDiscount, adjustment.amount) };
```
This is a no-op when `orderPlacedQuantity === quantity`, keeps zero-quantity lines at `0`, and still prevents the unit price going negative (`amount / divisor ≤ remainingUnitValue`).

Note that the alternative — changing the read-side divisor to `quantity`, as suggested in #3098 — is the cleaner model but is not backwards compatible for existing data: adjustments already stored against `orderPlacedQuantity` would suddenly be divided by a smaller quantity and render as inflated (up to 100%) discounts until every affected order is recalculated. A migration/recalculation story would be needed.

**Additional context**

- When did this start: it is not a regression; #3098 reports it on v3.0.0 and it is unchanged on master (the `Math.max(...)` divisor is still present).
- Consistency: fully deterministic — it occurs on every line where `quantity < orderPlacedQuantity` and a promotion applies.
- Production consequence: the wrong percentage propagates into invoicing/ERP exports. In our case a B2B order exported line discounts of `4` and `38` instead of `40`, so the ERP received figures that did not match the contractual customer discount.
- Related: #3098 (same root cause, closed stale), #4811 (order-level discount redistribution after cancellation — cites #3098 and the same `getUnitAdjustmentsTotal` / `getLineAdjustmentsTotal` behaviour), #5097 / #5101 (divide-by-zero in the same `discounts` getter, fixed).
- I'm happy to open a PR with the `addAdjustment()` fix plus unit coverage for the reduced-quantity case if the maintainers agree with the approach.

---

wakqasahmed:
# Description

Fixes #5127. Continuation of #3098.

## The bug

When an `OrderLine`'s quantity is reduced below `orderPlacedQuantity` (a partial
cancellation, or reducing a line's quantity via a `modifyOrder` order
modification), the applied promotion discount was silently rescaled — a 40%
discount could collapse to ~4% after cutting quantity from 20 to 2, exactly as
described in #5127 and, for the order-level case, in #3098.

## Root cause: two mismatched quantity bases for the same `adjustment.amount`

`OrderLine.addAdjustment()` (write side) and `getUnitAdjustmentsTotal()` /
`getLineAdjustmentsTotal()` (read side) assumed the same `Adjustment.amount`
meant different things depending on *which* code path last touched the
line's quantity, but the getters didn't know which:

- **`PROMOTION` adjustments** (item- and line-level `PromotionAction`s) are
  always freshly recomputed against the `OrderLine`'s **current** quantity —
  `OrderCalculator.applyOrderItemPromotions()` calls `line.clearAdjustments()`
  and rebuilds them via `addAdjustment()` on every recalculation
  (`order-calculator.ts:196-217`). So the stored total is already correctly
  scaled to the current quantity. Reading it back by dividing by
  `Math.max(orderPlacedQuantity, quantity)` (the pre-fix code) **double-scaled**
  it whenever a quantity reduction triggered a fresh recalculation (e.g. via
  `modifyOrder`), on top of the correct scaling already baked into the stored
  amount — this is the exact bug reported in #5127.

- **`DISTRIBUTED_ORDER_PROMOTION` adjustments** (an order-level promotion's
  per-line share, from `applyOrderPromotions`) are redistributed by *weight*
  across lines on every recalculation, and a single remaining line can end up
  freshly assigned the *entire* order-level discount regardless of its own
  quantity. Dividing by `orderPlacedQuantity` here is a **deliberate,
  pre-existing mechanism** that keeps a partially cancelled/refunded line's
  per-unit share pinned to what it was at placement, so its surviving units
  don't absorb a windfall meant for units that no longer exist. I confirmed
  this is intentional (not part of the bug) by initially "fixing" it uniformly
  and immediately breaking the existing `refunds correct amount when
  order-level promotion applied` e2e test — this behaviour is now explicitly
  preserved and left untouched.

- **`OrderModifier.cancelOrderByOrderLines()`** reduces `quantity` directly via
  a raw repository `.update()`, without going through `OrderCalculator` at
  all — so `PROMOTION`-type adjustments were left stale at the
  pre-cancellation total after a partial cancellation. This method now
  rescales `PROMOTION`-type adjustments (only) proportionally to the new
  quantity when cancelling, so the same "already current-quantity-scaled"
  invariant holds after a partial cancellation as it does after a
  `modifyOrder`-triggered recalculation.

## The fix

`getUnitAdjustmentsTotal()`, `getLineAdjustmentsTotal()` and the `discounts`
getter now pick the quantity basis **per adjustment**, based on its `type`:
`this.quantity` for `PROMOTION`, `Math.max(orderPlacedQuantity, quantity)` for
everything else (`DISTRIBUTED_ORDER_PROMOTION` and `OTHER`) — via a new
`quantityBasisFor()` helper. This makes the read side always match whichever
basis the write side actually used for that adjustment type, instead of
assuming one basis for all adjustments.

`OrderModifier.cancelOrderByOrderLines()` now rescales `PROMOTION`-type
adjustments to the new quantity, matching what a full recalculation would
have produced. `DISTRIBUTED_ORDER_PROMOTION`/`OTHER` adjustments are left
untouched there, consistent with their read-side basis.

I considered instead moving *all* proration to write time (rescale
everything on cancellation, always divide by current quantity on read), which
would have been a smaller code diff, but it collapses the deliberate
order-level "pinned to placement" behaviour above and breaks that existing,
passing e2e test — so I ruled it out in favour of the type-aware fix, which
is a few more lines but changes zero externally-observed behaviour outside of
the actual bug.

## Tests

- `order-calculator.spec.ts`: new unit test reproducing the reported 40% → ~4%
  discount decay after a `modifyOrder`-style recalculation at a reduced
  quantity (fails without the fix, passes with it).
- `order-modification.e2e-spec.ts`: two new e2e tests — one going through
  `modifyOrder` (adjustOrderLines + refund), one through a partial
  `cancelOrder({ lines: [...] })` — both reducing quantity from 20 to 2 under
  a 50% item promotion and asserting the discount ratio survives. The
  `modifyOrder` test fails on the pre-fix code (`expected 296172 to be
  155880`); the `cancelOrder` test happens to already pass on both old and
  new code (the pre-existing read-side formula already handled that specific
  path correctly) but guards the new write-side rescale against regression.
- Confirmed the existing `refunds correct amount when order-level promotion
  applied` e2e test (order-level/`DISTRIBUTED_ORDER_PROMOTION` case) still
  passes unchanged, and the `order-line.entity.spec.ts` #5101 NaN-regression
  unit test still passes unchanged.
- Full run of `order-modification.e2e-spec.ts` (73/73), `order-promotion.e2e-spec.ts`
  (66/67, 1 pre-existing skip), `order.e2e-spec.ts` (93/93),
  `order-cancel-shipping.e2e-spec.ts` (2/2), `order-multi-vendor-promotion.e2e-spec.ts`
  (6/6) and `shop-order.e2e-spec.ts` (104/104) — all green.

# Breaking changes

Yes - for existing data. `OrderLine.discounts`, `discountedLinePrice(WithTax)`,
`proratedLinePrice(WithTax)` etc. now return the *correct* discount amount in
the specific case (quantity reduced below `orderPlacedQuantity`) where they
previously returned an incorrectly rescaled (too small) discount for lines
recalculated via `modifyOrder`. Any code relying on the previous, incorrect
numbers in that narrow case would see different (larger, correct) output.

For lines that were **partially cancelled via `cancelOrder` and never
subsequently went through a `modifyOrder` recalculation**, this PR also
changes what `PROMOTION`-type adjustments are read against (see review
discussion below), so their stored `adjustment.amount` needs rescaling to
match. A `rescaleOrderLinePromotionAdjustments()` migration helper is now
exported from `\@vendure/core` (`migration-utils`) to run as part of your
upgrade migration - see its docblock for usage. The helper skips lines whose
`PROMOTION` amounts were last written by a `modifyOrder` recalculation, which
are already correct.

# Update, addressing review

- Added `rescaleOrderLinePromotionAdjustments()` (in `migration-utils`) and corrected the
  breaking-changes note above for existing partially-cancelled orders.
- Added an e2e test (through both `modifyOrder` and `cancelOrder`) covering an order that
  carries both a `PROMOTION` and a `DISTRIBUTED_ORDER_PROMOTION` at once, since
  `DefaultOrderLineDiscountDistributionStrategy.getWeight()` reads the exact value this PR
  changes for a reduced line.
- Moved the write-side rescale onto the entity (now `OrderLine.setQuantityRescalingAdjustments()`),
  next to `quantityBasisFor()`, and had `cancelOrderByOrderLines()` assign the rescaled
  quantity/adjustments onto the in-memory `OrderLine` before publishing its `OrderLineEvent`,
  so a subscriber no longer sees stale pre-cancellation values.
- Added `order-line.entity.spec.ts` unit tests for `PROMOTION` alone, both types together, and
  `orderPlacedQuantity: 0`.
- Trimmed the `quantityBasisFor()` docblock and de-duplicated the "before the fix..." narration
  that had been repeated across three files down to one explanation, with the others pointing at
  #5127.
- Reworked the flagged assertions in `order-calculator.spec.ts` and
  `order-modification.e2e-spec.ts` to pin expected values to a baseline captured independently
  of the code path under test, rather than deriving them from another getter on the same object.
  Added a length check before indexing `discounts[0]` in the e2e test.
- `quantityBasisFor()` is now a `switch` over `AdjustmentType`.
- Reverted the unrelated import-order changes in `order-modifier.ts` and
  `order-calculator.spec.ts`.
- Left the `orderLine.quantity > 0` guard in the rescale in place, since a lineInput requesting a
  no-op cancellation on an already-fully-cancelled line does reach it; added a comment explaining
  why.
- On the rounding-drift point: agreed it's real and bounded (confirmed by the same brute-force
  check across quantities 2-30), but fixing it needs either extra stored precision or reworking
  cancellation to route through a full recalculation, which felt like a larger change than this
  fix warranted - happy to follow up separately if you'd rather not carry it even at that bound.

# Update, addressing the migration review

- `rescaleOrderLinePromotionAdjustments()` no longer rescales every row matching
  `quantity < orderPlacedQuantity`. It reconstructs each candidate line's history from two further
  queries: every `order_modification.createdAt` belonging to an order that holds a candidate line,
  and every `ORDER_CANCELLATION` history entry for those orders, whose `data.lines` carries the
  quantity cancelled per line. The modification side is grouped per *order* and the cancellation
  side per *line*, because `modifyOrder()` ends in `applyPriceAdjustments()`, which rewrites
  `PROMOTION` amounts for every line in the order rather than only the ones the `OrderModification`
  references - so cancelling line B and then modifying line A still leaves B's amounts correct.
  A line with no modification at all is scaled from `orderPlacedQuantity` as before; a line whose
  order's latest modification is more recent than every one of its cancellations is skipped, since
  those amounts were already rewritten; a line cancelled after that modification is scaled from its
  current quantity plus the quantities cancelled since. Where a modification and a cancellation
  share a timestamp their order is unknowable from stored data, so the helper throws before writing
  anything and names the lines needing an explicit basis via `ambiguousOrderLineQuantityBases`.
- Table and column names come from `queryRunner.connection.getMetadata()` rather than being
  hardcoded, so `entityPrefix`, a schema or a custom naming strategy no longer break the queries.
- Added `v3_8_orderline_promotion_rescale.spec.ts`: 15 tests covering each of those branches, the
  equal-timestamp refusal and the two explicit-basis overrides, the metadata-resolved names, the
  parameter binding and the chunked update.
- `cancelOrderByOrderLines()` no longer zeroes a fully-cancelled line's `PROMOTION` amounts:
  `setQuantityRescalingAdjustments()` assigns the new quantity but leaves the adjustments alone
  when that quantity is 0, so the record of the applied discount survives for plugins and exports.
- Dropped `escLiteral()`. Both migration statements now use
  `queryRunner.connection.driver.createParameter()` with bound values, and the per-row loop is
  replaced by a chunked `UPDATE ... SET adjustments = CASE id WHEN ? THEN ? ... END WHERE id IN (...)`.
- Moved the `quantity === 0` guard into `quantityBasisFor()`, replaced the `orderPlacedQuantity: 0`
  unit test with a `PROMOTION` adjustment on a `quantity === 0` line, and had
  `setQuantityRescalingAdjustments()` assign `this.quantity` itself.
- Combo e2e tests: dropped the total-based assertion and the derived `totalWithTax` check, capture
  each line's order-level share before the change, and tell the two discount types apart with
  `discount.type`. The promotions they create are deleted in an `afterAll`, and the duplicated
  `CanceledOrderFragment`/guard now live once at the top-level describe.
- Reverted the last prettier hunk in `order-modifier.ts`. Note the repo's `bun.lock` pins prettier
  3.7.4 while a bare `npx prettier` resolves 3.9.6, which formats that union the other way -
  `npx prettier@3.7.4 --check` is clean on every file touched here.

_The description of the migration's behaviour above was brought in line with the current
implementation by @biggamesmallworld while reviewing; no code was changed._

# Screenshots

N/A — logic fix, no UI changes.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5143"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789483559&installation_model_id=20193&pr_number=5143&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5143&signature=af7742817b22bf7bbe3b845504e75b89e749a58a095dde14a4f53026b1baa2ab"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/order-modification.e2e-spec.ts
+++ b/packages/core/e2e/order-modification.e2e-spec.ts
@@ -2880,11 +2880,16 @@ describe('Order modification', () => {
 
             // A partial cancellation does not redistribute the order-level discount, so the
             // reduced line keeps the per-unit share it was assigned at placement, and the
-            // untouched line keeps its share in full.
-            expect(orderLevelShareOf(remainingDiscountedLine) / remainingDiscountedLine.quantity).toBeCloseTo(
-                orderLevelShareOf(placedLines.get('T_1')!) / placedLines.get('T_1')!.quantity,
-                6,
+            // untouched line keeps its share in full. The per-unit rate itself is exact, but
+            // `orderLevelShareOf` is a line total rounded to the nearest minor unit, so scale the
+            // placement rate by the reduced quantity and round the same way rather than comparing
+            // two already-rounded per-unit values, which can differ by a fraction of a cent.
+            const placedDiscountedShareLine = placedLines.get('T_1')!;
+            const expectedRemainingShare = Math.round(
+                (orderLevelShareOf(placedDiscountedShareLine) / placedDiscountedShareLine.quantity) *
+                    remainingDiscountedLine.quantity,
             );
+            expect(orderLevelShareOf(remainingDiscountedLine)).toBe(expectedRemainingShare);
             expect(orderLevelShareOf(remainingLines.get('T_4')!)).toBe(
                 orderLevelShareOf(placedLines.get('T_4')!),
             );
--- a/packages/core/e2e/order-modification.e2e-spec.ts
+++ b/packages/core/e2e/order-modification.e2e-spec.ts
@@ -2642,7 +2642,8 @@ describe('Order modification', () => {
             orderGuard.assertSuccess(placedOrder);
 
             const placedLine = placedOrder.lines[0];
-            expect(placedLine.discountedUnitPriceWithTax).toBe(placedLine.unitPriceWithTax / 2);
+            const placedUnitPriceWithTax = placedLine.unitPriceWithTax;
+            expect(placedLine.discountedUnitPriceWithTax).toBe(placedUnitPriceWithTax / 2);
 
             const transitionOrderToState = await adminTransitionOrderToState(placedOrder.id, 'Modifying');
             orderGuard.assertSuccess(transitionOrderToState);
@@ -2664,13 +2665,12 @@ describe('Order modification', () => {
             const modifiedLine = modifyOrder.lines.find(l => l.id === placedLine.id)!;
             expect(modifiedLine.quantity).toBe(2);
             expect(modifiedLine.orderPlacedQuantity).toBe(20);
-            // Before the fix, `OrderCalculator.applyPriceAdjustments` (triggered by `modifyOrder`)
-            // already recomputes `adjustments` freshly against the *new* quantity of 2 - i.e. the
-            // stored adjustment total is already correctly scaled. The read-side getters then divided
-            // that already-correct total by `Math.max(orderPlacedQuantity, quantity)` a second time,
-            // silently rescaling the 50% discount down to `quantity / orderPlacedQuantity` (2/20 = 10%)
-            // of its correct value.
-            expect(modifiedLine.discountedLinePriceWithTax).toBe(modifiedLine.linePriceWithTax / 2);
+            // Both assertions are pinned to `placedUnitPriceWithTax`, captured before the
+            // modification, rather than to `modifiedLine`'s own fields - see #5127 for why the
+            // pre-fix read-side getters silently rescaled the discount down to 10% of its correct
+            // value here.
+            expect(modifiedLine.linePriceWithTax).toBe(placedUnitPriceWithTax * 2);
+            expect(modifiedLine.discountedLinePriceWithTax).toBe(placedUnitPriceWithTax);
         });
 
         it('cancelOrder: partially cancelling a line keeps the 50% discount ratio on the remaining units', async () => {
@@ -2726,7 +2726,203 @@ describe('Order modification', () => {
             // The remaining 2 units should still carry the full 50% discount: half of their combined
             // list price, i.e. exactly one unit's price.
             expect(remainingLine.proratedLinePriceWithTax).toBe(unitPriceWithTax);
+            expect(remainingLine.discounts.length).toBe(1);
             expect(remainingLine.discounts[0].amountWithTax).toBe(-unitPriceWithTax);
+        });
+
+        // #5127 — DefaultOrderLineDiscountDistributionStrategy.getWeight() reads
+        // line.proratedLinePriceWithTax, so an order-level DISTRIBUTED_ORDER_PROMOTION's split
+        // across lines depends on the very value this fix changes for a reduced line. These two
+        // tests put an item-level PROMOTION and an order-level DISTRIBUTED_ORDER_PROMOTION on the
+        // same order, then reduce the discounted line's quantity, to guard the split against
+        // corruption from that interaction.
+        it('modifyOrder: order-level discount split survives reducing one line of a multi-line order', async () => {
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'ITEM_5127_COMBO_MODIFY',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: productsPercentageDiscount.code,
+                            arguments: [
+                                { name: 'discount', value: '50' },
+                                { name: 'productVariantIds', value: JSON.stringify(['T_1']) },
+                            ],
+                        },
+                    ],
+                    translations: [
+                        { languageCode: LanguageCode.en, name: '50% off T_1 (#5127 combo modify)' },
+                    ],
+                },
+            });
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'ORDER_5127_COMBO_MODIFY',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: orderFixedDiscount.code,
+                            arguments: [{ name: 'discount', value: '500' }],
+                        },
+                    ],
+                    translations: [
+                        { languageCode: LanguageCode.en, name: '$5 off order (#5127 combo modify)' },
+                    ],
+                },
+            });
+
+            await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_1',
+                quantity: 20,
+            } as any);
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_4',
+                quantity: 1,
+            } as any);
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'ITEM_5127_COMBO_MODIFY' });
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'ORDER_5127_COMBO_MODIFY' });
+            await proceedToArrangingPayment(shopClient);
+            const placedOrder = await addPaymentToOrder(shopClient, testSuccessfulPaymentMethod);
+            orderGuard.assertSuccess(placedOrder);
+
+            const placedDiscountedLine = placedOrder.lines.find(l => l.productVariant.id === 'T_1')!;
+            const placedOtherLine = placedOrder.lines.find(l => l.productVariant.id === 'T_4')!;
+            const placedUnitPriceWithTax = placedDiscountedLine.unitPriceWithTax;
+
+            const transitionOrderToState = await adminTransitionOrderToState(placedOrder.id, 'Modifying');
+            orderGuard.assertSuccess(transitionOrderToState);
+
+            const { modifyOrder } = await adminClient.query(modifyOrderDocument, {
+                input: {
+                    dryRun: false,
+                    orderId: placedOrder.id,
+                    adjustOrderLines: [{ orderLineId: placedDiscountedLine.id, quantity: 2 }],
+                    refund: {
+                        paymentId: placedOrder.payments![0].id,
+                        reason: 'reduced quantity (#5127 combo modify)',
+                    },
+                },
+            });
+            orderWithModificationsGuard.assertSuccess(modifyOrder);
+
+            const modifiedDiscountedLine = modifyOrder.lines.find(l => l.id === placedDiscountedLine.id)!;
+            const modifiedOtherLine = modifyOrder.lines.find(l => l.id === placedOtherLine.id)!;
+
+            // The item-level 50% discount on the reduced line survives, independently of the
+            // order-level discount below.
+            expect(modifiedDiscountedLine.discountedLinePriceWithTax).toBe(placedUnitPriceWithTax);
+
+            // The order-level $5 discount is still split, in full, across the two lines - it is
+            // neither dropped nor double-counted by combining it with the item-level discount.
+            const discountedLineOrderShare =
+                modifiedDiscountedLine.proratedLinePriceWithTax -
+                modifiedDiscountedLine.discountedLinePriceWithTax;
+            const otherLineOrderShare =
+                modifiedOtherLine.proratedLinePriceWithTax - modifiedOtherLine.discountedLinePriceWithTax;
+            expect(discountedLineOrderShare + otherLineOrderShare).toBe(-500);
+            expect(modifiedOtherLine.discountedLinePriceWithTax).toBe(placedOtherLine.unitPriceWithTax);
+
+            expect(modifyOrder.totalWithTax).toBe(modifyOrder.subTotalWithTax + modifyOrder.shippingWithTax);
+        });
+
+        it('cancelOrder: order-level discount split survives partially cancelling one line of a multi-line order', async () => {
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'ITEM_5127_COMBO_CANCEL',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: productsPercentageDiscount.code,
+                            arguments: [
+                                { name: 'discount', value: '50' },
+                                { name: 'productVariantIds', value: JSON.stringify(['T_1']) },
+                            ],
+                        },
+                    ],
+                    translations: [
+                        { languageCode: LanguageCode.en, name: '50% off T_1 (#5127 combo cancel)' },
+                    ],
+                },
+            });
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'ORDER_5127_COMBO_CANCEL',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: orderFixedDiscount.code,
+                            arguments: [{ name: 'discount', value: '500' }],
+                        },
+                    ],
+                    translations: [
+                        { languageCode: LanguageCode.en, name: '$5 off order (#5127 combo cancel)' },
+                    ],
+                },
+            });
+
+            await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_1',
+                quantity: 20,
+            } as any);
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_4',
+                quantity: 1,
+            } as any);
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'ITEM_5127_COMBO_CANCEL' });
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'ORDER_5127_COMBO_CANCEL' });
+            await proceedToArrangingPayment(shopClient);
+            const placedOrder = await addPaymentToOrder(shopClient, testSuccessfulPaymentMethod);
+            orderGuard.assertSuccess(placedOrder);
+
+            const placedDiscountedLine = placedOrder.lines.find(l => l.productVariant.id === 'T_1')!;
+            const placedOtherLine = placedOrder.lines.find(l => l.productVariant.id === 'T_4')!;
+            const unitPriceWithTax = placedDiscountedLine.unitPriceWithTax;
+
+            const { cancelOrder } = await adminClient.query(cancelOrderDocument, {
+                input: {
+                    orderId: placedOrder.id,
+                    lines: [{ orderLineId: placedDiscountedLine.id, quantity: 18 }],
+                    reason: 'partial cancellation (#5127 combo cancel)',
+                },
+            });
+            canceledOrderGuard.assertSuccess(cancelOrder);
+
+            const { order } = await adminClient.query(getOrderWithLineDiscountsDocument, {
+                id: placedOrder.id,
+            });
+            const remainingDiscountedLine = order!.lines.find(l => l.id === placedDiscountedLine.id)!;
+            const remainingOtherLine = order!.lines.find(l => l.id === placedOtherLine.id)!;
+
+            expect(remainingDiscountedLine.quantity).toBe(2);
+            expect(remainingDiscountedLine.orderPlacedQuantity).toBe(20);
+
+            // The order-level discount is present on both lines; the item-level discount only on
+            // the discounted one. Tell them apart that way, rather than by adjustmentSource format.
+            const otherLineSources = new Set(remainingOtherLine.discounts.map(d => d.adjustmentSource));
+            const itemDiscounts = remainingDiscountedLine.discounts.filter(
+                d => !otherLineSources.has(d.adjustmentSource),
+            );
+            const discountedLineOrderLevelDiscounts = remainingDiscountedLine.discounts.filter(d =>
+                otherLineSources.has(d.adjustmentSource),
+            );
+
+            // The item-level 50% discount on the remaining 2 units still equals half their
+            // combined list price, independently of the order-level discount split below.
+            expect(itemDiscounts.length).toBe(1);
+            expect(itemDiscounts[0].amountWithTax).toBe(-unitPriceWithTax);
+
+            // The order-level $5 discount is still split, in full, across both lines - neither
+            // dropped nor double-counted by combining it with the item-level discount.
+            const orderLevelDiscountTotal =
+                summate(discountedLineOrderLevelDiscounts, 'amountWithTax') +
+                summate(remainingOtherLine.discounts, 'amountWithTax');
+            expect(orderLevelDiscountTotal).toBe(-500);
         });
     });
 
--- a/packages/core/e2e/order-modification.e2e-spec.ts
+++ b/packages/core/e2e/order-modification.e2e-spec.ts
@@ -2606,6 +2606,130 @@ describe('Order modification', () => {
         });
     });
 
+    describe('reducing OrderLine quantity below orderPlacedQuantity keeps the discount ratio (#5127)', () => {
+        type CanceledOrderFragment = FragmentOf<typeof canceledOrderFragment>;
+        const canceledOrderGuard: ErrorResultGuard<CanceledOrderFragment> = createErrorResultGuard(
+            input => !!input.lines,
+        );
+
+        it('modifyOrder: reducing quantity via order modification keeps the 50% discount', async () => {
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'HALF_5127_MODIFY',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: productsPercentageDiscount.code,
+                            arguments: [
+                                { name: 'discount', value: '50' },
+                                { name: 'productVariantIds', value: JSON.stringify(['T_1']) },
+                            ],
+                        },
+                    ],
+                    translations: [{ languageCode: LanguageCode.en, name: '50% off T_1 (#5127 modify)' }],
+                },
+            });
+
+            await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_1',
+                quantity: 20,
+            } as any);
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'HALF_5127_MODIFY' });
+            await proceedToArrangingPayment(shopClient);
+            const placedOrder = await addPaymentToOrder(shopClient, testSuccessfulPaymentMethod);
+            orderGuard.assertSuccess(placedOrder);
+
+            const placedLine = placedOrder.lines[0];
+            expect(placedLine.discountedUnitPriceWithTax).toBe(placedLine.unitPriceWithTax / 2);
+
+            const transitionOrderToState = await adminTransitionOrderToState(placedOrder.id, 'Modifying');
+            orderGuard.assertSuccess(transitionOrderToState);
+
+            // Reduce the quantity from 20 down to 2, well below the orderPlacedQuantity of 20.
+            const { modifyOrder } = await adminClient.query(modifyOrderDocument, {
+                input: {
+                    dryRun: false,
+                    orderId: placedOrder.id,
+                    adjustOrderLines: [{ orderLineId: placedLine.id, quantity: 2 }],
+                    refund: {
+                        paymentId: placedOrder.payments![0].id,
+                        reason: 'reduced quantity (#5127)',
+                    },
+                },
+            });
+            orderWithModificationsGuard.assertSuccess(modifyOrder);
+
+            const modifiedLine = modifyOrder.lines.find(l => l.id === placedLine.id)!;
+            expect(modifiedLine.quantity).toBe(2);
+            expect(modifiedLine.orderPlacedQuantity).toBe(20);
+            // Before the fix, `OrderCalculator.applyPriceAdjustments` (triggered by `modifyOrder`)
+            // already recomputes `adjustments` freshly against the *new* quantity of 2 - i.e. the
+            // stored adjustment total is already correctly scaled. The read-side getters then divided
+            // that already-correct total by `Math.max(orderPlacedQuantity, quantity)` a second time,
+            // silently rescaling the 50% discount down to `quantity / orderPlacedQuantity` (2/20 = 10%)
+            // of its correct value.
+            expect(modifiedLine.discountedLinePriceWithTax).toBe(modifiedLine.linePriceWithTax / 2);
+        });
+
+        it('cancelOrder: partially cancelling a line keeps the 50% discount ratio on the remaining units', async () => {
+            await adminClient.query(createPromotionDocument, {
+                input: {
+                    couponCode: 'HALF_5127_CANCEL',
+                    enabled: true,
+                    conditions: [],
+                    actions: [
+                        {
+                            code: productsPercentageDiscount.code,
+                            arguments: [
+                                { name: 'discount', value: '50' },
+                                { name: 'productVariantIds', value: JSON.stringify(['T_1']) },
+                            ],
+                        },
+                    ],
+                    translations: [{ languageCode: LanguageCode.en, name: '50% off T_1 (#5127 cancel)' }],
+                },
+            });
+
+            await shopClient.asUserWithCredentials('hayden.zieme12@hotmail.com', 'test');
+            await shopClient.query(addItemToOrderWithCustomFieldsDocument, {
+                productVariantId: 'T_1',
+                quantity: 20,
+            } as any);
+            await shopClient.query(applyCouponCodeDocument, { couponCode: 'HALF_5127_CANCEL' });
+            await proceedToArrangingPayment(shopClient);
+            const placedOrder = await addPaymentToOrder(shopClient, testSuccessfulPaymentMethod);
+            orderGuard.assertSuccess(placedOrder);
+
+            const placedLine = placedOrder.lines[0];
+            const unitPriceWithTax = placedLine.unitPriceWithTax;
+
+            // Cancel 18 of the 20 units, going through `OrderModifier.cancelOrderByOrderLines()` - a
+            // path which, unlike `modifyOrder`, does *not* recalculate promotions from scratch.
+            const { cancelOrder } = await adminClient.query(cancelOrderDocument, {
+                input: {
+                    orderId: placedOrder.id,
+                    lines: [{ orderLineId: placedLine.id, quantity: 18 }],
+                    reason: 'partial cancellation (#5127)',
+                },
+            });
+            canceledOrderGuard.assertSuccess(cancelOrder);
+
+            const { order } = await adminClient.query(getOrderWithLineDiscountsDocument, {
+                id: placedOrder.id,
+            });
+            const remainingLine = order!.lines.find(l => l.id === placedLine.id)!;
+
+            expect(remainingLine.quantity).toBe(2);
+            expect(remainingLine.orderPlacedQuantity).toBe(20);
+            // The remaining 2 units should still carry the full 50% discount: half of their combined
+            // list price, i.e. exactly one unit's price.
+            expect(remainingLine.proratedLinePriceWithTax).toBe(unitPriceWithTax);
+            expect(remainingLine.discounts[0].amountWithTax).toBe(-unitPriceWithTax);
+        });
+    });
+
     async function adminTransitionOrderToState(id: string, state: string) {
         const result = await adminClient.query(adminTransitionToStateDocument, {
             id,
--- a/packages/core/e2e/order-modification.e2e-spec.ts
+++ b/packages/core/e2e/order-modification.e2e-spec.ts
@@ -220,9 +220,18 @@ const getOrderWithLineDiscountsDocument = adminGraphql(`
                 id
                 quantity
                 orderPlacedQuantity
+                unitPriceWithTax
+                linePriceWithTax
+                discountedUnitPriceWithTax
+                discountedLinePriceWithTax
+                proratedUnitPriceWithTax
                 proratedLinePriceWithTax
+                productVariant {
+                    id
+                }
                 discounts {
                     adjustmentSource
+                    type
                     amount
                     amountWithTax
                 }
@@ -317,6 +326,11 @@ describe('Order modification', () => {
     type OrderWithModificationsFragment = FragmentOf<typeof orderWithModificationsFragment>;
     const orderWithModificationsGuard: ErrorResultGuard<OrderWithModificationsFragment> =
         createErrorResultGuard(input => !!input.id);
+
+    type CanceledOrderFragment = FragmentOf<typeof canceledOrderFragment>;
+    const canceledOrderGuard: ErrorResultGuard<CanceledOrderFragment> = createErrorResultGuard(
+        input => !!input.lines,
+    );
 
     // Type aliases for fragment types
     type OrderWithLinesFragment = FragmentOf<typeof orderWithLinesFragment>;
@@ -2508,11 +2522,6 @@ describe('Order modification', () => {
     describe('cancelling an order with a line added by a modification', () => {
         const CODE_HALF_PRICE = 'HALF_PRICE';
 
-        type CanceledOrderFragment = FragmentOf<typeof canceledOrderFragment>;
-        const canceledOrderGuard: ErrorResultGuard<CanceledOrderFragment> = createErrorResultGuard(
-            input => !!input.lines,
-        );
-
         it('order with discounts can still be read after cancellation', async () => {
             await adminClient.query(createPromotionDocument, {
                 input: {
@@ -2607,15 +2616,12 @@ describe('Order modification', () => {
     });
 
     describe('reducing OrderLine quantity below orderPlacedQuantity keeps the discount ratio (#5127)', () => {
-        type CanceledOrderFragment = FragmentOf<typeof canceledOrderFragment>;
-        const canceledOrderGuard: ErrorResultGuard<CanceledOrderFragment> = createErrorResultGuard(
-            input => !!input.lines,
-        );
+        const createdPromotionIds: string[] = [];
 
-        it('modifyOrder: reducing quantity via order modification keeps the 50% discount', async () => {
-            await adminClient.query(createPromotionDocument, {
+        async function createItemPromotion(couponCode: string, name: string) {
+            const { createPromotion } = await adminClient.query(createPromotionDocument, {
                 input: {
-                    couponCode: 'HALF_5127_MODIFY',
+                    couponCode,
                     enabled: true,
                     conditions: [],
                     actions: [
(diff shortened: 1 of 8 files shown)


wakqasahmed:

wakqasahmed:
Hi maintainers — the CLA Assistant Lite bot already confirmed above that all contributors have signed the CLA ✍️ ✅ (posted 2026-08-16), and `mergeable` shows `MERGEABLE`. However, the `CLAAssistant` check run itself still shows `FAILURE` — it looks like it never re-ran after the signature was recorded, so the status is just stale/cosmetic. Could someone re-trigger that check or manually confirm mergeability so this doesn't sit blocked on a check that's out of sync with the bot's own confirmation? Thanks!

wakqasahmed:
recheck

wakqasahmed:
Following up again — the CLA was signed on 2026-08-16 and the CLA Assistant Lite bot confirmed it right here in this thread ("All contributors have signed the CLA ✍️ ✅"), but the separate `CLAAssistant` check run is still showing the stale `failure` result from just before that signature landed, and it's never re-run since (checked directly against the check run for this PR's head commit — no re-run since 2026-08-16T14:46:01Z). `recheck` didn't retrigger it when I tried on 08-22.

Could a maintainer manually re-run/clear that check, or confirm it's safe to merge despite the stale status? Happy to do anything needed on our end — just don't have permission to rerequest that specific check run myself. Thanks for your patience on this one.

biggamesmallworld (changes requested):
Thanks for the detailed write-up. The diagnosis is right: the read/write basis mismatch is real, and the two adjustment types genuinely do need different treatment. The problem is that this converts a pure read-time formula into persisted state, and existing rows were written under the old contract.

## Blocking

### 1. Existing partially-cancelled orders read as negative on upgrade

The `PROMOTION` read basis moves from `Math.max(orderPlacedQuantity, quantity)` to `quantity` (`order-line.entity.ts:481-486`), but the compensating rescale is applied only at write time in `cancelOrderByOrderLines`. Rows written before this change never got that rescale, so their stored `amount` is still the total for the pre-cancellation quantity while the new read divides by the current quantity.

I built an `OrderLine` matching a row written by the pre-fix `cancelOrderByOrderLines` (20 units at 2975 with a 50% item promotion, 18 units cancelled, so `quantity: 2`, `orderPlacedQuantity: 20`, `adjustments[0].amount: -29750`) and ran it on both branches:

| | master | this PR |
|---|---|---|
| `linePriceWithTax` | 5950 | 5950 |
| `discounts[0].amountWithTax` | -2975 | **-29750** |
| `discountedLinePriceWithTax` | 2975 | **-23800** |
| `proratedLinePriceWithTax` | 2975 | **-23800** |

There is no read-side clamp on this path: the `Math.max(maxDiscount, ...)` guard in `addAdjustment()` only runs on write. So refunds, `Order.discounts`, the prorated getters and any reporting over historical orders all pick up the inflated value.

This needs a migration that rescales existing `PROMOTION` adjustments on lines where `quantity < orderPlacedQuantity`, and the "Breaking changes: None" line in the description needs correcting.

### 2. The change feeds into order-level discount distribution, with no test

`DefaultOrderLineDiscountDistributionStrategy.getWeight()` returns `line.proratedLinePriceWithTax`, which is exactly the value this PR changes for a reduced line. So on any order carrying both an item-level and an order-level promotion, this alters how the `DISTRIBUTED_ORDER_PROMOTION` is split across *all* lines, not just the reduced one.

None of the added tests put both adjustment types on the same order after a quantity reduction. Please add one covering that combination through both the `modifyOrder` and `cancelOrder` paths. If it turns out the weights shift, that is a second thing to reconcile before this can land.

## Should fix

**The rule now lives in two places that cannot see each other.** `OrderLine.quantityBasisFor()` decides the read basis, and a hand-written scale factor in `order-modifier.ts` maintains the invariant that basis depends on. The only thing binding them is a paragraph of prose in each file pointing at the other. Moving the write side onto the entity, say `orderLine.rescaleAdjustmentsForQuantity(newQuantity)`, gives a future change to `quantityBasisFor` one obvious sibling to update.

**Stale entity handed to the event.** `order-modifier.ts:355-361` writes the rescaled adjustments and new quantity via `repository.update()`, then publishes `new OrderLineEvent(ctx, order, orderLine, 'cancelled')` with the untouched in-memory `orderLine`. A subscriber reading `orderLine.discounts` sees pre-cancellation numbers. The stale quantity predates this PR, but this adds a second stale field to it. Assigning both onto the entity before the update fixes both.

**Assertions derived from the object under test.** `order-modification.e2e-spec.ts:2674`:

```ts
expect(modifiedLine.discountedLinePriceWithTax).toBe(modifiedLine.linePriceWithTax / 2);
```

If `linePriceWithTax` itself regressed, this still passes. Same shape in `order-calculator.spec.ts:492`, where `Math.round(order.lines[0].linePriceWithTax * 0.6)` also re-implements the production rounding, so the test tracks the implementation rather than pinning the behaviour. `listPrice` is 2975 and quantity is 2, so please write the integer.

**Missing unit tests where they are cheapest.** `order-line.entity.spec.ts` already has a `createOrderLine(quantity, orderPlacedQuantity)` helper and currently only covers `DISTRIBUTED_ORDER_PROMOTION`. This PR introduces per-type basis branching and does not touch that file. A `PROMOTION` adjustment alone, both types on one line, and a `PROMOTION` adjustment on a line with `orderPlacedQuantity: 0` would cover the whole new function in a few lines.

**Docblock length.** Twenty lines of comment on a four-line function (`order-line.entity.ts:461-480`), and most of it belongs in the PR description. Lines like "has no documented scaling contract to assume otherwise" are arguing with a reviewer inside the source file. Keep the one non-obvious fact, that the two types have different bases and why the distributed one is pinned to placement, and let the issue link carry the rest.

**Change narration, three copies.** The order-modifier comment, the unit test comment and the e2e test comment all repeat the same "before the fix, the read-side getters then divided that already-correct total by `Math.max(...)` a second time" paragraph. A reader two years from now has never seen the pre-fix code. Say what the value should be and why, once.

Also `remainingLine.discounts[0].amountWithTax` at `order-modification.e2e-spec.ts:2729` indexes without checking length, so a regression reports a TypeError rather than a number mismatch.

## Minor

**Rounding drift in the rescale.** `const adjustmentScaleFactor = orderLine.quantity > 0 ? newQuantity / orderLine.quantity : 0` then `roundMoney(adjustment.amount * adjustmentScaleFactor)` rounds off the previously rounded value, so cancelling a line one unit at a time compounds. Nothing recomputes afterwards, since `cancelOrderByOrderLines` ends in `calculateOrderTotals()` rather than `applyPriceAdjustments()`, so it is permanent. I brute-forced amounts from -1 to -5000 across quantities 2 to 30 stepping down one unit at a time: worst case diverges from the one-shot value by 1 minor unit. Small enough that I would not block on it, but rescaling from an untouched original rather than the running value avoids it entirely.

**`quantityBasisFor` hides its third case in a fall-through.** `PROMOTION` gets an `if`, `DISTRIBUTED_ORDER_PROMOTION` and `OTHER` share a bare `return`. A `switch` costs two lines and makes the next `AdjustmentType` addition a compile-time question instead of a silent default.

**Unrelated import reordering.** `order-modifier.ts` and `order-calculator.spec.ts` both move `order/order.entity` above `order-line/order-line.entity` for reasons unconnected to this change. Looks like an editor import sorter disagreeing with the repo's. Please revert those hunks.

**Dead guard.** The `orderLine.quantity > 0 ?` check in the scale factor: I could not find a path that reaches this loop with a zero-quantity line. If there isn't one, drop it rather than leaving a reader wondering what it protects.

## Summary

Blocking before merge: the migration for existing partially-cancelled orders, a test covering `PROMOTION` plus `DISTRIBUTED_ORDER_PROMOTION` on a reduced line, and correcting the breaking-changes claim.

wakqasahmed:
Thanks for the thorough review, this is genuinely helpful. I've pushed a follow-up commit addressing everything:

**Blocking**

- Existing partially-cancelled orders reading as negative: added `rescaleOrderLinePromotionAdjustments()` as a new export in `migration-utils` (following the pattern of the existing `v3_6_*` helpers), and corrected the "Breaking changes: None" claim in the PR description - it's genuinely a data migration for anyone with partially-cancelled orders. I've documented a real limitation in its docblock: it can't distinguish a line whose `PROMOTION` amount is still on the `orderPlacedQuantity` basis (needs rescaling) from one that's already been rescaled by a later `modifyOrder` recalculation, since there's no stored field to tell them apart. I called that out explicitly rather than pretending it's airtight.
- Order-level distribution test: added two new e2e tests (`modifyOrder` and `cancelOrder`) with both a `PROMOTION` and a `DISTRIBUTED_ORDER_PROMOTION` on the same multi-line order, reducing the discounted line's quantity and asserting the order-level split still sums to the full discount across both lines, and that the item-level discount survives independently of it.

**Should fix**

- Moved the rescale off `order-modifier.ts` and onto the entity as `OrderLine.rescaleAdjustmentsForQuantity()`, right next to `quantityBasisFor()`.
- `cancelOrderByOrderLines()` now assigns the rescaled quantity/adjustments onto the in-memory `OrderLine` before the `repository.update()` call and before publishing the `OrderLineEvent`, so a subscriber sees current values on both fields.
- Reworked the flagged assertions in `order-calculator.spec.ts` and `order-modification.e2e-spec.ts` so they no longer derive the expected value from another getter on the same object under test - `order-calculator.spec.ts` now asserts the literal integers (35700 / 3570), and the e2e test pins its expectations to `unitPriceWithTax` captured before the modification. Added the missing `discounts.length` check before indexing `discounts[0]`.
- Added `order-line.entity.spec.ts` coverage for a `PROMOTION` adjustment alone, both adjustment types on one line, and `orderPlacedQuantity: 0`.
- Trimmed the `quantityBasisFor()` docblock down to the one non-obvious fact, and de-duplicated the "before the fix..." narration that was repeated in `order-modifier.ts`, `order-calculator.spec.ts` and the e2e spec - now stated once, with the others pointing at #5127.

**Minor**

- `quantityBasisFor()` is now a `switch` over `AdjustmentType`.
- Reverted the unrelated import reordering in `order-modifier.ts` and `order-calculator.spec.ts` - you were right, that wasn't this repo's own import order (confirmed against `prettier-plugin-organize-imports`, which put them straight back to the original order).
- On the dead guard: I actually found a path that reaches it - a `lineInputs` entry requesting a 0-quantity (no-op) cancellation on a line that's already fully cancelled would otherwise divide 0/0. Left it in place and added a one-line comment explaining why, rather than removing it.
- On the rounding drift: agreed it's real, and your bound (worst case 1 minor unit) matches what I'd expect. Fixing it properly means either storing extra precision or routing cancellation through a full recalculation, which felt like a bigger change than this fix warrants - I've left it as is for now, happy to open a separate issue/PR for it if you'd rather not carry even that bound.

All existing and new unit tests in `order-line.entity.spec.ts` and `order-calculator.spec.ts` pass locally. I wasn't able to get the e2e suite running in my local environment this round (unrelated tooling issue on my end, not a code problem) so I'm relying on CI for `order-modification.e2e-spec.ts` - flagging that in case CI turns up something I should look at.

biggamesmallworld (changes requested):
Blockers from the last round are addressed: migration helper, breaking-changes note, `rescaleAdjustmentsForQuantity()` on the entity, non-stale `OrderLineEvent`, and the combo test I asked for.

That combo test is now failing CI on all four DBs:

```
modifyOrder: order-level discount split ...  expected -337 to be -500   (:2825)
cancelOrder: order-level discount split ...  expected -117 to be -500   (:2925)
```

The assertion is wrong, not the code. I ran both tests against master's `order-line.entity.ts` and `order-modifier.ts` with your test file: the `cancelOrder` one fails there too, at -141. So "the order-level discount still sums to the full amount after a quantity reduction" was never true. It is the direct consequence of pinning `DISTRIBUTED_ORDER_PROMOTION` to `orderPlacedQuantity`, which this PR deliberately preserves. (The `modifyOrder` variant fails earlier on master, on the item-level assertion, which is #5127 itself and is genuinely fixed here.)

Please assert the invariant the pinning actually provides instead:

- the reduced line's *per-unit* order-level share is unchanged from placement
- the untouched line's order-level share is unchanged in full

Both hold before and after this change, and they are what would break if someone later altered `quantityBasisFor()`. Drop the `+ otherLineOrderShare).toBe(-500)` line from both tests.

The whole-order shortfall is real but pre-existing and out of scope here. Worth its own issue.

Unrelated: the `default-search-plugin` "updates index when a Product is deleted" failure in the same run is a known flake. Re-run CI once the assertions are updated.

biggamesmallworld (changes requested):
Thanks, this addresses the read/write basis properly and the new unit tests do fail against master, so they are pinning real behaviour. The remaining work is on the migration helper.

## Blocking

### 1. The migration helper needs to exclude modified lines

`rescaleOrderLinePromotionAdjustments()` selects every row where `quantity < orderPlacedQuantity` and multiplies its `PROMOTION` amounts by `quantity / orderPlacedQuantity`. A line whose quantity was reduced through `modifyOrder` does not need that: `OrderModifier.modifyOrder()` ends in `applyPriceAdjustments()` (`order-modifier.ts:650`), which recomputes `PROMOTION` amounts against the new quantity. Those rows still match the `WHERE` clause, so the helper scales an already correct amount a second time. On a 50% discount reduced from 20 units to 2, that turns a correct -2975 into -297. It is #5127 again, in the opposite direction.

The docblock says there is no stored field that distinguishes the two cases. There is one. `OrderModificationLine` (a `ChildEntity` of `OrderLineReference`) links an `orderLineId` to a `modificationId`, so a line that went through a modification is reachable by join, and `order_modification.createdAt` compared against the latest `Cancellation.createdAt` for the same line tells you which happened last. Narrow the query with that rather than documenting the hole. Since interleaved modifications and cancellations on one line are possible, the comparison needs to be per line, not per order.

Please also add unit coverage for the helper. It is exported public API that rewrites money, and it currently has none.

### 2. Fully cancelling a line destroys its stored discount

`cancelOrderByOrderLines()` calls `rescaleAdjustmentsForQuantity(newQuantity)` for every line input including a full cancellation, where `newQuantity` is 0. The scale factor is then 0, every `PROMOTION` amount is written as 0, and `cancelOrder` does not recalculate afterwards (`order.service.ts:1917`), so the zeroes are permanent.

The getters return 0 either way, since both short-circuit on `quantity === 0`, so this changes nothing on the read side. What it changes is the record: a plugin, report or accounting export reading `orderLine.adjustments` on a cancelled line now sees zero where the applied discount used to be. Skip the rescale when `newQuantity` is 0.

### 3. Build the migration SQL with parameters

`escLiteral()` hand-rolls quote doubling and the result is interpolated into the `UPDATE` along with a `JSON.stringify` blob. Backslash is an escape character inside string literals on mysql and mariadb under the default `sql_mode`, so a promotion description containing a quote or a backslash reaches the database altered: the JSON escape `\"` is unescaped by the driver and the stored value no longer parses as written. `queryRunner.query(sql, params)` takes parameters, and the sibling helpers in `migration-utils/` use it. Use it here and delete `escLiteral()`.

Two smaller things in the same function: `WHERE id = '123'` compares an integer primary key against a string literal, and the loop issues one round trip per row.

### 4. Prettier fails on `order-modifier.ts`

`npx prettier --check packages/core/src/service/helpers/order-modifier/order-modifier.ts` flags the file. The unrelated hunk is at line 592, where the multiline union return cast was collapsed onto one line. The import reordering was reverted as described, but this one was not. Revert it too.

## Should fix

**The combo tests assert the total, not the split.** `expect(discountedLineOrderShare + otherLineOrderShare).toBe(-500)` passes for `-500 / 0` as readily as for the correct split, and the split across lines is the thing the test exists to guard, since `getWeight()` reads `proratedLinePriceWithTax`. Capture each line's share before the modification and assert both.

**`expect(modifyOrder.totalWithTax).toBe(modifyOrder.subTotalWithTax + modifyOrder.shippingWithTax)`** derives both sides from the same response, so it holds for any internally consistent set of wrong numbers. Drop it or pin it to an integer.

**The `orderPlacedQuantity: 0` unit test no longer covers anything.** My fault for naming that case: after this change `PROMOTION` never divides by `orderPlacedQuantity`, so the test passes on master too. The case worth having is a `PROMOTION` adjustment on a line with `quantity === 0`, which is where `quantityBasisFor()` returns 0.

**Move the `quantity === 0` guard into `quantityBasisFor()`.** It returns 0 for `PROMOTION` when the line is empty, and that is safe only because all three call sites check first. Given #5101 was a NaN regression in this same code, the guard belongs in the helper.

**Have `rescaleAdjustmentsForQuantity()` assign `this.quantity`.** Both callers must remember to set it afterwards, and the only thing enforcing that is a line of docblock. Folding the assignment in removes the ordering contract.

**Tidy the new e2e tests.** Six promotions are created and none deleted; line 1804 of the same file has the `afterAll` and `deletePromotionDocument` pattern to follow. `CanceledOrderFragment` and `canceledOrderGuard` are declared twice, at 2510 and 2610. The two discount types can be told apart with `discount.type` instead of set membership across lines, which currently works only because T_4 falls outside `productVariantIds`.

**Comments.** "See #5127 for the full reasoning" appears four times. The reasoning that matters is one sentence and belongs in `quantityBasisFor()` alone. "to keep the invariant `quantityBasisFor()` relies on" should name the invariant: `PROMOTION` amounts are stored scaled to the current quantity. The comment recording why the tests tell discounts apart by source rather than format is review correspondence, it belongs in this thread rather than the file.

## On the rounding drift

Agreed on leaving it. Worth an issue rather than a blocker, with one addition: it compounds across successive partial cancellations of the same line, which my single-reduction check would not have caught.

Fix the migration and the three other blocking items and I will approve. The core change itself I am happy with.

wakqasahmed:
Thanks — you were right that the signal exists, and the helper was double-scaling exactly as you described.

`rescaleOrderLinePromotionAdjustments()` now decides per line instead of rescaling everything that matches the `WHERE`. The candidate query pulls three more things for each row: the latest `stock_movement.createdAt` where `type = 'CANCELLATION'` for that line, the latest `order_modification.createdAt` reached through that line's `OrderModificationLine` rows (matched on `order_line_reference.modificationId IS NOT NULL`, which avoids depending on the discriminator value), and the sum of those rows' quantity deltas. From those it picks a basis:

- No modification at all: the whole reduction came from cancellations, so the amounts are still on the placement basis and get scaled by `quantity / orderPlacedQuantity`, as before.
- A modification with no later cancellation: skipped entirely. This is your -2975 case — `applyPriceAdjustments()` already wrote the right number and the helper now leaves it alone.
- Modified and then cancelled: rescaled, but from `orderPlacedQuantity + sum(modification deltas)` rather than `orderPlacedQuantity`, which is the quantity the last modification left the line at. A line placed at 20, modified down to 10 and then cancelled to 2 scales by 2/10, not 2/20.

One case I could not close, and it is now in the docblock rather than glossed over: cancelling stock that was allocated but never fulfilled records a `Release`, not a `Cancellation`, and `Release` rows are written by `modifyOrder` just as much as by `cancelOrderByOrderLines`, so they say nothing about which came last. On a line that also has a modification, there is no way to order the two, and the helper leaves that line at its pre-migration amounts rather than risk shrinking a correct one. Lines that were only ever cancelled are unaffected, since they have no modification timestamp to compare against at all.

The new `v3_8_orderline_promotion_rescale.spec.ts` drives the helper against a fake `QueryRunner` that records every statement. `leaves a line reduced by an OrderModification untouched` is the regression test for the case you gave — 20 units, 50% off, reduced to 2 by `modifyOrder`, stored at -2975 — and it asserts no `UPDATE` is issued at all; on the previous helper that row would have been rewritten to -297. `rescales a modified-then-cancelled line from its post-modification quantity` pins the interleaved basis, and `leaves a line untouched when the modification is more recent than the cancellation` covers the other ordering.

On the other three:

Full cancellation — the method is now `OrderLine.setQuantityRescalingAdjustments()`, which assigns `this.quantity` itself (so neither caller has to remember to) and skips the rescale entirely when the new quantity is 0, leaving the stored amounts as the record of what had applied. Two unit tests in `order-line.entity.spec.ts` cover the scale and the leave-alone.

SQL — `escLiteral()` is gone. Both statements go through `queryRunner.connection.driver.createParameter()` so the placeholders come out right per dialect, the id is passed as the value from the row rather than a quoted string, and the per-row loop is replaced by one chunked `UPDATE ... SET adjustments = CASE id WHEN ? THEN ? ... END WHERE id IN (...)` of up to 100 rows. A test asserts the JSON never appears in the SQL text and that a description containing both a quote and a backslash comes back byte-identical.

Prettier — reverted. Worth flagging for whoever hits it next: `bun.lock` pins prettier 3.7.4, but a bare `npx prettier` in a fresh checkout here resolved 3.9.6, which collapses that union onto one line, which is how it crept back in. `npx prettier@3.7.4 --check` is clean on every file this PR touches.

From the should-fix list I have also moved the `quantity === 0` guard into `quantityBasisFor()`, swapped the `orderPlacedQuantity: 0` test for a `PROMOTION` adjustment on a `quantity === 0` line, deleted the six promotions in an `afterAll`, hoisted the duplicated `CanceledOrderFragment`/guard to the top-level describe, switched the combo tests to `discount.type`, and cut the repeated `#5127` narration back to the one explanation in `quantityBasisFor()`.

On the combo assertions — I dropped the total and the derived `totalWithTax` check. For the `cancelOrder` path nothing recalculates, so I now pin the reduced line's per-unit order-level share to its placement value and the untouched line's share to its full placement value, both captured before the cancellation. For the `modifyOrder` path the discount really is redistributed by weight, so I assert the reduced line still carries a share and that the untouched line's share cannot have shrunk, and left it there rather than pin numbers that the whole-order shortfall makes unstable. If you would rather see exact values on that one, tell me what the split should be and I will pin it.

I could not run the e2e suite in this environment, so the four e2e tests are unverified since the last CI run — the unit suites (`order-line.entity.spec.ts`, `order-calculator.spec.ts`, the new migration spec) are green locally, and `prettier@3.7.4 --check` passes. Happy to open the two follow-up issues you mentioned, for the rounding drift and the whole-order shortfall, unless you would rather file them yourself.

wakqasahmed:
Pushed a fix for CI, no production code change:

- The PR title was capitalized (sentence-case), which `subject-case` flags. Retitled to `fix(core): reconcile read/write quantity basis for OrderLine discounts (#5127)`.
- The new `cancelOrder` combo test was failing on all four CI databases: `expected -21 to be close to -21.25`. The assertion divided two already-rounded line totals (order-level share / quantity) and compared with `toBeCloseTo(...,6)`, but the share is rounded to the nearest minor unit *before* the division, so scaling down from 20 units to 2 introduces up to half a cent of drift on the per-unit figure — not a real regression. Rewrote it to scale the placement rate by the reduced quantity and round the same way production does, then compare totals directly.

`deploy` and the yarn `test_package_managers` failures are unrelated (missing `VERCEL_TOKEN` on fork PRs, and an unrelated ESM-linking error in `@vendure/create` itself) — not touching those.

packages/core/src/migration-utils/v3_8_orderline_promotion_rescale.ts
```diff
@@ -0,0 +1,313 @@
...
+                ol.${orderLineQuantity} AS ${esc('quantity')},
+                ol.${orderLinePlacedQuantity} AS ${esc('orderPlacedQuantity')},
+                ol.${orderLineAdjustments} AS ${esc('adjustments')}
+         FROM ${orderLineTable} ol
+         WHERE ol.${orderLineQuantity} < ol.${orderLinePlacedQuantity}
+           AND ol.${orderLineQuantity} > 0
+           AND ol.${orderLinePlacedQuantity} > 0`,
+    );
+
+    const modifications: ModificationRow[] = await queryRunner.query(
```
wakqasahmed, line 194: **Blocking:** This reconstructs recalculation history only from `OrderModificationLine` rows that reference the candidate line, but `modifyOrder()` calls `applyPriceAdjustments()` on the whole order and `applyOrderItemPromotions()` clears and rebuilds `PROMOTION` adjustments for every `order.lines` entry. For example: partially cancel line B from 20 to 10, then modify only line A. The later modification rewrites line B to the correct quantity-10 basis, but B has no modification-line row here; `modifications.length === 0` therefore treats it as placement-basis data and scales the correct amount by 10/20 again. Please reconstruct the latest `OrderModification` per order (joining through the OrderLine order relation) and add a two-line regression for cancel-B then modify-A.

wakqasahmed: Addressed in 0c64fb1. Candidate rows now carry the metadata-resolved order ID, and modification history reads `OrderModification.order` directly and groups timestamps per order, so a later modification of line A is also recognized as the recalculation point for line B. The new two-line cancel-B/modify-A regression failed before the fix by scaling B from -10000 to -5000 and now leaves both lines untouched. Focused migration spec: 15/15; ESLint and Prettier clean.

packages/core/src/entity/order-line/order-line.entity.ts
```diff
@@ -13,8 +13,8 @@ import { Channel } from '../channel/channel.entity';
 import { CustomOrderLineFields } from '../custom-entity-fields';
 import { EntityId } from '../entity-id.decorator';
 import { Money } from '../money.decorator';
-import { Order } from '../order/order.entity';
 import { OrderLineReference } from '../order-line-reference/order-line-reference.entity';
+import { Order } from '../order/order.entity';
```
biggamesmallworld, line 17: This moves `Order` below `OrderLineReference`, which trips the repo's `import/order` rule (`eslint` warns here) and is the last of the unrelated import churn from the first round. Put it back above `OrderLineReference`, where the merge base has it.

packages/core/src/migration-utils/v3_8_orderline_promotion_rescale.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+                        return databaseName ? { databaseName } : undefined;
+                    },
+                    findRelationWithPropertyPath: (propertyPath: string) => {
+                        const databaseName = names.relations?.[propertyPath];
+                        return databaseName ? { joinColumns: [{ databaseName }] } : undefined;
+                    },
+                };
+            },
+        },
+        query: vi.fn((sql: string, params: any[] = []) => {
```
biggamesmallworld, line 109: All 15 tests drive this mock, so they check the SQL as a string and never as SQL. I ran the generated statements against a real sqlite and a real postgres by hand and they are fine, so this is not blocking. Recording it because the next person to change the query has no test that would catch a syntax or type-coercion error. I will open a follow-up for one real-engine case rather than ask you for it here.

packages/core/e2e/order-modification.e2e-spec.ts
```diff
@@ -2606,6 +2615,287 @@ describe('Order modification', () => {
...
+            // order-level discount below.
+            expect(modifiedDiscountedLine.discountedLinePriceWithTax).toBe(placedUnitPriceWithTax);
+            expect(modifiedOtherLine.discountedLinePriceWithTax).toBe(placedOtherLine.unitPriceWithTax);
+
+            // The order-level $5 discount is still split across both lines. `modifyOrder` runs a
+            // full recalculation, which redistributes it by weight, so the reduced line's much
+            // smaller weight can only move share towards the untouched line - it must not leave
+            // either line without one.
+            const modifiedLines = await getLinesByVariant(placedOrder.id);
+            expect(orderLevelShareOf(modifiedLines.get('T_1')!)).toBeLessThan(0);
```
biggamesmallworld, line 2833: These two inequalities do catch the corruption the test exists for: a reduced line whose prorated price went hugely negative loses its share entirely, and `T_4` absorbing more would fail the second one. They would not catch a partial drift in the split. The post-modification distribution is deterministic, so the two shares could be pinned as integers instead. Not worth another round on its own.

biggamesmallworld (approved):
All four blocking items from the last round are addressed and the gates are green. Ready to merge.

## Blocking items

**1. Migration helper must exclude modified lines.** Addressed, and by a better route than the one I proposed. I asked for a per-line modification lookup through `OrderModificationLine`. That would have been wrong: `modifyOrder()` ends in `applyPriceAdjustments()`, which rewrites `PROMOTION` amounts for every line in the order, not just the ones the `OrderModification` references. `getStoredQuantityBasis()` groups modifications per order and cancellations per line out of `ORDER_CANCELLATION` history, which is why the cancel-B-then-modify-A case you found in your own re-read now comes out right. The three branches are correct: no modification at all scales from `orderPlacedQuantity`; a modification later than every cancellation is skipped; a cancellation after the last modification scales from `quantity + sum(post-modification cancelled quantities)`. Throwing on an equal-timestamp history rather than guessing is the right call for a helper that rewrites money.

**2. Fully cancelling a line destroys its stored discount.** Addressed. `setQuantityRescalingAdjustments()` assigns the quantity and leaves the adjustments alone at 0.

**3. Build the migration SQL with parameters.** Addressed. `escLiteral()` is gone, every statement binds through `driver.createParameter()`, and the per-row loop is now a chunked `CASE ... WHEN` update. The IDs are bound as the values the `SELECT` returned, so the integer-against-string comparison went away with it.

**4. Prettier on `order-modifier.ts`.** Addressed, and thanks for pinning down the 3.7.4-vs-3.9.6 difference. `--check` is clean on all six changed files here.

Unit coverage for the helper: 15 tests across every branch, plus the parameter binding and the chunking.

## Should-fix items

All addressed. The combo tests capture each line's share before the change, the self-derived `totalWithTax` assertion is gone, the `orderPlacedQuantity: 0` case is now a `quantity === 0` case, the guard moved into `quantityBasisFor()`, `setQuantityRescalingAdjustments()` assigns `this.quantity` itself, the e2e promotions are deleted in an `afterAll`, the duplicated `CanceledOrderFragment`/guard is down to one, `discount.type` tells the two types apart, and the repeated `#5127` narration is down to one place.

## What I ran

- `packages/core` unit suite: 78 files, 1192 tests, all pass.
- `prettier --check` and `eslint` on the changed files: clean, apart from one import-order warning noted inline.
- `tsc`: no errors in any source file this PR touches. The 653 errors in the package are pre-existing ones in e2e and spec files, from codegen output that is not generated locally.
- Call-site audit: `quantityBasisFor()` is private with three call sites, all inside the entity, and all three short-circuit on `quantity === 0` before calling, so the `basis === 0 ? 1 : basis` line is defence in depth rather than load-bearing. No stale references to the old `rescaleAdjustmentsForQuantity` name.
- The helper's spec drives a fake `QueryRunner`, so its SQL had not reached a database engine. I ran the generated statements against a real sqlite and a real postgres over a three-line fixture covering cancel-only, modify-then-cancel and cancel-then-modify. Both engines execute them and produce -2975, -2000 and untouched, as intended. MySQL and MariaDB are unverified, but the statements are the same shape as the sqlite ones.

## One edit I made

The "Update, addressing the migration review" section still described the earlier approach: the latest `stock_movement.createdAt` of type `CANCELLATION`, modifications reached through `OrderModificationLine`, and a `Release`-shaped blind spot documented in the docblock. None of that is what the helper does now, and that text feeds the changelog, so I rewrote the section to match the current implementation rather than send this back for another round. The code is untouched. Please check I have not misdescribed anything.

I will open a follow-up for a real-engine test of the helper rather than hold this up for it.

PR #5144, opened 2026-08-17
feat(create): add TanStack Start storefront option

dlhck:
# Description

Adds a generic storefront registry to `@vendure/create` and exposes TanStack Start alongside Next.js in interactive and CI creation flows. Starter definitions now own repository references, scripts, environment contracts, and documentation metadata so additional storefronts can be added cleanly. No related issue.

# Breaking changes

No; the legacy `--with-storefront` flag continues to select Next.js.

# Screenshots

Not applicable for this CLI-only change.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5144"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789542097&installation_model_id=20193&pr_number=5144&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5144&signature=52b5b6f8dae7c4232bf73efdc4b6bc3859aa07c0e0c880152a2b2e645bff50dd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx
+++ b/docs/docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx
@@ -2,7 +2,7 @@
 title: "MultiChannelStockLocationStrategy"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts" sourceLine="32" packageName="@vendure/core" since="3.1.0" />
+<GenerationInfo sourceFile="packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts" sourceLine="35" packageName="@vendure/core" since="3.1.0" />
 
 The MultiChannelStockLocationStrategy is an implementation of the [StockLocationStrategy](/reference/typescript-api/products-stock/stock-location-strategy#stocklocationstrategy).
 which is suitable for both single- and multichannel setups. It takes into account the active
--- a/docs/docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx
+++ b/docs/docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx
@@ -2,7 +2,7 @@
 title: "MultiChannelStockLocationStrategy"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts" sourceLine="35" packageName="@vendure/core" since="3.1.0" />
+<GenerationInfo sourceFile="packages/core/src/config/catalog/multi-channel-stock-location-strategy.ts" sourceLine="32" packageName="@vendure/core" since="3.1.0" />
 
 The MultiChannelStockLocationStrategy is an implementation of the [StockLocationStrategy](/reference/typescript-api/products-stock/stock-location-strategy#stocklocationstrategy).
 which is suitable for both single- and multichannel setups. It takes into account the active
--- a/docs/docs/reference/typescript-api/request/request-context.mdx
+++ b/docs/docs/reference/typescript-api/request/request-context.mdx
@@ -45,7 +45,7 @@ class RequestContext {
     userHasPermissions(permissions: Permission[]) => boolean;
     userHasAllPermissions(permissions: Permission[]) => boolean;
     serialize() => SerializedRequestContext;
-    copy() => RequestContext;
+    copy(channel?: Channel) => RequestContext;
     req: Request | undefined
     apiType: ApiType
     channel: Channel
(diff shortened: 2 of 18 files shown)


michaelbromley (commented):
Thanks for this — the registry design is the right call. `StorefrontId` is derived from `STOREFRONT_STARTERS` rather than maintained as a parallel union, and each starter carries its own `packageScripts` and `environment` functions, so adding a third starter needs no change to `create-vendure-app.ts`. Pulling `selectStorefront()` out of the duplicated quick-start and manual prompt paths was a good cleanup, and the new spec file uses exact assertions rather than `toBeTruthy()`.

A few things to sort before merge.

## Code

**1. Unrelated changes in the diff.** The PR includes `packages/core/src/service/services/shipping-method.service.ts` and five regenerated docs files (`order-service.mdx`, `request-context.mdx`, `shipping-method-service.mdx`, `multi-channel-stock-location-strategy.mdx`, `manifest.json`). These come from master changes the branch has not been rebased onto, picked up when the docs generator ran. Please rebase so the file list only shows the storefront work.

**2. `quoteEnvironmentValue` has an untested throw path.** `storefront-starters.ts` is otherwise well covered, but the specs only exercise a value containing `#` and `"`. Nothing covers the backtick fallback, and nothing covers the error branch. The value being escaped comes from the user's project name, so both branches are reachable from ordinary input.

**3. The `--with-storefront` fallback is not testable.** This line in `create-vendure-app.ts` is the whole backward-compatibility contract:

```ts
const selectedCiStorefront = options.storefront ?? (options.withStorefront ? 'nextjs' : undefined);
```

It sits in top-level script code, so no test can reach it. That includes the case where both flags are passed. Extracting it into an exported function would match what you have already done with `selectStorefront()` and `parseStorefrontId()`.

**4. The starter repos' own `dev` and `start` scripts are replaced.** `configureStorefrontPackageJson` spreads `storefront.packageScripts(context)` over whatever the downloaded `package.json` contains, so `@vendure/create` owns those two script entries. If `tanstack-starter-vendure` or `nextjs-starter-vendure` ever needs a flag on `dev`, the scaffolded project drops that flag and the CLI output says nothing about it. Worth a comment recording that the starter repos must keep those two scripts free of required flags.

**5. `ref: 'v1.0.0'` for TanStack, `ref: 'main'` for Next.js.** A commit to `nextjs-starter-vendure` on `main` reaches every new project immediately, with no way to scaffold against a known-good state. If that is deliberate, say so in a comment; otherwise pin both.

## Prose

**6. `packages/create/README.md`** — "The legacy `--with-storefront` flag continues to select the Next.js starter."

"Legacy" and "continues to" both describe the change rather than the behaviour, so a reader who never saw this PR gets nothing from them. The sentence also never says which flag to use instead.

> `--with-storefront` also selects the Next.js starter. Use `--storefront nextjs` instead.

**7. `packages/create/README.md`** — "In CI mode, include one of the available storefront starters. Currently supported values are `tanstack` and `nextjs`."

Two sentences carrying one fact, and "currently" dates the sentence without naming a version.

> Selects the storefront to scaffold in CI mode. Supported values are `tanstack` and `nextjs`.

**8. `packages/create/src/create-vendure-app.ts:101`** — `'Include Next.js storefront with --ci (deprecated)'`

"(deprecated)" tells the reader to stop using the flag without telling them what replaces it.

> `'Include the Next.js storefront with --ci. Deprecated: use --storefront nextjs'`

**9. `packages/create/src/storefront-starters.ts`** — `'Environment values cannot contain single, double, and backtick quotes together.'`

This names no key, no value and no action, and the user hitting it is midway through a scaffold. The only value that can trigger it is the project name, so the message can say exactly what to do. Passing the key into `quoteEnvironmentValue` would let it read:

> `Cannot write ${key} to the storefront env file: the value contains a single quote, a double quote and a backtick, leaving no character to quote it with. Choose a project name without one of those characters.`

**10. `packages/create/src/storefront-starters.ts`** — both starters describe themselves as "A <Framework> storefront with routing, internationalization, and checkout".

These render as the hint beside the label in the select prompt, where the label already gives the framework name. Two options with identical claimed features leave the user nothing to choose on. The hints should name what actually differs between the two, such as build tooling or deployment target.

**11. `packages/create/src/helpers.ts:114`** — the concatenated strings produce `You are running Node v20.10.0.Vendure requires Node ...`. The space is missing after the first sentence. This predates the PR, but the surrounding lines are already being edited.

**12. `packages/create/README.md:7`** — "Vendure Create requires [Node.js](https://nodejs.org/en/) v18+ to be installed." This does not match `REQUIRED_NODE_VERSION`. Also predates the PR, and also in a file this PR already touches.

PR #5151, opened 2026-08-18
fix(core): preserve OrderLine relation custom fields when merging orders

### LINKED ISSUE #5073: Relation custom fields on OrderLine are lost when a guest order is merged on login
When a signed-out customer's order is merged into an existing customer order at login, any relation custom field on the carried-over `OrderLine`s is silently set to null. Scalar custom fields survive.

`toMergedOrderLine` copies `line.customFields`, which exposes a relation under its field name (`foo`) and does not load it by default. The merged lines are then re-inserted via `addItemsToOrder`, and `CustomFieldRelationService.updateRelations` only attaches a relation when it finds the GraphQL input key (`fooId`). That key is never present, so the relation is dropped without any error.

Lines already on the existing order keep their relations, because `updateRelations` skips absent keys. Only lines carried over from the guest order are affected.

## Reproduction

Define a relation custom field on `OrderLine`, then:

1. Sign in as a customer and add an item, so an active order exists.
2. Sign out, and as a guest add an item with the relation custom field set. Confirm it is set on the guest line.
3. Sign in as the same customer.
4. The line carried over from the guest order now has the relation set to null.

## Versions

Reproduced on 3.6.5. `toMergedOrderLine` is unchanged on `master` (3.7.1).

## Interaction with #5012

On the `minor` branch, non-list relation custom fields now expose a `<name>Id` property on the entity, so `line.customFields` carries the id and the merge path should work by side effect.

List relation custom fields are explicitly out of scope there, since their ids live in a join table rather than a column, so those would still be dropped by a merge. This seems worth a regression test covering both cases, rather than relying on the id property being present.

## Workaround

For anyone hitting this before a fix lands, a custom `OrderMergeStrategy` can read the relation custom fields back and re-express them under their input keys. Reading the relation custom field configs from `ConfigService` at `init()` keeps this generic, and `getGraphQlInputName` gives the correct key for both single (`fooId`) and list (`fooIds`) relations:

```ts
export class RelationPreservingMergeStrategy implements OrderMergeStrategy {
    private connection: TransactionalConnection;
    private relationFields: RelationCustomFieldConfig[] = [];

    init(injector: Injector) {
        this.connection = injector.get(TransactionalConnection);
        this.relationFields = injector
            .get(ConfigService)
            .customFields.OrderLine.filter(
                (config): config is RelationCustomFieldConfig => config.type === 'relation',
            );
    }

    async merge(ctx: RequestContext, guestOrder: Order, existingOrder: Order) {
        const relationIds = await this.getRelationCustomFieldIds(ctx, guestOrder.lines);
        // ...then spread `relationIds.get(guestLine.id)` into the customFields of each
        // MergedOrderLine built from a guest line.
    }

    private async getRelationCustomFieldIds(ctx: RequestContext, lines: OrderLine[]) {
        const result = new Map<ID, Record<string, ID | ID[]>>();
        if (!this.relationFields.length || !lines.length) {
            return result;
        }
        const linesWithRelations = await this.connection.getRepository(ctx, OrderLine).find({
            where: { id: In(lines.map((l) => l.id)) },
            relations: this.relationFields.map((config) => `customFields.${config.name}`),
        });
        for (const line of linesWithRelations) {
            const customFields: Record<string, ID | ID[]> = {};
            for (const config of this.relationFields) {
                const relation = (line.customFields as Record<string, any>)?.[config.name];
                if (config.list) {
                    if (Array.isArray(relation) && relation.length) {
                        customFields[getGraphQlInputName(config)] = relation.map((r) => r.id);
                    }
                } else if (relation) {
                    customFields[getGraphQlInputName(config)] = relation.id;
                }
            }
            if (Object.keys(customFields).length) {
                result.set(line.id, customFields);
            }
        }
        return result;
    }
}
```

One related note while looking at this: `OrderModifier.customFieldsAreEqual` skips the comparison for a relation custom field whenever the incoming input value is falsy. Adding a variant that already sits on a line with a relation set, this time without that relation, therefore matches the existing line and merges into it rather than creating a separate one. Happy to raise that separately if it is worth its own issue.

deep231w:
Relation-type custom fields on `OrderLine` are not loaded by default, which breaks guest order merging on login.

## Problem

`MergeOrdersStrategy.findCorrespondingLine` decides whether two lines are the same by comparing `JSON.stringify(line.customFields)`. Because the relation is never loaded, every line's relation custom field reads as absent, so two lines of the same variant carrying *different* relations compare equal and get merged into one. The guest line's relation is then lost.

With this config:

```ts
customFields: {
    OrderLine: [
        { name: 'relationField', type: 'relation', entity: Product, graphQLType: 'Product' },
    ],
}
```

the database holds the FK, but the loaded entity does not:

```text
customFieldsRelationfieldid = 1
```

```ts
customFields: { inscription: 'bar', relationField: null }
```

## Solution

`OrderService.mergeOrders` now hydrates the relation custom fields on both the guest order and the existing order before handing them to `OrderMerger`, and only when `configService.customFields.OrderLine` actually declares a relation field.

The hydrated values are written under the GraphQL input name (`relationFieldId`, or `relationFieldIds` for a list) because that is the shape `OrderMerger.getLinesToInsert` passes on to `addItemsToOrder`, which resolves relations via `CustomFieldRelationService.updateRelations`.

The load is scoped to this one call site. `findOne` is unchanged, so no other caller pays for extra joins.

## Result

Lines that differ only by a relation custom field are no longer merged, and the guest line's relation survives the merge.

## Tests

Three cases added to `order-merge.e2e-spec.ts`:

- guest-only relation is preserved through the merge
- lines with different relations stay separate
- lines with the same relation merge, taking the guest quantity

Fixes #5073
# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5151"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789604726&installation_model_id=20193&pr_number=5151&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5151&signature=df7af1926b468150dcc8a994d1ed694640428f6c469f25aa4dfca418d234ebe1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


deep231w:

biggamesmallworld (changes requested):
Thanks — diagnosis is right, and I confirmed the new spec fails 3/5 on master and passes 5/5 with the fix (also covers `list: true`). Two blockers:

### 1. Merge behaviour regression

Hydrating relations makes `findCorrespondingLine` treat lines that differ only by a relation custom field as distinct, so the guest line goes to `linesToInsert`. Downstream still expects the GraphQL input shape (`singleId`), not a hydrated `{ single: Product }`, so `customFieldsAreEqual` skips the check and quantities are summed instead of replaced.

```
existing: T_1 qty 1, single=T_4
guest:    T_1 qty 3, single=T_1

master:  1 line, qty 3, single=T_4
this PR: 1 line, qty 4, single=T_4
```

Guest relation is lost either way; this PR also breaks the documented replace semantics. Convert to `<name>Id` at the merger (or accept both shapes downstream), and add a test for this case.

### 2. Scope the load

Bug is one call site (`order.service.ts:2188`). Don't force-load every OrderLine relation custom field on every `findOne` (incl. `activeOrder`). Add to the default relation list when `relations` is omitted, or pass them only at that call. Prefer deriving from `configService.customFields.OrderLine` (same as `order-modifier.ts:845`), not TypeORM metadata.

### Nits

- Format the diff (`bun run format`).
- Fold into `order-merge.e2e-spec.ts` (~15 lines); drop tests 1/3/4; use `loginResultGuard` and assert by variant id + line count.
- Comment should explain *why* hydration is needed on the merge path, not the obvious join.

biggamesmallworld (changes requested):
Closer. Scoping is right now (blocker 2 done), and the two new tests pass locally.

Blocker 1 is only half fixed. Hydration is guest-only, so the existing order's lines still carry no relation data. Same variant with the *same* relation on both orders now regresses:

```
existing: T_1 qty 1, relationField=T_4
guest:    T_1 qty 3, relationField=T_4

master:  1 line, qty 3   (replace)
this PR: 1 line, qty 4   (summed)
```

Guest line gets `{ relationFieldId: 'T_4' }`, existing line gets `{}`, `findCorrespondingLine` misses, the line goes to `linesToInsert`, and `addItemToOrder` sums the quantity. Normalize both orders (or do the conversion inside the merger) and add this case as a test:

```ts
existingOrderLines: [{ productVariantId: 'T_1', quantity: 1, customFields: { relationFieldId: 'T_4' } }],
guestOrderLines:    [{ productVariantId: 'T_1', quantity: 3, customFields: { relationFieldId: 'T_4' } }],
// expect 1 line, quantity 3
```

Also: `packages/core/e2e/order-merge-custom-fields.e2e-spec.ts` should be deleted. The tests were folded into `order-merge.e2e-spec.ts` as asked, but the original file is still in the diff, duplicating coverage. It is also the only file failing `bun run format`.

biggamesmallworld (approved):

PR #5152, opened 2026-08-18
fix(dashboard): transform JSX when compiling plugin config imports

biggamesmallworld:
Takeover of #3726 by @hans-rollingridges-dev, who diagnosed this a year ago and is credited as co-author on the commit. The original was opened against the pre-#4561 `ts.createProgram` implementation and no longer applied, so the one-line fix is rebased onto the current `ts.transpileModule` compiler here.

## The bug

The dashboard's config compiler walks the Vendure config's import graph and transpiles each file with `ts.transpileModule()`. No `jsx` option was set, so any `.tsx` file reachable from the config, for example server-side React used for PDF invoice or email template rendering, was emitted with its JSX untransformed.

TypeScript reports no diagnostic for this, so the failure only surfaces when the compiled config is imported, and it surfaces as a misleading `Could not find a variable exported as VendureConfig with the name "config"`. That error names the wrong thing entirely, which is why the original report took work to pin down.

## The fix

`jsx: ts.JsxEmit.ReactJSX` in `packages/dashboard/vite/utils/compiler.ts`.

## Coverage

The original had none. Added here: a `vite/tests/fixtures-jsx` fixture whose plugin imports a `.tsx`, and `vite/tests/jsx-compilation.spec.ts` running it in both commonjs and esm mode.

With the `jsx` line removed, both cases fail with the exact `Could not find a variable exported as VendureConfig` error from the original report. With it in place the full `vite/tests` suite passes: 14 files, 110 tests.

## Verification

- `bunx vitest run vite/tests` — 14 files passed, 1 skipped; 110 tests passed, 4 skipped
- `bunx tsc --project tsconfig.vite.json --noEmit` — exit 0
- `bunx prettier --check` on the changed files — pass
- eslint not run: `packages/dashboard/eslint.config.js` scopes to `**/*.{js,jsx}`, so both changed files report "File ignored because no matching configuration was supplied"

Relates to #3726

Co-authored-by: RollingRidges <hans@rollingridges.dev>

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5152"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789632036&installation_model_id=20193&pr_number=5152&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5152&signature=cea8c28333925f632603fa3a23024242ea4d9a1716dd9a6f53c3908189909ea4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/reference/typescript-api/common/currency-code.mdx
+++ b/docs/docs/reference/typescript-api/common/currency-code.mdx
@@ -2,7 +2,7 @@
 title: "CurrencyCode"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/common/src/generated-types.ts" sourceLine="1135" packageName="@vendure/common" />
+<GenerationInfo sourceFile="packages/common/src/generated-types.ts" sourceLine="1150" packageName="@vendure/common" />
 
 ISO 4217 currency code
 
--- a/docs/docs/reference/typescript-api/common/job-state.mdx
+++ b/docs/docs/reference/typescript-api/common/job-state.mdx
@@ -2,7 +2,7 @@
 title: "JobState"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/common/src/generated-types.ts" sourceLine="2390" packageName="@vendure/common" />
+<GenerationInfo sourceFile="packages/common/src/generated-types.ts" sourceLine="2405" packageName="@vendure/common" />
 
 The state of a Job in the JobQueue
 
--- a/docs/docs/reference/typescript-api/common/language-code.mdx
+++ b/docs/docs/reference/typescript-api/common/language-code.mdx
@@ -2,7 +2,7 @@
 title: "LanguageCode"
 generated: true
 ---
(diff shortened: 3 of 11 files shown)


HouseinIsProgramming (approved):
Good fix, clean takeover of #3726. The one-line jsx emit option is the right place for this, and the fixture reproduces the original misleading error exactly.

PR #5154, opened 2026-08-18
fix(telemetry-plugin): bump opentelemetry to 0.222.x (ghsa-q7rr-3cgh-j5r3, ghsa-45rx-2jwx-cxfr)

biggamesmallworld:
Closes [Dependabot alert 639](https://github.com/vendurehq/vendure/security/dependabot/639). Relates to OSS-715.

## The vulnerabilities

**GHSA-q7rr-3cgh-j5r3.** The Prometheus exporter in `@opentelemetry/sdk-node` below 0.217.0 crashes the Node process on a malformed HTTP request to the metrics endpoint. `_requestHandler` calls `new URL()` with no error handling, so a request like `http://` throws an uncaught `TypeError`. The endpoint is unauthenticated and binds to `0.0.0.0` by default. CVSS 7.5. Patched in 0.217.0, which returns a 400 instead.

**GHSA-45rx-2jwx-cxfr (CVE-2026-59892).** `@opentelemetry/propagator-jaeger` below 2.9.0 calls `decodeURIComponent` on incoming `uber-trace-id` and `uberctx-*` headers with no error handling. A bare `%` throws an uncaught `URIError` during context extraction and terminates the process. Patched in 2.9.0.

Vendure never instantiates the Prometheus exporter, and the Jaeger propagator is only reachable if you set `OTEL_PROPAGATORS=jaeger` or supply your own `NodeSDK` config. Both alerts apply to the dependency regardless.

### Span attribute names change for every user of the plugin

The 0.221 instrumentations removed the `OTEL_SEMCONV_STABILITY_OPT_IN` switch and emit only the stable semantic conventions. There is no opt-out. This applies whether or not you pass your own `instrumentations` array, because the curated default uses the same packages.

| Old name | New name |
|---|---|
| `http.method` | `http.request.method` |
| `http.status_code` | `http.response.status_code` |
| `http.url` | `url.full` |
| `http.target` | `url.path` and `url.query` |
| `http.scheme` | `url.scheme` |
| `http.host` | `server.address` |
| `http.flavor` | `network.protocol.version` |
| `http.user_agent` | `user_agent.original` |
| `http.client_ip` | `client.address` |
| `net.peer.name`, `net.host.name` | `server.address` |
| `net.peer.port`, `net.host.port` | `server.port` |
| `net.peer.ip` | `network.peer.address` |
| `db.system` | `db.system.name` |
| `db.statement` | `db.query.text` |
| `db.name` | `db.namespace` |
| `db.user`, `db.connection_string` | removed, no replacement |
| `http.request_content_length`, `http.response_content_length` | removed, no replacement |

Span names and `http.route` are unchanged.

Grafana dashboards and alerts keyed on the old names will return nothing. The telemetry guide now carries this table plus step by step migration instructions, including an OpenTelemetry Collector `transform` fallback that copies the new attributes back onto the old names for anyone who cannot update dashboards immediately.

### Query parameters are redacted on incoming spans

0.222 extends sensitive query parameter redaction to spans for incoming requests. Earlier versions only did this for outgoing requests.

If a request URL carries any of `sig`, `Signature`, `AWSAccessKeyId`, `X-Goog-Signature`, `X-Amz-Signature`, `X-Amz-Credential` or `X-Amz-Security-Token`, the value is replaced with `REDACTED` in the `url.query` attribute. Server spans do not set `url.full`, so no other attribute changes, and the request itself is untouched. The new `redactedQueryParamsServer` option on `HttpInstrumentation` overrides the list.

### Other breaking changes in range

- **0.222** Four `fail-fast` changes to `sdk-node` covering Propagator, MeterProvider, TracerProvider and Resource creation. All four only affect declarative config-file setup through `startNodeSDK()`. Users of `new NodeSDK()`, which is what the Vendure docs use, are unaffected.
- **0.215** `LogRecordExporter` now requires a `forceFlush()` method. Affects custom log exporters only.
- **0.220** `new BatchLogRecordProcessor(exporter)` becomes `new BatchLogRecordProcessor({ exporter })`. The docs preload example is updated.
- **0.207** `OTEL_TRACES_EXPORTER=jaeger` no longer resolves. Jaeger has native OTLP support.
- **0.208 to 0.221** `NoopLogger`, `ProxyLogger`, `NOOP_LOGGER_PROVIDER` and `ProxyLoggerProvider` removed from `api-logs`. Affects direct imports only.
- **core 2.6** `OTEL_RESOURCE_ATTRIBUTES` parsing follows the spec strictly, so one malformed entry voids the whole variable. The documented `service.name=...` form is unaffected.

Everything else in range was gRPC, browser/XHR, declarative config, or metrics-reader paths the plugin does not touch.

## Why every `@opentelemetry/*` package moves together

OpenTelemetry for JavaScript publishes one release under two version numbers. Packages it treats as stable get a `2.x` number (`core`, `resources`, `sdk-trace-base`); packages it treats as experimental get a `0.2xx` number (`sdk-node`, `sdk-logs`, the exporters, the instrumentations). Both halves ship together, minutes apart, and the experimental packages pin the stable ones to that release's exact version:

| `sdk-node` | pins `core` / `resources` / `sdk-trace-base` at |
|---|---|
| 0.220.0 | 2.9.0 |
| 0.221.0 | 2.10.0 |
| 0.222.0 | 2.11.0 |

So `0.222.0` and `2.11.0` are the same release wearing two numbers, not a new thing and an old thing.

This matters because a caret on a `0.x` version pins the minor, while a caret on a `2.x` version does not. Asking for `^0.221.0` alongside `^2.10.0` freezes the experimental half at the July release while letting the stable half drift to the August one, which requests half of one release and half of another. npm resolves that by installing both.

That was measurable. Resolving the previous revision's dependency set without a lockfile produced **5 duplicated packages**, with `@opentelemetry/core` appearing at 16 separate paths. Moving to the matched 0.222 / 2.11 pair and pinning the seven siblings `sdk-node` pins exactly brings that to **0 duplicated packages**. The `instrumentation-*` packages stay on carets, because nothing pins them and the carets still let patches through.

This also adds `@opentelemetry/api-logs` and `@opentelemetry/sdk-trace-base` to `dependencies`. Both are imported by the plugin (`src/config/otel-logger.ts`, `src/instrumentation.ts`) and resolved until now only because `sdk-node` and `sdk-logs` hoisted them, which fails under pnpm and Yarn PnP.

## Why 0.222 and not 0.219 or 0.221

An earlier revision pinned 0.219, which is above the Prometheus patch floor and below the semantic convention break, so telemetry output was unchanged. That is no longer viable: the Jaeger propagator fix only lands in the 0.221 line, so the attribute rename became unavoidable, which is why this targets `minor`.

Given the rename has to happen either way, this goes to the current head of the ecosystem rather than sitting one release behind it. 0.222 adds no semantic convention changes on top of 0.221, its breaking changes do not reach `new NodeSDK()` users, every contrib instrumentation now has a release built against `^0.222.0`, and the resulting dependency tree has no duplicates at all.

## Docs

- `docs/docs/guides/how-to/telemetry/index.mdx`: the two breaking changes now live in their own top-level `Upgrading from earlier versions` section, carrying the rename table, a four step dashboard migration, and the incoming-span query parameter redaction. They previously sat inside `Customising instrumentation`, which is opt-in material, so they were invisible to anyone not customising `instrumentations` and absent from the page's table of contents. The install step now links to the new section.
- Contrib pins in both install examples updated to the releases built against `@opentelemetry/instrumentation@0.222.0`: `auto-instrumentations-node@0.80.0`, `instrumentation-kafkajs@0.31.0`, `-express@0.70.0`, `-graphql@0.70.0`, `-ioredis@0.70.0`, `-mysql2@0.68.0`, `-nestjs-core@0.68.0`, `-pg@0.74.0`.
- The preload example now has its own install line, so users no longer depend on hoisting.

## Verification

- Diffed the published `instrumentation-http` tarballs. Both 0.221.0 and 0.222.0 emit the same 17 stable `ATTR_*` constants, with no legacy `SEMATTRS_*` imports and no `OTEL_SEMCONV_STABILITY` references. `build/src/semconv.js` still exists but no longer holds the stability switch; it now carries a copy of unstable convention definitions, currently only `user_agent.synthetic.type`.
- Same diff on `instrumentation-pg` 0.73.0 to 0.74.0, `-mysql2` 0.67.0 to 0.68.0 and `-ioredis` 0.69.0 to 0.70.0. All three emit an identical attribute set across the bump, so the rename table above covers database spans and needs no changes for 0.222.
- `span.updateName()` is identical in both `http.js` builds, so span names stay `GET /shop-api`.
- Confirmed `exporter-prometheus@0.222.0` `_requestHandler` catches the `URL` throw and returns `400 Bad Request`, and that `propagator-jaeger` resolves to 2.11.0, above the 2.9.0 floor.
- Resolved the published dependency set with npm and no lockfile: 40 OpenTelemetry packages, 0 duplicated. `bun.lock` likewise now holds exactly one copy of every `@opentelemetry/*` package.
- End to end with the documented dev-server env vars: a local collector received `/v1/traces`, `/v1/logs` and `/v1/metrics`.
- Plugin builds clean. `dev-server` typechecks with no OpenTelemetry errors.

The [earlier manual verification comment](https://github.com/vendurehq/vendure/pull/5154#issuecomment-5329047725) on this PR was run against the 0.219 revision. Its conclusion that attribute names are unchanged no longer holds.

## Follow-up, not in this PR

`@vendure/telemetry-plugin/preload` could re-export `NodeSDK`, `BatchSpanProcessor`, `BatchLogRecordProcessor` and the two exporters. That would remove the separate install line from the docs and make a version mismatch impossible by construction.


biggamesmallworld:
## Manual verification

Ran the documented instrumented setup end to end against this branch, since nothing in CI exercises the telemetry plugin.

### The fix

The vulnerable path is never reached by the plugin itself, so it was tested directly against the exporter. A raw socket sends `GET http:// HTTP/1.1`, which is the malformed request-target that breaks `new URL()`:

| `@opentelemetry/exporter-prometheus` | Result |
|---|---|
| 0.200.0 (before) | uncaught `TypeError: Invalid URL`, process exits |
| 0.219.0 (after) | `HTTP/1.1 400 Bad Request`, process survives |

### No regression in emitted telemetry

Started `bun run dev:server:instrumented` in `packages/dev-server` with the `jaeger`, `loki` and `grafana` services from the root `docker-compose.yml`, then sent Shop API queries.

Traces arrived in Jaeger under service `vendure-dev-server` with the expected tree:

```
POST /shop-api
 └ middleware - corsMiddleware / _cookieSession / jsonParser
   └ graphql.resolve products / activeChannel
     └ ChannelService.getChannelFromToken, ListQueryBuilder.applyWhereCondition
       └ pg.query:SELECT vendure-dev
```

Spans from Vendure's own `@Instrument` decorator are present, so the `ENABLE_INSTRUMENTATION_ENV_VAR` path set by `getSdkConfiguration` still works.

The attribute keys across all collected spans:

```
http.method   http.status_code   http.route   http.url   http.target
http.flavor   http.scheme        http.host    http.user_agent
net.peer.name net.peer.ip        net.peer.port net.host.name
db.system     db.statement       db.name      db.user   db.connection_string
```

These are the old semantic conventions, identical to what 0.200 emitted. There are no occurrences of `http.request.method`, `server.address` or `db.namespace`, which confirms that `instrumentation-http`, `-express`, `-graphql`, `-nestjs-core` and `-pg` are all unchanged in output. Existing dashboards and alerts keyed on these names keep working. This is the reason for pinning 0.219 rather than 0.221.

Logs arrived in Loki under `service_name=vendure-dev-server`, which confirms `OTEL_LOGS_EXPORTER` and `OTEL_RESOURCE_ATTRIBUTES` still resolve after the 0.212 move of environment variable handling into `sdk-node` and the stricter `OTEL_RESOURCE_ATTRIBUTES` parsing in core 2.6.

### Not covered

A consumer installing the published package resolves its own dependency tree rather than this workspace's hoisting, so CI's install is the only check on that.

Separately, `src/instrumentation.ts` imports `@opentelemetry/api-logs` and `@opentelemetry/sdk-trace-base` while neither appears in `dependencies`. They resolve transitively through `sdk-node` today. That predates this PR and is left alone here, but this bump changes what gets hoisted, so it is worth fixing separately.

packages/telemetry-plugin/package.json
```diff
@@ -35,20 +35,20 @@
...
+        "@opentelemetry/instrumentation-graphql": "^0.67.0",
+        "@opentelemetry/instrumentation-http": "^0.219.0",
+        "@opentelemetry/instrumentation-ioredis": "^0.67.0",
+        "@opentelemetry/instrumentation-mysql2": "^0.65.0",
+        "@opentelemetry/instrumentation-nestjs-core": "^0.65.0",
+        "@opentelemetry/instrumentation-pg": "^0.71.0",
+        "@opentelemetry/instrumentation-runtime-node": "^0.32.0",
+        "@opentelemetry/resources": "^2.8.0",
+        "@opentelemetry/sdk-logs": "^0.219.0",
+        "@opentelemetry/sdk-node": "^0.219.0",
```
mehringer68: Two packages are imported by this plugin but not declared here, and this PR moves both.

* `@opentelemetry/sdk-trace-base`: `src/instrumentation.ts:19` imports `BatchSpanProcessor`, `ConsoleSpanExporter` and `SimpleSpanProcessor`. The lockfile moves it 2.0.0 to 2.8.0.
* `@opentelemetry/api-logs`: `src/config/otel-logger.ts:1` imports `logs` and `SeverityNumber`. The lockfile moves it 0.200.0 to 0.219.0.

Both resolve today only because `sdk-node` and `sdk-logs` hoist them. Under pnpm or Yarn PnP the published package fails to resolve them at runtime.

This is not a deliberate policy of leaning on `sdk-node` to supply things: `@opentelemetry/resources` and `@opentelemetry/context-async-hooks` are also transitive deps of `sdk-node`, and both are declared explicitly right here. So it is an oversight, and this PR is a good moment to correct it since it is touching every one of these versions anyway.

```json
"@opentelemetry/api-logs": "^0.219.0",
"@opentelemetry/sdk-trace-base": "^2.8.0",
```

mehringer68 (commented):
The dependency analysis holds up, and I checked the load-bearing claim independently rather than taking it on trust. `@opentelemetry/instrumentation@0.219.0` initialises `semconvStabilityFromStr` to `SemconvStability.OLD` and only moves on an explicit `OTEL_SEMCONV_STABILITY_OPT_IN` token, and the four instrumentations with a semconv migration in range (pg 0.71.0, mysql2 0.65.0, ioredis 0.67.0, nestjs-core 0.65.0) all read that env var. Old attribute names survive by default. 0.219 is the right pick.

Two things to fix before merge. One is inline on `package.json`. The other is the docs.

## The telemetry guide hands users the exact skew this PR avoids

`docs/docs/guides/how-to/telemetry/index.mdx` installs the OpenTelemetry packages with no version pinned, at line 206:

```bash
npm install @opentelemetry/auto-instrumentations-node
```

and at line 227:

```bash
npm install @opentelemetry/instrumentation-http \
            @opentelemetry/instrumentation-express \
            ...
```

Both resolve to the 0.221 line today. `@opentelemetry/auto-instrumentations-node@0.79.0` depends on `@opentelemetry/instrumentation: ^0.221.0`, and `@opentelemetry/instrumentation-http` latest is `0.221.0`.

I diffed the published 0.221.0 and 0.219.0 tarballs:

* 0.219.0 `build/src/http.js` contains both `SemconvStability.OLD` and `SemconvStability.STABLE`, gated on `OTEL_SEMCONV_STABILITY_OPT_IN`. Default output is `http.method`, `net.peer.name`, `http.url`.
* 0.221.0 has no semconv stability code at all. The old attribute constants are gone from the build. Only `ATTR_HTTP_REQUEST_METHOD`, `ATTR_SERVER_ADDRESS` and `ATTR_URL_FULL` remain. The rename is unconditional, with no opt-out.

The guide's own callout at line 267 says that passing `instrumentations` **replaces** the curated default rather than merging with it. So the user's 0.221 `HttpInstrumentation` is not a duplicate sitting alongside ours, it is the only one registered. Every HTTP span silently changes attribute names, and their existing dashboards and alerts stop matching.

The duplicate install is guaranteed as well, not merely likely. 0.221.0 pins `@opentelemetry/core` to exactly `2.10.0` and `@opentelemetry/instrumentation` to exactly `0.221.0`, against the `^0.219.0` in this PR. No dedupe is possible, so users also get a second `InstrumentationBase` distinct from the one `getSdkConfiguration()` registers against.

The description anticipates this and argues it away:

> Users pinning their own `@opentelemetry/*` packages should align them with the 0.219 line, otherwise their processors run against a newer `NodeSDK`. That skew is possible with any bump, not new here.

True, and beside the point. This is the bump that makes the currently published guide wrong, and the fix is pinning versions in two code fences.

Concretely: pin the contrib packages to the versions in `packages/telemetry-plugin/package.json`, use `@opentelemetry/auto-instrumentations-node@0.77.0` for Option 1 and `@opentelemetry/instrumentation-kafkajs@0.28.0` for Option 2, since those are the releases built against `@opentelemetry/instrumentation@0.219.0`.

While that file is open: the `preload.ts` example at line 144 imports `@opentelemetry/sdk-logs`, `@opentelemetry/sdk-node`, `@opentelemetry/sdk-trace-base` and both exporters, but the only install instruction on the page is `npm install @vendure/telemetry-plugin`. Users are being told to depend on hoisting. Same root cause as the inline comment below.

docs/docs/guides/how-to/telemetry/index.mdx
```diff
@@ -263,6 +275,15 @@ const config = getSdkConfiguration({
 });
 ```
 
+:::warning
+Install these instrumentations at the versions shown, not at `latest`. From
+`@opentelemetry/instrumentation-http@0.221.0` onwards the HTTP instrumentation emits only the
+stable semantic conventions, with no opt-out: `http.method` becomes `http.request.method` and
+`net.peer.name` becomes `server.address`. Since your `instrumentations` array replaces the
```
michaelbromley, line 282: this is not easy to understand

michaelbromley (dismissed):
This should target the `minor` branch as it has the potential to break existing apps due to the APIs of the otel packages changing

docs/docs/guides/how-to/telemetry/index.mdx
biggamesmallworld: I've re-written this to be more clear

biggamesmallworld: I've also included a short migration guide for otel consuming applications

michaelbromley (approved):

PR #5156, opened 2026-08-19
feat(dashboard): user-aware navigation and route visibility context

kevmtt:
# Description

The dashboard currently has no way to decide what to show based on who is logged in. `NavMenuItem.requiresPermission` is the only per-viewer condition available, and permissions answer a different question: what a user may call on the API, not what they should see in the UI. Floor staff who legitimately hold `ReadCatalog` and `ReadOrder` cannot have the catalog and order entries hidden, and marketplace sellers cannot have menu visibility controlled independently of their roles.

This adds a user context to the dashboard and one way to use it in the nav menu. Background and API discussion in https://github.com/vendurehq/vendure/discussions/5153.

What is new:

- `DashboardUserContext`, exposed through the `useDashboardUserContext()` hook. Carries the active administrator, their roles, their Administrator custom fields, the accessible channels, the active channel, `hasPermissions()` and `isSuperAdmin`.
- `isVisible` on nav menu items and sections. A predicate evaluated on every nav render, ANDed with `requiresPermission`.
- `setNavVisibility` and `keepOnlyNavEntries` helpers, which compose predicates rather than replacing them. A plain spread of `isVisible` silently discards a predicate another plugin already set. Both are meant to be called from the existing function form of `navSections`, which is how a plugin reaches an entry it did not declare.
- `NavMenuTarget`, the `{ sections, items }` shape both helpers take. Sections and nested items are named separately, because the two may carry the same id: the built-in Customers section and Customers item both do.
- `BUILT_IN_NAV_SECTION_IDS` and `BUILT_IN_NAV_ITEM_IDS`. `defaults.ts` and the built-in shortcut map are built from them, so neither can drift from the ids they name. Both helpers also warn in development when an id matches no entry, which is otherwise silent and, with `keepOnlyNavEntries`, hides every other entry.
- `requiresPermission` on a nav section is now applied. It typechecked before, through the `Omit` of `NavMenuItem`, and did nothing. No built-in section declares one, so nothing in a stock dashboard changes.

Supporting changes:

- `CurrentUserInformation` now selects the administrator's roles.
- Administrator custom fields load through a separate query gated on server config, because `CurrentUserInformation` is itself the query that unblocks the server config request which populates the custom field map.
- Nav filtering and sorting moved out of `nav-main.tsx` into a pure `resolveNavMenu()`, which is where the new rules are applied and where the tests point. It takes a `userContextPending` flag, so while the administrator custom fields are in flight only the entries carrying a predicate are withheld and the rest of the sidebar paints immediately.
- `registerRoutes` builds nav items with a spread instead of field by field, which was silently dropping any newly added optional field.

There is no new nav extension point. An earlier revision added `navMenuTransforms`, functions run on every nav render with the user context. That is dropped: the function form of `navSections` already receives the whole `NavMenuConfig` after every array-form registration, so it can attach an `isVisible` predicate to any entry, and that predicate is what runs per render with the context. Two nav extension points with different timing cannot be merged afterwards, and this one would have put third-party functions in the sidebar render path on every render. `keepOnlyNavEntries` gained an optional `when` predicate so a whitelist can still depend on the user without one.

Hiding a nav entry changes what the dashboard renders and nothing else. A user who can call an operation on the Admin API can still call it whether or not a link is shown. The guide, the type docblocks and both helpers all say so, and real access control still belongs on the server.

Everything is additive for existing installs. `isVisible` is optional and absent means visible, no built-in id changes, and the function form of `navSections` keeps its exact current timing and signature. The one behaviour change is that `requiresPermission` on a section is honoured rather than ignored.

# Breaking changes

None.

# Screenshots

```TS
/**
 * Role code the fulfillment staff are given. Change this in one place if you name
 * the role differently when creating it in the dashboard.
 */
const FULFILLMENT_ROLE_CODE = 'fullfillment';

/**
 * keepOnlyNavEntries is a whitelist: everything not named here is hidden, including
 * entries contributed by other plugins. 'secret' is listed so the Pssst section from
 * SecretPagePlugin survives for fulfillment staff who have the flag set.
 */
const KEEP_FOR_FULFILLMENT = { sections: ['fulfillment', 'secret'] };

const isFulfillmentStaff = (ctx: DashboardUserContext) =>
    ctx.roles.some(role => role.code === FULFILLMENT_ROLE_CODE);

defineDashboardExtension({
    navSections: [
        {
            id: 'fulfillment',
            title: 'Fulfillment',
            icon: TruckIcon,
            order: 250,
            placement: 'top',
        },
    ],
    routes: [fulfillmentOrderList],
});

// A second call, because `navSections` is either the array form or the function form,
// and this plugin needs both: the array form declares the section that the route above
// attaches to, the function form adjusts the assembled menu.
defineDashboardExtension({
    // Presentation only. The role still needs ReadOrder, and a fulfillment user can
    // reach any other route by typing its URL.
    navSections: config => keepOnlyNavEntries(config, KEEP_FOR_FULFILLMENT, isFulfillmentStaff),
});

/**
 * Visible only when the Administrator custom field is set. Reading customFields is
 * safe here without a readiness check of its own: the framework withholds this entry
 * until the user context has resolved, so a rule never observes them as absent.
 */
const canSeeSecret = (ctx: DashboardUserContext) => ctx.administrator?.customFields?.showSecretPage === true;

defineDashboardExtension({
    navSections: [
        {
            id: 'secret',
            title: 'Pssst!',
            icon: SparklesIcon,
            order: 900,
            placement: 'top',
            isVisible: canSeeSecret,
        },
    ],
    routes: [secretPage],
});
```

| Header | Image |
|--------|--------|
| User with permission to see Orders, doesn't see Orders navigation because of a role | image |
| Admin Account received access to a secret page via custom field | image |
| After having the customField toggled true | image
|

# Checklist

📌 Always:

👍 Most of the time:

Tests: unit tests across `resolve-nav-menu`, `nav-menu-helpers`, `dashboard-user-context`, `nav-main` and the two new hooks, plus a core e2e assertion that `activeAdministrator` resolves the nested role channels, which is the relation depth the dashboard query needs. The existing dashboard e2e suite still runs in a browser to confirm login and sidebar rendering work end to end.

Docs: the navigation guide has a new User-Aware Navigation section, and the generated reference pages are regenerated.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5156"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789720807&installation_model_id=20193&pr_number=5156&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5156&signature=6ec44a6d6b432aba5146fdec0b251e1ad51380b737b0b681b511c5c07426307e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kevmtt:
I can provide/commit the code with the test plugins as well if required

biggamesmallworld (changes requested):
Two things to fix before merge. Everything else I found is in a separate comment and is non-blocking.

### 1. A throwing predicate can un-hide an entry another plugin hid

`nav-menu-helpers.ts:13` composes as `ctx => added(ctx) && existing(ctx)`. The `added`-first ordering correctly protects the case covered by `nav-menu-helpers.spec.ts:114`, where the *existing* predicate throws. The inverse is not covered: when `existing` hides and `added` throws, the composed predicate throws, `isVisibleFor` fails open, and the entry is shown.

```
existing hides + added throws  ->  false becomes true
```

That breaks the guarantee stated in `docs/docs/guides/extending-the-dashboard/navigation/index.mdx:312`: "any plugin can hide an entry and no plugin can force one to appear."

Minimal fix:

```ts
return ctx => {
    let a;
    try {
        a = added(ctx);
    } catch {
        a = true; // a broken predicate must not un-hide
    }
    return a && existing(ctx);
};
```

Worth a test mirroring the one at line 114, with the throw on the other side.

### 2. `useAdminCustomFields` runs for every install

`use-dashboard-user-context.ts:31` calls it unconditionally, and its `enabled` at `use-admin-custom-fields.ts:78` only checks `serverConfig` and `user?.id`. So a vanilla dashboard with no `isVisible` predicate and no transform still pays an extra Admin API round trip on boot for a feature it does not use.

`nav-main.tsx:156` already computes `hasUserDependentRules`. That signal belongs in the `enabled` condition, not only in the render gate.

biggamesmallworld:
Non-blocking follow-ups from the same pass. Take or leave any of these.

**Types**

- `(ctx: DashboardUserContext) => boolean` is written out at `nav-menu-helpers.ts:5`, `nav-menu-extensions.ts:137` and `types/navigation.ts:133`. Export one named type and use it in all three.
- `DashboardNavSectionDefinition` is a standalone interface that re-lists fields from `NavMenuSection`, so it has already drifted: `NavMenuSection` extends `Omit<NavMenuItem, 'url' | 'shortcut'>` and therefore has `requiresPermission`, which the definition type does not offer. `DashboardNavMenuItemDefinition` one file up derives correctly. Pre-existing, not introduced here, but this PR adds another hand-copied field to it.
- `dashboard-user-context.ts` carries four hand-maintained mirrors of the same entity (lines 11, 31, 67, 132). `BuildDashboardUserContextInput.administrator` in particular is `DashboardAdministrator` minus `customFields`, so it can be written as an `Omit<...>` rather than a fourth copy.

**Duplication**

- The active-channel lookup at `use-dashboard-user-context.ts:36` is the same code as the one in `usePermissions`. The comment says "Same idiom as usePermissions". Extract it once and have both call it.

**Small things**

- `resetNavMenuWarnings` (`resolve-nav-menu.ts:46`) carries `@description` and `@since 3.8.0`, which are the docs-generation markers, but `resolve-nav-menu.js` is not exported from `index.ts` and only the specs import it. Drop the tags, or scope `warnedIds` per `resolveNavMenu` call so tests do not need a global reset.
- `keepOnlyNavItems(config, [])` blanks the whole sidebar with no warning and no test. An empty list from a mistyped config variable is a plausible mistake with a confusing symptom. Worth either a dev-mode warning or a test that records the behaviour as intentional.
- The orphan `sectionId` warning in `addNavMenuItem` is a good change but unrelated to user-aware navigation. It would get a better look in its own PR. Also, it reads more naturally as the `else` of the branch above rather than re-testing `sectionIndex === -1` afterwards.

**Comments and prose**

Comment density is high: 47 of 103 lines in `nav-menu-helpers.ts`, 39 of 111 in `use-admin-custom-fields.ts`, 22 of 57 in `use-dashboard-user-context.ts`.

Some of it earns its place. The `andPredicate` ordering rationale, the `placeholderData: undefined` note and the "do not hoist `addCustomFields`" comment all explain a why that is not reconstructable from the code. Keep those. The rest:

- "presentation only ... never authorization" appears seven times in hand-written source and docs, which propagates to fifteen occurrences once the reference pages are generated. Once in the `DashboardUserContext` docblock and once in the guide is enough.
- Comments that narrate the change rather than describe the code: "Field-by-field construction silently dropped new optional fields such as isVisible" in `logic/navigation.ts`, and "Ported unchanged from nav-main.tsx" in `resolve-nav-menu.ts`. The reader never saw the previous version, and the second is wrong as soon as anyone edits the function.
- `use-dashboard-user-context.ts:41` explains that a hook returning a fresh object literal would break memoization. The dependency array already says that.
- In the guide: "and this trips people up" (line 316) and "One consequence is worth stating plainly" (line 318) carry no information. At line 333, "not formally guaranteed" and "not covered by a stability policy" are the same statement twice.

For what it is worth, the parts I liked: extracting `resolveNavMenu` as a pure function and pointing the tests at it is the right call and would have been worth doing on its own; the tests use exact assertions rather than `toBeDefined()`; and chaining transforms in a registered array beats the last-writer-wins `navMenuModifiers` pattern.

biggamesmallworld:
Both blockers from my last pass are addressed. I ran the nav and hook specs locally on the branch, 61 tests pass across the 6 files.

On the predicate composition: wrapping `added` and warning through `warnOnce` is the right shape, and the `resolveNavMenu` test that asserts the entry is absent from the resolved output rather than merely false in isolation is the one that actually pins the guarantee. On the custom fields: `hasUserDependentRules` now gates the request as well as the render, and it is covered at all three levels.

`dashboard lint` is still red. All three findings are in files this PR adds:

- `packages/dashboard/src/lib/framework/user-context/dashboard-user-context.ts:56` has `// eslint-disable-next-line @typescript-eslint/no-empty-interface`. That rule is not present in the current config, so the directive itself is the error: `Definition for rule '@typescript-eslint/no-empty-interface' was not found`. This is the only hard error, and it is what fails the job.
- `packages/dashboard/src/lib/framework/nav-menu/nav-menu-extensions.ts:189`, unused `no-console` disable. Warning only.
- `packages/dashboard/src/lib/framework/nav-menu/resolve-nav-menu.ts:39`, same. Warning only.

Deleting all three lines makes eslint clean on these files, which I confirmed locally. The empty `AdministratorCustomFields` interface is not flagged once the stale directive is gone, so it can stay as it is.

I have edited the PR title to lowercase the subject, which clears the `conventional commit format` check.

One non-blocking observation on the composition fix, for a later pass rather than this one: only `added` is wrapped, so an `isVisible` written directly in a nav config that throws still escapes and fails open. That is not a regression and it cannot un-hide a predicate composed through the helpers, since every helper-added predicate goes through the wrapper.

biggamesmallworld (dismissed):
### Scope: drop the `navMenuTransforms` extension point, keep everything else

Both blockers from my earlier passes are fixed and the branch is green. What follows is a scope decision about the public API surface, not a defect in your work.

`navMenuTransforms` duplicates something the dashboard already ships. The function form of `navSections` receives the whole `NavMenuConfig` and returns a new one. It runs once at load, so it cannot itself decide anything about the user. It does not need to. It can attach an `isVisible` predicate, which this PR then evaluates on every render with the context:

```ts
navSections: config => setNavVisibility(config, ['catalog', 'orders'], ctx => !isFloorStaff(ctx)),
```

I ran this against your branch. Both helpers work unchanged when called from the function form, `hasUserDependentRules` still detects the attached predicate through its scan of the registry items, and the composition guarantee still holds, because a later plugin's modifier receives the earlier plugin's config.

```ts
const builtIns: NavMenuConfig = {
    sections: [
        { id: 'catalog', title: 'Catalog', placement: 'top', items: [{ id: 'products', title: 'Products', url: '/products' }] },
        { id: 'fulfillment', title: 'Fulfillment', placement: 'top', items: [{ id: 'pos', title: 'POS', url: '/pos' }] },
    ],
};
const isFloorStaff = (c: DashboardUserContext) => c.roles.some(r => r.code === 'floor-staff');

it('setNavVisibility attached at load time hides a built-in at render time', () => {
    // what a plugin returns from `navSections: config => ...`
    const atLoad = setNavVisibility(builtIns, ['catalog'], c => !isFloorStaff(c));
    // note the empty transforms array
    expect(resolveNavMenu(atLoad, ctx('floor-staff'), []).map(s => s.id)).toEqual(['fulfillment']);
    expect(resolveNavMenu(atLoad, ctx('admin'), []).map(s => s.id)).toEqual(['catalog', 'fulfillment']);
});

it('two plugins compose without clobbering', () => {
    const a = setNavVisibility(builtIns, ['catalog'], c => !isFloorStaff(c));
    const b = setNavVisibility(a, ['catalog'], () => true); // later plugin cannot un-hide
    expect(resolveNavMenu(b, ctx('floor-staff'), []).map(s => s.id)).toEqual(['fulfillment']);
});
```

Three tests pass, `keepOnlyNavItems` included. Your fulfillment example produces the same sidebar through `navSections` as it does through `navMenuTransforms`.

So I would like to drop:

- the `navMenuTransforms` field on `DashboardExtension`
- `addNavMenuTransform`, `getNavMenuTransforms` and `NavMenuTransform`
- `applyTransforms` and the third parameter of `resolveNavMenu`
- the `navMenuTransforms` registry entry and its type
- the `getNavMenuTransforms().length > 0` branch in `hasUserDependentRules`

and keep everything else: `setNavVisibility`, `keepOnlyNavItems`, `andPredicate` and its fail-open guarantee, `DashboardUserContext`, `useDashboardUserContext`, `useAdminCustomFields`, `isVisible`, the pure `resolveNavMenu`, the `registerRoutes` spread fix and the orphan `sectionId` warning.

Two reasons for cutting it. It puts third-party functions in the sidebar render path on every render. It also commits us to two nav extension points with different timing, which we cannot merge afterwards. The existing function form has too little adoption to justify that split: in this repo it has one consumer, `packages/dev-server/test-plugins/nav-modifier-plugin`, which exists to demonstrate it, and in one large production Vendure project I checked, 9 plugins call `defineDashboardExtension` and 3 declare `navSections`, all array form.

What this loses is real but narrower than it first looks. A predicate can only hide. A transform which retitles an entry for one role, or inserts one only some users get, cannot be written with a predicate. Nobody has asked for that yet, and when they do the cheaper answer is to pass the context to the function form and move it to render time, which those same adoption numbers make close to free.

Two things which would help either way, and which the guide needs if the function form becomes the documented route:

- Export the built-in nav ids as constants. `keepOnlyNavItems(config, ['orders'])` depends on strings in `defaults.ts` which we have never declared stable, and an id which matches nothing hides every other entry silently.
- Document the `navSections` plus `setNavVisibility` idiom in the User-Aware Navigation section. It becomes the way to reach entries your plugin did not declare.

kevmtt:
Dropped, in 39cbaa203. Gone: the `navMenuTransforms` field, `NavMenuTransform`, `addNavMenuTransform`, `getNavMenuTransforms`, the registry entry and its type, `applyTransforms` and the third parameter of `resolveNavMenu`, and the transform branch in `hasUserDependentRules`. Everything on your keep list stays.

I verified your claim before cutting: phase 1 of `executeDashboardExtensionCallbacks` registers every array-form section and every route's `navMenuItem`, phase 2 runs the modifier functions and writes the result back with `setNavMenuConfig`, so a predicate attached from the function form is in the registry array that `NavMain` scans. Your two tests are in `resolve-nav-menu.spec.ts` now, plus one in `define-dashboard-extension.spec.ts` covering the round trip through the registry.

Two things came out of doing it.

`keepOnlyNavItems` needed a predicate. The guide's example was "floor staff see only the POS section", and `keepOnlyNavItems(config, ['pos'])` from the function form hides everything else from everybody, because the function form has no context. It now takes an optional third argument:

```ts
navSections: config => keepOnlyNavItems(config, ['pos'], isFloorStaff),
```

`setNavVisibility` already took a predicate, so it needed nothing.

A plugin that declares its own section and also adjusts the assembled menu needs two `defineDashboardExtension` calls, since `navSections` is either the array form or the function form. The array form has to run first so a route's `navMenuItem.sectionId` has somewhere to attach. That is the fulfillment test plugin's shape now, and the guide says it.

On your other two points:

Built-in ids are exported as `BUILT_IN_NAV_SECTION_IDS` and `BUILT_IN_NAV_ITEM_IDS`. `defaults.ts` is built from them rather than repeating the strings, so a constant cannot drift from the id it names. Separately, both helpers now warn in development when an id matches no entry, which covers the case a constant cannot: another plugin's id, or a typo in one.

The guide's User-Aware Navigation section is rewritten around `navSections` plus the helpers, including the composition rules, the whitelist consequence and the two-call note. The reference pages are regenerated.

63 unit tests pass across the nav menu, user context and the two hooks; all 526 `src/lib` tests pass.

biggamesmallworld (approved):

biggamesmallworld (approved):

PR #5157, opened 2026-08-19
feat(cli): add Console project linking

dlhck:
# Description

Adds `vendure console link`, `status`, and `unlink` with secure browser approval, strict v1 manifest validation, safe monorepo project selection, and atomic local writes. Updates the CLI documentation and generated-project source-control guidance, with unit, HTTP integration, and compiled CLI e2e coverage. Relates to [PDEV-49](https://linear.app/vendure/issue/PDEV-49/deliver-add-vendure-console-linkstatusunlink-cli-commands).

# Breaking changes

None.

# Screenshots

Not applicable; this is a CLI-only change.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5157"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789731917&installation_model_id=20193&pr_number=5157&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5157&signature=912184f944fce99a7908eb7864e57ad80e056734cbffe661d9e563d2fa1d010c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -218,7 +218,7 @@
           "title": "CLI",
           "slug": "cli",
           "file": "docs/guides/developer-guide/cli/index.mdx",
-          "lastModified": "2026-08-19T13:44:35+02:00"
+          "lastModified": "2026-08-24T10:48:37+02:00"
         },
         {
           "title": "AI-assisted Development",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -218,7 +218,7 @@
           "title": "CLI",
           "slug": "cli",
           "file": "docs/guides/developer-guide/cli/index.mdx",
-          "lastModified": "2026-08-17T11:43:39+02:00"
+          "lastModified": "2026-08-19T13:44:35+02:00"
         },
         {
           "title": "AI-assisted Development",
@@ -2780,7 +2780,7 @@
                   "title": "MultiChannelStockLocationStrategy",
                   "slug": "multi-channel-stock-location-strategy",
                   "file": "docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
                 },
                 {
                   "title": "ProductVariantPriceCalculationStrategy",
@@ -2874,7 +2874,7 @@
                   "title": "RequestContext",
                   "slug": "request-context",
                   "file": "docs/reference/typescript-api/request/request-context.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
                 },
                 {
                   "title": "RequestContextService",
@@ -3096,7 +3096,7 @@
                   "title": "OrderService",
                   "slug": "order-service",
                   "file": "docs/reference/typescript-api/services/order-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
                 },
                 {
                   "title": "OrderTestingService",
@@ -3186,7 +3186,7 @@
                   "title": "ShippingMethodService",
                   "slug": "shipping-method-service",
                   "file": "docs/reference/typescript-api/services/shipping-method-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
                 },
                 {
                   "title": "SlugService",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -218,7 +218,7 @@
           "title": "CLI",
           "slug": "cli",
           "file": "docs/guides/developer-guide/cli/index.mdx",
-          "lastModified": "2026-08-24T11:50:56+02:00"
+          "lastModified": "2026-08-24T13:59:38+02:00"
         },
         {
           "title": "AI-assisted Development",
@@ -2780,7 +2780,7 @@
                   "title": "MultiChannelStockLocationStrategy",
                   "slug": "multi-channel-stock-location-strategy",
                   "file": "docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
                 },
                 {
                   "title": "ProductVariantPriceCalculationStrategy",
@@ -2874,7 +2874,7 @@
                   "title": "RequestContext",
                   "slug": "request-context",
                   "file": "docs/reference/typescript-api/request/request-context.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
                 },
                 {
                   "title": "RequestContextService",
@@ -3096,7 +3096,7 @@
                   "title": "OrderService",
                   "slug": "order-service",
                   "file": "docs/reference/typescript-api/services/order-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
                 },
                 {
                   "title": "OrderTestingService",
@@ -3186,7 +3186,7 @@
                   "title": "ShippingMethodService",
                   "slug": "shipping-method-service",
                   "file": "docs/reference/typescript-api/services/shipping-method-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
                 },
                 {
                   "title": "SlugService",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -218,7 +218,7 @@
           "title": "CLI",
           "slug": "cli",
           "file": "docs/guides/developer-guide/cli/index.mdx",
-          "lastModified": "2026-08-24T10:48:37+02:00"
+          "lastModified": "2026-08-24T11:04:10+02:00"
         },
         {
           "title": "AI-assisted Development",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -218,7 +218,7 @@
           "title": "CLI",
           "slug": "cli",
           "file": "docs/guides/developer-guide/cli/index.mdx",
-          "lastModified": "2026-08-24T11:04:10+02:00"
+          "lastModified": "2026-08-24T11:50:56+02:00"
         },
         {
           "title": "AI-assisted Development",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -2780,7 +2780,7 @@
                   "title": "MultiChannelStockLocationStrategy",
                   "slug": "multi-channel-stock-location-strategy",
                   "file": "docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx",
-                  "lastModified": "2026-08-13T17:32:23+01:00"
+                  "lastModified": "2026-08-17T12:00:11+02:00"
                 },
                 {
                   "title": "ProductVariantPriceCalculationStrategy",
@@ -2874,7 +2874,7 @@
                   "title": "RequestContext",
                   "slug": "request-context",
                   "file": "docs/reference/typescript-api/request/request-context.mdx",
-                  "lastModified": "2026-08-13T14:40:56+02:00"
+                  "lastModified": "2026-08-17T12:00:11+02:00"
                 },
                 {
                   "title": "RequestContextService",
@@ -3096,7 +3096,7 @@
                   "title": "OrderService",
                   "slug": "order-service",
                   "file": "docs/reference/typescript-api/services/order-service.mdx",
-                  "lastModified": "2026-08-13T17:32:23+01:00"
+                  "lastModified": "2026-08-17T12:00:11+02:00"
                 },
                 {
                   "title": "OrderTestingService",
@@ -3186,7 +3186,7 @@
                   "title": "ShippingMethodService",
                   "slug": "shipping-method-service",
                   "file": "docs/reference/typescript-api/services/shipping-method-service.mdx",
-                  "lastModified": "2026-08-13T14:40:56+02:00"
+                  "lastModified": "2026-08-17T12:00:11+02:00"
                 },
                 {
                   "title": "SlugService",
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -4031,13 +4031,13 @@
                   "title": "UseDetailPage",
                   "slug": "use-detail-page",
                   "file": "docs/reference/dashboard/detail-views/use-detail-page.mdx",
-                  "lastModified": "2026-07-20T15:59:41+02:00"
+                  "lastModified": "2026-08-24T12:05:27Z"
                 },
                 {
                   "title": "UseGeneratedForm",
                   "slug": "use-generated-form",
                   "file": "docs/reference/dashboard/detail-views/use-generated-form.mdx",
-                  "lastModified": "2026-07-22T14:03:14Z"
+                  "lastModified": "2026-08-24T12:05:27Z"
                 }
               ],
               "lastModified": "2026-01-28T14:49:21+01:00"
--- a/packages/cli/src/commands/console/console.ts
+++ b/packages/cli/src/commands/console/console.ts
@@ -193,11 +193,6 @@ async function runConsoleCommand(
 }
 
 export function resolveConsoleEndpoints(env: NodeJS.ProcessEnv): ConsoleEndpoints {
-    if (env.VENDURE_CONSOLE_URL?.trim() || env.VENDURE_CONSOLE_API_URL?.trim()) {
-        throw new Error(
-            'Use the link-specific VENDURE_CONSOLE_LINK_URL and VENDURE_CONSOLE_LINK_API_URL variables for this command.',
-        );
-    }
     const consoleOverride = env.VENDURE_CONSOLE_LINK_URL?.trim() || undefined;
     const apiOverride = env.VENDURE_CONSOLE_LINK_API_URL?.trim() || undefined;
     if (Boolean(consoleOverride) !== Boolean(apiOverride)) {
--- a/packages/cli/src/commands/console/console.ts
+++ b/packages/cli/src/commands/console/console.ts
@@ -1,0 +1,593 @@
+import { confirm, isCancel, log } from '@clack/prompts';
+import { ChildProcess, spawn } from 'node:child_process';
+
+import { CliCommandExit } from '../../shared/cli-command-exit';
+import { isNonInteractiveEnvironment, withInteractiveTimeout } from '../../utilities/utils';
+
+import {
+    ManifestReadResult,
+    ProjectLinkManifest,
+    parseProjectLinkManifest,
+    readProjectLinkManifest,
+    removeProjectLinkManifest,
+    resolveProjectRoot,
+    writeProjectLinkManifestAtomic,
+} from './project-link-manifest';
+
+const DEFAULT_CONSOLE_URL = 'https://console.vendure.io';
+const DEFAULT_CONSOLE_API_URL = 'https://api.vendure.io';
+const POLL_INTERVAL_MS = 2_000;
+const REQUEST_TIMEOUT_MS = 10_000;
+const MAX_POLL_ATTEMPTS = 3;
+const RETRY_DELAYS_MS = [500, 1_000];
+
+export interface ConsoleCommandOptions {
+    project?: string;
+    force?: boolean;
+}
+
+export interface ConsoleReporter {
+    error(message: string): void;
+    info(message: string): void;
+    success(message: string): void;
+    warn(message: string): void;
+    url(value: string): void;
+}
+
+export interface ConsoleCommandDependencies {
+    cwd: string;
+    env: NodeJS.ProcessEnv;
+    fetch: typeof globalThis.fetch;
+    isNonInteractive: () => boolean;
+    now: () => number;
+    openUrl: (url: string) => Promise<void>;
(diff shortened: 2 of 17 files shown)


packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+}
+
+function pickTarget(projectRoot: string, chain: GitignoreFile[]): GitignoreFile {
+    const projectFile = chain.find(file => file.scope === 'project');
+    if (projectFile) {
+        return projectFile;
+    }
+    if (chain.length > 0) {
+        return chain[chain.length - 1];
+    }
```
biggamesmallworld: **Precedence bug: this write is a silent no-op.**

`collectGitignoreChain` pushes from `projectRoot` upward, so `chain[chain.length - 1]` is the file closest to the git root. Git resolves in favour of the **deepest** matching `.gitignore`, so any intermediate file that also matches wins.

Reproduced against real git:

```
root/.gitignore   node_modules            (tool writes **/.vendure/* + !**/.vendure/project.json here)
apps/.gitignore   **/.vendure/*           (pre-existing)

result:  { kind: 'updated', path: 'root/.gitignore' }
$ git check-ignore -v apps/vendure/.vendure/project.json
apps/.gitignore:1:**/.vendure/*   apps/vendure/.vendure/project.json
```

User is told the manifest can be committed; it is still ignored.

Deleting this branch falls through to the project-local return below, which is always correct because the deepest file wins. Verified: the manifest becomes committable (`check-ignore` exit 1) and `.vendure/cache.db` stays ignored (exit 0). It also fixes the `$HOME` case noted on `findGitRoot`.

```suggestion
```

packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+    return output;
+}
+
+function ensureGitignoreInRepo(projectRoot: string, gitRoot: string): ProjectLinkGitignoreResult {
+    const projectGitignorePath = getProjectLinkGitignorePath(projectRoot);
+    const chain = collectGitignoreChain(projectRoot, gitRoot);
+    const dirty = new Map<string, string>();
+
+    for (const file of chain) {
+        if (!hasDirectoryIgnore(file.content)) {
```
biggamesmallworld: **Rewriting every chain file changes tracking for unrelated projects.**

`isDirectoryIgnore` matches any path-prefixed form, so an unrelated `other-app/.vendure/` is replaced with a repo-wide `**/.vendure/*` plus `!**/.vendure/project.json`. Reproduced:

```
before:  other-app/.vendure/project.json -> check-ignore exit 0 (ignored)
after:   other-app/.vendure/project.json -> check-ignore exit 1 (NOT ignored, now shows in git status)
```

Linking `apps/vendure` un-ignored a file in `other-app`, which the user never named.

The ancestor rewrite is justified in principle (a root `.vendure/` does block a deeper `!.vendure/project.json`), but the fix should be scoped to the project being linked:

```
!apps/vendure/.vendure/
apps/vendure/.vendure/*
!apps/vendure/.vendure/project.json
```

packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+            continue;
+        }
+        const next = applyProjectLinkGitignoreRules(file.content, file.scope === 'project' ? 'local' : 'nested');
```
biggamesmallworld: Prettier fails (>110 cols). Same at line 117, and in `console.ts` and `project-link-manifest.ts`; `prettier --write` on the four files will settle it.

```suggestion
        const next = applyProjectLinkGitignoreRules(
            file.content,
            file.scope === 'project' ? 'local' : 'nested',
        );
```

packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+        if (next !== file.content) {
+            file.content = next;
+            dirty.set(file.path, next);
+        }
+    }
+
+    if (isProjectCovered(projectRoot, chain)) {
+        return writeDirty(dirty, coveringGitignorePath(projectRoot, chain), dirty.size > 0 ? 'updated' : 'unchanged');
```
biggamesmallworld:
```suggestion
        return writeDirty(
            dirty,
            coveringGitignorePath(projectRoot, chain),
            dirty.size > 0 ? 'updated' : 'unchanged',
        );
```

packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+        }
+        const parent = path.dirname(current);
+        if (parent === current) {
+            return files;
+        }
+        current = parent;
+    }
+}
+
+function findGitRoot(start: string): string | undefined {
```
biggamesmallworld: This walks to the filesystem root. A user with a dotfiles repo at `~` and a project at `~/code/myproj` with no `.git` of its own gets `gitRoot = ~`, and the CLI appends to their home `.gitignore`. Reproduced; nothing was created in the project directory.

Worth squaring with the risk posture elsewhere: `--allow-custom-console` gates merely *talking to* a non-default URL, while this writes outside the project with no prompt.

The `pickTarget` suggestion fixes this case. A stop condition here would still be worth having.

packages/cli/src/commands/console/project-link-gitignore.ts
```diff
@@ -0,0 +1,370 @@
...
+    const projectGitignorePath = getProjectLinkGitignorePath(projectRoot);
+    try {
+        const gitRoot = findGitRoot(projectRoot);
+        if (!gitRoot) {
+            return ensureSingleGitignore(projectGitignorePath, 'local');
+        }
+        return ensureGitignoreInRepo(projectRoot, gitRoot);
+    } catch (error) {
+        const reason = error instanceof Error ? error.message : String(error);
+        return { kind: 'failed', path: projectGitignorePath, reason };
```
biggamesmallworld: `writeDirty` can modify several files. If it succeeds on the first and throws on the second, this reports the *project* `.gitignore` as failed when it was not, and never mentions the ancestor it already changed.

Same shape on the success path: `writeDirty` returns one path and `reportProjectLinkGitignore` prints it, so a run that rewrote two files tells the user about one. Confirmed in a repo with `.vendure/` in both the root and `apps/`: both files modified, one reported.

packages/cli/src/commands/console/console.ts
```diff
@@ -0,0 +1,729 @@
...
+            chunks.push(decoder.decode(value, { stream: true }));
+        }
+        chunks.push(decoder.decode());
+        return chunks.join('');
+    } finally {
+        try {
+            reader.releaseLock();
+        } catch {
+            // cancel() already released the lock
+        }
```
biggamesmallworld: `cancel()` does not release the lock, so this comment is not the reason. Nor does `releaseLock()` throw here. Node 22:

```
after cancel():        releaseLock OK (no throw)
with pending read():   releaseLock OK (no throw)
                       pending read REJECTED: TypeError: Invalid state: Releasing reader
```

The catch is unreachable and the comment states something false.

```suggestion
        reader.releaseLock();
```

docs/manifest.json
```diff
@@ -2780,7 +2780,7 @@
                   "title": "MultiChannelStockLocationStrategy",
                   "slug": "multi-channel-stock-location-strategy",
                   "file": "docs/reference/typescript-api/products-stock/multi-channel-stock-location-strategy.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
```
biggamesmallworld, line 2783: `lastModified` rolled **backwards** here (17 Aug -> 13 Aug). Unrelated to this PR; the generator ran against a stale checkout.

```suggestion
                  "lastModified": "2026-08-17T12:00:11+02:00"
```

docs/manifest.json
```diff
@@ -2874,7 +2874,7 @@
                   "title": "RequestContext",
                   "slug": "request-context",
                   "file": "docs/reference/typescript-api/request/request-context.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
```
biggamesmallworld, line 2877: `lastModified` rolled **backwards** here (17 Aug -> 13 Aug). Unrelated to this PR; the generator ran against a stale checkout.

```suggestion
                  "lastModified": "2026-08-17T12:00:11+02:00"
```

docs/manifest.json
```diff
@@ -3096,7 +3096,7 @@
                   "title": "OrderService",
                   "slug": "order-service",
                   "file": "docs/reference/typescript-api/services/order-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T17:32:23+01:00"
```
biggamesmallworld, line 3099: `lastModified` rolled **backwards** here (17 Aug -> 13 Aug). Unrelated to this PR; the generator ran against a stale checkout.

```suggestion
                  "lastModified": "2026-08-17T12:00:11+02:00"
```

docs/manifest.json
```diff
@@ -3186,7 +3186,7 @@
                   "title": "ShippingMethodService",
                   "slug": "shipping-method-service",
                   "file": "docs/reference/typescript-api/services/shipping-method-service.mdx",
-                  "lastModified": "2026-08-17T12:00:11+02:00"
+                  "lastModified": "2026-08-13T14:40:56+02:00"
```
biggamesmallworld, line 3189: `lastModified` rolled **backwards** here (17 Aug -> 13 Aug). Unrelated to this PR; the generator ran against a stale checkout.

```suggestion
                  "lastModified": "2026-08-17T12:00:11+02:00"
```

biggamesmallworld (changes requested):
The protocol half is in good shape: capped response bodies, `redirect: 'error'`, no secret in the URL or output, atomic manifest write, and tests with real assertions.

The `.gitignore` half needs work. I reproduced three of these against real `git` in throwaway repos rather than reading alone.

**Blocking**

1. `pickTarget` targets the outermost `.gitignore`, but git honours the deepest match. Reports success while the manifest stays ignored.
2. The ancestor rewrite converts an unrelated `other-app/.vendure/` into a repo-wide rule, un-ignoring `project.json` for projects the user did not name.
3. `findGitRoot` walks to the filesystem root and can append to the user's home `.gitignore`, with no prompt.
4. Multiple files are written, one path is reported, and the failure path names the wrong file.
5. `docs/manifest.json` carries four unrelated entries with `lastModified` rolled backwards.
6. Prettier fails on four files.

Items 1 and 3 are covered by the suggestion on `pickTarget`; 5 and 6 have suggestions attached. 2 and 4 need a decision from you.

Worth considering: collapsing this module to project-local writes plus one targeted ancestor fix. That is roughly 100 lines instead of 370, and it removes the class of bug rather than the instances.

**Minor**

- `PROJECT_LINK_MANIFEST_RELATIVE_PATH` exists but nothing outside its own file uses it; `.vendure/project.json` is spelled out independently in the gitignore module, `console.ts`, the create template, and the docs.
- `uuid` and `nonEmptyString` are declared three times: exported from `project-link-validation.ts`, then re-declared privately in `project-link-manifest.ts` and `console.ts`.
- Two `.git` walks: `findGitRoot` and the loop in `findManifestSearchBoundary`.
- `CommandInterruptedError` takes an `exitCode` that is `130` at all six call sites. The 143 comes from `interruptedExitCode`, not the field.
- The docs do not mention that `console link` may modify `.gitignore` files outside the project directory. That is the one thing a user needs to know before running it.

**Tests**

Good coverage on the protocol side. The gaps are all in the part that is broken: no test for gitignore precedence, none that the ancestor rewrite leaves unrelated rules alone, and several assert only `.kind` without checking which files were touched. `project-link-gitignore.spec.ts:199` does check; that pattern applied throughout would have caught 1 and 2.

dlhck:
Thanks for the detailed Git reproductions. Fixed in `b8edea25f`. The command now writes one `.gitignore` per run, defaults to the project-local file, and adds only a project-specific ancestor exception when a monorepo directory rule blocks Git traversal. The new tests run `git check-ignore` for precedence, unrelated-project isolation, ancestor rules, and incidental outer repositories. I also restored the four manifest timestamps, fixed formatting, removed the unreachable stream-lock catch, reused the manifest path and validators, consolidated the `.git` walk, simplified interrupt errors, and documented the ancestor-file exception. Validation: CLI build passed; all 30 test files passed with 358 tests passed and 1 skipped; Prettier check passed for the four reviewed TypeScript files.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f. The default target is now the project-local .gitignore, and a real Git regression covers an intermediate .gitignore overriding the repository root.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f. Ancestor directory rules stay intact. The command adds exceptions for only the selected project, and the regression confirms other-app remains ignored.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f. Prettier now passes on all four reviewed TypeScript files.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f as part of the formatting pass.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f. Ancestor writes are limited to detected monorepo boundaries. An incidental outer repository now gets no ancestor changes, with a regression test for that case.

packages/cli/src/commands/console/project-link-gitignore.ts
dlhck: Fixed in b8edea25f. Each run now writes at most one .gitignore, so the reported success or failure path is the exact file touched and partial multi-file writes are gone.

packages/cli/src/commands/console/console.ts
dlhck: Fixed in b8edea25f. The unreachable catch and incorrect comment are removed; releaseLock is called directly.

dlhck: Fixed in b8edea25f. The unreachable catch and incorrect comment are removed; releaseLock is called directly.

docs/manifest.json
dlhck: Restored to 2026-08-17T12:00:11+02:00 in b8edea25f.

docs/manifest.json
dlhck: Restored to 2026-08-17T12:00:11+02:00 in b8edea25f.

docs/manifest.json
dlhck: Restored to 2026-08-17T12:00:11+02:00 in b8edea25f.

docs/manifest.json
dlhck: Restored to 2026-08-17T12:00:11+02:00 in b8edea25f.

dlhck:
Follow-up on the generated manifest: the docs bot regenerated those four values after the first push. I merged the latest `minor` in `5c11d0cd9`, so the PR now has the current merge base and the four unrelated timestamp changes are gone. The manifest patch now contains only the CLI page timestamp.

michaelbromley (dismissed):
**Must change**

`packages/cli/src/commands/console/project-link-gitignore.ts` implements far more than the requirement needs. The requirement is small: existing scaffolded projects ignore `.vendure/`, and that line must become `.vendure/*` plus `!.vendure/project.json` so the manifest can be committed. The file instead implements a general gitignore-pattern detector and rewriter. It walks the ancestor chain up to the git root, classifies every non-comment line in every file along that chain via regex, decides which single file covers the project, and rewrites whatever pattern it finds.

Two problems with that approach:

- gitignore resolution is order-sensitive. The code appends `!.vendure/project.json` without checking whether another pattern, for example a catch-all `*.json` in a nested `.gitignore`, still ignores `project.json`.
- The 280-line spec pins the parser's current behaviour. Tests of a hand-written parser lock in what it does, not whether that matches git's actual gitignore semantics.

Please scope it down: when a file contains the known `.vendure/` (or `.vendure`) line, replace it with the two-line block; when no file contains it, append the block. If the general ancestor-chain rewriting is needed for a concrete monorepo case, add a comment which explains why a hand-written regex parser was chosen over an existing gitignore-matching library.

**Should change**

- `console.ts` and `project-link-manifest.ts` each define private `uuid()` and `nonEmptyString()` functions which wrap the shared `uuidValue`/`nonEmptyStringValue` validators and differ only in the error message. One shared helper which takes the message would remove the duplication.
- `consoleCommand()` records interruption in two places: the closured `interruptedExitCode` set by the SIGINT/SIGTERM handlers, and the `exitCode` on a thrown `CommandInterruptedError`. The catch block reconciles them with `interruptedExitCode ?? error.exitCode`. A comment which explains why both channels exist and which one wins would save the next reader real time.

**Worth a second look (not blocking)**

- `link`, `status` and `unlink` all resolve the project root through `resolveProjectRoot`, so a read-only `status` call fails when an ancestor directory has a conflicting manifest. Surfacing the conflict from `status` may be deliberate; please confirm the coupling is intended.
- In `link()`, the "replace existing manifest?" confirmation runs before `resolveConsoleEndpoints` validates the environment, so a user can confirm and then hit a config error. Nothing is written until approval completes, so swapping the order is a polish fix only.
- The `docs/manifest.json` changes move several `lastModified` timestamps backwards (for example `RequestContext` and `OrderService` from 2026-08-17 to 2026-08-13). This looks like the regeneration script picking up stale git blame during a rebase; worth checking separately from this PR.

**What is solid**

The security handling is careful: `redirect: 'error'` on every fetch, the verification URL checked against the polling secret so the secret never reaches a printed link, capped response bodies, temp-file-then-rename writes with cleanup on failure, and exact-field manifest validation. The tests match that standard: exact assertions, coverage of expired requests, 429/503 backoff sequences, SIGINT mid-poll and cross-root conflicts, and several tests assert the polling secret never appears in printed output.

dlhck:
@michaelbromley Addressed in a3319acd1.

- Replaced the gitignore parser and ancestor traversal with a project-local update. It replaces only the exact scaffolded `.vendure/` or `.vendure` line. For non-scaffolded projects, it appends the two required rules or creates the project `.gitignore`.
- Added a real Git regression for a non-scaffolded project. The manifest is committable and other `.vendure` files stay ignored.
- Kept the shared validation helpers already present on the branch.
- Documented interruption precedence. SIGINT and SIGTERM retain their process exit codes. Prompt cancellation and external aborts use 130.
- Moved endpoint validation before the replacement confirmation.
- Kept the cross-root manifest check for `status` deliberately. It prevents the command from reporting one project while a conflicting ancestor manifest is present.
- The later docs regeneration commits restore the unrelated manifest timestamps.

Verified with the CLI build, the CLI suite with 352 passing and 1 skipped test, and all 4 Console command end-to-end tests. Could you take another look?

biggamesmallworld (approved):
Fixes verified against the new head. I re-ran the same throwaway-repo harness from the previous review:

```
precedence no-op   -> writes apps/vendure/.gitignore; manifest committable (exit 1), cache.db still ignored (exit 0)
unrelated clobber  -> other-app/.vendure/project.json untouched (exit 0)
$HOME write       -> project directory only
multi-file report  -> single file written, single path reported
```

370 lines down to 76, project-local only, and it reads better for it. Also confirmed: `docs/manifest.json` no longer carries the backwards entries, the unreachable `try`/`catch` around `releaseLock` is gone, `PROJECT_LINK_MANIFEST_RELATIVE_PATH` is now the single source for the path, `uuid`/`nonEmptyString` are down to one copy, and there is one `.git` walk instead of two.

Two things I am not holding the PR for.

**Prettier still fails on two files**, `console.spec.ts:152` and `project-link-manifest.ts:287`, both wrap/indent. I ran 3.3.3 without `prettier-plugin-organize-imports` and the repo pins `^3.2.5`, so worth a quick confirm with the repo's own binary before you trust my count.

**One case the simplification gives up**, noted so it is on record rather than as a request. A directory-form `.vendure/` in an *ancestor* `.gitignore` defeats the project-local `!.vendure/project.json`, because git will not descend into an excluded directory to re-include a file. The rules get written and the run reports success:

```
root/.gitignore: .vendure/
reported: created apps/vendure/.gitignore
manifest committable: false
```

The old ancestor-rewriting code handled this, so it is a regression, but one that came from my own suggestion and is the right trade: it bought fixing three bugs that fire on ordinary setups, against one that needs a hand-placed ancestor rule. Nothing in the template or CLI produces that rule, so I do not think it is worth pre-emptive code. The one part that is wrong regardless of how the rule got there is the message at `console.ts:319` claiming the manifest can be committed. If you ever want to close it, `git check-ignore -q` on the manifest after writing would let git resolve it rather than re-deriving the semantics.

I have not read the `/v1` endpoint commit, which landed after my last pass. CI is green.

biggamesmallworld (approved):

PR #5164, opened 2026-08-21
fix(core): prevent orphan-nullification of surcharges on Order save

thomslbua:
## Description

`Surcharge.order` is a nullable `ManyToOne` and `Order.surcharges` its `OneToMany`, with no `orphanedRowAction` set. TypeORM defaults that to `nullify`, so saving an `Order` diffs the passed `surcharges` array against the DB and **orphan-nullifies** (`UPDATE surcharge SET "orderId" = NULL`) every row whose id is missing from the in-memory array — silently: no error, no event, no log.

The in-memory array goes stale easily: core mutations (`addItemToOrder`, `adjustOrderLine`, `applyCouponCode`, `setShippingAddress`, …) snapshot the order at entry and publish `OrderLineEvent` *before* the save runs — any **blocking** event handler that adds/removes/re-adds a surcharge in that window (or a concurrent request doing so) gets its surcharge row detached by the subsequent save.

We hit this in production: negative discount surcharges (loyalty points / gift vouchers, maintained by a blocking `OrderLineEvent` handler that re-clamps them on repricing) vanished from orders while their side effects remained — customers paid full price although a discount had been applied, with zero trace in any log.

### Reproduction sketch

1. Register a blocking `OrderLineEvent` handler that removes + re-adds a surcharge on the order (the re-added row gets a new id).
2. Call `adjustOrderLine` on an order carrying a surcharge.
3. The surcharge row ends up with `orderId = NULL` — detached by the save in `applyPriceAdjustments`, which still held the pre-event snapshot. A cross-request race between any cart mutation and `addSurchargeToOrder` reproduces it without a blocking handler.

### Fix

The PR originally omitted `surcharges` from the save in `applyPriceAdjustments`, which closes the path the bug was reported on. But the same hazard is present in every other full-`Order` save, and `getOrderOrThrow` loads `surcharges` by default:

- `OrderService.removeItemsFromOrder`
- `OrderService.addCustomerToOrder`
- `PromotionService.addPromotionsToOrder`
- `OrderModifier.cancelOrder`
- plugin code doing `getRepository(ctx, Order).save(order)`, which is the pattern documented on `ActiveOrderStrategy`

So the fix belongs on the relation, not at each call site:

```ts
@ManyToOne(type => Order, order => order.surcharges, {
    onDelete: 'CASCADE',
    orphanedRowAction: 'disable',
})
order: Order;
```

Two things worth knowing about that option:

- TypeORM 0.3 reads it from the inverse (`ManyToOne`) side of a `OneToMany`, not from the `OneToMany` itself (`OneToManySubjectBuilder` checks `relation.inverseRelation.orphanedRowAction`), so setting it on `Order.surcharges` has no effect.
- It is applied when TypeORM builds the persistence subjects. No schema change, no migration.

Nothing relies on the orphan behaviour: `removeSurchargeFromOrder` deletes the row rather than detaching it, and `OrderModifier.modifyOrder` saves each `Surcharge` with its `order` already set. The bind path in `OneToManySubjectBuilder` runs before the orphan check, so newly added surcharges are still attached as before.

The `surcharges` omission in `applyPriceAdjustments` is kept, now purely for the same reason `lines` and `shippingLines` are omitted: the save has no need to diff a relation it never writes.

The guarantee is documented on `Order.surcharges`, so it reaches the generated API reference where a plugin author will look.

### Test

`packages/core/e2e/order.e2e-spec.ts` adds a Surcharge, loads the Order, then adds a second Surcharge through `OrderService` so the loaded handle really is stale, saves that handle, and asserts both rows are still attached. It goes through a plain `Order` save rather than `applyPriceAdjustments`, so it pins the relation guarantee for all call sites. Without the fix it reports `expected [ 200 ] to deeply equal [ -500, 200 ]`: the surcharge added since the load is the one that disappears, which is the production symptom. The test also covers the removal path, which must keep deleting the row.

### Not addressed here

`OrderLine.order`, `ShippingLine.order`, `Payment.order` and `Surcharge.orderModification` still default to `nullify`, so the same hazard exists for them. They are not symmetric with `Surcharge.order` and do not belong in this fix:

- `lines` **cannot** take `orphanedRowAction: 'disable'` as things stand. `OrderService.removeItemsFromOrder` filters `order.lines` and saves the Order specifically to make the removal stick before repricing, so it depends on the nullify.
- `shippingLines` and `payments` are never detached through the array in core (shipping lines are added and removed with explicit relation queries), so they look safe, but each needs its own check and regression test.

Repricing also calculates `Order.subTotal` and `Order.subTotalWithTax` from the same snapshot array, so a persisted total can disagree with the surcharge rows even though the rows now survive. Filed separately as #5332.

## Breaking changes

No API change and no migration. Two behaviour changes worth naming:

- Code which detached a `Surcharge` by removing it from `order.surcharges` and saving the Order no longer detaches it. Use `OrderService.removeSurchargeFromOrder`, which deletes the row.
- Because `applyPriceAdjustments` omits the relation, that save no longer binds a `Surcharge` which was persisted *without* its `order` property set. Core is unaffected — both writers save the row with `order` first — but a plugin which pushed an unsaved `Surcharge` onto `order.surcharges` and relied on repricing to attach it will now lose it silently. Set `order` on the `Surcharge`, or use `addSurchargeToOrder`.

There is no issue number for this: the bug was reported directly in this PR.

### DIFF
--- a/docs/docs/reference/typescript-api/entities/order.mdx
+++ b/docs/docs/reference/typescript-api/entities/order.mdx
@@ -160,6 +160,11 @@ Surcharges are arbitrary modifications to the Order total which are neither
 ProductVariants nor discounts resulting from applied Promotions. For example,
 one-off discounts based on customer interaction, or surcharges based on payment
 methods.
+
+Surcharges are attached and removed with `OrderService.addSurchargeToOrder` and
+`OrderService.removeSurchargeFromOrder`. Saving an Order does not detach a
+Surcharge which is missing from this array, and repricing the Order does not attach one
+which was saved without its `order` property set.
 ### couponCodes
 
 <MemberInfo kind="property" type={`string[]`}   />
--- a/docs/docs/reference/typescript-api/entities/surcharge.mdx
+++ b/docs/docs/reference/typescript-api/entities/surcharge.mdx
@@ -21,7 +21,10 @@ class Surcharge extends VendureEntity {
     @Column('simple-json')
     taxLines: TaxLine[];
     @Index()
-    @ManyToOne(type => Order, order => order.surcharges, { onDelete: 'CASCADE' })
+    @ManyToOne(type => Order, order => order.surcharges, {
+        onDelete: 'CASCADE',
+        orphanedRowAction: 'disable',
+    })
     order: Order;
     @Index()
     @ManyToOne(type => OrderModification, orderModification => orderModification.surcharges)
--- a/packages/core/e2e/order.e2e-spec.ts
+++ b/packages/core/e2e/order.e2e-spec.ts
@@ -12,8 +12,10 @@ import {
 import { omit } from '@vendure/common/lib/omit';
 import { pick } from '@vendure/common/lib/pick';
 import {
+    Order,
     OrderService,
     RequestContextService,
+    TransactionalConnection,
     defaultShippingCalculator,
     defaultShippingEligibilityChecker,
     manualFulfillmentHandler,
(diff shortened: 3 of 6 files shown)


thomslbua:

thomslbua:
recheck

biggamesmallworld (dismissed):
The diagnosis is right and this is the correct fix. Nothing in the repricing path writes surcharge columns: both tax calculation strategies only read `order.surcharges`, `Order.surcharges` has no `cascade`, and surcharge rows are written only by `addSurchargeToOrder`, `removeSurchargeFromOrder` and `OrderModifier.modifyOrder`.

Four things before it can merge.

**1. The comment above the omit list describes the wrong reason.**

`lines` and `shippingLines` were omitted for performance, and the comment says so: "The other omissions prevent the save function from doing more work than necessary." `surcharges` is omitted for correctness. A reader who takes the comment at its word will delete the entry as a performance tweak, or answer a "surcharge not persisting" report by removing it.

State the mechanism and the invariant in the comment: TypeORM nullifies the `orderId` of any persisted surcharge row missing from this array, the array is a snapshot taken before blocking event handlers run, and surcharge rows are written only by `addSurchargeToOrder`, `removeSurchargeFromOrder` and `OrderModifier.modifyOrder`.

**2. Add the regression test.**

The blocking handler is not needed to pin the invariant. Load an order with a persisted surcharge, set `order.surcharges = []` to stand in for the stale snapshot, call `applyPriceAdjustments`, then assert the row still has its `orderId`. A second test driving a blocking `OrderLineEvent` handler is welcome on top of that, but this one is what stops a later refactor undoing the fix.

**3. The same bug is live in `order-modifier.ts:539-541`.**

```ts
if (input.surcharges?.length) {
    await this.connection.getRepository(ctx, Order).save(order, { reload: false });
}
```

Each `Surcharge` is constructed with `order` set and saved at line 520, so the child save already writes `orderId`. `order.surcharges` is loaded by then. The only thing this parent save contributes is the orphan-nullification you are fixing, behind an `if` which fires precisely when surcharges are in play. Omit `surcharges` from it, or drop the save. If only one call site is fixed, the next occurrence looks identical and a reader grepping for the fix finds it already applied.

**4. "Breaking changes: None" is wrong.**

A plugin can persist a `Surcharge` without setting `order`, push it onto `order.surcharges`, and rely on this save to bind `orderId`. That plugin will now silently stop attaching surcharges. It is the same trap `lines` and `shippingLines` already set, and it is unlikely, but name it in the PR body and the changelog. Add the invariant to the `addSurchargeToOrder` docblock as well, since that is where a plugin author looks.

The test matrix has not run on this PR, only CLA, CodeRabbit and Vercel. I will approve the workflows so the unit and e2e suites report.

biggamesmallworld (approved):
Approving. Note for the record that I wrote the current commit, so this is not an independent review of that code: the diagnosis and the original report are the contributor's, the fix as it now stands is mine.

What changed since my review of 4 Sept:

- The fix moved from the `applyPriceAdjustments` omit list to the relation itself. `Surcharge.order` now sets `orphanedRowAction: 'disable'`. Omitting the relation at one save site left the same hazard in `removeItemsFromOrder`, `addCustomerToOrder`, `PromotionService.addPromotionsToOrder`, `OrderModifier.cancelOrder` and in plugin code, because `getOrderOrThrow` loads `surcharges` by default.
- TypeORM 0.3 reads `orphanedRowAction` from the inverse side of a `OneToMany`, so it has to go on `Surcharge.order`. On `Order.surcharges` it does nothing. Verified against `OneToManySubjectBuilder`.
- Regression test in `order.e2e-spec.ts` goes through a plain `Order` save rather than through repricing, so it pins the guarantee for every call site. With the fix removed it reports `expected [ 200 ] to deeply equal [ -500, 200 ]`: only the surcharge added after the snapshot was taken disappears, which is the reported symptom.
- The plugin-facing guarantee is on `Order.surcharges`, so it reaches the generated reference.
- `OrderLine.order`, `ShippingLine.order`, `Payment.order` and `Surcharge.orderModification` are untouched and still nullify. `lines` cannot take the same option: `removeItemsFromOrder` saves the Order specifically to make a line removal stick before repricing, so it depends on the nullify. The other two are listed in the PR body for a separate pass.

Still open, not a blocker for this PR: repricing computes totals from the same snapshot array, so a total can be persisted that does not match the surcharge rows. The rows survive now, the total can still be stale. Needs its own fix.

Test matrix on this head: mysql, mariadb and postgres e2e pass, unit tests pass on 20.x, 22.x and 24.x, build and codegen pass. The sqljs job first failed on `EEXIST: mkdir .../e2e/__data__` inside `SqljsInitializer.populate`, which is a harness race at server init, and is rerunning. `deploy` fails because fork PRs get no `VERCEL_TOKEN`.

PR #5168, opened 2026-08-23
fix(core): free job queue slot when status update fails

### LINKED ISSUE #5167: A failed job_record update permanently wedges a JobQueue with concurrency 1
**Describe the bug**

`ActiveQueue.onFailOrComplete` (`packages/core/src/job-queue/polling-job-queue-strategy.ts`) awaits
`this.jobQueueStrategy.update(job)` before calling `this.removeJobFromActive(job)`:

```js
async onFailOrComplete(job) {
    await this.jobQueueStrategy.update(job);   // <- throws
    this.removeJobFromActive(job);             // <- never runs
}
```

`onFailOrComplete` itself is only ever invoked from inside a `.finally()` whose own rejection is
swallowed by a `.catch()` that just logs:

```js
.finally(() => {
    nextJob.off('progress', onProgress);
    cancellationSub.unsubscribe();
    return this.onFailOrComplete(nextJob);
})
.catch((err) => {
    Logger.warn(`Error updating job info: ${JSON.stringify(err)}`);
});
```

If `jobQueueStrategy.update(job)` throws — e.g. `SqlJobQueueStrategy`'s settling `UPDATE`
times out because the database is briefly congested — the job is never spliced out of
`activeJobs`. The polling loop that starts new jobs is:

```js
const runningJobsCount = this.activeJobs.length;
for (let i = runningJobsCount; i < this.concurrency; i++) { ... }
```

With `concurrency: 1` (`DefaultJobQueuePlugin.init({ concurrency: () => 1 })`, or any queue
configured that way), `activeJobs.length` stays at 1 forever once this happens, so the loop body
never runs again — the queue is silently dead until the process restarts. No error surfaces
beyond the one throttled `Error updating job info` warning; nothing else indicates the queue has
stopped.

**To Reproduce**

1. Configure a queue at `concurrency: 1` (e.g. via `DefaultJobQueuePlugin.init({ concurrency: () =>
   1 })`).
2. Make the settling `UPDATE` fail once for a job in that queue — either a paused/congested
   database (e.g. set `query_timeout` low in `dbConnectionOptions.extra` and stall a query), or by
   stubbing `SqlJobQueueStrategy.prototype.update` to reject on its first call for a given job.
3. Let the job resolve or reject normally.
4. Observe that no further job in that queue is ever processed, even though new jobs continue to
   be added to it — until the worker process is restarted.

**Expected behavior**

`removeJobFromActive` should run regardless of whether persisting the job's terminal state
succeeded. A failed status write must not permanently consume a concurrency slot.

**Actual behavior**

The concurrency slot is never released. The `JobQueue` for that name stops processing jobs
entirely, with no further error or log line beyond the one throttled warning at the moment of
failure.

**Screenshots/Videos**

N/A — this is a background job-queue processing bug, not a UI issue.

**Error logs**

Production worker log sequence (`update-search-index` queue, `concurrency: 1`) at the moment the
queue wedged:

```
23:32  Job 60747 [update-search-index] failed and will not retry.
23:33  Error updating job info: {"query":"UPDATE \"public\".\"job_record\" SET … WHERE \"id\" = $13
       AND \"settledAt\" IS NULL","parameters":[60747,"update-search-index",…,"FAILED",0,null,
       "Query read timeout","2026-08-22T23:31:37.620Z","2026-08-22T23:32:37.801Z",1,0,1,60747],…}
```

Database evidence at the time this was investigated: last `COMPLETED` row in the queue at
`23:31:24`; one `RUNNING` row (job 60747) that never settled; 58 `PENDING` rows queued up behind
it; plus 7 older `RUNNING` rows from 2026-06-28/06-29 that are earlier, unnoticed occurrences of
the same failure mode (the queue silently died then too, and only "recovered" because the worker
happened to be redeployed for an unrelated reason).

- Nodejs version: 24 (node:24-alpine); engines requires >=22
- Database (mysql/postgres etc): PostgreSQL 17
- Operating System (Windows/macOS/Linux): Linux (Alpine, Docker) in production; Windows 11 for development
- Browser (if applicable): N/A — worker process, no browser involved
- Package manager (npm/yarn/pnpm): npm (npm workspaces monorepo)

**Configuration**

```ts
dbConnectionOptions: {
    type: 'postgres',
    extra: {
        max: 10,
        connectionTimeoutMillis: 10000,
        idleTimeoutMillis: 30000,
        statement_timeout: 60000,
        idle_in_transaction_session_timeout: 30000,
        query_timeout: 60000,
    },
},
// ...
DefaultJobQueuePlugin.init({
    useDatabaseForBuffer: true,
    concurrency: queueName => /* every queue in this app resolves to */ 1,
}),
```

**Minimal reproduction**

Not attached as a runnable repo — reproducible directly from the excerpted source with the steps
above: any `JobQueueStrategy.update()` implementation that can throw (the shipped
`SqlJobQueueStrategy` can, on any transient DB error) combined with `concurrency: 1` reproduces
this deterministically.

**Workaround**

Patch `SqlJobQueueStrategy.prototype.update` at runtime so it never rejects — retry the write
once, and if that also fails, log and resolve anyway. `onFailOrComplete` awaits this call, so a
swallowed error still lets `removeJobFromActive` run and the concurrency slot gets released. Trade-
off: the `job_record` row can be left stale (its terminal state never persisted); a periodic
worker-boot recovery step settles any row still in a non-terminal state. See
`apps/shop-api/src/config/patch-job-queue-update.ts` and `apps/shop-api/src/config/job-record-recovery.ts`
in this repository for the exact implementation.

**Suggested fix**

```js
async onFailOrComplete(job) {
    try {
        await this.jobQueueStrategy.update(job);
    } finally {
        this.removeJobFromActive(job);
    }
}
```

**Additional context**

This was found while investigating a production incident where the `update-search-index` queue
stopped processing entirely for over 30 minutes with no operator-visible error. The root trigger
was unrelated DB congestion (a scheduled task overlapping itself — see the companion issues
"Scheduled task timeout releases the lock but leaves the task running, allowing unbounded
overlapping executions" and "preventOverlap cannot protect a task that runs longer than its own
schedule interval"), but this bug is what turned a transient few-second slowdown into a permanent
outage requiring a restart.

wakqasahmed:
Fixes #5167.

`ActiveQueue.onFailOrComplete()` in `polling-job-queue-strategy.ts` awaits `jobQueueStrategy.update(job)` and only calls `removeJobFromActive(job)` on the next line. If that `update()` call throws — which `SqlJobQueueStrategy` can do on any transient DB error — the job is never spliced out of `activeJobs`. With `concurrency: 1` the polling loop's `for (let i = runningJobsCount; i < this.concurrency; i++)` then never runs again, so the queue is permanently wedged until the process restarts.

The fix wraps the `update()` call in try/finally so `removeJobFromActive()` always runs regardless of whether the update succeeds. The `.finally()`/`.catch()` chain that already surrounds the call to `onFailOrComplete()` still logs the warning when the update throws, so no error visibility is lost — the only change is that the concurrency slot is no longer permanently consumed by a job whose terminal state failed to persist.

I added a test in `polling-job-queue-strategy.spec.ts` that uses `InMemoryJobQueueStrategy` with `concurrency: 1`, mocks `update()` to throw on the settling call for the first job, and asserts a second queued job still gets processed afterward. Without the fix this test hangs/fails because the second job never gets picked up.

I traced the fix through the exact code path by hand against the current source and I'm confident in the change itself, but I want to flag that I wasn't able to actually execute `npm run test` for this PR — the machine I'm working from ran out of disk space partway through `npm install` for this monorepo, and I didn't want to hold the fix back for a fresh issue on that account. Happy to re-run and confirm once I have a clean environment, or if a maintainer/CI gets to it first that works too.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5168"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790080199&installation_model_id=20193&pr_number=5168&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5168&signature=ebae52371169686318062e80a9695bb3b60a05a528ec7eba849b4b696ae68e4b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
+++ b/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
@@ -1,0 +1,54 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { InMemoryJobQueueStrategy } from './in-memory-job-queue-strategy';
+import { Job } from './job';
+
+describe('PollingJobQueueStrategy', () => {
+    let strategy: InMemoryJobQueueStrategy;
+
+    beforeEach(() => {
+        strategy = new InMemoryJobQueueStrategy({ concurrency: 1, pollInterval: 10 });
+        strategy.init({
+            get() {
+                return { isWorker: false };
+            },
+        } as any);
+    });
+
+    afterEach(() => {
+        strategy.destroy();
+    });
+
+    it('releases the concurrency slot even when the settling update() throws', async () => {
+        const originalUpdate = strategy.update.bind(strategy);
+        let updateCallCount = 0;
+        vi.spyOn(strategy, 'update').mockImplementation(async (job: Job) => {
+            updateCallCount++;
+            if (updateCallCount === 2) {
+                // Simulate the settling update for the first job failing,
+                // e.g. a transient DB error.
+                throw new Error('simulated update failure');
+            }
+            return originalUpdate(job);
+        });
+
+        await strategy.add(new Job({ id: 'job-1', queueName: 'test', data: {} }));
+        await strategy.add(new Job({ id: 'job-2', queueName: 'test', data: {} }));
+
+        const processed: string[] = [];
+        const process = async (job: Job) => {
+            processed.push(job.id as string);
+            return true;
+        };
+        await strategy.start('test', process);
+
+        await vi.waitFor(
+            () => {
+                expect(processed).toEqual(['job-1', 'job-2']);
+            },
+            { timeout: 2000, interval: 20 },
+        );
+
+        await strategy.stop('test', process);
+    });
+});
--- a/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
+++ b/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
@@ -15,16 +15,25 @@ describe('PollingJobQueueStrategy', () => {
         } as any);
     });
 
-    afterEach(() => {
+    let activeProcess: ((job: Job) => Promise<any>) | undefined;
+
+    afterEach(async () => {
+        // strategy.destroy() does not touch the ActiveQueue timer, so without
+        // an explicit stop() the polling loop keeps running against a
+        // torn-down mock in the next test. Calling stop() here — rather than
+        // at the end of each test body — means it still runs even when the
+        // test fails on an assertion or a waitFor timeout above it.
+        if (activeProcess) {
+            await strategy.stop('test', activeProcess);
+            activeProcess = undefined;
+        }
         strategy.destroy();
     });
 
     it('releases the concurrency slot even when the settling update() throws', async () => {
         const originalUpdate = strategy.update.bind(strategy);
-        let updateCallCount = 0;
         vi.spyOn(strategy, 'update').mockImplementation(async (job: Job) => {
-            updateCallCount++;
-            if (updateCallCount === 2) {
+            if (job.id === 'job-1' && job.isSettled) {
                 // Simulate the settling update for the first job failing,
                 // e.g. a transient DB error.
                 throw new Error('simulated update failure');
@@ -40,6 +49,7 @@ describe('PollingJobQueueStrategy', () => {
             processed.push(job.id as string);
             return true;
         };
+        activeProcess = process;
         await strategy.start('test', process);
 
         await vi.waitFor(
@@ -48,7 +58,37 @@ describe('PollingJobQueueStrategy', () => {
             },
             { timeout: 2000, interval: 20 },
         );
+    });
 
-        await strategy.stop('test', process);
+    it('releases the concurrency slot even when the initial (pre-process) update() throws', async () => {
+        const originalUpdate = strategy.update.bind(strategy);
+        let updateCallCount = 0;
+        vi.spyOn(strategy, 'update').mockImplementation(async (job: Job) => {
+            updateCallCount++;
+            if (updateCallCount === 1) {
+                // Simulate the initial "mark as running" update for the first
+                // job failing before process() ever runs.
+                throw new Error('simulated update failure');
+            }
+            return originalUpdate(job);
+        });
+
+        await strategy.add(new Job({ id: 'job-1', queueName: 'test', data: {} }));
+        await strategy.add(new Job({ id: 'job-2', queueName: 'test', data: {} }));
+
+        const processed: string[] = [];
+        const process = async (job: Job) => {
+            processed.push(job.id as string);
+            return true;
+        };
+        activeProcess = process;
+        await strategy.start('test', process);
+
+        await vi.waitFor(
+            () => {
+                expect(processed).toEqual(['job-2']);
+            },
+            { timeout: 2000, interval: 20 },
+        );
     });
 });
--- a/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
+++ b/packages/core/src/job-queue/polling-job-queue-strategy.spec.ts
@@ -91,4 +91,45 @@ describe('PollingJobQueueStrategy', () => {
             { timeout: 2000, interval: 20 },
         );
     });
+
+    it('keeps an in-flight initial update visible to shutdown, so stop() waits for it', async () => {
+        let releaseUpdate: (() => void) | undefined;
+        const updateGate = new Promise<void>(resolve => {
+            releaseUpdate = resolve;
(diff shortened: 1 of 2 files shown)


biggamesmallworld (changes requested):
Diagnosis is right and `try/finally` is the correct minimal change. On your open question: I ran the test both ways. Passes with the fix, fails without (`["job-1"]` vs `["job-1", "job-2"]`). It discriminates correctly, so that is not a reason to hold the PR.

Four things before merge.

**1. Same bug twenty lines earlier, unguarded.** `polling-job-queue-strategy.ts:123-124`:

```ts
this.activeJobs.push(nextJob);
await this.jobQueueStrategy.update(nextJob);
```

If this throws, it unwinds to the `catch` at line 168. `process()` never runs, so the `.finally()`, `onFailOrComplete()` and `removeJobFromActive()` never run either. Same wedge. I confirmed it on your branch with your fix applied, throwing on this call instead of the settling one: `concurrency: 1`, `pollInterval: 10`, after 800ms `processed` was `[]`.

Line 124 is the first write of the job's lifecycle, under exactly the connection pressure that causes the transient error you describe, so it is the likelier trigger of the two. Push to `activeJobs` only after the update succeeds, or reuse the shape from `onFailOrComplete()`. Needs a test.

**2. `JSON.stringify(err)` renders `{}`** (line 164). `Error` has no enumerable own properties, so the operator sees `Error updating job info: {}`. Pre-existing, but this PR makes that line the only signal a terminal state was dropped. Log `err.message` and `err.stack`.

**3. Jobs are left in RUNNING with no recovery path.** `getNextAndSetAsRunning()` (`sql-job-queue-strategy.ts:127-129`) selects only PENDING and RETRYING, and `update()` carries `.andWhere('settledAt IS NULL')`. Your fix also drops the job from `activeJobs`, so shutdown's `defer()` misses it. There is no reaper. Right trade, but note it in the description and open a follow-up.

**4. `strategy.stop()` is unreachable when the test fails** (it sits after the `waitFor` that throws). `afterEach` is not a backstop: `PollingJobQueueStrategy` has no `destroy()` override, and neither `InjectableJobQueueStrategy.destroy()` nor `InMemoryJobQueueStrategy.destroy()` touches the `ActiveQueue` timer. The polling loop keeps running against a mocked `update()`. Move it to `afterEach`.

Minor: `updateCallCount === 2` encodes call ordering. `job.id === 'job-1' && job.isSettled` says what you mean and survives an update call being added upstream (`job.isSettled` is at `job.ts:78`). Commit scope should be `fix(core):`.

wakqasahmed:
Pushed fixes for all 4 points, and filed the no-reaper gap as a tracked follow-up: #5187 (per point 3, not fixing that here — this PR's trade-off is still right, just doesn't include recovery).

wakqasahmed:
Pushed a follow-up fix for the shutdown-ordering gap CodeRabbit's second pass flagged: delaying the `activeJobs.push()` until after the initial `update()` resolves meant the job was invisible to `awaitRunningJobsOrTimeout()` during that await, so a shutdown racing it could resolve before the job ever ran.

Now the job is pushed to `activeJobs` first, and removed again in a `catch` if the initial update rejects — so it stays visible during shutdown, and the concurrency slot still isn't wedged on failure. Added a test that delays the initial update, calls `stop()` while it's in flight, and asserts `stop()` waits for the job to actually process rather than resolving early.

Same disclosure as before: traced by hand against the current source, but I still don't have a clean environment to run `npm run test` in (`npm install` on this monorepo doesn't complete in the sandbox I'm working from). Flagging again rather than holding the fix back — happy to have CI or a maintainer confirm.

biggamesmallworld (approved):
Thanks @wakqasahmed 🙏  nice work!

PR #5169, opened 2026-08-23
fix(dashboard): populate extension registry before signalling extensions loaded

### LINKED ISSUE #5159: Dashboard: Order Detail Query cannot be extended
**Describe the bug**
It's not possible to extend the detail query for Orders via `extendDetailDocument`

**To Reproduce**

1. Checkout and start https://github.com/LucidityDesign/vendure-extend-order-query
2. Go to Customer Detail Page and check the GraphQL query: user roles id and permission is loaded. Detail query is extended ✅ 
3. Go to Order List Page: channels defaultLanguageCode is loaded. List Query is extended ✅ 
4. Go to Order Detail Page: channels defaultLanguageCode is not loaded. Detail Query is not extended ❌ 

**Expected behavior**
Order Detail query should be extended in `MyNewFeaturePlugin`

**Actual behavior**
Order Detail query extension is ignored.

**Screenshots/Videos**

**Error logs**

```
Paste error logs here
```

- Nodejs version: v24.13.0
- Browser (if applicable): Edge

**Configuration**

See https://github.com/LucidityDesign/vendure-extend-order-query/tree/main/src/plugins/my-new-feature

**Minimal reproduction**
https://github.com/LucidityDesign/vendure-extend-order-query

**Workaround**
Solution from claude (take it with a grain of salt because the other extensions work):

```diff
diff --git a/src/app/main.tsx b/src/app/main.tsx
index 6bbeaad346f8b5796898819857a89a9f4b3cf195..b696180912593ee3ccb93afef34cbdd6134583ff 100644
--- a/src/app/main.tsx
+++ b/src/app/main.tsx
@@ -12,7 +12,7 @@ import { useAuth } from '@/vdb/hooks/use-auth.js';
 import { useServerConfig } from '@/vdb/hooks/use-server-config.js';
 import { defaultLocale, dynamicActivate } from '@/vdb/providers/i18n-provider.js';
 import { AnyRoute, createRouter, RouterOptions, RouterProvider } from '@tanstack/react-router';
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 import { createPortal } from 'react-dom';
 import { createRoot } from 'react-dom/client';
 
@@ -136,6 +136,7 @@ function BootSplash() {
 function App() {
     const [i18nLoaded, setI18nLoaded] = React.useState(false);
     const { extensionsLoaded } = useDashboardExtensions();
+    const hasExecutedExtensionCallbacks = useRef(false);
     useEffect(() => {
         // With this method we dynamically load the catalogs
         void dynamicActivate(defaultLocale, () => {
@@ -144,11 +145,15 @@ function App() {
         registerDefaults();
     }, []);
 
-    useEffect(() => {
-        if (extensionsLoaded) {
-            executeDashboardExtensionCallbacks();
-        }
-    }, [extensionsLoaded]);
+    if (extensionsLoaded && !hasExecutedExtensionCallbacks.current) {
+        // Run synchronously during render (not in an effect) so the extension
+        // registry is populated before the first route's own render/useMemo
+        // reads it. Guarded to run exactly once, since the registration
+        // functions append to arrays without deduping and re-running would
+        // duplicate every registered detail form / data table / nav item.
+        hasExecutedExtensionCallbacks.current = true;
+        executeDashboardExtensionCallbacks();
+    }
 
     if (!i18nLoaded || !extensionsLoaded) {
         // Show a minimal full-screen splash so the user sees that the app is
```

**Root cause** (in `node_modules/.../@vendure/dashboard/src/app/main.tsx`):

- `defineDashboardExtension()` only *queues* a callback; the actual registration into the `detailQueryDocumentRegistry` happens in `executeDashboardExtensionCallbacks()`.
- `App` (main.tsx:136-165) calls `executeDashboardExtensionCallbacks()` inside a `useEffect` gated on `extensionsLoaded` (lines 147-151), and only renders `<AppProviders><InnerApp/></AppProviders>` (mounting the router and the first routed page) once `extensionsLoaded` is already true.
- React commits effects child-first: when `App` re-renders and mounts `InnerApp`/router/first page for the very first time, that new subtree's own render (including the `useMemo` in `useExtendedDetailQuery`, called from `use-detail-page.ts:301`) runs *before* `App`'s own effect (`executeDashboardExtensionCallbacks`) fires — because that effect belongs to a component that is merely updating, not mounting.
- So if the very first page rendered in the session happens to be an order-detail page (hard refresh / deep link to `/orders/:id`), `getDetailQueryDocuments('order-detail')` reads an empty registry, and the query is built without our extension. React Query caches that unextended result under a key that doesn't encode the query shape (`['DetailPage', queryName, variables]`).

**Additional context**

JillK83:
## Summary

Fixes #5159

When a user hard-refreshes or deep-links directly to a detail page (e.g. `/orders/1`), 
query extensions registered via `extendDetailDocument` in `defineDashboardExtension` 
were silently ignored and the page loaded with the base query only.

## Root cause

1. `extensionsLoaded` flips `true` inside `useDashboardExtensions`'s `.then()`
2. React re-renders `App`, which now mounts `InnerApp` → router → the detail page
3. During this render, `useExtendedDetailQuery`'s `useMemo` reads `detailQueryDocumentRegistry` — it is empty
4. After render, React flushes effects — `App`'s `useEffect([extensionsLoaded])` fires `executeDashboardExtensionCallbacks()`, populating the registry
5. But the `useMemo` is keyed on static deps `[detailQuery, pageId]` and never recomputes — extensions are permanently invisible for that page load

## Fix

Move `executeDashboardExtensionCallbacks()` into the `runDashboardExtensions().then()` 
chain in `use-dashboard-extensions.ts`, before `setExtensionsLoaded(true)`. This 
guarantees the registry is populated before `App` re-renders `InnerApp`.

## Changes

- `packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts` — call `executeDashboardExtensionCallbacks()` before `setExtensionsLoaded(true)`
- `packages/dashboard/src/app/main.tsx` — remove now-redundant import and `useEffect`

## Testing

Existing orders e2e suite passes: `sales/orders.spec.ts` (22/22)

A dedicated e2e test for this fix would require scaffolding a test plugin that registers an `extendDetailDocument` extension on the order detail page, which is outside the scope of this PR. The fix is verifiable by reading the commit: the 
registry is guaranteed to be populated before `setExtensionsLoaded(true)` triggers a re-render. Happy to add a test if the maintainers can advise on the right approach for testing extension registration in the e2e suite.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5169"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790091317&installation_model_id=20193&pr_number=5169&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5169&signature=9668d8945590ad8a2a5cc4d424cf7d2f7e57d4ed63406ceb55c7488d1142acac"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
+++ b/packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
@@ -3,6 +3,11 @@ import { runDashboardExtensions } from 'virtual:dashboard-extensions';
 
 import { executeDashboardExtensionCallbacks, onExtensionSourceChange } from './define-dashboard-extension.js';
 
+// Extension registration writes into the global registry and must run exactly once
+// per page load. This module-level flag guards against duplicate execution caused by
+// React Strict Mode's mount/unmount/remount cycle (and any additional hook consumers).
+let extensionCallbacksExecuted = false;
+
 /**
  * @description
  * This hook is used to load dashboard extensions via the `virtual:dashboard-extensions` module,
@@ -15,10 +20,28 @@ export function useDashboardExtensions() {
     const [reloadCount, setReloadCount] = useState(0);
 
     useEffect(() => {
-        void runDashboardExtensions().then(() => {
-            executeDashboardExtensionCallbacks();
-            setExtensionsLoaded(true);
-        });
+        void runDashboardExtensions()
+            .catch(err => {
+                // eslint-disable-next-line no-console
+                console.error('Failed to load dashboard extensions', err);
+            })
+            .finally(() => {
+                try {
+                    if (!extensionCallbacksExecuted) {
+                        // Set before invoking so a concurrent second resolution
+                        // (Strict Mode) can't re-enter and register twice.
+                        extensionCallbacksExecuted = true;
+                        executeDashboardExtensionCallbacks();
+                    }
+                } catch (err) {
+                    // eslint-disable-next-line no-console
+                    console.error('Error executing dashboard extension callbacks', err);
+                } finally {
+                    // Always leave the loading screen, even if registration threw,
+                    // so the app never hangs on the boot splash.
+                    setExtensionsLoaded(true);
+                }
+            });
         onExtensionSourceChange(() => {
             // Setting this state var is only really done
             // in order to force a re-render of components using this hook.
--- a/packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
+++ b/packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
@@ -1,7 +1,7 @@
 import { useEffect, useState } from 'react';
 import { runDashboardExtensions } from 'virtual:dashboard-extensions';
 
-import { onExtensionSourceChange } from './define-dashboard-extension.js';
+import { executeDashboardExtensionCallbacks, onExtensionSourceChange } from './define-dashboard-extension.js';
 
 /**
  * @description
@@ -15,7 +15,10 @@ export function useDashboardExtensions() {
     const [reloadCount, setReloadCount] = useState(0);
 
     useEffect(() => {
-        void runDashboardExtensions().then(() => setExtensionsLoaded(true));
+        void runDashboardExtensions().then(() => {
+            executeDashboardExtensionCallbacks();
+            setExtensionsLoaded(true);
+        });
         onExtensionSourceChange(() => {
             // Setting this state var is only really done
             // in order to force a re-render of components using this hook.
--- a/packages/dashboard/src/app/main.tsx
+++ b/packages/dashboard/src/app/main.tsx
@@ -5,7 +5,6 @@ import {
     clearCustomFieldsMap,
     setCustomFieldsMap,
 } from '@/vdb/framework/document-introspection/add-custom-fields.js';
-import { executeDashboardExtensionCallbacks } from '@/vdb/framework/extension-api/define-dashboard-extension.js';
 import { useDashboardExtensions } from '@/vdb/framework/extension-api/use-dashboard-extensions.js';
 import { useExtendedRouter } from '@/vdb/framework/page/use-extended-router.js';
 import { useAuth } from '@/vdb/hooks/use-auth.js';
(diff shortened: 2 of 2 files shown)


JillK83:

packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
```diff
@@ -15,7 +15,10 @@ export function useDashboardExtensions() {
     const [reloadCount, setReloadCount] = useState(0);
 
     useEffect(() => {
-        void runDashboardExtensions().then(() => setExtensionsLoaded(true));
+        void runDashboardExtensions().then(() => {
+            executeDashboardExtensionCallbacks();
```
LucidityDesign: When I test this, `executeDashboardExtensionCallbacks` is called multiple times

packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
```diff
@@ -15,7 +15,10 @@ export function useDashboardExtensions() {
...
+            setExtensionsLoaded(true);
+        });
```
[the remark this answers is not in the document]
JillK83: Fixed in the latest push. Wrapped executeDashboardExtensionCallbacks() in try/finally so setExtensionsLoaded(true) always runs even if a callback throws. Also added .catch() on runDashboardExtensions() so a rejected load doesn't wedge the boot splash either

packages/dashboard/src/lib/framework/extension-api/use-dashboard-extensions.ts
JillK83: Thanks for testing and catching this. The multiple-execution issue was React Strict Mode invoking the effect twice. I added a module-level guard flag that's set before the call so the second invocation skips registration entirely. Pushed in the latest commit

LucidityDesign:
Please see the comment: https://github.com/vendurehq/vendure/issues/5159#issuecomment-5407397440

PR #5170, opened 2026-08-24
perf(core): reduce server and worker bootstrap time

michaelbromley:
# Description

Reduces the cold-start time of the Vendure server and worker. All changes are internal implementation changes: nothing importable from `@vendure/core` changes, and the generated GraphQL schemas are verified unchanged (admin SDL byte-identical before/after; shop SDL identical in content, with only the ordering of type definitions in the printed SDL shifting).

Measured results (medians, server totals include a health-check round-trip after bootstrap resolves):

| target | before | after | after + warm `NODE_COMPILE_CACHE` |
|---|---|---|---|
| server total (M4) | 744ms | 679ms | 657ms |
| worker total (M4) | 414ms | 314ms | 300ms |
| server + 20 synthetic plugins (M4) | 871ms | 729ms | 647ms |
| server total (Docker, 1 vCPU / 2GB) | ~1721ms | ~1386ms | ~1235ms |
| worker total (Docker, 1 vCPU / 2GB) | ~608ms | ~464ms | ~387ms |

The four optimizations:

1. **Batch plugin API extensions into a single `extendSchema` pass.** Each plugin's schema extension used to trigger a full schema rebuild, so the cost grew with plugin count. All extension documents are now merged with `concatAST` and applied in one pass. Extension documents were already all evaluated against the un-extended schema (the `.map()` ran to completion before the extend loop started), so visibility semantics are unchanged.
2. **Replace `stitchSchemas` with a single-pass type merge.** Four generator functions (list options, auth types, permissions enum, active order types) used `stitchSchemas` with a single subschema just to add or replace a handful of named types, paying for federation machinery (wrapper schema construction, delegation wiring) that this use case does not need. The new `mergeTypesIntoSchema` helper does one `rewireTypes` pass instead. Measured at ~104ms per boot across the six call sites.
3. **Stop loading heavy dependencies eagerly at import time.** Files that only need decorators from `@nestjs/graphql` now import from its `decorators` subpath instead of the package barrel, which was eagerly pulling in the drivers, federation, `@graphql-tools/*`, `subscriptions-transport-ws` and `fast-glob`. `express` and `http-proxy-middleware` are now required lazily at their single call sites. This removes roughly 500 modules from the eager require graph. The worker benefits most, since it never serves GraphQL but paid for the full graph.
4. **Cache GraphQL type definition loading across the shop and admin API builds.** The 31 `common/*.graphql` files were globbed, read and merged twice, once per API. They are now cached per glob pattern, with the merge order preserved so the output is unchanged.

The PR also adds a benchmark harness at `scripts/bootstrap-bench/` (per-phase timing, CPU-profile analyzers, a schema-equivalence dump script, and a synthetic plugin set modeled on real-world plugin behaviour) so these numbers can be reproduced and future regressions caught. `scripts/bootstrap-bench/README.md` documents usage, including how to benchmark under constrained CPU with Docker, and `FINDINGS.md` records the full attribution of where boot time goes plus deployment guidance (the `NODE_COMPILE_CACHE` column above requires no code change; it is a Node runtime feature that deployments can enable via an env var and a Docker build-stage warm-up).

Verification:
- Full core unit suite passes.
- Full core e2e suite passes on the master-based branch this was developed on; on this minor-based branch the schema-sensitive suites (auth, authentication-strategy, active-order-strategy, custom-permissions, custom-fields, collection, facet) pass.
- Admin and shop SDL dumped before and after and compared, for both a plain config and a 20-plugin config with custom fields and API extensions.

One behaviour note for reviewers: the printed order of fields inside merged input types (for example `OrderFilterParameter`) is preserved by an explicit reordering helper, since `stitchSchemas` used to emit SDL-declared fields first and codegen output is sensitive to field order. The only observable difference is the top-level ordering of type definitions in the printed shop SDL, which carries no semantic meaning.

# Breaking changes

None.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5170"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790163620&installation_model_id=20193&pr_number=5170&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5170&signature=e928e94b7abc6c1dafe07ef8fa8009702005f2966d78413d81122a36a24c0e9e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -229,7 +229,8 @@
       "version": "3.7.2",
       "dependencies": {
         "@apollo/server": "^4.12.2",
-        "@graphql-tools/stitch": "^10.1.16",
+        "@graphql-tools/merge": "^9.1.7",
+        "@graphql-tools/utils": "^10.11.0",
         "@nestjs/apollo": "~13.1.0",
         "@nestjs/common": "^11.0.12",
         "@nestjs/core": "^11.0.12",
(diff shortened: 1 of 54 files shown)


michaelbromley:
Addressed review findings:

- `packages/core/package.json` now declares `@graphql-tools/merge` and `@graphql-tools/utils` (which the code calls directly) and no longer declares `@graphql-tools/stitch`, which has no remaining references. Lockfile updated.
- Added unit tests: `merge-types-into-schema.spec.ts` covers add, replace-and-rewire, additional-schema merging with main-schema precedence, introspection-type exclusion and root-type preservation; `generate-auth-types.spec.ts` and `generate-permissions.spec.ts` lock down the placeholder-replacement behaviour with exact `printType` assertions. To make the generators testable under vitest they now import from `graphql/index.js`, the same dual-package-hazard workaround `generate-list-options.ts` already used.
- Ran prettier on `generate-list-options.ts` and the other touched files.
- The four `@nestjs/graphql` deep imports are centralized in `packages/core/src/common/nestjs-graphql-internals.ts` with the risk documented once; if a future `@nestjs/graphql` release moves those files there is now a single place to fix.
- `mergeTypesByPathsCached` carries a comment flagging that its merge options mirror a private `@nestjs/graphql` implementation detail, with the SDL-dump script named as the verification step when bumping that package.
- Trimmed the speculative future-work section from `scripts/bootstrap-bench/FINDINGS.md`; that material lives in issue tracking instead.

Note for reviewers on scope: the two dashboard docs pages (`use-detail-page.mdx`, `use-generated-form.mdx`) and the manifest change in this diff come from the docs-regeneration bot catching up documentation for a feature that already exists on `minor`; they are not part of this change.

michaelbromley:
Addressed the SonarCloud quality gate failures and the inline findings:

**Reliability (the critical bug finding):** the type-path sort in `mergeTypesByPathsCached` now uses an explicit comparator. Deliberately a code-unit comparator rather than `localeCompare`: the goal of that sort is to reproduce the default-sort file ordering that `GraphQLTypesLoader` applies to globbed paths, and `localeCompare` is locale-dependent and would diverge from it. Verified the generated SDL is byte-identical before and after by rebuilding both states and diffing all four dumps (plain/heavy x admin/shop). The same function's `Promise.all` callback is now `async`, resolving the promise-aggregator finding on the same line.

**Security (the path-injection and PATH-resolution findings):** the three benchmark scripts that take path arguments (`analyze-profile.js`, `nc-analyze-calltree.js`, `dump-sdl.js`) now resolve the argument and refuse anything outside the repository or the OS temp dir, and `bench.js` spawns `process.execPath` instead of resolving `node` from PATH.

**Code smells:** `node:` protocol prefixes across the bench scripts, `.at(-1)`, optional chaining, removed an unused destructured binding and a useless constructor in `heavy-plugins.js`, and `readonly` on the injected members of both fulltext search resolvers.

Not changed: the `.filter(notNullOrUndefined)` in the cache stays, since `mergeTypesByPaths` is typed as returning `string | null` and the filter is the honest way to narrow it.

PR #5174, opened 2026-08-25
fix(core): warn when migration patterns stop matching

biggamesmallworld:
Supersedes #5039. The diagnosis and the callback seam are @alex-hahn's work; this carries them forward with the detection gated correctly and the second half of the issue wired up. Opened as a new PR because #5039 had been stalled for four weeks and the original branch is blocked on an unsigned CLA.

## Description

`runMigrations()` returns an empty array in two different situations:

1. every migration has already been applied, and
2. the configured `migrations` glob patterns matched no files at all.

Both CLI entry points derive their report from that array alone, so the second case is presented as `No pending migrations found`. A database that is out of sync with the entity schema is therefore indistinguishable from an up-to-date one, and `vendure migrate --run` exits `0` having applied nothing.

The patterns are resolved relative to the current working directory, so this happens when the command is run from an unexpected directory, or when the patterns point at compiled output (e.g. `dist/migrations/*.js`) that has not been built.

## Approach

The condition is not visible in the return value, so it is handed to the caller through a new optional `RunMigrationsOptions.onDiagnostic` callback which both CLI entry points render. Reporting it any other way does not work:

- **Logging it directly does not work.** `log()` in `migrate.ts` is a no-op whenever `VENDURE_RUNNING_IN_CLI` is set, because the CLI owns the output. That suppression is why the condition is currently invisible.
- **Throwing would break the default project template.** The scaffold in `packages/create/templates/vendure-config.hbs` configures `migrations: [path.join(__dirname, './migrations/*.+(js|ts)')]` before any migration file exists, so a freshly created project would fail `vendure migrate --run` with a non-zero exit.

Two diagnostics go through the callback.

**`no-migrations-matched`.** Zero loaded migration classes on its own is not evidence of a problem: it is also what a project looks like before its first migration is authored, which is how the `create` scaffold ships. Warning on that alone reports an alarming, factually-true non-problem to exactly the users least able to evaluate it, and it regressed an existing CLI e2e case. The discriminator is the `migrations` table. Zero classes loaded while the database has migrations recorded as applied can only mean the patterns stopped matching, and that is also the dangerous real-world case: a deploy pointed at `dist/migrations/*.js` that was never built, against a database with history. Against a fresh database `MigrationExecutor.getExecutedMigrations()` returns `[]` rather than throwing, so no special-casing is needed.

**`schema-out-of-sync`.** `checkMigrationStatus()` already computed precisely the warning #5001 asks for and then discarded it through the suppressed `log()`, on every single `vendure migrate --run`. It now goes through the same seam. This costs nothing extra to detect and is the data-safety half of the issue.

The callback receives data (`patterns`, `cwd`, `queries`) rather than a pre-formatted English string, so a programmatic consumer can act on the condition instead of only printing it. The exit code is deliberately unchanged.

Two smaller fixes fall out of this:

- Both CLI entry points build their report from one `buildMigrationReport()`, rather than repeating the same block.
- A warning is no longer rendered with `log.success()`'s green glyph, and the multi-line detail no longer hangs outside the clack box when passed to `spinner.stop()`.

## Tests

- `packages/cli/e2e/migrate-command.e2e-spec.ts` — a new case drives real TypeORM against real sqlite: apply a migration, empty the migrations directory, run again. Verified that it fails on `master` and passes here. The existing `should report no pending migrations when none exist` case, which #5039 regressed, passes untouched.
- `packages/cli/src/commands/migrate/migration-report.spec.ts` — 5 tests over the report builder's branches.

Full `packages/cli` unit suite passes (277 tests), the migrate e2e suite passes (19 tests), and `tsc` reports no new errors in either package.

Note when verifying locally: the CLI e2e suite resolves `@vendure/core` through `packages/core/dist`, so it will happily pass against a build that predates the change. Build `packages/core` first.

Reference docs for the new types are left to the "Generate & commit docs" job.

## Not covered

`revertLastMigration()` has the same `log()` suppression, which #5001 also calls out. It does not currently run the schema check at all, so wiring it up is new behaviour rather than surfacing something already computed. Left for a follow-up.

The reporter's own symptom is still unexplained. Their glob matches, so `connection.migrations.length` is non-zero in their scenario and `no-migrations-matched` never fires. They see exit 0 in ~3s with no output at all, before any migration work happens. I could not reproduce it: every config-load failure I could construct against the e2e fixture exits non-zero with a message. That root cause needs a repro from the reporter (their tsconfig, plus `NODE_DEBUG=module npx vendure migrate --run`), which is why this is `Relates to` rather than `Fixes`.

Relates to #5001

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5174"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790240985&installation_model_id=20193&pr_number=5174&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5174&signature=9617df41971842f3fb587fde8e06b8ddc91d57a2d5ac47d491c1188a3a5d1be0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


gabriellbui (commented):
Three things I think need changing before this merges, plus one question about the release branch.

### The schema-drift warning now prints in the CLI, where it was suppressed before

`packages/core/src/migrate.ts:114-117`

The new `report()` helper calls `options?.onDiagnostic?.(diagnostic)` and then `log()`. Only `log()` checks `VENDURE_RUNNING_IN_CLI`, so the callback fires regardless and the CLI renders whatever it receives. `checkMigrationStatus` was previously invisible from the CLI for that reason. It no longer is.

`buildMigrationReport` sets `hasWarnings` for any diagnostic, so `handleNonInteractiveMode` in `packages/cli/src/commands/migrate/migrate.ts:86-92` also switches from `log.success` to `log.warn`.

The case that shows this: a fresh project with no migration files and an empty database, which is what `@vendure/create` gives you before the first migration is generated. `createSchemaBuilder().log()` returns the full schema as up-queries there. So the first `vendure migrate --run` on a new project prints a warning containing every CREATE TABLE, index, and foreign key, where it used to print one green line. The existing test at `packages/cli/e2e/migrate-command.e2e-spec.ts:199` covers exactly that scenario and still passes, because it asserts with `toContain`.

Surfacing schema drift in the CLI may well be worth doing, but it is a separate decision from #5001 and this makes it unbounded. Suggest dropping `schema-out-of-sync` from the callback, or keeping `hasWarnings` scoped to `no-migrations-matched` and capping the rendered queries.

### The new warning blames the working directory when the pattern is absolute

`packages/core/src/migrate.ts:163-167`

The message says patterns "are resolved relative to the current directory" and prints the cwd. But `packages/create/templates/vendure-config.hbs:49` scaffolds `path.join(__dirname, './migrations/*.+(js|ts)')`, and the CLI's own fixture at `packages/cli/e2e/fixtures/test-project/src/vendure-config.ts:15` does the same. Absolute paths ignore the cwd, so a user on a stock config goes looking for a working-directory problem that isn't there. In that setup the real cause is nearly always an unbuilt `dist`.

Suggest printing the cwd line only when at least one pattern fails `path.isAbsolute`, and for absolute patterns saying that nothing on disk matches the path.

### The `onDiagnostic` doc describes behaviour the option doesn't control

`packages/core/src/migrate.ts:91-97`

> These conditions are printed to the console by default, but that output is suppressed when running from the Vendure CLI so that the CLI can render them itself.

Suppression comes from `VENDURE_RUNNING_IN_CLI` alone (`migrate.ts:386-388`). Passing `onDiagnostic` doesn't change it. Someone calling `runMigrations` from their own deploy script and supplying the callback to render diagnostics their own way gets the built-in yellow text too. Either name the env var in the comment, or give the option a silent mode so the sentence becomes true.

### Question: is `master` the right target for the new exports?

`CLAUDE.md` puts bug fixes on `master` and new features on `minor`. The behaviour fix is clearly a bug fix. But this also adds to the public surface of `@vendure/core`, all tagged `@since 3.7.2` against a `3.7.2` package version:

- `describeDiagnostic()`
- `MigrationDiagnostic`, `NoMigrationsMatchedDiagnostic`, `SchemaOutOfSyncDiagnostic`, `RunMigrationsOptions`
- a second optional parameter on `runMigrations()`

Was that deliberate, or should the exported API go to `minor` and only the internal detection land as a patch?

biggamesmallworld:
Thanks, all three were fair. Fixed in 008e02adc.

### Schema drift in the CLI

You were right that this was unbounded, and the fresh-project case is the one that matters. `checkMigrationStatus` now returns early unless the database has migrations recorded as applied:

```ts
// Against a database with no migration history the schema builder reports the entire schema
// as pending, which is what a project looks like before its first migration is applied. That
// is not drift, so there is nothing useful to say about it.
const executed = await new MigrationExecutor(connection).getExecutedMigrations();
if (!executed.length) {
    return;
}
```

That is the same discriminator `no-migrations-matched` already uses. A schema builder reporting every table as pending against a database with nothing applied is not drift, it is a database that has not been set up yet.

I added the assertion to the existing case at `migrate-command.e2e-spec.ts:199` so it bites rather than passing on `toContain`:

```ts
expect(result.hasWarnings).toBe(false);
```

Confirmed it fails without the gate and passes with it.

`describeDiagnostic` also caps the rendered queries at 10 and appends `...and N more changes`. The full list stays on the diagnostic object, so a programmatic consumer still gets everything.

I kept `hasWarnings` covering both diagnostics rather than scoping it to `no-migrations-matched`. With the history gate, drift only fires on a database that has been migrated and has since diverged, which is worth the warning channel. If you would rather it stayed green there, say so and I will narrow it.

### The cwd on absolute patterns

Fixed. `describeDiagnostic` checks `path.isAbsolute` and only mentions the working directory when at least one pattern is relative. For absolute patterns it now says nothing on disk matches and to check that compiled output has been built, which as you say is nearly always the real cause on a stock config.

### The `onDiagnostic` doc

Corrected to name the env var:

> These conditions are also printed to the console, unless the `VENDURE_RUNNING_IN_CLI` environment variable is set, which the Vendure CLI does so that it can render them itself. Passing this callback does not suppress that output.

I went with naming it rather than adding a silent mode. It makes the sentence true for no extra API surface, and a caller who wants their own rendering can set the variable.

Also picked up CodeRabbit's `Successfully ran 1 migrations`.

### Branch target

Deliberate, and I would like to keep it on `master`.

The two diagnostics are not separable in practice. They share the `report()` seam, the `MigrationDiagnostic` union, `describeDiagnostic`, `buildMigrationReport` and `hasWarnings`. Splitting means landing that mechanism on `master` and then reopening the same four files on `minor` to add one union member and one switch case, in exchange for a versioning distinction no consumer can observe. Everything added is additive and optional: no export removed, `runMigrations`'s first parameter unchanged, `hasWarnings` optional, exit codes unchanged.

The underlying bug is a patch-shaped data-safety problem, `migrate --run` exiting 0 against a database it did not touch, so I would rather it went out with the patch than waited for a minor.

Verified after the changes: `packages/cli` unit suite 279 passing, migrate e2e 19 passing.

gabriellbui (commented):
Thanks, all three earlier points look handled. One problem came in with the new guard in `checkMigrationStatus` (`packages/core/src/migrate.ts`).

```ts
const executed = await new MigrationExecutor(connection).getExecutedMigrations();
if (!executed.length) {
    return;
}
```

An empty `migrations` table doesn't mean the database is empty. A schema built with `synchronize: true` has all its tables and no migration history, because TypeORM only writes to that table when it runs a migration. The guard treats those two states as the same, so it now hides a real mismatch.

Where that bites, and it's the path described in `docs/docs/guides/developer-guide/migrations/index.mdx`:

1. Dev has a local database built with `synchronize: true`. The `migrations` table is empty.
2. They set `synchronize: false` and add a custom field.
3. They run `vendure migrate --run`.
4. Before this commit: it printed the `ALTER TABLE` they need to generate a migration for, the same message the docs show at `index.mdx:38-39`.
5. Now: "No pending migrations found", and `hasWarnings` is false.

That message only exists at `migrate.ts:170`. Nothing checks the schema at server startup, so there is no other place they'd hear about it.

Suggest the guard asks whether the database has a schema at all, for example whether the core tables exist, rather than whether any migration has been recorded.

I read this rather than ran it, so worth confirming against a synchronize-built database.

biggamesmallworld:
Right, and confirmed against a synchronize-built database. Fixed in 91939b1ef.

The guard now asks whether the database has a schema at all, not whether anything has been recorded as applied:

```ts
if (!(await hasAnyEntityTable(connection))) {
    return;
}
```

`hasAnyEntityTable` runs `queryRunner.getTables()` over `connection.entityMetadatas.map(m => m.tablePath)` and returns whether any of them came back. No tables means a project which has not been set up, where the whole schema being pending is expected. Tables with an empty `migrations` table is the `synchronize: true` case, and drift there is real.

Reproduced your path in the CLI e2e suite: apply a migration, then `DELETE FROM migrations` and drop a table so the schema no longer matches, then empty the migrations directory. That leaves exactly the state you describe: tables present, no history, no migration files. It fails on the previous guard with `hasWarnings` false and passes now with the `ALTER TABLE` in the message.

The fresh-project case you raised first still stays green, because an empty database has no tables. `packages/cli/e2e/migrate-command.e2e-spec.ts:199` asserts that with `hasWarnings` false.

Two smaller things from CodeRabbit's second pass, in the same commit:

- `describeDiagnostic` no longer says all patterns resolve against the working directory when the `migrations` array mixes absolute and relative entries. The cwd is now a trailing line, printed only when at least one pattern is relative, and scoped to those.
- `connection.close()` and `resetConfig()` are in a nested `finally`, so an `onDiagnostic` callback that throws cannot leak the connection or leave global config installed.

Also merged `master` in to clear the `docs/manifest.json` conflict.

Verified: `packages/cli` unit suite 286 passing, full CLI e2e 48 passing.

gabriellbui (commented):
The table check reads much better than the migration-history one, and the CLI side is in good shape. Four things I'd like sorted before this goes in. Only the first one is a real design question, the other three are small.

### 1. Five new exported names, and we can't take them back

`packages/core/src/index.ts` gains `describeDiagnostic` plus four types. Once those ship in a release we're committed to them.

Looking at what the two diagnostic shapes actually carry: `patterns` and `cwd` get read in one place, to decide whether to print a line about the working directory. `queries` gets read in one place, to slice off the first ten and prefix each with a dash. Nothing acts on either. They're both just turned into text.

So the structure isn't buying anything today. Could both collapse into one shape carrying the already-rendered lines?

```ts
export interface MigrationDiagnostic {
    code: 'no-migrations-matched' | 'schema-out-of-sync';
    lines: string[];
}
```

`code` is the one piece the CLI genuinely needs, so it can swap "No pending migrations found" for "No migration files could be loaded". Core renders the wording where it detects the condition, `describeDiagnostic` stays module-private, and we export two names instead of five.

If someone later wants the raw glob list back, adding a field is easy. Removing three exported types isn't.

### 2. The ten-query cap ended up in the wrong package

When I asked for the drift list to be capped I meant in the CLI. It's landed in `describeDiagnostic`, which is also what core prints for itself.

So someone running `runMigrations(config)` from a deploy script, no callback and no CLI, used to get every pending statement and could paste them straight into a migration. Now they get ten and a count, with no way to see the rest short of changing their code. Can the cap move to the CLI renderer?

### 3. `hasAnyEntityTable` repeats the database work the next line does

`packages/core/src/migrate.ts:159`:

```ts
if (!(await hasAnyEntityTable(connection))) { return; }
const builderLog = await connection.driver.createSchemaBuilder().log();
```

`hasAnyEntityTable` calls `queryRunner.getTables()` over every entity table. TypeORM's `log()` opens its own query runner and calls `getTables()` over the same set (`RdbmsSchemaBuilder.js:101-107`). `getTables` never checks its cache before loading, and the two runners don't share one anyway, so we pay for it twice on every `migrate --run`.

Each load is not one query. On Postgres it's separate statements against `information_schema.tables`, `information_schema.columns` joined to `pg_attribute`, `pg_constraint`, `pg_class` for indices and again for foreign keys, each carrying one `OR` term per table. Core alone declares 65 entities, and `entityMetadatas` is bigger than that once junction tables are counted.

`queryRunner.hasTable()` on one known core table answers the same question in a single statement. It does hardcode a table name, which I think is fine when the question is only "has this database been set up".

### 4. `@since 3.7.2` should be `3.7.4`

3.7.3 shipped on 1 September, so the earliest this can appear in is 3.7.4. Five spots in `packages/core/src/migrate.ts`, lines 40, 62, 79, 88 and 189. The generated docs preview is already showing the wrong version.

---

Two things I'm happy to accept as they are, noting them so they're on the record rather than overlooked:

- The early return at the top of `detectUnmatchedPatterns` gives up as soon as any migration class loaded. A config mixing a direct class import with a glob, or one with two globs where only one breaks, won't warn. That's no worse than today and the single-glob case from the issue is caught, so I'd take the limitation. Worth a line of comment saying so.
- `migrate --run` still exits 0 on a warning. Keep it. Changing exit codes in a patch would break people's pipelines.

One last small thing: `packages/cli/src/commands/migrate/migration-report.spec.ts` fails `prettier --check` around line 54.

gabriellbui (approved):

PR #5171, opened 2026-08-25
fix(dashboard): never use a browser's translation service on the dashboard

knoid:
# Description

Given that the dashboard has per device language settings, translating it shouldn't be needed. In fact, this was why the app was breaking for me. This new attribute should prevent translation services from offering translations or even auto-translating.

# Breaking changes

No breaking changes.

# Screenshots

This error shows up when the dashboard is being automatically translated.
image

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5171"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790217776&installation_model_id=20193&pr_number=5171&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5171&signature=4efe28ced58a2c4b29ffa56be59e899767c181e2147a9f1d0bf0d27418f1578f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #5175, opened 2026-08-25
fix(dashboard): match the add-payment option exactly

biggamesmallworld:
# Description

`dashboard e2e (3)` has been failing on unrelated PRs. Same signature every time:

```
strict mode violation: locator('[role="dialog"]').getByRole('option', { name: /Test Payment/ })
resolved to 2 elements
```

Payment method options are labelled `${name} (${code})` (`add-manual-payment-dialog.tsx:109`). `createNewOrder` creates `Test Payment (test-payment)` at runtime. `payment-methods.spec.ts` creates `E2E Test Payment (e2e-test-payment)`, so the unanchored `/Test Payment/` in `orders.spec.ts:605` matches both from that spec's create test until its update test renames the row. Specs run in parallel, so the window is real.

The knock-on effect is what took the shard down. `orders.spec.ts:12` sets `mode: 'serial'`, so the failure retried the whole `Orders` block from the top. On the retry `should create, configure, and complete a draft order` ran against a database the first pass had already populated and timed out after a minute, and the remaining 17 tests in the file were skipped.

Two changes:

- Match `Test Payment (test-payment)` exactly.
- Scope the guard in `createNewOrder` to `code: { eq: "test-payment" }`. It previously created the method only when the payment method list came back empty, which is the same race in reverse: if `payment-methods.spec.ts` creates its method first, the list is non-empty, `test-payment` is never created, and the exact locator matches nothing.

Observed on #5090, #5171 and #5173 within 24 hours, always shard 3, always tests 75 and 86. That shard matrix does not run on master pushes, which is why it went unnoticed.

# Breaking changes

None. Test-only.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/docs/docs/reference/graphql-api/shop/mutations.mdx
+++ b/docs/docs/reference/graphql-api/shop/mutations.mdx
@@ -224,7 +224,7 @@ type Mutation {
   }}
 >
 {`"""
-Register a Customer account with the given credentials. There are three possible registration flows:
+Register a Customer account with the given credentials. There are four possible registration flows:
 
 _If \`authOptions.requireVerification\` is set to \`true\`:_
 
@@ -238,6 +238,19 @@ _If \`authOptions.requireVerification\` is set to \`true\`:_
 _If \`authOptions.requireVerification\` is set to \`false\`:_
 
 3. The Customer _must_ be registered _with_ a password. No further action is needed - the Customer is able to authenticate immediately.
+
+_Whatever the setting, if an account already exists for the email address through another authentication strategy
+(for example an SSO provider) and has no password yet:_
+
+4. **The supplied password is never stored.** A verificationToken is created and emailed to the address, and this mutation
+   answers with a generic success so that it does not reveal whether the account exists. The password is set only when that
+   token is passed to the \`verifyCustomerAccount\` mutation _with_ the chosen password, which proves the caller controls the
+   mailbox. This holds even when \`requireVerification\` is \`false\`, so the Customer cannot be authenticated straight after
+   registering. Registering again issues a fresh token and sends the email again.
+
+In every flow the caller-supplied \`firstName\`, \`lastName\`, \`phoneNumber\` and custom fields are ignored whenever a User already
+exists for the email address, since the caller has not proven they own it. This includes an account an administrator created
+earlier. A Customer with no User, such as one left by a guest checkout, is not an account and its details are still filled in.
 """
 type Mutation {
   registerCustomerAccount(input: RegisterCustomerInput!): RegisterCustomerAccountResult!
@@ -609,7 +622,9 @@ type Mutation {
   }}
 >
 {`"""
-Verify a Customer email address with the token sent to that address. Only applicable if \`authOptions.requireVerification\` is set to true.
+Verify a Customer email address with the token sent to that address. Applicable whenever a verificationToken was issued:
+that is when \`authOptions.requireVerification\` is set to true, and also when a password was registered against an account
+that already existed through another authentication strategy, whatever that setting is.
 
 If the Customer was not registered with a password in the \`registerCustomerAccount\` mutation, the password _must_ be
 provided here.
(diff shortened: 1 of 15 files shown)


gabriellbui (commented):
Two things on the test file, both about the payment method the new locator asks for.

### The method may not exist when the test runs

`createNewOrder` creates "Test Payment" only when the payment method list comes back empty:

```ts
// orders.spec.ts:838
const { paymentMethods } = await client.gql(`query { paymentMethods { items { id } } }`);
if (paymentMethods.items.length === 0) {
```

`e2e/fixtures/initial-data.ts:16` is `paymentMethods: []`, so this helper is the only thing that creates it.

`payment-methods.spec.ts:46` creates "E2E Test Payment". Specs run in parallel against one shared server, and nothing orders these two files. If that create lands before the order tests first call `createNewOrder`, the list is no longer empty, `test-payment` is never created, and the new exact locator matches zero options. The test then times out at `orders.spec.ts:612`.

The old `/Test Payment/` passed in that case because it matched `E2E Test Payment (e2e-test-payment)`. So this swaps a strict mode violation for a 10 second timeout, in the same serial block, with the same retry cascade behind it.

It usually heals: `payment-methods.spec.ts:102` deletes the row, and the guard re-runs on the next call. It stops healing when that spec fails before its delete test, since it is serial too and skips the rest.

Simplest fix is to stop racing for the row and seed it. `InitialData.paymentMethods` takes a name and a handler, and the populator derives the code from the name (`populator.ts:315`, `normalizeString(name, '-')`), so this gives exactly `Test Payment (test-payment)` before any spec starts:

```ts
// e2e/fixtures/initial-data.ts
paymentMethods: [
    {
        name: 'Test Payment',
        handler: {
            code: 'dummy-payment-handler',
            arguments: [{ name: 'automaticSettle', value: 'true' }],
        },
    },
],
```

That needs the cached seed databases deleted once (`rm -f packages/dashboard/e2e/__data__/*.sqlite`, files only, keep the directory). It also lets the guard at :838 go, and the comment below with it.

If you would rather keep the change minimal, scoping the check to the code fixes the race on its own:

```ts
paymentMethods(options: { filter: { code: { eq: "test-payment" } } })
```

### The new comment describes it wrongly

```ts
// orders.spec.ts:605
// `payment-methods.spec.ts` creates "E2E Test Payment" and a substring match
// resolves to both while that spec is mid-run.
```

Nothing seeds this method, so "the seeded method" (comment, title and description) points a reader at a fixture that does not exist. That is what hides the guard above.

"mid-run" is not the window either. `payment-methods.spec.ts:85` renames the method to "E2E Updated Payment", which no longer contains "Test Payment", so the overlap runs from the create test to the update test. And if that spec fails anywhere before the rename, the row keeps the colliding name for the rest of the run and every retry.

Seeding the method removes the coupling and the comment can go. If the comment stays, it should say the row is created at runtime and only deleted by the last test in the other file.

biggamesmallworld:
Both correct, thanks. Fixed in 04f03bf.

**The race.** Took the second option, scoping the guard to the code:

```ts
const { paymentMethods } = await client.gql(
    `query { paymentMethods(options: { filter: { code: { eq: "test-payment" } } }) { items { id } } }`,
);
```

I went with this over seeding `InitialData.paymentMethods` because the seed databases in `e2e/__data__/` are not keyed on the fixture contents. `SqljsInitializer` reuses the file if it exists, so seeding means every dev has to know to delete the cache or they get a confusing failure on an unrelated branch. CI is fine either way. Scoping the filter fixes the race without that footgun.

**The comment.** Rewritten to say where the method actually comes from, and the window is now the create-to-rename span rather than "mid-run":

```ts
// Options are labelled `${name} (${code})`. Match `test-payment` (created by
// `createNewOrder`) exactly: `payment-methods.spec.ts` creates "E2E Test Payment",
// and a substring match resolves to both until that spec renames or deletes it.
```

Title and description updated too, "the seeded method" is gone from both.

Ran `orders.spec.ts` and `payment-methods.spec.ts` together, 30 passed.

gabriellbui (approved):

PR #5180, opened 2026-08-25
fix(core): validate custom fields passed to registerCustomerAccount

JayMikeMill:
`addRegisterCustomerCustomFieldsInput()` adds a `customFields` field to `RegisterCustomerInput` whenever the Customer entity has public, writable custom fields, so those fields can be set during registration.

`CustomFieldProcessingInterceptor` never processed them. It builds its set of recognised inputs as `Create<Entity>Input` / `Update<Entity>Input`, and `RegisterCustomerInput` matches neither, so a `validate` function on such a field runs on `updateCustomer` but is silently skipped on `registerCustomerAccount`. A rejected value produced an account that was created successfully and carried an invalid custom field, with nothing reported.

The fix registers `RegisterCustomerInput` in `createInputsWithCustomFields` in the constructor, and maps it to `Customer` in `getEntityNameFromInputType()`. `hasCustomFields()` and `shouldApplyDefaults()` already consult that set, so they need no changes. Registration creates a Customer, so treating it as a create input is correct and defaults apply as they do elsewhere.

`validateInput()` also had its own copy of the entity-name derivation as an inline regex; it now calls `getEntityNameFromInputType()` so there is one implementation. The two are equivalent for every existing input (`CreateProductInput` -> `Product`, `OrderLineCustomFieldsInput` unchanged).

The existing `OrderLineCustomFieldsInput` special case is deliberately left as it is, to keep the diff additive.

### Tests

The new coverage lives in its own spec, `custom-field-register-customer.e2e-spec.ts`, rather than in `custom-fields.e2e-spec.ts`. Customer in that suite has only the readonly `score` field, which is what makes `writeableNonLocalizedFields` emit the `customFields: JSON` fallback shape on `UpdateCustomerInput`, and that shape is what lets "throws on attempt to update readonly field when no other custom fields defined" reach the interceptor. Adding a writable Customer custom field there generates `UpdateCustomerCustomFieldsInput` instead and breaks that test, which is the only coverage of the JSON-fallback readonly path.

The new spec asserts that an invalid value is rejected, that no Customer is left behind when it is, that a valid value is persisted on the entity, and that defaults are applied on registration.

Verified by reverting the interceptor change: 3 of the 4 new tests fail without it. All 8 `custom-field*` specs and `shop-auth.e2e-spec.ts` pass with it.

### Breaking changes

A behaviour change rather than a breaking API change: a `validate` function that previously only ran on update will now also reject invalid values at registration. That is the intent of the fix, but it may surface validators that were never exercised on this path before.


JayMikeMill:

biggamesmallworld (dismissed):
Diagnosis is right, but two things need fixing before this can land.

**1. This breaks an existing test.**

`Customer` currently has only `score`, which is readonly. `writeableNonLocalizedFields` filters readonly out (`graphql-custom-fields.ts:94`), so the length-0 branch emits `extend input UpdateCustomerInput { customFields: JSON }`. That is what lets `custom-fields.e2e-spec.ts:490` ("throws on attempt to update readonly field when no other custom fields defined") reach the interceptor and get `The custom field "score" is readonly`.

Adding a public writable `validateOnRegister` to `Customer` generates `UpdateCustomerCustomFieldsInput` instead, so `{ score: 5 }` is now rejected by GraphQL validation with a different message and that test fails. No e2e job ran on this PR, so CI did not catch it.

Please put the new field on an entity that has other writable custom fields, or otherwise keep the `Customer` JSON-fallback shape intact. Do not just update the expected string: that test is the only coverage of the JSON-fallback readonly path.

**2. The regression is untested.**

The bug is that an account gets created carrying an invalid custom field. Neither test asserts the account is not created. Add an admin `customers` query confirming `register-validation-invalid@test.com` does not exist. Also assert that `validateOnRegister` actually persisted on the Customer in the success case: validation passing and the value surviving to the entity are separate claims.

**3. Simpler mechanism.**

`createInputsWithCustomFields` (line 33) already expresses "this is a create input", and both `hasCustomFields` (84) and `shouldApplyDefaults` (110) already consult it. Adding

```ts
this.createInputsWithCustomFields.add('RegisterCustomerInput');
```

in the constructor, plus the entity-name lookup in `getEntityNameFromInputType`, gets the same behaviour with no changes to `hasCustomFields` or `shouldApplyDefaults`, and no third mechanism alongside the Create/Update sets and `OrderLineCustomFieldsInput`. As written, "RegisterCustomerInput is a create input" is stated in a fourth place, and a future entry that happens to be an update input would silently get defaults applied.

**Minor**

- `isNonStandardInputWithCustomFields` guards on `!!this.configService.customFields[entityName]`, but `default-config.ts:228` sets `Customer: []` and an empty array is truthy. The guard never gates anything. Drop it or use `.length`.
- The doc comment on `nonStandardInputEntityNames` is four lines of bug report. One line does it: `RegisterCustomerInput carries Customer custom fields but is not named CreateCustomerInput.` The rest belongs in the commit message.
- In the test validator, the `typeof value === 'string'` guard is noise given the field is typed `string`. Compare with `intListWithValidation` just above.

The `getEntityNameFromInputType` de-duplication in `validateInput` is a genuine improvement, and the equivalence holds.

biggamesmallworld (approved):
Approving after the rework.

The mechanism is now a single line in the constructor plus the entity-name mapping, so `hasCustomFields()` and `shouldApplyDefaults()` are untouched and there is no fourth place stating that this is a create input. The `validateInput()` de-duplication stands.

Test coverage moved to its own spec so that Customer in `custom-fields.e2e-spec.ts` keeps only the readonly `score` field, which preserves the `customFields: JSON` fallback shape on `UpdateCustomerInput` and with it the only coverage of the JSON-fallback readonly path. The new spec asserts that no Customer is left behind when validation rejects a value, that a valid value is persisted on the entity, and that defaults are applied on registration.

Full e2e matrix passes on sqljs, mariadb, mysql and postgres. The failing `deploy` check is the Vercel dashboard deploy, which cannot run on fork PRs.

Follow-up, not blocking: the same gap exists for `UpdateActiveAdministratorInput`, `ModifyOrderInput`, `UpdateOrderAddressInput` and the four wrapper inputs that nest `OrderLineCustomFieldsInput`. I will open an issue for those.

Issue #5179, opened 2026-08-25
Channel switcher still shows deleted channel until page refresh (dashboard)


Ryrahul:
**Describe the bug**
In the new React dashboard, the **channel switcher** at the top of the dashboard does not update when a channel is deleted. After deleting a channel, the deleted channel is still shown in the channel switcher dropdown until a full page refresh. (Creating a new channel does update the switcher correctly — only delete is affected.)

Relates to #5177 

**To Reproduce**

1. Go to **Settings → Channels**.
2. Select one or more channels and delete them via the bulk delete action.
3. Open the **channel switcher** dropdown at the top of the dashboard.
4. See that the deleted channel is **still listed** in the switcher.

**Expected behavior**
When a channel is deleted, it should immediately disappear from the channel switcher, without requiring a page refresh — mirroring the create-channel behaviour, which already updates the switcher correctly.

**Actual behavior**
The channel switcher continues to show the deleted channel until the page is hard-refreshed.

**Screenshots/Videos**
https://www.loom.com/share/f9d59896ef114a8fb7547f65bf16b5ab

**Error logs**
No errors are thrown — this is a stale cache issue, not a runtime error.

```
N/A
```

**Environment (please complete the following information):**

- @vendure/core version: 3.7.2
- Nodejs version: 22.22.3
- Database (mysql/postgres etc): N/A (dashboard-only, cache invalidation issue)
- Operating System (Windows/macOS/Linux): macOS 15.5
- Browser (if applicable): Any (Chrome/Safari)
- Package manager (npm/yarn/pnpm): npm

**Configuration**
Not configuration-related; reproducible on a default dashboard install.

```typescript
// N/A
```

**Minimal reproduction**
Reproducible on any standard Vendure dashboard install with more than one channel. No custom code required.

**Workaround**
Hard-refresh the page after deleting a channel.

Ryrahul:
Fixed in #5181 — the channel bulk delete now calls `refreshChannels()`, so the deleted channel is removed from the channel switcher without a page refresh.

biggamesmallworld:
Fixed by #5181 (and #5228 for deleting the active channel), both on master.

PR #5181, opened 2026-08-25
fix(dashboard): refresh channel switcher after deleting channels

Ryrahul:
Relates to #5179

# Description

The **channel switcher** at the top of the dashboard did not update when a channel was deleted — the deleted channel stayed in the switcher dropdown until a full page refresh. (Creating a channel already updated the switcher correctly; only delete was affected.)

The switcher renders its list from `ChannelProvider`, which is built primarily from the auth provider's `me.channels` (`staleTime: Infinity`). Create/update channel flows call `refreshChannels()` in their `onSuccess`, which invalidates `['currentUser']`, `['channels']` and `['activeChannel']`, so the switcher updates. The delete flow went through the generic `DeleteBulkAction` with no `onSuccess`, so it only refetched the Channels list-page table and never invalidated the queries the switcher depends on.

This PR wires `refreshChannels()` into the channel delete bulk action via the `DeleteBulkAction` `onSuccess` prop, reusing the exact same refresh path as create/update. The deleted channel now disappears from the switcher immediately, without a page refresh.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5181"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790261400&installation_model_id=20193&pr_number=5181&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5181&signature=edc23c22dad5a6d68da07203782e436793e78453c67ae71bc700b379511124f6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/settings/channels.spec.ts
+++ b/packages/dashboard/e2e/tests/settings/channels.spec.ts
@@ -151,6 +151,18 @@ test.describe('Channels CRUD', () => {
         await lp.expectSuccessToast();
 
         await expect(lp.getRows().filter({ hasText: 'e2e-test-channel' })).toHaveCount(0);
+
+        // #5179 — the deleted channel must also disappear from the channel
+        // switcher in the sidebar, not just the list. The switcher renders from
+        // the channel provider (me.channels), which the delete bulk action
+        // refreshes via refreshChannels(). Crucially there is no page.reload()
+        // here: a reload would repopulate the switcher from scratch and mask the
+        // bug this asserts.
+        const sidebar = page.locator('[data-slot="sidebar"]');
+        await sidebar.getByRole('button').first().click();
+        const switcherMenu = page.locator('[data-slot="dropdown-menu-content"]');
+        await expect(switcherMenu).toBeVisible();
+        await expect(switcherMenu.getByText('e2e-test-channel')).toHaveCount(0);
     });
 });
 
--- a/packages/dashboard/src/app/routes/_authenticated/_channels/components/channel-bulk-actions.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_channels/components/channel-bulk-actions.tsx
@@ -1,13 +1,16 @@
 import { DeleteBulkAction } from '../../../../common/delete-bulk-action.js';
 import { deleteChannelsDocument } from '../channels.graphql.js';
 import { BulkActionComponent } from '@/vdb/framework/extension-api/types/index.js';
+import { useChannel } from '@/vdb/hooks/use-channel.js';
 
 export const DeleteChannelsBulkAction: BulkActionComponent<any> = ({ selection, table }) => {
+    const { refreshChannels } = useChannel();
     return (
         <DeleteBulkAction
             mutationDocument={deleteChannelsDocument}
             entityName="channels"
             requiredPermissions={['DeleteChannel']}
+            onSuccess={refreshChannels}
             selection={selection}
             table={table}
         />


biggamesmallworld (changes requested):
This one is right, and it is right for the reason the other two were not. Rather than adding a local `invalidateQueries` call, you reached for `refreshChannels()`, the same path `channels_.$id.tsx:174` already uses for create and update. That keeps the knowledge of which keys the switcher depends on (`currentUser`, `channels`, `activeChannel`) inside `channel-provider.tsx` where it belongs, so the bulk action does not need to know any of it. Three lines, no new abstraction. Good instinct.

Requesting changes for one thing only: the test.

## Please extend the existing e2e test

`packages/dashboard/e2e/tests/settings/channels.spec.ts:141` already has `should bulk-delete the test channel`. Its final assertion is:

```ts
await expect(lp.getRows().filter({ hasText: 'e2e-test-channel' })).toHaveCount(0);
```

That passes both with and without your fix, because the list table was never the broken part. The switcher was (`packages/dashboard/src/lib/components/layout/channel-switcher.tsx`).

Add to that same test, after `expectSuccessToast()`: open the channel switcher and assert `e2e-test-channel` is gone from it. The important constraint is no `page.reload()` anywhere between the delete and the assertion, since a reload fixes the bug you are fixing and would make the test pass for the wrong reason.

Please do not add a new spec file for this. Extending the existing test is the right shape.

Run it with:

```
CI=true VITE_TEST_PORT=5176 bunx playwright test --config e2e/playwright.config.ts e2e/tests/settings/channels.spec.ts --reporter=list
```

from `packages/dashboard`.

## Worth a manual check, not blocking

Try deleting the channel you are currently switched into, and mention in the description what happens. `refreshChannels()` invalidates `activeChannel` straight away, while `localStorage` still holds the deleted channel's token. The provider does recover, at `channel-provider.tsx:204`, but only once `channels` has refetched, so there is a brief window where a request goes out under a token for a channel that no longer exists.

To be fair, that window arguably existed before your change too, and you may just be surfacing it sooner. I am not asking you to fix it here. I would just like to know what it looks like on screen before this merges.

## Nitpick, leave it alone

`delete-bulk-action.tsx:128` calls `onSuccess?.()` unconditionally, after counting failures. So a delete that fails with `NOT_DELETED`, for example on the default channel, still fires all three refetches. That is pre-existing behaviour of the shared component and not something this PR should take on.

Add the switcher assertion and I will approve. Thanks for the run of these, and sorry about the other two.

Ryrahul:
https://www.loom.com/share/1b827b0ca5a04a3faf8a67916dc74bb3

@biggamesmallworld 

Manual check (video): deleting the channel you're currently switched into leaves the dashboard on a broken active channel until a hard refresh — it doesn't self-recover.



**Pinned cause:** `['activeChannel']`  uses `retry: false`. On delete, `refreshChannels()` invalidates it and it refetches **while localStorage still holds the deleted channel's token** , the recovery effect  only runs later, after `me.channels` updates. That request fails under the dead token and, with `retry: false`, is never retried. The recovery effect then falls into the "pick first available channel" branch , which fixes the token but  unlike the token-out-of-sync branch at  does **not** re-invalidate `['activeChannel']`. So the active channel stays stuck on the failed response.

I've kept this PR scoped to the switcher fix as you asked and haven't touched this here. I've got the fix ready locally and verified it recovers without a reload — it just mirrors the sibling branch: after picking the first channel, invalidate `['activeChannel']`.

```ts
} else if (channels.length > 0) {
    // If no selected channel is set, use the first available channel
    const defaultChannel = channels[0];
    setSelectedChannelId(defaultChannel.id);
    setChannelTokenInLocalStorage(defaultChannel.token);
    // The active channel query may have been refetched (e.g. by refreshChannels()
    // after deleting the current channel) while localStorage still held the
    // now-deleted channel's token, and it is not retried (retry: false).
    // Invalidate it now that the token points at a valid channel so the active
    // channel recovers without a full page reload.
    queryClient.invalidateQueries({ queryKey: ['activeChannel', isAuthenticated] });
}
```

Happy to open this as a separate PR once this one lands.

biggamesmallworld (approved):
Nice work! 🙏

Issue #5182, opened 2026-08-25
Zone dropdown on Tax Rate / Channel pages shows stale list until page refresh (dashboard)


Ryrahul:
**Describe the bug**
In the new React dashboard, the **Zone** dropdown (`ZoneSelector`) does not reflect changes to zones until a full page refresh. After creating a new zone, it is missing from the dropdown; after deleting a zone, the deleted one still appears. This dropdown is used on the **Tax Rate** detail/create page (zone selection) and on the **Channel** detail page (default tax zone / default shipping zone).

This is the same class of bug as the tax category selector (#5177) and the channel switcher (#5179).

**To Reproduce**

1. Go to **Settings → Zones** and create a new zone (e.g. "TestZone").
2. Navigate to **Settings → Tax rates → Create new tax rate** (or open the Channel detail page).
3. Open the **Zone** dropdown.
4. See that the newly created zone is **not** listed. Conversely, if you delete a zone and open the dropdown, the deleted one is **still** listed (until a hard refresh, or until the 5-minute react-query `staleTime` elapses).

**Expected behavior**
The Zone dropdown should immediately reflect newly created and deleted zones, without requiring a page refresh.

**Actual behavior**
The dropdown shows a stale list. Newly created zones are missing and deleted zones remain, until the page is hard-refreshed.

**Screenshots/Videos**
N/A

**Error logs**
No errors are thrown — this is a stale cache issue, not a runtime error.

```
N/A
```

**Environment (please complete the following information):**

- @vendure/core version: 3.7.2
- Nodejs version: 22.22.3
- Database (mysql/postgres etc): N/A (dashboard-only, cache invalidation issue)
- Operating System (Windows/macOS/Linux): macOS 15.5
- Browser (if applicable): Any (Chrome/Safari)
- Package manager (npm/yarn/pnpm): npm

**Configuration**
Not configuration-related; reproducible on a default dashboard install.

```typescript
// N/A
```

**Minimal reproduction**
Reproducible on any standard Vendure dashboard install with more than one zone. No custom code required.

**Workaround**
Hard-refresh the page (or wait 5 minutes for the query to become stale) after creating or deleting a zone.

=

Issue #5177, opened 2026-08-25
Dashboard: tax category selector not updated after create/delete


Ryrahul:
**Describe the bug**
In the new React dashboard, the **Tax category** dropdown on the Tax Rate detail/create page does not reflect changes to tax categories until a full page refresh. After creating a new tax category, it is missing from the dropdown; after deleting a tax category, the deleted one still appears in the dropdown.

**To Reproduce**

1. Go to **Settings → Tax categories** and create a new tax category (e.g. "Reduced Rate").
2. Navigate to **Settings → Tax rates → Create new tax rate** (or open an existing one).
3. Open the **Tax category** dropdown.
4. See that the newly created tax category is **not** listed. Conversely, if you delete a tax category and open the dropdown, the deleted one is **still** listed.

**Expected behavior**
The Tax category dropdown on the Tax Rate page should immediately reflect newly created and deleted tax categories, without requiring a page refresh.

**Actual behavior**
The dropdown shows a stale list. Newly created categories are missing and deleted categories remain, until the page is hard-refreshed (or the 5-minute react-query `staleTime` elapses).

**Screenshots/Videos**

https://www.loom.com/share/1f586e19f9cd42b3ae16e7d4310f2775

**Error logs**
No errors are thrown — this is a stale cache issue, not a runtime error.

```
N/A
```

**Environment (please complete the following information):**

- @vendure/core version: 3.7.2
- Nodejs version: 22.22.3
- Database (mysql/postgres etc): N/A (dashboard-only, cache invalidation issue)
- Operating System (Windows/macOS/Linux): macOS 15.5
- Browser (if applicable): Any (Chrome/Safari)
- Package manager (npm/yarn/pnpm): npm

**Configuration**
Not configuration-related; reproducible on a default dashboard install.

```typescript
// N/A
```

**Minimal reproduction**
Reproducible on any standard Vendure dashboard install with more than one tax category. No custom code required.

**Workaround**
Hard-refresh the page (or wait 5 minutes for the query to become stale) after creating or deleting a tax category.

PR #5178, opened 2026-08-25
fix(dashboard): refresh tax category selector after mutations

Ryrahul:
Relates to #5177

# Description

The **Tax category** dropdown on the Tax Rate detail/create page in the dashboard was showing a stale list. `TaxCategorySelector` caches its options under the react-query key `['taxCategories']` with a 5-minute `staleTime`, but the tax category create/update and delete flows never invalidated that key. As a result, newly created tax categories were missing from the dropdown and deleted ones remained, until a full page refresh.

This PR invalidates `['taxCategories']` after the relevant mutations, following existing dashboard conventions:

- **Create/update** (`_tax-categories/tax-categories_.$id.tsx`): call `queryClient.invalidateQueries({ queryKey: ['taxCategories'] })` in the detail page `onSuccess`.
- **Delete** (`_tax-categories/components/tax-category-bulk-actions.tsx`): pass `invalidateQueries={['taxCategories']}` to the shared `DeleteBulkAction`.

The dropdown now reflects created and deleted tax categories immediately, without a page refresh.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

---
[...]

PR #5183, opened 2026-08-25
fix(dashboard): refresh zone selector after mutations

Ryrahul:
Relates to #5182

# Description

The **Zone** dropdown (`ZoneSelector`) in the dashboard was showing a stale list. It is used on the **Tax Rate** detail/create page (zone selection) and on the **Channel** detail page (default tax zone / default shipping zone). `ZoneSelector` caches its options under the react-query key `['zones']` with a 5-minute `staleTime`, but the zone create/update and delete flows never invalidated that key. As a result, newly created zones were missing from the dropdown and deleted ones remained, until a full page refresh.

This PR invalidates `['zones']` after the relevant mutations, following existing dashboard conventions 

- **Create/update** (`_zones/zones_.$id.tsx`): call `queryClient.invalidateQueries({ queryKey: ['zones'] })` in the detail page `onSuccess`.
- **Delete** (`_zones/components/zone-bulk-actions.tsx`): pass `invalidateQueries={['zones']}` to the shared `DeleteBulkAction`.

The dropdown now reflects created and deleted zones immediately, without a page refresh.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5183"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790262618&installation_model_id=20193&pr_number=5183&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5183&signature=efc6da394f3723513b1633656163b9881a7cc7eb0b3d5a7c42c6d0f1a40ab130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_zones/components/zone-bulk-actions.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_zones/components/zone-bulk-actions.tsx
@@ -14,6 +14,7 @@ export const DeleteZonesBulkAction: BulkActionComponent<any> = ({ selection, tab
             mutationDocument={deleteZonesDocument}
             entityName="zones"
             requiredPermissions={['DeleteZone']}
+            invalidateQueries={['zones']}
             selection={selection}
             table={table}
         />
--- a/packages/dashboard/src/app/routes/_authenticated/_zones/zones_.$id.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_zones/zones_.$id.tsx
@@ -16,6 +16,7 @@ import {
 import { detailPageRouteLoader } from '@/vdb/framework/page/detail-page-route-loader.js';
 import { useDetailPage } from '@/vdb/framework/page/use-detail-page.js';
 import { Trans, useLingui } from '@lingui/react/macro';
+import { useQueryClient } from '@tanstack/react-query';
 import { createFileRoute, useNavigate } from '@tanstack/react-router';
 import { toast } from 'sonner';
 import { ZoneCountriesTable } from './components/zone-countries-table.js';
@@ -43,6 +44,7 @@ function ZoneDetailPage() {
     const navigate = useNavigate();
     const creatingNewEntity = params.id === NEW_ENTITY_PATH;
     const { t } = useLingui();
+    const queryClient = useQueryClient();
 
     const { form, submitHandler, entity, isPending, resetForm } = useDetailPage({
         pageId,
@@ -59,6 +61,7 @@ function ZoneDetailPage() {
         params: { id: params.id },
         onSuccess: async data => {
             toast.success(creatingNewEntity ? t`Successfully created zone` : t`Successfully updated zone`);
+            await queryClient.invalidateQueries({ queryKey: ['zones'] });
             resetForm();
             if (creatingNewEntity) {
                 await navigate({ to: `../$id`, params: { id: data.id } });


biggamesmallworld:
Thanks for this, and for #5178 before it. The diagnosis is right and the fix works, so I want to be clear that nothing here is wrong. I'm going to close it anyway, because I think we should fix the cause rather than the sixteenth symptom, and I'd rather you get the credit for that PR than for this one.

## Why this is symptom-level

This is the second byte-identical fix in two days. #5178 did it for `['taxCategories']`, this one does it for `['zones']`. There are fourteen more sites waiting.

The shape of the bug is: a shared selector component opts into a `staleTime`, and from then on every mutation site anywhere in the app is obliged to invalidate a cache key it does not own and cannot see. Nothing links the two. If someone renames the key in `zone-selector.tsx`, both invalidations in this PR silently become no-ops, and no type error or lint rule catches it. That obligation is unbounded: any future page that mutates a zone inherits it, and will forget.

## The underlying issue

The dashboard's `QueryClient` already sets `staleTime: 0` globally (`packages/dashboard/src/app/app-providers.tsx:12`). React Query's default is refetch-on-mount, which is exactly the behaviour these dropdowns want. Every one of these bugs is a per-site opt-in *out* of correct behaviour, in exchange for caching that buys us nothing: these are `take: 100` dropdown queries that fire once when a detail page mounts.

So the fix is to delete the opt-ins, not to add invalidations. That removes the obligation entirely instead of documenting it.

## What a replacement PR should do

Remove `staleTime` from the queries backing mutable entities. I count 16 lines across 14 files:

| File | Query key | Current `staleTime` |
|---|---|---|
| `lib/components/shared/channel-selector.tsx` | `['channels']` | 5m |
| `lib/components/shared/tax-category-selector.tsx` | `['taxCategories']` | 5m |
| `lib/components/shared/zone-selector.tsx` | `['zones']` | 5m |
| `lib/components/shared/customer-group-selector.tsx` | `['customerGroups']` | 5m |
| `lib/components/shared/language-selector.tsx` | `['availableGlobalLanguages']` | 5m |
| `lib/components/shared/country-selector.tsx` | `['countries', search]` | 1h |
| `lib/components/shared/customer-address-form.tsx` | `['availableCountries']` | 24h |
| `app/routes/_authenticated/_shipping-methods/components/test-address-form.tsx` | `['availableCountries']` | 24h |
| `lib/components/shared/product-variant-selector.tsx` | | 5m |
| `lib/components/shared/customer-selector.tsx` | | 1m |
| `lib/components/shared/asset/asset-gallery.tsx` | | 5m |
| `app/routes/_authenticated/_products/components/add-option-group-dialog.tsx` | | 1m |
| `app/routes/_authenticated/_collections/components/move-collections-dialog.tsx` | | 5m (x2) |
| `app/routes/_authenticated/_collections/collections.tsx` | | 5m (x2) |

Two of those are worth calling out. `['availableCountries']` at 24 hours means an admin can add a country and not see it in the address form for a day. And `collections.tsx:106` already carries a comment describing this exact bug being worked around locally, which is a good sign the caching is not paying for itself.

Then revert what is now redundant: the four lines in this PR, and the four in #5178. They become dead once the selectors refetch on mount.

Leave these alone. They are server config, immutable until the server restarts, and the cache is doing real work:

- `configurable-operation-selector.tsx`, `configurable-operation-multi-selector.tsx`
- `fulfillment-handler-selector.tsx`, `fulfill-order-dialog.tsx` (x2)
- `duplicate-entity-dialog.tsx`
- `lib/providers/auth.tsx` and `lib/providers/server-config.tsx` (both `Infinity`)

One judgement call I'd leave to you: `asset-gallery.tsx` is the heaviest query in the list, so if refetch-on-mount is noticeable there, keep its `staleTime` and say so in the PR. The rest should just go.

## Tests

This PR needs one, and so did #5178. Missing invalidation is invisible to typecheck, lint and unit tests, which is why two rounds of it landed unnoticed.

One e2e test covering the class is enough, not sixteen. `packages/dashboard/e2e/tests/settings/zones.spec.ts` already exists, and `e2e/tests/regression/breadcrumb-stale-after-mutation.spec.ts` is precedent for this exact bug shape. Create a zone, open the tax rate page, assert it appears in the dropdown without a reload.

## Practical notes

- Target `master`.
- Use `Fixes #5182` and `Fixes #5177` in the body, since one PR closes both.
- Expect a net negative diff. That is the point.

Sorry to send you round again on something that already works. Two identical fixes in two days is a signal about our code, not about your PRs, and I'd rather act on it now than merge fourteen more. Happy to review the replacement as soon as it's up.

PR #5184, opened 2026-08-25
feat(core): add auth options disable last login update property

biggamesmallworld:
# Description

Adds a new `authOptions.disableLastLoginUpdate` config option (default `false`). When set to `true`, Vendure skips the `User.lastLogin` write on every successful login.

On installations with high authentication traffic, that write causes lock contention on the `user` table (the original report cited 250k+ updates/day). This option lets those installations opt out of tracking `lastLogin`.

This is a take-over of #3253 by [@alexisvigoureux](https://github.com/alexisvigoureux), rebased onto current `minor`. The original author is credited via `Co-authored-by`, and #3253 has been closed in favour of this PR.

Changes made on top of the original PR, from the review feedback there:
- Removed the incorrect `user.roles` condition. Roles are loaded just above that check, so the flag never actually took effect.
- Set `@since 3.8.0` on the new option.
- Added e2e coverage for both the default and the disabled behaviour.

Relates to #3253

# Breaking changes

None. The option defaults to `false`, so existing behaviour is unchanged.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have added or updated test cases (`packages/core/e2e/disable-last-login-update.e2e-spec.ts`: default updates `lastLogin`; with the flag set to `true`, login leaves it null)
- [ ] I have updated the README if needed (not needed; the option is documented via the generated `AuthOptions` reference docs)

### DIFF
--- a/packages/core/e2e/disable-last-login-update.e2e-spec.ts
+++ b/packages/core/e2e/disable-last-login-update.e2e-spec.ts
@@ -1,0 +1,75 @@
+import { mergeConfig } from '@vendure/core';
+import { createTestEnvironment } from '@vendure/testing';
+import path from 'path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import { initialData } from '../../../e2e-common/e2e-initial-data';
+import { TEST_SETUP_TIMEOUT_MS, testConfig } from '../../../e2e-common/test-config';
+
+import { getCustomerDocument, getCustomerListDocument } from './graphql/shared-definitions';
+
+describe('authOptions.disableLastLoginUpdate (default false)', () => {
+    const { server, adminClient, shopClient } = createTestEnvironment(testConfig());
+
+    beforeAll(async () => {
+        await server.init({
+            initialData,
+            productsCsvPath: path.join(__dirname, 'fixtures/e2e-products-minimal.csv'),
+            customerCount: 1,
+        });
+        await adminClient.asSuperAdmin();
+    }, TEST_SETUP_TIMEOUT_MS);
+
+    afterAll(async () => {
+        await server.destroy();
+    });
+
+    it('updates User.lastLogin on successful login', async () => {
+        const { customers } = await adminClient.query(getCustomerListDocument);
+        const customer = customers.items[0];
+
+        const before = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(before.customer?.user?.lastLogin).toBeNull();
+
+        await shopClient.asUserWithCredentials(customer.emailAddress, 'test');
+
+        const after = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(after.customer?.user?.lastLogin).not.toBeNull();
+    });
+});
+
+describe('authOptions.disableLastLoginUpdate (true)', () => {
+    const { server, adminClient, shopClient } = createTestEnvironment(
+        mergeConfig(testConfig(), {
+            authOptions: {
+                disableLastLoginUpdate: true,
+            },
+        }),
+    );
+
+    beforeAll(async () => {
+        await server.init({
+            initialData,
+            productsCsvPath: path.join(__dirname, 'fixtures/e2e-products-minimal.csv'),
+            customerCount: 1,
+        });
+        await adminClient.asSuperAdmin();
+    }, TEST_SETUP_TIMEOUT_MS);
+
+    afterAll(async () => {
+        await server.destroy();
+    });
+
+    it('does not update User.lastLogin on successful login', async () => {
+        const { customers } = await adminClient.query(getCustomerListDocument);
+        const customer = customers.items[0];
+
+        const before = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(before.customer?.user?.lastLogin).toBeNull();
+
+        await shopClient.asUserWithCredentials(customer.emailAddress, 'test');
+
+        const after = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(after.customer?.user?.lastLogin).toBeNull();
+    });
+});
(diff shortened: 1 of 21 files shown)


packages/core/src/service/services/auth.service.ts
```diff
@@ -99,8 +99,10 @@ export class AuthService {
         if (ctx.session && ctx.session.activeOrderId) {
             await this.sessionService.deleteSessionsByActiveOrderId(ctx, ctx.session.activeOrderId);
         }
-        user.lastLogin = new Date();
-        await this.connection.getRepository(ctx, User).save(user);
+        if (!this.configService.authOptions.disableLastLoginUpdate) {
+            user.lastLogin = new Date();
+            await this.connection.getRepository(ctx, User).save(user);
+        }
```
gabriellbui, line 104: This `save` is the only place that persists the `User` object an `AuthenticationStrategy.authenticate()` returns. With the option on, any field a strategy set on that user (`verified`, custom fields) is dropped with no error and no log line.

Nothing in this repo depends on it. Custom strategies do: the Facebook example in `docs/docs/guides/core-concepts/auth/index.mdx:369-370` ends with `user.verified = true; return user;` and no save of its own. An integrator who followed that shape and then enables this option loses the write, and a user created unverified is never marked verified, so the next login returns `NotVerifiedError`.

Keep the save unconditional, gate only the assignment:

```suggestion
        if (!this.configService.authOptions.disableLastLoginUpdate) {
            user.lastLogin = new Date();
        }
        await this.connection.getRepository(ctx, User).save(user);
```

That leaves the write on the `user` row, so it does not solve the contention this PR is for. A `repository.update({ id: user.id }, { lastLogin: new Date() })` in place of `save()` would be one narrow UPDATE with no SELECT and no `user_roles` join-table diff, and might make the option unnecessary. Was the contention measured against `save()` specifically?

packages/core/e2e/disable-last-login-update.e2e-spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        const { customers } = await adminClient.query(getCustomerListDocument);
+        const customer = customers.items[0];
+
+        const before = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(before.customer?.user?.lastLogin).toBeNull();
+
+        await shopClient.asUserWithCredentials(customer.emailAddress, 'test');
+
+        const after = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(after.customer?.user?.lastLogin).toBeNull();
```
gabriellbui: This test passes when the login fails.

`asUserWithCredentials` (`packages/testing/src/simple-graphql-client.ts:138-154`) does not throw on a failed login. Its `LOGIN` document selects `... on ErrorResult`, so an `InvalidCredentialsError` or `NotVerifiedError` comes back as ordinary data and the method returns the error object. The return value is discarded here.

If the seed password changes, or shop login regresses, `lastLogin` stays null because nobody logged in, and this still passes. It cannot fail for the reason it exists. It also never checks that login still works with the option on, which is the actual risk of the change.

```suggestion
        const login = await shopClient.asUserWithCredentials(customer.emailAddress, 'test');
        expect(login.identifier).toBe(customer.emailAddress);

        const after = await adminClient.query(getCustomerDocument, { id: customer.id });
        expect(after.customer?.user?.lastLogin).toBeNull();
```

packages/core/e2e/disable-last-login-update.e2e-spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        const { customers } = await adminClient.query(getCustomerListDocument);
+        const customer = customers.items[0];
+
+        const before = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(before.customer?.user?.lastLogin).toBeNull();
+
+        await shopClient.asUserWithCredentials(customer.emailAddress, 'test');
+
+        const after = await adminClient.query(getCustomerDocument, { id: customer.id });
+        expect(after.customer?.user?.lastLogin).not.toBeNull();
```
gabriellbui: This assertion always passes. If `customer` is undefined (wrong id, permissions change, query shape drift) the expression is `undefined`, and `expect(undefined).not.toBeNull()` passes. The one assertion proving the default behaviour still works would pass against a query that returned nothing.

The `toBeNull()` calls at lines 32, 68 and 73 are safe by accident, since `undefined` is not `null`.

Capture a timestamp before the login and assert against it:

```suggestion
        expect(new Date(after.customer!.user!.lastLogin).getTime()).toBeGreaterThan(startOfTest);
```

gabriellbui (commented):
Three findings, all traced through the code.

packages/core/src/service/services/auth.service.ts
biggamesmallworld: Good catch, taken as suggested in 99dfbf0c6. The save is now unconditional and only the `lastLogin` assignment is gated.

On the contention question: with no changed columns TypeORM issues no UPDATE, so the option still removes the write on the `user` row, which is what the original report was about. The remaining SELECT does not take a row lock.

I have not measured `save()` vs `update()` myself, and I cannot re-measure the original reporter's workload. Swapping `save()` for a narrow `update()` would drop the strategy-set fields on every login, not just when the option is on, so I have kept it out of this PR.

packages/core/e2e/disable-last-login-update.e2e-spec.ts
biggamesmallworld: Right, fixed in 99dfbf0c6. Both tests now assert `login.identifier` so a failed login fails the test, which also covers "login still works with the option on".

packages/core/e2e/disable-last-login-update.e2e-spec.ts
biggamesmallworld: Fixed in 99dfbf0c6. Now captures a timestamp before the login and asserts `lastLogin` is at or after it, with 1s of slack for DBs that store datetimes at second precision. Also switched the optional chaining to `!` so a missing customer throws instead of passing.

gabriellbui (approved):

PR #5192, opened 2026-08-26
fix(dashboard): fix country dropdown in address forms

### LINKED ISSUE #5191: Country dropdown on customer address form is incomplete and doesn't show an existing address's country
**Describe the bug**
In the new React dashboard, the **Country** dropdown on the customer address form does not show all countries, and does not correctly display the country of an existing address.

Two related symptoms:

1. **Newly created countries are missing from the dropdown.** After creating a new (enabled) country under Settings → Countries, it does not appear in the country dropdown of the customer address form — even after a full page refresh.
2. **An existing address does not show its country when editing.** Opening an existing customer address for editing, the Country field does not display / pre-select the country that was previously saved.

**To Reproduce**

Symptom 1 (new country missing):
1. Go to **Settings → Countries** and create a new country (make sure it is enabled).
2. Go to a **Customer → Addresses** and add or edit an address.
3. Open the **Country** dropdown.
4. Scroll through the full list — the newly created country is not present.

Symptom 2 (existing address country not shown):
1. Open a **Customer** that has a saved address.
2. Edit one of the existing addresses.
3. Observe the **Country** field — the previously saved country is not shown/selected.

**Expected behavior**
1. All enabled countries (including newly created ones) should appear in the Country dropdown, ideally sorted by name.
2. When editing an existing address, the Country field should display and pre-select the address's saved country.

**Actual behavior**
1. Only a subset of countries appears; newly added countries are missing even after a refresh.
2. The Country field is blank / not pre-selected when editing an address whose country is not among the shown subset.

**Screenshots/Videos**
https://www.loom.com/share/3ae648bfe2eb4f688cee2b629c840097

**Error logs**
No errors are thrown — the country is simply not present in the dropdown.

```
N/A
```

- Nodejs version: 22.22.3
- Operating System (Windows/macOS/Linux): macOS 15.5
- Browser (if applicable): Any (Chrome/Safari)

**Configuration**
Not configuration-related; reproducible on a default dashboard install with enough countries that the newly created one falls outside the shown subset.

```typescript
// N/A
```

**Minimal reproduction**
Reproducible on any standard Vendure dashboard install: create an enabled country, then open a customer address form and inspect the Country dropdown. Also visible when editing an existing address whose country is not in the shown subset.

**Affected areas**
- The Country selector on the customer address form (customer detail → addresses, and draft order address).
- The same country dropdown on the shipping method test-address form.

**Additional context**

- Found while recording a demo video for the dashboard selector-staleness fix (#5177 / #5182): after creating a country to test the selector behaviour, it never showed up in the address-form country dropdown, and editing an existing address surfaced the second symptom.
- Happens consistently.
- The two symptoms appear related: an address whose country is not among the countries the dropdown loads will neither be shown in edit mode nor selectable.
- The dropdown appears to load only a subset of countries (and unsorted), so countries beyond that subset — including newly created ones — never render.

Ryrahul:
Fixes #5191

# Description

The **Country** dropdown on the customer address form had two problems:

1. **Countries could be missing from the list.** A country that existed and was returned by the API was absent from the dropdown (verified: the API returned all countries including the "missing" one, but it wasn't in the rendered options).
2. **Editing an existing address didn't pre-select its country.** Opening an existing address for editing showed the placeholder instead of the saved country.

Both came down to the country `<Select>` not following the pattern used by the other entity selectors in the dashboard:

- The `GetAvailableCountries` query fetched countries **unsorted**, unlike `country-selector.tsx` / `customer-group-selector.tsx` which sort by name. Adding `sort: { name: ASC }` makes all countries render in a predictable order.
- The `<Select>` used `value={field.value || undefined}` with no `SelectValue` label resolver, so a value set before the item list loaded was discarded and never re-applied. Aligning it with the working selectors (`value={field.value ?? ''}`, `onValueChange={value => value && field.onChange(value)}`, and a `<SelectValue>` children function that resolves the label from the loaded data) fixes the pre-selection.

Applied in both places that use this dropdown:
- `packages/dashboard/src/lib/components/shared/customer-address-form.tsx` (customer address add/edit, draft-order addresses)
- `packages/dashboard/src/app/routes/_authenticated/_shipping-methods/components/test-address-form.tsx` (shipping-method test address)

Adds e2e tests covering both: the dropdown includes a newly created country in sorted order, and editing an existing address pre-selects its country.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5192"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790340275&installation_model_id=20193&pr_number=5192&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5192&signature=01ee875eff4a9c6b50061de507e3ccce7952b7d713a64653faff3922e9fa9f76"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/e2e/tests/customers/customers.spec.ts
+++ b/packages/dashboard/e2e/tests/customers/customers.spec.ts
@@ -53,55 +53,77 @@ test('should show new history entries after updating the customer', async ({ pag
     await expect(page.getByText('Customer details updated').first()).toBeVisible();
 });
 
-// #5191 — the country dropdown on the customer address form must include every
-// enabled country, sorted by name. The query returned countries unsorted, and
-// Base UI's Select drops items whose order does not match its own collation, so
-// a country whose name sorted between two existing ones was missing from the
-// dropdown.
-test('address form country dropdown includes a newly created country', async ({ page }) => {
-    const client = new VendureAdminClient(page);
-    await client.login();
+// #5191 — the country dropdown on the customer address form.
+test.describe('Address form country dropdown', () => {
+    let countryId = '';
+    let customerId = '';
 
-    // A name that sorts in the middle of the list (between the seeded "Austria"
-    // and "Canada"), which is exactly where the unsorted-order bug dropped items.
-    const suffix = Date.now();
-    const countryName = `Belgium E2E ${suffix}`;
-    const countryCode = `QE${suffix.toString().slice(-4)}`;
+    // Runs even if the test body throws, so created entities never leak.
+    test.afterEach(async ({ page }) => {
+        if (!countryId && !customerId) return;
+        const client = new VendureAdminClient(page);
+        await client.login();
+        if (customerId) {
+            await client.gql(`mutation ($id: ID!) { deleteCustomer(id: $id) { result } }`, {
+                id: customerId,
+            });
+            customerId = '';
+        }
+        if (countryId) {
+            await client.gql(`mutation ($id: ID!) { deleteCountry(id: $id) { result } }`, {
+                id: countryId,
+            });
+            countryId = '';
+        }
+    });
 
-    const countryResult = await client.gql(
-        `mutation CreateCountryForAddressTest($input: CreateCountryInput!) {
-            createCountry(input: $input) { id }
-        }`,
-        {
-            input: {
-                code: countryCode,
-                enabled: true,
-                translations: [{ languageCode: 'en', name: countryName }],
+    // The dropdown must include every enabled country, sorted by name. The query
+    // returned countries unsorted, and Base UI's Select drops items whose order
+    // does not match its own collation, so a country whose name sorted between
+    // two existing ones was missing from the dropdown.
+    test('includes a newly created country in sorted order', async ({ page }) => {
+        const client = new VendureAdminClient(page);
+        await client.login();
+
+        // A name that sorts in the middle of the list (between the seeded "Austria"
+        // and "Canada"), which is exactly where the unsorted-order bug dropped items.
+        const suffix = Date.now();
+        const countryName = `Belgium E2E ${suffix}`;
+        const countryCode = `QE${suffix.toString().slice(-4)}`;
+
+        const countryResult = await client.gql(
+            `mutation CreateCountryForAddressTest($input: CreateCountryInput!) {
+                createCountry(input: $input) { id }
+            }`,
+            {
+                input: {
+                    code: countryCode,
+                    enabled: true,
+                    translations: [{ languageCode: 'en', name: countryName }],
+                },
             },
-        },
-    );
-    const countryId = countryResult.createCountry.id as string;
-    expect(countryId).toBeTruthy();
+        );
+        countryId = countryResult.createCountry.id as string;
+        expect(countryId).toBeTruthy();
 
-    const customerResult = await client.gql(
-        `mutation CreateCustomerForAddressTest($input: CreateCustomerInput!) {
-            createCustomer(input: $input) {
-                ... on Customer { id }
-                ... on ErrorResult { errorCode message }
-            }
-        }`,
-        {
-            input: {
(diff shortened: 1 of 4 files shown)


Ryrahul:
This pr needs some changes after #5190 is merged , will work and fix the conflicts after that merge takes place

michaelbromley (changes requested):
The change to the `value` prop is correct. The explanation attached to it is wrong, and that explanation is now a permanent comment in the test file. Replacing it is the main thing to do before merge.

## The stated root cause does not exist

The description and three comments in `packages/dashboard/e2e/tests/customers/customers.spec.ts` claim:

> Base UI's Select drops items whose order does not match its own collation, so a country whose name sorted between two existing ones was missing from the dropdown.

I searched `@base-ui/react/select/` and found nothing resembling collation. `SelectGroup` renders a `<div role="group">` with an `aria-labelledby`. It has no effect on which items register or render. Sorting a list also does not add or remove items from it.

A later reader will take that comment as fact. Please remove it.

## What actually fixes the pre-selection bug

```diff
-value={field.value || undefined}
+value={field.value ?? ''}
```

`SelectRoot` calls `useControlled()` at mount. That call decides once whether the component is controlled or uncontrolled. On the first render of the add-address form `field.value` is `''`, so `field.value || undefined` passed `undefined`, and the Select was uncontrolled from then on. `form.reset` later set `field.value` for an existing address, and `SelectRoot` ignored the new prop. Passing `field.value ?? ''` keeps the prop defined from the first render, so the Select stays controlled.

Please put that in the comment in place of the collation claim.

## The missing-countries symptom has no explanation here

`adminListQueryLimit` defaults to 1000 in `packages/core/src/config/default-config.ts:84`. No test or dev config overrides it, so the query returned every country. Sorting cannot restore an item the query already returned. Symptom 1 in issue 5191 has a cause this diff does not identify. Please reproduce it against the default limit before treating it as fixed.

## Test coverage does not match the description

The fix is applied in both files, but only the customer address form has an e2e test. `packages/dashboard/e2e/tests/settings/shipping-methods.spec.ts` already exists and is where a test for the `test-address-form.tsx` path belongs. Add one there, or drop the claim that both paths are covered.

## Hardcoded string

`test-address-form.tsx` has `<SelectValue placeholder="Select a country">`. `.claude/rules/dashboard.md` forbids hardcoded user-facing strings, and `customer-address-form.tsx` already uses ``t`Select a country` ``. The diff touches this element, so please wrap it.

## What is correct

- `sort: { name: ASC }` is valid syntax and `name` is a valid `CountrySortParameter` field.
- The tests extend `customers.spec.ts` rather than adding a new file, which follows the convention.
- The `VendureAdminClient` helper and `afterEach` cleanup match the rest of the suite.
- `value ?? ''` with `onValueChange={value => value && field.onChange(value)}` matches `zone-selector.tsx`.

Ryrahul:
@michaelbromley  Really sorry i am pretty sure i had closed this Pr before, maybe due to some issue on github side its still open. I have worked on this seperately will open another pr

PR #5196, opened 2026-08-27
fix: publish OrderEvent when customer is assigned to order

### LINKED ISSUE #5189: OrderService.addCustomerToOrder and mergeOrders publish no OrderEvent
**Describe the bug**

`OrderService.addCustomerToOrder` and `OrderService.mergeOrders` publish no event. A plugin cannot react when a guest order gets a customer, or when a guest cart is merged into the customer's existing cart.

The admin path publishes: `OrderService.updateOrderCustomer` calls `addCustomerToOrder` and then publishes `OrderEvent(ctx, order, 'updated', targetCustomer)` (`order.service.ts` line 576 on master). The shop paths do not:

- `setCustomerForOrder` mutation calls `addCustomerToOrder` (`shop-order.resolver.ts`).
- Login with an active guest order: `SessionService` calls `addCustomerToOrder`, and `mergeOrders` when the customer already has an active order.

Neither path publishes an `OrderEvent`. This is the same on 3.7.1 and on master.

**To Reproduce**

1. Subscribe to `OrderEvent` in a plugin and log `event.type`.
2. As a guest, add an item to the cart. `OrderEvent` with `type: 'created'` fires.
3. Call `setCustomerForOrder`, or log in as a customer with no active order.
4. No `OrderEvent` fires. `order.customer` is set.
5. Log in as a customer that has an active order. `mergeOrders` runs, the guest order is deleted, its lines move to the customer's order. No `OrderEvent` fires for either order.

**Expected behavior**

`addCustomerToOrder` publishes `OrderEvent(ctx, order, 'updated', input)`, as `updateOrderCustomer` does. `mergeOrders` publishes `OrderEvent(ctx, mergedOrder, 'updated')` and `OrderEvent(ctx, guestOrder, 'deleted')`.

**Actual behavior**

No event. A plugin that stores customer-derived data on the order (in our case a B2B company reference) has no hook between sign-in and checkout. It can set the value on the next `OrderLineEvent` or on the `ArrangingPayment` transition, so the order carries no company until then.

**Environment**

- @vendure/core version: 3.7.1 (also checked `packages/core/src/service/services/order.service.ts` on master, lines 2036 and 2146)
- Nodejs version: 24.11.0
- Database: postgres

Kathircpe:
closes #5189.

# Description

`addCustomerToOrder` now publishes `OrderEvent(ctx, order, 'updated', customer)` internally. This ensures all code paths — admin (`updateOrderCustomer`), shop (`setCustomerForOrder`), draft order (`setCustomerForDraftOrder`), and login-merge — emit the event when a customer is associated with an order.

The duplicate event publish in `updateOrderCustomer` was removed. The raw `save()` in `mergeOrders` was replaced with a call to `addCustomerToOrder`, which also gains coupon code revalidation.

# Note

# Breaking changes

The new code delegates to addCustomerToOrder, which also validates coupon codes (lines 2049-2059) and removes any that are invalid for the newly assigned customer. This means if a guest had coupon codes that don't apply to their registered account, they'll be silently stripped during merge.

This is arguably the correct behavior — but it's a behavioral change beyond the scope of "fire an OrderEvent."

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5196"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790422988&installation_model_id=20193&pr_number=5196&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5196&signature=6d4e67ad1161c2e8f15411aa9ae681ebb8bea70325ee3dbe066969e6957c0f47"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/e2e/order-merge.e2e-spec.ts
+++ b/packages/core/e2e/order-merge.e2e-spec.ts
@@ -1,9 +1,11 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
+    EventBus,
     mergeConfig,
     MergedOrderLine,
     MergeOrdersStrategy,
     Order,
+    OrderEvent,
     OrderMergeStrategy,
     Product,
     RequestContext,
@@ -454,5 +456,21 @@ describe('Order merging', () => {
         expect(result.lines[0].productVariant.id).toBe('T_1');
         expect(result.lines[0].quantity).toBe(3);
         expect(result.lines[0].customFields.relationField?.id).toBe('T_4');
+    });
+
+    // #5189 — login with an existing active cart and no guest cart must not publish OrderEvent
+    it('login with existing cart and no guest cart publishes zero OrderEvents', async () => {
+        const eventBus = server.app.get(EventBus);
+        const events: OrderEvent[] = [];
+        const subscription = eventBus.ofType(OrderEvent).subscribe(event => events.push(event));
+
+        // Log in as a customer who already has an active order, with no guest cart
+        await shopClient.asUserWithCredentials(customers[0].emailAddress, 'test');
+
+        // Allow any async event delivery to settle
+        await new Promise(resolve => setTimeout(resolve, 100));
+
+        expect(events).toHaveLength(0);
+        subscription.unsubscribe();
     });
 });
--- a/packages/core/e2e/order.e2e-spec.ts
+++ b/packages/core/e2e/order.e2e-spec.ts
@@ -12,12 +12,14 @@ import {
 import { omit } from '@vendure/common/lib/omit';
 import { pick } from '@vendure/common/lib/pick';
 import {
+    EventBus,
     OrderService,
     RequestContextService,
     defaultShippingCalculator,
     defaultShippingEligibilityChecker,
     manualFulfillmentHandler,
     mergeConfig,
+    OrderEvent,
 } from '@vendure/core';
 import {
     ErrorResultGuard,
@@ -2142,6 +2144,28 @@ describe('Orders resolver', () => {
                     type: HistoryEntryType.ORDER_CUSTOMER_UPDATED,
                 },
             ]);
+        });
+
+        it('setOrderCustomer publishes exactly one OrderEvent', async () => {
+            const eventBus = server.app.get(EventBus);
+            const events: OrderEvent[] = [];
+            const subscription = eventBus.ofType(OrderEvent).subscribe(event => events.push(event));
+
+            const { setOrderCustomer } = await adminClient.query(setOrderCustomerDocument, {
+                input: {
+                    orderId,
+                    customerId: customers[2].id,
+                    note: 'Event test',
+                },
+            });
+
+            // Allow any async event delivery to settle
+            await new Promise(resolve => setTimeout(resolve, 100));
+
+            expect(events).toHaveLength(1);
+            expect(events[0].type).toBe('updated');
+            expect(events[0].order.id).toBe(orderId);
+            subscription.unsubscribe();
         });
     });
 
--- a/packages/core/e2e/shop-order.e2e-spec.ts
+++ b/packages/core/e2e/shop-order.e2e-spec.ts
@@ -6,8 +6,10 @@ import {
     Asset,
     defaultShippingCalculator,
     defaultShippingEligibilityChecker,
+    EventBus,
     manualFulfillmentHandler,
     mergeConfig,
+    OrderEvent,
 } from '@vendure/core';
 import { createErrorResultGuard, createTestEnvironment, ErrorResultGuard } from '@vendure/testing';
 import { fail } from 'assert';
@@ -999,6 +1001,29 @@ describe('Shop orders', () => {
             expect(customer.lastName).toBe('Person');
             expect(customer.emailAddress).toBe('test@test.com');
             expect(customer.id).toBe(createdCustomerId);
+        });
+
+        it('setCustomerForOrder publishes exactly one OrderEvent', async () => {
+            const eventBus = server.app.get(EventBus);
+            const events: OrderEvent[] = [];
+            const subscription = eventBus.ofType(OrderEvent).subscribe(event => events.push(event));
+
+            const { setCustomerForOrder } = await shopClient.query(setCustomerDocument, {
+                input: {
+                    emailAddress: 'event-test@test.com',
+                    firstName: 'Event',
+                    lastName: 'Test',
+                },
+            });
+            orderResultGuard.assertSuccess(setCustomerForOrder);
+
+            // Allow any async event delivery to settle
+            await new Promise(resolve => setTimeout(resolve, 100));
+
+            expect(events).toHaveLength(1);
+            expect(events[0].type).toBe('updated');
+            expect(events[0].order.id).toBe(setCustomerForOrder.id);
+            subscription.unsubscribe();
         });
 
         describe('address handling', () => {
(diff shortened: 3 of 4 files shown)


Kathircpe:
order.service.ts:2221-2228 and 2240-2253

Not introduced by this PR, but clearly visible in the diff context and directly adjacent to the change. linesToDelete is processed twice:

2221: if (order && linesToDelete) {  // Block 1: removeItemFromOrder per line
2240: if (order && linesToDelete) {  // Block 2: removeItemsFromOrder batch
removeItemFromOrder is a thin wrapper around removeItemsFromOrder (line 950: return this.removeItemsFromOrder(ctx, orderId, [orderLineId])). So Block 1 calls the batch method N times (once per line), then Block 2 calls it once more with all N lines. The second pass would attempt to delete already-deleted lines. This is a genuine bug in the merge flow that predates this PR, but since the new addCustomerToOrder call lands right after it, the merge event may reflect an order state affected by double-processing. Worth a separate fix.

biggamesmallworld (changes requested):
Thanks for this. The core idea is right: making `addCustomerToOrder` the single place that publishes the event, and deleting the duplicate publish in `updateOrderCustomer`, is a better design than four call sites each remembering to publish. Two things need fixing before it can merge.

### 1. The merge path now publishes a spurious event on every login

`mergeOrders` runs on **every** `createNewAuthenticatedSession`, not only when a guest cart exists (`session.service.ts:120`). With no guest order, `OrderMerger.merge` falls through to the final branch and returns `{ order: existingOrder }` (`order-merger.ts:51-55`). There is then nothing to delete, insert, modify or remove, so execution goes straight to:

```ts
order = await this.addCustomerToOrder(txCtx, order, customer);
```

The order already belongs to that customer. Nothing changes, but an `OrderEvent(ctx, order, 'updated', customer)` is now published.

The problem is not the volume, it is that the event is untrue. `OrderEvent`'s `input` is typed `Customer | ModifyOrderInput | { customFields: any }` (`order-event.ts:7`), so passing a `Customer` is precisely the signal subscribers use to detect "a customer was just assigned to this order". Here, none was. Note also that this login path currently publishes zero `OrderEvent`s, so this is a new event where there was none, not one more among many.

`updateOrderCustomer` already guards this at line 557:

```ts
if (currentCustomer?.id === customerId) {
    // No change in customer, so just return the order as-is
    return order;
}
```

Now that `addCustomerToOrder` owns the publish, it needs to own that guard too:

```ts
if (idsAreEqual(order.customer?.id, customer.id)) {
    return order;
}
```

One caveat on the relation load: `getOrderOrThrow`'s default relations do include `customer`, but the `Order`-instance overload of `addCustomerToOrder` takes whatever the caller passes, so `order.customer` may be undefined simply because it was never hydrated. Compare against the `customerId` FK rather than letting "not loaded" read as "changed".

### 2. Decide on the coupon revalidation

Your own description has a "Breaking changes" section noting that guest coupon codes are now silently stripped during merge, and concluding it is "a behavioral change beyond the scope of 'fire an OrderEvent'". I agree with that assessment, so it should not ship as an unremarked side effect. Either drop it or commit to it with a test and a changeset entry.

There is a smaller robustness point behind it. The old code was `order.customer = customer` followed by `save(order, { reload: false })`, which is about as close to unthrowable as a write gets. The new path can reach `removeCouponCode`, which calls `applyPriceAdjustments`, and a custom `PromotionCondition` or `PromotionAction` can throw from there. That lands in the catch at line 2276, which rolls the merge transaction back and returns `existingOrder`. Nothing is destroyed, but if `existingOrder` was empty or undefined the customer signs in to an empty cart while their guest order sits orphaned in the database, with only a server log to explain it. Narrow, but it is a failure mode the previous code did not have.

Happily, the guard in point 1 fixes both: returning early when the customer is unchanged skips the coupon loop entirely on the login path.

### 3. Tests

The test checkbox is unticked on a PR whose entire observable output is an event, and the linked issue includes a reproduction. `packages/core/e2e/shop-order.e2e-spec.ts` and `packages/core/e2e/customer.e2e-spec.ts` already cover these mutations. Please assert exact counts rather than "at least one", since the count is the whole point of removing the duplicate publish:

- `setCustomerForOrder` publishes exactly one `OrderEvent` with `type: 'updated'` and the correct customer.
- `updateOrderCustomer` still publishes exactly one, not zero and not two. This is the regression the PR is most likely to introduce.
- Logging in with an existing active cart and no guest cart publishes zero `OrderEvent`s. This one fails against the current diff.

### 4. Docs

`addCustomerToOrder` is a public, documented service method whose doc comment is still the one line "Associates a Customer with the Order." It now publishes an event that plugin authors are expected to subscribe to. Please say so, with an `@since` for the release this lands in, matching how `updateOrderCustomer` documents `@since 2.2.0`.

Kathircpe:
done

biggamesmallworld (changes requested):
Not ready yet. CI is red on all four databases, and two points from the last round are still open.

### Test failures

Three e2e failures, reproducible on sqljs, postgres, mysql and mariadb:

**`order.e2e-spec.ts:2165` — "setOrderCustomer publishes exactly one OrderEvent"**, got zero. The test assigns `customers[2]`, but the "update order customer" test 30 lines above already assigned that same customer. `updateOrderCustomer` hits its own early return at `order.service.ts:557` and never reaches `addCustomerToOrder`. Use a customer that is not already on the order.

**`shop-order.e2e-spec.ts:1025`** — `expected 1 to be 'T_1'`. `events[0].order.id` is the raw entity id; the assertion compares it to a GraphQL-encoded id. Compare against the decoded id.

**`shop-order.e2e-spec.ts` — "customer default Addresses are updated after payment"** now fails. It passed before this PR. The new test was inserted into the middle of the `ordering as anonymous user` block and mutates the shared active order that the later test depends on. Move it to the end of the block, or give it its own order.

### 1. The guard

The guard is in, but it reads `order.customer?.id`. That is the case I flagged last time: `addCustomerToOrder`'s `Order`-instance overload takes whatever the caller passes, so an unhydrated `customer` relation reads as "changed". This is live in `mergeOrders`, where `order` is reassigned by `addItemsToOrder`, `adjustOrderLines` and `removeItemsFromOrder` before it reaches the call. `Order.customerId` exists (`order.entity.ts:99`); compare against that:

```ts
if (idsAreEqual(order.customerId, customer.id)) {
    return order;
}
```

### 2. Coupon revalidation

Still unresolved. `mergeOrders` continues to delegate to `addCustomerToOrder`, so when the guest order is the one retained, its coupon codes are revalidated against the newly assigned customer and silently stripped. Your own PR description calls this "a behavioral change beyond the scope of 'fire an OrderEvent'", and I agree. Either avoid the coupon path on merge, or keep it with a test covering the stripping. Right now it ships unremarked.

### 3. Small one

`@since 3.7.2` is wrong. 3.7.2 and 3.7.3 are both published, so this lands in 3.7.4.

Issue #3409, opened 2025-03-11
Timestamps persisted incorrectly to Postgres despite DB timezone is set to UTC


gabriellbui:
**Describe the bug**
This is basically a duplicate of https://github.com/typeorm/typeorm/issues/9627. It's mainly here to track the timestamp issue within Vendure as well.

In TypeORM columns with `timestamp` or `timestamp without time zone` type is read as local date, instead of UTC, causing the timezone offset to be lost.

This happens with automatically created values as well (e.g., `@CreateDateColumn()`). In Vendure, the base `VendureEntity` applies the default `@CreateDateColumn()` to `createdAt` and `@UpdateDateColumn()` to `updatedAt`, causing them to persist with incorrect timestamps.

**To Reproduce**
Follow the "To Reproduce" section in this issue: [TypeORM Issue #9627](https://github.com/typeorm/typeorm/issues/9627)

**Expected behavior**
Entities extending the base VendureEntity should have the correct `createdAt` and `updatedAt` timestamps

**Environment (please complete the following information):**

Dependency | Version
-- | --
Operating System | macOS sequoia 15.3.1
Node.js version | 22.14.0
Typescript version | 5.2.2
TypeORM version | 0.3.20
PostgreSQL version | 15.4

**Related issues**
https://github.com/typeorm/typeorm/issues/9627
https://github.com/typeorm/typeorm/issues/2220
https://github.com/typeorm/typeorm/issues/8296
https://github.com/typeorm/typeorm/issues/5841
https://github.com/typeorm/typeorm/issues/7184
https://github.com/vendure-ecommerce/vendure/issues/1375

**Suggested solution**
Use the `timestamptz` type which is an abbreviation for type `timestamp with time zone`. It should internally store them in UTC, but converted to the session's time zone when queried.

```
  @CreateDateColumn({ type: "timestamptz" })
  createdAt: Date;

  @UpdateDateColumn({ type: 'timestamptz' })
  updatedAt: Date;
```

An alternative is to configure the driver to interpret timestamps as UTC inside your project

```
pg.defaults.parseInputDatesAsUTC = true

const dateParser = pg.types.getTypeParser(pg.types.builtins.TIMESTAMP)
pg.types.setTypeParser(pg.types.builtins.TIMESTAMP, (val: string) =>
  dateParser(`${val}Z`),
)
```

DanielBiegler:
Have you tried setting your .env variable to:

```
TZ=UTC
```

[...]

PR #5203, opened 2026-08-28
feat(core): add timezone option for scheduled tasks

### LINKED ISSUE #5202: Scheduled tasks have no timezone option; cron expressions are evaluated in the process timezone
> [!NOTE]
> **Authorship:** This report was researched and written by Claude AI Fable 5 working under my direction, and reviewed by me before posting. Every source reference below was verified against current `master` before posting.

**Describe the bug**

`@vendure/core` constructs `croner` `Cron` instances in two places and never passes a `timezone` option. Without it, croner evaluates the cron expression in the **process** timezone. There is no timezone option on `SchedulerOptions`, `ScheduledTaskConfig` or `DefaultSchedulerPluginOptions`, and `SchedulerService.createCronJob` is private, so there is no userland workaround other than setting `TZ` on the process.

I am filing this as a bug rather than a feature request because that workaround conflicts with the accepted resolution of #3409. `VendureEntity.createdAt` / `updatedAt` use `@CreateDateColumn()`, which maps to `TIMESTAMP` (without time zone) on Postgres, and node-postgres both writes and reads that column in process-local wall-clock time. #3409 (closed 2025-03-13) resolved that problem with the advice "set `TZ=UTC`". The only way to get local wall-clock scheduling today is `TZ=Europe/Stockholm`. The two pieces of advice are mutually exclusive: you can have correct naive timestamps, or local wall-clock schedules, but not both.

The docs make the gap invisible: the scheduled-tasks guide's headline examples are `cron.everyDayAt(3, 0)` and `cron.everyDayAt(0, 0)`, and neither the guide nor the API reference mentions timezones or UTC anywhere. "Every day at 03:00" silently means "03:00 in whatever timezone the worker container happens to have", and drifts by an hour twice a year across DST transitions.

**To Reproduce**

1. Run the server with `TZ=UTC` (as advised in #3409).
2. Register a scheduled task intended to run at 02:00 local time in a non-UTC zone:

   ```ts
   new ScheduledTask({
       id: 'nightly-sync',
       schedule: '0 2 * * *',
       async execute() { /* ... */ },
   });
   ```

3. Observe the resolved next run time.

**Expected behavior**

Some supported way to state that `0 2 * * *` means 02:00 in a given IANA zone, independently of the process timezone.

**Actual behavior**

The expression is evaluated in the process timezone (UTC), so the task runs at 03:00 or 04:00 local depending on DST. Setting `TZ=Europe/Stockholm` fixes the schedule but reintroduces the naive-timestamp problem from #3409.

**Source references**

Both call sites, verified on `master` @ 4631092 (2026-08-27); identical in v3.7.2:

- `packages/core/src/scheduler/scheduler.service.ts` — `createCronJob()`, lines 168–173: `new Cron(schedule, { name: task.id, protect: ... }, ...)`. No `timezone`.
- `packages/core/src/plugin/default-scheduler-plugin/stale-task.service.ts` — `getScheduleIntervalMs()`, line 73: `new Cron(scheduleString)`. No options at all.

The second call site matters independently: `getScheduleIntervalMs` feeds both stale-lock detection and `DefaultSchedulerStrategy.computeLockHoldMs`, and the value is cached for the lifetime of the process. If only one of the two call sites became timezone-aware, the computed interval and the actual job cadence would drift apart — for daily patterns the drift shows up across DST transitions (a 23h or 25h day against an assumed 24h), for windowed expressions it is larger. This is why the fix should resolve the timezone through one shared helper used by both.

Worth noting: croner (`^10.0.1`, already a dependency of `@vendure/core`) exposes `timezone?: string` on `CronOptions`. The fix is a pass-through of an option the dependency already supports — no new dependency, no new scheduling logic.

**Proposed fix**

Additive and opt-in:

1. Add `timezone?: string` (IANA identifier) to `SchedulerOptions` as a global default.
2. Add `timezone?: string` to `ScheduledTaskConfig` as a per-task override.
3. Resolve the effective value through a **single shared helper** (per-task → global → `undefined`) used by **both** call sites, so the scheduler and the stale-task interval calculation cannot diverge.
4. Validate the identifier at bootstrap with a clear error naming the task, rather than letting croner throw a bare `RangeError` on first schedule.

The default remains `undefined`, which preserves today's process-local behaviour, so this is not a breaking change.

**I have a working implementation of exactly this (including tests and docs) and will open a PR against `minor` right after filing this issue.** A functionally equivalent patch — the same timezone pass-through at both call sites — has been running in a production 3.7.1 store since late July, with the process in UTC and schedules evaluated in `Europe/Stockholm`, without issues. One open question: given #3451 and discussion #4321 (adding `timezone` to User / Channel), you may prefer a different name here to avoid future ambiguity — `cronTimezone` or similar. Happy to rename in the PR to whatever you prefer.

**Related (not duplicates)**

- #1719 — `timestamp without time zone` for entity timestamps (open)
- #3409 — closed (2025-03-13) with the `TZ=UTC` recommendation that this issue conflicts with
- #3451 / discussion #4321 — `timezone` on User / Channel; relevant to naming
- #3804 (merged 2025-09) — added the second `new Cron` call site (as `TaskService`; moved and renamed to `StaleTaskService` in a follow-up refactor)
- #5165 / #5166 — recent scheduler defects (lock lifetime) filed by another user; listed as related context in the same subsystem, not duplicates

- @vendure/core version: 3.7.2 (also verified on `master` @ 4631092, 2026-08-27)
- Nodejs version: 24.x
- Operating System (Windows/macOS/Linux): Linux (Docker) / Windows

comega-johan:
> [!NOTE]
> **Authorship:** This PR was implemented by Claude AI Fable 5 working under my direction, and reviewed by me before submission (the commit carries a `Co-Authored-By` trailer).

> [!NOTE]
> **Maintainer note (@biggamesmallworld):** the commits after the first are mine. They apply my review feedback, a second review pass, and a rebase onto `minor`. The description below covers the branch as it now stands.

Fixes #5202.

Cron schedules for scheduled tasks are evaluated in the process timezone, and the only way to influence that is the `TZ` environment variable, which conflicts with the `TZ=UTC` resolution of #3409 for correct naive-timestamp persistence. This PR adds an opt-in IANA timezone option:

- **`SchedulerOptions.timezone?: string`** — default for all tasks.
- **`ScheduledTaskConfig.timezone?: string`** — per-task override, also settable via `task.configure({ timezone })` for built-in tasks such as `cleanSessionsTask`.
- **One shared helper** (`getScheduleTimezone()`) resolves per-task → global → `undefined`, and is used by **both** `Cron` construction sites, `SchedulerService.createCronJob()` and `StaleTaskService.getScheduleIntervalMs()`, so the job cadence and the stale-lock interval cannot diverge.
- **Blank values are treated as unset.** An empty or whitespace-only string, such as a defined-but-empty env var, falls through to the next level. Surrounding whitespace is trimmed off the resolved value.
- **Bootstrap validation.** `schedulerOptions.timezone` and each task's own `timezone` are validated separately, so the error names the option or the task that carries the bad value:
  `Invalid timezone "Europe/Stockholmm" configured for the \`schedulerOptions.timezone\` option.`
  The check runs before the scheduler-strategy check, so it also fires under `NoopSchedulerStrategy` and when no tasks are configured.
- **`ScheduledTask.timezone`** is a nullable field on the GraphQL type and on `TaskInfo`. The dashboard and the Admin UI both select it and render it next to the schedule description. `scheduleDescription` stays the plain cronstrue sentence.
- **`StaleTaskService` no longer memoises the schedule interval.** Two consecutive runs are 23 or 25 hours apart around a DST transition, so a cached value was wrong for the rest of the process lifetime. It is computed from the next two runs on every call.
- **Docs:** new "Timezones" section in the scheduled-tasks guide, plus regenerated reference docs.

No new dependency: croner `^10.0.1` (already a dependency) supports `CronOptions.timezone`; the change is a pass-through.

`RuntimeVendureConfig.schedulerOptions` uses the `Required<Omit<..., 'timezone'>> & ...` pattern already established for `entityOptions.entityIdStrategy`, since `timezone` has no default value.

## Verification

- `packages/core` unit suite: 1553 passing, of which 26 are new.
- The new specs also pass under `UTC`, `Europe/Stockholm`, `America/Chicago` and `Pacific/Chatham`, so they do not depend on the process timezone.
- `default-scheduler-plugin` e2e against sqljs, which exercises the `StaleTaskService` constructor through real DI and asserts the new GraphQL field.
- Full monorepo build, and `tsc` over the dashboard.

Beyond the test suite: a functionally equivalent patch of the same two call sites (applied via pnpm patch against 3.7.1) has been running in a production store since late July, process in UTC and schedules evaluated in `Europe/Stockholm`.

# Breaking changes

None at runtime. The default is `undefined`, which preserves process-local evaluation exactly.

Two type-level notes:

- `RuntimeVendureConfig['schedulerOptions']` (and the `ConfigService#schedulerOptions` getter) is no longer assignable to `Required<SchedulerOptions>`, since `timezone` stays optional. Reads and writes of all existing members are unaffected.
- `TaskInfo` gains a required `timezone: string | null` member. Only `SchedulerService` constructs a `TaskInfo`, so this affects code which builds one by hand.

# Screenshots

N/A. The dashboard and Admin UI task lists gain a timezone in parentheses after the schedule description, and show nothing extra when no timezone is configured.

# Checklist

📌 Always:

👍 Most of the time:


biggamesmallworld (dismissed):
Thanks for this, and for the unusually thorough PR description. The core design is right: resolving the timezone through a single helper used by both `SchedulerService.createCronJob()` and `StaleTaskService.getScheduleIntervalMs()` is exactly what prevents the job cadence and the stale-lock interval from disagreeing. The `Required<Omit<SchedulerOptions, 'timezone'>> & SchedulerOptions` shape follows the existing `entityOptions` precedent rather than inventing a new one, and the tests assert exact instants instead of just checking for definedness. I ran the new specs locally: 19 passing.

Two things need to change before this can merge, plus a handful of smaller items.

## Blocking

**1. The docs file has an unrelated Prettier pass that breaks the Shiki highlight markers.**

`docs/docs/guides/developer-guide/scheduled-tasks/index.mdx` picked up a wholesale reformat, and Prettier moved several `// [!code highlight]` comments onto their own lines:

```diff
-            generateSitemapTask.configure({ // [!code highlight]
-                params: { // [!code highlight]
+            generateSitemapTask.configure({
+                // [!code highlight]
+                params: {
+                    // [!code highlight]
```

The marker highlights the line it sits on and is then stripped, so `generateSitemapTask.configure({` loses its highlight and the reader gets a highlighted blank line instead. Same damage at `config.schedulerOptions.tasks.push(`, `SitemapPlugin.init({`, and `if (this.processContext.isWorker) {`.

Please revert every hunk in that file that is not the new `## Timezones` section, including the frontmatter quote style and the whitespace-only changes.

**2. The timezone should be a field, not a suffix on `scheduleDescription`.**

```ts
scheduleDescription: pattern ? `${cronstrue.toString(pattern)}${timezoneSuffix}` : 'unknown',
```

`scheduleDescription` is a GraphQL field consumed by the dashboard table. Baking the timezone into it means any client that wants the value has to regex a parenthetical out of a cronstrue sentence. Please add `timezone: String` to the `ScheduledTask` GraphQL type and to `TaskInfo`, and leave `scheduleDescription` as the cron description alone. The UI can render the two together.

This also affects the existing assertion in `packages/core/e2e/default-scheduler-plugin.e2e-spec.ts:86`, which compares `scheduleDescription` to an exact string. It passes today only because no e2e config sets a timezone.

I realise your PR description lists this as a deliberate follow-up. I would rather have the field now than ship a string format that clients start parsing.

## Should fix

**Return the trimmed value.** In `getScheduleTimezone`:

```ts
const timezone = taskTimezone?.trim() ? taskTimezone : schedulerOptions.timezone;
return timezone?.trim() ? timezone : undefined;
```

The stated motivation for trimming is an environment variable carrying stray whitespace, but `' Europe/Stockholm '` passes both guards and is returned with the spaces intact. `Intl` then rejects it (`Invalid time zone specified:  Europe/Stockholm `) and bootstrap fails. It fails loudly rather than silently, so this is not a blocker, but it is the case that actually bites. Worth a spec case too: the current tests cover `''` and `'   '` only.

**Validation is unreachable under the Noop strategy.** `assertValidTimezone` is only called from `createCronJob`, and `onApplicationBootstrap` returns before the task loop when no scheduler strategy is configured. Low impact, since tasks do not run in that setup anyway, but the commit message says the identifier is validated at bootstrap and that is only conditionally true. Either move the validation ahead of the strategy check, or soften the claim.

**The memoised interval is now DST-dependent.** `getScheduleIntervalMs` caches into `taskIntervalMap` with no invalidation. Your own test shows a 25 hour interval computed near the Stockholm fall-back; compute it near spring-forward instead and a 23 hour interval is cached for the process lifetime, shortening the stale-lock threshold. It only affects tasks whose interval exceeds 23 hours, and it is arguably pre-existing for anyone running a non-UTC `TZ`, but this PR makes it reachable for UTC processes. Either cache the larger of the two gaps, or add a comment explaining why the drift is acceptable.

**Lint.** The four new files produce 6 `import/order` warnings. Please clear them.

**No e2e coverage.** The unit tests cover the resolution logic well, but nothing exercises the timezone through real DI wiring, where `StaleTaskService` just gained a `ConfigService` constructor parameter.

**Trim the comments.** The docblock on `getScheduleTimezone` reproduces the rationale from the commit message, which is where it belongs; one line saying the task timezone wins over the global one is enough. Three sentences narrate the change rather than describe the code, and should go: `"preserving the behaviour of versions prior to the introduction of this option"`, `"which preserves the behaviour of prior versions"`, and `"When no timezone is configured, the behaviour is unchanged from prior versions."` A reader arriving at these docs never saw the previous version.

## Minor

- `if (timezone != null)` before `assertValidTimezone`: the helper returns `string | undefined`, so `null` cannot occur. A truthiness check is clearer.
- `catch (e: any)` binds `e` and never uses it. Use `catch {`.
- `assertValidTimezone` validates against `Intl`, but croner is what has to accept the value. `Intl` accepts offset forms such as `'+05:00'`; if croner rejects those, you get the raw `TypeError` this function exists to avoid. Either construct the `Cron` inside the try, or document that only IANA names are supported.
- Wrapping the `configure()` signature across three lines leaked into the generated reference docs as `Partial<             Pick<...`. Keep it on one line.

On naming: `timezone` is fine, no need to rename.

biggamesmallworld (approved):
Approving. Every point from my earlier review is addressed, on the original commit or on the five I pushed on top.

Since that review:

- The timezone is a `ScheduledTask.timezone` GraphQL field and a `TaskInfo` member, not a suffix on `scheduleDescription`. The dashboard and the Admin UI both select it and render it after the schedule description.
- `schedulerOptions.timezone` and each task's own `timezone` are validated separately, so the error names the option or the task that holds the bad value. The check runs before the scheduler-strategy check, so it fires under `NoopSchedulerStrategy` and when no tasks are configured.
- `getScheduleTimezone()` returns the trimmed value, so `' Europe/Stockholm '` resolves.
- `StaleTaskService` no longer caches the schedule interval. Two consecutive runs are 23 or 25 hours apart around a DST transition, so the cached value was wrong for the rest of the process lifetime.
- The Prettier pass over the scheduled-tasks guide is reverted, so the Shiki highlight markers are intact.
- The e2e suite covers the new field through real DI, and the import/order warnings are gone.

Disclosure: commits 2 to 6 are mine, written with Claude under my direction, so this approval covers @comega-johan's original commit plus my own follow-up work rather than an independent review of all six. The branch has been rebased onto `minor`. CI is green on the full matrix: build, unit tests on 20.x, 22.x and 24.x, e2e on sqljs, postgres, mysql and mariadb, dashboard e2e and lint, codegen and generated-docs validation.

Thanks for a well-argued PR, and for the production evidence behind it.

PR #5208, opened 2026-08-31
fix(core): evaluate SQLite regex filters with re2js instead of native re2

HouseinIsProgramming:
Replaces the native `re2` dependency added in #4956 with the pure-JavaScript `re2js`, as a plain dependency of `@vendure/core`.

## Why

`re2` is a native module. It compiles at install time, needs `node-gyp`, and needs a build-script allowlist entry for pnpm, yarn and bun. It sat in `optionalDependencies`, so a failed build was not an install error. The package was absent and `regex` filters fell back to the backtracking engine.

`re2js` is the same RE2 algorithm written in JavaScript. It has no build step, so it cannot fail to install, and no package manager needs an allowlist entry for it.

## What changed

- `@vendure/core` depends on `re2js`. `re2` and its `optionalDependencies` entry are gone.
- `sqlite-regexp-function.ts` adapts `re2js` to the `RegExp` shape the rest of the file uses. The fallback to the built-in engine and the start-up warning are removed.
- The engine is resolved on first use, not at import. `parse-filter-params.ts` imports this module and sits on the path of every list query on every backend, so a top-level import would make a Postgres or MySQL deployment parse 246 kB of engine it can never reach. Both entry points already return early for non-SQLite.
- `@vendure/create` no longer adds `re2js` to a SQLite scaffold, because every project already has it.
- The root `trustedDependencies` entry for `re2` is removed.

Dropping `re2` also drops `node-gyp`, `nan` and `install-artifact-from-github` from the lockfile. `re2js` has no dependencies of its own, and adds 868 kB.

## Upgrade step: a breaking change on SQLite and SQL.js

`re2` never shipped. #4956 merged to `minor` and its merge commit is in no tag, so 3.7.2 evaluates SQLite `regex` filters with the built-in `RegExp`, where lookaround and backreferences work.

RE2 supports neither. From this release a `regex` filter on `better-sqlite3` or `sqljs` which uses lookaround or a backreference returns a `UserInputError` instead of a result, rejected up-front by `assertRegexFilterEngineCompatible` rather than failing part-way through the query. Such filters have to be rewritten without those constructs. The changelog says so.

Nothing else is needed. Package managers install a dependency, so every SQLite project gets the bounded engine on upgrade. Other databases evaluate the pattern themselves and are unaffected.

## Review points from @michaelbromley

The first revision made `re2js` an optional peer dependency, which package managers never install on their own. The fix therefore depended on the user reading the changelog and adding a package, and the docs then had to explain that gap. Both review points follow from the optional peer, so this revision drops it.

**"I thought that it is solved for sqlite now with this PR?"** Yes. `re2js` installs with `@vendure/core`, so SQLite and SQL.js match with RE2 on every project, with nothing to add and no fallback left to hit.

**"This warning is overly alarming."** The warning admonition is gone, and so is the advice to expose `regex` behind a permission. That advice was not actionable: `regex` is defined on `StringOperators` in the common schema, so it is on every `PaginatedList` query on both APIs and cannot be turned off. What is left is three paragraphs: prefer `contains`, prefer PostgreSQL or SQLite if you do expose `regex` filters, and the note that RE2 rejects lookaround and backreferences.

## Verification

| Check | Result |
|---|---|
| `packages/core` build (`tsc`) | clean |
| `packages/core` full unit suite | 1444 passed, 101 files |
| `list-query-builder` unit specs | 119 passed |
| `list-query-builder` e2e on sqljs | 112 passed, 1 skipped |
| `packages/create` `helpers.spec.ts` | 57 passed |
| eslint on the changed files | clean |

Checked against the built `dist`, not the TypeScript source: the emitted CJS resolves `re2js` through its `require` export condition, and the engine stays unloaded until a SQLite filter needs it.

```
after importing the module:      re2js loaded = false
after a Postgres regex filter:   re2js loaded = false
after a SQLite regex filter:     re2js loaded = true
```

### Before and after

The pattern is `^(.|..)+$Z` from GHSA-jgm3-qmp2-c4p7. The trailing `Z` after `$` makes every match attempt fail, so the backtracking engine tries every partition of the input into 1- and 2-character pieces. "Before" is the built-in `RegExp`, which is what a SQLite project on 3.7.2 gets. "After" is the function this branch registers as SQLite's `REGEXP`.

```
chars | before (built-in RegExp) | after (RE2 via re2js)
------|--------------------------|----------------------
   24 |                   3.4 ms |               2.1 ms
   28 |                   4.0 ms |               0.1 ms
   32 |                  27.1 ms |               0.0 ms
   36 |                 190.5 ms |               0.1 ms
   40 |                    1.2 s |               0.1 ms
   44 |                    8.2 s |               0.1 ms
   48 |                   57.0 s |               0.1 ms
```

Every four characters multiply the built-in engine's work by about six. The event loop is blocked for the whole 57 seconds of the last row, because SQLite calls this function synchronously. RE2 stays at 0.1 ms. `sqlite-regexp-function.spec.ts` asserts the "after" column on every CI run.

### Pre-existing CI failure

`test_package_managers (yarn)` fails on this branch. It also fails on `minor` and on `master`, for example [run 33381604885](https://github.com/vendurehq/vendure/actions/runs/33381604885). It is a yarn PnP and ESM loader error inside `@vendure/create` itself, `request for 'node:process' is from a module not been linked`, and is unrelated to this change.

### Generated docs drift

`be12cac0a` (the docgen bot) carries reference pages for source files this PR does not touch: `bootstrap.mdx`, a new `deduplicate-translations.mdx`, `deserialized-cached-session.mdx`, and five others. That is committed drift between `minor` and what the generator produces, not a change from this branch. Left alone here, since the bot regenerates and re-commits it on every push.


docs/docs/guides/how-to/paginated-list/index.mdx
```diff
@@ -267,6 +267,39 @@ query {
...
+single long string, and MariaDB pays that cost for every row it reads. A filter over a large table
+holds a database connection for seconds. Enough concurrent regex queries take every connection in
+the pool, and other requests wait for one to free up.
+
+SQLite and SQL.js match the pattern in the Node.js process. A slow pattern blocks the event loop, so
+the server answers nothing else until the match finishes. Install the `re2js` package on those two
+databases. Vendure then matches with the RE2 engine, which limits what any one pattern can cost.
+
+:::warning
+Do not let untrusted input reach a `regex` filter on MariaDB, SQLite or SQL.js. A pattern from a
```
michaelbromley: I feel this is not easily actionable. Every paginated list query allows regex filters, so how are you supposed to prevent usage of these? It raises the question "why expose a dangerous feature in so many places?"

And I thought that it is solved for sqlite now with this PR? And for MariaDB it is merely slow, and not event-loop blocking. So this warning is overly alarming IMO.

michaelbromley (changes requested):

PR #5226, opened 2026-09-01
fix(dashboard): remove arbitrary 1000 max on promotion usage limits

### LINKED ISSUE #5225: Promotion usageLimit UI capped at 1000 but not enforced by backend; blocks always-on promotions
**Describe the bug**
The Promotion detail page in the Dashboard caps the `usageLimit` and `perCustomerUsageLimit` fields at a maximum of `1000` via a client-side-only HTML `max` attribute. The backend (`@vendure/core`) does not enforce any such maximum, so the two are inconsistent. The `1000` limit is not a real business rule — it can be trivially bypassed via the Admin API (or by typing a larger value directly into the input, since the HTML `max` attribute only constrains the spinner arrows, not typed input).

**Real-world impact**
This `1000` cap is a genuine operational blocker for common, high-volume promotions. For example, a store-wide promotion like "Free shipping over Rs 500" is auto-applied to nearly every qualifying order, so its usage count reaches 1000 very quickly. Once the limit is hit the promotion stops applying, and store operators are forced to manually create a brand-new promotion every time — repeatedly, and often mid-campaign. For an always-on promotion, capping usage at 1000 does not make sense at all, and there is no UI way to set a higher value.

**To Reproduce**

1. Go to the Dashboard > Promotions > (any promotion) detail page
2. Locate the "Usage limit" / "Per customer usage limit" fields
3. Observe the spinner is capped at 1000 (`max={1000}`)
4. Alternatively, call the `createPromotion`/`updatePromotion` Admin API mutation with `usageLimit: 50000` — it is accepted, persisted, and enforced without error, contradicting the UI cap

**Expected behavior**
The UI and backend should agree. Either:
- The `1000` cap should be removed/raised in the UI if it is not an intended constraint (this is the likely correct fix — an always-on promotion should support a high or unlimited usage count), OR
- The backend should enforce the same limit (return an error result / throw `UserInputError`) so the constraint cannot be bypassed via the API.

**Actual behavior**
The `max={1000}` exists only in the Dashboard UI. The backend accepts and enforces any `Int` value (up to the DB signed 32-bit integer ceiling, ~2.1 billion). No validation exists in the GraphQL schema, the `Promotion` entity, or `PromotionService.createPromotion` / `updatePromotion`.

**Screenshots/Videos**
N/A

**Error logs**
N/A (no error is produced — that is part of the problem; the API silently accepts values > 1000)

```
No error logs
```

- @vendure/core version: (current master / dev)
- Nodejs version:
- Browser (if applicable): any
- Package manager (npm/yarn/pnpm):

**Configuration**
N/A — reproducible with default configuration.

```typescript
// Default Vendure configuration
```

**Workaround**
Set the higher `usageLimit` directly via the Admin API `updatePromotion` mutation (the backend accepts values > 1000), bypassing the Dashboard cap. Leaving the field empty/null also means no limit is enforced at all.

**Additional context**
- The `1000` value appears arbitrary and undocumented.
- The HTML `max` attribute on a number input does not block typed values, so even in the UI a user may be able to enter more than 1000 depending on the form's Zod/react-hook-form validation (which does not appear to mirror this cap).
- Question for maintainers: is `1000` an intended business constraint, or a leftover UI guardrail? Given that always-on promotions (e.g. free shipping thresholds) routinely exceed 1000 uses, the cap seems unintended and the fix direction (remove/raise the UI cap vs. add backend validation) depends on your answer.

Ryrahul:
# Description

Removes the arbitrary `max={1000}` cap on the `usageLimit` and `perCustomerUsageLimit` fields in the Promotion detail page.

This cap existed only as a client-side HTML attribute in the Dashboard. The backend (`@vendure/core`) never enforced it — both fields are plain nullable `Int` columns, and a `null` value means "unlimited". This made the UI and backend inconsistent, and the limit could already be bypassed via the Admin API.

The 1000 cap was a real operational blocker for common always-on promotions (e.g. "free shipping over a threshold") which are auto-applied to most orders and reach 1000 uses quickly, forcing operators to recreate the promotion repeatedly. Both fields now keep `min={0}`; leaving them empty continues to mean unlimited.

Fixes #5225

# Breaking changes

None. This only removes a client-side input constraint; no API, schema, or persisted behavior changes.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5226"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790838110&installation_model_id=20193&pr_number=5226&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5226&signature=939b9bdc7c01f4769572c76f10bef82923479682ee5db02a57437157e8b85d20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/app/routes/_authenticated/_promotions/promotions_.$id.tsx
+++ b/packages/dashboard/src/app/routes/_authenticated/_promotions/promotions_.$id.tsx
@@ -211,7 +211,7 @@ function PromotionDetailPage() {
                             name="perCustomerUsageLimit"
                             label={<Trans>Per customer usage limit</Trans>}
                             render={({ field }) => (
-                                <NumberInput {...field} value={field.value ?? ''} min={0} max={1000} />
+                                <NumberInput {...field} value={field.value ?? ''} min={0} />
                             )}
                         />
                         <FormFieldWrapper
@@ -219,7 +219,7 @@ function PromotionDetailPage() {
                             name="usageLimit"
                             label={<Trans>Usage limit</Trans>}
                             render={({ field }) => (
-                                <NumberInput {...field} value={field.value ?? ''} min={0} max={1000} />
+                                <NumberInput {...field} value={field.value ?? ''} min={0} />
                             )}
                         />
                     </DetailFormGrid>


Ryrahul:
Happy to add a `max` limit back if the maintainers feel a hard cap is desirable — just let me know what value would be suitable.

My reasoning for removing it entirely: the backend stores these as nullable 32-bit `Int` columns (with `null` meaning unlimited), so the only technically meaningful ceiling is the DB integer max (`2147483647`). If you'd prefer a guardrail, I'd suggest either capping at that DB max, or adding it as server-side validation in `PromotionService` so the constraint is enforced consistently rather than only in the UI.

Whatever you prefer, I'm glad to update.

michaelbromley (approved):

PR #4801, opened 2026-06-01
fix(dashboard): allow selecting nullable select items

casperiv0:
# Description

Allows selecting an empty option in dropdown/select, mimicking the Angular Admin UI to support nullable custom fields.

Currently, we can't use the dashboard because custom fields that are nullable are init as an empty string throwing the following error:

```
Failed to create promotion
GraphQL Request Error: The custom field "featureType" value [""] is invalid. Valid options are ['welcome-gift', 'we-forest', 'loyalty-level1', 'loyalty-level2', 'loyalty-level3', 'birthday-gift', 'product-variant-upsell']
```

# Breaking changes

Nope!

# Screenshots

image

# Checklist

📌 Always:
- [x] I have set a clear title
- [x] My PR is small and contains a single feature
- [x] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [x] I have added or updated test cases
- [x] I have updated the README if needed

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/4801"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782912425&installation_id=128408429&pr_number=4801&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F4801&signature=d1a47650d8976a0cef7e31f6d619aa97a799a468c24fba45fabc8064c6bb991d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

PR #5242, opened 2026-09-01
fix(dashboard): keep create forms valid with non-nullable custom fields

### LINKED ISSUE #5241: [dashboard] Non-nullable int/float/datetime custom field makes create forms start invalid with no visible error
**Describe the bug**

A `nullable: false` custom field of type `int`, `float` or `datetime` makes the matching dashboard **create** form invalid at first render, and nothing on the page says so — Create just stays disabled. The schema (from the custom-field config) requires the field; the form seeds it `null` (from the GraphQL input type, where every custom field is nullable); react-hook-form surfaces no error until a field-level validation runs.

This is **not** #5045 / PR #5057, which exempted `readonly` fields only; an editable `nullable: false` field is still required and still seeded `null` on `master` as of 2026-09-01 (section below).

Severity: a workaround exists (below), so not critical — the part worth fixing is the silence. Built-in Collection, Product, Facet, ShippingMethod and StockLocation create pages are among those affected.

**To Reproduce**

1. Add the custom field from **Configuration** below.
2. Start the server and open `/stock-locations/new` in the dashboard.
3. Fill in the name (and any other genuinely required field).
4. **Create** stays disabled; the field is an empty number input, no error text.
5. Type `0` into `defaultMinStock` — the button enables immediately.

**Expected behavior**

The form is valid from first render. Core refuses to boot unless a `nullable: false` custom field has a `defaultValue` ([validate-custom-fields-config.ts:85-95](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/entity/validate-custom-fields-config.ts#L85-L95)), and that value becomes the column's SQL `DEFAULT` ([register-custom-entity-fields.ts:220-234](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/entity/register-custom-entity-fields.ts#L220-L234)). Omitting the field therefore writes exactly the configured default — the form never needed a value; requiring one is wrong in principle.

**Actual behavior**

The form starts invalid; Create stays disabled until the admin happens to type into that field. No message points them there.

**Screenshots/Videos**

First: name filled in — Create greyed out, `defaultMinStock` empty, no error text. Second: with the fix, field still empty, Create enabled.

Image
Image

**Error logs**

The only evidence is the `console.log` at [`use-generated-form.tsx:203`](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx#L203):

```json
{ "customFields": { "defaultMinStock": { "message": "Expected number, received null", "type": "invalid_type" } } }
```

**Root cause**

Two sources of truth disagree in [`form-schema-tools.ts`](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts):

1. The Zod schema comes from the **custom-field config**: `applyCustomFieldModifiers` ([254-267](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L254-L267)) leaves the field required when `nullable === false` and not readonly:

```ts
if (customField.nullable !== false || customField.readonly) {   // line 260
    modifiedType = modifiedType.optional().nullable();
}
```

2. The defaults come from the **GraphQL input type**: `getDefaultValueFromField` ([439-493](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L439-L493), via [354-373](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L354-L373)) branches on the *GraphQL* nullability; `getGraphQlInputType`/`getGraphQlType` ([graphql-custom-fields.ts:702-713](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/api/config/graphql-custom-fields.ts#L702-L713), [727-751](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/api/config/graphql-custom-fields.ts#L727-L751)) never append `!`, so an `int` custom field is a nullable `Int` in `Create*Input` and seeds `null` ([466](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L466); post-processing at [393-418](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L393-L418) covers only structs and nullable selects).

Detail routes gate their button on `formState.isValid` ([`collections_.$id.tsx:139-147`](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/app/routes/_authenticated/_collections/collections_.$id.tsx#L139-L147)).

- @vendure/core version: 3.7.2 (also on `master`)
- @vendure/dashboard version: 3.7.2
- Nodejs version: 24.14.1
- Operating System (Windows/macOS/Linux): macOS 26.5.2
- Browser (if applicable): Chromium 151.0.7922.34 (headless, via Playwright 1.62.1)

**Configuration**

```typescript
// vendure-config.ts
customFields: {
    StockLocation: [
        {
            name: 'defaultMinStock',
            type: 'int',
            nullable: false,
            defaultValue: 0, // required, or the server refuses to boot
            label: [{ languageCode: LanguageCode.en, value: 'Default minimum stock' }],
        },
    ],
},
```

**Minimal reproduction**

The **Configuration** above reproduces it on a stock `@vendure/create` project; repository on request.

**Workaround**

Type a value (e.g. `0`); the button enables immediately. Hard to discover: the field renders empty, no error — and with `ui: { component: 'currency-form-input' }` it even looks filled, `MoneyInput` rendering `null` as `0.00` ([money-input.tsx:29](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/components/data-input/money-input.tsx#L29), [:35](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/components/data-input/money-input.tsx#L35), [use-local-format.ts:33-38](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/hooks/use-local-format.ts#L33-L38)). Dropping `nullable: false` also works, but gives up the constraint.

**Why this is not a duplicate of #5045**

PR #5057 (`1faa4c1dae`) fixed #5045 by adding the `|| customField.readonly` escape shown above, rescuing `readonly: true` fields only. For an editable `nullable: false` field both operands are false: the schema stays required, the `null` default still fails. Related but different: #4741 / PR #4868 (unexplained disabled submit buttons) would make this failure visible, not fix it.

**Proposed solution**

Two changes in `form-schema-tools.ts`, making the schema mirror the GraphQL input:

- `applyCustomFieldModifiers`: give `nullable: false` fields `.optional()` — accept a missing value, still reject explicit `null`.
- Default seeding: for `nullable: false`, omit the key instead of seeding `null`.

- **Create.** Absent key: validation passes, the mutation omits it, the column `DEFAULT` applies. `stripNullNullableFields` ([utils.ts:182-204](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/utils.ts#L182-L204), called on create from [use-generated-form.tsx:240-242](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/use-generated-form.tsx#L240-L242)) already strips null nullable custom fields before submit anyway; only the client-side Zod gate blocks the form.
- **Update.** The entity supplies a real value (NOT NULL column), so nothing changes. Clearing the input yields `null`, which `.optional()` still rejects — the button disables as today, correctly: the database would reject that write.

The obvious one-liner — extending #5057's `.optional().nullable()` to `nullable: false` — is wrong: it would let a cleared field through on update, producing a NOT NULL violation.

This does not revert PR #4339 (fixes #4155) or reopen #4328 (PR #4414) — both concern *nullable* fields, which keep defaulting to `null`; only `nullable: false` fields change.

Other types are unaffected — `list` seeds `[]` ([440](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L440)), `boolean`/string kinds seed `false`/`''`, `relation` gets `z.any()`, struct sub-fields are hardcoded `nullable: true` ([39](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/framework/form-engine/form-schema-tools.ts#L39)) — while `datetime` shares the `null` seed and the fix.

Optional follow-up: expose `defaultValue` ([custom-field-types.ts:106](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/config/custom-field/custom-field-types.ts#L106)) on the `CustomField` GraphQL interface ([custom-field-types.graphql:1-14](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/core/src/api/schema/common/custom-field-types.graphql#L1-L14)) so the dashboard ([server-config.tsx:8-26](https://github.com/vendurehq/vendure/blob/9f84f9c184b5bb80b4a2310e20b0221d63ea45a2/packages/dashboard/src/lib/providers/server-config.tsx#L8-L26)) can display the configured default.

Opened as #5242.

**Additional context**

- Started on 3.7.2 after adding `nullable: false` to existing numeric custom fields; reproduces on every load. Detail (edit) pages are unaffected — the loaded entity supplies a real value.

brmk:
# Description

Fixes #5241.

An `int`, `float` or `datetime` custom field with `nullable: false` makes the create form invalid as soon as it loads. Create stays disabled and no error is shown. This affects any create page with such a field, including Collection, Product, Facet, ShippingMethod and StockLocation.

The Zod schema comes from the custom field config and marks the field as required. The default values come from the GraphQL input type, where every custom field is nullable, so the field starts as `null` and fails validation. React Hook Form computes `isValid` on mount without setting `formState.errors`, so nothing is displayed.

**Fix:** `createFormSchemaFromFields` takes an `isCreateForm` flag. `use-generated-form` sets it from `!entity`, the same check its submit path already uses for `stripNullNullableFields`.

- **Create:** the field accepts `null`. On submit, `stripNullNullableFields` removes it from the payload, and the column's SQL `DEFAULT` is used. There is always a default: core won't start if a `nullable: false` custom field has no `defaultValue`.
- **Update:** the schema is the same as on `master`. Clearing the field disables Update, because the column is `NOT NULL`.

On update pages `useDetailPage` loads the entity with `useSuspenseQuery`, so `entity` exists on the first render and the flag is correct from the start.

Not covered: a `nullable: false` `string` field with a `pattern` defaults to `''`, which the regex rejects. That is a separate cause, and nullable string fields with a pattern have it too.

# Breaking changes

No API or schema changes. When the field is left empty on create, the configured `defaultValue` is saved without the admin seeing it. Showing the default in the form would need `defaultValue` on the `CustomField` GraphQL type.

# Screenshots

Before / after on `/stock-locations/new` with a single `nullable: false` `int` custom field (`defaultMinStock`): Create greyed out with the field empty, then Create enabled with that same field still empty.

# Checklist

📌 Always:

👍 Most of the time:
- [ ] I have updated the README if needed (not applicable)

# Tested

- Unit (`form-schema-tools.spec.ts`): the #5057 check is back unchanged (the update schema rejects `null` and `undefined`). The create schema accepts `null` for `int`, `float` and `datetime`. #4328 and #5045 still pass. The new cases fail on `master`.
- E2E (`catalog/custom-fields.spec.ts`): creating a product without filling `warrantyMonths` saves `12`, and clearing it on update disables Update. The field was added to `e2e-shared-config.ts`, the same way as in #4801 and #4988.


biggamesmallworld:
Merge lane session `c3882454` (@biggamesmallworld) is working this pull request.

Current stage: Stage 2: prove the bug is real. Claim expires 2026-09-15T12:17:02Z unless refreshed.

biggamesmallworld:
Thanks for this, @brmk. We verified the change independently and it holds up. We cannot push to this branch, because GitHub does not offer maintainer edits on forks owned by an organisation.

What we verified, against `master` at 5f983a342:

- On `master`, a create page with an untouched non-nullable int custom field keeps the Create button disabled and shows no error. The existing Stock Locations CRUD e2e suite fails for the same reason, so this blocks a shipped flow.
- With your change, both pass: `catalog/custom-fields.spec.ts` 24 tests and `settings/stock-locations.spec.ts` 12 tests. Dashboard unit: 533 tests. Type check and prettier pass.
- At runtime on postgres, the create mutation body is `{"input":{"name":"…","description":"","customFields":{}}}`. The key is omitted rather than sent as null, and the stored value is the configured default. That is the mechanism the fix rests on.
- The choice matches core. Core refuses to boot without `defaultValue` on a `nullable: false` field and writes it as the column default, and omitting the field has always written that default through the Admin API, the Shop API and plugins. The dashboard was the only place that blocked it, and it blocked int, float and datetime while allowing string, boolean and list.
- We tried to make the lenient create schema stay active on an update page. Forced by hand it does let a cleared field submit an explicit null, and the server rejects it with `The custom field "defaultMinStock" value cannot be set to null`, leaving the stored value unchanged. It is not reachable on its own: `useDetailPage` reads the entity through `useSuspenseQuery`, and instrumenting the hook showed `isCreateForm=false` on every render of an update page.

Three changes we would like on the branch:

1. The comment saying `entity` arrives after the first render, so the flag flips once, is wrong. Nothing renders before the entity is there, so it never flips. Say what the code guarantees instead.
2. Two branches in the create-form path have identical bodies. Collapse them into one condition and move the explanation into the JSDoc.
3. Add e2e coverage for both directions: create with an untouched non-nullable field persists the configured default, and clearing the field on update keeps the button disabled. `e2e/tests/catalog/custom-fields.spec.ts` is the right file, per the dashboard CLAUDE.md.

Two notes on scope. `convertEmptyStringsToNull` skips `String`, so a non-nullable string field still persists `''` rather than its configured default; that is unchanged by this PR. The dashboard renders no required indicator at all, so there is no marker pointing at a field that is not required. Showing one, with the default value, would need `defaultValue` on the `CustomField` GraphQL interface, which is a separate decision.

Your branch also overlaps #4868 in `use-generated-form.tsx`, so whichever lands second will need a rebase.

brmk:
Done.

1. The comment now states what `useSuspenseQuery` guarantees on update pages. It doesn't say the flag never changes: after a successful create, `products_.$id.tsx` navigates to `../$id` without remounting, so it goes from true to false once, right after the form is reset.
2. Collapsed to `nullable !== false || readonly || isCreateForm`, reasoning moved to the JSDoc.
3. Two tests in `catalog/custom-fields.spec.ts`, using a new `warrantyMonths` field (`int`, `nullable: false`, `defaultValue: 12`) in `e2e-shared-config.ts`. The update test makes the form dirty and valid before clearing the field, so the disabled button can only come from validation.

I also removed the "timing nuance" and "required indicator" paragraphs from the description; both were wrong.

Issue #3043, opened 2024-09-05
Introduce a CacheStrategy for multi-instance data caching


michaelbromley:
## The Problem

There are several places where Vendure core makes use of caching in order to significantly improve performance:

- We have the TtlCache class which is used in several places to cache calculated values in memory for a given time
- We have the SelfRefreshingCache type which is used to cache all zones, channels and tax rates in memory because they are accessed extremely frequently

The issue with those solutions is that they are in-memory-only, and therefore local to the specific server/worker instance.

Being in-memory has two major downsides:

1. **Duplication**: each instance of the server & worker maintains its own cache. On a typical deployment with 2x server & 2x worker processes, this means 4x the work needs to be done to fill up 4 caches. Load-balancing means that they will be filled unevenly, which is inefficient.
2. **Invalidation**: we cannot implement an invalidation method which works across multiple instances. E.g. the TaxRate cache: we would like to keep it cached forever until a TaxRate is changed/added/removed. We could naively set up an EventBus subscriber to clear the cache when we detect a change. However, the change event will _only_ be published on _one instance_. Any other instances will not know about that event, and their cache is now stale.

This second point is the reason we use TTLs with relatively short times. Again, leading to more work being done on the database.

It is also the reason we do not use caching in other scenarios that could radically improve performance: we cannot reliably invalidate a cache that is not shared by all instances.

**Example**

This issue was originally motivated by an investigation I am conducting into the performance of the order-related mutations. Using a prototype of this caching approach, I was able to speed up my benchmark by ~2.5x and cut the p(95) response time from 6.98s to 3s.

## Proposed Solution

I propose introducing a shared caching mechanism into the core: `CacheStrategy`. This will be strategy-based allowing you to decide whether you want to store that cache in:

- the database
- Redis
- some other key-value storage system

The `CacheStrategy` would replace all existing caching mechanisms mentioned above, and would unlock the opportunity to make huge performance gains in currently slow areas like:

- Checking Promotions on large orders
- Evaluating collection filters

Because the cache is shared, it means as soon as _one_ instance has cached a value, it will be available to _all_ instances.

## Design

At the most basic, the `CacheStrategy` will implement the typical cache methods: `get()` `add()`, `delete()`.

It should also support key eviction via TTL which would be configurable per key.

It should be able to store JSON-like data, i.e. any serializable JS data structure, just like we already support with the job queue.

Here's a sketch of how it would look:

```ts
[...]

Issue #2233, opened 2023-06-15
ChannelService.findAll cache uses the API query limit


goroka:
**Describe the bug**
Internal cache for channelservice (allChannels) refresh/findAll uses the default query limit of the API/Shop API instead of being unlimited.  (default 1000)

**To Reproduce**
Create a multi channel site with over 1000+ channels. 
On the admin page by selecting the channel that it is not in this 1k channel, the vendure triggers a CHANNEL-NOT-FOUND exception

**Expected behavior**
Should find it and the limit for the cached allChannels should be independent from the API query limite

**Environment (please complete the following information):**
 - @vendure/core version: 2.0.1
 - Nodejs version 18.16
 - Database (mysql/postgres etc): postgres

Issue #988, opened 2021-07-14
Cache strategy for channels in ChannelService


karel-advantitge:
**Is your feature request related to a problem? Please describe.**
The ChannelService as-is holds an array of all channels in memory, which is updated whenever a channel is created/updated. This makes the vendure instance stateful and can lead to invalid/out-of-sync data if a project uses multiple vendure instances.

**Describe the solution you'd like**
Querying from the database on every call would solve this, but since the data is used on every call (specifically `getChannelFromToken`) this would become a bottleneck.
Querying the session data presented a similar bottleneck, so I think we need a similar solution.
Similar to the `SessionCacheStrategy` config, I'd propose a `ChannelCacheStrategy`

michaelbromley:
Yes this needs to be solved. Let's consider the approaches:

1. **Cache nothing, fetch from DB when needed**
    ✔ Data always correct
    ✔ Simple, no extra config.
    ❌ Performance impact. It would not be as bad as the session request, since there are only 2 top-level joins. The session query joins 3 levels deep over a much larger data set. 
2. **Use a cache with TTL**
    Use the [`TtlCache` class](https://github.com/vendure-ecommerce/vendure/blob/85051b8e37fe6786e643892f3e700e8c9e65671a/packages/core/src/common/ttl-cache.ts#L7) to cache the channels for a limited period of time, e.g. 30 seconds.
    ✔ Performance would be similar to current solution
    ✔ No extra config to add to the API
    ❌ Still possibility of getting stale data up to <ttl> seconds after a change.
3. **Configurable cache strategy**
    ✔ Can externalize the cache store and share between instances
    ✔ Full control over caching
    ❌ Increases the config API surface

I personally like the simplicity of option 2, but in some circumstances I can imagine even a short amount of inconsistency might not be acceptable.

Investigating further, it can be seen that a similar in-memory caching technique is used in [PromotionService.activePromotions](https://github.com/vendure-ecommerce/vendure/blob/8653e834d4d626907f08107f661ac0e3923e1249/packages/core/src/service/services/promotion.service.ts#L273), which has the exact same issues as with the ChannelService.

Therefore it might make more sense to define a generalized caching strategy, which would encompass Sessions, Channels, Promotions and anything else needed in the future. This could be a single config field, which would deprecate and succeed the SessionCacheStrategy. 

I'm not exactly sure of what the API of a general cache strategy would look like, but to me that's quite an elegant solution. 

Thoughts?

[...]

PR #5248, opened 2026-09-01
feat(core): add configurable Channel and Zone cache strategies

rafabernad:
# Description

This PR adds configurable `ChannelCacheStrategy` and `ZoneCacheStrategy` extension points.

The default implementations use two cache levels:

- a short-lived process-local L1 for hot lookups;
- a longer-lived L2 backed by the globally configured `CacheStrategy` through `CacheService`.

Channels are cached individually by token, with a dedicated key for the default Channel. A miss loads only the requested Channel and its default tax and shipping Zones. Concurrent misses for the same Channel are coalesced into a single lookup.

Zone lists use the same L1/L2 model. The strategy receives the `RequestContext` and a loader callback, allowing custom implementations to control cache behavior without changing `ZoneService`.

Channel and Zone mutations immediately update or invalidate local entries and their corresponding L2 entries. Other processes can retain an L1 value until its short TTL expires.

`CacheService` is the sole lifecycle owner of the configured `CacheStrategy`. Initialization is lazy and single-flight, so early cache access and `onModuleInit()` share the same initialization attempt. Teardown only runs after successful initialization.

## Configuration

No configuration changes are required. The default configuration uses:

- `entityOptions.channelCacheTtl` and `entityOptions.zoneCacheTtl` for the L1 TTL, both defaulting to 30 seconds;
- a 5-minute L2 TTL;
- a maximum of 10,000 Channel L1 keys;
- the globally configured `systemOptions.cacheStrategy` for L2 storage.

The Channel defaults can be customized:

```ts
import { DefaultChannelCacheStrategy, VendureConfig } from '@vendure/core';

    entityOptions: {
        channelCacheTtl: 30_000,
        channelCacheStrategy: new DefaultChannelCacheStrategy({
            l1CacheSize: 10_000,
            l2CacheTtl: 5 * 60_000,
            cachePrefix: 'vendure-channel-cache',
        }),
    },
};
```

## Performance

A local compiled-code microbenchmark produced the following hot-L1 results:

| Cached Channels | Array scan baseline | L1 map lookup |
| ---: | ---: | ---: |
| 1 | 0.252 µs | 0.392 µs |
| 10 | 0.286 µs | 0.409 µs |
| 100 | 0.714 µs | 0.412 µs |
| 1,000 | 4.914 µs | 0.418 µs |
| 10,000 | 47.974 µs | 0.443 µs |

For small Channel counts, local bookkeeping adds approximately 0.1–0.2 µs relative to an array scan. The lookup remains O(1) as the Channel count grows. An L2 hit using the in-memory `CacheStrategy`, including deserialization and entity rehydration, measured approximately 4.2 µs. External backends additionally include their network or storage latency and are accessed only after an L1 miss.

## Related issues

Feature request: [Discussion #5247](https://github.com/vendurehq/vendure/discussions/5247)  
Relates to #988  
Relates to #3043  
Relates to #2233

## Testing

- Added unit coverage for L1 hits, expiry and capacity limits.
- Added L2 serialization and Channel, Zone, Country and Date rehydration coverage.
- Added mutation invalidation and single-flight lookup coverage.
- Added `CacheStrategy` lifecycle coverage for early access, concurrent initialization, retry and teardown.
- Full `@vendure/core` unit suite: 1,543 passed.
- Channel E2E suite: 21 passed.
- Zone E2E suite: 14 passed.
- `@vendure/common` and `@vendure/core` TypeScript builds completed successfully.

# Breaking changes

None. Both strategy properties are optional and have default implementations.

# Screenshots

Not applicable.

# Checklist

📌 Always:

👍 Most of the time:


rafabernad:

michaelbromley (changes requested):
Thanks for this. The diagnosis is correct and the problem is worth fixing. The solution is larger than the problem needs, and below is where I think it should land.

## The benchmark measures the wrong thing

The benchmark table compares an array scan against a map lookup. A difference of microseconds does not matter here, because it sits far below the cost of resolving a single GraphQL field.

The problem is in the code this PR deletes. `ChannelService.createCache()` ran `listQueryBuilder.build(Channel, {}, { relations: ['defaultShippingZone', 'defaultTaxZone'], ignoreQueryLimits: true })` every 30 seconds in every instance. A deployment with 10,000 Channels therefore ran a 10,000 row query with two joins every 30 seconds per instance. Each instance also held 10,000 hydrated Channel entities and their Zones in memory, including Channels which received no traffic. The `ignoreQueryLimits: true` flag is there because a deployment hit the query limit and Channels above it became unreachable.

Loading one Channel per token on a cache miss fixes all of that. That change is the part of this PR I want, and it does not need anything else in the PR to work.

## The L2 does not pay for itself

`Channel.token` is declared `@Column({ unique: true })`, so a lookup by token reads a single indexed row.

The L2 replaces that read with a Redis round trip of similar latency, a recursive deserialize, and entity reconstruction. For Zones the cost is higher. One query becomes a fetch and deep parse of every Country with its eager translations, which is roughly 130KB of JSON on the default dataset. That runs every 30 seconds in every instance.

The L2 does not reduce staleness either. L1 keeps a 30 second TTL and nothing invalidates it across instances. A Channel updated on one instance therefore stays stale on the others for up to 30 seconds, with or without the L2.

The default `systemOptions.cacheStrategy` is `InMemoryCacheStrategy`. In a default deployment the shared L2 is therefore a second Map in the same process. Every L1 expiry runs a serialize walk on write, and a deserialize walk with entity reconstruction on read, to move data between two Maps in one heap. For a single instance deployment that is slower than the code it replaces.

There is a second reason to drop it. All Channel and Zone mutations are wrapped in `@Transaction()`, and `createChannel`, `updateChannel` and `refreshCachedZones` write to the cache before the transaction commits. A rollback currently leaves stale data in one instance for 30 seconds. Writing through to a shared cache with a five minute TTL spreads that same stale data across the whole cluster.

## The strategies

`systemOptions.cacheStrategy` is already the extension point for cache storage, and `CacheService` is already the way a service reaches it. A `ChannelCacheStrategy` and a `ZoneCacheStrategy` add a second layer of cache configuration above one which already exists.

Once `ChannelService` loads a single Channel per token, I cannot name a caching requirement the core implementation would fail to serve. Each strategy we add is a public API we support indefinitely and a documentation page we maintain, so I would rather add one when a requirement arrives that the core implementation cannot meet.

## The CacheService change should survive

`ConfigModule.onApplicationBootstrap` initializes `cacheStrategy` today, which is after the first code paths that read the cache. Your lazy single-flight init in `CacheService` removes that ordering requirement. The same gap is already worked around elsewhere:

```ts
// default-session-cache-strategy.ts
// We use the `?` here because there is a case where in the SessionService,
// the clearSessionCacheOnDataChange() method may be invoked during bootstrap prior to
// the cacheService being initialized in the `init()` method above.
return this.cacheService?.invalidateTags(this.tags);
```

I will take that fix independently of the caching work.

## What I would like to see

1. In `ChannelService`, replace the all-Channels cache with a per-token cache holding one entry per Channel. Bound its size, keep the existing `channelCacheTtl`, and load a single Channel on a miss.
2. Drop the Zone changes. Zone count is bounded by geography and Zones change rarely, so the existing cache is adequate.
3. Drop both strategy interfaces and the L2.
4. Split the `CacheService` initialization fix into its own PR against `master`.

Point 1 carries the whole scaling benefit, in roughly the number of lines the current cache occupies.

I know this asks you to remove most of what you wrote. The investigation behind it is sound, and the problem you found is one we needed to fix. I would rather land the part which carries the benefit and add configuration later if a requirement for it appears.

PR #5250, opened 2026-09-01
feat: add custom field support for Role entity

### LINKED ISSUE #5249: Support custom fields on the Role entity
## Describe the feature

Role is the only core entity with admin CRUD that cannot have custom fields. Every other entity in the `CustomFields` config type accepts them, so this fails to compile and does nothing at runtime:

```ts
customFields: {
    Role: [{ name: 'note', type: 'string' }],
},
```

This is useful for storing extra data on roles, for example an ID from an external permission system or an internal note about what a role is for.

Adding Role support means:

- The `CustomFields` config type accepts a `Role` key.
- The Admin API exposes `customFields` on the `Role` type and on `CreateRoleInput` / `UpdateRoleInput`.
- `createRole` and `updateRole` persist the values.
- The Dashboard role detail page shows and saves the fields.

Role has no translations, so `localeString` and `localeText` fields stay unsupported. The server already logs an error when a config defines those on a non-translatable entity, and that check covers Role without extra work.

gabriellbui:
## Description

Adds custom field support to the Role entity, the last core entity with admin CRUD that could not have them.

With this change, a project can write:

```ts
customFields: {
    Role: [{ name: 'note', type: 'string' }],
},
```

and the Admin API exposes `customFields` on the `Role` type and on `CreateRoleInput` / `UpdateRoleInput`. The Dashboard role detail page shows and saves the fields. Role has no translations, so `localeString` / `localeText` fields stay unsupported; the existing bootstrap check already logs an error for those on non-translatable entities.

### What changed

Core, hand-written:

- `custom-entity-fields.ts` — new empty `CustomRoleFields` class. No database columns exist until a project defines fields, so no migration is needed.
- `role.entity.ts` — implements `HasCustomFields`, adds the embedded `customFields` column.
- `custom-field-types.ts` / `default-config.ts` — `Role` added to the config type and the defaults.
- `role.service.ts` — create and update now copy `input.customFields` onto the entity and call `CustomFieldRelationService.updateRelations` after saving, so relation-type fields persist too. Everything else (schema extension, validation, list filtering and sorting) picks Role up from the config key with no Role-specific code.

Dashboard:

- `roles.graphql.ts` — `customFields` added to the role detail query, for TypeScript types only; the framework injects the runtime selection.
- `roles_.$id.tsx` — the form loads `entity.customFields` and renders the standard `CustomFieldsPageBlock`.

Generated files regenerated via the root codegen scripts. Note: the dashboard's `graphql-env.d.ts` also picks up two pieces of schema drift that predate this PR (the old `metricSummary` types removed after the DashboardPlugin renamed them, and the new administrator avatar field and mutation).

### Testing

Four new e2e tests in `packages/core/e2e/custom-fields.e2e-spec.ts`: create persists a Role custom field, the `role` query reads it back, `updateRole` changes it, and updating only the description leaves it intact.

Runs: `custom-fields.e2e-spec.ts` 61/61, `role.e2e-spec.ts` 31/31, wider rerun of related suites 205 passed / 1 skipped. `@vendure/common` + `@vendure/core` build clean, dashboard `check-types` clean.

Fixes #5249

## Breaking changes

None.

## Checklist

📌 Always:
- [x] My PR is small and focused
- [x] There is a linked issue

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5250"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790879336&installation_model_id=20193&pr_number=5250&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5250&signature=a1aa91dd07885318ab3b033dca0b4cff7da10cda1b8c46144ee9ba69be0c9b56"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld (changes requested):
Core change looks good, follows the Seller/Zone pattern. A few things first.

**Are the Angular admin-ui changes needed?** We only maintain the React dashboard now, so `9d51147` may be worth dropping. If it stays, `role-list.component` needs `.addCustomFieldFilters()`, `.addCustomFieldSorts()` and `<vdr-dt2-custom-field-column>` too, otherwise the fields are invisible in the list (cf. `seller-list.component`).

Changes requested:

1. **Drop the `graphql-env.d.ts` hand-patch in `7284367`.** The metric types should stay gone: that file is generated from the live schema at `localhost:3000/admin-api`, `dev-config.ts` loads `DashboardPlugin` only, and the dashboard only queries `dashboardMetricSummary`. Nothing in `packages/dashboard/src` references `metricSummary`. The next regeneration will drop them again anyway. Please also split the unrelated drift in that file (administrator avatar, `deprecated`, `ConfigArgDefinition.defaultValue`, `ReadSecret`) into its own commit.

2. **Bind the `updateRelations` return in `createRoleForChannels`.** It currently discards it and returns the in-memory `newRole`, which only works because `updateRelations` mutates in place. `update` and `SellerService.create` both use the return value.

3. **Tests.** No coverage for the list filtering/sorting the description says comes for free; `does not delete primitive values` asserts `primitive` and `single` but not `multi`, which is the one the per-field merge could clobber; no case for setting a field back to `null`.

4. `RoleService` gained a required constructor parameter, so "Breaking changes: None" is not accurate for subclassers. `minor` is still the right target.

Non-blocking: `Role.customFields` is now in the shop schema and reachable via `Customer.user.roles`, and custom fields default to `public: true`. Also `getAllRolesWithChannels` serializes Roles into the cache behind every permission check.

gabriellbui:
@biggamesmallworld Thanks. What changed, and where I disagree.

Done:

- Dropped the Angular admin-ui commit. Reverting is safe: `add-custom-fields.ts` only adds a `customFields` selection when an entity has fields configured, so the role pages send the same queries as before.
- Reverted the metric types patch, they stay gone.
- Added tests for filtering and sorting the roles list by a custom field, and for setting one back to null.
- `create` was returning the in-memory role where `ZoneService.create` re-reads it. Now it re-reads too.

Where I disagree:

- Binding the `updateRelations` return cannot change anything. It has one `return entity;` and mutates in place, so the value is always the argument you passed in. Our own `update` discards it as well, and so do most call sites in core.
- `multi` being unasserted matches all nine other entity blocks in that spec. It also cannot be clobbered: the loop skips any field whose ids are `undefined`.
- Setting a relation to null is already tested, directly below the test you quoted. The gap was a plain field, which I added.
- On breaking changes, `CONTRIBUTING.md` lists database schema, GraphQL API, `VendureConfig` and behaviour changes. Service constructors are not on that list, and `c68120a4e` made this same change to `ZoneService` and shipped in v3.5.1, a patch.

On the shop API: every entity in the shop schema exposes `customFields`, including User, Channel and Seller. Role is no different, so I kept the default and operators opt out per field with `public: false`.

One deviation from your first point. The unrelated drift in the typed schema file is not split into its own commit, since we are keeping this response as a single commit. It is described in the commit body.

biggamesmallworld (approved):
Looks cool!

PR #5255, opened 2026-09-02
fix(asset-server-plugin): derive served Content-Type from the file bytes

biggamesmallworld:
## Problem

`getFileNameFromParameters` keeps the source asset's extension when a transform request has no `format` param, and the served `Content-Type` was derived from that file name. So `/<asset>.svg?w=100` was served as `image/svg+xml` even though sharp had rasterised the body to a PNG.

That mime value also decides the security headers added in #5102: the mislabelled response got the markup CSP (`sandbox`) and `Content-Disposition: attachment`, so a browser downloaded the transformed image instead of rendering it. With `X-Content-Type-Options: nosniff` it could not recover by sniffing either.

The mislabelling is pre-existing; #5102 is what made it visible.

## Change

`getMimeType(fileName)` is replaced by `resolveMimeType(buffer, fileName)`, used by both serve paths (`sendAsset` and `generateTransformedImage`). The file's own signature takes precedence, with the file name as the fallback for text-based formats which have no signature to detect.

One guard: `file-type` reports every XML document as `application/xml`, so for a `+xml` file name the more specific name-derived type is kept. Without it, real SVG sources would have been served as `application/xml`.

This also corrects any other asset whose bytes disagree with its name, and it fixes the cache-hit path as well as the first request, since the cached file is still stored under the source extension.

## Test

`serves a transformed SVG as the rasterised type, not as SVG` in `asset-server-plugin.e2e-spec.ts` fetches `test.svg?w=100` twice, so it covers the generated response and the later cache hit, and asserts `image/png` with the non-markup headers. It fails on master and passes here.

Full `asset-server-plugin` e2e suite: 41/41 passing.

## Known limitation

The cache file on disk still carries the source extension, so `S3AssetStorageStrategy.writeFileFromBuffer` records the same stale `ContentType` on the S3 object. Assets served through the asset server are unaffected because the type is now resolved at serve time; a URL pointing directly at the S3 object would not be. Left out of this fix.

Relates to #5102

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5255"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790941235&installation_model_id=20193&pr_number=5255&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5255&signature=b04ad4110812a2cbc23287090d87aeb7456bdd27fc7dfb6cfcbae3de25932238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
+++ b/packages/asset-server-plugin/e2e/asset-server-plugin.e2e-spec.ts
@@ -262,6 +262,21 @@ describe('AssetServerPlugin', () => {
             expect(res.headers.get('x-content-type-options')).toBe('nosniff');
             expect(res.headers.get('content-security-policy')).toBe(BASE_CSP);
         });
+
+        // A rasterised SVG used to be served as `image/svg+xml`, which sandboxed and
+        // force-downloaded it. Checked twice: once generated, once from cache.
+        it('serves a transformed SVG as the rasterised type, not as SVG', async () => {
+            const svgAsset = await uploadAsset('test.svg');
+
+            for (const pass of ['generated', 'from cache']) {
+                const res = await fetch(`${svgAsset.source}?w=100`);
+
+                expect(res.headers.get('content-type'), pass).toContain('image/png');
+                expect(res.headers.get('content-disposition'), pass).toBeNull();
+                expect(res.headers.get('x-content-type-options'), pass).toBe('nosniff');
+                expect(res.headers.get('content-security-policy'), pass).toBe(BASE_CSP);
+            }
+        });
     });
 
     describe('unexpected input', () => {
--- a/packages/asset-server-plugin/src/asset-server.ts
+++ b/packages/asset-server-plugin/src/asset-server.ts
@@ -98,10 +98,7 @@ export class AssetServer {
             const key = this.getFileNameFromParameters(req.path, params);
             try {
                 const file = await this.assetStorageStrategy.readFileToBuffer(key);
-                let mimeType = this.getMimeType(key);
-                if (!mimeType) {
-                    mimeType = (await getFileType(file))?.mime || 'application/octet-stream';
-                }
+                const mimeType = await this.resolveMimeType(file, key);
                 res.contentType(mimeType);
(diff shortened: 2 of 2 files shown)


biggamesmallworld:
Raised against my own PR from a classification audit. The precedence inversion opens a path this change does not cover.

A file named `.html`, `.svg` or `.xml` whose bytes carry a recognised binary signature, such as a GIF or PDF header, now resolves to that binary type. Binary types are not in `MARKUP_MIME_TYPES`, so the response loses the sandbox CSP and the `Content-Disposition: attachment` that its filename used to earn it.

`X-Content-Type-Options: nosniff` is what stops that becoming stored XSS from an admin holding only `CreateAsset`. This PR leaves that header alone, but the safety of the served type now rests on it rather than on the mime choice. Neither the description nor the tests say so.

Two smaller points:

- `getFileType` now runs on every asset request rather than only on a `mime.lookup` miss, so buffer inspection sits on the hot serving path.
- The transform fallback moved from `image/jpeg` to `application/octet-stream`, which forces a download where an image previously rendered.

The XML guard is correct. `application/xml` is in `MARKUP_MIME_TYPES`, so a file detected as XML stays sandboxed either way.

Also worth a follow-up issue rather than a paragraph in the description: `s3-asset-storage-strategy.ts:255` still sets `ContentType: mime.lookup(fileName)` at write time. Different mechanism and only reachable by bypassing the asset server, but the standing policy on partial application asks for the follow-up to be committed rather than noted.

PR #5260, opened 2026-09-03
test(core): add e2e coverage for createFromFileStream without a RequestContext

### LINKED ISSUE #5204: Add e2e coverage for AssetService.createFromFileStream() without a RequestContext
`AssetService.createFromFileStream()` has an overload that takes no `RequestContext`. Nothing in the repo tests it, and the only internal caller (`AssetImporter`, `asset-importer.ts:45`) always passes a `ctx`. The overload exists for plugin code.

That gap is why #4651 went unnoticed: the path broke in v3.6 and CI stayed green for a release cycle.

#4662 fixes the bug. The test that came with it was dropped during review and should land separately.

### What to add

A trailing `describe` at the end of `packages/core/e2e/asset.e2e-spec.ts`. It has to go last, because earlier blocks in that file assert absolute asset counts and a persisted asset shifts them (`filter by tags > empty array`, line 411).

```ts
// https://github.com/vendurehq/vendure/issues/4651
describe('createFromFileStream without a RequestContext', () => {
    const assetGuard: ErrorResultGuard<Asset> = createErrorResultGuard(input => input.id != null);

    it('creates the default translation with the default Channel languageCode', async () => {
        const assetService = server.app.get(AssetService);
        const defaultChannel = await server.app.get(ChannelService).getDefaultChannel();

        const result = await assetService.createFromFileStream(
            Readable.from(['test file content']),
            'test-file.pdf',
        );

        assetGuard.assertSuccess(result);
        expect(result.name).toBe('test-file.pdf');
        expect(result.translations.map(t => t.languageCode)).toEqual([
            defaultChannel.defaultLanguageCode,
        ]);
        expect(result.channels.map(c => c.id)).toEqual([defaultChannel.id]);
    });
});
```

`.pdf` is already in the suite's `permittedFileTypes`, so no config change is needed. `AssetService` and `Readable` are already imported by #4662; `Asset` and `ChannelService` need adding to the `@vendure/core` import on line 4.

The assertions cover both halves of the fix. `languageCode` is the reported bug. The channel assertion covers the second half: the old `RequestContext.empty()` path also fed a channel-less context into `assignToCurrentChannel`.

### Verifying

From `packages/core`:

```
bun run e2e asset.e2e-spec.ts
```

On current master the suite is 23 green. With `asset.service.ts` reverted to its pre-#4662 state, only the new test fails, with `NOT NULL constraint failed: asset_translation.languageCode`.

Relates to #4651

wakqasahmed:
Closes #5204.

Adds the regression test that #4662 fixed but didn't land with, covering `AssetService.createFromFileStream()`'s no-`RequestContext` overload — the path that broke in v3.6 (#4651) with CI staying green the whole cycle because nothing exercised it.

Added as a trailing `describe` block at the end of `asset.e2e-spec.ts` so it doesn't shift the absolute asset counts the earlier blocks assert on (the file's `filter by tags > empty array` block, for one).

To make sure this is actually pinned to the fix and not a false-positive, I reverted `asset.service.ts` back to its pre-#4662 state locally and reran just this test: it fails with `NOT NULL constraint failed: asset_translation.languageCode`, exactly as expected, then re-applied the fix and reran the full file - back to 27/27 green.

## Test plan

- `bun run e2e asset.e2e-spec.ts` from `packages/core` - 27 passed (was 26 before this PR)
- Isolated the new test with `-t "creates the default translation"` - passes on its own
- Reverted the `asset.service.ts` fix from #4662 and confirmed the new test is the one that fails, with the exact error described in #5204

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5260"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791010504&installation_model_id=20193&pr_number=5260&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5260&signature=ccf300e5bce032dffafbe8d5c08739c845d87a05fd0908def8826ac7da63d789"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


gabriellbui:
Thanks for this, @wakqasahmed. Merging it now.

We pushed a commit to your branch and are merging it ourselves rather than leaving review comments and waiting for you to act on them. That is our standard process for well-bounded changes now, and it is worth explaining so it does not come as a surprise. Pull requests arrive faster than we have maintainer hours to review them, and most of the ones that stall do so in the round trip between a review comment and a follow-up push.

What we changed on top of your work, all in `packages/core/e2e/asset.e2e-spec.ts`:

- Streamed real PDF bytes instead of the string `test file content`. Plain text has no magic-byte signature, so content detection returned nothing and the file passed permission checks on the `.pdf` extension alone. Tighten content detection later and the test fails with a `MimeTypeError` that looks unrelated to what it guards. A `Buffer` also takes the same byte path as the real caller in `AssetImporter`, rather than the object-mode string path.
- Added a read-back of the Asset with its `channels` relation. The returned entity carries the ids assigned in memory before the save, so on its own it cannot show the join row was written. Your in-memory assertion stays, because that is the one that catches a missing `ctx.channelId`.
- Moved the ordering note above so it covers both trailing `describe` blocks, since each now creates an Asset.

All 27 tests in that file pass. We also confirmed the test still fails if the request context fallback is removed, with `expected [ undefined, 1 ] to deeply equal [ 1 ]`.

If you disagree with any of it, say so and we will revisit. Thanks again.

gabriellbui (approved):

PR #5267, opened 2026-09-05
fix(dashboard): fix country selection in address forms

### LINKED ISSUE #5191: Country dropdown on customer address form is incomplete and doesn't show an existing address's country

psang39:
# Description

Fixes #5191

The Dashboard country selectors had two related issues:

- Enabled countries were returned in insertion order, causing newly-created countries to appear at the end of an otherwise alphabetical list.
- A saved country code could be set before the asynchronously-loaded country options were available, leaving the Select trigger without the correct label.

This change:

- Sorts enabled countries by name in the existing single list query.
- Keeps the country Select controlled from its initial render.
- Extracts a shared `CountrySelect` for the customer and shipping-method test-address forms.
- Resolves selected country labels after the country data loads and displays a translated placeholder when no country is selected.
- Adds an accessible name to the customer address edit trigger.
- Shares the shipping test-address local-storage key with the E2E tests.

Regression coverage includes:

- A unit test for the single-page country query result.
- An E2E test for pre-selecting a saved customer address country.
- E2E tests for the empty and saved country states in the shipping-method test-address form.

Validation performed:

- Dashboard source tests: 362 passed.
- Targeted country query unit test: 1 passed.
- Customer and shipping-method Playwright suites: 22 passed.
- Dashboard type-check: passed.
- Dashboard production build: passed.
- Formatting checks: passed.

# Breaking changes

None.

# Screenshots

N/A

# Checklist

📌 Always:

👍 Most of the time:


psang39:

biggamesmallworld (dismissed):
Thanks for the tests and the writeup. Three things to change.

**1. The pagination loop is not the fix, and the PR body says it is.**

`adminListQueryLimit` defaults to 1000 (`default-config.ts:84`), the seed set is 248 countries, and `ListQueryBuilder` already falls back to that limit when `take` is omitted (`list-query-builder.ts:643`). On a default install the old query returned every enabled country in one request.

The old query was missing `sort`, so countries came back in insertion order and a newly created one landed at the bottom of an otherwise alphabetical list. `sort: { name: 'ASC' }` is the one-line fix for symptom 1. Drop the loop, or keep it and describe it in the PR body as hardening for a lowered `adminListQueryLimit`.

**2. The `SelectValue` children function kills the placeholder.**

Base UI documents the placeholder as "overridden by `children` if specified". The function returns `undefined` when nothing is selected and while loading, so the new-address form renders an empty trigger. Both e2e tests only cover the pre-filled case. Handle the empty case in the function and drop the unreachable `placeholder` prop. The callback argument is typed `any`, not `string`.

**3. Extract a shared `CountrySelect`.**

`customer-address-form.tsx:170-190` and `test-address-form.tsx:200-223` are identical for 20 lines, except one uses ``t`Select a country` `` and the other `<Trans>`. `use-available-countries.ts` is already the shared home for this.

**Smaller:**

- `fetchAllAvailableCountries` returns a fake GraphQL envelope, and reports `totalItems: 0` on the empty-page path when the server said `1`. The unit test asserts that wrong value. Return the array.
- Derive the item type from the document instead of restating the selection set.
- `value => value && field.onChange(value)` silently drops a change event. Use an `if` with a comment if clears are meant to be ignored.
- `expect(customerId).toBeTruthy()` discards the `ErrorResult` message the query already selects.
- The edit trigger in `customer-address-card.tsx:123` is a bare `EditIcon` with no accessible name, hence the `.locator('..')` traversal. Add `aria-label`, then use `getByRole('button', { name: 'Edit address' })`.
- `shipping-methods.spec.ts:44`: use `LS_KEY_SHIPPING_TEST_ADDRESS` from `src/lib/constants.ts:17`.
- The unit test asserts `sort`/`filter` literals copied from the function, and never covers the single-page case that runs in production.
- The new shipping test sits in a describe block called "Shipping Methods CRUD" and is not CRUD.

psang39:
Thanks for the detailed review. I’ve addressed the requested changes:

- Removed the pagination loop and added `name: ASC` sorting to the existing single countries query.
- Extracted a shared `CountrySelect` and fixed the empty/loading placeholder state while keeping the Select controlled.
- Changed the fetch helper to return the country array directly and derived the item type from the GraphQL document.
- Improved the customer E2E error handling and added an accessible name to the edit-address trigger.
- Added coverage for the empty placeholder and single-page query behavior, and moved the shipping test-address coverage outside the CRUD describe block.

One implementation detail: importing `LS_KEY_SHIPPING_TEST_ADDRESS` directly from `constants.ts` caused Playwright’s Node runtime to load `virtual:admin-api-schema` through `schema-enums.ts`. I moved the storage keys to a dependency-free `storage-keys.ts`, re-exported them from `constants.ts`, and used the same canonical constant in the E2E test.

Validation completed:
- Dashboard source tests: 362 passed
- Playwright customer and shipping suites: 22 passed
- Type-check and production build passed
- Formatting checks passed

michaelbromley:

michaelbromley:
Thanks for this, @psang39. We're going to merge it.

extra review round trip.

What we changed:

- Moved the country select into its own component file, so the countries hook is plain TypeScript again rather than a module holding a component.
- Deleted the hook test, which only exercised a function extracted to make it testable, and inlined that function back into the query.
- Added a test for the country ordering, which nothing covered.
- Gave the country select an accessible name and a loading skeleton, matching the existing zone selector.
- Made the ordering test wait for the popup, because it read the option list before the popup had opened.
- Made the shipping method test retry its click, because a refetch from a parallel worker can leave the sheet closed.
- Explained in a comment why the storage keys sit in their own module.
- Explained in a comment why an empty country value is ignored, rather than leaving the guard unexplained.

Verification

Editing an address on master, and with the fix:

Edit address dialog on master with Full Name, Street and City populated but Country reading Select a country

The same dialog with the fix, Country reading United States of America

The country list on master, and with the fix:

Open country dropdown on master with Belgium listed last, below United States of America

The same dropdown with the fix, Belgium in alphabetical position and the selected country check-marked

michaelbromley (approved):
Reproduced both symptoms of #5191 on origin/master (country not pre-selected when editing a saved address, and a new country sorting last), confirmed both pass with the fix, and verified the before/after rendering of the dialog and the dropdown. Root cause is Base UI's useControlled latching the Select uncontrolled when value is undefined on first render. Full CI green with no build or test job skipped.

PR #5269, opened 2026-09-05
feat(core): allow field resolver enhancers to be configured

### LINKED ISSUE #5268: Global NestJS interceptors never run for @ResolveField() resolvers, with no way to opt in
**Describe the bug**

`configure-graphql-module.ts` hardcodes the Apollo driver's `fieldResolverEnhancers` to `['guards']` for
both the Shop API and the Admin API:

```ts
// packages/core/src/api/config/configure-graphql-module.ts
return {
    path: '/' + options.apiPath,
    typeDefs: printSchema(builtSchema),
    include: [options.resolverModule],
    inheritResolversFromInterfaces: true,
    fieldResolverEnhancers: ['guards'],
    // ...
};
```

So an interceptor a plugin registers with Nest's `APP_INTERCEPTOR` token runs for top-level queries and
mutations and never for a field resolver. Same for an exception filter registered with `APP_FILTER`.
Nothing fails and nothing is logged. The plugin compiles, the field resolver returns the right data, and
the interceptor is simply absent from that code path.

We lost hours to this twice on the same project, months apart. Both times the unit tests were green,
because they construct the resolver class and call the method directly, which bypasses Nest's enhancer
pipeline altogether. The second time round it took a while to remember that the cause was one hardcoded
array inside core.

The concrete need behind both incidents was production performance logging: a global interceptor that
times every resolver and logs the slow ones. Most of the expensive work in our API happens inside field
resolvers — custom fields, DataLoader-backed relations — which is exactly the code path the interceptor
never saw. As shipped, there is no way to build that kind of slow-resolver logging without
hand-decorating every field resolver in every plugin.

**To Reproduce**

```ts
@Injectable()
class LoggingInterceptor implements NestInterceptor {
    intercept(context: ExecutionContext, next: CallHandler) {
        const info = GqlExecutionContext.create(context).getInfo<GraphQLResolveInfo>();
        Logger.info(`${info.parentType.name}.${info.fieldName}`);
        return next.handle();
    }
}

@Resolver('Product')
class ProductFieldResolver {
    @ResolveField()
    somethingCustom() {
        return 'a value';
    }
}

@VendurePlugin({
    adminApiExtensions: {
        resolvers: [ProductFieldResolver],
        schema: gql`
            extend type Product {
                somethingCustom: String!
            }
        `,
    },
    providers: [{ provide: APP_INTERCEPTOR, useClass: LoggingInterceptor }],
})
export class MyPlugin {}
```

Run `{ products(options: { take: 1 }) { items { id somethingCustom } } }` against the Admin API.

**Expected behavior**

Two log lines: `Query.products` and `Product.somethingCustom`.

**Actual behavior**

One log line: `Query.products`. The `somethingCustom` field resolves correctly, so the only signal that
the interceptor was skipped is its own missing output.

**Suggested fix**

Not a change of default. Nest documents `fieldResolverEnhancers` as a performance tradeoff, and it is a
real one: with `interceptors` enabled, every global interceptor runs once per resolved field, so a list
query returning 100 items runs them 100 extra times. Vendure's own interceptors are in that set too.
`['guards']` is the right default, and it is what makes `@Allow()` work on field resolvers.

What is missing is a way to opt in, and a line in the docs saying what the default leaves out. Something
like `apiOptions.fieldResolverEnhancers`, defaulting to `['guards']`, passed through to the driver config
for both APIs.

The change in core is minimal: the hardcoded array becomes a config lookup whose default is the same
`['guards']`, so existing installations see no behavioural difference unless they set the new option.

The current workaround is to put `@UseInterceptors()` on each field resolver by hand, which defeats the
point of registering the interceptor globally, and which you only think of once you already know what is
wrong.

**Environment**

- @vendure/core version: 3.7.2 (the same line is present on `minor`)
- Nodejs version: 22
- Database: postgres
- Operating System: Linux
- Package manager: npm

brmk:
# Description

Closes #5268.

Adds `apiOptions.fieldResolverEnhancers`, so that the NestJS enhancers applied to GraphQL field resolvers can be configured.

`configure-graphql-module.ts` hardcoded `fieldResolverEnhancers: ['guards']` for both the Shop API and the Admin API. That means an interceptor a plugin registers with Nest's `APP_INTERCEPTOR` token — or an exception filter registered with `APP_FILTER` — runs for top-level queries and mutations and never for a `@ResolveField()` resolver. Nothing fails and nothing is logged, so the only symptom is a missing side effect. We hit this twice on the same project, and unit tests stayed green both times because they call the resolver method directly and bypass Nest's enhancer pipeline.

The default is deliberately unchanged. Nest documents this as a performance tradeoff and it is a real one: with `interceptors` enabled every global interceptor runs once per resolved field, and Vendure's own interceptors are in that set. So this only makes the choice available, and documents what the default leaves out.

Changes:

- `ApiOptions.fieldResolverEnhancers?: Enhancer[]` (Nest's own `Enhancer` type, imported as a type-only import so the `@nestjs/graphql` barrel is not pulled in at require time), `@since 3.8.0`.
- `defaultConfig` sets `['guards']`; `configure-graphql-module.ts` reads it from `configService.apiOptions`, matching how `introspection` and `csrfPrevention` are already handled there. Both APIs pick it up.
- A note in the "Global NestJS middleware" section of the API layer guide, since that page currently says these middleware classes apply "to all requests".

Verification:

- `packages/core/e2e/field-resolver-enhancers.e2e-spec.ts` — a plugin registers a global interceptor that records every `Type.field` it fires for, and resolves a `@ResolveField()` probe on `Product`. One server on the default config, one with `['guards', 'interceptors']`. Both pass; reverting the source change makes the second test fail, as it should.
- `packages/core`: `bun run test` — 1518 passed. `e2e` for `auth`, `plugin`, `apollo-server-plugin`, `csrf-prevention` — 44 passed.
- `bun run docs:generate-typescript-docs` (only the reference pages whose content actually changed are included; the sourceLine-only renumbering churn it also produces is left out).

# Breaking changes

None. The default value is the previously hardcoded one.

# Screenshots

N/A.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5269"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791189317&installation_model_id=20193&pr_number=5269&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5269&signature=91280f8ee584e7caf8c65c16038ce67163e2fdb62795e066b31fd8a24f62b128"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/guides/developer-guide/the-api-layer/index.mdx
+++ b/docs/docs/guides/developer-guide/the-api-layer/index.mdx
@@ -196,6 +196,14 @@ export const config: VendureConfig = {
 };
 ```
 
+:::note
+Guards are the only enhancer applied to field resolvers (those declared with `@ResolveField()`) by default,
+so a globally-registered interceptor or exception filter will not fire when a field resolver runs. If you
+need it to, opt in with the
+[`apiOptions.fieldResolverEnhancers`](/reference/typescript-api/configuration/api-options#fieldresolverenhancers)
+config property.
+:::
+
 
 
 ### Apollo Server plugins
--- a/docs/docs/reference/typescript-api/configuration/api-options.mdx
+++ b/docs/docs/reference/typescript-api/configuration/api-options.mdx
@@ -2,7 +2,7 @@
 title: "ApiOptions"
 generated: true
 ---
-<GenerationInfo sourceFile="packages/core/src/config/vendure-config.ts" sourceLine="84" packageName="@vendure/core" />
+<GenerationInfo sourceFile="packages/core/src/config/vendure-config.ts" sourceLine="85" packageName="@vendure/core" />
 
 The ApiOptions define how the Vendure GraphQL APIs are exposed, as well as allowing the API layer
 to be extended with middleware.
@@ -27,6 +27,7 @@ interface ApiOptions {
     middleware?: Middleware[];
     trustProxy?: TrustProxyOptions;
     apolloServerPlugins?: ApolloServerPlugin[];
+    fieldResolverEnhancers?: Enhancer[];
     introspection?: boolean;
     inputValidation?: {
         /**
@@ -217,6 +218,37 @@ allow the extension of the Apollo Server, which is the underlying GraphQL server
 
 Apollo plugins can be used e.g. to perform custom data transformations on incoming operations or outgoing
 data.
+### fieldResolverEnhancers
+
+<MemberInfo kind="property" type={`Enhancer[]`} default={`['guards']`}  since="3.8.0"  />
+
+Controls which NestJS enhancers are applied to GraphQL field resolvers, i.e. resolvers
+declared with `@ResolveField()`. The value is passed through to the `@nestjs/graphql`
+module option of the same name, and applies to both the Shop API and the Admin API.
+
+The default of `['guards']` is what makes the [Allow](/reference/typescript-api/request/allow-decorator#allow) decorator work on a field
+resolver. Interceptors and exception filters are left out because a field resolver runs
+once per resolved field rather than once per operation, so a list query which resolves the
+field on each of 100 items would run every global interceptor 100 extra times.
+
+The consequence of that default is easy to miss: an interceptor registered by a plugin
+with Nest's `APP_INTERCEPTOR` token (or an exception filter registered with `APP_FILTER`)
+runs for top-level queries and mutations only. On a `@ResolveField()` resolver it never
+fires, and nothing is logged to say so. Add `'interceptors'` if a plugin of yours depends
+on one seeing field resolution.
+
+Note that enabling an enhancer here also enables Vendure's own interceptors and exception
+filter on every resolved field, so measure the effect on your heaviest list queries.
+
+*Example*
+
+```ts
+const config: VendureConfig = {
+  apiOptions: {
+    fieldResolverEnhancers: ['guards', 'interceptors'],
+  },
+};
+```
 ### introspection
 
 <MemberInfo kind="property" type={`boolean`} default={`true`}  since="1.5.0"  />
--- a/packages/core/e2e/field-resolver-enhancers.e2e-spec.ts
+++ b/packages/core/e2e/field-resolver-enhancers.e2e-spec.ts
@@ -1,0 +1,96 @@
+import { mergeConfig } from '@vendure/core';
+import { createTestEnvironment, SimpleGraphQLClient } from '@vendure/testing';
+import gql from 'graphql-tag';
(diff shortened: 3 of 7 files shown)


michaelbromley:
Hi,

Thanks for the PR. I do not want to accept this design, however, because allowing the user to modify that NestJS setting makes it very easy to set up unexpected performance issues.

Let's say you add "interceptors" to cover the use case in your original issue. Now the 4 built-in global interceptors we ship in core are all running on every field resolver too, as well as global interceptors defined by any plugin.

Alternatives:

- @vendure/telemetry-plugin. It enables @opentelemetry/instrumentation-graphql by default. That
patches the GraphQL executor and emits a span per resolver, field resolvers included, with timings
and parent/child nesting. You get traces rather than log lines, and it covers field resolvers without
changing the enhancer config.

- An Apollo Server plugin. apiOptions.apolloServerPlugins is already public API, and Apollo Server
has a per-field hook:
```ts
const perfPlugin: ApolloServerPlugin = {
    async requestDidStart() {
        return {
            async executionDidStart() {
                return {
                    willResolveField({ info }) {
                        const start = process.hrtime.bigint();
                        return () => {
                            const ms = Number(process.hrtime.bigint() - start) / 1e6;
                            if (10 < ms) {
                                Logger.info(`${info.parentType.name}.${info.fieldName}: ${ms}ms`);
                            }
                        };
                    },
                };
            },
        };
    },
};
```

`willResolveField` runs for every field. `info.parentType.name` and `info.fieldName` hold the same
values you were reading off `GqlExecutionContext`. The returned function runs when the field
resolves, so measuring the duration needs no observable.

michaelbromley:
However, a PR containing the clarifying callout for the docs is certainly welcome

PR #5277, opened 2026-09-05
fix(core): log stack trace and error class for failed scheduled tasks

### LINKED ISSUE #5276: Scheduled task failures log only `error.message`, so an empty message produces a blank log line
**Describe the bug**

When a `ScheduledTask` throws, `DefaultSchedulerStrategy.runTask()` logs `error.message` and nothing else. No stack trace, no error class, no `cause` chain. If the message is empty, the log line ends after the colon and there is nothing left to act on.

`Logger.error(message, context, trace?)` already takes a stack trace as its third argument. Many other core call sites pass it (`cache.service.ts`, `order.service.ts`, `exception-logger.filter.ts`). The scheduler does not.

https://github.com/vendurehq/vendure/blob/master/packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts#L124-L129

```ts
} catch (error) {
    let errorMessage = 'Unknown error';
    if (error instanceof Error) {
        errorMessage = error.message;
    }
    Logger.error(`Scheduled task "${task.id}" failed with error: ${errorMessage}`);
```

**To Reproduce**

1. Register a scheduled task whose `execute` rejects with an `Error` that has no message, e.g. `throw new SomeError()` where `class SomeError extends Error {}`.
2. Wait for the task to run (or trigger it manually from the Dashboard).
3. Read the server log.

**Expected behavior**

The log line identifies the failure well enough to start debugging: a stack trace, and something non-empty to name the error.

**Actual behavior**

The whole diagnostic output is one line, and it is blank after the colon.

**Error logs**

This is the real output from our production worker on 2026-08-24. It is the complete record of the failure:

```
error Scheduled task "esputnik-sync-contacts" failed with error:
```

The task had been failing on a schedule for some time before anyone noticed there was nothing to read.

- @vendure/core version: 3.7.2 (also present on `master` at 3.7.3)
- Nodejs version: 22
- Browser (if applicable): n/a

**Minimal reproduction**

```ts
import { ScheduledTask } from '@vendure/core';

class SyncError extends Error {}

export const failingTask = new ScheduledTask({
    id: 'failing-task',
    schedule: cron => cron.everyMinute(),
    execute: async () => {
        throw new SyncError();
    },
});
```

The same blank line appears for any error whose message happens to be empty — an HTTP client that throws a response object, a library that constructs errors from an empty upstream body, a rethrow that loses the message and keeps only `cause`.

**Workaround**

We wrapped every registered task's `execute` in a plugin that re-logs the failure with `error.stack` and the `cause` chain before rethrowing. That is ~330 lines to restore what the core log line should already contain. It is also fragile: the wrapper plugin has to be the last entry in the `plugins` array, or tasks registered by later plugins are never wrapped.

**Additional context**

`lastResult` on the `scheduled_task` record stores the same message, so the Dashboard shows the same empty string. Fixing the message fixes both.

Two things would make the line useful, both small:

1. Pass `error.stack` as the third argument to `Logger.error`.
2. Fall back to the error's class name when the message is empty, and append the `cause` chain — `cause` does not appear in `error.stack`.

Note that for `class SyncError extends Error {}` the stack header reads `Error:`, not `SyncError:`, so the class name is genuinely extra information rather than a duplicate of the trace.

brmk:
# Description

Fixes #5276.

When a `ScheduledTask` throws, `DefaultSchedulerStrategy.runTask()` logs `error.message` and nothing else. If the message is empty, the log line ends after the colon.

This is the complete output our production worker gave us for a failing task on 2026-08-24:

```
error Scheduled task "esputnik-sync-contacts" failed with error:
```

No stack, no error class, no `cause`.

## What changed

The failure path now logs `util.inspect(error)`.

Node already renders everything this needs, so there is no hand-rolled formatting in the fix. Checked on both ends of the supported range (`^20.19.0 || >=22.12.0`) — Node v20.19.0 and v24.14.1 produce identical output:

| case | `util.inspect` output |
| --- | --- |
| `class EsputnikSyncError extends Error {}` thrown with `''` | header is `EsputnikSyncError` — the constructor name, even though `error.name` is `'Error'` and the stack header is bare |
| `cause` chain `Error` → `Error` → `'ECONNRESET'` | all three rendered, ending in `[cause]: 'ECONNRESET'` |
| `cause` cycle | `<ref *1>` / `[Circular *1]`, no infinite loop |
| non-`Error` throw (string, plain object, `null`) | `'boom'`, `{ code: 'ECONNRESET', errno: -54 }`, `null` |
| `AggregateError` | `[errors]: [ ... ]` with each sub-error |
| extra own properties | appended, e.g. `{ status: 502, url: '...' }` |

The first version of this PR walked the `cause` chain by hand. As @gabriellbui pointed out in review, that walk stopped as soon as a `cause` was not an `Error` — which is exactly the `error.cause = 'ECONNRESET'` case. `inspect` handles it, so the hand-rolled helper is gone.

**No `trace` argument is passed to `Logger.error` any more.** `DefaultLogger` renders `trace` by appending it to the message (`message + '\n' + trace`), and the `inspect` output already contains the stack, so passing `error.stack` as well would print the stack twice.

**`lastResult` is deliberately *not* the `inspect` output.** That value is persisted on the `scheduled_task` record and surfaced in the Dashboard, where a multi-line stack dump does not belong. It keeps a short single-line label: `error.message`, falling back to the constructor name (`constructor.name` rather than `name`, because a subclass that does not set `name` still inherits `'Error'` from `Error.prototype`). So the Dashboard's error column stops being blank without becoming a wall of text.

## Before / after

Both lines are the real output of the new e2e case, for a task throwing `class EmptyMessageError extends Error {}` with `error.cause = 'ECONNRESET'`.

Before (on `master`):

```
Scheduled task "error-test-job" failed with error:
```

After:

```
Scheduled task "error-test-job" failed with error: EmptyMessageError:
    at Object.execute (.../packages/core/e2e/default-scheduler-plugin.e2e-spec.ts:47:43)
    at ScheduledTask.execute (.../packages/core/src/scheduler/scheduled-task.ts:141:28)
    at DefaultSchedulerStrategy.runTask (.../packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts:114:28)
    ... {
  cause: 'ECONNRESET'
}
```

## Tests

Per review, the separate `default-scheduler-strategy.spec.ts` added by the first version of this PR is removed, and coverage lives in the existing `packages/core/e2e/default-scheduler-plugin.e2e-spec.ts` instead: one task that throws an empty-message subclass with a string `cause`, asserted through `TestingLogger.errorSpy`, no mocks beyond the task itself. It also asserts that `trace` is undefined and that `lastResult` is the short `{ error: 'EmptyMessageError' }` label.

The case fails on `master` with `expected 'Scheduled task "error-test-job" failed with error: ' to contain 'EmptyMessageError'`, and passes with the fix.

`packages/core`: e2e spec green (7 tests), unit suite green (82 files, 1251 tests). `bun run build:core-common` green.

# Breaking changes

None. For an error with a non-empty message, the log line keeps the same prefix and gains the stack, any `cause` chain and any extra properties. `lastResult.error` is unchanged except for errors whose message is empty, where it now carries the class name instead of an empty string.

# Screenshots

None — the log output above is the relevant evidence.

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts
+++ b/packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts
@@ -1,6 +1,7 @@
 import { UpdateScheduledTaskInput } from '@vendure/common/lib/generated-types';
 import { Cron } from 'croner';
 import ms, { type StringValue } from 'ms';
+import { inspect } from 'node:util';
 
 import { Injector } from '../../common';
 import { assertFound } from '../../common/utils';
@@ -122,17 +123,13 @@ export class DefaultSchedulerStrategy implements SchedulerStrategy {
             );
             Logger.verbose(`Scheduled task "${task.id}" completed successfully`);
         } catch (error) {
-            let errorMessage = 'Unknown error';
-            if (error instanceof Error) {
-                errorMessage = error.message;
-            }
-            Logger.error(`Scheduled task "${task.id}" failed with error: ${errorMessage}`);
+            Logger.error(`Scheduled task "${task.id}" failed with error: ${inspect(error)}`);
             await this.connection.rawConnection.getRepository(ScheduledTaskRecord).update(
                 { taskId: task.id },
                 {
                     lastExecutedAt: new Date(),
                     lockedAt: null,
-                    lastResult: { error: errorMessage } as any,
+                    lastResult: { error: errorLabel(error) } as any,
                 },
             );
         } finally {
@@ -356,4 +353,15 @@ export class DefaultSchedulerStrategy implements SchedulerStrategy {
             this.tasks.set(taskId, { task: task.task, isRegistered: true });
         }
     }
+}
+
+/**
+ * `constructor.name` rather than `name` because a subclass which does not set `name`
+ * still inherits `'Error'` from `Error.prototype`.
+ */
+function errorLabel(error: unknown): string {
+    if (!(error instanceof Error)) {
+        return 'Unknown error';
+    }
+    return error.message || error.constructor.name;
 }
--- a/packages/core/e2e/default-scheduler-plugin.e2e-spec.ts
+++ b/packages/core/e2e/default-scheduler-plugin.e2e-spec.ts
@@ -1,5 +1,5 @@
 import { ConfigService, DefaultSchedulerPlugin, mergeConfig, ScheduledTask } from '@vendure/core';
-import { createTestEnvironment } from '@vendure/testing';
+import { createTestEnvironment, TestingLogger } from '@vendure/testing';
 import path from 'path';
 import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
 
@@ -15,8 +15,11 @@ const MAX_HOLD_MS = 5_000;
 
 const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
 
+class EmptyMessageError extends Error {}
+
 describe('Default scheduler plugin', () => {
     const taskSpy = vi.fn();
+    const testingLogger = new TestingLogger(() => vi.fn());
     // One task per hold-window test so DB state can't leak between them.
     const holdSpyBlocking = vi.fn();
     const holdSpyManual = vi.fn();
@@ -32,6 +35,16 @@ describe('Default scheduler plugin', () => {
                         async execute(injector) {
                             taskSpy();
                             return { success: true };
+                        },
+                    }),
+                    new ScheduledTask({
+                        id: 'error-test-job',
+                        description: 'A test job which throws an error with an empty message',
+                        schedule: cron => cron.everySaturdayAt(0, 0),
+                        async execute(injector) {
+                            const error = new EmptyMessageError('');
+                            (error as Error & { cause?: unknown }).cause = 'ECONNRESET';
+                            throw error;
                         },
                     }),
                     new ScheduledTask({
@@ -56,6 +69,7 @@ describe('Default scheduler plugin', () => {
                 runTasksInWorkerOnly: false,
             },
             plugins: [DefaultSchedulerPlugin.init({ manualTriggerCheckInterval: 50 })],
+            logger: testingLogger,
         }),
     );
 
@@ -78,7 +92,7 @@ describe('Default scheduler plugin', () => {
 
     it('get tasks', async () => {
         const { scheduledTasks } = await adminClient.query(getTasksDocument);
-        expect(scheduledTasks.length).toBe(3);
+        expect(scheduledTasks.length).toBe(4);
         const testJob = scheduledTasks.find(t => t.id === 'test-job');
         if (!testJob) throw new Error('test-job not found');
         expect(testJob.description).toBe("A test job that doesn't do anything");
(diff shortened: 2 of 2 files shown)


packages/core/src/plugin/default-scheduler-plugin/default-scheduler-strategy.ts
```diff
@@ -357,3 +358,24 @@ export class DefaultSchedulerStrategy implements SchedulerStrategy {
...
+    const seen = new Set<unknown>();
+    let current: unknown = error;
+    while (current instanceof Error && !seen.has(current)) {
+        seen.add(current);
+        parts.push(current.message || current.constructor.name || 'Error');
+        // `Error.cause` is not in the `lib` used by this package, but is populated at runtime.
+        current = (current as Error & { cause?: unknown }).cause;
+    }
+    return parts.join(' caused by: ');
+}
```
gabriellbui: I think it stops as soon as `error.cause` is not an error. For example: `error.cause = 'ECONNRESET'`

I think node already does what you are trying to do here and also handles the above scenario. Could you check if [inspect](https://nodejs.org/api/util.html#utilinspectobject-options) from node can fully replace this? 

If yes, one case in `default-scheduler-plugin.e2e-spec.ts` with a task that throws an empty-message subclass, asserting on TestingLogger.errorSpy without any mocks should be sufficient. We don't need a new spec file for this.

brmk: Good catch on the non-Error cause — that was exactly the hole, and `inspect` covers it. Checked on Node 20.19 and 24.14 (both ends of the supported range), identical output for all of: empty-message subclass, `cause` chain ending in a string, `cause` cycles, non-Error throws, and `AggregateError`.

Concretely, for `class EmptyMessageError extends Error {}` with `error.cause = 'ECONNRESET'`:

```
- Scheduled task "error-test-job" failed with error:
+ Scheduled task "error-test-job" failed with error: EmptyMessageError:
+     at Object.execute (...)
+     ... {
+   cause: 'ECONNRESET'
+ }
```

So `describeError` is gone. I also dropped the `trace` argument, since `DefaultLogger` appends it to the message and `inspect` already includes the stack — passing both printed it twice.

One deliberate split: `lastResult.error` does *not* get the `inspect` blob. It's persisted and shown in the Dashboard, so it keeps a short single-line label (`error.message` falling back to `constructor.name` — `name` is still `'Error'` on a subclass that doesn't set it).

Spec file removed; the case now lives in `default-scheduler-plugin.e2e-spec.ts` asserting on `errorSpy`, no mocks. It fails on master with `expected '...failed with error: ' to contain 'EmptyMessageError'`.

gabriellbui (changes requested):
@brmk thanks for the change, and for the rework after the first round.

The code is good. One request before approval, and one nit.

**Remove the comments.** Each one repeats what the code below it shows, or defends a decision that belongs in the commit message. In `default-scheduler-strategy.ts`, the block above `Logger.error`. In the e2e spec, the one above `EmptyMessageError`, the `Error.cause` note in `execute`, the three "...the class name... ...the cause... ...the stack..." lines, and the "persisted record" line. Keep a single `// #5276` above the test. The `errorLabel` JSDoc can stay, trimmed to the one sentence about why `constructor.name` and not `name`.

**Nit.** After `instanceof Error`, `constructor` always exists and `constructor.name` is only empty for a class written with no name at all, so `?.` and `|| 'Error'` cover nothing. Fine to drop them, fine to leave them.

brmk:
Done in 433c9577e7: comments removed as listed, `errorLabel` trimmed, and the dead `?.`/`|| 'Error'` dropped.

gabriellbui (approved):

PR #5279, opened 2026-09-05
fix(core): do not record promotions that apply no discount

### LINKED ISSUE #5278: Promotions that apply zero discount still consume usageLimit / perCustomerUsageLimit
**Describe the bug**

A Promotion whose conditions pass but whose actions produce no discount is still attached to the Order, and still consumes `usageLimit` / `perCustomerUsageLimit`.

**To Reproduce**

1. Create a Promotion with a coupon code, `perCustomerUsageLimit: 1`, no conditions, and a single `products_percentage_discount` action limited to variant A.
2. As a customer, add variant B to the cart.
3. Apply the coupon code. It is accepted. `order.discounts` is empty and the total does not change.
4. Complete the order and pay.
5. Start a new order, add variant A, apply the same coupon code.

**Expected behavior**

Step 5 succeeds. The customer never received a discount in step 3, so no usage was consumed.

**Actual behavior**

Step 5 returns `CouponCodeLimitError` — "Coupon code cannot be used more than once per customer". The order placed in step 4 holds a row in `order_promotions_promotion` despite `order.discounts` being empty.

**Where it happens**

`OrderCalculator.applyOrderItemPromotions` calls `this.addPromotion(order, promotion)` outside the `if (adjustment)` branch, so passing `Promotion.test()` is enough to record the promotion. `applyOrderPromotions` and `applyShippingPromotions` do the same. `PromotionService` then counts usages off those rows — `countPromotionUsages`, `countPromotionUsagesForCustomer` and `getUsageCountsBatch` all `innerJoin('order.promotions')` and never look at whether an adjustment exists.

- @vendure/core version: 3.7.2, re-verified against master at 523b573
- Nodejs version: 22
- Database: postgres
- Operating System: Linux
- Package manager: npm

**Additional context**

We hit this on our production store on 2026-08-05. We were running a coupon promotion whose actions used facet-based exclusions, so a cart holding only excluded items passed the conditions and received nothing. 89 orders were placed with that coupon that day. 51 of them got zero discount, and every one of those 51 consumed the customer's single allowed use. Those customers applied a coupon, got nothing, and could never use it again. Support tickets followed.

The data gives you no way to tell the two apart. The join row for an order that got 50% off and the join row for an order that got nothing are identical; you have to load the order's lines and shipping lines and sum the adjustments to find out.

Free-shipping promotions are the case to be careful with: a promotion can legitimately discount nothing on the items and still discount the shipping, so "did this promotion do anything" has to look at shipping lines too. Actions with `onActivate` side effects are the other one — the documented free-gift action returns 0 from `execute()` until its side effect has added the gift line, so it has to stay recorded on the first pass. It also has to stay recorded when the side effect subsequently fails: the documented action logs the error and continues if the gift cannot be added, and nothing in the return value distinguishes that from success, so a side-effect promotion keeps consuming a usage either way.

**Workaround**

We delete the zero-discount join rows on `OrderPlacedEvent` and let Vendure's own counting stay authoritative. It only affects orders placed after we deployed it — the rows already in the table stay, and a promotion exhausted before then stays exhausted. It works, but it lives outside `@vendure/core` and has to survive every upgrade, which is why we would rather fix it upstream.

brmk:
# Description

Fixes #5278.

A Promotion whose conditions pass but whose actions produce no adjustment anywhere is still attached to the Order, so it still consumes `usageLimit` / `perCustomerUsageLimit`. The full bug report is at the bottom of this description; the short version is that a customer who applies a coupon and gets no discount can never use that coupon again.

`OrderCalculator.applyPriceAdjustments` now ends with `removeIneffectivePromotions(order)`, which keeps a Promotion on `order.promotions` only if it adjusted an OrderLine or a ShippingLine.

Design notes, in case any of these are the wrong call:

**The check is at the end of the calculation, not at the `addPromotion()` call sites.** There are three of them (`applyOrderItemPromotions`, `applyOrderPromotions`, `applyShippingPromotions`) and a Promotion that discounts nothing on the first OrderLine may still discount the third one, the Order total, or the shipping. Filtering once at the end covers all three targets and leaves the behaviour during the calculation exactly as it was, in case a `PromotionCondition` reads `order.promotions` mid-run.

**"Produced an adjustment" is read from the raw `adjustments` arrays on the lines and shipping lines,** not from `Order.discounts`. Both `OrderLine.addAdjustment()` and `applyShippingPromotions` already refuse to store a zero-amount adjustment, so the presence of an adjustment is itself the signal. Going through `Order.discounts` would have dragged in the quantity and tax maths, which can round a real discount to a zero-amount `Discount` on a fully-cancelled line.

**Promotions with side effects are kept even at zero discount.** This is the subtle case. The documented free-gift plugin (`docs/docs/guides/core-concepts/promotions/index.mdx`) has an action whose `execute()` returns 0 until the gift line exists, and an `onActivate()` that adds it. `PromotionService.runPromotionSideEffects()` decides what to activate by diffing `order.promotions`, so filtering on discount alone would mean the promotion is never recorded, `onActivate` never runs, the gift is never added, and `execute()` keeps returning 0 forever. `PromotionAction.hasSideEffects` and `Promotion.hasSideEffects` were added so the filter can make that exception. Such a promotion consumes a usage even when it discounts nothing, which seems right — the side effect ran.

The case that exemption does not cover is a side effect that runs and *fails*: the documented free-gift `onActivate` logs the error and continues when the gift variant cannot be added, so an Order whose gift was out of stock keeps the promotion and burns a usage having received nothing. Distinguishing that would mean giving `onActivate` a way to report failure back to the calculator, which is a wider API change than this fix. Promotions with side effects therefore keep the old behaviour by design, and only the plain zero-discount case changes.

**No config option or strategy gates the new behaviour.** Recording a promotion the customer received nothing from is not a policy any store would deliberately choose — it is simply the bug — so a flag would preserve it only for the stores that have not noticed yet, and would leave `PromotionService`'s usage counting with two meanings instead of one. The contract is documented in the promotions guide next to the free-gift example, because it does have one consequence for plugin authors: a custom action that delivers its benefit by mutating the Order inside `execute()` and returns 0 should declare a no-op `onActivate` to keep being recorded.

**Coupon codes are untouched.** `order.couponCodes` is separate from `order.promotions`. A code the customer typed stays on the Order even when it discounts nothing: it is still a valid code, `Promotion.test()` still passes, and removing it would be a surprising, unexplained change in the cart. It just no longer counts as a usage. Concretely, a customer can now apply a zero-effect coupon, place the order, and apply the same coupon again on their next order.

**Removal still works.** `applyPriceAdjustments` starts from `order.promotions = []` and rebuilds the list, so the filter is a narrower rebuild, not a new deletion path. A promotion that was never recorded is simply absent from both sides of the `runPromotionSideEffects` diff, and `deactivate()` on a promotion without side-effect functions is a no-op.

Tests:

- `order-calculator.spec.ts` — a new `promotions which apply no discount` block: not added when nothing is discounted; added when one of several lines is discounted; removed without disturbing the other Promotion applied to the same Order; removed once it stops discounting; added when it only discounts shipping; added when its action has a side effect.
- `order-promotion.e2e-spec.ts` — a zero-discount coupon with `perCustomerUsageLimit: 1`: the placed Order has no promotions, and the same customer can apply the coupon again on a cart where it does discount something.

The regression tests fail on `master` (three of the `order-calculator.spec.ts` cases and the e2e case), and the remaining "must still be attached" cases pass either way, so they only guard against over-removal.

Verified with:

```
packages/core $ npm test                      # 82 files, 1257 tests
packages/core $ npm run e2e -- promotion      # 4 files, 100 tests, 1 skipped
packages/core $ npm run e2e -- order          # 21 files, 503 tests, 1 skipped
```

`order-promotion.e2e-spec.ts` and `promotion-side-effects.e2e-spec.ts` were additionally run against postgres 16 and mysql 8 (73 tests each) as well as sqljs.

`tsc -p ./build/tsconfig.build.json --noEmit` is clean, `docs $ npm run test:mdx` passes, and `eslint` on the touched files reports only warnings that are already present on `master`.

# Breaking changes

No API or schema change, but it is a change in current behaviour, so flagging it explicitly:

- `Order.promotions` no longer lists a Promotion that discounted nothing. Any storefront reading that field to decide whether to show "coupon applied" needs `order.couponCodes` instead — which is the accurate source anyway.
- Consequently `usageLimit` and `perCustomerUsageLimit` are consumed less often. This applies to new calculations only: the existing rows in `order_promotions_promotion` are not migrated, so a promotion already exhausted by past zero-discount orders stays exhausted until those orders are recalculated — which in practice means an admin-API order modification, not an upgrade.
- `OrderModifier.modifyOrder` re-runs `applyPriceAdjustments` on a placed Order and saves the result, so a modification that removes the discounted OrderLine now also drops the Promotion and releases the usage. That is arguably the right reading of "this Order no longer benefits from the promotion", but it is a behaviour change worth naming.

Happy to retarget this at `minor` or `major` if you consider that a behaviour break rather than the fix.

# Screenshots

n/a

# Checklist

📌 Always:

👍 Most of the time:

---

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5279"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791198797&installation_model_id=20193&pr_number=5279&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5279&signature=cc00ffbd37fdcbd87790c50a12b5518d322ddcf90ca74e1d45987228fec2c4be"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/guides/core-concepts/promotions/index.mdx
+++ b/docs/docs/guides/core-concepts/promotions/index.mdx
@@ -260,6 +260,10 @@ function lineContainsIds(ids: ID[], line: OrderLine): boolean {
 export class FreeGiftPromotionPlugin {}
 ```
 
+:::note[Promotions which discount nothing are not recorded]
+A Promotion is only recorded on `Order.promotions` — and therefore only consumes its `usageLimit` and `perCustomerUsageLimit` — if it actually adjusted an OrderLine or a ShippingLine. The exception is a Promotion with at least one action defining `onActivate`/`onDeactivate`: the free gift action above returns `0` from `execute()` until its side effect has added the gift line, so such a Promotion is kept whatever it discounts. If you write a custom action which delivers its benefit by mutating the Order inside `execute()` and returns `0`, declare a no-op `onActivate` so that the Promotion continues to be recorded.
+:::
+
 ### Dependency relationships
 
 It is possible to establish dependency relationships between a PromotionAction and one or more PromotionConditions.
--- a/packages/core/e2e/order-promotion.e2e-spec.ts
+++ b/packages/core/e2e/order-promotion.e2e-spec.ts
@@ -2218,6 +2218,93 @@ describe('Promotions applied to Orders', () => {
         });
     });
 
+    describe('promotion which applies no discount', () => {
+        const couponCode = 'NO_DISCOUNT_ON_THIS_CART';
+        let promotion: PromotionFragment;
+
+        beforeAll(async () => {
+            promotion = await createPromotion({
+                enabled: true,
+                name: '50% off item-5000 only',
+                couponCode,
+                perCustomerUsageLimit: 1,
+                conditions: [],
+                actions: [
+                    {
+                        code: productsPercentageDiscount.code,
+                        arguments: [
+                            { name: 'discount', value: '50' },
+                            {
+                                name: 'productVariantIds',
+                                value: `["${getVariantBySlug('item-5000').id}"]`,
+                            },
+                        ],
+                    },
+                ],
+            });
+        });
+
+        afterAll(async () => {
+            await deletePromotion(promotion.id);
+        });
+
+        function addGuestCustomerToOrder() {
+            return shopClient.query(setCustomerDocument, {
+                input: {
+                    emailAddress: 'no-discount-guest@test.com',
+                    firstName: 'No Discount',
+                    lastName: 'Customer',
+                },
+            });
+        }
+
+        it('is not added to the Order and does not consume a usage', async () => {
+            await shopClient.asAnonymousUser();
+            await shopClient.query(addItemToOrderDocument, {
+                // Not one of the Promotion's ProductVariants, so the action discounts nothing
+                productVariantId: getVariantBySlug('item-1000').id,
+                quantity: 1,
+            });
+            const { setCustomerForOrder } = await addGuestCustomerToOrder();
+            orderResultGuard.assertSuccess(setCustomerForOrder);
+
+            const { applyCouponCode } = await shopClient.query(applyCouponCodeDocument, { couponCode });
+            orderResultGuard.assertSuccess(applyCouponCode);
+
+            // The customer typed the code, so it stays on the Order even though it discounts nothing
+            expect(applyCouponCode.couponCodes).toEqual([couponCode]);
+            expect(applyCouponCode.discounts).toEqual([]);
+            const orderCode = applyCouponCode.code;
+
+            await proceedToArrangingPayment(shopClient);
+            const order = await addPaymentToOrder(shopClient, testSuccessfulPaymentMethod);
+            orderResultGuard.assertSuccess(order);
+            expect(order.state).toBe('PaymentSettled');
+
+            const { orderByCode } = await shopClient.query(getOrderPromotionsByCodeDocument, {
+                code: orderCode,
+            });
+            expect(orderByCode!.promotions).toEqual([]);
+
+            // The customer's single allowed usage is still available
+            await shopClient.asAnonymousUser();
+            await shopClient.query(addItemToOrderDocument, {
+                productVariantId: getVariantBySlug('item-5000').id,
+                quantity: 1,
+            });
+            const { setCustomerForOrder: setCustomerForSecondOrder } = await addGuestCustomerToOrder();
+            orderResultGuard.assertSuccess(setCustomerForSecondOrder);
+
+            const { applyCouponCode: secondUse } = await shopClient.query(applyCouponCodeDocument, {
+                couponCode,
+            });
+            orderResultGuard.assertSuccess(secondUse);
+
+            expect(secondUse.discounts.length).toBe(1);
+            expect(secondUse.totalWithTax).toBe(3000);
+        });
+    });
+
     // https://github.com/vendurehq/vendure/issues/710
     it('removes order-level discount made invalid by removing OrderLine', async () => {
         const promotion = await createPromotion({
--- a/packages/core/src/config/promotion/promotion-action.ts
+++ b/packages/core/src/config/promotion/promotion-action.ts
@@ -306,6 +306,17 @@ export abstract class PromotionAction<
         this.onDeactivateFn = config.onDeactivate;
     }
 
+    /**
+     * @description
+     * Whether this action defines any of the side-effect functions. Such an action can have an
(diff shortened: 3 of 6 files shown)


biggamesmallworld (approved):
Approving. Verified locally against sqljs at 4cbbc79.

- `order-calculator.spec.ts`: 54 pass, including the six new cases.
- `order-promotion.e2e-spec.ts`: 67 pass, 1 skipped.
- Confirmed the e2e case is a real regression test rather than a passing assertion: with `removeIneffectivePromotions(order)` commented out and core rebuilt, it fails on `expected [ { id: 'T_27' } ] to deeply equal []` and the rest of the suite stays green.

Two non-blocking notes.

The docs note tells plugin authors to declare a no-op `onActivate` so their action keeps being recorded. A function whose only purpose is to be detected by `!!this.onActivateFn` reads as dead code, so the next person to touch that plugin deletes it and nothing fails to tell them. Worth rewording at least, or moving the signal to an explicit flag on `PromotionActionConfig`.

`applyOrderPromotions` has no zero-discount test. Low risk, since the filter compares `adjustment.adjustmentSource` against `getSourceId()` with no per-path branching, and the existing `orderPercentageDiscount` / `orderFixedDiscount` blocks already cover the over-removal direction.

One thing worth a decision rather than a fix: `master` back-merges to `minor`, so this ships in a patch. `OrderModifier.modifyOrder` re-runs the calculation on a placed Order and saves, so an admin edit to a settled order can now drop a Promotion and release a usage. That is the only path by which existing join rows move, and you have already flagged it in the description. Happy either way, but it is the piece I would want a maintainer to sign off on explicitly.

PR #3614, opened 2025-06-25
perf(jobs,orders) added indexes and change default sort order for jobs

oidt:
# Description

Performance issues on jobs createdAt index

# Breaking changes

No

# Screenshots

You can add screenshots here if applicable.

# Checklist

📌 Always:
- [X] I have set a clear title
- [X] My PR is small and contains a single feature
- [X] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ ] I have added or updated test cases
- [ ] I have updated the README if needed

## Summary by CodeRabbit

- **New Features**
	- Improved performance for order and job queue listings by adding database indexes on relevant date fields.
	- The job queue list now defaults to sorting by creation date in descending order for easier viewing of recent jobs.

PR #5287, opened 2026-09-05
perf(core): add missing indexes on STI discriminators and productId

### LINKED ISSUE #5286: STI discriminator indexes are silently discarded for entities with custom fields; search_index_item.productId is unindexed
**Describe the bug**

Five core columns are missing indexes.

Four of them are single-table-inheritance discriminators, and TypeORM already tries to index them.
`EntityMetadataBuilder.createKeysForTableInheritance()` pushes an `IndexMetadata` for the
discriminator column of every STI entity, but for some entities that index never reaches the schema.
Which ones it reaches depends on whether the entity has an embedded column:

`computeEntityMetadataStep2()` rebuilds the index array as

```ts
entityMetadata.indices = entityMetadata.embeddeds.reduce(
    (indices, embedded) => indices.concat(embedded.indicesFromTree),
    entityMetadata.ownIndices,
);
```

With zero embeddeds `reduce` returns the seed — `ownIndices` itself, by reference — so the index
pushed by `createKeysForTableInheritance()` is also in `ownIndices` and survives. With at least one
embedded, `concat()` returns a new array, the push lands on that copy, and the next rebuild starts
from `ownIndices` again and discards it. `build()` re-runs `computeEntityMetadataStep2()` for every
column with a generation strategy, and every Vendure entity's `id` has one, so the rebuild always
happens.

Net effect: an STI base entity that has a `customFields` embed loses its automatic discriminator
index.

| STI table | `customFields` embed | Discriminator indexed |
|---|---|---|
| `authentication_method` | no | yes |
| `order_line_reference` | no | yes |
| `history_entry` | yes | **no** |
| `session` | yes | **no** |
| `region` | yes | **no** |
| `stock_movement` | yes | **no** |

A minimal standalone STI entity has no embeds and does get the index, so this does not reproduce
outside an app that uses embedded columns.

Separately, `search_index_item.productId` has no index at all. It is not a relation, so there is no FK
index, and it is not part of the primary key (`languageCode`, `productVariantId`, `channelId`) nor a
prefix of any other index — while `IndexerController.removeSyntheticVariants()` deletes by
`{ productId, sku: '', price: 0 }` per product and the MySQL search strategy joins its term-score
subquery on `inner_productId = si.productId`.

**To reproduce**

1. Start Vendure against Postgres with `synchronize: true` (or generate a migration).
2. `\d history_entry`, `\d session`, `\d region`, `\d stock_movement`, `\d search_index_item`.
3. No index on the discriminator columns, or on `search_index_item.productId`.
4. Contrast with `\d authentication_method` and `\d order_line_reference`, which do have theirs.

**Expected behavior**

The discriminator columns are indexed — TypeORM's own `createKeysForTableInheritance()` says that is
the intent — and `search_index_item.productId`, which the indexer filters on, is indexed.

**Actual behavior**

Four of the six STI discriminator columns, and `search_index_item.productId`, are plain unindexed
columns.

**Environment**

- @vendure/core version: 3.7.2 (also reproduced on current `minor`)
- Nodejs version: 22
- Database: postgres
- Operating System: Linux
- Package manager: npm

**Additional context**

Production store, `history_entry` ~2M rows and `session` ~1.5M rows before cleanup. On 2026-05-27 an
admin-list COUNT over history entries sequentially scanned the table during peak traffic, pinning the
database and slowing checkout; `EXPLAIN` pointed at the discriminator filter.

We now maintain these indexes out of band — created by hand in the database and documented
internally — because they sit on core-owned tables, and `migration:generate` keeps emitting
`DROP INDEX` for anything core's entity metadata does not know about. Fighting that drift on every
migration is the reason the indexes belong in the entity definitions upstream.

PR #5287 fixes this: it adds `@Index(['discriminator'])` / `@Index(['type'])` at the entity level for
all four affected STI entities (`HistoryEntry`, `Session`, `Region`, `StockMovement`) and `@Index()`
on `SearchIndexItem.productId`. An explicit `@Index` lands in `ownIndices` and therefore survives the
metadata rebuild, and it also trips TypeORM's existing dedup guard so no duplicate index is created.

Upgrade note for anyone hitting this before the fix ships: TypeORM emits a plain `CREATE INDEX`, which
on PostgreSQL takes a `SHARE` lock and blocks writes for the duration of the build. On a large table,
create the index by hand with `CREATE INDEX CONCURRENTLY` under the name TypeORM generates, so the
generated migration finds it already present. Besides `history_entry`, that matters most for
`stock_movement`, which gains a row for every stock movement of every order line and so grows with
order volume without ever being pruned.

brmk:
# Description

Fixes #5286.

Adds indexes on `history_entry.discriminator`, `session.type`, `region.discriminator`, `stock_movement.discriminator` and `search_index_item.productId`.

TypeORM is supposed to index every STI discriminator. `createKeysForTableInheritance()` pushes the index onto `entityMetadata.indices`, but for an entity with an embedded column that array is a copy, and `computeEntityMetadataStep2()` later rebuilds it from `ownIndices`, which drops the index. The rebuild always runs, because `VendureEntity.id` has a generation strategy. Every core STI base with a `customFields` embed loses its discriminator index. `authentication_method` and `order_line_reference` have no embed and keep theirs. An explicit `@Index()` goes into `ownIndices`, so it is kept, and the existing duplicate check stops TypeORM adding a second one. The code path is the same in TypeORM 1.1.0. Details in #5286.

`search_index_item.productId` never had an index. The indexer deletes by it once per product in every reindex batch (`removeSyntheticVariants()`), and the MySQL search strategy joins on it.

What the indexes are used for:

- `search_index_item.productId`: the reindex deletes above. This is the only one whose effect grows with data size.
- `region.discriminator`: the only filter in the countries and provinces lists, but on a small table.
- `history_entry.discriminator`, `session.type`, `stock_movement.discriminator`: no core query uses them. Core does query subclasses such as `Allocation` and `AuthenticatedSession`, but always together with an indexed `orderLineId`, `userId` or `token`. They restore the TypeORM default for code that queries through a child repository, where TypeORM adds the discriminator filter automatically.

# Breaking changes

Yes, same as #3614: users need to run a DB migration that only adds indexes. TypeORM generates a plain `CREATE INDEX`, which blocks writes on PostgreSQL while it builds. On large tables you can create the index with `CREATE INDEX CONCURRENTLY` first, using the name from the generated migration.

Please add the `typeorm-v1` label.

# Checklist

📌 Always:

👍 Most of the time:

### Tests

`entity-indexes.spec.ts` builds the schema on sqljs, reads it back with `QueryRunner.getTables()`, and checks that exactly one index starts with each target column. `authentication_method.type` and `order_line_reference.discriminator` are controls. With the new decorators removed, the five targets fail and the controls pass.


biggamesmallworld:
Merge lane session `c3882454` (@biggamesmallworld) is working this pull request.

Current stage: Stage 7: push. Claim expires 2026-09-15T12:17:02Z unless refreshed.

biggamesmallworld:
Thanks for this, @brmk. We verified the change independently and it holds up. We cannot push to this branch, because GitHub does not offer maintainer edits on forks owned by an organisation.

What we verified, against `minor` at 396df71e6:

- On `minor`, four STI discriminator columns and `search_index_item.productId` have no index on postgres, mariadb and sqljs. `authentication_method.type` and `order_line_reference.discriminator` keep their automatic index, which matches the table in #5286.
- With the change applied, all five columns are indexed. A postgres `history_entry` count at 2M rows goes from a sequential scan (26,667 buffers, 37.9 ms) to an index-only scan (91 buffers, 6.3 ms).
- The generated migration is five `CREATE INDEX` statements on postgres and mariadb, with no drops or renames. It applies, no-ops on a second run, reverts, and re-applies.
- Core e2e passes on sqljs, postgres and mariadb 11.5 (490, 491 and 491 tests). Core unit: 1548 tests.
- The locking warning in your commit footer is accurate. On postgres at 2M rows the index build took 1087 ms and a concurrent insert waited 692 ms. MariaDB InnoDB did not block.

Two changes we would like on the branch:

1. `entity-indexes.spec.ts` reads `getMetadataArgsStorage().indices`, which holds decorator arguments. The defect happens after those are read: `createKeysForTableInheritance()` pushes an index for the discriminator, and `computeEntityMetadataStep2()` then rebuilds `indices` from `ownIndices` and discards it for any entity with an embedded column. The spec therefore stays green even when the column ends up unindexed. Build the schema on sqljs and read the indexes back instead, assert exactly one index per column, and add `authentication_method` and `order_line_reference` as a control group.
2. The description claims more benefit than the queries show. Captured SQL shows core never filters on `history_entry.discriminator` or `stock_movement.discriminator`, and session resolution is served from cache. `region.discriminator` is the sole predicate of the countries list, and `search_index_item.productId` is used directly. No query gets slower, so this affects the description rather than the change.

We have our version of the spec change ready as a patch if you want it.

One question for you: CONTRIBUTING says schema changes target `major`. This PR targets `minor` with a BREAKING footer, which follows #3614. We are keeping it on `minor`.

brmk:
Both done.

1. `entity-indexes.spec.ts` builds the schema on sqljs and reads indexes through `QueryRunner.getTables()`. It asserts exactly one index leads with each column, with `authentication_method.type` and `order_line_reference.discriminator` as controls. With the five `@Index` lines reverted, the five targets fail and the controls pass.
2. Description rewritten. One correction: `stock_movement.discriminator` and `session.type` do appear in core queries (the `Allocation`/`Sale`/`Release`/`Cancellation` lookups in `order-modifier.ts`, and `deleteSessionsByUser`), but always alongside an indexed `orderLineId` or `userId`, so the discriminator index is never chosen. `history_entry` is as you described.

I also shortened the `BREAKING CHANGE` footer, since it goes into the changelog as is.

TypeORM 1.1.0 has the same code path (`EntityMetadataBuilder`: the discriminator index is pushed in `createKeysForTableInheritance()` and dropped by the step-2 rerun), so a `typeorm-v1` run should agree.

brmk:
Following up on my own note above: I went ahead and trimmed the `BREAKING CHANGE` footer rather than leaving it for you to decide, so the branch is now `5d1fef3` + `6d3ef02` (was `faa2202` + `210aad6`).

It is a message-only rewrite. The fix commit's diff is byte-identical to the one you verified, and the branch tree is unchanged — `git diff 210aad6 HEAD` is empty. Only two things changed in the message: the footer is three sentences instead of ten, keeping the migration requirement and the `CREATE INDEX CONCURRENTLY` pointer and dropping the `SHARE` lock / InnoDB detail that is already in the PR body, and the commit now carries `Fixes #5286`, which it was missing. The body also gained a short paragraph stating which of the five indexes actually serve a query, so the commit does not claim more than the PR body admits.

Sorry for the extra round — I would rather you re-read one footer than have that land in `CHANGELOG.md` at four times the length of #3614's.

PR #5295, opened 2026-09-05
feat(dashboard): add defaultColumnFilters to ListPage

### LINKED ISSUE #5294: ListPage has no way to declare default column filters
**`ListPage` has no way to declare default column filters**

### Is your feature request related to a problem? Please describe.

`ListPage` lets a route declare `defaultSort`, `defaultVisibility` and `defaultColumnOrder`:

https://github.com/vendurehq/vendure/blob/9a2ad2bf3ded595bc95ffcbf34a94c8645e662ef/packages/dashboard/src/lib/framework/page/list-page.tsx#L265-L274

but there is nothing for column filters. That leaves no hook for the common "hide these rows unless the user asks for them" case.

Our concrete instance: a store of ~1000 products, some of which are models that are permanently gone and will never restock. A small plugin flags those as archived — deliberately not deletion, and not "disabled" either, because archived products stay live on the storefront as sold-out pages. They are indexed by Google and still bring in organic traffic, so removing them would 404 indexed URLs and burn that SEO value. In the admin product list, though, they are pure noise: day-to-day catalog work never touches them. So the list wants to open with an `isArchived = false` column filter applied, while an admin who does need an archived model — to un-archive one that has restocked, say — just clears the filter. There is no way to express that through the public API, so we patch `@vendure/dashboard` itself in every deploy.

It generalises to any plugin that adds a soft-hide flag: draft articles, out-of-season items, deactivated records. Each of them wants its list page to open in a filtered state that the user can escape.

### Describe the solution you'd like

A `defaultColumnFilters?: ColumnFiltersState` prop on `ListPage`, applied only while the user has no filter state of their own, and behaving like user-set filters once applied — visible as chips, editable, clearable.

The part that needs care is what "no filter state of their own" means. Filters persist to `tableSettings[pageId].columnFilters`, and a user who clears every filter has made a deliberate choice that must not be overwritten by the defaults on the next visit. That requires "never configured" and "cleared" to be distinguishable in the saved settings, which they currently are not — `DataTable` reports its filter state from an effect that also runs on mount, so visiting a page is enough to save an empty one:

https://github.com/vendurehq/vendure/blob/9a2ad2bf3ded595bc95ffcbf34a94c8645e662ef/packages/dashboard/src/lib/components/data-table/data-table.tsx#L436-L451

The sibling view options already read the way the new prop would need to: `useViewOptionDefaults` treats the *absence* of a saved value, not its emptiness, as the signal that the code defaults still apply.

https://github.com/vendurehq/vendure/blob/9a2ad2bf3ded595bc95ffcbf34a94c8645e662ef/packages/dashboard/src/lib/hooks/use-view-option-defaults.ts#L33-L38

Guarding that mount-time write fixes the state saved from then on, but not the state already out there: every existing user has an empty entry saved for every list page they have visited. If `[]` were read as "cleared", the feature would be inert on any installation that is not brand new.

It does not have to be, and nothing already saved needs rewriting. `columnFilters` can carry three states instead of two:

- **absent, or `[]`** — no preference, so the defaults apply. `[]` is what the old mount-time write produced, and a deliberate clear made by that same version produced exactly the same value; the two are indistinguishable, so `[]` cannot honestly be read as either, and reading it as "no preference" is what lets the defaults reach the users who already carry it.
- **`null`** — the user cleared every filter here. What a clear writes from this version on. (`null` rather than deleting the key, so the sentinel survives `JSON.stringify` on its way to `localStorage` and the SettingsStore.)
- **a non-empty array** — the user's own filters.

The honest cost is one visible edge: a user who really had cleared every filter on a page that later declares defaults sees those defaults one more time, and clears them once more — which writes `null` and sticks. On a page with no defaults the distinction is invisible either way. That is the same one-time cost a migration would have, without rewriting anyone's persisted settings, without a version schema, and without a burst of writes on first load after the upgrade.

### Describe alternatives you've considered

Filtering in `transformVariables` instead. That works for the "always hidden" case but not this one, since the user can never turn it off — the filter is invisible to the UI, so there is no chip to clear.

Patching `@vendure/dashboard`, which is what we do now and would rather not.

brmk:
# Description

Closes #5294.

Adds a `defaultColumnFilters` prop to `ListPage`, so a route can declare column filters that apply until the user configures their own:

https://github.com/vendurehq/vendure/blob/18ca3a955955abe99528b737cdec21d034ab9311/packages/dashboard/src/lib/framework/page/list-page.tsx#L278-L288

`ListPage` already accepts `defaultSort`, `defaultVisibility` and `defaultColumnOrder`, but there is no equivalent for filters. Our case: a store of ~1000 products, some of which are models that are permanently gone and will never restock. A small plugin flags those as archived — deliberately not deletion, and not "disabled" either, because archived products stay live on the storefront as sold-out pages. They are indexed by Google and still bring in organic traffic, so removing them would 404 indexed URLs and burn that SEO value. In the admin product list, though, they are pure noise: day-to-day catalog work never touches them. So the list wants to open with an `isArchived = false` column filter applied, while an admin who does need an archived model — to un-archive one that has restocked, say — just clears the filter. Without this prop the only way to get that default is to patch `@vendure/dashboard` itself, which is what we do today.

```tsx
<ListPage
    pageId="product-list"
    title="Products"
    listQuery={productListQuery}
    // hide archived products until the user says otherwise
    defaultColumnFilters={[{ id: 'isArchived', value: { eq: false } }]}
/>
```

The defaults behave exactly as if the user had set them: they render as active filter chips, and can be edited or removed.

## The behaviour contract

1. **Defaults apply on a first visit.** A user who has never configured filters on the page gets the route's defaults.
2. **User edits win and persist.** Once the user changes the filters, their choice is saved and the defaults are no longer consulted.
3. **Clearing all filters persists as cleared.** Removing every filter is a deliberate choice, saved as such — the defaults do not come back on the next visit or remount.
4. **Changing a route's defaults does not clobber existing user state.** A user who already configured filters keeps them; only users with no filter state of their own see the new defaults.

## Making point 3 possible

Point 3 is the interesting one, and it needed a fix below the new prop.

Filters for a page are saved in the user settings under `tableSettings[pageId].columnFilters` (`providers/user-settings.tsx`), the same store that holds `columnVisibility`, `columnOrder` and `pageSize`. There is a precedent for how to express a default there — `useViewOptionDefaults` resolves the saved column order against the code defaults with `??`, so the *absence* of a saved value is what marks the defaults as still applicable:

https://github.com/vendurehq/vendure/blob/9a2ad2bf3ded595bc95ffcbf34a94c8645e662ef/packages/dashboard/src/lib/hooks/use-view-option-defaults.ts#L33-L38

That rule could not hold for filters, because `DataTable` fired `onFilterChange` from an effect that also ran on mount:

https://github.com/vendurehq/vendure/blob/9a2ad2bf3ded595bc95ffcbf34a94c8645e662ef/packages/dashboard/src/lib/components/data-table/data-table.tsx#L436-L451

Merely visiting a list page therefore reported the initial filter state and `ListPage` persisted it, so "has never configured filters here" and "has cleared every filter" both ended up as `[]`. Under those conditions a `persisted?.length ? persisted : defaults` check re-applies the defaults over a state the user deliberately cleared, on every remount — and, since the first visit freezes whatever the defaults were into the user's settings, a route that later changes its defaults can never deliver them.

So `DataTable` now reports filter changes only when the filters actually change (`columnFiltersEqual`, which also replaces the duplicated inline `JSON.stringify` comparison in the same effect):

https://github.com/vendurehq/vendure/blob/18ca3a955955abe99528b737cdec21d034ab9311/packages/dashboard/src/lib/components/data-table/data-table.tsx#L437-L445

The mount run now reports nothing, so visiting a page saves no filter state at all.

Applying a saved view still reports its filters directly (`DataTableProvider.handleApplyView`), including an empty set, which is correct — the user chose that view.

## The `[]` already out there

Fixing the mount-time write fixes the state written from here on, but not the state that is already saved. Every user who has ever opened a list page has an empty `columnFilters` entry for it, written by the old unguarded effect. If `[]` meant "cleared", the new prop would be inert on any existing installation for precisely the users it is meant for; it would only ever work for a brand-new account.

Rather than rewrite those entries, `columnFilters` carries a third state. An emptied filter set is now saved as `null`:

https://github.com/vendurehq/vendure/blob/18ca3a955955abe99528b737cdec21d034ab9311/packages/dashboard/src/lib/framework/page/list-page.tsx#L718-L727

and resolution reads all three:

https://github.com/vendurehq/vendure/blob/18ca3a955955abe99528b737cdec21d034ab9311/packages/dashboard/src/lib/framework/page/list-page.tsx#L477-L488

| Saved `columnFilters` | Meaning | Resolves to |
| --- | --- | --- |
| absent | user has never configured filters here | `defaultColumnFilters` |
| `[]` | no preference — see below | `defaultColumnFilters` |
| `null` | user cleared every filter | `[]` |
| `[...]` | user's own filters | the saved filters |

`[]` is read as *no preference* rather than as a clear. It is what the old mount-time write produced, and a deliberate clear made by that same version produced exactly the same value, so the two are indistinguishable — it cannot honestly be read as either, and reading it as "no preference" is what lets the defaults reach users who already carry it. From this version on nothing writes `[]` for a page: a clear writes `null`, and merely visiting writes nothing.

The `null` is written on every page, not only those declaring defaults. It is what the store means by "cleared" now, it costs nothing on a page with no defaults, and a page can gain defaults later. `null` rather than deleting the key, because the sentinel has to survive `JSON.stringify` on the way to `localStorage` and the SettingsStore, which `undefined` would not.

There is one visible consequence, worth stating plainly: **a user who really had cleared every filter on a page that later declares defaults sees those defaults one more time, and clears them once more.** That second clear writes `null` and sticks. On a page with no `defaultColumnFilters` the distinction is invisible either way.

Nothing already stored is rewritten, there is no version stamp, no migration ladder, and no write burst on first load after the upgrade. The only change to `user-settings.tsx` is the type and its docstring.

## Asynchronous settings and the mount gate

`DataTable` seeds its filter state from props once, on mount. The user settings resolve asynchronously, so a table that mounts before the server-side SettingsStore responds keeps filter state that the arriving settings may contradict. That divergence between the chips and the query is pre-existing, and on pages without defaults it is invisible in practice; with defaults it is not, because the default chips would sit on screen while the list query ran unfiltered.

The settings provider already exposes `settingsReady` for exactly this, with a docstring telling one-shot consumers to wait for it. `ListPage` now does: it defers mounting the data table until `settingsReady`, **only** when both `defaultColumnFilters` and `pageId` are set. Scoping the gate to the new feature means no existing list page changes how it mounts. A narrower prop→state sync effect inside `DataTable` was considered instead and rejected: `columnFilters` arrives as a fresh array on every render, so the effect would also fire on the re-render caused by the user's own edit — at which point the prop still holds the pre-persistence value and the edit would be reverted under them.

While the gate is closed the block renders nothing. A loading skeleton would read better than an empty block, and `DataTable` already has one — but it is driven by `isLoading` on a *mounted* table, and mounting the table is precisely what the gate exists to prevent, so using it would mean duplicating the skeleton markup in `ListPage`. Left minimal on purpose; happy to add one if you would prefer it.

The residual limitation is unchanged and left as it was: on a list page **without** `defaultColumnFilters`, a table that mounts before the server settings arrive still keeps whatever filter state it was seeded with. Fixing that generally means reconciling props into `DataTable`'s state, which is a larger change than this PR should carry.

# Breaking changes

None. The new prop is optional, and pages that do not use it behave as before.

Two behavioural changes beyond the new prop:

- Visiting a list page no longer writes an empty `columnFilters` entry into the user's saved table settings.
- Clearing every filter on a list page now saves `null` where it previously saved `[]`. `TableSettings['columnFilters']` widens to `ColumnFiltersState | null` accordingly. No saved settings are read differently on upgrade except that an existing `[]` stops suppressing a page's defaults, and nothing already stored is rewritten.

`onColumnVisibilityChange` and `onPageChange` emit on mount in the same way `onFilterChange` used to, and are deliberately left alone here. Neither currently has a consumer that needs to distinguish "never set" from "set to the default", so changing them would be scope this PR does not need; the same guard would apply if that changes.

## Tests

Each half of the change is covered by tests that fail without it. Against a naive `persisted ?? defaults` resolution, the `null` and legacy-`[]` unit tests both fail; against a naive `setTableSettings(pageId, 'columnFilters', filters)` write, the two `null`-persistence unit tests fail and the e2e clear-all test fails with `Received: []`. The mount-gate test fails without the gate, and the "visiting saves nothing" e2e fails without the `DataTable` mount fix. The existing filter e2e suites (`catalog/product-list.spec.ts`, `regression/issue-4730-*`, 20 tests) pass unchanged, twice over.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5295"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791237235&installation_model_id=20193&pr_number=5295&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5295&signature=86852f140a88745cce92f874e7191af112db6e16658eea2e7f0869c8777fa7d5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/reference/dashboard/list-views/data-table.mdx
+++ b/docs/docs/reference/dashboard/list-views/data-table.mdx
@@ -4,7 +4,7 @@ generated: true
 ---
 ## DataTable
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table.tsx" sourceLine="258" packageName="@vendure/dashboard" since="3.4.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table.tsx" sourceLine="259" packageName="@vendure/dashboard" since="3.4.0" />
 
 A data table which includes sorting, filtering, pagination, bulk actions, column controls etc.
 
@@ -21,7 +21,7 @@ Parameters
 
 ## DataTableProps
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table.tsx" sourceLine="129" packageName="@vendure/dashboard" since="3.4.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/components/data-table/data-table.tsx" sourceLine="130" packageName="@vendure/dashboard" since="3.4.0" />
 
 Props for configuring the [DataTable](/reference/dashboard/list-views/data-table#datatable).
 
--- a/docs/docs/reference/dashboard/list-views/list-page.mdx
+++ b/docs/docs/reference/dashboard/list-views/list-page.mdx
@@ -4,7 +4,7 @@ generated: true
 ---
 ## ListPage
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/page/list-page.tsx" sourceLine="524" packageName="@vendure/dashboard" since="3.3.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/page/list-page.tsx" sourceLine="581" packageName="@vendure/dashboard" since="3.3.0" />
 
 Auto-generates a list page with columns generated based on the provided query document fields.
 
@@ -103,7 +103,7 @@ Parameters
 
 ## ListPageProps
 
-<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/page/list-page.tsx" sourceLine="37" packageName="@vendure/dashboard" since="3.3.0" />
+<GenerationInfo sourceFile="packages/dashboard/src/lib/framework/page/list-page.tsx" sourceLine="38" packageName="@vendure/dashboard" since="3.3.0" />
 
 Props to configure the [ListPage](/reference/dashboard/list-views/list-page#listpage) component.
 
@@ -121,6 +121,7 @@ interface ListPageProps<T extends TypedDocumentNode<U, V>, U extends ListQuerySh
     additionalColumns?: AC;
     defaultColumnOrder?: (keyof ListQueryFields<T> | keyof AC | CustomFieldKeysOfItem<ListQueryFields<T>>)[];
     defaultSort?: SortingState;
+    defaultColumnFilters?: ColumnFiltersState;
     defaultVisibility?: Partial<
         Record<keyof ListQueryFields<T> | keyof AC | CustomFieldKeysOfItem<ListQueryFields<T>>, boolean>
     >;
@@ -377,6 +378,33 @@ Allows you to specify the default sorting applied to the table.
 
 ```tsx
 defaultSort={[{ id: 'updatedAt', desc: true }]}
+```
+### defaultColumnFilters
+
+<MemberInfo kind="property" type={`ColumnFiltersState`}  since="3.8.0"  />
+
+Allows you to specify column filters which are applied when the current user has not
+yet configured any filters for this page. They behave exactly as if the user had set
+them: they show up as active filter chips, and can be edited or removed.
+
+Once the user edits the filters, their choice is persisted and these defaults are no
+longer applied — including when the user removes every filter. Clearing the filters is
+persisted as a deliberate choice, so the defaults do not reappear on the next visit.
+
+Requires `pageId` to be set, since the persisted table settings are what distinguishes
+"not configured yet" from "cleared by the user". Without a `pageId` the filters live
+only in the URL, which cannot express that difference, and the defaults are ignored.
+
+*Example*
+
+```tsx
+ <ListPage
+   pageId="product-list"
+   listQuery={productListQuery}
+   title="Products"
+   // hide archived products until the user says otherwise
+   defaultColumnFilters={[{ id: 'isArchived', value: { eq: false } }]}
+ />
 ```
 ### defaultVisibility
 
--- a/packages/dashboard/e2e/tests/components/list-page-filter-persistence.spec.ts
+++ b/packages/dashboard/e2e/tests/components/list-page-filter-persistence.spec.ts
@@ -1,0 +1,154 @@
+import { type Page, expect, test } from '@playwright/test';
+
+import { BaseListPage } from '../../page-objects/list-page.base.js';
+
+// Must match `LS_KEY_USER_SETTINGS` in src/lib/constants.ts. Inlined rather than imported:
+// that module pulls in the generated GraphQL schema enums, which do not belong in the
+// Playwright process.
+const USER_SETTINGS_KEY = 'vendure-user-settings';
+const PAGE_ID = 'product-list';
+
+/**
+ * Reads the saved column filters for a page. `saved` is what distinguishes "no value has
+ * been saved" from a saved `null` — the "user cleared every filter" sentinel — since
+ * `undefined` cannot survive the trip out of `page.evaluate` and both arrive as `null`.
+ */
+async function readSavedColumnFilters(page: Page, pageId: string) {
+    return page.evaluate(
+        ([key, id]) => {
+            const settings = JSON.parse(localStorage.getItem(key) || '{}');
+            const tableSettings = settings.tableSettings?.[id] ?? {};
+            return {
+                saved: Object.prototype.hasOwnProperty.call(tableSettings, 'columnFilters'),
+                value: tableSettings.columnFilters ?? null,
+            };
+        },
+        [USER_SETTINGS_KEY, pageId] as const,
+    );
+}
+
+/**
+ * Puts the page back into the "user has never configured filters here" state that a
+ * first-time visitor sees. The stored auth state is shared between tests, so the saved
+ * settings cannot be assumed to be empty at the start of a test.
+ */
+async function resetSavedTableSettings(page: Page, pageId: string) {
+    await page.evaluate(
+        ([key, id]) => {
+            const settings = JSON.parse(localStorage.getItem(key) || '{}');
+            if (settings.tableSettings) {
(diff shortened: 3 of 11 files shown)


brmk:
Note on the dashboard e2e failures: both failing specs (`Facets › facets CRUD › should create a new facet` and `translation-placeholders.spec.ts:234`) reproduce locally on a clean `minor` checkout with this PR's changes fully reverted — `minor` was actually one failure worse in my run (`Facet values › should navigate to facet value detail` also failed there). So they are pre-existing on the base branch, not introduced here. The PR's own e2e (`list-page-filter-persistence.spec.ts`) and the product-list suite pass twice in a row locally.

PR #5300, opened 2026-09-07
feat(cli): support commands that have both an action and subcommands

michaelbromley:
# Description

The CLI extension contract could express a command with an action, or a group of subcommands, but not both. `assertNode` rejected any node declaring both, and `isCliCommandGroup` treated every node with a `subcommands` array as a pure group.

The hosted Cloud CLI has two commands that need both shapes: `deploy`, which runs on its own and parents `plan` and `teardown`; and `backup db`, which runs on its own inside the `backup` group and parents `list` and `status`. Changing the public hierarchy to avoid this would break compatibility for existing users of those commands.

`CliCommandDefinition` now takes an optional `subcommands` array, so a command may keep its action and parent further commands. Pure group behaviour is unchanged: a node with subcommands and no action still prints its help and exits non-zero when run without a subcommand.

The single `isCliCommandGroup` check was doing three different jobs, which only coincided because the two shapes could never overlap. They are now separate:

- `hasCliSubcommands` — the node parents further commands. This is what tree walking and option scoping key on.
- `isRunnableCliCommand` — the node has an action. This is what action registration and decoration key on.
- `isCliCommandGroup` — the node has subcommands and no action. Used only for error wording.

A command that has subcommands shares its options with everything below it whether or not it also runs an action, so the stricter group rule applies to the new shape: it may not repeat a flag an ancestor already shares. Conflict messages that used to say "the command group" now say "the command" when the node runs an action of its own.

Three behaviours are worth calling out.

A command that has subcommands may not also declare positional `arguments`. The first word after the command name would be ambiguous, since it could equally name a subcommand or fill an argument, and there is no way to tell the two apart. Declaring both is rejected when the plugin is registered, in keeping with every other ambiguity in this contract. Neither hosted consumer needs the combination: `deploy` and `backup db` both take options only.

That rule is what makes the second one sound. Commander resolves the first word against the subcommand names before it considers the action, so a word that named no subcommand is always a mistake. It is reported through Commander's own `unknownCommand`, which produces the same message, the same "did you mean" suggestion, the same output channel and the same exit code that a group produces for the same mistake:

```
$ vendure deploy plann
error: unknown command 'plann'
(Did you mean plan?)
```

Commander also leaves out its implicit `help` subcommand once a command has an action, so a runnable parent is given one explicitly and `vendure deploy help` works the way `vendure config help` does. It is skipped when the plugin declares its own `help`, which Commander would otherwise list twice.

A decorator extending a runnable parent is given a frozen deep copy, so it can see that the command it wraps has subcommands without being handed the nodes the registry goes on to use.

Two documents said a command entry with `subcommands` is a group, which contradicted the new shape: the CLI guide at `docs/docs/guides/developer-guide/cli/index.mdx` and the package architecture note at `packages/cli/src/README.md`. Both are corrected, and the guide has a section for the new shape.

Relates to CLO-710.

# Breaking changes

None. The new `subcommands` property on `CliCommandDefinition` is optional, and every existing command, group and extension behaves as before. The new word-handling and help behaviour applies only to a command that declares `subcommands` alongside an `action`, which was not previously expressible.

One rule does reach an existing shape: a command *group* may no longer declare positional `arguments` either. Nothing in this repository does, and Commander never gave those arguments to anything, so the rule turns a silent no-op into a registration error.

# Screenshots

n/a

# Checklist

📌 Always:

👍 Most of the time:

### DIFF
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -549,10 +549,9 @@ commands without loading the plugin.
 
 ### Nested commands
 
-A command entry that has `subcommands` instead of an `action` is a group. A
-group only exists to hold the commands below it. Running it without a
-subcommand prints its help to stderr and exits 1. Groups can be nested to any
-depth.
+A command entry that has `subcommands` and no `action` is a group. A group only
+exists to hold the commands below it. Running it without a subcommand prints its
+help to stderr and exits 1. Groups can be nested to any depth.
 
 ```ts
 import type { Command } from 'commander';
@@ -615,6 +614,53 @@ by the CLI at every level, so `vendure --help` lists `config`,
 lists the arguments and the options that are valid there. Unknown subcommands,
 unknown options and missing arguments fail through the CLI's normal error path
 with exit code 1.
+
+### Commands that run and also hold subcommands
+
+A command may declare an `action` and `subcommands` together. Running it on its
+own runs the action; naming one of its subcommands runs that instead. Use this
+where a command already exists and needs commands underneath it without changing
+what the bare command does.
+
+```ts
+commands: [
+    {
+        name: 'deploy',
+        description: 'Deploy the application',
+        options: [{ long: '--env <name>', description: 'Target environment', required: true }],
+        action: async (options, command, context) => {
+            // `vendure deploy --env staging`
+            return 0;
+        },
+        subcommands: [
+            {
+                name: 'plan',
+                description: 'Show what a deploy would change',
+                // `vendure deploy plan --env staging`
+                action: async (options, command, context) => {
+                    console.log(context.inheritedOptions.env); // 'staging'
+                    return 0;
+                },
+            },
+        ],
+    },
+],
+```
+
+Its `options` are shared with every command below it, the same as a group's, so
+`--env` above reaches `deploy plan` through `context.inheritedOptions`. The
+command itself reads `--env` from its own options argument.
+
+The CLI resolves the first operand against the subcommand names before it
+considers the action, so a subcommand always wins. An operand that names no
+subcommand is reported as an unknown command rather than reaching the action, so
+`vendure deploy plann` fails instead of deploying.
+
+That protection needs a spare operand to catch. A command that declares
+`arguments` as well as `subcommands` cannot tell a mistyped subcommand from a
+value — with `arguments: [{ name: 'target' }]`, `vendure deploy plann` deploys
+`plann`. Prefer options to positional arguments on a command that has
+subcommands.
 
 ### Extending an existing command
 
@@ -727,7 +773,8 @@ they belong to:
 
 - `rootOptions` on the plugin are added to `vendure` itself, so every command
   can be given them.
-- `options` on a command group apply to every command inside that group.
+- `options` on a command that has subcommands apply to every command below it,
+  whether that command is a group or runs an action of its own.
 
 A shared option can be given anywhere on the command line once the command that
 declares it has been reached. These two are equivalent — same flag, same value,
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -651,16 +651,19 @@ Its `options` are shared with every command below it, the same as a group's, so
 `--env` above reaches `deploy plan` through `context.inheritedOptions`. The
 command itself reads `--env` from its own options argument.
 
-The CLI resolves the first operand against the subcommand names before it
-considers the action, so a subcommand always wins. An operand that names no
-subcommand is reported as an unknown command rather than reaching the action, so
-`vendure deploy plann` fails instead of deploying.
+A command that has subcommands cannot also declare `arguments`. The first word
+after the command name would be ambiguous, since it could equally name a
+subcommand or fill an argument, so declaring both is rejected when the plugin is
+registered. Take an option instead.
 
-That protection needs a spare operand to catch. A command that declares
-`arguments` as well as `subcommands` cannot tell a mistyped subcommand from a
-value — with `arguments: [{ name: 'target' }]`, `vendure deploy plann` deploys
-`plann`. Prefer options to positional arguments on a command that has
-subcommands.
+That means a word which names no subcommand is always a mistake, and it is
+reported as an unknown command rather than reaching the action:
+
+```bash
+$ npx vendure deploy plann
+error: unknown command 'plann'
+(Did you mean plan?)
+```
 
 ### Extending an existing command
 
--- a/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
+++ b/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
@@ -169,18 +169,33 @@ describe('CLI plugin commands that also have subcommands', () => {
     it('keeps the shared root options in scope below a runnable parent', async () => {
         const result = await project.runCliCommand(['deploy', 'plan', '--token', 'tok', '--json']);
 
-        expect(parseCloudResult(result.stdout).inherited).toMatchObject({ token: 'tok', json: true });
+        expect(parseCloudResult(result.stdout).inherited).toEqual({ token: 'tok', json: true });
     });
 
-    it('rejects a mistyped subcommand rather than running the parent', async () => {
+    it('reports a mistyped subcommand rather than running the parent', async () => {
         const result = await project.runCliCommand(['deploy', 'plann'], { expectError: true });
 
         expect(result.exitCode).not.toBe(0);
-        expect(result.stderr).toContain('too many arguments');
+        expect(result.stderr).toContain("unknown command 'plann'");
         expect(result.stdout).not.toContain('CLOUD_RESULT');
     });
 
-    it('lists a runnable parent subcommands and its own options in help', async () => {
+    it('keeps a help subcommand on a command that has an action', async () => {
+        const result = await project.runCliCommand(['deploy', 'help']);
+
+        expect(result.stdout).toMatch(/^\s+plan\s+Show what a deploy would change$/m);
+        expect(result.stdout).not.toContain('CLOUD_RESULT');
+    });
+
+    it('lists the subcommands of a runnable parent nested inside a group', async () => {
+        const result = await project.runCliCommand(['backup', 'db', '--help']);
+
+        expect(result.stdout).toContain('Usage: vendure backup db');
+        expect(result.stdout).toMatch(/^\s+list\s+List database backups$/m);
+        expect(result.stdout).toMatch(/^\s+status\s+Show the status of a backup$/m);
+    });
+
+    it("lists a runnable parent's subcommands and its own options in help", async () => {
         const result = await project.runCliCommand(['deploy', '--help']);
 
         expect(result.stdout).toContain('Usage: vendure deploy [options] [command]');
--- a/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
+++ b/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
@@ -177,6 +177,8 @@ describe('CLI plugin commands that also have subcommands', () => {
 
         expect(result.exitCode).not.toBe(0);
         expect(result.stderr).toContain("unknown command 'plann'");
+        // The same suggestion a command group gives for the same mistake.
+        expect(result.stderr).toContain('(Did you mean plan?)');
         expect(result.stdout).not.toContain('CLOUD_RESULT');
     });
 
@@ -185,6 +187,20 @@ describe('CLI plugin commands that also have subcommands', () => {
 
         expect(result.stdout).toMatch(/^\s+plan\s+Show what a deploy would change$/m);
         expect(result.stdout).not.toContain('CLOUD_RESULT');
+    });
+
+    it('rejects a plugin whose command has both arguments and subcommands', async () => {
+        const broken = createTestProject('cli-plugin-ambiguous-parent');
+        try {
+            installCliPluginFixture(broken, 'ambiguous-parent-cli-plugin');
+            const result = await broken.runCliCommand(['--help']);
+
+            expect(result.stderr).toContain('declares both positional arguments and subcommands');
+            // The plugin is skipped rather than taking the CLI down with it.
+            expect(result.stdout).toContain('plugins');
+        } finally {
+            broken.cleanup();
+        }
     });
 
     it('lists the subcommands of a runnable parent nested inside a group', async () => {
--- a/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
+++ b/packages/cli/e2e/cli-plugin-commands.e2e-spec.ts
@@ -116,6 +116,81 @@ describe('CLI plugin nested commands', () => {
     });
 });
 
+describe('CLI plugin commands that also have subcommands', () => {
+    let project: CliTestProject;
+
+    beforeAll(() => {
+        project = createTestProject('cli-plugin-runnable-parent');
+        installCliPluginFixture(project, 'cloud-cli-plugin');
+    });
+
+    afterAll(() => {
+        project?.cleanup();
+    });
+
+    it('runs the parent action when no subcommand is given', async () => {
+        const result = await project.runCliCommand(['deploy', '--env', 'staging']);
+
+        expect(result.exitCode).toBe(0);
+        const parsed = parseCloudResult(result.stdout);
+        expect(parsed.command).toEqual(['deploy']);
+        expect(parsed.options).toEqual({ env: 'staging' });
+    });
+
+    it('runs a subcommand of a runnable parent', async () => {
+        const plan = await project.runCliCommand(['deploy', 'plan']);
+        const teardown = await project.runCliCommand(['deploy', 'teardown']);
+
+        expect(parseCloudResult(plan.stdout).command).toEqual(['deploy', 'plan']);
+        expect(parseCloudResult(teardown.stdout).command).toEqual(['deploy', 'teardown']);
+    });
+
+    it('shares a parent option with its subcommands', async () => {
+        const result = await project.runCliCommand(['deploy', 'plan', '--env', 'staging']);
+
+        const parsed = parseCloudResult(result.stdout);
+        expect(parsed.inherited.env).toBe('staging');
+        // The parent owns the option, so it is not one of the subcommand's own.
+        expect(parsed.options).toEqual({});
+    });
+
+    it('runs a runnable parent nested inside a group, and its subcommands', async () => {
+        const db = await project.runCliCommand(['backup', 'db', '--label', 'nightly']);
+        const list = await project.runCliCommand(['backup', 'db', 'list']);
+        const status = await project.runCliCommand(['backup', 'db', 'status', '--label', 'nightly']);
+
+        expect(parseCloudResult(db.stdout).command).toEqual(['backup', 'db']);
+        expect(parseCloudResult(db.stdout).options).toEqual({ label: 'nightly' });
+        expect(parseCloudResult(list.stdout).command).toEqual(['backup', 'db', 'list']);
+        expect(parseCloudResult(status.stdout).command).toEqual(['backup', 'db', 'status']);
+        expect(parseCloudResult(status.stdout).inherited.label).toBe('nightly');
+    });
+
+    it('keeps the shared root options in scope below a runnable parent', async () => {
+        const result = await project.runCliCommand(['deploy', 'plan', '--token', 'tok', '--json']);
+
+        expect(parseCloudResult(result.stdout).inherited).toMatchObject({ token: 'tok', json: true });
+    });
+
+    it('rejects a mistyped subcommand rather than running the parent', async () => {
+        const result = await project.runCliCommand(['deploy', 'plann'], { expectError: true });
+
+        expect(result.exitCode).not.toBe(0);
+        expect(result.stderr).toContain('too many arguments');
+        expect(result.stdout).not.toContain('CLOUD_RESULT');
+    });
+
+    it('lists a runnable parent subcommands and its own options in help', async () => {
+        const result = await project.runCliCommand(['deploy', '--help']);
+
+        expect(result.stdout).toContain('Usage: vendure deploy [options] [command]');
+        expect(result.stdout).toMatch(/^\s+plan\s+Show what a deploy would change$/m);
+        expect(result.stdout).toMatch(/^\s+teardown\s+Tear the deployment down$/m);
+        expect(result.stdout).toMatch(/^\s+--env /m);
+        expect(result.stdout).toContain('Global Options:');
+    });
+});
+
 describe('CLI plugin help output', () => {
     let project: CliTestProject;
 
@@ -137,6 +212,7 @@ describe('CLI plugin help output', () => {
         expect(result.stdout).toMatch(/^\s+config \[options\]\s+Manage Cloud configuration$/m);
         expect(result.stdout).toMatch(/^\s+backup\s+Manage backups$/m);
         expect(result.stdout).toMatch(/^\s+restore\s+Restore from a backup$/m);
+        expect(result.stdout).toMatch(/^\s+deploy \[options\]\s+Deploy the application$/m);
         for (const option of ['--token', '--project', '--environment', '--json']) {
             expect(result.stdout).toMatch(new RegExp(`^\\s+${option}`, 'm'));
         }
--- a/packages/cli/e2e/fixtures/cli-plugins/ambiguous-parent-cli-plugin/cli-plugin.js
+++ b/packages/cli/e2e/fixtures/cli-plugins/ambiguous-parent-cli-plugin/cli-plugin.js
@@ -1,0 +1,25 @@
+/**
+ * A CLI plugin whose command declares both positional arguments and
+ * subcommands. The first word after `deploy` would be ambiguous, so the plugin
+ * is rejected when it is loaded.
+ */
+const { defineCliPlugin } = require('@vendure/cli');
+
+module.exports = defineCliPlugin({
+    id: '@vendure-e2e/ambiguous-parent-cli-plugin',
+    commands: [
+        {
+            name: 'deploy',
+            description: 'Deploy the application',
+            arguments: [{ name: 'target', description: 'What to deploy' }],
+            action: async () => 0,
+            subcommands: [
+                {
+                    name: 'plan',
+                    description: 'Show what a deploy would change',
+                    action: async () => 0,
+                },
+            ],
+        },
+    ],
+});
--- a/packages/cli/e2e/fixtures/cli-plugins/ambiguous-parent-cli-plugin/package.json
+++ b/packages/cli/e2e/fixtures/cli-plugins/ambiguous-parent-cli-plugin/package.json
@@ -1,0 +1,12 @@
+{
+    "name": "@vendure-e2e/ambiguous-parent-cli-plugin",
+    "version": "1.0.0",
+    "private": true,
+    "main": "cli-plugin.js",
+    "vendure": {
+        "cliPlugin": "./cli-plugin.js",
+        "cliCommands": [
+            "deploy"
+        ]
+    }
+}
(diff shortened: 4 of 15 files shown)


michaelbromley:
Thanks Nigel — the false guarantee was a real one and I have taken the whole list. Pushed as c0e1dd4c1.

**1. The documented guarantee (must).** Confirmed your repro. `allowExcessArguments(false)` only rejects operands *beyond* the declared count, so a declared positional swallows the typo and the deploy runs. The consequence clause was false and it was in the contract docs and the PR description.

I could not enforce it as written: a mistyped subcommand and a positional value are the same token, so nothing can tell them apart. Rather than forbid `arguments` on a runnable parent — which would be a bigger contract restriction than this issue calls for, and I cannot see the Cloud commands to know the cost — I made the docs true and added the tests that pin the real behaviour:

- `cannot tell a mistyped subcommand from a value when a positional is declared` — asserts exit 0 and `positionals: ['plann']`, the case the doc used to deny.
- `reports a second argument that names no subcommand` — the guard still fires past the declared positionals.

The contract now states the limit plainly and tells an author to prefer options to positional arguments on a command with subcommands. The PR description says the same.

**2. Two policies for excess arguments (should).** You were right that this was an ad-hoc local fix, and the fix was better than either option you offered. `allowExcessArguments(false)` is gone. A runnable parent now checks for itself whether an operand is still spare past its declared positionals, and reports `error: unknown command 'plann'` — the same wording a pure group already produces.

That answers the arity-message point at the same time: the user is told what actually went wrong rather than about argument counts. And it gives the guard a reason specific to the shape rather than a general policy, so leaves needing no exemption. Making every leaf strict would change behaviour for every existing built-in and plugin command, which is outside this issue.

**3. The `help` subcommand.** Adding it, not documenting it. `command.addHelpCommand()` on a runnable parent restores `vendure deploy help`. I checked the collision you would have asked about next: a plugin that declares its own `help` subcommand still wins and no duplicate is added. Covered by a unit test and an e2e test.

**4. Formatted error text as data (should).** Fixed, and further than you suggested. `DeclaredOption.sharedBy?: string` is now `declaredBy: CliCommandNode`, and `ExtensionOptionScope.subtree.describedAs` is now `subtree.target`. Carrying the node rather than a second boolean means there is nothing derived to keep in step: the call sites ask `hasCliSubcommands(declared.declaredBy)` for the rule and call `describeCommand` only when a conflict is actually being reported. No string is formatted on the happy path.

**5. `freezeCommand` (should).** Dropped, as you preferred. Note that just removing the explicit `subcommands` line would have made it worse — the spread would have passed the live array through unfrozen — so it is destructured out: `const { subcommands, ...rest } = command`. The freeze comment's promise is true again and `CliCommandDecoratorInput.command` needed no change, since it already documents options, arguments and description.

**6. Stale docs (must).** All six fixed, including the guide. `docs/.../cli/index.mdx` now says a group is a command with `subcommands` and no `action`, has a new section for the runnable shape with a worked example, and the shared-options bullet covers both. The guide documents the positional ambiguity too, so a user meets it before they hit it.

`@since 3.8.0` dropped from `CliCommandParent` — it stays unexported.

**7. Tests.** `toMatchObject` is `toEqual`. The extension-sharing test now decorates `deploy plan` and asserts `inheritedOptions` equals `{ dryRun: true }` from a real run, rather than reading help text. `backup db --help` is covered at both unit and e2e level. I also added the mirror I had spotted myself: a runnable parent declaring a flag the root already shares, which was only tested in the other direction.

On `toHaveLength(1)` — you were right that it asserted nothing. Rather than change idiom in two of five sibling tests, both accept-tests now assert the shape they accepted: that the action is a function and the subcommand names are what was declared.

**8. Prose.** The sharing rule lives in the `hasCliSubcommands` docblock once and names the three callers; the other four sites point at it or say it in passing. The three restating comments are gone. Commander is named as the actor. The non-function-action error now suggests fixing the action first. Both test names have their apostrophe.

Verification after the changes: 489 CLI unit tests and 82 e2e tests pass, `tsc` clean, eslint and prettier clean on everything touched.

michaelbromley (commented):
> **Nigel code review — round one (published after the fact).**
>
> Everything below the line is the unedited output of a local, read-only `nigel:nigel` agent
> invocation. It was run against `8308d6617`, the head at the time of review; the PR head is now
> `c0e1dd4c1`. Nothing has been rewritten, summarised, omitted or added.
>
> This review was produced as private agent output and was not posted at the time, so my
> point-by-point reply (the earlier comment on this PR) was published before the findings it
> answers. Publishing it now repairs that: the review and the reply can be read together.

---

Ey up. I pulled PR 5300 with `gh pr diff`, confirmed the local worktree is on the same head (`8308d6617`), built the package, ran the CLI unit suite (484 passed), both plugin e2e specs (30 passed), and then poked the new shape myself with a few throwaway specs. Here's what I found.

## Verdict: needs work, but the bones are good

The three-way split of `isCliCommandGroup` into `hasCliSubcommands` / `isRunnableCliCommand` / `isCliCommandGroup` is the right call. One predicate doing three jobs because two shapes could never overlap is exactly the sort of accidental coincidence that rots, and pulling it apart before adding the third shape is proper craftsmanship. The registration rewrite in `registerNode` is smaller and clearer than what it replaces. Where it falls down is a documented guarantee that the code does not deliver, and a set of stale comments that contradict the new model.

## The documented guarantee is false

`/Users/micha/.paseo/worktrees/1vmlbsy0/clo-710-runnable-groups/packages/cli/src/shared/cli-command-definition.ts:104-107` promises:

```
 * An argument beyond the declared positionals is rejected rather than passed to
 * the action, so a mistyped subcommand cannot silently run the parent.
```

The second clause is wrong the moment the parent declares a positional. I ran it:

```
deploy declares arguments: [{ name: 'target' }] and subcommands: [plan]
$ vendure deploy plann
exit 0, action ran with commandPath ['deploy'], positionals ['plann']
```

`allowExcessArguments(false)` only rejects arguments *beyond* the declared count. With one optional positional, a typo'd subcommand fills the slot and the deploy runs. That is precisely the failure mode the PR description calls out as the reason the guard exists, and the guard does not cover it. The PR body repeats the same claim, so the wrong statement is in two places.

The tests dance right around this hole. `command-registry.spec.ts` has "gives a subcommand name precedence over a positional argument", which uses `deploy api` — a *valid* value. There is no test combining a declared positional with a mistyped subcommand. If there were, it would fail the doc.

Pick one and commit to it. Either narrow the sentence to "a parent that declares no positionals rejects a stray argument", or actually enforce it — reject the first operand when it is not a declared subcommand name and the parent has subcommands. If you keep the current behaviour, say plainly that a parent with positionals cannot tell a typo from a value.

## Two policies for excess arguments in one CLI

A plain leaf swallows rubbish silently:

```
restore has no arguments, no subcommands
$ vendure restore oops
exit 0, action ran, positionals []
```

A runnable parent rejects it. A pure group says `unknown command 'lst'`. So the same user mistake now produces three different outcomes depending on a property of the command the user cannot see. That is an ad-hoc local fix, not a policy. `allowExcessArguments(false)` is the right default for every command the host registers — put it in `registerNode` unconditionally and let the one behaviour change stand on its own, or explain in the comment at `command-registry.ts:59-64` why leaves are exempt.

While you are there: the message a user gets for a typo is `error: too many arguments for 'deploy'. Expected 0 arguments but got 1.` A user who typed `deploy plann` is being told about argument arity. The group path manages `unknown command 'lst'`. If you can produce the better message for the shape you are adding, do it — a `command.argument` count is not what went wrong in the user's head.

Also worth knowing, because nobody has written it down: Commander does not add its implicit `help` subcommand to a command that has an action handler. So `vendure config help` prints help but `vendure deploy help` either errors with "too many arguments" or, if `deploy` declares a positional, runs the deploy with `target = 'help'`. Confirmed both. Either document it on the `subcommands` prop or add `help` explicitly.

## Formatted error text stored as data

Two places now keep a rendered message where a flag would do:

- `command-registry-store.ts:585-589` — `DeclaredOption.sharedBy?: string`, checked as `if (declared.sharedBy)` at lines 216 and 302.
- `command-registry-store.ts:393-395` — `ExtensionOptionScope.subtree.describedAs`.

The field is doing double duty as a boolean and as presentation, and `listCommandOptions` now formats a string for every parent node in the tree whether or not anything conflicts. `DeclaredOption` already carries `path`; give it `sharesWithSubtree: boolean` and call `describeCommand` at the two message sites. Same for `describedAs`. Conflict detection should not know how errors are worded.

## `freezeCommand` now hands decorators the live subtree

`command-registry-store.ts:526-545`. The comment above it says the copy "means it could not have reached the registered definition even if it did not". With `subcommands` added, that is no longer true in any useful sense: `Object.freeze([...command.subcommands])` freezes the new array, but the elements are the same node objects that `replaceNodeAtPath` reuses. A decorator can reach `frozen.subcommands[0].description` and mutate the registered tree. The same hole exists for `options`, but exposing an entire subtree makes it worth deciding rather than inheriting.

A decorator wraps an action. It has no business seeing the subtree at all. My preference is to drop `subcommands` from `freezeCommand` entirely. If you keep it, `CliCommandDecoratorInput.command` at `cli-command-definition.ts:198-200` still says "Read its `options`, `arguments` and `description`" and needs updating, and the freeze comment needs its promise walked back.

## Stale docs, including public ones

You were meticulous about renaming "group" to "command with subcommands" in the store, then left six places saying the old thing. Four of them are public API docs that ship to users:

- `cli-command-definition.ts:31-33` — "plus the options of every command group the command is nested in". A runnable parent is not a group.
- `cli-command-definition.ts:46-48` — "a flag declared on a group that an ancestor already shares is rejected at registration". Now also true of runnable parents.
- `cli-plugin.ts:242-246` — "A command may repeat a flag one of its groups shares". `assertMatchesInheritedShape` is only reached by leaves, but the flag can now come from a runnable parent.
- `cli-plugin.ts:275-278` — "A group's options are shared with everything below it". This function is now the one applied to runnable parents, which is the whole point of the change.
- `command-registry.ts:131-134` — "rejects a flag shared by both a group and the root, or by a group and one of its ancestors".
- `docs/docs/guides/developer-guide/cli/index.mdx:552` — "A command entry that has `subcommands` instead of an `action` is a group." The user-facing guide for this exact feature now contradicts the API. Line 730 ("`options` on a command group apply to every command inside that group") needs the same treatment.

The guide is a must-fix. A new public shape with no documentation and a guide that says it is impossible is worse than not shipping it.

Smaller: `CliCommandParent` at `cli-command-definition.ts:155-161` carries `@since 3.8.0`, the tag this repo uses for published API, but it is never exported from `src/index.ts`. Drop the tag or export the type.

## Tests

Genuinely good: the exit-code test (`uses the numeric result of a parent action as the exit code`), the option-not-in-scope test asserting the exact `unknown option '--env'` text, and the pair of "given before / given after the subcommand" tests. The comment `// The decorator wraps the parent action only, so running a subcommand does not go through it.` earns its place.

What is missing or weak:

1. **The gap above.** A test with `arguments` *and* `subcommands` *and* a mistyped subcommand. Currently the untested case is the one where the documented behaviour is wrong.
2. `e2e/cli-plugin-commands.e2e-spec.ts:172` uses `toMatchObject({ token: 'tok', json: true })`. It is the only `toMatchObject` in the whole suite, and the sibling unit test uses `toEqual`. `inherited` is fully determined here, so `toEqual` works and actually catches an option leaking in. Weaken assertions only when you have a reason.
3. `command-extensions.spec.ts` "shares an option added to a runnable parent with its subcommands" only asserts the help text lists `--dry-run` under Global Options. The test name promises the value reaches the subcommand. Assert `inheritedOptions.dryRun` from an actual run, the way `command-registry.spec.ts` does for parent-declared options.
4. `expect(plugin.commands).toHaveLength(1)` in the two new accept-tests asserts an array length the test itself wrote — `defineCliPlugin` returns its input. It tells you only "did not throw". I can see it is the file's existing style, so I will not die on this hill, but `expect(() => defineCliPlugin(...)).not.toThrow()` says what is meant.
5. No test that `backup db --help` lists its subcommands, even though a runnable parent nested inside a group is one of the two shapes the whole PR exists for. The help assertions only cover top-level `deploy`.

## Prose

One sentence is restated five times across the change: "A parent shares its options with everything below it whether or not it runs an action." It is at `command-registry.ts:53-54`, `cli-plugin.ts:184-185`, and paraphrased at `cli-command-definition.ts:164-166`, `command-registry-store.ts:425-426` and `:608-610`. Put it in the `hasCliSubcommands` docblock once and let the rest be `{@link}`s. Five copies means five things to update the next time the rule shifts.

`command-registry.ts:43` — `// Narrowed up front: a node may both run an action and have subcommands.` restates the two lines below it, and "Narrowed up front" has no actor. Cut it.

`command-registry-store.ts:505-506` — `// The spread carries any subcommands through, so extending a runnable parent leaves the tree below it as it was.` explains what `...target` does. That belongs in the commit message, not above the code.

`cli-command-definition.ts:103` — "The first argument is matched against the subcommand names before the action is considered" is passive with no actor. Commander does the matching, and naming it tells the reader where to go when it misbehaves.

`cli-command-definition.ts:165-167` — "this rather than `isCliCommandGroup` is what the option and tree-walking rules key on". "tree-walking rules" is a coined noun and "key on" is a metaphor. Say which functions call it.

`cli-plugin.ts:190-191` — the error tells an author to "Omit it to make X a command group", but the author who wrote a broken action almost certainly wanted a runnable parent. Suggest fixing the action first.

Two test names want an apostrophe: "lists a runnable parent subcommands and its own options in help" (both the e2e and unit versions).

## Priorities

Must change before merge:
1. Fix or correct the "a mistyped subcommand cannot silently run the parent" claim at `cli-command-definition.ts:104-107`, and in the PR description. Add the test that covers positionals plus a typo.
2. Update `docs/docs/guides/developer-guide/cli/index.mdx` — the guide currently says the feature you just built cannot exist.
3. Fix the five stale "group" comments, four of which are published API docs.

Should change:
4. Decide one policy for excess arguments across all runnable commands, or write down why leaves are exempt.
5. Replace `sharedBy` and `describedAs` with booleans, and format at the message site.
6. Drop `subcommands` from `freezeCommand`, or fix the comment that claims decorators cannot reach the registered tree.
7. Strengthen the extension-option-sharing test to assert the value, not the help text; swap the lone `toMatchObject` for `toEqual`.

Nitpicking:
8. `@since 3.8.0` on the unexported `CliCommandParent`.
9. The repeated rule sentence, the three restating comments, the two test names.
10. Consider adding `help` to a runnable parent, or documenting that it has none.

michaelbromley (commented):
> **Nigel code review — round two.**
>
> Everything below the line is the unedited output of a local, read-only `nigel:nigel` agent
> invocation, run against `c0e1dd4c1` — the current head of this PR. Nothing has been rewritten,
> summarised, omitted or added.
>
> Published before any fixes or reply, so the findings are on the record ahead of the response to
> them.

---

Got the diff for PR 5300 at `c0e1dd4c1` via `gh pr diff`, and the local worktree is on that exact commit, so I reviewed the real thing. I also ran the CLI unit suite (178 pass), eslint (clean on the changed files), `tsc`, and a throwaway probe spec against Commander 11.1.0 to check the two behavioural claims the code makes.

# Verdict

The core refactor is genuinely good. Splitting one overloaded predicate into `hasCliSubcommands` / `isRunnableCliCommand` / `isCliCommandGroup` is exactly the right call, and the fact that every call site had to pick one of the three is proof the old check was doing three jobs by accident. The option-scoping rules were extended correctly and the tests around them are real tests.

But the third commit — the "unknown subcommand" guard — is the weak part, and it does not do what the PR description says it does. Two specific claims in the code comments are wrong, and I verified both.

# Design issues

**1. The stray-operand guard bypasses Commander's error path, and the result is worse than a group's.** `/Users/micha/.paseo/worktrees/1vmlbsy0/clo-710-runnable-groups/packages/cli/src/shared/command-registry.ts:74-82`:

```ts
process.stderr.write(`error: unknown command '${unknownSubcommand}'\n`);
process.exit(1);
```

The PR body says this matches "what a pure group already does". It does not. I ran both through the same harness:

- pure group, `backup dbb` → `error: unknown command 'dbb'\n(Did you mean db?)\n`
- runnable parent, `deploy plann` → `error: unknown command 'plann'\n`

Commander's `unknownCommand()` builds a `suggestSimilar` hint from the visible subcommands and routes through `this.error()`, which respects `configureOutput().writeErr` and `exitOverride()`. Writing to `process.stderr` and calling `process.exit` directly does neither. So the shape whose entire selling point is "behaves like a group" gets a strictly worse error than a group, and any host that overrides Commander's output or exit gets bypassed.

Use `command.error(\`error: unknown command '${name}'\`, { code: 'commander.unknownCommand' })` at minimum, and build the "Did you mean" hint from `command.commands.map(c => c.name())` so the two shapes actually match.

The reason the tests didn't catch this: `run-cli.ts:51-57` spies on both `process.exit` and `process.stderr.write` as well as configuring Commander's output, so the harness can't tell the two channels apart. That's a test that mirrors the implementation closely enough to be blind to it.

**2. The `addHelpCommand()` comment is wrong, and it produces a duplicate.** `command-registry.ts:59-65` says "A `help` the plugin declared itself wins, and no second one is added." Half right. Commander's default is `this.commands.length && !this._actionHandler && !this._findCommand('help')` — note the third clause. Calling `addHelpCommand()` sets `_addImplicitHelpCommand = true` unconditionally, which discards that guard, and `Help.visibleCommands()` then pushes an implicit help command on top of the declared one. Probe output for a runnable parent with its own `help` subcommand:

```
Commands:
  plan            plan
  help            my own help
  help [command]  display help for command
```

`RESERVED_COMMANDS` only guards the top-level name, so a plugin declaring `deploy help` is reachable. One-line fix: only call `addHelpCommand()` when `!subcommands.some(s => s.name === 'help')`. Then fix the comment to say what the code does.

**3. The whole guard rests on a Commander default nobody has written down.** `strayOperand` (`command-registry.ts:103-105`) only ever gets to run because `commander@11` sets `_allowExcessArguments = true`. Commander flipped that default to "error" in a later major. The moment someone bumps `"commander": "^11.0.0"`, `vendure deploy plann` starts producing `error: too many arguments` from Commander before the action runs, and the guard becomes dead code with tests that still pass for the wrong reason. Either call `.allowExcessArguments(true)` explicitly on a runnable parent so the dependency is stated in code, or name the version in the docblock.

**4. Documenting a footgun where this codebase otherwise rejects one.** `assertNode` rejects a flag shadowed at two levels, a sub-option nested too deep, an option whose value shape disagrees with an inherited one, a group with no subcommands. Every ambiguity in this system is a registration error. Then `arguments` + `subcommands` — whose failure mode is `vendure deploy plann` silently deploying `plann` — gets a paragraph in the guide and a test pinning the bad behaviour. That is inconsistent with everything around it. `assertNode` should reject the combination in `/Users/micha/.paseo/worktrees/1vmlbsy0/clo-710-runnable-groups/packages/cli/src/shared/cli-plugin.ts:174-199`. If the hosted Cloud CLI genuinely needs both, say so in the PR and I'll shut up, but "prefer options to positional arguments" is advice, not a contract.

**5. `freezeCommand` deletes `subcommands` from a public-API object.** `command-registry-store.ts:529-540`:

```ts
const { subcommands, ...rest } = command;
```

`options` and `arguments` get a frozen shallow copy; `subcommands` gets removed entirely. A decorator receives a `CliCommandDefinition` whose `subcommands` is `undefined` and has no way to know that's a lie rather than a leaf. The justification — "no amount of freezing at this level would protect" — is equally true of `options`, which you froze anyway. Either freeze a shallow copy like the neighbours, or document the omission on `CliCommandDecoratorInput.command`. Right now it is a silent behaviour change on a public surface with no test.

# Abstraction problems

**6. Two places now carry a whole node to answer a boolean.** `DeclaredOption.declaredBy` (`command-registry-store.ts:581-586`) replaced `isGroupOption: boolean`, and `ExtensionOptionScope.subtree.target` (`:397-404`) holds the target purely so `describeCommand(subtree.target, scope.path)` can be recomputed three times. Both exist only to answer "group or command?" when rendering a message. Store the rendered name once — `subtree: { name: describeCommand(target, path), parents, leaves }` — and the message formatter stops re-interrogating the tree.

**7. `ExtensionOptionScope` carries both `path: string[]` and `label: string`, where `label === path.join(' ')`.** Two representations of one fact in a five-field interface. Drop `label`.

# Comment and doc quality

**8. `hasCliSubcommands`' docblock enumerates its own call sites** (`cli-command-definition.ts`, the `hasCliSubcommands` block): "That is why `registerCommands` passes them down..., why `assertCliPlugin` applies..., and why the collision checks in `CommandRegistry`...". This rots the first time a call site moves or a fourth one appears, and nothing will fail when it does. A predicate's docblock should say what it returns and why the distinction exists. The "why" paragraph is good; the call-site inventory is a liability.

**9. The `arguments` + `subcommands` caveat appears four times.** In `CliCommandDefinition.subcommands`, in `strayOperand`'s docblock, in `docs/docs/guides/developer-guide/cli/index.mdx`, and in a test comment — near-verbatim each time. Pick the guide as the home, leave one sentence and a pointer at the other three.

**10. `/Users/micha/.paseo/worktrees/1vmlbsy0/clo-710-runnable-groups/packages/cli/src/README.md:29-42` was not updated** and now contradicts the change: "A command that groups further commands declares `subcommands` **instead of** an action", followed by a `CliCommandDefinition` block with no `subcommands` field. That file is the architecture doc a maintainer reads first, and lines 453-460 of it were quoted in the diff context, so it was open. The checklist says the README was updated; the docs-site guide was, this one wasn't.

**11. "Operand" leaks into the user guide.** `index.mdx`: "The CLI resolves the first operand against the subcommand names", "That protection needs a spare operand to catch." A plugin author does not know what an operand is — that's Commander's internal vocabulary. Say "the first word after the command name". The second sentence is also a metaphor doing no work.

# Tests

**12. The change introduces a TypeScript error.** `packages/cli/src/shared/command-registry-store.spec.ts:469`:

```
error TS2339: Property 'action' does not exist on type 'CliCommandGroupDefinition'.
```

`backup` is cast to `CliCommandGroupDefinition`, which has no `action`, then the test asserts `typeof backup.action`. The package already has pre-existing `tsc` errors elsewhere so CI evidently doesn't gate on this, but don't add to the pile.

**13. Two tautological tests.** `command-registry-store.spec.ts` `'accepts a command that has both an action and subcommands'` and `'accepts a runnable command with subcommands nested inside a group'` both construct an object literal, pass it to `defineCliPlugin` (which returns it), and then assert that the returned object still has the properties written three lines above. `expect(deploy.subcommands?.map(n => n.name)).toEqual(['plan', 'teardown'])` tests `Array.prototype.map`. The only real assertion is "it did not throw" — write that, or delete them and let the `registerCommands` tests carry the weight.

**14. `expect(() => registry.applyPlugin(plugin)).toThrow(...)` twice in a row** on the same registry, lines 681-684 of `command-extensions.spec.ts`, once per regex. It happens to be safe because `applyPlugin` throws before committing the draft, but it silently depends on that. Assert once and match both fragments.

**15. `recordingLeaf` no longer builds leaves.** `command-registry.spec.ts:26-50` — the docblock says "A leaf that records what the host handed to it", and the new fixtures call it as `recordingLeaf('deploy', ..., { subcommands: [...] })`. Rename it `recordingCommand`.

**16. Gaps.** Nothing covers: the decorator's `command` argument having no `subcommands` (item 5, a real behaviour change); the duplicate `help` listing (item 2); a runnable parent replaced with `replaces: true`; a pure group nested under a runnable parent. The last one matters because it's the only untested combination of the three node shapes.

# Priorities

Must change before merge:
- Item 1 — route the unknown-subcommand error through `command.error()` and include the suggestion, so the claim of parity with a group is true.
- Item 2 — guard `addHelpCommand()` against a declared `help`, and correct the comment.
- Item 10 — update `packages/cli/src/README.md`.
- Item 12 — the new `tsc` error.

Should change:
- Item 4 — reject `arguments` + `subcommands` at registration, or justify keeping it in the PR body.
- Item 3 — pin the `allowExcessArguments` assumption in code.
- Item 5 — freeze `subcommands` like the other arrays, or document the omission.
- Items 13 and 15 — the tautological tests and the misnamed helper.

Nitpicking:
- Items 6, 7, 8, 9, 11, 14, 16.

Cleared up after myself: the probe spec I wrote to `packages/cli/src/shared/__tests__/probe.spec.ts` has been deleted, the tree is as I found it.

michaelbromley:
Round two taken in full, pushed as `9d07eb632`. You were right on all four must-fixes, and I verified each one before changing anything rather than taking it on trust. Three of them contradicted claims I had made in this PR, so those are corrected in the description too.

**1. The guard bypassed Commander, and the parity claim was false.** Reproduced exactly: a group gives `unknown command 'dbb'\n(Did you mean db?)\n` through the configured `writeErr`, mine gave a bare message straight to `process.stderr` with no suggestion and no respect for `exitOverride`.

Fixed by calling Commander's own `unknownCommand()`, so the message, the suggestion, the channel and the exit code are not merely similar to a group's but produced by the same code. It is a public method Commander calls internally and has never put in its typings, so it goes through a small helper that casts, and falls back to the typed `command.error(...)` if a future version drops it — losing the hint but keeping the channel and exit code right.

Your diagnosis of why the tests missed it was the more useful half. `run-cli.ts` merged both channels into one string, so it was structurally incapable of catching this. It now records `commanderStderr` and `processStderr` separately, with `stderr` kept as the combined view. The new test asserts the runnable parent and the group produce matching suggestions and exit codes, and that `processStderr` is empty. It fails against the old implementation.

**2. `addHelpCommand()` duplicated a declared `help`.** Confirmed. My earlier claim that "no duplicate is added" was wrong because I had checked the registered command list rather than the rendered output — registration has one entry, the help listing printed two rows. Now guarded by `!subcommands.some(s => s.name === 'help')`, with the comment corrected to say what the code does, and a test asserting the declared description appears and `help [command]` does not.

**3. The undeclared `allowExcessArguments` dependency.** Now stated in code as `command.allowExcessArguments(true)` on a runnable parent, so a future Commander default flip cannot silently turn the guard into dead code.

**4. Rejecting `arguments` + `subcommands`.** Taken, and you were right that documenting it was inconsistent with everything around it. I checked with the hosted consumers rather than guessing: `deploy` and `backup db` both take options only, so nothing needs the ambiguous shape. `assertNode` now rejects the combination for groups as well as runnable parents, with contract, registration and e2e coverage including a fixture plugin that gets skipped rather than taking the CLI down.

This removed rather than moved the problem — the caveat you flagged as repeated four times (item 9) is simply gone, along with the test that pinned the bad behaviour.

**5. `freezeCommand` deleting `subcommands`.** Neither of your two options, because both lose something: freezing a shallow copy still hands over the live child nodes, and documenting the omission leaves a decorator unable to tell a runnable parent from a leaf. It is now a recursive frozen deep copy, so `subcommands` is visible and nothing reachable through it is writable. Two tests: one proving a decorator can distinguish the two shapes, one proving that writing to a child throws and the registered tree is untouched.

**6 and 7.** Item 6 is the one place I have not done what you asked, because rounds one and two pulled in opposite directions. Round one said "conflict detection should not know how errors are worded" and I replaced a rendered string with the node. Round two says the node is too much to answer a boolean and asks for the rendered name back. Both critiques are fair and the fix that satisfies both is neither: `DeclaredOption.sharedBy` is now `'group' | 'command' | undefined`. That is a classification rather than prose, it is the only bit the wording needs, and nothing is formatted unless a conflict is actually reported. Say the word if you still want the string. Item 7 done — `label` dropped, derived from `path` where needed.

**8, 11.** The call-site inventory is out of the `hasCliSubcommands` docblock, which now states the distinction and why it is wider than `isCliCommandGroup`. "Operand" is out of the user guide in favour of "the first word after the command name".

**10.** `packages/cli/src/README.md` updated: the `CliCommandDefinition` block has `subcommands`, and the prose covers the runnable shape and the arguments rule. The checklist item was ticked for the docs-site guide and I had missed this file; the description now names both.

**12.** Fixed, and the reason it escaped is worth recording. `tsconfig.cli.json` uses `"files": ["./src/cli.ts", "./src/index.ts"]`, so it compiles only the graph reachable from the two build entry points and contains no spec files at all. My "tsc clean" was true of what it ran and I had over-read it. I now typecheck the specs separately as well, which immediately caught a second error I would otherwise have shipped: a `/s` regex flag above the package's ES target.

**13, 14, 15.** The two accept-tests now assert the predicates rather than re-reading properties written three lines above. The doubled `toThrow` is one assertion matching both fragments. `recordingLeaf` is `recordingCommand`.

**16.** All four gaps covered: the decorator's view of `subcommands`, the duplicate `help` listing, `replaces: true` on a runnable parent (asserting the old subtree goes with it), and a pure group nested under a runnable parent.

Verification: 495 unit tests, 83 e2e tests, build-graph typecheck clean, spec typecheck clean, eslint clean on every file touched, prettier clean, package builds.

PR #5301, opened 2026-09-07
docs(core): document that /health no longer runs health check strategies

### LINKED ISSUE #5270: Health check docs still describe the pre-3.6.0 behaviour
## Is your feature request related to a problem? Please describe.

The health check documentation describes behaviour that was removed in v3.6.0.

`/health` today returns an unconditional `{ "status": "ok" }`. The controller has no dependencies and
runs no indicators — #4442 reduced it to a single `return { status: 'ok' }`. The strategies configured
in `systemOptions.healthChecks` are still initialised, but nothing ever calls their indicator
functions. `HealthCheckRegistryService.healthIndicatorFunctions` has no readers left in the codebase.

The docs still say the opposite:

- [Health Checks guide](https://docs.vendure.io/guides/core-concepts/healthchecks/): "The `/health`
  endpoint aggregates all strategy results into a single response ... If any strategy reports a
  failure, the endpoint returns an unhealthy status". It also recommends using it as a readiness
  probe to "delay traffic until the application is ready", and tells you to write custom strategies
  whose "results are included in the aggregated `/health` response".
- [Using Docker](https://docs.vendure.io/guides/deployment/using-docker/) shows a sample response with
  the old Terminus `info` / `error` / `details` envelope and a `database: { status: "up" }` indicator.
  That response shape no longer exists.
- The API reference, generated from JSDoc, repeats it. `SystemOptions.healthChecks` documents
  `@default [TypeORMHealthCheckStrategy]`; the actual default in `default-config.ts` is `[]`.
  `HealthCheckRegistryService` still says registered indicators "will also appear in the Admin UI's
  system status view" — that view was removed in the same PR.

Nothing in the docs indicates any of this changed. The `@deprecated` tags say to prefer
infrastructure-level checks, which reads as a recommendation rather than a statement that the feature
has already stopped working.

This is worth fixing beyond tidiness. The failure it invites is quiet and it is expensive. You read
the guide, conclude `/health` verifies the database, and point your load balancer's target-group check
at it. You now believe you have a readiness check that pulls an instance out of rotation when its
database connection goes bad. You do not — every instance will report healthy through a total database
outage. Someone who has been running Vendure since before 3.6.0 has the inverse problem: their custom
`HealthCheckStrategy` silently stopped running on upgrade, and the docs still tell them it works.

Worth noting that the outcome here is the right one. Gating `/health` on the database and wiring it to
a load balancer is how a one-minute database slowdown becomes a full outage: every instance fails the
check simultaneously, the balancer kills them all, and the replacements cold-boot for a minute or more
into the same slow database and fail the same check. Restarting application processes was never going
to fix a database. #4441 made the right call. The docs just never caught up, so the reasoning is not
available to the people who need it. In fact the v3.6.0 scope in #4441 itself lists "Update documentation to recommend
infrastructure-level health checks instead" — that is the one item that never landed.

## Describe the solution you'd like

Update the documentation to match the code:

- Rewrite the Health Checks guide around what `/health` actually is — a liveness signal that the
  process is up and serving HTTP — with a note that the aggregation behaviour was removed in v3.6.0.
- Explain the liveness/readiness distinction and why the built-in checks were removed, so the guidance
  carries its reasoning. Include a probe example, and a `startupProbe` suggestion since Vendure's boot
  time makes a naive `livenessProbe` restart pods mid-startup.
- Give the people who had custom strategies somewhere to land: those checks are inert, remove them,
  set `healthChecks: []` to silence the startup warning, and move the monitoring to the layer that can
  act on it.
- Fix the sample response in the Docker guide.
- Fix the JSDoc that feeds the API reference, including the incorrect `@default`.

An e2e test asserting that `/health` returns `{ "status": "ok" }` while a throwing strategy is
configured would pin the contract. There is no test covering the route at present, which is part of
why the divergence went unnoticed.

## Describe alternatives you've considered

Adding a separate `GET /health/live` alongside `/health`, so the liveness question has a dedicated
endpoint. This is unnecessary — `/health` already has exactly those semantics, so the new route would
be byte-for-byte identical to the existing one, and it would add public API surface to a module
scheduled for removal in v4.0.0. The gap is that `/health` is not documented as a liveness endpoint,
not that one is missing.

Leaving the docs until the v4.0.0 removal. That leaves the misleading text in place across the whole
v3 line, and the people most likely to be misled are the ones deploying now.

## Additional context

Happy to open the PR — I have the change ready.

gabriellbui:
# Description

Fixes #5270. Reported by @brmk, supersedes #5271.

Since v3.6.0 (#4442) `/health` returns an unconditional `{ "status": "ok" }` and the strategies in `systemOptions.healthChecks` are not executed. The Health Checks guide, the Docker guide and the JSDoc still described the old aggregated response and the old `[TypeORMHealthCheckStrategy]` default.

Changes:

- Health Checks guide rewritten around the real response, with a "Changed in v3.6.0" note, the reason the dependency checks were removed, and the `healthChecks: []` migration step.
- Docker guide: old Terminus sample response replaced with the real one.
- `@default` on `systemOptions.healthChecks` corrected to `[]`.
- Each deprecated health check class gets one sentence saying it is not executed since v3.6.0. Sentences describing removed behaviour are deleted. The `@deprecated` tags on `SystemOptions.healthChecks` and `HealthCheckStrategy` now state the fact instead of giving advice.
- Reference pages regenerated with `bun run docs:generate-typescript-docs`.

Compared with #5271 this leaves out the probe tutorial, the per-page warning blocks and the e2e test. The reference pages are removed in v4.0.0, so one sentence each is enough.

# Breaking changes

None. Docs and JSDoc only.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5301"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791375515&installation_model_id=20193&pr_number=5301&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5301&signature=a43de143070b527e6ac3b964b947e736d6a5c10e888ec870d2870dca10e027b0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/docs/guides/core-concepts/healthchecks/index.mdx
+++ b/docs/docs/guides/core-concepts/healthchecks/index.mdx
@@ -1,60 +1,32 @@
 ---
 title: "Health Checks"
 metaTitle: "Health Check Endpoint for Monitoring Vendure"
-metaDescription: "How Vendure exposes a /health endpoint for monitoring system dependencies like the database and external services."
+metaDescription: "What the Vendure /health endpoint reports and how to use it as a liveness probe."
 ---
 
-Vendure exposes a **`/health`** endpoint that reports the status of critical system dependencies.
-This endpoint is essential for production deployments where load balancers, container orchestrators,
-and monitoring tools need to know whether the application is healthy and ready to serve traffic.
+Vendure exposes a **`/health`** endpoint on the server. It is a plain REST route which returns
+`{ "status": "ok" }` as soon as the application is accepting HTTP requests.
 
-## How it works
+The response is unconditional. `/health` does not check the database or any other dependency. It only
+tells you that this process is running and serving HTTP, which is what a liveness probe needs.
 
-Each configured health check strategy probes a specific dependency and reports back with a status.
-The `/health` endpoint aggregates all strategy results into a single response.
+:::warning[Changed in v3.6.0]
 
-A healthy response indicates that all dependencies are reachable and functioning correctly. If any
-strategy reports a failure, the endpoint returns an unhealthy status, signaling that the system
-may not be able to serve requests reliably.
+Before v3.6.0, `/health` ran the strategies configured in `systemOptions.healthChecks` and returned a
+503 if any of them failed. Since v3.6.0 those strategies are not executed. The strategy classes and
+`HealthCheckRegistryService` will be removed in v4.0.0. Remove any custom strategies and set
+`healthChecks: []` to silence the startup warning. See [#4441](https://github.com/vendurehq/vendure/issues/4441) for the reasoning.
 
-## Built-in strategies
+:::
 
-Vendure ships with two health check strategies out of the box:
+Checking dependencies from the application health endpoint was removed on purpose. If `/health` failed
+whenever the database was slow, an orchestrator would restart every instance at once, and restarting the
+application cannot fix the database. Monitor dependencies with their own alerting and keep the
+application probe shallow.
 
-- **`TypeORMHealthCheckStrategy`** — verifies that the database connection is alive by executing
-  a simple query. Since the database is the most critical dependency for any Vendure instance,
-  this strategy is enabled by default.
-
-- **`HttpHealthCheckStrategy`** — checks the availability of an external HTTP service. This is
-  useful when your Vendure instance depends on external APIs, such as a payment gateway or a
-  third-party inventory system.
-
-## Configuration
-
-Health checks are configured via the `systemOptions.healthChecks` property in the
-[`VendureConfig`](/reference/typescript-api/configuration/vendure-config/). You can enable or disable
-individual strategies and configure their parameters to match your infrastructure.
-
-## Use cases
-
-The `/health` endpoint serves several important operational purposes:
-
-- **Load balancer probes** — configure your load balancer to poll `/health` and route traffic
-  only to healthy instances.
-- **Container orchestration** — use the endpoint as a Kubernetes liveness or readiness probe
-  to automatically restart unhealthy pods or delay traffic until the application is ready.
-- **Monitoring dashboards** — integrate with monitoring tools to track uptime and get alerted
-  when dependencies become unreachable.
-
-## Custom strategies
-
-When your application has dependencies beyond the database and HTTP services, you can implement
-the `HealthCheckStrategy` interface to create custom checks. For example, a custom strategy might
-verify connectivity to a Redis cache, an Elasticsearch cluster, or a message broker.
-
-Custom strategies are added to the same `systemOptions.healthChecks` configuration array alongside
-the built-in strategies, and their results are included in the aggregated `/health` response.
+The worker can serve the same unconditional `/health` response. See the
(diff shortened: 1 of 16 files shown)


biggamesmallworld (changes requested):
Facts all check out: `/health` returns an unconditional `{ "status": "ok" }`, the default is `[]`, and anchoring on v3.6.0 instead of "now" is the right call. Three leftovers before this can go in, all of them the thing this PR set out to remove.

**1. Generated file edited by hand.**

`docs/docs/reference/typescript-api/health-check/health-check-strategy.mdx:45` gained "Since v3.6.0 this method is never called." That sentence is not in the source JSDoc for `getHealthIndicator` (`packages/core/src/config/system/health-check-strategy.ts:32-36`), so the next `docs:generate-typescript-docs` run deletes it. Put it on the JSDoc and regenerate.

**2. `health-check-registry.service.ts:9-14` contradicts itself.**

Line 6 says registered indicators are not executed since v3.6.0. Lines 9-11 then say "the preferred way to implement a custom health check is by creating a new `HealthCheckStrategy` and then passing it to the `systemOptions.healthChecks` array", followed by a worked example of the registry path. Both read as live instructions for a dead code path. Delete them or mark them as pre-v3.6.0.

**3. `typeorm-health-check-strategy.ts:26-29` documents an impossible failure.**

```
// The default key is "database" and the default timeout is 1000ms
// Sometimes this is too short and leads to false negatives in the
// /health endpoint.
```

A strategy that is never executed cannot cause a false negative. The lead-in "It is added via the `systemOptions.healthChecks` array" has the same problem. Say the example shows the pre-v3.6.0 configuration and drop the timeout comment. `http-health-check-strategy.ts:35` still tells readers to add `new TypeORMHealthCheckStrategy()` in its example.

**Not blocking, up to you**

`terminus-compat.ts:57-59` still says the `causes` payload is forwarded to the `/health` response. Not in this diff, so a follow-up is fine.

There is no test anywhere that hits `/health`. The e2e spec went away in #4442 and nothing replaced it, which is how the docs drifted in the first place. One assertion on `{ status: 'ok' }` would keep this page honest. Your call whether it belongs here or in a separate PR.

gabriellbui:
Thanks @biggamesmallworld, addressed all three points in bb9872659.

dlhck (approved):

PR #5302, opened 2026-09-07
fix(dashboard): correct "State" translation across locales to mean condition

cuyl:
# Description

In the dashboard i18n catalogs, `msgid "State"` (used in payment details, order lists, and job queue) was mistranslated in several languages as a geographic or administrative division (such as region, province, or US state) rather than a lifecycle condition or status.

This PR updates the translation to mean condition/status in the following locales:
- **ne** (Nepali): `"राज्य"` (province/realm) -> `"अवस्था"` (state/condition)
- **ru** (Russian): `"Регион"` (region) -> `"Состояние"` (state/condition)
- **sv** (Swedish): `"Region"` (region) -> `"Tillstånd"` (state/condition)
- **tr** (Turkish): `"İl"` (province) -> `"Durum"` (state/condition)
- **uk** (Ukrainian): `"Регіон"` (region) -> `"Стан"` (state/condition)
- **zh_Hans** (Simplified Chinese): `"州/省"` (state/province) -> `"状态"` (state/condition)
- **zh_Hant** (Traditional Chinese): `"州/省"` (state/province) -> `"狀態"` (state/condition)

# Breaking changes

None.

# Checklist

📌 Always:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5302"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791382274&installation_model_id=20193&pr_number=5302&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5302&signature=339f3a912b8efb3e5a52f90a7758405cbf213bb852ab33fba6fd931e3d175bf0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -274,7 +274,7 @@ After implementing your changes, stage the changes and commit them. Refer to the
 
 ```
 git add .
-git commit -m "type(scope): Message in present tense"
+git commit -m "type(scope): message in present tense"
 ```
 
 ### Creating a pull request
@@ -354,8 +354,12 @@ myNewApi: number;
 This repo uses [Conventional Commits](https://www.conventionalcommits.org).
 
 ```
-type(scope): Message in present tense
+type(scope): message in present tense
 ```
+
+The subject must not start with a capital letter: `@commitlint/config-conventional` rejects sentence-case,
+start-case, pascal-case and upper-case subjects.
+
 `type` may be one of:
 * **feat** (new feature)
 * **fix** (bug fix)
@@ -380,7 +384,7 @@ type(scope): Message in present tense
 If a commit affects more than one package, separate them with a comma:
 
 ```
-fix(core,common): Fix the thing
+fix(core,common): fix the thing
 ```
 
 If a commit applies to no particular package (e.g. a tooling change in the root package.json), the scope can be omitted.
@@ -394,7 +398,7 @@ Please also make your pull request against the `major` branch rather than `maste
 Example:
 
 ```
-feat(core): Add new field to Customer
+feat(core): add new field to Customer
 
 Relates to #123. This commit adds the "foo" field to the Custom entity.
 
--- a/docs/docs/guides/core-concepts/healthchecks/index.mdx
+++ b/docs/docs/guides/core-concepts/healthchecks/index.mdx
@@ -1,60 +1,32 @@
 ---
 title: "Health Checks"
 metaTitle: "Health Check Endpoint for Monitoring Vendure"
-metaDescription: "How Vendure exposes a /health endpoint for monitoring system dependencies like the database and external services."
+metaDescription: "What the Vendure /health endpoint reports and how to use it as a liveness probe."
 ---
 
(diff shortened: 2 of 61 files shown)


cuyl:

michaelbromley:
Thanks for this, @cuyl. It is merged.

We pushed one commit to your branch and merged it ourselves, rather than leaving review comments and waiting for you to act on them. That is our standard process for well-bounded fixes now, and it is worth explaining so it does not come as a surprise.

The number of pull requests reaching this project has grown faster than the maintainer hours available to review them. Most pull requests that stall do so in the round trip between a review comment and a follow-up push. Taking that step ourselves is how we keep the queue moving.

What we added on top of your work:

- Corrected the same translation in `nl.po`, which had `msgid "State"` as `"Staat"`. Dutch `Staat` reads as a nation or a US state, the same class of mistranslation your seven locales had. It is now `"Status"`, which `nl.po` already uses for the `"Status"` message, so it is the established word in that file rather than a new choice. Checking all 28 catalogs, Dutch was the only one still wrong after your change.

Before merging we checked the thing that could have made this change wrong: whether `msgid "State"` is shared with an address field, which would mean one translation had to serve both senses. It is not. The message has four call sites, all lifecycle state — `fulfillment-details.tsx`, `payment-details.tsx`, the order list column in `orders.tsx`, and the job queue column in `job-queue.tsx`. The geographic sense has its own separate messages, `"State/Province"` and `"State / Province"`, which keep their region wording in every locale you touched, so address forms are unaffected.

Two notes, neither of which needs anything from you. In Turkish, Simplified Chinese and Traditional Chinese, `"State"` now renders the same as that locale's `"Status"` — `Durum`, `状态`, `狀態`. No screen shows both labels, and those languages do not draw the distinction English does, so this reads as correct rather than as a collision. And Spanish, Portuguese, Italian, French and Polish were left alone deliberately: `Estado`, `Stato`, `État` and `Stan` already mean condition as well as territory.

Since this is a translation change, there is no test that can assert a word's meaning. The evidence above is the call-site check instead.

If you disagree with any of the changes, say so and we will revisit them. Thanks again for the fix.

michaelbromley (approved):
Verified that msgid "State" has four call sites, all lifecycle state, and that the geographic sense uses the separate "State/Province" messages which keep their region wording in all seven locales. Swept all 28 catalogs; nl.po was the only one still wrong and is fixed here. Full CI green.

PR #5313, opened 2026-09-08
fix(dashboard): show field name instead of form path as custom field …

### LINKED ISSUE #5246: Dashboard shows a form path instead of a label for localized custom fields
**Describe the bug**

In the Dashboard, a `localeString` or `localeText` custom field with no matching `label` shows an internal form path as its label, for example `translations.0.customFields.tagline`.

A custom field's `label` holds one entry per language. The Dashboard picks the entry matching the operator's display language. When no entry matches, it falls back to a name it prints above the input. For plain fields that fallback is the field name, which reads fine. For localized fields it is the full path into the form value, which includes the translation row index, so the label changes as you switch content language.

**To Reproduce**

1. Declare a localized custom field with no `label`:

```typescript
customFields: {
    Seller: [{ name: 'tagline', type: 'localeString' }],
},
```

2. Open any seller in the Dashboard.
3. The field above the input reads `translations.0.customFields.tagline`.

Declaring a `label` for a language the operator is not using does the same thing:

```typescript
customFields: {
    Seller: [
        {
            name: 'tagline',
            type: 'localeString',
            label: [{ languageCode: LanguageCode.de, value: 'Slogan' }],
        },
    ],
},
```

With the Dashboard display language set to English, the label is again `translations.0.customFields.tagline`.

**Expected behavior**

The label falls back to the field's own name, `tagline`, as it does for plain custom fields.

**Actual behavior**

The label is the form path, `translations.0.customFields.tagline`. Switching the content language from English to German changes it to `translations.1.customFields.tagline`.

**Screenshots**

Image

Image

**Environment**

- Nodejs version: 24.13.0
- Database: sqlite
- Operating System: macOS
- Browser: Chrome 152
- Package manager: bun

**Where it comes from**

`packages/dashboard/src/lib/components/shared/custom-fields-form.tsx` draws every custom field label:

```tsx
<FieldLabel htmlFor={fieldId}>{getTranslation(fieldDef.label) ?? fieldName}</FieldLabel>
```

Three of the five branches that render a field pass `fieldDef.name` as `fieldName`, which gives `tagline`. Two pass `field.name` from the form controller, which gives the full path. Those two are the localized branch (line 252) and the branch for a plain field with a custom form component (line 278), where the path is `customFields.tagline`.

The label lookup has no cross-language fallback:

```tsx
const getTranslation = input =>
    typeof input === 'string' ? input : input?.find(t => t.languageCode === displayLanguage)?.value;
```

**Suggested fix**

Pass `fieldDef.name` in both branches, so a missing label falls back to the field name everywhere. Optionally also fall back to the field name inside the label lookup when a `label` exists but has no entry for the display language.

**Workaround**

Declare a `label` entry for every display language your admin users work in.

Ismatulla:
Two of the five branches in CustomFieldItem passed the react-hook-form controller's field.name (e.g. translations.0.customFields.tagline) as the label fallback instead of fieldDef.name (e.g. tagline), used by the other three branches. This made localized custom fields with no matching label show an internal form path that also changed when switching content language.

Fixes #5246

# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5313"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791442713&installation_model_id=20193&pr_number=5313&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5313&signature=f26f7f6e602494d4a146736b18bb50dc43bab593ee65a0d19600916d1e475efc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Ismatulla:

Ismatulla:
recheck

michaelbromley:
Thanks for this, @Ismatulla. It is merged.

We pushed a commit to your branch and merged it ourselves, rather than leaving review comments and waiting for you to act on them. That is our standard process for well-bounded fixes now, and it is worth explaining so it does not come as a surprise.

The number of pull requests reaching this project has grown faster than the maintainer hours available to review them. Most pull requests that stall do so in the round trip between a review comment and a follow-up push. Taking that step ourselves is how we keep the queue moving.

What we added on top of your work:

- Kept the form path as the element id. `CustomFieldFormItem` built `fieldId` from the same `fieldName` prop it used for the label, so changing that prop to `fieldDef.name` also changed the id. The product variant page renders this form once per price, so every price row would then have produced the same `field-<name>` id. The prop is now `fieldPath`, which carries the form control's path and feeds only the id, and the label fallback reads `fieldDef.name` from the field definition the component already has. Splitting the two means the label is right and the id stays unique.
- Applied the same `fieldPath` to the three branches that were already passing `fieldDef.name`. Those had the duplicate id problem before this pull request, so this fixes it for struct fields, struct lists and plain fields as well.
- Added an end-to-end test in `packages/dashboard/e2e/tests/catalog/custom-fields.spec.ts`, with a `localeString` custom field carrying no label in the e2e fixture. It fails on master and passes with your fix, so the behaviour stays covered if someone refactors this later.
- Merged master into the branch. That was not needed here, since your branch was only a day old, and it is why the commit list on this pull request is longer than the change. The merge is squashed, so master gets one commit.

We also reproduced the reported bug independently before merging. On master the SEO tab renders the label as `translations.0.customFields.seoKeywords`, which Playwright captured verbatim in its failure context, and the test cannot find a label reading `seoKeywords`. With the fix, all 22 tests in that suite pass.

If you disagree with any of the changes, say so and we will revisit them. Thanks again for the fix.

michaelbromley (approved):
Reproduced #5246 with a new e2e fixture: on master the label renders translations.0.customFields.seoKeywords, captured verbatim in Playwright's failure context. Split fieldName into fieldPath for the element id and fieldDef.name for the label fallback, which also fixes the duplicate DOM ids the three non-locale branches had already. all-passed green with build, unit 20/22/24, e2e on four databases and all four dashboard e2e shards. The deploy failure is the Vercel token being unavailable to fork PRs.

PR #5314, opened 2026-09-08
feat(cli): let plugins run after a Console project link

dlhck:
`vendure console link` owns linking: it resolves the Console endpoints, approves custom ones, opens the browser, polls for approval, validates the Project Link Manifest, writes it atomically, and updates `.gitignore`. A plugin that needs its own setup after linking can now extend that flow without replacing the `console` command.

## What this adds

**`CliPlugin.afterConsoleLink`** runs after the manifest is written. It receives the resolved project root, manifest and path, Console endpoints, link outcome, abort signal, reporter, confirmation function, terminal mode, and `--force` state.

Each hook gets its own context, manifest, and endpoint objects. Hooks run in plugin order, and the first failure stops the remaining hooks without removing the manifest.

**Repeating a link repairs it.** In an already-linked project, the command reuses the manifest and runs hooks with `outcome: 'repaired'`. It creates no second Project Link. `--force` creates a different Project Link and runs hooks with `outcome: 'linked'`. `--yes` answers every confirmation owned by the CLI without changing repair into relink behavior.

**Official Console origins are one paired rule.** Production and staging app/API origins must be used as matching pairs. `endpoints.official` is `'production'`, `'staging'`, or `undefined` for custom and loopback origins. Public API and documentation remain specific to the Console integration.

**The shared registry stores extension values generically.** `CliCommandContext.getPluginExtensions()` returns registered values by extension-point name. It preserves each extension value opaquely, while the `console` command maps `afterConsoleLink` values into its local hook dependency. This removes the mutable global hook accessor and leaves future extension metadata intact.

**`CLI_PLUGIN_EXTENSION_POINTS`** lets a plugin refuse to load when the installed CLI does not support its extension. The full marker list is pinned by a test.

The registry also rejects every duplicate plugin ID and deduplicates repeated package entries, including plugins that contribute no command.

## Checks

Environment: Darwin arm64, Bun 1.4.0, Node v24.16.0.

- Built `@vendure/common` and `@vendure/core` so the doctor suites could collect.
- `bun run build` in `packages/cli`: passed.
- Focused Console/shared suite: 8 files passed, 177 tests passed.
- Full serialized CLI Vitest suite: 36 files passed, 553 tests passed, 1 skipped.
- PR-file ESLint: 6 errors and 128 warnings. All 6 errors are in `shared/resolve-cli-plugins.ts` at code already present on `minor`; that file is unchanged from the reviewed head. The seventh error previously reported was a `status` name shadow in `console.ts`, which this update removes.

Relates to EE-308

### DIFF
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -793,11 +793,12 @@ export default defineCliPlugin({
 });
 ```
 
-The hook receives the resolved `projectRoot`, the `manifest` as written and its
-`manifestPath`, the Console `endpoints` the link was made against, the
-`AbortSignal` the command runs under, the command's `reporter`, a `confirm`
-function for yes/no questions, `isNonInteractive`, and whether `--force` was
-given.
+The hook receives the resolved `projectRoot`, the `manifest` and its
+`manifestPath`, the Console `endpoints` the link was made against, whether this
+run linked or repaired (`outcome`), the `AbortSignal` the command runs under,
+the command's `reporter`, a `confirm` function for yes/no questions,
+`isNonInteractive`, and whether `--force` was given. Each hook gets its own
+context, so one hook cannot change what a later one reads.
 
 Do not replace the `console` command to do this. Linking stays one
 implementation, so every project links the same way whether or not a plugin is
@@ -812,16 +813,36 @@ it.
   command says so rather than reporting that nothing changed.
 - Hooks run in `vendure.cli.plugins` order, and the first to throw stops the
   rest.
-- Running `vendure console link` again creates a **new** Project Link rather than
-  repairing the existing one. Work that has to be repeatable, such as replacing
-  a credential a developer has lost, belongs in a command of the plugin's own
-  that reads the existing manifest. Do not send people back to `console link`
-  for it.
+- It has to be repeatable, because `vendure console link` repairs a link as well
+  as making one (see below).
 
 Throwing is for work that failed, not for work that cannot be done here. A hook
-that needs an answer it cannot get, such as a sign-in with no terminal attached,
-should say what is missing through `reporter` and return. The command still
-exits 0, because linking is what it was asked to do and it did it.
+that needs something it cannot get in this run should say what is missing
+through `reporter` and return. The command still exits 0, because linking is
+what it was asked to do and it did it.
+
+`context.confirm` asks a yes/no question and rejects when `isNonInteractive` is
+`true`, since there is nobody to answer. Check `isNonInteractive` first.
+Multi-choice questions belong in the plugin's own command, where it owns the
+whole screen.
+
+#### Repeating a link repairs it
+
+Run `vendure console link` in a project that is already linked and it reuses the
+manifest on disk and runs the hooks again, with `context.outcome` set to
+`'repaired'`. It does not create a second Project Link, which would abandon the
+first in Console over a problem that is local to the checkout.
+
+That makes `vendure console link` the command a developer runs to put back
+something the project needs and no longer has, such as a credential a hook
+writes. A plugin does not need a repair command of its own, and should send
+people here instead.
+
+`--force` is how a checkout is linked to a **different** Console Project: it
+skips the reuse, creates a new Project Link, and the hooks run with an `outcome`
+of `'linked'`. A hook that rotates a remote credential should treat `'repaired'`
+as the case that needs confirming, because the credential it is replacing may
+still be live for someone else.
 
 `context.endpoints.areDefault` is `false` when the link was made against
 endpoints set by `VENDURE_CONSOLE_LINK_URL` and `VENDURE_CONSOLE_LINK_API_URL`
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -367,6 +367,16 @@ npx vendure console unlink --force
 
 `--force` does not bypass ambiguous project selection or a manifest that belongs to another project root.
 
+Running `vendure console link` in a project that is already linked repairs it rather than replacing it: the manifest
+is reused and any plugin setup runs again. Where a plugin is registered, that is confirmed first, because the manifest
+may have arrived with a clone. Use `--yes` to answer that confirmation in advance:
+
+```bash
+npx vendure console link --yes
+```
+
+`--force` is different: it links this checkout to a **different** Console Project, which creates a new Project Link.
+
 ### Manifest and source control
 
 The v1 manifest contains identity metadata only:
@@ -857,13 +867,13 @@ The pair is matched as a pair, so a production app origin with a staging API is
 not official. A loopback pair is not official either; it is a development and
 test convenience.
 
-It says where this run will talk, not where the manifest came from. The manifest
-records no origin, so on a repair the Console that issued its project and
-account identifiers is unknown, and a `'production'` answer does not vouch for
-them. A hook that acts on the identifiers, rather than only on the origin,
+`official` says where this run will talk, not where the manifest came from. The
+manifest records no origin, so on a repair the Console that issued its project
+and account identifiers is unknown, and a `'production'` answer does not vouch
+for them. A hook that acts on the identifiers, rather than only on the origin,
 should treat a repair as unverified provenance.
 
-It is a fact, not a permission. Approving a custom endpoint approves it for
+`official` is a fact, not a permission. Approving a custom endpoint approves it for
 creating a Project Link, not for receiving anything secret. A hook that holds
 credentials checks `official` before it reads a credential file or sends a
 request, and refuses when it is `undefined`. Do not draw the same conclusion
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -790,7 +790,7 @@ an `afterConsoleLink` hook:
 import { defineCliPlugin } from '@vendure/cli';
 
 export default defineCliPlugin({
-    id: '@example/platform-cli-plugin',
+    id: '@example/setup-cli-plugin',
     commands: [],
     afterConsoleLink: async context => {
         // Everything the command already worked out, so the hook never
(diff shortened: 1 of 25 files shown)


michaelbromley (changes requested):
The feature design is right. `afterConsoleLink` is the correct shape, running hooks after the manifest is written is the correct ordering, and giving each hook its own context with a cloned manifest is a real property backed by a real test: `console-link-hook.spec.ts:289` mutates from the first hook and checks the second is unaffected. `console-origins.ts` uses literals, matches origins as a pair, requires https, and rejects credentials, path, query and fragment, with lookalike-host and subdomain cases covered.

The PR body claims 545 tests pass and 1 skipped. I get 506 collected, 501 passed, 1 skipped, and 4 files failing to collect. The failures are all `doctor.spec.ts` needing a built `@vendure/core`, which is pre-existing and unrelated to this diff. The console and shared subset is 284 passed and clean. The code is fine and the number in the body does not reproduce.

## Must fix

**1. The exit-code test does not assert the exit code or the reporter level.** `packages/cli/src/commands/console/console-link-hook.spec.ts:336`

```ts
it.each([0, 2])('says what survived when a hook exits with code %i', async exitCode => {
    ...
    await expect(
        runLink(root, registry, { reporter: recordingReporter(messages) }),
    ).rejects.toBeInstanceOf(CliCommandExit);
```

Commit `13d2f015b` says the test "asserted only the error type, so a regression that rewrote the code to 1 would have passed it". That is still what this test does. Nothing asserts `error.exitCode`.

The behaviour that commit added cannot be tested here either. `console.ts:433` calls `reporter.warn` for code 0 and `reporter.error` for anything else. `recordingReporter` at line 427 pushes every level into one flat array, so the level is gone before the assertion runs. `it.each([0, 2])` therefore runs the same two assertions twice, and nothing covers the choice between `warn` and `error`. Capture the level with the message, assert `.rejects.toMatchObject({ exitCode })`, and assert the level per case.

**2. `--force` over a valid manifest has no test.** This is the only route to creating a second Project Link in an already-linked checkout, it is the flag the guide leads with, and it is the branch where `outcome` must be `'linked'` and `context.force` must be `true`. Searching for `force: true` in `console.spec.ts` gives three hits: a project-root mismatch failure, a replacement of an `'{invalid'` manifest, and one for `unlink`. This PR rewrote the test at line 442 to use `'{invalid'` and did not add the valid-manifest case anywhere else. Nothing asserts `context.force === true` either; `console-link-hook.spec.ts:59` only asserts `false`. Add a test: valid manifest plus `--force`, fetch called, new manifest written, hook sees `outcome: 'linked'` and `force: true`.

**3. The `--force` help text describes the old behaviour.** `packages/cli/src/commands/console/command.ts` says "Confirm replacement or unlink without an interactive prompt". The guide says `--force` links this checkout to a different Console Project, which creates a new Project Link. That is now its main meaning on `link`, and `vendure console --help` does not say it. `command.spec.ts` asserts the flag exists and asserts nothing about its text.

## Should fix

**Two pairing rules for the same origins.** `resolveConsoleEndpoints` at `console.ts:233` still enforces its own production-only rule, and `console-origins.ts` has a paired table that covers staging as well. The two disagree: a production console with a staging API throws, but a staging console with `https://api.evil.test` passes and only reaches the custom-endpoint prompt. Express the rule once in `console-origins.ts`: if exactly one half of the pair is an official origin, refuse.

**`--yes`, `--force` and `--allow-custom-console` answer three overlapping subsets of the prompts.** The command has four confirmations: replace manifest, remove manifest, custom endpoints, and run plugin setup. `--force` answers the first two, `--allow-custom-console` the third, and `--yes` only the fourth. Someone who passes `--yes` to a scripted run still hits three prompts, which is why the help text needs two sentences to explain what the flag does not do. Either make `--yes` answer every CLI confirmation and leave `--force` its separate meaning, or name the flag for the one prompt it answers.

**The registry hard-codes one command's extension point.** `command-registry-store.ts:80` and `cli-plugin.ts:1` both import from `../commands/console/console-link-hook`, so the generic plugin registry in `shared/` knows about the `console` command. `CLI_PLUGIN_EXTENSION_POINTS` promises that entries are only ever added, so a second extension point needs another field in `RegistryState`, another getter, another global setter and another line in `cli.ts`. A `Map` keyed by extension point holds any number of them without a new field each.

**Two mechanisms for one dependency.** `ConsoleCommandDependencies.hooks` is dependency injection. `getConsoleLinkHooks()` at `console.ts:110` reads a mutable module-level global that `cli.ts` sets. Tests use the first and production uses the second. The comment at `console-link-hook.ts:176` acknowledges this, but the fix is to make `CliCommandDefinition` carry the registry or a hook accessor. Nothing breaks today in a one-shot CLI, and `setConsoleLinkHooks` is still an exported mutable global in a package that also exports a plugin API.

**`ConsoleReporter` is in the wrong file.** It is defined in `console-link-hook.ts`, imported by `console.ts`, and re-exported from `console.ts` at line 42 so three test files do not have to change their imports. It is the command's output interface. Give it its own module and update the three imports.

**Pin `CLI_PLUGIN_EXTENSION_POINTS`.** The docblock says entries are never removed or renamed, and `console-link-hook.spec.ts:228` only checks that two entries are present. You pinned the origin literals in a dedicated spec so an edit fails a test. Do the same here, with `toEqual` and the full list.

**`isNonInteractive` is read twice.** `createConsoleLinkContext` snapshots `dependencies.isNonInteractive()` into the context, and `confirmForHook` calls it again at call time. Snapshot once and pass the boolean.

## Comments and prose

The code carries a lot of justification that belongs in the PR description.

The docblock on `repair()` at `console.ts:304-322` is 19 lines and mostly argues the design decision rather than describing the function. `runConsoleLinkHooks` explains why rolling back would be worse, `createConsoleLinkContext` explains what a shared object would let a plugin do, and the block at `console.ts:249-255` is seven lines of rationale above one function call. Each is defensible on its own. Together they read as a review thread pasted into the file.

Some comments describe the change rather than the code. `usesCustomRemoteEndpoints` at `console.ts:925-932` explains what an earlier version said wrongly, which tells a reader who never saw the diff nothing about what the function does. The same sentence appears again in `console.spec.ts`. The first line, "Whether these origins are a remote Console that is not Vendure's", is the whole comment.

Several sentences are copied into three or four places. "It is a fact, not a permission" is in `console-link-hook.ts:43`, in `console-origins.ts`, in the guide and in the PR body. "Refusing staging is as wrong as accepting an unknown host" is in the type docblock and in `console-origins.spec.ts:39`. Each copy has to be edited by hand, and a missed one leaves two files saying different things.

The guide at `docs/docs/guides/developer-guide/cli/index.mdx` restates around 130 lines of what the `ConsoleLinkContext` and `ConsoleLinkHook` docblocks already say. The type already points at the guide, so let the guide own the normative text and cut the docblocks to a line each. The guide also argues a threat model at a reviewer rather than a user. The paragraph about two environment variables sending a hook to staging reaches the useful instruction four clauses in, which is to compare against `'production'` rather than against `undefined`. Lead with that and drop the rest.

"Mint a second Project Link" appears in the code, the guide and the PR body. The API creates one. Say create.

## Smaller test points

- `console-origins.spec.ts:21` and `:32` both assert the staging pair resolves to `'staging'`, identically. Collapse into one test.
- `console-origins.spec.ts:48` tests a trailing slash, which the production caller cannot produce because `baseUrl()` returns `url.origin`.
- `console-link-hook.spec.ts:62` asserts `context.signal.aborted` is `false` after a successful run, which could not be otherwise.
- `console-link-hook.spec.ts:141` loops three actions and asserts once at the end, so a regression in one of the three does not say which. Use `it.each`.
- No test covers a hook seeing `official: 'production'` on the link path. Staging is only covered on repair.

dlhck:
@michaelbromley I addressed every point in the review in `c6c9a0f5f00525dd6baf79d20b5853332c6758d4`.

Must-fix changes:

- The hook-exit test now asserts the exact `CliCommandExit.exitCode` and records the reporter method. Exit 0 must use `warn`; exit 2 must use `error`.
- A valid-manifest `--force` test now verifies both requests, the new manifest on disk, `outcome: 'linked'`, and `force: true`.
- `--force` help now states that it creates a different Project Link or removes the current link without confirmation. The text is pinned by a test.

Should-fix changes:

- Official origin validation now has one rule in `console-origins.ts`. Any official app or API origin must be paired with the matching origin from the same environment, including staging and cross-environment mismatches.
- `--yes` now answers every CLI-owned confirmation: custom endpoints, manifest replacement/removal, and repair setup. It does not answer plugin-owned questions and does not force a relink. `--allow-custom-console` remains the narrow non-interactive approval for custom endpoints.
- The command registry now stores extensions in a generic map and exposes them through `CliCommandContext.getPluginExtensions()`. Values remain opaque and retain object identity. The `console` command performs its own shallow hook mapping. `cli-plugin.ts` still imports `ConsoleLinkHook` only to type the public `CliPlugin.afterConsoleLink` field; the registry and command context have no Console-specific storage.
- The mutable global hook setter/getter is gone. Production and tests now use the same dependency path through the command context.
- `ConsoleReporter` has its own module, and all imports and public exports point to it.
- The extension marker test now pins the complete list with `toEqual`.
- Non-interactive mode is sampled once per hook context and reused by `confirm`; a test makes the source return different values and verifies one read.

Prose and smaller test changes:

- I reduced the long implementation-rationale comments and API docblocks, left the guide as the normative explanation, led with the production comparison rule, removed repeated slogans, and changed “mint” to “create.”
- I collapsed the duplicate staging assertion, removed the caller-impossible trailing-slash test and successful-signal assertion, converted the three-action loop to `it.each`, and added a production-link hook test.
- The PR body now reports reproducible counts instead of the old 545-test claim.

Validation on this checkout (Darwin arm64, Bun 1.4.0, Node v24.16.0):

- `bun run build` in `packages/cli`: passed.
- Focused suite: 8 files, 177 tests passed.
- Full CLI suite, after building `@vendure/common` and `@vendure/core`, serialized with one worker: 36 files passed; 553 tests passed; 1 skipped.
- ESLint across PR CLI files: 6 errors, 128 warnings. The six errors are the non-null assertion, shadowing, and template-expression findings in `shared/resolve-cli-plugins.ts`; each exists on `minor`, and that file is unchanged from the reviewed head. The prior seventh error was the `status` shadow in `console.ts`, now fixed.
- Current-head CI completed with 27 successful checks, 8 intentional skips, 0 failures, and 0 pending. It covered build and codegen, unit tests on Node 20/22/24, PostgreSQL/MySQL/MariaDB/sql.js e2e, five package-install consumers, docs/MDX, Sonar, commit format, CLA, CodeRabbit, and previews.

There were no inline review threads to resolve.

PR #5315, opened 2026-09-08
feat(cli): settle a Console session with the Project Link approval

dlhck:
**Stacked on #5314 (stack #5316). Review that first; this PR's base is that branch, not `minor`.**

**Both land in `minor` together, then a prerelease is published.** #5314 adds the hook and runs it; it does not add `ConsoleLinkContext.session`, which arrives here. A plugin registered against #5314 alone therefore runs normally and finds no session, which is the absence of the auth capability rather than a broken hook — a consumer is expected to report incomplete credential setup and carry on. An intermediate canary cut between the two would behave that way; final consumer validation targets a release containing both.

Linking already takes the developer through a browser approval, and Console already mints a CLI Session from a signed-in browser. They were two approvals because nothing joined them: the link approval could not settle a login, so a plugin needing a session had to start a second browser round trip of its own.

Console now advertises `cli-auth` on the create response and accepts a login request alongside a decision (vendurehq/console#129, merged). This carries one: it generates an S256 challenge, binds a loopback callback before the browser opens, puts the five parameters Console reads on the verification URL it already returned, and exchanges the code at Console's `POST /v1/auth/cli/token`. The result reaches a plugin as `ConsoleLinkContext.session`.

The route, the grant types and the field names are Console's, so every client of that exchange speaks the same one.

## Nothing here can fail the link

The poll owns the link and the callback owns only the session. The manifest is written before the exchange runs, so an interrupt during the login cannot leave an approved Project Link with nothing on disk. Beyond that, the link stands with no session and a report saying which when:

- no plugin registered a hook, so no token is minted at all;
- the origins are not an official Console;
- the Console does not advertise `cli-auth`;
- the shell is attached over SSH, so a browser cannot return to this machine;
- the callback port cannot be bound;
- the browser could not be opened here, in which case the URL offered for another machine carries no callback address;
- the approval happened elsewhere, was refused, or the exchange did not complete.

The listener has no deadline of its own; the link expiry governs. It is accepted briefly after the poll returns, because approval mints the code and returns the redirect in one transaction, so the browser's navigation and the next poll run on unrelated clocks.

The CLI writes no credential. A plugin that stores the session owns that, including keeping a user-scoped token out of the project directory. A repair never carries a session: it asks Console nothing and opens no browser.

## Verification

The unit and integration tests use a contract fixture and a real loopback listener, with no network and no database: the official origins open the gate, every request is served in memory, and the stub browser follows the approval to the advertised callback. They assert that the exchange proves possession of the verifier behind the challenge, that the fallback URL carries no callback address, and that the manifest is on disk before the exchange can be interrupted.

Separately, the Console repository extracted this module read-only and drove its own acceptance suite with it against a real Console on real PostgreSQL: 15 cases pass, covering both the `cli` and `create` client values.

The auth primitives are deliberately not exported from the package index. A loopback-listener factory with no origin check is not something to hand plugin authors on a semver promise; a cross-repository test pins the module path instead.

## Checks

Typecheck clean, 584 tests pass and 1 skipped, lint at the 7 errors already present on `minor`. `build_and_test.yml` filters `pull_request` to `master`, `major` and `minor`, so a PR based on a feature branch gets no substantive run; the full matrix is dispatched against this branch through the workflow's declared `workflow_dispatch` trigger instead.

Relates to EE-308

### DIFF
--- a/docs/docs/guides/developer-guide/cli/index.mdx
+++ b/docs/docs/guides/developer-guide/cli/index.mdx
@@ -804,8 +804,36 @@ export default defineCliPlugin({
 ```
 
 The context contains `projectRoot`, `manifest`, `manifestPath`, `endpoints`,
-`outcome`, `signal`, `reporter`, `confirm`, `isNonInteractive`, and `force`.
-Each hook gets its own manifest and endpoint objects.
+`outcome`, the optional `session`, `signal`, `reporter`, `confirm`,
+`isNonInteractive`, and `force`. Each hook gets its own manifest and endpoint
+objects. A session is also copied when the hook requested one and the login
+succeeded.
+
+A hook does not receive a Console session unless it requests one. Wrap a hook
+with `requiresSession: true` when it needs the session from the same browser
+approval:
+
+```ts
+export default defineCliPlugin({
+    id: '@example/credential-cli-plugin',
+    commands: [],
+    afterConsoleLink: {
+        requiresSession: true,
+        hook: async context => {
+            if (!context.session) {
+                context.reporter.warn('No Console session was obtained.');
+                return;
+            }
+            await storeCredential(context.session);
+        },
+    },
+});
+```
+
+The session is user-scoped and can authorize access across the user's Console
+projects. The CLI does not store it. If a plugin stores the session, use an OS
+credential store or a user-specific configuration directory outside the
+Vendure project. Never write it to the project directory or commit it.
 
 Do not replace the `console` command for post-link setup. Register the hook.
 
--- a/packages/cli/src/commands/console/cli-auth.spec.ts
+++ b/packages/cli/src/commands/console/cli-auth.spec.ts
@@ -178,24 +178,46 @@ describe('parseConsoleSession()', () => {
 
     it('accepts a response with no refresh token', () => {
         expect(
-            parseConsoleSession(
-                { access_token: 'vcli_access', token_type: 'Bearer', expires_in: 60 },
-                now,
-            ).refreshToken,
+            parseConsoleSession({ access_token: 'vcli_access', token_type: 'Bearer', expires_in: 60 }, now)
+                .refreshToken,
         ).toBeUndefined();
     });
 
     it.each([
         ['no access token', { token_type: 'Bearer', expires_in: 60 }, /invalid access token/],
-        ['an empty access token', { access_token: '', token_type: 'Bearer', expires_in: 60 }, /invalid access token/],
-        ['an unsupported token type', { access_token: 'a', token_type: 'Basic', expires_in: 60 }, /token type/],
+        [
+            'an empty access token',
+            { access_token: '', token_type: 'Bearer', expires_in: 60 },
+            /invalid access token/,
+        ],
+        [
+            'an unsupported token type',
+            { access_token: 'a', token_type: 'Basic', expires_in: 60 },
+            /token type/,
+        ],
         ['no lifetime', { access_token: 'a', token_type: 'Bearer' }, /token lifetime/],
-        ['a non-numeric lifetime', { access_token: 'a', token_type: 'Bearer', expires_in: '60' }, /token lifetime/],
+        [
+            'a non-numeric lifetime',
+            { access_token: 'a', token_type: 'Bearer', expires_in: '60' },
+            /token lifetime/,
+        ],
         // A lifetime at or before now is a dead token described as a live one.
         ['a zero lifetime', { access_token: 'a', token_type: 'Bearer', expires_in: 0 }, /token lifetime/],
-        ['a negative lifetime', { access_token: 'a', token_type: 'Bearer', expires_in: -3600 }, /token lifetime/],
-        ['an absurd lifetime', { access_token: 'a', token_type: 'Bearer', expires_in: 1e15 }, /token lifetime/],
-        ['a fractional lifetime', { access_token: 'a', token_type: 'Bearer', expires_in: 1.5 }, /token lifetime/],
+        [
+            'a negative lifetime',
+            { access_token: 'a', token_type: 'Bearer', expires_in: -3600 },
+            /token lifetime/,
+        ],
+        [
+            'an absurd lifetime',
+            { access_token: 'a', token_type: 'Bearer', expires_in: 1e15 },
+            /token lifetime/,
+        ],
+        [
+            'a fractional lifetime',
+            { access_token: 'a', token_type: 'Bearer', expires_in: 1.5 },
+            /token lifetime/,
+        ],
         ['a response that is not an object', 'nope', /malformed token response/],
     ])('refuses a response with %s', (_label, value, message) => {
         expect(() => parseConsoleSession(value, now)).toThrow(message);
--- a/packages/cli/src/commands/console/cli-auth.spec.ts
+++ b/packages/cli/src/commands/console/cli-auth.spec.ts
@@ -219,6 +219,16 @@ describe('parseConsoleSession()', () => {
             /token lifetime/,
         ],
         ['a response that is not an object', 'nope', /malformed token response/],
+        [
+            'an empty refresh token',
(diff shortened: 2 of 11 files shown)


michaelbromley (changes requested):
The security shape here is right: PKCE, the callback bound before the browser opens, `127.0.0.1` rather than `localhost`, the origin gate checked before any token is requested, the manifest written before the exchange, and no credential written by the CLI itself. Two things need work: which hooks receive the session, and the amount of comment around the code.

## Must change before merge

**1. Any registered hook causes a session to be minted, and every hook receives it.** `startConsoleLogin` gates on nothing but `dependencies.hooks.length === 0`. A plugin that registered `afterConsoleLink` to write a config file now receives an access token and a refresh token. The TSDoc on `ConsoleSession` describes those as worth as much as the person's Console password across every project they own. Adding `requiresSession?: boolean` to `RegisteredConsoleLinkHook` and filtering on it is a three-line change today. After this ships in `minor` it is a semantics change to a public plugin API.

**2. The session is shared by reference while the manifest is cloned.** `createConsoleLinkContext` clones the manifest with `structuredClone`, under a comment explaining that a shared object would let the plugin listed first decide what the plugin listed second sees. Two lines below, `session: inputs.session` hands one object to every hook. One hook can set `context.session.accessToken` and change what the next hook reads. Clone or freeze the session too.

**3. The losing grace-period sleep is never cancelled.** `completeConsoleLogin` races `login.code()` against `dependencies.sleep(CALLBACK_GRACE_MS, signal)`. When the code wins, which is the normal case, nothing clears the sleep. `abortableSleep` builds a plain `setTimeout` and neither clears nor unrefs it, and the success path does not call `process.exit`. The process stays alive for two seconds after printing its last line. Give that sleep its own `AbortController` and abort it once the race settles.

**4. `expect(test.delays).toContain(CALLBACK_GRACE_MS)` asserts on a value two constants share.** `CALLBACK_GRACE_MS` and `POLL_INTERVAL_MS` are both `2_000`, so the assertion cannot tell which sleep it observed, and a poll retry in either test would satisfy it. The sleep stub keys on the same value at `if (milliseconds === CALLBACK_GRACE_MS)`. Assert `toEqual([CALLBACK_GRACE_MS])`, and give the stub a distinct value to match on.

**5. The CLI guide now states something false.** `docs/docs/guides/developer-guide/cli/index.mdx` lists what a hook receives, member by member, and `session` is missing from that list. The same paragraph ends "Each hook gets its own context, so one hook cannot change what a later one reads", which stops being true the moment the session is shared. Fixing point 2 restores it; the sentence has to be checked either way. The warning about keeping the token out of the project directory lives only in a TSDoc comment, which a plugin author will not open. Put it in the guide.

## Should change

6. `CliAuthClient = 'cli' | 'create'` and the optional `client?` parameter on `cliAuthSearchParams` exist for a test in another repository. Nothing here passes `'create'`. Hardcode `'cli'` and let the Console repo's fixture supply its own value.
7. `CLI_AUTH_CAPABILITY` is exported from `packages/cli/src/index.ts`, but `request.supports` never reaches `ConsoleLinkContext`, so a plugin author cannot act on the constant. Drop it until one can. `ConsoleSession` and `CLI_AUTH_CAPABILITY` are also missing the `@since 3.8.0` that every neighbouring exported type carries.
8. `vi.spyOn(cliAuth, 'startLoopbackCallback')` bypasses `ConsoleCommandDependencies`, which every other test in the directory uses, and couples the test to the module's internal shape. Simulate the bind failure through the injection point.
9. Cut the comment volume. `cli-auth.ts` is 73 comment lines out of 227. The word "settle" does four jobs across the diff: settling a login, settling a link, `settle?.(code)` resolving a promise, and meaning an approval rather than a refusal. Lines to remove:
    - the `shellIsRemote` sentence about what being wrong the other way costs
    - the `MAX_TOKEN_LIFETIME_SECONDS` line defining the bound by what it is not
    - the `noSessionFromThisLink` sentence about running the command twice
    - the "used to" narration in `link()`, which describes an ordering no reader has seen
    - `// The first and largest claim of the correction:` in `console-login.spec.ts`, which points at a review thread

    The `server.once('error', reject)` comment is the one to keep: it names a hazard the code cannot show.
10. Add a test for two registered hooks both receiving the session, which is the behaviour point 2 turns on. Add one for the loopback port being released after `link()` returns. Closing twice is tested in isolation, but nothing checks that the command closes it at all.
11. `console-login.spec.ts` adds a third copy of `vendureProject()`, a second identical `jsonResponse()` and a third fake Console responder, and defines a second `runLink` in a directory where `console-link-hook.spec.ts` already has one driving a real HTTP server. `console.fixtures.ts` already holds `createResponse()` and is where these belong.

## Smaller points

12. `url.searchParams.set` in `startConsoleLogin` overwrites any of the five names Console already placed on `verificationPath`. Today that path carries `?link=<id>` only, so a Console that adds one of those names later would break the exchange with no warning.
13. `parseConsoleSession` drops a `refresh_token` that is not a non-empty string, while every other malformed field in that function throws.
14. The non-official-Console branch is the only one of four that prints nothing, and a test asserts that silence. A developer running against a custom Console gets no reason for the missing session.
15. `console-login.spec.ts` names the result of `runLink()` `test` in every case. Call it `run` or `result`.

The `parseConsoleSession` table is thorough: exact error matching, both sides of the lifetime bound, non-integer and non-object inputs. The interrupt-during-exchange test pins the manifest ordering the right way, asserting exit 130, the file on disk, and the absence of the "no manifest changed" line together.

dlhck:
@michaelbromley Addressed in `a735061f3` after merging the lower PR fixes from `c6c9a0f5f`.

Session handling and cleanup:

- A plain `afterConsoleLink` function remains a generic hook and never causes a session to be minted. A hook opts in with `{ requiresSession: true, hook }`. Only opted-in hooks receive the session.
- Each opted-in hook receives a structured clone of the session. A test mutates the first hook's token and verifies that the second hook still reads the original token.
- The callback grace delay now has its own abort controller. The command aborts the losing delay when the callback wins. The grace value is 2,500 ms, distinct from the 2,000 ms poll interval, and the tests assert the exact delay list.
- The loopback callback factory is now a command dependency, so the bind-failure test uses the injection point instead of spying on the module.
- The link-level test verifies that the loopback port is closed after `link()` returns. The close path already released it, so this point needed coverage rather than a production change.

API, validation, and documentation:

- `cliAuthSearchParams()` always sends `client=cli`. I removed `CliAuthClient` and the optional client argument. A Console-side fixture that needs `create` must now supply that value itself.
- `CLI_AUTH_CAPABILITY` is no longer public. The capability string is private to the console command. `ConsoleSession` and the new public hook registration types have `@since 3.8.0`.
- A verification URL that already contains any reserved auth parameter is preserved. The optional login is skipped with a specific warning instead of overwriting the query.
- An omitted refresh token remains valid. A present empty or non-string refresh token now makes the token response invalid, consistent with the other present malformed fields.
- A session-requesting hook against a custom or loopback Console now gets the reason that no session was offered. Generic hooks remain silent because they did not request one.
- The guide lists the optional session, shows the explicit opt-in form, states that each hook gets its own copy, and warns plugins to use an OS credential store or user-specific directory outside the project. It says never to commit the session.
- I removed the requested review-history and threat-model comments, renamed the overloaded callback resolver, renamed `test` results in `console-login.spec.ts`, and moved the duplicated project, response, and fake Console helpers into `console.fixtures.ts`.

Validation on macOS arm64, Node 24.16.0, Bun 1.4.0:

- `bun run build:core-common`: passed.
- `packages/cli` `bun run build`: passed.
- Full CLI suite outside the socket-restricted runner: 38 files passed, 600 tests passed, 1 skipped, 0 failed.
- Focused auth and hook set: 3 files passed, 67 tests passed.
- Docs validation: 856 files passed, 4,970 code blocks checked, 0 errors.
- ESLint on the 10 changed TypeScript files: 0 errors, 26 `require-await` warnings in async test callbacks.

The repository-wide ESLint process exhausted both 4 GB and 8 GB heaps before it emitted diagnostics, so I am not attributing or repeating the earlier seven-error baseline.

PR #5321, opened 2026-09-08
test(core): cover order-level side-effect promotions being kept

brmk:
Follow-up to the review on #5279.

**Test.** Adds the one case in `promotions which apply no discount` that nothing covered: a `PromotionOrderAction` whose `execute` returns `0` but defines `onActivate` is kept on `Order.promotions`. This is the contract the docs note describes, and it was only asserted for item-level actions. Removing `|| promotion.hasSideEffects` from `removeIneffectivePromotions` fails exactly the two side-effect tests (item-level and Order-level).

**Docs.** Shortens the note in the promotions guide to say that the presence of an `onActivate` hook, not its body, is what keeps a Promotion recorded.

Relates to #5279


gabriellbui:
Merge lane session `53590d7e` (@gabriellbui) is working this pull request.

Current stage: 1 (isolate). Claim expires 2026-09-10T15:22:11Z unless refreshed.

gabriellbui:
Thanks for this, @brmk. One thing to hand back before it goes in.

The new test does not exercise `applyOrderPromotions`. I put an early `return` at the top of `applyOrderPromotions` and ran the spec. 21 of the 55 tests in `order-calculator.spec.ts` fail, and `is not added to the Order when its action is Order-level` still passes.

`applyOrderItemPromotions` (`order-calculator.ts:214`) is why. That loop runs over every promotion whose conditions pass and calls `addPromotion` whatever the action type. So the item loop records the promotion, and `removeIneffectivePromotions` removes it again. An assertion on final state cannot tell which loop recorded it.

Commenting out `removeIneffectivePromotions` shows the same thing. The new test then fails alongside the pre-existing item-level `is not added to the Order`.

Two ways forward, whichever you prefer:

1. Keep the case as a cheap check that a second action type behaves the same way, and drop the paragraph in the description saying it covers `applyOrderPromotions`.
2. Cover something no test reaches today. A `PromotionOrderAction` returning `0` with an `onActivate` defined must be kept. That is the contract the docs paragraph you edited describes, and no test asserts it for order actions.

Two smaller points:

- `createOrderPromotion` is `createItemPromotion` with the parameter type changed. `Promotion` already accepts `promotionActions?: Array<PromotionAction<any>>`, so widening the existing helper's parameter removes the need for a second one.
- The `id = 1` parameter on the new helper is never passed.

The docs fact is correct and worth keeping. `hasSideEffects` is `!!(onActivateFn || onDeactivateFn)`, so an action counts as having side effects when either hook is defined, whatever the body does. The paragraph could be shorter, since the sentence above it already states the exception.

The two red checks are unrelated to your change. Both fail because pull requests from forks run without repository secrets and with reduced write permissions.

brmk:
Thanks @gabriellbui, went with option 2. The Order-level zero-discount case is replaced by one asserting a `PromotionOrderAction` returning `0` with an `onActivate` is kept. The two helpers are merged into a single `createPromotion(action: PromotionAction<any>, id = 1)`, and the docs note is shortened. Description updated accordingly.

gabriellbui (approved):

Issue #5245, opened 2026-09-01
Problems with @vendure/create on Windows with Powershell


michaelbromley:
Follow the getting-started guide as a customer would and log every stall with a timestamp + where it happened

FINDINGS:
I installed Core following the getting-started guide, as a buyer would. It works, but three things stopped me before I saw a shop.

1. The install command fails on Windows. npx @vendure/create is blocked by PowerShell's default script policy on a clean Windows 11 machine. So does npm run dev, printed on your own success screen. The fix is npx.cmd / npm.cmd, and it isn't mentioned anywhere. Every Windows user hits this at step one
2. The Quick Start storefront doesn't boot on Windows. Windows Application Control blocks the Next.js native binary, and the bundler it defaults to can't run without it. Next.js prints the answer — add --webpack — but it scrolls past inside a long stream of successful server logs, so it reads as "everything started fine." I only found it by reading back through the output as when I tried to see it locally I got an error and went to investigate
3. Partial failure looks like success. The server, dashboard and APIs came up, so the terminal looked healthy while a third of what I installed was dead. Nothing at the end summarised what actually started.

None of this is broken product, it's a getting-started path written and tested on a Mac.
Worth deciding whether Windows is a supported first-run platform as also in the docs I dont see any mention of it

-----

Another thing I noticed is that the verification link, email, etc. points at the port 8080, more info below also on my install:

Version: Vendure 3.7.2, @vendure/create 3.7.2, Windows 11, Node 24.15.0, SQLite
Steps to reproduce
1.npx @vendure/create my-shop → Quick Start → include the storefront
2. npm run dev
3. Register a new customer account on the storefront at localhost:3001
4. Open the generated email in apps/server/static/email/test-emails (or the dev mailbox at /mailbox)

Expected: the verification link points at the storefront the installer just created.

Actual: it points at http://localhost:8080/verify?token=.... Nothing runs on 8080. The storefront runs on 3001 and does have a working /verify route — localhost:3001/verify returns 307, localhost:8080/verify refuses the connection.

apps/server/src/vendure-config.ts hardcodes localhost:8080 for verifyEmailAddressUrl, passwordResetUrl and changeEmailAddressUrl, with a comment saying it assumes a storefront on 8080 --> The same scaffold writes NEXT_PUBLIC_SITE_URL=http://localhost:3001/ into the storefront's own .env.local. The two halves of one generated project disagree about the port.

[...]

PR #5338, opened 2026-09-09
fix(create): use real storefront port in scaffolded vendure-config

wakqasahmed:
# Description

Part of #5245 — just the storefront-port paragraph, not the rest of that issue (which is a broader Windows getting-started report).

`vendure-config.hbs` hardcodes `http://localhost:8080` for `verifyEmailAddressUrl`, `passwordResetUrl` and `changeEmailAddressUrl`, even on `@vendure/create` scaffolds where a storefront is included. The actual storefront dev port is scanned at scaffold time (`findAvailablePort`, starting from 3001) and passed to the storefront's own `package.json`/env, but that value never made it into `generateSources()`'s Handlebars context, so the config template had no way to reference it.

This threads `storefrontPort` through `generateSources()` in `gather-user-responses.ts` and templates it into the three URLs. The port scan itself now happens inside `generateSources()` (once a storefront is selected) instead of afterward in `create-vendure-app.ts` — it was one scan before and still is, just moved earlier so the template can see the result. When no storefront is selected, it falls back to the `STOREFRONT_PORT` constant as a placeholder.

Added a test in `gather-user-responses.spec.ts` covering both cases (storefront selected → real scanned port in the config, no storefront → placeholder default), and ran the full `packages/create` test suite plus a `tsc --noEmit` typecheck.

# Breaking changes

None.

# Checklist

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5338"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791580997&installation_model_id=20193&pr_number=5338&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5338&signature=b2eaaa59e57b379913f5e0e06c526cccb74230b419965bd3468324716eceaacb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


packages/create/src/gather-user-responses.ts
```diff
@@ -278,11 +279,19 @@ async function generateSources(
...
     const assetPath = (fileName: string) => path.join(__dirname, '../assets', fileName);
 
     registerEscapeSingleHelper();
 
+    // Only scan for a real port when a storefront was actually selected - this is the
+    // same port that create-vendure-app.ts later hands to the storefront's own dev
+    // server, so the two must agree. When no storefront is selected there is no real
+    // port to point at, so STOREFRONT_PORT is just a placeholder default.
+    const storefrontPort = answers.storefront
+        ? await findAvailablePort(Math.max(STOREFRONT_PORT, port + 1), PORT_SCAN_RANGE)
```
wakqasahmed, line 291: The commit message says the old code "scanned twice, which could land on two different ports." Checked `origin/master` before this PR: `generateSources()` never scanned for a storefront port at all - the only `findAvailablePort` call for the storefront lived in `create-vendure-app.ts`, and it ran *after* `generateSources()` returned. So there was one scan, not two, and it's the one this diff removes below. The refactor itself is fine (moving the scan earlier so the template can see it is exactly the right fix), but the stated rationale for it doesn't match what the pre-fix code actually did - might be worth correcting the commit message/description so the git history doesn't record a bug that didn't exist.

packages/create/src/create-vendure-app.ts
```diff
@@ -224,23 +224,6 @@ export async function createVendureApp(
...
 
-    // Find an available storefront port if including storefront
-    if (includeStorefront) {
-        const storefrontPortSpinner = spinner();
-        storefrontPortSpinner.start(`Establishing storefront port...`);
-        try {
-            // Start scanning from the higher of STOREFRONT_PORT or serverPort + 1
-            // to avoid conflicts with the server port
-            const storefrontStartPort = Math.max(STOREFRONT_PORT, port + 1);
-            storefrontPort = await findAvailablePort(storefrontStartPort, PORT_SCAN_RANGE);
```
wakqasahmed, line 235: This deleted block's try/catch gave a clean failure path if the port scan ran out of range: stop the spinner with a red message, `outro(e.message)`, `process.exit(1)`. Now that the same `findAvailablePort` call lives inside `generateSources()` (gather-user-responses.ts), there's no catch around it there - if it throws, it'll propagate up and only get caught by the generic top-level `.catch(err => { log(err); process.exit(1); })` in create-vendure-app.ts's entrypoint. Still exits 1, so not a crash, but it trades the polished spinner/outro message for a raw logged error. Given `PORT_SCAN_RANGE` makes this a narrow window to actually hit, probably not worth blocking on, but flagging since it is a real (if minor) UX regression from this refactor.

packages/create/src/gather-user-responses.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+
+describe('getCiConfiguration', () => {
+    beforeAll(() => {
+        // Normally registered once in create-vendure-app.ts before any config is generated.
+        registerTemplateHelpers(getPackageManagerInfo('npm'));
+    });
+
+    afterEach(() => {
+        vi.clearAllMocks();
```
wakqasahmed: `afterEach` calls `vi.clearAllMocks()`, which clears call history but not the `mockResolvedValue` set on `findAvailablePortMock` in the first test. Doesn't affect these two tests today - the second one asserts the mock is never called at all - but if a third test relies on the mock having no canned resolution, or the two tests get reordered, the leftover `mockResolvedValue(4321)` from test 1 would leak in silently. `vi.resetAllMocks()` (or `findAvailablePortMock.mockReset()`) would be more robust here.

wakqasahmed (commented):
Went through this cold against #5245's port paragraph. The actual fix is correct and the tests are real (mocked `findAvailablePort` via `./helpers`, and they'd fail pre-fix since the old template always contained `localhost:8080`). Two things worth a look before merge, both inline. Also independently verified the lint-bypass justification in the commit message: checked out `origin/master` at 818327d1c (this PR's own base) into a clean worktree, ran `bun install`, then ran eslint on both a file this PR touches (`gather-user-responses.ts`) and one it doesn't (`packages/core/.../order.service.ts`) - both crash identically with `TypeError: ...getTypeAtLocation is not a function`, caused by `@typescript-eslint/eslint-plugin@8.69.0` paired with `@typescript-eslint/parser@5.62.0` (peer dep wants `^8.69.0`). That's a real, pre-existing, repo-wide breakage from the same-day dependabot bump, not something this PR caused - the `--no-verify` is justified.

packages/create/src/create-vendure-app.ts
wakqasahmed: Fixed — wrapped the config-generation call in its own spinner/try-catch so a storefront port scan failure (which now happens inside generateSources) gets the same polished outro+exit treatment the old dedicated try/catch gave it, instead of falling through to the generic top-level handler.

packages/create/src/gather-user-responses.ts
wakqasahmed: You're right, thanks for checking — pre-fix there was only one storefront port scan (in create-vendure-app.ts, after generateSources returned), not two. I've corrected the description below; the actual fix (move the scan earlier so the template can see it) doesn't depend on that claim being true either way.

packages/create/src/gather-user-responses.spec.ts
wakqasahmed: Fixed — afterEach now uses vi.resetAllMocks() instead of vi.clearAllMocks().

biggamesmallworld:
Merge lane session `88abe85f` (@biggamesmallworld) is working this pull request.

Current stage: 1 (isolate). Claim expires 2026-09-10T15:37:31Z unless refreshed.

biggamesmallworld:

extra review round trip.

What we changed:

- Removed the spinner around the configuration prompts, because the spinner erased each prompt as it rendered and stopped Ctrl+C cancelling.
- Made the unit tests generate the template assets they read, so the tests pass on a clean checkout.
- Added a comment to the generated config saying the port is a placeholder when no storefront was created.
- Added a regression test covering the scaffolded config and the storefront agreeing on one port.

biggamesmallworld (approved):
Reproduced the wrong port on base with an own test, verified the fix end to end (scan landed on 3003; config, storefront package.json and .env.local all agree), and confirmed 71 unit tests pass on a clean tree. Fixed the spinner that was erasing the interactive prompts and the missing template assets that broke CI.

PR #5339, opened 2026-09-09
fix(dashboard): restore display component precedence over core cell fn

### LINKED ISSUE #5201: `displayComponents` registry is bypassed for any column where the core supplies a `cell` function
> [!NOTE]
> **Authorship:** This report was researched and written by Claude AI Fable 5 working under my direction, and reviewed by me before posting. Every source reference below was verified against current `master` before posting.

**Describe the bug**

The dashboard extension API documents `displayComponents` as the way to control how a data table cell is rendered. It works only for columns that the core does not already customise. In `use-generated-columns.tsx`, a column's cell renderer short-circuits to the core-supplied `cell` function whenever `customizeColumns` provides one, and `CellWrapper` — the only place a registered `pageId_blockId_field` key is ever looked up — is then never rendered. Registration fails silently: `addDisplayComponent` writes into a Map without validating that the key targets a column that will ever consult the registry.

This appears to be a regression rather than the original design: commit 2e9f8629 ("fix(dashboard): Fix data table component overrides", 2025-10-14) checked the registry first and fell back to the core `cell` function. #4064 (merged 2025-12-18) replaced that with the current short-circuit while fixing an unrelated facet-values issue.

The practical consequence is that the columns most likely to need customising — the money columns — are exactly the ones that cannot be customised.

(This is related to, but distinct from, #4642: that issue is about a *new* custom-field column never appearing at all; this one is about an *existing* column rendering while the registered display component is silently ignored. Likely the same underlying area — happy for these to be linked or merged if you see one root cause.)

**To Reproduce**

The documented example from [Customizing List Pages](https://docs.vendure.io/guides/extending-the-dashboard/customizing-pages/customizing-list-pages/) works, and changing one string breaks it.

1. Register a display component exactly as documented:

   ```ts
   defineDashboardExtension({
       dataTables: [{
           pageId: 'product-detail',
           blockId: 'product-variants-table',
           displayComponents: [{ column: 'sku', component: MyCustomCell }],
       }],
   });
   ```

   `MyCustomCell` renders. `sku` happens to have no `cell` entry in that table's `customizeColumns`.

2. Change `column: 'sku'` to `column: 'price'` — same page, same block, same table.

**Expected behavior**

`MyCustomCell` renders in the price column, or a warning explains why it cannot.

**Actual behavior**

Nothing happens. No error, no warning. `price` and `priceWithTax` have hardcoded `cell` functions in `customizeColumns`, so `CellWrapper` is never rendered for them and the registry is never consulted.

Note that both examples in the documentation land on working columns only by accident: the guide's first example targets `sku` (the one column on `product-variants-table` without a core `cell`), and its second targets `slug` on `product-list` (which overrides only `name` and `description`). The same guide opens with "you can customize any existing data table" and recommends `displayComponents` as the most direct approach for existing tables. A user copying the documented example and pointing it at `name`, `price` or `priceWithTax` on the very same table gets silence.

**Source references**

All verified on `master` @ 4631092 (2026-08-27); byte-identical in v3.7.2. The same logic is present on the `minor` branch, though line numbers there differ slightly.

`packages/dashboard/src/lib/components/data-table/use-generated-columns.tsx` — the short-circuit (lines 142–155):

```tsx
// If a custom cell function is provided, use it directly (like additionalColumns does).
// This preserves the same behavior and prevents cell unmounting issues.
// Only use CellWrapper for columns without custom cells.
const cellFn =
    typeof customCell === 'function'
        ? customCell
        : (cellContext: CellContext<any, any>) => (
              <CellWrapper
                  cellContext={cellContext}
                  fieldInfo={fieldInfo}
                  isCustomField={isCustomField}
                  displayComponentId={displayComponentId}
              />
          );
```

Within this file, `getDisplayComponent(displayComponentId)` is called only inside `CellWrapper` (line 352), and `displayComponentId` (computed at lines 137–140) is passed nowhere else.

`packages/dashboard/src/lib/framework/extension-api/display-component-extensions.tsx` — `addDisplayComponent` (lines 37–58) warns on key collision but performs no other validation, so a key that can never be reached is indistinguishable from a typo.

`packages/dashboard/src/app/routes/_authenticated/_products/components/product-variants-table.tsx` — `customizeColumns` (lines 79–112) supplies `cell` for `name`, `currencyCode`, `price`, `priceWithTax` and `stockLevels`, but not for `sku`. This is why the documented example works.

**Secondary consequence: order views**

The same mechanism makes the order views' price display fixed. `MoneyGrossNet` (`packages/dashboard/src/app/routes/_authenticated/_orders/components/money-gross-net.tsx`, lines 11–18) renders gross above net. Gross-first is a reasonable default and I am not proposing it be changed. The issue is that there is no supported way to express the opposite preference in a B2B context, because:

- there is no setting (`UserSettings` has no price display field; `pricesIncludeTax` is a channel setting governing input and storage, not display);
- the display component registry is bypassed for those columns, per the above;
- `edit-order-table.tsx` (draft order and order modification views) builds its `ColumnDef[]` by hand and passes them straight to `useReactTable`, never going through `useGeneratedColumns` at all;
- `MoneyGrossNet` cannot be reused from an extension — it lives under `src/app/routes/`, outside the `src/lib` barrel. Notably, the barrel *does* already export comparable table cells (`order-table-cell-components.js`: `CustomerCell`, `OrderStateCell`, `OrderMoneyCell`, …), so the precedent for exporting core cells exists; this component simply sits on the wrong side of the boundary.

I am aware of the `pageBlocks` escape hatch — `position: { blockId: 'order-table', order: 'replace' }` does work. It requires reimplementing the whole table, including custom field columns, view options and the totals rows, without being able to reuse `MoneyGrossNet`. That is workable but disproportionate to "show net first".

**Workaround**

What we ended up doing instead, running in production since mid-August: a pnpm patch that edits `money-gross-net.tsx` directly. It works, but it forks a core file and has to be re-applied on every Vendure upgrade — which is the situation the extension API exists to avoid.

**Proposed fix**

In rough order of cost:

1. **Make the failure loud.** Warn when a registered display component key resolves to a column whose cell is core-supplied, so a silently ignored registration is distinguishable from a typo.
2. **Let the registry win.** Restore the pre-#4064 precedence (registry first, core `cell` as fallback), while keeping the memoised `CellWrapper` structure so the cell-unmounting fix from #4064 is not undone.
3. **Export `MoneyGrossNet`** (and comparable display primitives) from the public entry point, so extensions replacing a block can reuse them instead of forking them.
4. Optionally, a `displayPricesWithTax`-style user or channel preference for the admin order views — but (1) and (2) alone would be enough for extensions to solve this themselves.

**Additional context**

#5159 (order detail query extension silently ignored) and discussion #4332 describe other cases where a dashboard extension registration is accepted and then silently does nothing. This report supplies the concrete mechanism for the `displayComponents` case, which may be useful for the systemic view already labelled on #4642.

- @vendure/core version: 3.7.2 (`@vendure/dashboard@3.7.2`; also verified on `master` @ 4631092, 2026-08-27, and on the `minor` branch)
- Nodejs version: 24.x
- Database (mysql/postgres etc): postgres (not relevant to this issue)
- Operating System (Windows/macOS/Linux): Windows/Linux
- Browser (if applicable): any

wakqasahmed:
# Description

Fixes #5201.

`useGeneratedColumns` decides how to render each column's cell like this: if `customizeColumns` supplies a `cell` function, use it directly; otherwise render through `CellWrapper`, which is the only place that looks up the `displayComponents` registry (`addDisplayComponent`/`generateDisplayComponentKey`). That means any column where the core already supplies a `cell` — money columns being the obvious example — can never be overridden through the registry. Registering a display component for that `pageId_blockId_field` key silently does nothing.

This was a regression from #4064. Before that PR (since #2e9f8629), the cell renderer checked the registry first and only fell back to the custom cell function if nothing was registered. #4064 replaced that with today's straight short-circuit while fixing an unrelated facet-values re-render/unmounting issue.

The fix restores the registry-first precedence: a column now checks whether a display component is registered for its key, and only bypasses `CellWrapper` for the direct-custom-cell-function path when nothing is registered there. When nothing's registered, behavior is unchanged from today (still bypasses `CellWrapper`, so #4064's fix stays intact).

Added a test alongside `use-generated-columns.tsx` covering both cases: a registered display component wins over a core-supplied cell, and the core cell still renders when nothing is registered. Confirmed the new test fails against the current code and passes with the fix.

# Breaking changes

None. This restores prior, documented precedence for the `displayComponents` registry; a column with a registered display component now shows it instead of silently ignoring it.

# Screenshots

N/A — data-table cell rendering logic, covered by the added unit test.

# Checklist

📌 Always:

👍 Most of the time:
- [ ] I have updated the README if needed (not applicable — no docs describe this precedence)

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5339"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791583864&installation_model_id=20193&pr_number=5339&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5339&signature=5cea329c4349b62a52e713b47c0c7d28e406fbb28585f76fa37ade3fd8e3179b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/data-table/use-generated-columns.spec.tsx
+++ b/packages/dashboard/src/lib/components/data-table/use-generated-columns.spec.tsx
@@ -1,38 +1,51 @@
+import {
+    defineDashboardExtension,
+    executeDashboardExtensionCallbacks,
+} from '@/vdb/framework/extension-api/define-dashboard-extension.js';
 import { addDisplayComponent } from '@/vdb/framework/extension-api/display-component-extensions.js';
 import { PageBlockContext } from '@/vdb/framework/layout-engine/page-block-provider.js';
 import { PageContext } from '@/vdb/framework/layout-engine/page-provider.js';
-import { CellContext } from '@tanstack/react-table';
+import { CellContext, flexRender } from '@tanstack/react-table';
 import { renderToStaticMarkup } from 'react-dom/server';
 import { describe, expect, it } from 'vitest';
 
 import { useGeneratedColumns } from './use-generated-columns.js';
 
-const fakeCellContext = {
-    cell: { getValue: () => undefined },
-    row: { original: {} },
+// The display component registry is a module-level Map with no removal API, so an entry
+// registered by one test stays visible to every later test. Each test therefore uses its
+// own pageId; reusing one would let an earlier registration decide a later result.
+const BLOCK_ID = 'test-block';
+
+const PRICE = 1234;
+
+const fields = [
+    { name: 'sku', type: 'String', nullable: false, list: false, isPaginatedList: false, isScalar: true },
+    { name: 'price', type: 'Int', nullable: false, list: false, isPaginatedList: false, isScalar: true },
+];
+
+// Mirrors product-variants-table.tsx, which supplies a `cell` for `price` but not for `sku`.
+// That asymmetry is why the documented example works on `sku` and silently fails on `price`.
+const customizeColumns = {
+    price: { cell: () => <span>core-money-cell</span> },
+} as any;
+
+const cellContext = {
+    cell: { getValue: () => PRICE },
+    row: { original: { sku: 'SKU-1', price: PRICE } },
 } as unknown as CellContext<any, any>;
 
-function renderColumnCell(pageId: string, blockId: string, columnId: string, coreCellLabel: string) {
+/**
+ * Renders one generated column's cell the way TanStack renders it. `flexRender` calls
+ * `createElement` for any function cell, so a cell using hooks behaves here as it does in
+ * a real table; calling `column.cell(context)` directly would not.
+ */
+function renderColumnCell(pageId: string, columnId: string): string {
     const captured: { columns?: Array<{ id?: string; cell?: any }> } = {};
 
     function Harness() {
         const { columns } = useGeneratedColumns({
-            fields: [
-                {
-                    name: columnId,
-                    type: 'Int',
-                    nullable: false,
-                    list: false,
-                    isPaginatedList: false,
-                    isScalar: true,
-                },
-            ],
-            customizeColumns: {
-                [columnId]: {
-                    // Simulates a core-supplied `cell` function, e.g. the Money cell on price columns.
-                    cell: () => <span>{coreCellLabel}</span>,
-                },
-            } as any,
+            fields,
+            customizeColumns,
             includeSelectionColumn: false,
             includeActionsColumn: false,
         });
@@ -42,7 +55,7 @@ function renderColumnCell(pageId: string, blockId: string, columnId: string, cor
 
     renderToStaticMarkup(
         <PageContext.Provider value={{ pageId }}>
-            <PageBlockContext.Provider value={{ blockId, column: 'main' }}>
+            <PageBlockContext.Provider value={{ blockId: BLOCK_ID, column: 'main' }}>
                 <Harness />
             </PageBlockContext.Provider>
         </PageContext.Provider>,
@@ -52,35 +65,73 @@ function renderColumnCell(pageId: string, blockId: string, columnId: string, cor
     if (!column?.cell) {
         throw new Error(`Column "${columnId}" was not generated`);
     }
-    return renderToStaticMarkup(column.cell(fakeCellContext));
+    return renderToStaticMarkup(<>{flexRender(column.cell, cellContext)}</>);
 }
 
-describe('useGeneratedColumns', () => {
+describe('useGeneratedColumns display component precedence', () => {
     it('gives precedence to a registered display component over a core-supplied cell function', () => {
         const pageId = 'test-page-registered';
-        const blockId = 'test-block';
-        const columnId = 'price';
 
         addDisplayComponent({
             pageId,
-            blockId,
-            field: columnId,
+            blockId: BLOCK_ID,
+            field: 'price',
             component: () => <span>registered-display-component</span>,
         });
 
-        const html = renderColumnCell(pageId, blockId, columnId, 'core-cell');
+        expect(renderColumnCell(pageId, 'price')).toBe('<span>registered-display-component</span>');
+    });
 
-        expect(html).toContain('registered-display-component');
-        expect(html).not.toContain('core-cell');
+    it('passes the cell value to a display component that overrides a core-supplied cell', () => {
+        const pageId = 'test-page-value';
+
+        addDisplayComponent({
+            pageId,
+            blockId: BLOCK_ID,
+            field: 'price',
+            component: ({ value }) => <span>{`value:${value as number}`}</span>,
+        });
+
+        expect(renderColumnCell(pageId, 'price')).toBe(`<span>value:${PRICE}</span>`);
     });
 
     it('falls back to the core-supplied cell function when no display component is registered', () => {
-        const pageId = 'test-page-unregistered';
-        const blockId = 'test-block';
-        const columnId = 'price';
+        expect(renderColumnCell('test-page-unregistered', 'price')).toBe('<span>core-money-cell</span>');
+    });
 
-        const html = renderColumnCell(pageId, blockId, columnId, 'core-cell');
+    it('still renders a registered display component on a column with no core-supplied cell', () => {
+        const pageId = 'test-page-no-core-cell';
 
-        expect(html).toContain('core-cell');
+        addDisplayComponent({
+            pageId,
+            blockId: BLOCK_ID,
+            field: 'sku',
+            component: () => <span>sku-display-component</span>,
+        });
+
+        expect(renderColumnCell(pageId, 'sku')).toBe('<span>sku-display-component</span>');
+    });
+
+    it('renders the default display component when a column has neither a core cell nor a registration', () => {
+        expect(renderColumnCell('test-page-default', 'sku')).toBe(String(PRICE));
+    });
+
+    it('applies a display component registered through defineDashboardExtension', () => {
+        // The route real extensions take. registerDataTableExtensions defaults blockId to
+        // 'list-table', so the block has to be named explicitly to target this table.
+        const pageId = 'test-page-extension-api';
+
+        defineDashboardExtension({
+            dataTables: [
+                {
+                    pageId,
+                    blockId: BLOCK_ID,
+                    displayComponents: [{ column: 'price', component: () => <span>via-extension-api</span> }],
+                },
+            ],
+        });
+        executeDashboardExtensionCallbacks();
+
+        expect(renderColumnCell(pageId, 'price')).toBe('<span>via-extension-api</span>');
     });
 });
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/lib/components/data-table/use-generated-columns.tsx
```diff
@@ -138,12 +138,20 @@ export function useGeneratedColumns<T extends TypedDocumentNode<any, any>>({
                 pageId && pageBlock?.blockId
                     ? generateDisplayComponentKey(pageId, pageBlock.blockId, fieldInfo.name)
                     : undefined;
+            // A component registered via addDisplayComponent() must take precedence over a
+            // core-supplied `cell` function (e.g. the Money cell on price columns), otherwise
+            // registering an override for such a column silently does nothing.
+            const hasRegisteredDisplayComponent = !!(
```
wakqasahmed: Small note: this hoists the registry check to column-build time (inside the outer `useMemo`), whereas before this PR `getDisplayComponent` was only ever called inside `CellWrapper` at cell-render time - i.e. re-checked on every render. That's fine given `registerDataTableExtensions`/`addDisplayComponent` run synchronously during `executeDashboardExtensionCallbacks`, before any table mounts. But if a display component is ever registered *after* this memo has already run once for a given table (e.g. a lazily-loaded extension bundle), `hasRegisteredDisplayComponent` will be stuck `false` for that memo's lifetime since it's not in the `useMemo` dependency array (which also doesn't include `pageId`/`pageBlock`, pre-existing). Not asking for a change here, just flagging the assumption this relies on.

packages/dashboard/src/lib/components/data-table/use-generated-columns.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+        expect(html).not.toContain('core-cell');
+    });
+
+    it('falls back to the core-supplied cell function when no display component is registered', () => {
+        const pageId = 'test-page-unregistered';
+        const blockId = 'test-block';
+        const columnId = 'price';
+
+        const html = renderColumnCell(pageId, blockId, columnId, 'core-cell');
+
```
wakqasahmed, line 134: This second test pins down the unregistered fallback path (the one #4064 actually touched), which is good coverage for the branch-selection logic. Worth noting it passes against both pre-fix and post-fix code, so it's a regression guard for future changes to this file rather than proof of the #4064 fix itself - the original Sheet re-render/unmounting bug lived in a different rendering concern (table-wide re-render from state held in a cell) that a single `renderToStaticMarkup` pass can't exercise. No change requested, just calling out what this test does and doesn't cover for anyone revisiting this later.

wakqasahmed (commented):
Traced this against #4064 and the pre-#4064 commit (2e9f8629) mentioned in the issue. The bug report holds up exactly as described: on master, `getDisplayComponent` is only ever called inside `CellWrapper`, and `CellWrapper` is skipped entirely whenever `customizeColumns` supplies a `cell` fn, so a registered display component for a column like `price` is never consulted. #4064's actual fix (the Sheet state re-render issue) lived in facet-values-sheet.tsx/facets.tsx via context lifting, not in this short-circuit - so restoring registry-first precedence here doesn't touch what #4064 fixed. The no-registration path is byte-identical to today (customCell still bypasses CellWrapper directly), so the facet-values case stays exactly as #4064 left it.

Tests are meaningful: with the old code, `typeof customCell === 'function'` alone decides the branch, so test 1 renders `core-cell` and fails the `registered-display-component` assertion. With the fix it passes. Test 2 pins down the unregistered fallback, which is the #4064-relevant path, though it passes on both versions since it's not a regression test for the original re-render bug itself (that needs an actual multiple-render harness, which isn't here - not a blocker, just noting the boundary of what's covered).

On the 'does bypassing lose something essential' question: registered display components already fully replace default rendering for every column with no core `cell` (that's the existing contract), so extending that same replace-not-compose behavior to columns that do have a core `cell` isn't a new class of risk - it's consistent with how the registry already works everywhere else. `product-variants-table.tsx`'s `price`/`priceWithTax` cells are pure presentation (`<Money .../>`, no handlers), so no interactive behavior is lost there. `name`'s `DetailPageButton` link would be fully replaced if someone registered a display component for that key, but the component still receives `{...cellContext}` so `row.original.id` etc. are available to rebuild it - same trade-off as registering over any other column today.

One thing worth a second look (non-blocking): `hasRegisteredDisplayComponent` is now computed once per column inside the outer `useMemo`, whereas the registry lookup used to happen at cell-render time inside `CellWrapper` on every render. The `useMemo` deps (`fields, customizeColumns, rowActions, deleteMutation, additionalColumns, defaultColumnOrder`) already didn't include `pageId`/`pageBlock` before this PR, so that's pre-existing, but this PR adds a *new* dependency on registry state at build time that also isn't a dep. In the normal bootstrap flow (`executeDashboardExtensionCallbacks` runs synchronously before routes render) this is moot, but if any dashboard extension bundle ever registers a display component after a table's columns have already been memoized (dynamic import, lazy plugin load, etc.), the column would be permanently baked into the `customCell`-direct branch for that memo's lifetime, with no way for a later registration to take effect. Might be worth a comment noting the assumption, or moving the check inside `cellFn`'s render-time branch instead of hoisting it - not asking for a change, just flagging since it's the one place this PR's precedence decision is less dynamic than the code it replaces.

michaelbromley:

michaelbromley:

extra review round trip.

What we changed:

- Added the page and block ids to the generated columns memo dependencies, so a column built for one page cannot be reused on another.
- Extended the tests to cover the cell value reaching a registered display component, and registration through the public extension API.
- Documented that a display component replaces a column's built-in cell renderer, which nothing previously stated.

Verification

On master:

Product variants table where the Price column shows the built-in money cell, ignoring a registered display component

With the fix:

The same table where the Price column shows the registered display component, while Price with tax is unchanged

michaelbromley (approved):
Reproduced #5201 on origin/master (3 failing assertions in use-generated-columns.spec.tsx), confirmed the fix passes all 6, verified #4064's direct-call path still holds, and checked the before/after rendering of the Price column. Full CI green with no build or test job skipped.

PR #5349, opened 2026-09-10
fix(core): treat localeText custom fields as localized everywhere

### LINKED ISSUE #5328: localeText custom fields are missed where only localeString is checked
`localeText` custom fields are handled as a special case in several places in core, and two of those places only check for `localeString`. PR #5265 fixes one of them. This issue covers the other, and the underlying reason the pair keeps being missed.

## The remaining instance

`packages/core/src/data-import/providers/import-parser/import-parser.ts:319` decides whether a CSV column is translatable:

```ts
return customFieldConfig.type === 'localeString';
```

A `localeText` custom field is therefore not treated as translatable on import, so its value is not read from the translation columns of the CSV.

The same shape was just fixed in `parse-sort-params.ts`, where sorting by a `localeText` custom field built a column path with no `customFields` segment and threw `TypeError: Cannot read properties of undefined (reading 'databaseName')` (#5199).

## Why it keeps happening

The `localeString || localeText` pair is spelled out separately at each site that needs it:

- `entity/register-custom-entity-fields.ts:109`, `:240` — correct
- `api/config/graphql-custom-fields.ts:88` — correct
- `service/helpers/list-query-builder/parse-sort-params.ts` — fixed by #5265
- `data-import/providers/import-parser/import-parser.ts:319` — still wrong

Adding a third localized custom field type would mean finding all of these again.

Two options:

1. Export a predicate, for example `isLocalizedCustomFieldType(type)`, next to `CustomFieldType` in `config/custom-field/custom-field-types.ts`, and use it at every site.
2. For `parse-sort-params.ts` specifically, stop rebuilding the path from the custom field config and read `propertyPath` off the translation column metadata, the way the non-translated branch at line 44 already does. That removes the type check from this function entirely and is correct for any future localized type.

## Also worth checking

`config/custom-field/validate-custom-fields-config.ts:76` checks only for `localeString`. Whether that is deliberate has not been established.

Found while reviewing #5265.

Ismatulla:
`localeText` custom fields were missed at two sites which only checked for `localeString`:

- `ImportParser.isTranslatable()` returned false for a `localeText` custom field, so a CSV using its translation columns was rejected with "The '<field>' field is not translatable."
- `validateCustomFieldsConfig()` did not reject `localeText` custom fields configured on non-localized entities.

To stop the `localeString`/`localeText` pair from drifting apart again, this adds an exported `isLocalizedCustomFieldType()` predicate alongside the custom field types and uses it at every site that needs to distinguish localized custom fields.

The config validation error now names the offending type, so the existing `localeString` message is unchanged.

Fixes #5328

# Description

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5349"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791639039&installation_model_id=20193&pr_number=5349&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5349&signature=0109cdd464b186b81a53affc592b41652f051ca4da636090a03676dae7f4320c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/core/src/api/config/graphql-custom-fields.ts
+++ b/packages/core/src/api/config/graphql-custom-fields.ts
@@ -13,6 +13,7 @@ import {
     BaseTypedCustomFieldConfig,
     CustomFieldConfig,
     CustomFields,
+    isLocalizedCustomFieldType,
     StructCustomFieldConfig,
     StructFieldConfig,
 } from '../../config/custom-field/custom-field-types';
@@ -84,11 +85,9 @@ export function addGraphQLCustomFields(
             }
         }
 
-        const localizedFields = customEntityFields.filter(
-            field => field.type === 'localeString' || field.type === 'localeText',
-        );
+        const localizedFields = customEntityFields.filter(field => isLocalizedCustomFieldType(field.type));
         const nonLocalizedFields = customEntityFields.filter(
-            field => field.type !== 'localeString' && field.type !== 'localeText',
+            field => !isLocalizedCustomFieldType(field.type),
         );
         const writeableLocalizedFields = localizedFields.filter(field => !field.readonly);
         const writeableNonLocalizedFields = nonLocalizedFields.filter(field => !field.readonly);
--- a/packages/core/src/config/custom-field/custom-field-types.ts
+++ b/packages/core/src/config/custom-field/custom-field-types.ts
@@ -33,6 +33,22 @@ import { RequestContext } from '../../api/common/request-context';
 import { Injector } from '../../common/injector';
 import { VendureEntity } from '../../entity/base/base.entity';
 
+/**
+ * @description
+ * Returns `true` if the given custom field type is a localized type, i.e. one whose values are
+ * stored on the entity's translation entity rather than on the entity itself.
+ *
+ * Prefer this over comparing against `'localeString'` and `'localeText'` individually, so that
+ * every site which needs to know about localized custom fields stays correct if further
+ * localized types are added.
+ *
+ * @docsCategory custom-fields
+ * @since 3.7.4
+ */
+export function isLocalizedCustomFieldType(type: CustomFieldType | StructFieldType | undefined): boolean {
+    return type === 'localeString' || type === 'localeText';
+}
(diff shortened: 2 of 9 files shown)


Ismatulla:

Ismatulla:
ismatilloqoyliyev@gmail.com

Ismatulla:
recheck

gabriellbui:
Merge lane session `12056d64` (@gabriellbui) is working this pull request.

Current stage: 6 fix forward. Claim expires 2026-09-10T16:40:22Z unless refreshed.

gabriellbui:
Thanks for this, @Ismatulla. We're going to merge it.

extra review round trip.

What we changed:

- Narrowed the `isLocalizedCustomFieldType()` parameter to `CustomFieldType`, because struct field types can never be localized.
- Shortened the `isLocalizedCustomFieldType()` doc comment to the sentence that says what it returns, because `@docsCategory` publishes the comment to the API reference.

gabriellbui (approved):
Reproduced both #5328 defects on master with an independent test, verified the fix; core unit 1267 and e2e custom-fields/import specs pass locally, CI green on all build/unit/e2e jobs.

PR #5358, opened 2026-09-11
feat(cli): group the CLI help output by the package that provided each command

michaelbromley:
# Description

When a plugin extends the Vendure CLI, its commands and options are currently mixed in with the built-in ones in `vendure --help`, so there is no way to tell where anything came from. With `@vendure/cloud` installed that is a flat list of 33 commands. This came out of a Slack discussion about whether Cloud commands should be namespaced under `vendure cloud ...`; grouping the help gives the same clarity without making anyone type an extra word.

The help now groups both commands and shared options by the package that registered them:

```
Options:
  -V, --version                             output the version number
  -h, --help                                display help for command

Options from @vendure/cloud:
  --token <token>                           Cloud API token
  --project <slug>                          Cloud project slug

Commands:
  add [options]                             Add a feature to your Vendure project
  ...

Commands from @vendure/cloud:
  deploy [options]                          Deploy the current environment
  ...
```

Nothing about how a command is invoked changes. `vendure deploy` is still `vendure deploy`.

The heading uses the package name exactly as installed, so it names the thing to enable or remove. A command reached through `extendCommands` rather than registered outright stays under the heading it already had, because the command is still the one that declared it.

## Colour

Section headings are now bold, and a plugin heading's package name is also cyan, so a package's "Commands from" and "Options from" sections pair up visually. Commander applies this through its `styleTitle` hook, which covers every heading — `Usage:` and `Arguments:` are now bold too.

Colour is decoration only. The heading names the package in words, so a monochrome terminal, a colourblind reader or a piped log loses nothing.

Whether colour is emitted follows the usual rules, and both directions are covered by tests:

- No colour when stdout is not a terminal, or when `NO_COLOR` is set.
- Colour when `FORCE_COLOR` is set, **including when stdout is not a terminal**. This is worth knowing when writing tests: Vitest exports `FORCE_COLOR` to child processes when its own output is coloured, which it is when the suite runs through Lerna. `runCliCommand` in the CLI e2e helpers therefore returns output with colour stripped, and exposes `rawStdout` for a test that is about colour itself.

## `vendure plugins`

The listing now shows the commands each package provides, read from the plugin itself for an enabled one and from `vendure.cliCommands` for one that is only installed:

```
@vendure/cloud	enabled
	commands: project, config, backup, deploy, restore
```

## The commander upgrade, and the bug it exposed

Grouping needs Commander's `helpGroup()`, added in v14. `packages/cli` was still on `^11.0.0`, set in September 2023 and never touched since; `packages/dev-server` is on `^12` and `packages/create` on `^14`, so this was drift rather than a deliberate pin.

The upgrade turned up a latent bug. Commander 13 began throwing on duplicate option flags. `vendure add` declared `--selected-plugin` under both `-e` and `-s`, and `--selected-service` under both `-j` and `-a`, because every sub-option is flattened onto the same command. Commander 11 accepted the repeat, matched the first when parsing, and printed the flag twice in help. Under Commander 14 it throws while registering a built-in, so **every** `vendure` invocation would have failed, not just `add`.

The fix is in the `add` definition, which now declares each shared sub-option once with a description naming both parents. Nothing enforced this, which is why it was there in the first place: a plugin is validated by `assertCliPlugin` when it loads, so a duplicate flag is rejected by name, but the built-in definitions reach Commander unchecked. `builtins.spec.ts` now runs them through the same validation, so the built-ins are held to the rule plugins already were. Checked against the base commit, that test fails with `CLI plugin "builtins" declares "--selected-plugin" twice in options of command "add"`.

The first commit does this ahead of the upgrade, so no commit in this PR leaves the CLI broken.

## Known limitation

Only the root help is grouped. A subcommand's help shows shared options under `Global Options:`, which Commander builds as one flat list regardless of group, so it still mixes the CLI's own options with a plugin's. Grouping that would mean overriding `Help#formatHelp` and owning a copy of Commander's layout, which did not seem worth it — the root help is where the problem actually is.

# Breaking changes

None for users or plugin authors. `CliPlugin`, `CliCommandDefinition` and the rest of the public API are unchanged, and no command changes how it is invoked.

Internal signatures changed, none of them exported from the package:

- `CommandRegistry.toArray()` is now `getCommandTree()` and returns `{ node, source }` per command; `getRootOptions()` returns `{ option, source }`. They are returned paired rather than alongside a separate name-keyed map of sources, so the two cannot disagree about which commands exist.
- `registerCommands()` takes an options object instead of four positional arguments.
- `runCliCommand()` in the CLI e2e helpers returns colour-stripped `stdout` and `stderr`, plus `rawStdout` and `rawStderr`, and accepts an `env` option.

# Screenshots

See the help output above.

# Checklist

Always:

Most of the time:

## Testing

620 unit tests across 39 files and 88 e2e tests across 7 files, all passing, on Node 20, 22 and 24 and against sqljs, mysql, mariadb and postgres.

The e2e suite was verified both ways round — in the package directly and through Lerna as CI runs it — because only the second sets `FORCE_COLOR` and so exercises the coloured output.

Each commit was verified on its own, including the first on Commander 11 to confirm the option fix stands alone. The grouping and styling tests were mutation-checked: disabling each behaviour fails exactly the tests asserting it and leaves the negative controls passing.

`docs/docs/guides/developer-guide/cli/index.mdx` documents the grouping, and `packages/cli/src/README.md` covers the internals.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -195,10 +195,10 @@
       },
       "dependencies": {
         "@clack/prompts": "^0.7.0",
-        "@vendure/common": "3.7.0",
+        "@vendure/common": "3.7.2",
         "change-case": "^4.1.2",
         "chokidar": "^3.6.0",
-        "commander": "^11.0.0",
+        "commander": "^14.0.1",
         "cross-spawn": "^7.0.6",
         "dotenv": "^16.4.5",
         "fs-extra": "^11.2.0",
@@ -211,8 +211,8 @@
         "vite": "^7.3.1",
       },
       "devDependencies": {
-        "@vendure/core": "3.7.0",
-        "@vendure/testing": "3.7.0",
+        "@vendure/core": "3.7.2",
+        "@vendure/testing": "3.7.2",
         "cross-env": "^7.0.3",
       },
     },
@@ -3098,7 +3098,7 @@
 
     "comma-separated-tokens": ["comma-separated-tokens@2.0.3", "", {}, "sha512-Fu4hJdvzeylCfQPp9SGWidpzrMs7tTrlu6Vb8XGaRGck8QSNZJJp538Wrb60Lax4fPwR64ViY468OIUTbRlGZg=="],
 
-    "commander": ["commander@11.1.0", "", {}, "sha512-yPVavfyCcRhmorC7rWlkHn15b4wDVgVmBA7kV4QVBsF7kv/9TKJAbAXVTxvTnwP8HHKjRCJDClKbciiYS7p0DQ=="],
+    "commander": ["commander@14.0.3", "", {}, "sha512-H+y0Jo/T1RZ9qPP4Eh1pkcQcLRglraJaSLoyOtHxu6AapkjWVCy2Sit1QQ4x3Dng8qDlSsZEet7g5Pq06MvTgw=="],
 
     "comment-parser": ["comment-parser@1.3.1", "", {}, "sha512-B52sN2VNghyq5ofvUsqZjmk6YkihBX5vMSChmSK9v4ShjKf3Vk5Xcmgpw4o+iIgtrnM/u5FiMpz9VKb8lpBveA=="],
 
@@ -5132,7 +5132,7 @@
 
     "selfsigned": ["selfsigned@2.4.1", "", { "dependencies": { "@types/node-forge": "^1.3.0", "node-forge": "^1" } }, "sha512-th5B4L2U+eGLq1TVh7zNRGBapioSORUeymIydxgFpwww9d2qyKvtuPU2jJuHvYAwwqi2Y596QBL3eEqcPEYL8Q=="],
 
-    "semver": ["semver@7.7.4", "", { "bin": "bin/semver.js" }, "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA=="],
+    "semver": ["semver@7.8.5", "", { "bin": { "semver": "bin/semver.js" } }, "sha512-Y7/KDsb8LjooZpwaqGyulO6DQlksgCncchHGk+sZIY4SBvUocMBEFH5Ur1fI4dV+Jvl0w6cjvucaIi40puRioA=="],
 
     "semver-compare": ["semver-compare@1.0.0", "", {}, "sha512-YM3/ITh2MJ5MtzaM429anh+x2jiLVjqILF4m4oyQB18W7Ggea7BfqdH/wGMK7dDiMghv/6WG7znWMwUDzJiXow=="],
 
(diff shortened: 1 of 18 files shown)


michaelbromley:
Addressed the review. Three commits on top, no force-push.

## The blocker: `declareOption()`

Confirmed, and worse than described. I ran all three claims:

- `assertCliPlugin()` already calls `assertUniqueOptions()`, so a plugin gets `CLI plugin "@x/y" declares "--selected-plugin" twice in options of command "add"` and is skipped. The claim in my docblock and PR body that a plugin could hit this was wrong. Only the built-ins, which nothing validates, could reach that code.
- `--json` plus `--no-json` share an `attributeName`, so the guard silently deleted `--no-json`. Commander 14 supports that pair natively and parses it to `{ json: false }`.
- `-p, --plugin` plus `-p, --port` silently deleted `--port`, where Commander 14 throws `conflicting flag '-p'`.

I matched on `attributeName`; Commander matches on flag strings. That gap was the bug.

The dedup is gone. In its place `builtins.spec.ts` runs the built-in definitions through `assertCliPlugin`, which is the same validation a plugin gets, so there is now one policy rather than two. Checked against the base commit, that test fails with:

```
CLI plugin "builtins" declares "--selected-plugin" twice in options of command "add"
```

so it catches the original bug and keeps catching it. The `add` definition fix was the real fix and stands on its own.

## Formatting

Confirmed: `command-registry.ts` and `command-registry.spec.ts` were clean at base and not at the PR head. Prettier has been run over everything this PR touches. `plugins.ts` was already unformatted at base, so that is left alone.

## The vacuous e2e assertions

Confirmed. `sectionAfter()` has moved to `src/shared/__tests__/help-sections.ts` and both the unit and e2e specs import it, so there is one approach rather than two. The e2e tests now assert which section a command is listed in, and that the built-in section does *not* contain the plugin's commands. Its docblock says why offset comparison was wrong.

## Decomposition

`getCommandSources()` and `getRootOptionSources()` are gone. `getCommandTree()` and `getRootOptions()` now return `{ node, source }` and `{ option, source }`, so nothing is taken apart by name and reassembled, and the two cannot disagree.

That removed `runCli`'s fifth parameter as well, so the naming clash you spotted between `rootOptions` meaning definitions in one slot and sources in another no longer exists.

## `printJson`

Dropped the resolved `commands` field and kept `declaredCommands` and `loadedCommands`. One note on your framing: `declaredCommands` was already in this output before the PR and has an e2e test reading it, so removing it would break an existing contract. Keeping the two raw fields is your first option and preserves it.

`MAX_COMMANDS_IN_HINT` now sits above its use. The truncation difference is deliberate and now says so in one line: the picker hint shares a line with the package name, the text listing has a line of its own.

## Docs

`docs/docs/guides/developer-guide/cli/index.mdx` now documents the grouping with example output, says the heading uses the package name exactly as installed, and corrects the `cliCommands` description, which is now also what `vendure plugins` prints for a package that is not enabled. The `getCommandTree()` docblock says explicitly that an `extendCommands` target keeps its original heading, and why.

## Prose

Trimmed. The commander-version story is told once, in the commit that introduces the validation test. The ornamental comments are gone, `addOption`'s docblock is down to the one fact worth keeping, and the two comments in `add/command.ts` that explained an absence at the wrong site have been removed.

## Verification

620 unit tests across 39 files, 86 e2e tests across 7 files. Root help output is byte-identical to before these commits.

PR #5361, opened 2026-09-11
fix(dashboard): add json import attributes when compiling in esm mode

### LINKED ISSUE #5330: Dashboard config compilation does not support JSON imports in esm mode
#4808 taught the dashboard config compiler to copy imported JSON files into its output, so a config or plugin which imports JSON compiles and loads. That works for CommonJS, whose `require()` reads JSON directly. It does not work when `vendureDashboardPlugin` is configured with `module: 'esm'`.

## The problem

In ESM mode the compiler transpiles with `ts.ModuleKind.ESNext` and writes a `package.json` declaring `"type": "module"` into the output. The emitted config therefore carries the import unchanged:

```js
import configData from './config-data.json';
```

Node refuses to load a JSON module in ESM without an import attribute:

```
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module ".../config-data.json"
needs an import attribute of "type: json"
```

The author cannot fix this in their own source. Writing `with { type: 'json' }` in a file that targets CommonJS for the server build is rejected by TypeScript, and the server build is what most projects have.

#4808 added a warning naming the JSON files so the Node error has an explanation above it, and #5329 does the same for skipped `package.json` imports. Neither makes the ESM path work.

## What a fix needs

`packages/dashboard/vite/utils/compiler.ts` would need to emit JSON imports in a form Node accepts in ESM. Two approaches:

1. A transformer that adds `with { type: 'json' }` to import declarations resolving to `.json`. TypeScript 5.8 supports this through `factory.createImportAttributes`. Note that the existing path transformer only runs when `module === 'esm' && tsConfigInfo`, so this cannot simply be folded into it; it has to run whenever the module target is ESM.
2. Emit each JSON file as a real ES module (`export default {...}` in a `.js` file) and rewrite the specifier to match.

`convertExtension` in `packages/dashboard/vite/utils/path-transformer.ts` currently leaves `.json` alone, which is correct for approach 1 and would need changing for approach 2.

## Testing

Vitest resolves JSON imports without the attribute, so a test running under Vitest passes while a real project fails. The test has to run the emitted artifact under plain Node. `packages/dashboard/vite/tests/json-imports.spec.ts` has the fixture and the CommonJS coverage; its `esm` case deliberately asserts only that the file is copied, and says why.

## Scope

`module: 'esm'` is documented as a developer preview (`@since 3.5.1`), so this affects projects which opted into that flag and import JSON. Found while reviewing #4808; raised independently there by @HouseinIsProgramming and by CodeRabbit.

brbousnguar:
# Description

Fixes #5330

With `vendureDashboardPlugin({ module: 'esm' })`, a config or plugin that imports JSON compiled but could not be loaded. The compiler transpiles with `ModuleKind.ESNext` and writes `"type": "module"` into the output, so the emitted config kept a bare import:

```js
import configData from './config-data.json';
```

Node rejects that with:

```
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module ".../config-data.json" needs an import attribute of "type: json"
```

The author can't add the attribute in their own source, because TypeScript rejects `with { type: 'json' }` in a file that targets CommonJS for the server build.

This PR takes the first approach from the issue. `compileTypeScript` now runs a small `after` transformer whenever the target is ESM. It adds `with { type: 'json' }` to import and export declarations whose specifier ends in `.json` and that don't already carry attributes. The existing path transformer still runs only when `module === 'esm' && tsConfigInfo`. The JSON transformer is gated on `module === 'esm'` alone, so it also runs for projects without tsconfig paths. Both go into one `after` array. The path transformer passes `node.attributes` through, so an aliased JSON import ends up with both the rewritten path and the attribute. The emitted output now reads:

```js
import configData from './config-data.json' with { type: "json" };
```

A few details:
- Matching on the `.json` suffix is a superset of the rule `collectLocalSourceFiles` uses to copy a JSON file, so every copied JSON import gets the attribute. Bare specifiers such as `some-pkg/data.json` are never copied, and neither is a `package.json` (#5329), but they still get the attribute, because Node requires it for them too.
- Re-exports (`export { default as x } from './x.json'`) are handled too, because the collector follows export declarations.
- Dynamic `import()` is left alone. The collector doesn't follow it, so that JSON is never copied in the first place.
- CommonJS output is unchanged. It still emits `require("./config-data.json")`.
- The #4808 warning that said imported JSON is unsupported with `module: 'esm'` is removed, since it is no longer true.

`path-transformer.ts` is not touched. `convertExtension` leaving `.json` alone is what this approach needs.

# Breaking changes

None. Only the `module: 'esm'` path (developer preview) changes, and there it turns a load failure into a working config.

# Screenshots

n/a

# Testing

`json-imports.spec.ts`: the ESM case used to assert only that the JSON was copied. It now also loads the emitted `vendure-config.js` in a plain Node child process (`process.execPath --input-type=module`) and reads back one value per JSON import. It runs outside Vitest on purpose, because Vitest's loader accepts JSON without the attribute.

The `fixtures-json` fixture gains two cases, so the test covers:
- the config's own JSON import (`config.customFields.Product[0].name`);
- the plugin's transitive JSON import (`MyPlugin.sheetId`);
- a re-export, `export { default as reexportedData } from './reexported-data.json'`;
- an import that already has `with { type: 'json' }`. The test also checks that its emitted line still has exactly one `type:`, so the attribute isn't added twice.

The fixture's `package.json` is `"type": "commonjs"`, so an editor type-checking the fixture under NodeNext flags the pre-attributed import (TS2856). That's expected, and the line has a comment saying why it's there. `vite/tests` is excluded from `check-types`. The CommonJS case compiles the same fixture and still passes.

Before the fix, the new test fails with the error from the issue:

```
× should load a config which imports JSON in esm mode
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module "file:///.../__temp/json-esm/config-data.json" needs an import attribute of "type: json"
```

After the fix it passes. Commands run from `packages/dashboard`, after `bun run build:core-common` at the repo root:

```bash
bunx vitest run vite/tests/json-imports.spec.ts
bunx vitest run vite/ --exclude 'vite/tests/bundle-singleton*'   # 16 files, 135 tests pass
bun run check-types
bunx prettier --check vite/utils/compiler.ts vite/tests/json-imports.spec.ts vite/tests/fixtures-json/
```

I also checked by hand, with a throwaway fixture compiled in ESM mode and loaded under plain Node 22, that a tsconfig alias to a JSON file (`@data/z.json`) is rewritten to a relative path and gets the attribute.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5361"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791718313&installation_model_id=20193&pr_number=5361&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5361&signature=9ce707c0a52956cbb76fc7bab1d6f9b329d69c805b673d7818c55e7a11b3e36e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/vite/tests/fixtures-json/attributed-data.json
+++ b/packages/dashboard/vite/tests/fixtures-json/attributed-data.json
@@ -1,0 +1,1 @@
+{ "label": "attributed" }
--- a/packages/dashboard/vite/tests/fixtures-json/empty-attributes-data.json
+++ b/packages/dashboard/vite/tests/fixtures-json/empty-attributes-data.json
@@ -1,0 +1,1 @@
+{ "label": "empty-attributes" }
--- a/packages/dashboard/vite/tests/fixtures-json/reexported-data.json
+++ b/packages/dashboard/vite/tests/fixtures-json/reexported-data.json
@@ -1,0 +1,1 @@
+{ "label": "reexported" }
--- a/packages/dashboard/vite/tests/fixtures-json/vendure-config.ts
+++ b/packages/dashboard/vite/tests/fixtures-json/vendure-config.ts
@@ -4,12 +4,15 @@ import { VendureConfig } from '@vendure/core';
 // second one. TypeScript flags it here only because this fixture targets CommonJS.
 import attributedData from './attributed-data.json' with { type: 'json' };
 import configData from './config-data.json';
+// #5330 — an empty attribute list has no `type`, so the esm emit must still add one.
+import emptyAttributesData from './empty-attributes-data.json' with {};
 import { MyPlugin } from './my-plugin/src/my.plugin.js';
 
 // #5330 — a re-export of JSON needs the attribute in esm mode as much as an import does.
 export { default as reexportedData } from './reexported-data.json';
 
 export const attributedLabel = attributedData.label;
+export const emptyAttributesLabel = emptyAttributesData.label;
 
 export const config: VendureConfig = {
     apiOptions: { port: 3000 },
--- a/packages/dashboard/vite/tests/fixtures-json/vendure-config.ts
+++ b/packages/dashboard/vite/tests/fixtures-json/vendure-config.ts
@@ -1,7 +1,15 @@
 import { VendureConfig } from '@vendure/core';
 
+// #5330 — this import already has the attribute, so the esm emit must not add a
+// second one. TypeScript flags it here only because this fixture targets CommonJS.
+import attributedData from './attributed-data.json' with { type: 'json' };
 import configData from './config-data.json';
 import { MyPlugin } from './my-plugin/src/my.plugin.js';
+
+// #5330 — a re-export of JSON needs the attribute in esm mode as much as an import does.
+export { default as reexportedData } from './reexported-data.json';
+
+export const attributedLabel = attributedData.label;
 
 export const config: VendureConfig = {
     apiOptions: { port: 3000 },
--- a/packages/dashboard/vite/tests/json-imports.spec.ts
+++ b/packages/dashboard/vite/tests/json-imports.spec.ts
@@ -65,13 +65,15 @@ describe('compiling a config which imports .json files', () => {
         expect(attributedImport?.match(/type:/g)).toHaveLength(1);
 
         const script = [
-            'const { config, reexportedData, attributedLabel } = await import(process.argv[1]);',
+            'const { config, reexportedData, attributedLabel, emptyAttributesLabel } =',
+            '    await import(process.argv[1]);',
             'const [MyPlugin] = config.plugins;',
             'console.log(JSON.stringify([',
             '    config.customFields.Product[0].name,',
             '    MyPlugin.sheetId,',
             '    reexportedData.label,',
             '    attributedLabel,',
+            '    emptyAttributesLabel,',
             ']));',
         ].join('\n');
         const { stdout } = await execFileAsync(process.execPath, [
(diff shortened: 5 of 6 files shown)


brbousnguar:
Thanks for the review! Fixed: `needsJsonAttributes` now checks the existing import attributes for a `type` entry instead of treating any attribute list as already complete, and merges `type: 'json'` in when it's missing (keeping other attributes as-is). Added a regression fixture for a JSON import with an empty attribute list (`with {}`) to cover it.

brbousnguar:

biggamesmallworld:

biggamesmallworld:
Thanks for this, @brbousnguar. We're going to merge it.

What we changed:

- Corrected a comment in `compiler.ts` that said TypeScript emits the JSON import unchanged, which contradicts the transformer below it.

biggamesmallworld (approved):
Verified locally: ERR_IMPORT_ATTRIBUTE_MISSING reproduced on master and fixed by this change; CI build, unit and e2e ran green on fd4b1aa9.

Issue #5353, opened 2026-09-10
chore: @typescript-eslint plugin and parser are four majors apart, breaking lint on a fresh install


biggamesmallworld:
### Summary

The root `package.json` declares two incompatible major versions:

```json
"@typescript-eslint/eslint-plugin": "^8.68.0",
"@typescript-eslint/parser": "^5.54.1",
```

`bun.lock` resolves the plugin to `8.68.0`, which declares `"peerDependencies": { "@typescript-eslint/parser": "^8.68.0" }`. The parser resolves to `5.62.0`.

On a fresh install, type-aware rules then crash:

```
services.getTypeAtLocation is not a function
```

It crashes on files the change under test never touched, so `bun run lint` is unusable rather than merely noisy.

### Why it is not visible in every checkout

An existing `node_modules` can still hold the older plugin. In one checkout here, `node_modules/@typescript-eslint/eslint-plugin` is `5.62.0` and matches the parser, so linting works normally:

```
$ npx eslint packages/core/src/service/services/order.service.ts
✖ 5 problems (0 errors, 5 warnings)
```

The same repository, freshly installed per the lockfile, gets plugin `8.68.0` against parser `5.62.0` and crashes. So this reproduces for a new contributor and in CI, and does not reproduce for anyone whose `node_modules` predates the plugin bump.

### Suggested fix

Move the parser to `^8.68.0` to match the plugin. The two packages are released in lockstep and are not supported across majors.

Worth checking at the same time whether the eslint config still uses the flat-config or legacy shape the v8 plugin expects, since the parser has been four majors behind and the config may have been written against v5.

### Found via

A merge-lane session on #5338, where a worktree with a fresh `bun install` could not lint its own changes.

PR #5363, opened 2026-09-11
fix(testing): avoid races creating the sqljs data directory

### LINKED ISSUE #5322: @vendure/testing: SqljsInitializer.populate() has a TOCTOU mkdir race under parallel workers (EEXIST)
## Summary

`SqljsInitializer.populate()` in `@vendure/testing` creates its data directory with an `existsSync` check followed by a **non-recursive** `mkdirSync`. That check-then-act is a TOCTOU race: when the harness runs across more than one vitest worker (`pool: "forks"`, `maxWorkers > 1`) and several specs share the same `dataDir`, two workers that start `populate()` at the same instant on a cold directory both see it missing and both call `mkdirSync`, so the loser throws and fails its spec:

```
EEXIST: file already exists, mkdir '__data__'
```

Confirmed present on `master` today: `packages/testing/src/initializers/sqljs-initializer.ts`

```ts
async populate(populateFn: () => Promise<void>): Promise<void> {
    if (!fs.existsSync(this.dbFilePath)) {
        const dirName = path.dirname(this.dbFilePath);
        if (!fs.existsSync(dirName)) {
            fs.mkdirSync(dirName);          // <-- non-recursive + guarded by existsSync => TOCTOU race
        }
        // ...
        await populateFn();
    }
}
```

## When it bites

- Only under **parallel workers** (`maxWorkers > 1`) where **multiple specs share one `dataDir`** (e.g. a common `new SqljsInitializer("__data__")`) **and the directory is cold** (first run, or a gitignored `__data__` on CI). The default single-worker / warm-directory path never hits it, which is why it is easy to miss.
- It surfaces in downstream projects that parallelise their e2e suite through the harness. In ours (a Vendure 3.7.3 backend), moving the SQLite suite from 1 to 2 vitest forks made it reproducible on a cold `__data__`; pinning back to 1 worker hid it again.

This is a separate cause from the async-job assertion races catalogued in #4949 — those are in Vendure's own suite; this one is in the initializer itself and can hit any parallel consumer of `@vendure/testing`.

## Fix (one line)

`mkdirSync(..., { recursive: true })` is idempotent (no throw when the directory already exists) and also creates intermediate directories, so it removes both the race and the now-redundant `existsSync` guard:

```ts
const dirName = path.dirname(this.dbFilePath);
fs.mkdirSync(dirName, { recursive: true });
```

## Workaround (for downstream consumers, until this is released)

Pre-create the shared `dataDir` once in a vitest `globalSetup`, before any worker forks:

```ts
// vitest globalSetup
import { mkdirSync } from "node:fs";
export default function setup(): void {
    mkdirSync("__data__", { recursive: true }); // the relative dataDir handed to SqljsInitializer
}
```

Happy to open a PR with the one-line change if that is useful.

sridharkalaibala:
# Description

Fixes #5322. `SqljsInitializer.populate()` now creates the data directory with recursive `mkdirSync`, so a second test worker creating the directory between the check and mkdir no longer causes `EEXIST`. This also supports missing intermediate directories.

Added four tests covering the race, nested directory creation, an existing directory, and reuse of a cached database. The race test uses real filesystem calls and injects another worker's directory creation immediately before mkdir, making the original failure deterministic.

# Breaking changes

None. Existing database caches still skip population.

# Validation

- Before the fix: race and nested-directory tests fail with `EEXIST` and `ENOENT`; two compatibility cases pass.
- After the fix: all seven `@vendure/testing` unit tests pass.
- `bun run build:core-common`: passed.
- Testing package build (`tsc -p tsconfig.build.json`): passed.
- Focused source/test TypeScript check using the repository's compiler settings: passed.
- New test file passes Prettier; source retains existing surrounding formatting.
- ESLint could not complete because of the existing parser/plugin incompatibility tracked in #5353 (`services.getTypeAtLocation is not a function`). No dependency changes are included. The lint-dependent local commit hook was skipped after running the checks above; the commit message was checked separately.

Validated on Windows with Node 24 and Bun 1.4.2, installed with the frozen lockfile. Full database-backed ecommerce E2E suites were not run; these tests exercise directory setup and population callbacks directly. Prepared with AI assistance.

# Checklist

- [x] I have updated the README if needed (not needed)

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5363"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791744723&installation_model_id=20193&pr_number=5363&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5363&signature=bbd7595d2f6ee45eec5b5fb0a658088f53673fc22c8e9c3da28ea26c356d6a97"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/testing/src/initializers/sqljs-initializer.spec.ts
+++ b/packages/testing/src/initializers/sqljs-initializer.spec.ts
@@ -1,0 +1,76 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { SqljsInitializer } from './sqljs-initializer';
+
+describe('SqljsInitializer', () => {
+    let temporaryDir: string;
+
+    beforeEach(() => {
+        temporaryDir = fs.mkdtempSync(path.join(os.tmpdir(), 'vendure-sqljs-'));
+    });
+
+    afterEach(() => {
+        vi.restoreAllMocks();
+        fs.rmSync(temporaryDir, { recursive: true, force: true });
+    });
+
+    it('populates when another worker creates the directory just before mkdir', async () => {
+        const dataDir = path.join(temporaryDir, 'data');
+        const initializer = new SqljsInitializer(dataDir);
+        const options = await initializer.init('first.spec.ts', { type: 'sqljs' });
+        const mkdir = fs.mkdirSync;
+        vi.spyOn(fs, 'mkdirSync').mockImplementationOnce((directory, mkdirOptions) => {
+            // #5322: another worker wins the race after the missing-directory check.
+            mkdir(directory, { recursive: true });
+            return mkdir(directory, mkdirOptions);
+        });
+        const populate = vi.fn(async () => {
+            expect(fs.statSync(dataDir).isDirectory()).toBe(true);
+            expect(options.autoSave).toBe(true);
+            expect(options.synchronize).toBe(true);
+        });
+
+        await initializer.populate(populate);
+
+        expect(populate).toHaveBeenCalledOnce();
+        expect(options.autoSave).toBe(false);
+        expect(options.synchronize).toBe(false);
+    });
+
+    it('creates missing parent directories', async () => {
+        const dataDir = path.join(temporaryDir, 'nested', 'data');
+        const initializer = new SqljsInitializer(dataDir);
+        await initializer.init('nested.spec.ts', { type: 'sqljs' });
+        const populate = vi.fn(async () => {
+            expect(fs.statSync(dataDir).isDirectory()).toBe(true);
+        });
+
+        await initializer.populate(populate);
+
+        expect(populate).toHaveBeenCalledOnce();
+    });
+
+    it('populates a new database in an existing directory', async () => {
+        const initializer = new SqljsInitializer(temporaryDir);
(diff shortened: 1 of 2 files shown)


sridharkalaibala:

biggamesmallworld:

biggamesmallworld:
Thanks for this, @sridharkalaibala. We're going to merge it.

What we changed:

- Made the race test assert that its mocked `mkdirSync` ran, so the test cannot pass without reproducing the race.

biggamesmallworld (approved):
Verified locally: EEXIST race reproduced on master and fixed by this change; CI build, unit and e2e ran green on 94fe6c7f.

PR #5366, opened 2026-09-13
fix(dashboard): restore display component precedence for additionalColumns

### LINKED ISSUE #5347: additionalColumns never consult the display component registry
## Describe the bug

`useGeneratedColumns` builds two sets of columns. The generated ones consult the display component registry, so `addDisplayComponent` can override how a field renders. The columns supplied through `additionalColumns` never consult the registry at all, so registering a display component for one of them does nothing and gives no indication of why.

This is the same class of defect as #5201, in a different code path.

## Expected behaviour

Registering a display component for a column should override that column's rendering, whether the column was generated from the schema or supplied through `additionalColumns`. If that is not wanted for `additionalColumns`, the documentation should say so, because at the moment the API looks like it applies to every column.

## Additional context

Found while reviewing #5339, which fixes the precedence problem for generated columns. It was deliberately left out of scope there, because `additionalColumns` is a separate loop that #5339 does not touch and pulling it in would have widened a contained fix.

Worth deciding alongside #5346, since both are about how the display component registry is consulted.

wakqasahmed:
#5339 fixed display components registered via `addDisplayComponent()` being ignored for schema-derived columns in `useGeneratedColumns`, but that PR deliberately left `additionalColumns` alone since it's a completely separate loop. This issue is the natural follow-up: registering a display component for an `additionalColumns` entry currently does nothing at all, because that loop pushes the column through unchanged, whatever `cell` the extension itself supplied.

This applies the same precedence check to that loop: compute the same `generateDisplayComponentKey(pageId, blockId, id)` key, look it up via `getDisplayComponent`, and if something is registered, render that instead of the column's own cell. If nothing is registered, behavior is exactly as before.

Added two tests mirroring the existing precedence tests in this file - one where a registered component overrides an `additionalColumns` cell, one where an unregistered column still falls back to its own cell unchanged.

Closes #5347

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5366"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791856383&installation_model_id=20193&pr_number=5366&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5366&signature=6f1f61a06062ca8bb0a1354d35fff41c3c8e65a45992aa00a1a4bd58c16f5277"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/packages/dashboard/src/lib/components/data-table/use-generated-columns.spec.tsx
+++ b/packages/dashboard/src/lib/components/data-table/use-generated-columns.spec.tsx
@@ -39,13 +39,14 @@ const cellContext = {
  * `createElement` for any function cell, so a cell using hooks behaves here as it does in
  * a real table; calling `column.cell(context)` directly would not.
  */
-function renderColumnCell(pageId: string, columnId: string): string {
+function renderColumnCell(pageId: string, columnId: string, additionalColumns?: any): string {
     const captured: { columns?: Array<{ id?: string; cell?: any }> } = {};
 
     function Harness() {
         const { columns } = useGeneratedColumns({
             fields,
             customizeColumns,
+            additionalColumns,
             includeSelectionColumn: false,
             includeActionsColumn: false,
         });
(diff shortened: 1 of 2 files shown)


biggamesmallworld:

biggamesmallworld:

What we changed:

- Moved the display component lookup into one `resolveDisplayComponent` helper, so both column loops build the registry key the same way.
- Corrected a comment that called `CellWrapper` the only place the display component registry is read.
- Changed the additional-column test to read `row.original`, because an additional column's `value` is undefined outside the test.

biggamesmallworld (approved):
Verified locally: registered display component ignored for additionalColumns on master and rendered with this change; CI build, unit and e2e ran green on 9bc72875.

PR #5370, opened 2026-09-14
fix(core): recalculate Order totals from current Surcharges, not a st

### LINKED ISSUE #5332: Order totals can be persisted from a stale surcharges snapshot
**Describe the bug**

`OrderService.applyPriceAdjustments` calculates and persists `Order.subTotal` and `Order.subTotalWithTax` from the in-memory `order.surcharges` array. That array is a snapshot taken when the Order was loaded, so a Surcharge added or removed since the load is not reflected in the totals which get written.

The result is a persisted Order whose totals do not match its own surcharge rows. `Order.total` and `Order.totalWithTax` are derived from the persisted `subTotal` columns, so `addPaymentToOrder` can charge an amount which does not correspond to what the Order actually holds.

Chain:

- `DefaultOrderTaxCalculationStrategy` sums `order.surcharges` into the subtotal (`packages/core/src/config/tax/default-order-tax-calculation-strategy.ts:28`, `:52`). `OrderLevelTaxCalculationStrategy` does the same (`packages/core/src/config/tax/order-level-tax-calculation-strategy.ts:102`).
- `OrderCalculator.calculateOrderTotals` writes the result onto the Order (`packages/core/src/service/helpers/order-calculator/order-calculator.ts:376`).
- `subTotal` and `subTotalWithTax` are persisted columns (`packages/core/src/entity/order/order.entity.ts:170`), and `total` / `totalWithTax` are computed from them (`:245`, `:269`).
- `OrderService.applyPriceAdjustments` then saves the Order (`packages/core/src/service/services/order.service.ts:2462`).

The window is easy to hit, because every cart mutation loads the Order at entry and publishes `OrderLineEvent` before this save runs. Any blocking event handler which touches surcharges in that window, or any concurrent request doing so, leaves the totals computed from the pre-handler array.

**To Reproduce**

1. Register a blocking `OrderLineEvent` handler which adds a Surcharge to the Order, for example a loyalty or voucher discount of -500.
2. Call `adjustOrderLine` on that Order.
3. Read the Order back.

**Expected behavior**

The persisted `subTotal` / `subTotalWithTax` account for every Surcharge row attached to the Order at the time of the save, or the save is rejected so the caller can retry.

**Actual behavior**

The surcharge row exists and is attached, but the persisted totals were calculated without it. The Order shows a -500 surcharge and a total which does not include it. A subsequent `addPaymentToOrder` uses the stale `totalWithTax`.

**Environment**

- @vendure/core version: master
- Database: driver independent, this is application-level

**Additional context**

Found while reviewing #5164, which fixes a different symptom of the same stale snapshot: before that PR, saving the Order also detached surcharge rows it did not know about, by orphan-nullifying `surcharge.orderId`. That is fixed by `orphanedRowAction: 'disable'` on `Surcharge.order`, so the rows now survive. The totals can still be stale, which is why this is filed separately.

Possible directions, no strong preference:

- Reload the surcharge rows immediately before `calculateOrderTotals` in `applyPriceAdjustments`, so the totals are computed from what is persisted rather than from the snapshot.
- Detect the stale write instead of preventing it: optimistic lock on the Order, reject and retry the repricing.

The same question applies to `OrderLine` and `ShippingLine`, which are loaded as snapshots and feed the same totals. Surcharges are the easiest to hit because they are the relation a plugin is most likely to mutate from an event handler.

Ismatulla:
`OrderService.applyPriceAdjustments` calculated `Order.subTotal`/`subTotalWithTax` from the in-memory `order.surcharges` snapshot taken when the Order was loaded, so a Surcharge added or removed since then (e.g. by a blocking event handler) was not reflected in the persisted totals even though its row was correctly attached.

Reload surcharges from the DB immediately before recalculating totals, and reorder `removeSurchargeFromOrder` to delete the row before recalculating so it isn't reloaded back into the order.

Fixes #5332

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5370"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791983712&installation_model_id=20193&pr_number=5370&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5370&signature=0f10a69f2ea556059309aa198b5b867dace4c9945e94ab7e3360144a5ab1e1a9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Ismatulla:

Ismatulla:
recheck

Ismatulla:
recheck

Ismatulla:
recheck

PR #5396, opened 2026-09-16
test(core): stop job-queue service spec leaking state between tests

### LINKED ISSUE #5395: test(core): `job-queue.service.spec.ts` is order-dependent through leaked strategy state
`packages/core/src/job-queue/job-queue.service.spec.ts` constructs one `TestingJobQueueStrategy` and one `TestingJobBufferStorageStrategy` at module level and reuses both across all 13 tests. `beforeEach` builds a fresh Nest module, but `module.close()` only calls `strategy.destroy()`, which flips `hasInitialized` and clears the eviction timer. The strategy's `jobs` map, `unsettledJobs` queues and `concurrency`, and the buffer storage's buffered jobs, all survive from one test into the next.

The file therefore only passes in its current declaration order. Running it with `--sequence.shuffle` fails between 4 and 10 of the 13 tests, varying by seed:

```
bunx vitest run --config vitest.config.mts src/job-queue/job-queue.service.spec.ts --sequence.shuffle
Tests  10 failed | 3 passed (13)
Tests   4 failed | 9 passed (13)
Tests   6 failed | 7 passed (13)
```

Two dependencies are explicit. `with concurrency` sets `concurrency = 2` and never restores it; `buffers the specified jobs` asserts that two jobs on `buffer-test-queue-2` are both `RUNNING` after a single poll, which needs that second slot, and fails with `expected 'PENDING' to be 'RUNNING'` without it. `flushes and reduces buffered jobs` adds no jobs of its own and asserts a `'hello world'` result, which is the reduction of the two jobs buffered by the test before it. Less visibly, `sets long-running jobs to pending on destroy` and `should not start a queue if its name is in the active list` each leave a `PENDING` job in `unsettledJobs['test']`, which the next test to create a `test` queue picks up through its own process function. (#5293 already pins `concurrency = 2` explicitly in `processes existing jobs on start`, which relied on the same leak.)

Proposed fix: give `TestingJobQueueStrategy` a `reset()` that clears `jobs` and `unsettledJobs`, and `TestingJobBufferStorageStrategy` the equivalent for its buffers. Both fields are `protected` on the in-memory base classes, so no cast is needed. Call both from the outer `beforeEach` along with `concurrency = 1`, and have the two tests that genuinely need a second slot set it themselves — **before `createQueue()`**, because `ActiveQueue` reads `concurrency` into a `readonly` field in its constructor, so assigning it inside the test body is too late. `flushes and reduces buffered jobs` then buffers its own jobs.

This is test hygiene only; no runtime behaviour is affected. The reason to fix it is that a new test inserted in the wrong place, or any change to `with concurrency`, currently breaks unrelated tests in ways that are hard to attribute.

ivanvasyliuk:
# Description

`job-queue.service.spec.ts` shares one `TestingJobQueueStrategy` and one `TestingJobBufferStorageStrategy` at module level. `module.close()` calls `destroy()`, which only flips `hasInitialized` and clears the eviction timer, so the `jobs` map, the `unsettledJobs` queues, `concurrency` and the buffered jobs all survive into the next test.

The file was order-dependent as a result. Running it with `--sequence.shuffle` fails 4 to 10 of its 13 tests. Two tests passed only because of the leak: `buffers the specified jobs` needs the `concurrency = 2` that `with concurrency` leaves behind, and `flushes and reduces buffered jobs` asserted on the reduction of jobs buffered by the test before it.

Both testing strategies get a `reset()` — the fields are `protected` on the in-memory base classes, so no cast is needed — and the outer `beforeEach` calls them and restores `concurrency`. The two tests that genuinely need a second concurrency slot now set it themselves, before `createQueue()`, since `ActiveQueue` reads the value when the queue is constructed. `flushes and reduces buffered jobs` buffers its own jobs.

No runtime behaviour changes. Shuffled runs now pass 13/13.

Fixes #5395

# Breaking changes

None. Test-only change; no runtime, schema, GraphQL or config change.

# Screenshots

None. The shuffled-run results above are the relevant evidence.

# Checklist

📌 Always:

👍 Most of the time:

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5396"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792139864&installation_model_id=20193&pr_number=5396&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5396&signature=1577f8a220c5ae1f6970372675848e5f7fccad87f17285fa8d13819228a372cc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld (approved):
Looks good, thanks. Ran the spec with `--sequence.shuffle` five times locally, 13/13 each time. I'll open a small follow-up for a few spec nits.

PR #5400, opened 2026-09-16
feat(cli): support globally installed CLI and plugin packages

michaelbromley:
Makes the Vendure CLI usable when it is installed globally (`npm install -g @vendure/cli`), so its commands and the commands contributed by globally installed plugin packages work from any directory.

## The problem

Three separate things stopped this working.

**Plugin resolution was project-only.** Everything hung off one function that walked up from the current directory looking for a project `package.json`. Run from a home directory there is none, so no plugins loaded. Even with a project, a listed package had to be one of its direct dependencies, which a globally installed sibling never is.

**Built-in commands crashed.** `@vendure/core` is a devDependency of `@vendure/cli`, so a global install never brings it. Commands that import it at module level died with a raw `MODULE_NOT_FOUND` stack trace. Verified: a project with `@vendure/core` in its own `node_modules` still failed, because Node resolves from the requiring file's directory, not the working directory.

**Commands that need a project did not say so.** `vendure add` drew its intro, asked two questions, then failed on a missing tsconfig. A project check existed but was unreachable, sitting below a module-level import of the missing package.

### Commands declare whether they need a project

New `requiresProject` field on command definitions, honoured for built-ins and plugin commands alike. The host refuses before the first prompt and before the implementation is imported:

```
vendure migrate must be run from a Vendure project directory.
No package.json with a Vendure dependency was found in /home/user or any parent directory.
```

Declared on a group it applies to the subtree; a subcommand opts out with `requiresProject: false`. Outside a project, `--help` marks these commands rather than hiding them, so the command list is the same wherever the CLI runs:

```
  add [options]    Add a feature to your Vendure project *

  * Requires a Vendure project. You are not in one.
```

The check is deliberately loose — a `package.json` with a Vendure dependency at or above the current directory. It separates "not in a project" from "in a project" and promises nothing more, so a command with stricter needs still validates for itself.

`doctor` deliberately does **not** set it. Reporting that a directory is not a Vendure project is one of its checks, so refusing to run would remove the answer in the case that most needs it. It now degrades into a proper report instead of a stack trace.

### Project-scoped module resolution

Commands that need `@vendure/core` load the project's copy rather than the CLI's. This is a correctness fix independent of global installs: `generateMigration`, `preBootstrapConfig` and `getFinalVendureSchema` behave differently between versions and must match the version the server boots with, and the config singleton they read and write has to be the same instance the project's `vendure-config.ts` reaches. Both previously worked only because npm hoisting happened to put one copy where the CLI's own lookup landed.

Type-only imports are now `import type`, so naming the module costs nothing at runtime. `@nestjs/graphql` gets the same treatment, since it reads type paths that core defines.

### A global plugin scope

The single project context becomes a `PluginScope`. The project and global scopes differ in three things — where the allowlist lives, what makes a listed package eligible, and which packages the scope can enumerate — and share loading, validation and reporting.

The global allowlist lives in `cli.json` in the CLI config directory (`~/.config/vendure`, `%APPDATA%\vendure`, or `VENDURE_CLI_CONFIG_DIR`). Packages resolve from the CLI's own installation directory, where a globally installed plugin sits beside it. `pluginRoots` covers layouts the CLI does not otherwise reach.

```bash
npm install -g @vendure/cli @vendure/cloud
vendure plugins add --global @vendure/cloud
```

Both scopes apply at once. Global registers first and project second, so a project overrides the machine — the same rule that makes the last entry of a single list win. A package enabled in both loads once, from the project, so the copy pinned alongside the project's code runs.

Explicit activation is preserved. A globally installed package must not be able to add commands to every shell on the machine just by being installed.

Without `--global`, `add` and `remove` write to the project when there is one and to the machine-wide list when there is not. `vendure plugins` reports the scope of each package.

### Enabling without writing a file

`VENDURE_CLI_PLUGINS` is a comma-separated list added to whatever `cli.json` holds, for environments with no project and no writable home — a container image, or an ephemeral sandbox given to an agent. Read afresh each invocation, so `plugins remove --global` cannot countermand it; unset it instead.

## Verification

Tested against a simulated global install, with the CLI and a plugin as siblings in one `node_modules`, run from a directory with no project:

```
$ vendure project list
Unknown command "project".
It is provided by @vendure-e2e/cloud-cli-plugin, which is installed but not enabled.
Enable it with: vendure plugins add --global @vendure-e2e/cloud-cli-plugin

$ vendure plugins add --global @vendure-e2e/cloud-cli-plugin
Enabled CLI plugin @vendure-e2e/cloud-cli-plugin for this machine

$ vendure project list
CLOUD_RESULT {"command":["project","list"],...}
```

Project resolution was verified against a project whose `@vendure/core` is reachable only from the project and not from the CLI's own location. The CLI loaded the project's copy, which a plain `require` cannot do.

670 unit tests and 88 e2e tests pass. Type checking is clean. New coverage: the project gate and its inheritance, the help legend at each level, the global config file and environment variable, scope ordering, and cross-scope deduplication.

Two existing doctor e2e tests failed against an earlier version of the gate. They were correct and the gate was wrong, which is how `doctor` came to be excluded.

## Notes for review

- `requiresProject` is added to `CLI_PLUGIN_EXTENSION_POINTS`, so a plugin can detect whether the CLI it resolved honours it.
- Installed globally, `@vendure/cli` and a plugin are independent installs with nothing pinning them together. The existing capability check is the tool for this and is documented as mattering more here.
- Docs cover the new field including the `doctor` exception, and a new "Globally installed plugins" section.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5400"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792160791&installation_model_id=20193&pr_number=5400&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5400&signature=bfaa9287b58491c9a63f61051e5068376e3fb4153e355f916eef7415453556a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


michaelbromley:
Addressed the review findings in a1defaded.

**Scope conflation (must fix).** Confirmed and fixed. `findCliInstallLocation` now decides whether the enclosing `node_modules` is actually a global root, by looking for a Vendure project above it: a project's `node_modules` has one, a global prefix does not. The search starts at the parent, since `@vendure/cli` depends on `@vendure/common` and would otherwise look like a project itself. A project-local CLI now contributes no global root and no candidates, so a project dependency can never be listed or enabled as machine-wide. The lookup is injectable, and there are unit tests for the global, project-local and source-checkout layouts plus an e2e asserting a project dependency reports only `scope: project`.

**Config errors as packages (must fix).** Fixed. `CliPluginLoadResult` gained a `scopeErrors` channel, so a file is no longer reported as a package with advice to remove it. It now reads: `No CLI plugins were loaded from it. Fix <path>, or delete it to start over.`

**Dead code.** `buildEmptyScope`, `getCliPluginProjectContext`, `CliPluginProjectContext` and the unreachable branch in `assertPackageCanBeEnabled` are gone. Removing that branch forced a non-null assertion, so `resolveTargetScope` now returns the scope itself and there is nothing to assert.

**Eligibility composition.** Moved onto the scope as `scope.check()`. `checkScopeDeclaresPlugin` is deleted.

**The `env` coupling.** Replaced with an explicit `scopes` option. Specs that mean the project scope now say `scopes: ['project']`.

**Duplicated rules.** One `effectiveRequiresProject` helper, one `pluginsCommandFor` for the flag spelling.

**Text output column.** Kept, and called out below.

**Tests.** The weak assertions are tightened, the doctor field-assert test is gone (the ungated list already covers it), and there is a new `cli-global-plugins.e2e-spec.ts` running a copied CLI inside a simulated global prefix — the coverage gap that let both defects through.

**Comments and docs.** Trimmed the duplicated `requiresProject` JSDoc to a `{@link}`, cut the doctor explanation to the command file and the docs, rewrote the narrating comments in `project-core.ts` and the docs to describe the code, moved the policy comment onto `hasVendureDependency`, and named the `vendure` literal.

**Also in this commit:** the marker and legend are now coloured — yellow marker, dim legend — following `styleHelpTitle`'s existing pattern, with colour forced on in the tests.

Not changed, with reasoning: `vendure plugins` text output keeps the `scope` column, since scope is now load-bearing and the format is tab-separated for exactly this kind of extension. The `--config` versus `codemod` distinction is now stated in the codemod comment: a codemod's path names the files it rewrites, while `--config` names one file inside a project the command still needs the whole of.

676 unit and 94 e2e tests pass. Type checking clean.

mehringer68:
## Review

The `PluginScope` abstraction is the right shape: three genuine differences captured, everything else shared, and `resolve-cli-plugins-scopes.spec.ts` is real testing (temp-directory fixtures, exact `toEqual`, and the "loaded once, from the project" case distinguishes the two copies by command name rather than package name). Four things block merge.

### Must fix

**1. A hoisted monorepo `node_modules` is classified as a global install.**

`isGlobalNodeModules()` in `packages/cli/src/shared/resolve-cli-plugins.ts` asks whether `findVendureProjectRoot(dirname(nodeModulesDir))` is undefined. In a workspace monorepo the root `package.json` carries `workspaces` plus tooling and no `@vendure/*` dependency; the Vendure dep lives in `packages/server/package.json`, below the hoisted `node_modules`, not above it.

Verified with a fixture: CLI at `<root>/node_modules/@vendure/cli`, root manifest `{workspaces, devDependencies: {typescript}}`, server workspace depending on `@vendure/core`. `findCliInstallLocation()` returned `globalNodeModules: '<root>/node_modules'`.

Three consequences:
- every hoisted package declaring `vendure.cliPlugin` is listed twice by `vendure plugins`, once as `global`;
- the unknown-command hint prints `vendure plugins add --global @x` for a package that belongs to one repo;
- `listCandidates()` enumerates and resolves the entire hoisted tree. Measured against this repo's `node_modules` (1624 entries): 406 ms, and `maybeWriteInactivePluginsHint()` runs that synchronously before `parseAsync` on every invocation that is not `--help`, `--version` or `plugins`. Every `vendure dev` in such a repo pays it.

The existing test `does not treat a project node_modules as a global installation` passes only because its fixture puts `@vendure/cli` in the root manifest. Worth adding the workspace-root case. For the fix, "no Vendure dependency above" looks too weak a signal: either require that no `package.json` exists in any ancestor of the `node_modules`, or check the directory against npm's global prefix.

**2. `vendure plugins remove --global X` reports success while X stays enabled.**

In `packages/cli/src/commands/plugins/plugins.ts`, `readEnabledPlugins('global')` returns `getGlobalPluginAllowlist()`, which includes names from `VENDURE_CLI_PLUGINS`, so the "nothing to remove" guard passes. `removeGlobalPlugin()` then filters only `config.plugins`, writes an unchanged file, and the command logs `Disabled CLI plugin X for this machine`. The next invocation still loads it. The unit test `cannot remove a package the environment variable supplies` in `cli-global-plugin-config.spec.ts` freezes the behaviour rather than fixing the message. Better to detect the env-supplied name and say so: "X is enabled by VENDURE_CLI_PLUGINS; unset that variable instead."

**3. Two definitions of "in a project" inside one feature.**

The gate, the help legend, `requireFromProject` and the docs all use `findVendureProjectRoot` (a `package.json` with a Vendure dependency, walking up). `resolveTargetScope()` in `plugins.ts` uses `readCliProjectPackageJson()`, which goes through `resolveCliProjectRoot()` and falls back to `roots[0]`, the nearest `package.json` whatever it contains. So in an unrelated npm package directory, or a home directory with a stray `package.json`, `vendure plugins add @vendure/cloud` writes `vendure.cli.plugins` into that manifest instead of the machine-wide list, contradicting the rule the docs state. `findVendureProjectRoot` here too.

**4. Eight changed files fail `prettier --check`** against the repo's own config: `cli.ts`, `resolve-cli-plugins.ts`, `project-validation.ts`, `plugins.ts`, `command-registry-store.spec.ts`, `resolve-cli-plugins-scopes.spec.ts`, `cli-test-utils.ts`, `cli-global-plugins.e2e-spec.ts`. Import order in `cli.ts`, the misaligned `scopes: ['project'],` at 12 spaces inside 16-space blocks in `command-registry-store.spec.ts`, a double blank line in `cli-test-utils.ts`, an unused `join` import in `cli-global-plugins.e2e-spec.ts`. (`config-check.ts` and `production-check.ts` were already failing on `minor`.)

### Should fix, not blocking

**5. The Vendure-dependency check is duplicated inside one file.** `packages/cli/src/shared/project-validation.ts` already had `isVendureProjectDirectory()`, which reads `dependencies` + `devDependencies` and tests `dep.includes('@vendure/') || dep === 'vendure'`. The new `hasVendureDependency()` does the same job twenty lines below with `startsWith(VENDURE_PACKAGE_SCOPE)` and two new constants the older function does not use. `includes` and `startsWith` differ for a package like `my-@vendure/thing`. One predicate, two callers.

Related: the CLI now has four parent-walking loops (`resolveCliProjectRoot`, `findVendureProjectRoot`, `findNearestPackageDir`, `findContainingNodeModules`), each with its own `while (true) { ... parent === current ... }`. One `walkUp(startDir, predicate)` would carry all four.

**6. The scope objects are rebuilt, and `cli.json` re-parsed, on every call.** `getGlobalPluginScope()` calls `getGlobalPluginAllowlist(env)` (which calls `readGlobalCliConfig`) and then `readGlobalCliConfig(env)` again. `getPluginScopes()` runs afresh inside `discoverCliPlugins`, `resolveCliPlugins`, `getCliPluginScope` and `listInactiveCliPlugins`; `addPlugin()` alone builds them twice. Cheap on its own, but it compounds with finding 1.

**7. Tautological colour assertions** in `command-registry.spec.ts`: `expect(styleProjectMarker(colors)).toBe(colors.yellow('*'))` derives the expected value from the same helper the implementation calls, so it passes for any colour. Only the `createColors(false)` case asserts a literal. Assert the escape sequence or drop the two.

**8. `styleProjectMarker`/`styleProjectLegend` take a `colors` parameter no production call site passes.** `describeNode` and `addProjectLegend` both call them bare. Fix 7 and the parameter can go.

**9. Cross-scope shadowing can silence a working plugin.** `owningScopes()` gives a name to the last scope that lists it, so if a package is enabled in both and the project's copy fails its check (not a direct dependency, say), the global copy is never tried: "Listed in vendure.cli.plugins but is not a direct dependency" and no command, with a valid global copy sitting right there. Worth documenting next to the "project wins" rule at minimum.

**10. Test-harness duplication.** `stripAnsi` now exists in both `src/shared/__tests__/run-cli.ts` and `e2e/cli-test-utils.ts` with independent regexes. And `cli-global-plugins.e2e-spec.ts` does `const { installCliPluginFixture } = await import('./cli-test-utils');` inside a test while importing three other names from that module statically at the top.

### Question

`dev`, `build` and `migrate` are gated on cwd, but carry `--config <path>`, `--tsconfig <path>` and `--vite-config <path>`. The `codemod` comment suggests the distinction was thought through, but only for `--config`. Did `vendure migrate --config /elsewhere/vendure-config.ts` work from outside a project before this?

### Comments

Mostly disciplined. The `project-core.ts` rationale for loading the project's `@vendure/core` is exactly the kind of comment that saves someone an hour. The misses are all one kind, arguing with a reviewer in the source file:

- `effectiveRequiresProject()` in `cli-command-definition.ts` is `node.requiresProject ?? inherited` under a seven-line comment ending "two copies would be two things to change if the rule ever grows a third state". Keep the function, cut the defence.
- Change narration with nothing to anchor it: "The case that made the global scope wrong", "The case the original implementation got wrong", "Reporting those packages as machine-wide listed every project dependency twice". A reader in a year has no "original implementation". This is PR-description material, and the description already has it almost verbatim.
- `codemod/command.ts` spends five lines explaining why a field is absent and contrasting it with `--config` on three other commands. One sentence does it.
- `loadInto()` has a branch commented "The check above already passed, so this cannot happen" plus three lines defending it.
- Docs are solid. Minor: "It is deliberately loose: it separates 'not in a project at all' from 'in a project', and does not promise that a given command can succeed" defines the check by what it is not.

One type-side note, nothing to fix today: `requireProjectCore()` is typed `typeof import('@vendure/core')` from the CLI's own devDependency, so a project pinned to an older core is typed as if it exported everything the CLI's copy does.

michaelbromley:
Thanks Markus — all four blockers reproduced and fixed in b88d0551f.

**1. Hoisted workspace node_modules read as global.** Reproduced with your fixture: `globalNodeModules` came back as the repo root. This repository is itself the failing case — its root manifest has `workspaces` and no Vendure dependency.

On the two suggested signals: requiring no `package.json` in *any* ancestor rejects a real nvm install, because nvm ships one at the root of its own directory (verified on my machine: `~/.nvm/package.json` sits above `~/.nvm/versions/node/v24.15.0/lib/node_modules`). Comparing against npm's global prefix works — derived from `process.execPath` it matches `npm root -g` exactly, no subprocess — but it only recognises npm-shaped layouts.

What both were reaching for is narrower: **the directory that holds the node_modules has no package.json**. Every package manager puts `node_modules` beside the manifest it installs for, so a project always has one there, including a workspace root. A global root never does. One `fs.existsSync`. Tests cover the workspace root, the nvm-style ancestor manifest, a plain project and a source checkout.

**2. `plugins remove --global` reporting false success.** Reproduced — it logged `Disabled` and wrote an unchanged file. It now refuses: "Package X is enabled by the VENDURE_CLI_PLUGINS environment variable, which this command cannot change. Unset VENDURE_CLI_PLUGINS, or remove X from its value." Exit 1, with an e2e covering it. The unit test that froze the old behaviour is gone.

**3. Two definitions of "in a project".** Confirmed — in an unrelated npm package directory the CLI chose the project scope. `resolveTargetScope` now uses `findVendureProjectRoot`, same as the gate, the legend and the docs. `resolveCliProjectRoot` still picks *which* manifest holds the config once a project is established, which is a different question and its existing rule is right for it.

**4. prettier.** All eight files formatted. I left `config-check.ts` and `production-check.ts` alone since they were already failing on `minor`, as you noted — reformatting them would add unrelated hunks.

Also done: one `hasVendureDependency` predicate with `isVendureProjectDirectory` as a caller (the `includes` vs `startsWith` difference was real); a `walkUp` helper carrying three of the four loops; `readGlobalCliConfig` read once per scope build; colour assertions rewritten against literal escape sequences via a shared `ANSI` constant; `stripAnsi` moved to one module both harnesses import; the dynamic import in the e2e spec made static; cross-scope shadowing documented next to the "project wins" rule; and the narrating comments cut.

On point 8 — the `colors` parameter stays, because the tests now pass `createColors(true)` to force colour rather than depending on whether the suite runs in a TTY. Same shape as `styleHelpTitle`, which production also calls without it.

**Your question.** No — `vendure migrate --config /elsewhere` did not work from outside a project before this PR. `migrate` already called `validateVendureProjectDirectory()` on `minor`, which requires both a Vendure dependency *and* a `vendure-config` file in cwd. The new gate is strictly looser than what it already enforced. `dev`, `build` and `start` resolve the project from cwd via `resolveVendureProjectDirectory`, falling back to cwd, so they were cwd-dependent too — the gate makes that explicit instead of failing later. No behaviour lost; I have not widened it to honour the path options, which would be a separate change.

Noted and not changed: `requireProjectCore()` being typed from the CLI's own devDependency. Real, and not fixable without either bundling types per version or widening the return type to something useless.

678 unit and 96 e2e pass under CI's own conditions (`CI=true FORCE_COLOR=1`); typecheck clean; no new lint errors.

packages/cli/src/commands/doctor/checks/database-check.ts
```diff
@@ -1,4 +1,4 @@
-import { RuntimeVendureConfig } from '@vendure/core';
+import type { RuntimeVendureConfig } from '@vendure/core';
```
dlhck, line 1: [P1] Load TypeORM lazily from the project too

Changing the core import to a type import leaves the runtime `typeorm` import on the next line. TypeORM is not a production dependency of `@vendure/cli`, so importing `doctor.ts` from an isolated CLI installation fails with `Cannot find module 'typeorm'` before any check runs. This also breaks `vendure doctor --check project` outside a project, where the PR promises a diagnostic report. I reproduced the import failure with compiled CLI files and only its production dependencies linked. Resolve TypeORM from the project inside the database check, so checks that do not need a database can run without it. The global-install fixture currently links the entire repository node_modules and hides this failure.

packages/cli/src/commands/schema/generate-schema/generate-schema.ts
```diff
@@ -1,41 +1,43 @@
...
         const apiType = options.api === 'shop' ? 'shop' : 'admin';
-        const typePaths = apiType === 'shop' ? VENDURE_SHOP_API_TYPE_PATHS : VENDURE_ADMIN_API_TYPE_PATHS;
+        const typePaths =
+            apiType === 'shop' ? core.VENDURE_SHOP_API_TYPE_PATHS : core.VENDURE_ADMIN_API_TYPE_PATHS;
 
-        const runtimeConfig = await runPluginConfigurations(getConfig() as any);
+        const runtimeConfig = await core.runPluginConfigurations(core.getConfig() as any);
+        const { GraphQLTypesLoader } = requireFromProject<{
+            GraphQLTypesLoader: new () => GraphQLTypesLoaderType;
+        }>('@nestjs/graphql');
```
dlhck, line 44: [P1] Resolve GraphQL from the same project as the schema

Core and the types loader come from the project, but `printSchema`, `graphqlSync`, and `getIntrospectionQuery` still come from the CLI's module-level `graphql` import. An isolated global CLI cannot import this implementation because GraphQL is not a production dependency. Supplying GraphQL beside the CLI does not fix the separate-copy case: with two physical copies of the same GraphQL version, I reproduced `printSchema(projectSchema)` throwing `Cannot use GraphQLObjectType ... from another module or realm`, and `graphqlSync` rejecting the project's GraphQLSchema. Both output formats fail. Load these GraphQL functions through project-scoped resolution as well, and cover a CLI installation with dependencies separate from the project's.

packages/cli/src/shared/cli-global-plugin-config.ts
```diff
@@ -0,0 +1,148 @@
...
+    const configPath = getGlobalCliConfigPath(env);
+    if (!fs.existsSync(configPath)) {
+        return { config: {}, path: configPath };
+    }
+    try {
+        const parsed = fs.readJsonSync(configPath) as GlobalCliConfig;
+        if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+            return { config: {}, path: configPath, error: 'Expected a JSON object' };
+        }
+        return { config: parsed, path: configPath };
```
dlhck: [P2] Validate the global config fields before returning them

The outer-object check accepts values that the scope builder cannot consume. Actual calls to `resolveCliPlugins()` with `cli.json` containing `{"plugins":{}}` throw `object is not iterable`; `{"pluginRoots":"/opt/plugins"}` throws `.map is not a function`. Arrays containing non-string entries can also throw during resolution. These exceptions escape before command registration, so even help and plugin-management commands stop working. Validate both optional fields as arrays of strings and report invalid fields through the existing scope-error path.

packages/cli/src/shared/resolve-cli-plugins.ts
```diff
@@ -151,74 +214,137 @@ export function resolveCliProjectRoot(cwd: string = process.cwd()): string {
...
-                        ...entry,
-                        status: 'failed',
-                        reason: e?.message ?? String(e),
-                    });
-                }
-            }
-        }
+        discovered.set(
+            packageName,
+            validate && entry.status !== 'failed' ? withLoadedCommands(entry, packageName) : entry,
```
dlhck: [P2] Respect project ownership during discovery validation

`discoverCliPlugins({ validate: true })` validates each scope independently and imports both copies when a package is enabled globally and in the project. `vendure plugins --json`, text listing, and the interactive manager all use this path. I reproduced this with two plugin modules that record imports: startup through `resolveCliPlugins()` records only `project`, then discovery also records `global`. The overridden global module's initialization therefore executes inside the project, contrary to the documented guarantee that only the project's pinned copy loads. Keep both scope records for management, but apply the same ownership rule as startup before importing modules.

dlhck (changes requested):
## Standards

No material findings.

## Spec

Four findings require changes: two failures in globally installed built-in commands, invalid global config fields that stop CLI startup, and plugin discovery that executes an overridden global copy. Details and reproductions are in the inline comments.

Validation: 37 tests passed across project-validation, cli-global-plugin-config, and resolve-cli-plugins-scopes. Separate runtime checks reproduced all four findings. The full local suite was inconclusive because the borrowed dependency tree differs from this checkout and sandbox restrictions blocked local server tests. GitHub CI passes for b88d0551fa0a6ebe96dd8c789a5f5216fb5247de.

Standards: 0 material findings. Spec: 4 findings, with global doctor and schema failures at P1.

michaelbromley:
Thanks David — all four reproduced and fixed in b50fc3edd.

**1. TypeORM (P1).** Confirmed: `typeorm` is imported at module level in `database-check.ts` and is not a CLI dependency. Now loaded through `requireFromProject` inside the database check, so the checks that do not need a database run without it.

**2. GraphQL (P1).** Confirmed, and the realm half is the sharper problem — the schema is built by the project's copy, so printing it with the CLI's copy fails even at identical versions. `printSchema`, `graphqlSync` and `getIntrospectionQuery` now come from the project, alongside core and the types loader.

**3. Invalid config fields (P2).** Reproduced exactly: `{"plugins":{}}` killed `vendure --version` with `object is not iterable`. Both optional fields are now validated as arrays of strings and reported through the scope-error path:

```
Could not read /path/cli.json: Expected "plugins" to be an array of strings
No CLI plugins were loaded from it. Fix /path/cli.json, or delete it to start over.
3.7.2
```

**4. Ownership during discovery (P2).** Reproduced with your approach — two modules recording their own import. Discovery now applies the same ownership rule as startup before importing, while still listing both rows for management:

```
after resolveCliPlugins : PROJECT
after discoverCliPlugins: PROJECT
rows listed             : @example/dup@global=enabled, @example/dup@project=enabled
```

**Your point about the fixture was the important one.** It linked the repository's whole `node_modules`, so it could only ever prove the CLI *finds* a package, never that a global installation ships it — which is exactly why both P1s got through. It now links only the packages `@vendure/cli` declares, each to its real location so transitive dependencies still resolve. There is a test guarding the fixture itself (`commander` present, `typeorm`/`graphql`/`@vendure/core` absent), so if it ever regains them the guard fails rather than the suite quietly going blind again.

New e2e from an isolated installation: `doctor` produces a report, `schema` reports a missing project dependency rather than failing to import, and an invalid config leaves `--version` working.

**One more of the same kind, which the review did not reach.** Taking the general rule — any runtime import of a package not in `dependencies` is unreachable from a standalone install — I audited every import in `src` against `package.json`. It found `semver`, imported at runtime by `config-check.ts` and declared nowhere, with no declared dependency pulling it in; it resolves today only by monorepo hoisting. Unlike core, TypeORM and GraphQL it has no coupling to the project's runtime — it compares version strings — so the fix is for the CLI to ship it, and it is now in `dependencies`.

682 unit and 100 e2e pass under CI's conditions; typecheck clean; no new lint errors.

packages/cli/src/commands/doctor/checks/database-check.ts
michaelbromley: Confirmed and fixed in b50fc3edd. `typeorm` was imported at module level and is not a dependency of `@vendure/cli`, so an isolated installation failed on import before any check ran.

`DataSource` is now `import type`, and the value comes from `requireFromProject("typeorm")` inside the database check itself, so the checks that need no database run without it.

You were right about the fixture hiding it — it linked the repository's whole `node_modules`. It now links only the packages the CLI declares, and there is a guard test asserting `typeorm`, `graphql` and `@vendure/core` are *absent* from it, so the suite fails rather than going blind if that ever regresses. New e2e: `vendure doctor` from an isolated global install produces a report with no `Cannot find module`.

packages/cli/src/commands/schema/generate-schema/generate-schema.ts
michaelbromley: Confirmed and fixed in b50fc3edd. `printSchema`, `graphqlSync` and `getIntrospectionQuery` now come from `requireFromProject("graphql")`, alongside core and the types loader.

The separate-copy point is the one I would have missed on my own, so thank you for spelling it out. Supplying GraphQL beside the CLI looks like a fix and is not: the schema handed to these functions was built by the project's copy, so two physical copies of the same version still fail with "Cannot use GraphQLObjectType from another module or realm". Loading them from the project puts them in the same realm as the schema, which is the actual requirement — the missing dependency was only the symptom that surfaced first.

Covered by the isolated-install e2e, which asserts `schema` reports a missing project dependency rather than failing to import.

packages/cli/src/shared/cli-global-plugin-config.ts
michaelbromley: Confirmed and fixed in b50fc3edd. Both of your payloads reproduced exactly — `{"plugins":{}}` killed `vendure --version` with `object is not iterable`, and `{"pluginRoots":"/opt/plugins"}` with `.map is not a function`.

`plugins` and `pluginRoots` are now validated as arrays of strings, including non-string entries, and reported through the scope-error path you pointed at:

```
Could not read /path/cli.json: Expected "plugins" to be an array of strings
No CLI plugins were loaded from it. Fix /path/cli.json, or delete it to start over.
3.7.2
```

The version still prints, which is the point — a broken config must not take out the commands needed to fix it. That was already the stated intent for an unparseable file; a wrongly shaped one just was not covered. Unit tests for all three shapes, plus an e2e asserting `--version` still works.

packages/cli/src/shared/resolve-cli-plugins.ts
michaelbromley: Confirmed and fixed in b50fc3edd. I reproduced it your way, with two plugin modules that append to a log on import:

```
before:  after resolveCliPlugins : PROJECT
         after discoverCliPlugins: PROJECT, GLOBAL

after:   after resolveCliPlugins : PROJECT
         after discoverCliPlugins: PROJECT
         rows listed             : @example/dup@global=enabled, @example/dup@project=enabled
```

`discoverCliPlugins` now computes the same `owningScopes` map as startup and passes each scope a `mayLoad` predicate, so only the scope that startup would load from may import a module. Both rows are still listed, so `vendure plugins` can show a package is enabled in both places — which was your "keep both scope records for management" requirement.

Worth noting the guarantee was documented but only enforced on one of the two paths, so the docs were describing behaviour that the management commands did not have.

PR #5404, opened 2026-09-17
test(create): cover the interactive prompt paths

### LINKED ISSUE #5352: test(create): no automated coverage of the interactive scaffolder paths
### Summary

Every automated test of `@vendure/create` runs the scaffolder with `--ci`. `test_quickstart_postgres` and all three `test_package_managers` jobs pass `--ci`, and the unit tests drive `getCiConfiguration` directly. Nothing exercises the Quick Start or Manual paths, which are the ones a human uses.

Those two paths are where the interactive prompts live: Quick Start prompts twice, Manual prompts twelve times.

### What this let through

PR #5338 wrapped configuration generation in a `@clack/prompts` spinner. Both interactive modes prompt from inside that call, so the spinner was running while the user answered.

`spinner().start()` sets an 80ms interval writing `cursor.move(-999,0)` + `erase.down(1)`, and calls `core.block()`, which enables raw mode, opens a second readline interface on the same TTY, and installs a keypress handler that calls `process.exit(0)` on Ctrl+C before `checkCancel` sees the cancel symbol.

Driven under a PTY:

| | Ctrl+C exit code | outcome |
|---|---|---|
| without the spinner | 1 | cancelled cleanly via `checkCancel` |
| with the spinner | 0 | prompt erased by 15 spinner frames, `checkCancel` never runs |

The change passed review and a fully green CI. It was caught in a merge-lane session and fixed before merge, but only because someone drove the interactive path by hand.

### Suggested fix

A PTY-driven smoke test for Quick Start and Manual: spawn the CLI under a pseudo-terminal, answer the prompts, and assert the generated project. The verification above used `script` to allocate the PTY, so the mechanism is available without a new dependency.

Covering Ctrl+C is worth doing at the same time, since cancellation is what broke and it fails silently — exit code 0 looks like success.

biggamesmallworld:
# Description

Every automated test of `@vendure/create` runs with `--ci`, so nothing covered the Quick Start and Manual prompt paths. That gap let the #5338 spinner-around-prompts bug pass review and CI.

This adds `packages/create/src/interactive-prompts.spec.ts`. It imports the CLI entry point with `@clack/prompts`, `commander`, the logger, port probing and Docker detection stubbed, and `process.exit` replaced by a spy. The fake spinner counts running spinners, and the fake prompts record whether a spinner was running when each prompt was shown.

- **Quick Start:** mode, Docker fallback and storefront prompts, then cancel at the storefront prompt.
- **Manual (Postgres):** all 13 prompts, then cancel at the storefront prompt.
- **Cancel at the first prompt:** nothing runs after "Setup cancelled." and the exit.

Each test asserts that no prompt was shown while a spinner ran, and that the run ends with `checkCancel`'s cancel message followed by its exit.

No production code changes and no new dependencies. The spec runs in about 0.3s in the existing `unit tests` job and needs no built `lib/`, no Docker and no terminal.

Fixes #5352

# Testing

- `bunx vitest run` in `packages/create`: 74 passed (71 existing, 3 new), stable across repeated runs and with `CI=true`.
- With the #5338 spinner put back around config generation, the Quick Start and Manual tests fail and list the prompts shown under the spinner.
- With `process.exit(0)` removed from `checkCancel`, all 3 tests fail.
- `tsc --noEmit` is clean. ESLint crashes locally in `prefer-promise-reject-errors`, the same way on the existing `scaffolded-config-storefront-port.spec.ts`.

# Notes for reviewers

- **Differences from the issue:** the issue suggests a PTY test that also checks the generated project. This spec drives the prompt logic in-process instead, because a real scaffold needs network access and takes minutes. The generated config is already covered by `scaffolded-config-storefront-port.spec.ts`.
- **Exit code:** the issue's table says Ctrl+C exits with 1 without the spinner. On `master`, `checkCancel` calls `process.exit(0)`, and a PTY run of the built CLI confirmed exit 0 both with and without the spinner. The spec asserts the current `0`, so it detects a missed cancel by the event order ("Setup cancelled." then exit), not by the exit code. Switching cancellation to a non-zero exit is a separate behaviour change, and these assertions would need updating with it.
- **Separate bug found:** in `getQuickStartConfiguration`, `checkCancel(useSqlite)` checks the wrong variable. It should be `useSqliteResponse`. Ctrl+C at the Docker fallback prompt is ignored. This PR does not fix it.

# Breaking changes

None.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5404"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792226105&installation_model_id=20193&pr_number=5404&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5404&signature=b9129c100cce9f87878e1082494197a4ad569e708155db1d9d048e50c6025cbb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld:
Merge lane session `83fba0e9` (@biggamesmallworld) is working this pull request.

Current stage: Stage 0.5: claimed, waiting for a slot. Claim expires 2026-09-18T19:17:32Z unless refreshed.

PR #5410, opened 2026-09-19
fix: resolve dashboard package root by package.json instead of fixed depth (#4053)

### LINKED ISSUE #4053: Adding "../../../" to the vite output path breaks configuration in NX
This line https://github.com/vendure-ecommerce/vendure/blob/fc3cea8675dd8056211f19c2e41816b611375776/packages/dashboard/vite/vite-plugin-vendure-dashboard.ts#L286 doesn't allow correctly building the application within an NX workspace as it always moves the build out of the workspace.

And this means that all environments need the same 3 top level folders for this to work properly.

wakqasahmed:
# Description

Closes #4053

`getDashboardPackageRoot()` resolves `@vendure/dashboard`'s entry file via `import.meta.resolve` and then walks up a hardcoded `../../../` to get the package root. That only works when the entry resolves to `src/lib/index.ts`, which is this monorepo's own workspace-linked source path — 3 levels below the package root. Under an NX workspace (or any setup where the package resolves to a different depth, e.g. a published build), the fixed walk-up lands outside the package, breaking the dashboard build output path.

Replaced the fixed-depth walk-up with `findPackageRoot()`, which walks up from the resolved entry file looking for a `package.json` whose `name` is `@vendure/dashboard`, falling back to the old fixed-depth guess only if that search comes up empty.

# Breaking changes

None — same function signature and return value in the normal case, just resolved more robustly.

# Screenshots

N/A, build-tooling fix.

# Checklist

📌 Always:

👍 Most of the time:
- [x] I have added or updated test cases — added `packages/dashboard/vite/tests/dashboard-package-root.spec.ts` covering the 2-level and 3-level entry-depth cases, an unrelated-nested-`package.json` case, and the fallback path. Ran via vitest, all green.
- [ ] README update — not applicable, this is an internal implementation detail with no documented API change.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/vendure/pr/5410?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792381236&installation_model_id=20193&pr_number=5410&ref=codesmith_pr_footer&repository=vendurehq%2Fvendure&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fvendure%2Fpull%2F5410&signature=20ff8d17480775b882bb33c85325eaa12fc4989d4c55f15029a96ccd51bc6f32"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wakqasahmed (commented):
Reviewed cold against the linked issue (#4053) and the current source at `packages/dashboard/vite/vite-plugin-vendure-dashboard.ts`. Findings below; no blockers.

**1. Root cause — confirmed accurate.** `packages/dashboard/package.json`'s `main`/`module`/`exports["."]` all point to `./src/lib/index.ts`, which is exactly 3 path segments below the package root (`src` → `lib` → `index.ts`), matching the old `path.join(packagePath, '../../../')` assumption. I also pulled the file as of the commit referenced in #4053 (fc3cea86) and confirmed `getDashboardPackageRoot()` at that revision is the same function this PR touches — the fix is at the right spot, not a look-alike location.

**2. pnpm/npm symlink concern — non-issue.** This monorepo uses npm workspaces (root `package.json` `"workspaces": ["packages/*"]`, no `pnpm-workspace.yaml`), so `node_modules/@vendure/dashboard` is a plain symlink to `packages/dashboard`. For a downstream pnpm consumer, `import.meta.resolve` would resolve through the symlink into pnpm's `.pnpm` content-addressable store — but that store mirrors the package's own file tree verbatim (its `package.json` still sits at the root of that store entry), so `findPackageRoot`'s ancestor walk still finds a `package.json` named `@vendure/dashboard` at the expected relative depth. Not a real gap.

**3. 10-iteration limit — safe, but for a different reason than node_modules nesting.** The walk only needs to traverse from the resolved entry file up to *that package's own* root — e.g. `src/lib/index.ts` (2 hops) or a nested `dist/esm/index.js` build (a few more). It's bounded by how deep the entry sits *inside* the package's own output structure, not by how deeply `node_modules` itself is hoisted/nested in the host project (nesting elsewhere in the tree doesn't add hops between the entry file and its own package root). Given that, 10 is comfortably generous for any real published-package layout.

**4. New tests — correct, one minor note.** The 2-level/3-level/fallback cases are correct and match real entry-depth scenarios (`dist/index.js` vs `src/lib/index.ts`). The "unrelated nested `package.json`" case (test 3) is a slightly contrived setup as literally described — in practice `import.meta.resolve('@vendure/dashboard')` wouldn't resolve into `node_modules/some-other-dep`'s own subtree if resolution is otherwise working — but the property it exercises (must match by exact `name`, keep walking past a mismatched ancestor `package.json`) is a real and worthwhile invariant to lock in, so I'd keep it.

**5. Style/type check.** Diff reads clean against this repo's root `.eslintrc.js`: `import/order` is satisfied (new `node:fs` import is alphabetically placed between `node:crypto` and `node:url`), and the comment-only `catch {}` block is exempt from `no-empty: error` per ESLint's documented "block contains a comment" exception. Sync `fs.existsSync`/`readFileSync` usage matches the surrounding synchronous plugin-setup code. No obvious type-safety issues beyond the existing project-wide `no-unsafe-*` opt-outs.

Minor, non-blocking observation: `findPackageRoot` is now a named export reachable via the package's `./vite` subpath export, so it becomes part of `@vendure/dashboard`'s public surface rather than staying test-internal. Probably fine for a small utility, just flagging in case that wasn't deliberate.

Overall: root cause is correctly diagnosed and located, the fix is a strict improvement over the old fixed-depth guess (same result in the common case, correct in the previously-broken NX/deep-entry case), and the fallback preserves old behavior when no match is found. No functional or correctness issues found.

